PR #33733, opened 2019-03-09
DataZoom.Slider: fillColor must be fillerColor

tmtron:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://ecomfe.github.io/echarts-doc/public/en/option.html#dataZoom-slider.fillerColor>>

DataZoom.Slider: the correct property is `fillerColor` (not `fillColor`)


bilalucar (approved):
I checked the echarts documentation. Everything seems right.

Ovilia (approved):
Checked and approved.

PR #33741, opened 2019-03-09
[reserved-words] add type definitions for the reserved-words package

skovy:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/reserved-words/index.d.ts
+++ b/types/reserved-words/index.d.ts
@@ -1,0 +1,36 @@
+// Type definitions for reserved-words 0.1
+// Project: https://github.com/zxqfox/reserved-words#readme
+// Definitions by: Spencer Miskoviak <https://github.com/skovy>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+declare namespace reserved {
+    type Dialect =
+        | "es3"
+        | 3
+        | "es5"
+        | 5
+        | "es2015"
+        | 6
+        | "es7"
+        | 7
+        | "es6"
+        | "next";
+}
+
+interface Reserved {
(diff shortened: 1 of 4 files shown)


plantain-00 (approved):

types/reserved-words/index.d.ts
```diff
@@ -0,0 +1,36 @@
+// Type definitions for reserved-words 0.1
+// Project: https://github.com/zxqfox/reserved-words#readme
+// Definitions by: Spencer Miskoviak <https://github.com/skovy>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+declare namespace reserved {
```
SimonSchick: I don't think we need a namespace here since `module.exports` isn't written directly it's assigned but the only member is `check`.

eg. you can `export` everything directly.

```
export type Dialect = ...`;

expect function check...
```

Also you missed the `KEYWORDS` export.

types/reserved-words/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+        "noImplicitThis": true,
+        "strictNullChecks": true,
+        "baseUrl": "../",
+        "typeRoots": [
+            "../"
+        ],
+        "types": [],
+        "noEmit": true,
+        "forceConsistentCasingInFileNames": true,
+        "esModuleInterop": true,
```
SimonSchick: This won't be needed with my suggested changes.

SimonSchick (changes requested):

skovy:
Thanks for the feedback @SimonSchick! I addressed your comments, let me know if I missed anything 👍

SimonSchick (approved):

PR #32694, opened 2019-02-01
feat(graphql): add TArgs generic to more graphql types

mc0:
- [X] Use a meaningful title for the pull request. Include the name of the package modified.
- [X] Test the change in your own code. (Compile and run.)
- [X] Add or edit tests to reflect the change. (Run with `npm test`.)
- [X] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [X] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [X] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).
- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://graphql.org/graphql-js/type/#graphqlobjecttype
- [n/a] Increase the version number in the header if appropriate.
- [n/a] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

This builds off of some changes from #22789 and continues having TArgs be optional. This is not a breaking change. It's possible that `unknown` is a better option for TArgs; however, that would be TS3 specific and be a breaking change.

This change is to allow overriding TArgs in more places. For example, allowing `any` args or to restrict args further. This primarily impacts the `GraphQLFieldConfigMap` and `GraphQLObjectType` types.

Using `GraphQLFieldConfigMap` with generics:
```js
const Query = new GraphQLObjectType({
  fields: (): GraphQLFieldConfigMap<void, Context, any> => ({
    someField: getFieldConfig(),
  })
});
```

Using `GraphQLObjectType` with generics:
```js
const Query = new GraphQLObjectType<void, Context, any>({
  fields: () => ({
    someField: getFieldConfig(),
  })
});
```

PR #33743, opened 2019-03-10
Type check GraphQL args

mike-marcacci:
Having `TArgs` set to `{ [key: string]: any }` in the types produces errors when trying to refine it to a specific type, such as `{id: string}`.

image

- [ ] **N/A** Provide a URL to documentation or source code which provides context for the suggested changes: 
- [ ] **N/A** Increase the version number in the header if appropriate.
- [ ] **N/A** If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

### DIFF
--- a/types/graphql-relay/graphql-relay-tests.ts
+++ b/types/graphql-relay/graphql-relay-tests.ts
@@ -45,7 +45,7 @@ forwardConnectionArgs.first = { type: GraphQLInt };
 backwardConnectionArgs.before = { type: GraphQLString };
 backwardConnectionArgs.last = { type: GraphQLInt };
 // connectionDefinitions returns a connectionType and its associated edgeType, given a node type.
-const resolve: GraphQLFieldResolver<any, any> = (source, args, context, info) => {
+const resolve: GraphQLFieldResolver<any, any, any> = (source, args, context, info) => {
     context.flag = "f";
 };
 const fields: GraphQLFieldConfigMap<any, any> = {};
@@ -124,8 +124,8 @@ const idFetcher = (id: string, context: any, info: GraphQLResolveInfo) => {
     context.flag = "f";
 };
 const nodeDef = nodeDefinitions<any>(idFetcher, resolver);
-const nodeFieldConfig: GraphQLFieldConfig<any, any> = nodeDef.nodeField;
-const nodesFieldConfig: GraphQLFieldConfig<any, any> = nodeDef.nodesField;
+const nodeFieldConfig: GraphQLFieldConfig<any, any, any> = nodeDef.nodeField;
+const nodesFieldConfig: GraphQLFieldConfig<any, any, any> = nodeDef.nodesField;
 const interfaceType: GraphQLInterfaceType = nodeDef.nodeInterface;
 // toGlobalId takes a type name and an ID specific to that type name, and returns a "global ID" that is unique among all types.
 toGlobalId("t", "i").toLowerCase();
@@ -137,10 +137,10 @@ fgi.type.toUpperCase();
 const idFetcher2 = (object: any, context: any, info: GraphQLResolveInfo) => {
     return "";
 };
-const gif: GraphQLFieldConfig<any, any> = globalIdField("t", idFetcher2);
+const gif: GraphQLFieldConfig<any, any, any> = globalIdField("t", idFetcher2);
 // pluralIdentifyingRootField creates a field that accepts a list of non-ID identifiers (like a username) and maps them to their corresponding objects.
 const input: GraphQLInputType = GraphQLString;
-const prf: GraphQLFieldConfig<any, any> = pluralIdentifyingRootField({
+const prf: GraphQLFieldConfig<any, any, any> = pluralIdentifyingRootField({
     argName: "a",
     inputType: input,
     outputType: input,
--- a/types/graphql-relay/index.d.ts
+++ b/types/graphql-relay/index.d.ts
@@ -246,14 +246,14 @@ export interface MutationConfig {
  */
 export function mutationWithClientMutationId(
     config: MutationConfig
-): GraphQLFieldConfig<any, any>;
+): GraphQLFieldConfig<any, any, any>;
 
 // node/node.js
 
 export interface GraphQLNodeDefinitions {
     nodeInterface: GraphQLInterfaceType;
-    nodeField: GraphQLFieldConfig<any, any>;
-    nodesField: GraphQLFieldConfig<any, any>;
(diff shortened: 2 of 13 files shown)


mike-marcacci:
Also... I'm not sure where I'd go to propose this, since it would be a breaking change, but the order of `TContext` and `TArgs` should *definitely* be changed, since they are switched everywhere that they appear as arguments, which is quite confusing. Is there a way to make an issue and just tag a particular package?

mc0:
@mike-marcacci I think if, in your example, you set the `GraphQLQueryType` (not shown) `TArgs` to `any` you’ll get the desired result without this change. I had made a change to allow for this concept and contemplated using `any` for `TArgs` but using `any` is discouraged in this repo.
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32694

As far as switching `TArgs` and `TContext`: it looks like the the `GraphQLTypeResolver` implementation doesn’t take `TArgs` at all. That may have been the rational for the order but it is unfortunate.

mike-marcacci:
Hi @mc0 thanks for the reply. I don't see a `GraphQLQueryType` anywhere in the codebase, perhaps you meant doing this, which is my current workaround:

```ts
   new GraphQLObjectType<any, any, any>({
        name: "Query",
        fields: () => ({
            hello: {
                type: GraphQLString,
                args: {
                    id: {
                        type: GraphQLString
                    }
                },
                resolve: (source, context, args: {id: null | "string"}) => {
                    return "world";
                }
            }
        })
    })
```

I quite dislike this behavior, since it forces me to explicitly cast through `any` wherever args are used. It's not even possible to specify types on the `GraphQLObjectType` directly, as they are still incompatible with `{[key: string]: any}`:

```ts
// This has the same error:
new GraphQLObjectType<any, any, {
  id: {
    type: GraphQLString
  }
}>
```

At the end of the day, I don't think `{[key: string]: any}` is buying us much, and it makes it a bit more difficult to strongly type the args.

mc0:
My mistake for referencing the wrong name. You can define each of your fields as a more specific `GraphQLFieldConfig` instead of the object in your example. That way the `any` on the GraphQLObectType` doesn’t lose you any safety.

bradzacher:
Correct me if I'm wrong, but in reality there are a few of problems with having a top level `TArgs` on `GraphQLObjectType`, which makes a generic at the top level kind of silly for anything but the simplest of `GraphQLObjectType`s.

- `GraphQLObjectType` does not directly have arguments on it.
- Each `GraphQLFieldConfig` can (and probably will) have a different set of arguments on it.

I.e. this super simple case fails the generic usage:
```TS
new GraphQLObjectType<any, any, { id: string }>({
	fields: () => ({
        one: {
            // matches the generic
            args: {
                id: { type: GraphQLString },
            },
        },
        two: {
            // does not match the generic
            args: {
                foo: { type GraphQLNumber },
            },
        },
    }),
});
```

mike-marcacci:
@bradzacher - I completely agree with that. I'll go ahead and remove it entirely from these wrapping types.

mike-marcacci:
OK, I've done a more thorough audit of `TArgs` here, as opposed to just running find/replace. I also switched the order of the types where they continue to exist, so that they match the argument order. With these changes, it's now possible to do things like this:

```ts
new GraphQLObjectType<FooModel, { db: SomeDatabaseClient }>({
    name: "Foo",
    fields: () => ({
        hello: {
            type: GraphQLString,
            args: {
                id: {
                    type: GraphQLString
                }
            },
            resolve: (source, context, args: {id: null | "string"}) => {
                return "world";
            }
        }
    })
})
```

where source, context and args are all strongly typed.

mike-marcacci:
OK, everything passes now – I deliberately kept my added parameters required, since their order is different. This is, in a very minor sense, breaking, but I believe well worth whatever small hassle it may cause.

mike-marcacci:
As a "next steps" it's probably possible to perform a shallow type check of `TArgs` against `GraphQLFieldConfigArgumentMap` using `keyof`...

mike-marcacci:
OK... so I added type checking of the args config. You can now do something like this:

```ts
    new GraphQLObjectType({
        name: "Query",
        fields: () : {
            hello: GraphQLFieldConfig<any, { message: null | string }, any>
        } => ({
            hello: {
                type: GraphQLString,
                args: {
                    message: {
                        type: GraphQLString
                    }
                },
                resolve: (source, args, context) => {
                    return args.message || args.message;
                }
            }
        })
    })
```
and it will ensure the presence of `message` in the args config. As I'm doing this, I'm realizing it's quite possible using conditional type expressions to traverse the graphql configs and ensure strict conformity to types for both arguments and return types. I'm going to have to leave this here, though, as this is sufficient for my needs and totally tangential to what I'm actually supposed to be doing today!

mc0:
All of these changes now look like they'd be pretty extensive breaking changes.

Here is an example of what is possible currently:
```
const Query = new GraphQLObjectType<void, Context, any>({
  fields: () => ({
    example: (): GraphQLFieldConfig<void, Context, { id: string }> => ({
      args: {
        id: {
          type: new GraphQLNonNull(GraphQLID),
        },
      },
      resolve: (_, args, context) => {
        return args.id;
      },
      type: GraphQLID,
    }),
  }),
  name: 'Query',
});
```

mike-marcacci:
Hi @mc0, in your example the following would have changed:
1. You can remove the useless third parameter from `GraphQLObjectType<void, Context, any>`
2. You can update the parameter order of `GraphQLFieldConfig<void, Context, { id: string }>` to match the actual argument order (becoming `GraphQLFieldConfig<void, { id: string }, Context>`).
3. The config passed to `args` is now checked to match the type of `TArgs`. It will error if it's missing an argument config, or if it defines additional ones not declared by the type settings. In the future, I or someone else can actually enforce that the configured GraphQL type of the arg matches the TypeScript type...

mike-marcacci:
OK, so I actually started implementing deep type checks on args. This ensures that the `args` configs exactly match the type definitions, which also apply to the resolver. This has already helped me catch instances where I configured GraphQL with nullable inputs, but wrote types assuming they would be non-null.

Currently, the [lack of recursion support](https://github.com/Microsoft/TypeScript/issues/6230) in conditionals means that list depth is limited to one, before losing type safety and becoming `any`. It is possible to descend into custom input objects to check matching configs, but I haven't implemented this yet.

This currently only works in TypeScript 2.8+, since that's when conditionals were introduced, hence the failing tests. I'm not really sure how to proceed here.

PR #33746, opened 2019-03-10
[tampermonkey] Tampermonkey 4.8

nikolay-borzov:
If adding a new definition:

### DIFF
--- a/types/tampermonkey/tslint.json
+++ b/types/tampermonkey/tslint.json
@@ -1,0 +1,1 @@
+{ "extends": "dtslint/dt.json" }
--- a/types/tampermonkey/index.d.ts
+++ b/types/tampermonkey/index.d.ts
@@ -1,0 +1,397 @@
+// Type definitions for non-npm package Tampermonkey 4.x
+// Project: https://tampermonkey.net
+// Definitions by: Steven Wang <https://github.com/silverwzw>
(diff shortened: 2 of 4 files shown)


types/tampermonkey/tslint.json
```diff
@@ -0,0 +1,6 @@
+{ 
+  "extends": "dtslint/dt.json",
+  "rules": {
+    "no-unnecessary-generics": false
```
weswigham: I don't see anything that looks like an unneeded generic - why's this lint rule need to be ignored?

nikolay-borzov: It's here
```
declare function GM_xmlhttpRequest<TContext = any>(
    details: Tampermonkey.Request<TContext>
): Tampermonkey.AbortHandle<void>;
```

Use case
```
interface RequestContext {
    form: {
        name: string
    };
}

GM_xmlhttpRequest<RequestContext>({
    url: 'http://example.com/',
    onloadstart(response) {
        GM_log(response.context.form.name);
    },
});
```

weswigham: I'd prefer to see an inline suppression on that instance than a blanket suppression in the `.json` file, if you could, then - that way future editors don't accidentally add more they don't need to.

nikolay-borzov: Sure, I'll add an inline suppression

PR #33755, opened 2019-03-11
Deprecate types for packages that provide their own types

BendingBender:
If removing a declaration:


SimonSchick (approved):
At this rate DT will have no packages left to maintain in a year 😅

tcaesvk:
@BendingBender is it same purpose as [33631{(https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33631)?

Hoishin (approved):

tcaesvk (approved):

BendingBender:
@tcaesvk Exactly. And it will not be the last.

BendingBender:
@SimonSchick That would actually be rather good, don't you think?

ragnarok56 (approved):

SimonSchick:
@BendingBender imagine node shipping with typings 🤔

BendingBender:
only in my dreams...

types/got/got-tests.ts
```diff
@@ -6,7 +6,6 @@ import * as fs from 'fs';
 import * as http from 'http';
 import * as https from 'https';
 import * as url from 'url';
-import QuickLRU = require('quick-lru');
```
ikokostya: Is it possible import this module from `npm`?

BendingBender: It's not compatible in the sense that it uses TS 3.0+ features. This would require to move this definiton to TS 3.0 just to support a test case. I think it's not worth it.

PR #33756, opened 2019-03-11
Adds missing type for a parameter createExternalAccount when adding a bank account.

### LINKED ISSUE #33754: [@types/stripe] createExternalAccount(): second parameter requires string when it can be dictionary

- [x ] I tried using the `@types/xxxx` package and had problems.
- [ ] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [ ] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [ ] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @wjohnsto 

In the call for createExternalAccount

```
createExternalAccount(accId: string, data: accounts.IExternalAccountCreationOptions, response?: IResponseFn<cards.ICard | bankAccounts.IBankAccount>): Promise<cards.ICard | bankAccounts.IBankAccount>;
```

The datatype IExternalAccountCreationOptions has an issue as per stripe docemenation and doc string in @types/stripe

it is: 
```
interface IExternalAccountCreationOptions extends IDataOptionsWithMetadata {
            /**
             * When adding a card to a customer, the parameter name is source. When
             * adding to an account, the parameter name is external_account. The
             * value can either be a token, like the ones returned by our Stripe.js, or a
             * dictionary containing a user’s credit card details (with the options shown
             * below). Stripe will automatically validate the card.
             */
            external_account: string ;

            /**
             * Only applicable on accounts (not customers or recipients). If you set this to true (or if this is the first external account being added
             * in this currency) this card will become the default external account for its currency.
             */
            default_for_currency?: boolean;
        }
```

But should be

```
interface IExternalAccountCreationOptions extends IDataOptionsWithMetadata {
            /**
             * When adding a card to a customer, the parameter name is source. When
             * adding to an account, the parameter name is external_account. The
             * value can either be a token, like the ones returned by our Stripe.js, or a
             * dictionary containing a user’s credit card details (with the options shown
             * below). Stripe will automatically validate the card.
             */
            external_account: string | IBankAccountHash ;

            /**
             * Only applicable on accounts (not customers or recipients). If you set this to true (or if this is the first external account being added
             * in this currency) this card will become the default external account for its currency.
             */
            default_for_currency?: boolean;
        }
```

Can only create a bank account using a string instead of the dictionary currently unless using 
```as any```

### PR DESCRIPTION
**Pull request by yharaskrik**
Added ```IBankAccount``` as a type for ```IExternalAccountCreationOptions.external_account``` as per the Stripe documentation and doc string in @types/stripe. This type should allow for a tokenized string or a dictionary of data to create the external_account with. Fixes #33754

### CODE REVIEW

File: types/stripe/index.d.ts [hunk 1]
```diff
@@ -436,7 +437,7 @@ declare namespace Stripe {
              * dictionary containing a user’s credit card details (with the options shown
              * below). Stripe will automatically validate the card.
              */
-            external_account: string ;
+            external_account: string | IBankAccount;
```
galtalmor: Maybe I'm missing something, but I think this is a different type.
`IBankAccount` is the type returned from `createExternalAccount`, while this object (`IExternalAccountCreationOptions`) is the object sent to this function, which has a different structure.
It seems like this type of object already exists, though not separated to its own interface in `IAccountUpdateOptions -> external_account`, so I would suggest extracting it and using it instead here.

yharaskrik: Oh that would make sense, I must have missed that but that does make sense because I realized there a disconnect between the create and update calls for some of the other methods as well. Like creating an account not all options are available in the CreationOptions interface for it so maybe there are more missing types than I thought. I will take a look and compare them against the docs and update this PR when I do.

PR #33764, opened 2019-03-11
Add maxHeadersCount and headersTimeout properties to HTTPS module in @types/node

tyronedougherty:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/api/https.html#https_server_headerstimeout

### DIFF
--- a/types/node/http.d.ts
+++ b/types/node/http.d.ts
@@ -101,11 +101,13 @@ declare module "http" {
 
         setTimeout(msecs?: number, callback?: () => void): this;
         setTimeout(callback: () => void): this;
-        maxHeadersCount: number;
+        // https://nodejs.org/api/http.html#http_server_maxheaderscount
+        maxHeadersCount: number | null;
         timeout: number;
         /**
          * Limit the amount of time the parser will wait to receive the complete HTTP headers.
          * @default 40000
+         * {@link https://nodejs.org/api/http.html#http_server_headerstimeout}
          */
         headersTimeout: number;
         keepAliveTimeout: number;
--- a/types/node/http.d.ts
+++ b/types/node/http.d.ts
@@ -101,7 +101,11 @@ declare module "http" {
 
         setTimeout(msecs?: number, callback?: () => void): this;
         setTimeout(callback: () => void): this;
-        // https://nodejs.org/api/http.html#http_server_maxheaderscount
+        /**
+         * Limits maximum incoming headers count. If set to 0, no limit will be applied.
+         * @default 2000
+         * {@link https://nodejs.org/api/http.html#http_server_maxheaderscount}
+         */
         maxHeadersCount: number | null;
         timeout: number;
         /**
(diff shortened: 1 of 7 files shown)


SimonSchick (changes requested):
Please add a quick test 😄

j-oliveras (changes requested):
Can you add `maxHeadersCount` to node v10?

See https://nodejs.org/docs/latest-v10.x/api/https.html#https_server_maxheaderscount

types/node/https.d.ts
```diff
@@ -26,7 +26,13 @@ declare module "https" {
 
         setTimeout(callback: () => void): this;
         setTimeout(msecs?: number, callback?: () => void): this;
+        maxHeadersCount: number;
```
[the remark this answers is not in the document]
Flarna: Should be `number | null`, see https://github.com/nodejs/node/blob/v11.11.0/lib/https.js#L76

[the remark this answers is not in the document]
tyronedougherty: Done :)

tyronedougherty:
I've updated the PR with the requested changes :)

SimonSchick (approved):

Flarna:
`headersTimeout` has been added in NodeJs 10.14 but v10/index.d.ts tells 10.12. Could you please update the first line in v10/index.d.ts to 10.14?

Not sure if we should really add links to source code as JsDoc here as this is usually not what you want to see during programming. But I will not block because of this.

BrunoScheufler (approved):

r3nya (approved):

j-oliveras (approved):

types/node/http.d.ts
```diff
@@ -101,11 +101,13 @@ declare module "http" {
 
         setTimeout(msecs?: number, callback?: () => void): this;
         setTimeout(callback: () => void): this;
-        maxHeadersCount: number;
+        // https://nodejs.org/api/http.html#http_server_maxheaderscount
```
SimonSchick: Uh, please use actual JSDoc eg. `/**` or just copy/paste the docs directly.

tyronedougherty: I was just copying what I've seen throughout the typings... in fact none of the links in that file use JSDoc except the one that I added :/

tyronedougherty:
> `headersTimeout` has been added in NodeJs 10.14 but v10/index.d.ts tells 10.12. Could you please update the first line in v10/index.d.ts to 10.14?

Done!

> Not sure if we should really add links to source code as JsDoc here as this is usually not what you want to see during programming. But I will not block because of this.

I see it throughout the typings here. I think it's ok :)

types/node/http.d.ts
SimonSchick: Yeea, docs are a little bit inconsistent but every bit helps!

Flarna (approved):

SimonSchick (approved):

PR #33771, opened 2019-03-11
@types/react-navigation Add state to NavigationContainerComponent

Titozzz:
Fails with:
```
../react/index.d.ts(29,22): error TS2307: Cannot find module 'csstype'.

```

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: Not found 😢 ..

### DIFF
--- a/types/react-navigation/index.d.ts
+++ b/types/react-navigation/index.d.ts
@@ -33,6 +33,7 @@
 //                 Kenneth Skovhus <https://github.com/skovhus>
 //                 Aaron Rosen <https://github.com/azrosen92>
 //                 Haseeb Majid <https://github.com/hmajid2301>
+//                 Thibault Malbranche <https://github.com/titozzz>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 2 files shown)


types/react-navigation/index.d.ts
```diff
@@ -845,7 +846,8 @@ export interface NavigationContainerProps<S = {}, O = {}> {
 
 export interface NavigationContainerComponent
   extends React.Component<
-  NavigationContainerProps & NavigationNavigatorProps<any>
+  NavigationContainerProps & NavigationNavigatorProps<any>,
+  { nav: NavigationState | null }
```
jeroenvervaeke: Wouldn't it be better to introduce a `NavigationContainerState` interface which looks like this?

```ts
interface NavigationContainerState {
    nav?: NavigationState;
}
```

jeroenvervaeke (changes requested):

types/react-navigation/index.d.ts
Titozzz: I just aligned on how was declared the state inside NavigationContainer. Do you want me to change only one, both or none?

jeroenvervaeke: In that case I'll leave that to you, but I would change
`nav: NavigationState | null`
to
`nav?: NavigationState`
for the people who have strictNullCheck enabled

PR #33773, opened 2019-03-11
Propagate event type to components in react-big-calendar

TomasHubelbauer:
Currently components which accept the event always have its type set to `object` even though the props are already generic because nowhere else in the types is their generic argument actually set.

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: **this is about typing generic arguments, not sure what to point to in this case**


eps1lon (commented):
Could you add a test to see if typescript is passing the generic argument down properly i.e. some test that's using `<Calendar<T> components={components} />`?

TomasHubelbauer:
@eps1lon I updated an existing test which used it.

eps1lon (approved):

PR #33784, opened 2019-03-11
Fix lodash test exhibiting bug fixed in ts@next

weswigham:
There was a bug where when we had a secondary inference location that we didn't use, we'd infer a literal type where we shouldn't have. This is now fixed in `ts@next`, so the `ExpectType` check is a bit too strict.


weswigham:
Lodash OOMs on travis, but on devops like we use for TS builds it runs OK. Neat. Anyways, this fixes the breaks locally, so I'm going to merge this - fixing lodash's memory issue is bigger and not something I think I can really get into right now~

PR #33785, opened 2019-03-11
Fix ramda evolvable type for ts@next

weswigham:
Fixes one of the new errors in ramda's tests, per [this comment](https://github.com/Microsoft/TypeScript/pull/30215#issuecomment-470715781). One of the other issues is now tracked on TS by https://github.com/Microsoft/TypeScript/issues/30324 - I suspect it's actually the same root cause for all 3 of the other test failures.


weswigham:
Merging this since it removes an error from ramda ❤️

PR #33787, opened 2019-03-12
[redux-actions] strict null checks

oddui:
Please fill in this template.

### DIFF
--- a/types/redux-actions/index.d.ts
+++ b/types/redux-actions/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for redux-actions 2.3
+// Type definitions for redux-actions 2.6
 // Project: https://github.com/redux-utilities/redux-actions
 // Definitions by: Jack Hsu <https://github.com/jaysoo>,
 //                 Alex Gorbatchev <https://github.com/alexgorbatchev>,
--- a/types/redux-actions/index.d.ts
+++ b/types/redux-actions/index.d.ts
@@ -67,9 +67,9 @@ export type Reducer<State, Payload> = (state: State, action: Action<Payload>) =>
 
 export type ReducerMeta<State, Payload, Meta> = (state: State, action: ActionMeta<Payload, Meta>) => State;
 
-export type ReduxReducer<State, Payload> = (state: State | undefined, action: Action<Payload>) => State;
+export type ReduxCompatibleReducer<State, Payload> = (state: State | undefined, action: Action<Payload>) => State;
 
-export type ReduxReducerMeta<State, Payload, Meta> = (state: State | undefined, action: ActionMeta<Payload, Meta>) => State;
+export type ReduxCompatibleReducerMeta<State, Payload, Meta> = (state: State | undefined, action: ActionMeta<Payload, Meta>) => State;
 
 /** argument inferring borrowed from lodash definitions */
 export type ActionFunction0<R> = () => R;
@@ -153,13 +153,13 @@ export function handleAction<State, Payload>(
     actionType: string | ActionFunctions<Payload> | CombinedActionType,
     reducer: Reducer<State, Payload> | ReducerNextThrow<State, Payload>,
     initialState: State
-): ReduxReducer<State, Payload>;
+): ReduxCompatibleReducer<State, Payload>;
 
 export function handleAction<State, Payload, Meta>(
     actionType: string | ActionWithMetaFunctions<Payload, Meta> | CombinedActionType,
     reducer: ReducerMeta<State, Payload, Meta> | ReducerNextThrowMeta<State, Payload, Meta>,
     initialState: State
-): ReduxReducerMeta<State, Payload, Meta>;
+): ReduxCompatibleReducerMeta<State, Payload, Meta>;
 
 export interface Options {
     prefix?: string;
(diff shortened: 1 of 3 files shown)


types/redux-actions/index.d.ts
```diff
@@ -66,6 +67,10 @@ export type Reducer<State, Payload> = (state: State, action: Action<Payload>) =>
 
 export type ReducerMeta<State, Payload, Meta> = (state: State, action: ActionMeta<Payload, Meta>) => State;
 
+export type ReduxReducer<State, Payload> = (state: State | undefined, action: Action<Payload>) => State;
```
alexey-pelykh: Please elaborate on these changes, specifically why we can not just add `undefined` to a `Reducer` definition?

oddui: @alexey-pelykh thanks for the quick review. The existing `Reducer` definition is used as `handleAction`'s parameter which is guaranteed to be passed in the state (because of the required `initialState`). I thought adding this extra reducer definition could better describe the library's API.

alexey-pelykh: @oddui you're welcome! Why not to use it as `Reducer<State | undefined, Payload>` where it can be undefined? Instead of adding extra type

oddui: Hi @alexey-pelykh thanks for the suggestion. However that also sets the return type of the reducer to be `state | undefined`, which i don't think is correct.

alexey-pelykh: Valid point, then let's just make a better name than `ReduxReducer` vs `Reducer`, I'm not sure what's specific about Redux in the name.

oddui: Happy to change that. Did you have any suggestions? Thanks

sheetalkamat (changes requested):
Please address suggestions requested by @alexey-pelykh

types/redux-actions/index.d.ts
oddui: Hi @alexey-pelykh does `ReduxCompatibleReducer` sound better?

types/redux-actions/index.d.ts
```diff
@@ -148,13 +153,13 @@ export function handleAction<State, Payload>(
     actionType: string | ActionFunctions<Payload> | CombinedActionType,
     reducer: Reducer<State, Payload> | ReducerNextThrow<State, Payload>,
     initialState: State
-): Reducer<State, Payload>;
+): ReduxReducer<State, Payload>;
```
alexey-pelykh: @oddui what I don't get - why reducer type have to have state as undefined yet initialState is always defined?

types/redux-actions/index.d.ts
```diff
@@ -165,19 +170,19 @@ export function handleActions<StateAndPayload>(
     reducerMap: ReducerMap<StateAndPayload, StateAndPayload>,
     initialState: StateAndPayload,
     options?: Options
-): Reducer<StateAndPayload, StateAndPayload>;
+): ReduxReducer<StateAndPayload, StateAndPayload>;
```
alexey-pelykh: Same here, inialState is not allowed to be undefined, why reducer may receive undefined state?

types/redux-actions/index.d.ts
oddui: Hi @alexey-pelykh the return type here allows `undefined` as state, in which case it will pass initialState to the internal `reducer`. The internal `reducer`'s type is not changed `(state: State, action: Action<Payload>) => State`. Does this answer your question?

types/redux-actions/index.d.ts
```diff
@@ -148,13 +153,13 @@ export function handleAction<State, Payload>(
...
     initialState: State
-): Reducer<State, Payload>;
+): ReduxReducer<State, Payload>;
 
 export function handleAction<State, Payload, Meta>(
     actionType: string | ActionWithMetaFunctions<Payload, Meta> | CombinedActionType,
     reducer: ReducerMeta<State, Payload, Meta> | ReducerNextThrowMeta<State, Payload, Meta>,
     initialState: State
-): Reducer<State, Payload>;
+): ReduxReducerMeta<State, Payload, Meta>;
```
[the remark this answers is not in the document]
alexey-pelykh: So this change is only meaningful for cases when `handleAction(something)(undefined, action)` is called?

alexey-pelykh (approved):
Okies, figured it out

types/redux-actions/index.d.ts
alexey-pelykh: Yes, thanks

types/redux-actions/index.d.ts
alexey-pelykh: Disregard this

types/redux-actions/index.d.ts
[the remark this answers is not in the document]
alexey-pelykh: And this as well

types/redux-actions/index.d.ts
alexey-pelykh: Yes, this works!

types/redux-actions/index.d.ts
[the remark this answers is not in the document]
oddui: Yes. Before these changes it does not work with redux types when strictNullChecks is on.

Thanks @alexey-pelykh for the review. I'm glad we can push this through.

oddui:
@sheetalkamat can you have another look?

types/redux-actions/index.d.ts
alexey-pelykh: @oddui please don't forget to address this ^ suggested name sounds good to me

oddui: Oh yes. Just changed that. Please have a look @alexey-pelykh

ansachs:
@oddui we also use Actions without payloads, and this change would force us to add excessive 'undefined's throughout our codebase. It appears that the redux-actions package also allows undefined payloads.

We were thinking about making a pull request, would you have a problem with using bang (e.g. `action.payload!`) in the tests to indicate payload being defined or creating a new type extending BaseAction which requires payload: 

`type RequiredKeys<T, K extends keyof T> = Exclude<T, K> & Required<Pick<T, K>>`
`export type ActionWithPayload<Payload> = RequiredKeys<Action<Payload>, 'payload'>`

oddui:
Hi @ansachs I agree it's annoying to have `undefined` all over the place. Sorry about that.

Adding a new type with payload required sounds good to me. But i'm not too sure how you use that. Do you have to add new types for `createAction` to use the new action type?

ansachs:
it would only be required where a payload creator is necessary; however, I am more wondering if there was a need for 'strict null checks'. 

The behavior of redux-actions has not changed: a payload is not necessary. What was the motivation behind this change?

thanks

PR #33670, opened 2019-03-07
@storybook/addon-knobs: make type of return value of select same with options array element

alanhchoi:
When select options is an array of string literal union type, select must return the same type of value.

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/storybooks/storybook/blob/next/addons/knobs/README.md#select>

As written in the document above, we can pass `string[]` as select `options`. Thus, we can pass an array of value typed string literals union, e.g. `('Apple' | 'Banana' | 'Grapes')[]`. In this case, returned value from `select` should be `'Apple' | 'Banana' | 'Grapes'`.

PR #33796, opened 2019-03-12
@storybook/addon-knobs: update select options type to accept all types <option> tag's value attribute can take

alanhchoi:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/storybooks/storybook/blob/next/addons/knobs/src/components/types/Select.js#L24>

When an array is passed as `options` parameter of `select` function, each element of the array is used as a `value` attribute of an `<option>` HTML element. `value` can take `string`, `number`, `string[]` or `undefined` by the definition. So I updated type of `options` parameter to be either `ReadonlyArray<string>`, `ReadonlyArray<number>`, or `<ReadonlyArray<string[]>`.

Thank you @akameco for pointing it out in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33670#discussion_r264516189


akameco (approved):
LGTM 💯

PR #33798, opened 2019-03-12
[redux-form] Update Optional argument of push method

mrsekut:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://redux-form.com/8.1.0/examples/fieldarrays/>>

### DIFF
--- a/types/redux-form/index.d.ts
+++ b/types/redux-form/index.d.ts
@@ -14,6 +14,7 @@
 //                 Mohamed Shaaban <https://github.com/mshaaban088>
 //                 Ethan Setnik <https://github.com/esetnik>
 //                 Walter Barbagallo <https://github.com/bwlt>
+//                 Kota Marusue <https://github.com/mrsekut>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 import {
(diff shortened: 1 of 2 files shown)


bwlt:
@mrsekut what is your scenario? why don't you push an "empty" `FieldValue`?
The fieldarray example is a bit ambiguous. I think it should be changed with `fields.push("")` i.e. pushing a valid `FieldValue` which represent an empty hobby.
The crucial point is that when you use `fields.pop` you expect that it returns a `FieldValue`, but with your changes it returns undefined. It could be a source of various bugs

mrsekut:
The following is the case.
This is similar to the redux-form documentation, except that the form validation is not `required`.
If you submit without being filled in, it will be sent excluding 'undefined' internally.

In my project, I used `remove ()` to remove `field`, but there was no problem.
But, as you said, I felt that it could be a source of various bugs, not safe.
In fact, it seems that we can cope with this as we can force `push (undefined)` as it is.

I apologize for the trouble.
It is okay to have it closed.

sheetalkamat:
@mrsekut @bwlt I am not sure from the conversation here if this change is correct per library spec. Can you please either close the PR if this change is not needed or if change is right, review it so I can merge this in. Thanks.

43081j:
FYI it seems to be right: https://github.com/erikras/redux-form/blob/7b75f1b71ace0bb4805f754347dbeb7ef96a812f/src/__tests__/FieldArray.spec.js#L1430

a test would be nice, though.

PR #33799, opened 2019-03-12
(ramda) Fix traverse types in the all-arrays case

googol:
The traverse function is not properly typeable at this moment since we lack higher-kinded types. Specific cases are doable though, like the array-array case I've typed here. Users can add their own specific cases in their own typings, to extend this.

The new version uses the same type parameters as the ramda documentation. The type signature in the documentation has a small bug though, see https://github.com/ramda/ramda/pull/2797

The original type was
```hs
traverse :: ([U] -> A) -> (T -> U) -> [T] -> A
```

which is not the same as the correct type
```hs
traverse :: (a → f a) → (a → f b) → t a → f (t b)
```

Even if `A = f a` and `U = f b`, the input type of the `of` parameter is way wrong and the output of `traverse` and the `of` parameters should not be the same type.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://ramdajs.com/docs/#traverse

### DIFF
--- a/types/ramda/index.d.ts
+++ b/types/ramda/index.d.ts
@@ -2686,9 +2686,9 @@ declare namespace R {
          * sequence to transform the resulting Traversable of Applicative into
          * an Applicative of Traversable.
          */
-        traverse<T, U, A>(of: (a: U[]) => A, fn: (t: T) => U, list: ReadonlyArray<T>): A;
-        traverse<T, U, A>(of: (a: U[]) => A, fn: (t: T) => U): (list: ReadonlyArray<T>) => A;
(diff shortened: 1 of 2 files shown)


googol:
The errors in this PR are caused by a recent TS change. This PR removes one of the errors

KiaraGrouwstra (approved):

tycho01:
looks good! the change in input type from array to `ReadonlyArray` definitely seems like a step forward. in fact, that's probably good to do for all of the input array types, as it makes the types robust to plugging in alternate list implementations such as ImmutableJS's `List`.
for context, Ramda was written using old JS, meaning all it uses on arrays are for-loops, i.e. exactly those bits that `ReadonlyArray` exposes (length and index access).

PR #33302, opened 2019-02-22
fix: Cannot redeclare block-scoped variable 'console' conflict with @types/node

rami-res-zz:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [ ] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: Acoording to issue's context <<https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33275>>

fix issue:
#33275
It seems fixed also:
#33015
#16825
#33204
#26218

PR #33800, opened 2019-03-12
[react-native] allow redeclaring require when node typing is present

danilobuerger:
Fixes
```
Subsequent variable declarations must have the same type.  Variable 'require' must be of type 'NodeRequire', but here has type 'ReactNativeRequireFunction'.
```
Caused by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33302

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://facebook.github.io/react-native/docs/javascript-environment.html#polyfills

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -9297,9 +9297,11 @@ export const PointPropType: React.Validator<PointPropType>;
 export const ViewPropTypes: React.ValidationMap<ViewProps>;
 
 declare global {
-    type ReactNativeRequireFunction = (name: string) => any;
+    interface NodeRequire {
+        (id: string): any;
+    }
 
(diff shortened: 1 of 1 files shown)


alloy:
Makes sense 👍 

Do you think we can have test coverage for this in a separate test file that would also import the Node typings (probably by using `/// <reference path="..." />`)?

danilobuerger:
@alloy Unfortunately I don't know how. I just tried, but the build errored out with:
```
Error: Test files should not use '<reference path="" />'. 'react-native/test/globals-node.tsx' references '../../node/index.d.ts'.
```

alloy:
Hmm, bummer. @andy-ms is this something that could be allowed in the future?

For now, let’s go with this as-is then. Thanks for trying!

PR #33803, opened 2019-03-12
[styled-system] Update ResponsiveValue to accept generic type variables for breakpoints

ktranada:
**Context**:
I am using an object to define my breakpoints.

**Issue**:
As it currently stands, `ResponsiveValue` utilizes an index signature - accepting any/all propertiets.

**Proposal**:
Update `type ResponsiveValue<T>` to take in an additional generic type variable for breakpoints, while also providing a default (string index signature) - which was previously hardcoded. In doing so, type consumers will have the ability to extend and have their breakpoint objects be strongly typed.

--- 

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [Defining breakpoints as an object and then passing an object to configure responsive styles](https://styled-system.com/responsive-styles#using-objects)

### DIFF
--- a/types/styled-system/index.d.ts
+++ b/types/styled-system/index.d.ts
@@ -48,9 +48,11 @@ export function style(
     args: LowLevelStylefunctionArguments
 ): { [cssProp: string]: string };
 
-export type TBreakpointAliasStyledSystem<T> = { [key: string]: T };
 export type TLengthStyledSystem = string | 0 | number;
-export type ResponsiveValue<T, B> = T | Array<T | null> | B;
+export interface BreakpointAliasStyledSystem<T> {
+    [key: string]: T;
+}
+export type ResponsiveValue<T, B = BreakpointAliasStyledSystem<TLengthStyledSystem>> = T | Array<T | null> | B;
 
 /**
  * Converts shorthand margin and padding props to margin and padding CSS declarations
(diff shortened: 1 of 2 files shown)


alloy:
LGTM 👍 

Could you add some test coverage that actually uses the alias generic, though? (And I mean _additional_ test coverage, not by changing the existing tests that now exercise the default value of the type parameter.)

alloy (changes requested):
Hold that thought, @typescript-bot

ktranada:
@alloy Of course! Will get that in shortly - brain was a lil foggy at 330AM 🤪

ktranada:
@alloy In writing my tests, I've noticed that while the shape of the Breakpoint Alias is now strongly typed, the values are not.

For example:

```
 interface BreakpointAlias {
     small?: string | number;
     medium?: string | number;
     large?: string | number;
 }

 interface GenericBreakpointBoxProps<T = TLengthStyledSystem, B = BreakpointAlias> extends 
    FlexWrapProps<T, B>,
    FlexDirectionProps<T, B>,
    WidthProps<T, B>
{ }
```

While defining `BreakpointAlias` outside and passing it to the `GenericBreakpointsBoxProps` can reduce keystrokes, every property defined via the alias will now be of the same type. So `flexWrap` will accept any string | number when it should be typed to the `FlexWrapProperty` defined in `csstype`.

Please feel free to correct me, but it feels that in order to support this feature, the consumer will have to manually extend the props with the correct CSS Property.

For example:
```
interface BreakpointAlias<T> {
     small?: T;
     medium?: T;
     large?: T;
 }

interface FlexWrapProps extends 
  styledSystem.FlexWrapProps<string | number, BreakpointAlias<CSS.FlexWrapProperty>> {}

....
```

While this is a tradeoff, I still value the ability to have my aliases strongly typed. What are your thoughts?

ktranada:
Also, the need to extend Props to handle breakpoint aliases can come by a need-to-have basis. In that the consumer can decide which props will need to handle responsive styles via object.

alloy:
Ah, I see. In a [media query related library](https://github.com/artsy/react-responsive-media) of our own, we have [a single function](https://github.com/artsy/react-responsive-media/blob/9b546c24b2f9261b4c9874168cb5016ad1325142/src/Media.tsx#L300-L304) that creates the actual API to use, which also allows for making the entire generic API specific to your project’s breakpoints. Obviously this isn’t a change we can make to styled-system or the types, but perhaps what you could do is create a similar function/library that would wrap styled-system to give you a version of the API that has the breakpoint aliases made specific?

I do think that for this types package we should leave the simplest existing default in place and not break any projects, i.e. leave it as-is.

Does that sound good?

PR #33815, opened 2019-03-12
[@types/vscode] Add 1.14 Extension API

octref:
Please fill in this template.

If adding a new definition:
- [ ] 🚨 If this is for an NPM package, match the name. If not, do not conflict with the name of an NPM package. I didn't do this because similar to how [`@types/chrome`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/chrome/index.d.ts) matches [Chrome API](https://developers.chrome.com/extensions/api_index), `@types/vscode` should match the VS Code API described at https://code.visualstudio.com/api/references/vscode-api. The `vscode` NPM module contain a JS script to pull `vscode.d.ts` from microsoft/vscode manually, and we'll deprecate that package. See https://github.com/Microsoft/vscode/issues/70175 for details.
- [ ] 🚨 Create it with `dts-gen --dt`, not by basing it on an existing project. This is built upon https://github.com/Microsoft/vscode/blob/release/1.14/src/vs/vscode.d.ts.

---

Questions:

- Is this header OK? `// Type definitions for VS Code 1.14 extension development`, or should I change it to `// Type definitions for VS Code 1.14` for the versioning to work correctly?


types/vscode/tslint.json
```diff
@@ -0,0 +1,80 @@
+{
+	"extends": "dtslint/dt.json",
+	"rules": {
+		"adjacent-overload-signatures": false,
+		"array-type": false,
+		"arrow-return-shorthand": false,
+		"ban-types": false,
+		"callable-types": false,
+		"comment-format": false,
+		"dt-header": false,
```
weswigham: You cannot disable this rule (as you suspected) - publishing won't work if the header metadata isn't readable.

weswigham (changes requested):

types/vscode/tslint.json
octref: Thanks, resolved.

Is that the only field I should care about for publishing, or are there other fields affecting publishing too?

weswigham: Well, you need the lint rule to actually pass~

octref: Oops, sorry. Stupid me.

weswigham (approved):

Issue #33820, opened 2019-03-12
[@types/jasmine] Partial is not a good fit for objectContaining


silkentrance:
A PR for this is coming shortly, I just wanted to document this.

- [x] I tried using the `@types/jasmine` package and had problems.
- [x] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @SimonSchick @pe8ter 

## Rationale

The interface declaration for objectContaining defines the following

```
    interface ObjectContaining<T> {
        new (sample: Partial<T>): Partial<T>;

        jasmineMatches(other: any, mismatchKeys: any[], mismatchValues: any[]): boolean;
        jasmineToString(): string;
    }
...
    function objectContaining<T>(sample: Partial<T>): ObjectContaining<T>;
```

and in the test it uses the following to test that interface

```
    it("The 'toContain' matcher is also for finding an object containing distinct properties in an Array", () => {
        const a = [{ a: "foo" }, { a: "bar" }, { b: "baz" }];

        expect(a).toContain(jasmine.objectContaining({ a: "foo" }));
        expect(a).not.toContain({ a: "quux" });
    });
```

Now, with the ts compiler option `strictFunctionTypes` set to true, these tests will fail due to a mismatch in the interface, complaining about {a?:string, b:string} | {a:string, b?:string} and {a:string, b:undefined} and so on.

As of now, the option is set to false which basically hides that issue.

It would be better to not make the interface expect a Partial, as that is defined as

```
/**
 * Make all properties in T optional
 */
type Partial<T> = {
    [P in keyof T]?: T[P];
};
```

This will eventually cause ts to infer that the Partial is either {a:string, b?: string} or {b: string, a?: string}. And passing in different objects as samples to the objectContaining matcher will now result in error as the sample never matches the partial.

From my standpoint, I want to check whether an arbitrary object matches the interface. Therefore it must be declared as any and not limited to just the known properties.

## References

See also the typings for jest, which builds on jasmine. There, ObjectContaining is not generic and objectContaining expects a sample that is of type any.

SimonSchick:
Uhh, why was I tagged for this?

silkentrance:
[...]

PR #33821, opened 2019-03-12
[@types/jasmine] use any instead of Partial<T> in objectContaining and ObjectContaining matcher

silkentrance:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: #33820

### DIFF
--- a/types/jasmine/index.d.ts
+++ b/types/jasmine/index.d.ts
@@ -194,7 +194,7 @@ declare namespace jasmine {
 
     function arrayContaining<T>(sample: ArrayLike<T>): ArrayContaining<T>;
     function arrayWithExactContents<T>(sample: ArrayLike<T>): ArrayContaining<T>;
-    function objectContaining<T>(sample: Partial<T>): ObjectContaining<T>;
+    function objectContaining<T>(sample: any): ObjectContaining<T>;
     function createSpy(name?: string, originalFn?: Function): Spy;
 
     function createSpyObj(baseName: string, methodNames: SpyObjMethodNames): any;
@@ -236,8 +236,8 @@ declare namespace jasmine {
         jasmineToString(): string;
     }
 
-    interface ObjectContaining<T> {
-        new (sample: Partial<T>): Partial<T>;
+    class ObjectContaining<T> {
+        constructor(sample: any);
 
         jasmineMatches(other: any, mismatchKeys: any[], mismatchValues: any[]): boolean;
         jasmineToString(): string;
--- a/types/jasmine/jasmine-tests.ts
+++ b/types/jasmine/jasmine-tests.ts
@@ -854,19 +854,17 @@ describe("jasmine.objectContaining", () => {
     });
 
     it("matches objects with the expect key/value pairs", () => {
-        // not explictly providing the type on objectContaining only guards against
-        // missmatching types on know properties
+        // let the compiler try to infer the actual type
         expect(foo).not.toEqual(jasmine.objectContaining({
             a: 37,
-            foo: 2, // <-- this does not cause an error as the compiler cannot infer the type completely
-            // b: '123', <-- this would cause an error as `b` defined as number in fooType
+            foo: 2, // <-- this should not cause a compiler error as foo will be added in a future version of the interface
+            b: '123', // <-- this should not cause a compiler error as b will be refactored to string in a future version of the interface
         }));
 
-        // explictly providing the type on objectContaining makes the guard more precise
-        // as misspelled properties are detected as well
+        // provide an explicit type hint
         expect(foo).not.toEqual(jasmine.objectContaining<fooType>({
             bar: '',
-            // foo: 1, <-- this would cause an error as `foo` is not defined in fooType
+            foo: 1, // <-- this should not cause a compiler error because we expect foo to be available in a future version of the interface
         }));
     });
 
--- a/types/jasmine/tsconfig.json
+++ b/types/jasmine/tsconfig.json
@@ -12,7 +12,7 @@
         "noImplicitAny": true,
         "noImplicitThis": true,
         "strictNullChecks": true,
-        "strictFunctionTypes": false,
+        "strictFunctionTypes": true,
         "baseUrl": "../",
         "typeRoots": [
             "../"


cyungmann:
@silkentrance do you have an example of a problem this solves?

silkentrance:
@cyungmann yes, you own tests :smile:. They will work only if `strictFunctionTypes` is set to false. When set to true, the tests will fail, see the information provided in #33820.

```
jasmine-tests.ts:122:29 - error TS2345: Argument of type 'ObjectContaining<{ a: string; }>' is not assignable to parameter of type 'Expected<{ a: string; b?: undefined; } | { b: string; a?: undefined; }>'.
  Type 'ObjectContaining<{ a: string; }>' is not assignable to type 'ObjectContaining<{ a: string; b?: undefined; } | { b: string; a?: undefined; }>'.
    Types of parameters 'sample' and 'sample' are incompatible.
      Type 'Partial<{ a: string; b?: undefined; }> | Partial<{ b: string; a?: undefined; }>' is not assignable to type 'Partial<{ a: string; }>'.
        Type 'Partial<{ b: string; a?: undefined; }>' is not assignable to type 'Partial<{ a: string; }>'.
          Type '{ b: string; a?: undefined; }' is not assignable to type '{ a: string; }'.

122         expect(a).toContain(jasmine.objectContaining({ a: "foo" }));
                                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

jasmine-tests.ts:859:33 - error TS2345: Argument of type 'ObjectContaining<{ a: number; foo: number; }>' is not assignable to parameter of type 'Expected<fooType>'.
  Type 'ObjectContaining<{ a: number; foo: number; }>' is not assignable to type 'ObjectContaining<fooType>'.
    Types of parameters 'sample' and 'sample' are incompatible.
      Type 'Partial<fooType>' is not assignable to type 'Partial<{ a: number; foo: number; }>'.
        Property 'foo' is missing in type 'fooType' but required in type '{ a: number; foo: number; }'.

859         expect(foo).not.toEqual(jasmine.objectContaining({
                                    ~~~~~~~~~~~~~~~~~~~~~~~~~~
860             a: 37,
    ~~~~~~~~~~~~~~~~~~
...
862             // b: '123', <-- this would cause an error as `b` defined as number in fooType
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
863         }));
    ~~~~~~~~~~

  jasmine-tests.ts:861:13
    861             foo: 2, // <-- this does not cause an error as the compiler cannot infer the type completely
                    ~~~~~~
    'foo' is declared here.
```

cyungmann:
@silkentrance is this related to https://github.com/Microsoft/TypeScript/issues/13195, then?

silkentrance:
While this might be, it is more or less a misinterpretation of the original interface, see https://jasmine.github.io/2.0/introduction#section-Partial_Matching_with_%3Ccode%3Ejasmine.objectContaining%3C/code%3E

here, there is no mention of a Partial interface, just of a partial match, which in fact is different.

cyungmann (approved):

cyungmann (commented):
Whoops... Might actually see if anybody else has feedback (don't know if there's a better way to reset my feedback?)

cyungmann (changes requested):
.

cyungmann:
Ugh...apparently I don't know how to GitHub. @RyanCavanaugh do you know if there is a way to simply "reset" / "retract" my PR review feedback?


@silkentrance at any rate, I spent a little more time looking at this. It seems the first test can be solved with some additional type annotations:

```typescript
it("The 'toContain' matcher is also for finding an object containing distinct properties in an Array", () => {
    const a: Array<{a?: string, b?: string}> = [{ a: "foo" }, { a: "bar" }, { b: "baz" }];

    expect(a).toContain(jasmine.objectContaining<{a?: string, b?: string}>({ a: "foo" }));
    expect(a).not.toContain({ a: "quux" });
});
```

The first part of the second test does not seem to test a real use case and can probably be deleted (or have some type-casts to `any` added), as `fooType` does not contain `foo: number`:

```typescript
    it("matches objects with the expect key/value pairs", () => {
        // not explictly providing the type on objectContaining only guards against
        // missmatching types on know properties
        expect(foo).not.toEqual(jasmine.objectContaining<fooType>({
            a: 37,
            foo: 2, // <-- this does not cause an error as the compiler cannot infer the type completely
            // b: '123', <-- this would cause an error as `b` defined as number in fooType
        }));

        // explictly providing the type on objectContaining makes the guard more precise
        // as misspelled properties are detected as well
        expect(foo).not.toEqual(jasmine.objectContaining<fooType>({
            bar: '',
            // foo: 1, <-- this would cause an error as `foo` is not defined in fooType
        }));
    });
```

Your change seems to go against this bit under [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped#common-mistakes):
> `getMeAT<T>(): T`: If a type parameter does not appear in the types of any parameters, you don't really have a generic function, you just have a disguised type assertion. Prefer to use a real type assertion, e.g. `getMeAT() as number`. Example where a type parameter is acceptable: `function id<T>(value: T): T;`. Example where it is not acceptable: `function parseJson<T>(json: string): T;`. Exception: `new Map<string, number>()` is OK.

silkentrance:
@cyungmann how could this then work with the following test cases?

https://jasmine.github.io/2.0/introduction#section-Partial_Matching_with_%3Ccode%3Ejasmine.objectContaining%3C/code%3E

Here

```
  beforeEach(function() {
    foo = {
      a: 1,
      b: 2,
      bar: "baz"
    };
  });

  it("matches objects with the expect key/value pairs", function() {
    expect(foo).toEqual(jasmine.objectContaining({
      bar: "baz"
    }));
    expect(foo).not.toEqual(jasmine.objectContaining({
      c: 37
    }));
  });
```

c is definitely not part of the interface, yet the test should not fail because of some not matching interface but rather because of the not matching data.

cyungmann:
I do not understand why you would desire a test like that (the second test) when with the current typings the type-checker would check it at compile-time (vs at runtime). The test would always pass because of the disjoint types. I believe you could always do `expect(foo as any)` or similar if you really wanted to.

However, I would like to defer to other maintainers (I just don't know of a way to retract my request for changes).

@borisyankov @theodorejb @davidparsson @gmoothart @lukas-zech-software @Engineer2B @Roaders @devoto13 @FDIM @pe8ter could one of you take a look at this?

silkentrance:
Hm, considering an API that returns to you a given object or interface, say an OpenAPI defined JSON structure which is then mapped into a JS POJO.

Now, this data structure is the basis, for which there also exists some type of interface, e.g. an OpenAPI generated typescript package.

Let's also assume that the result returned by that API is a union of multiple interfaces.

Let's also assume that for a current change request, I want to implement the tests for these changes before hand and not when the change is reflected in the API, say a property of the SUT is renamed or removed or a new property is added.

Given the current approach, I am not able to define any tests for that, as the compiler will already warn about the mismatch and will reject the code with an error, unless I use more lax typescript compiler options, which might not be an option at all.

And, given the above information, any is also out of question as this is the same as the more lax compiler options.

I will rework the PR.

@cyungmann thanks for the valuable review!

silkentrance:
@cyungmann I have looked into this further and found that

```
    interface ObjectContaining<T> {
        new (sample: Partial<T>): Partial<T>;

        jasmineMatches(other: any, mismatchKeys: any[], mismatchValues: any[]): boolean;
        jasmineToString(): string;
    }
```

the constructor new is actually returning the wrong value, it should be ObjectContaining<T> or should it not?

As for sample, this should be any, as we do not want to prevent testing against future versions of the interface T, or would we?

silkentrance:
@cyungmann I would therefor propose the following change to the ObjectContaining interface

```
    interface ObjectContaining<T> {
        new (sample: any): ObjectContaining<T>;    
```

And we cannot have a union of Partial<T> | any as the type hint for sample (no-any-union).

silkentrance:
As for the current Jasmine interface, ObjectContaining prototype declares

```
  ObjectContaining.prototype.asymmetricMatch = function(other, customTesters) { ... }
```

returning a boolean.

Whereas the typings declare

```
        jasmineMatches(other: any, mismatchKeys: any[], mismatchValues: any[]): boolean;
```

Should this not be the same?

silkentrance:
@cyungmann also, ObjectContaining should be declared as a class instead of as an interface (no-misused-new)

silkentrance:
But then again, why would we need to expose these internal interfaces at all.

Both ArrayContaining and ObjectContaining, and others, represent AsymmetricMatcher objects, with a rather simple interface:

```
interface AsymmetricMatcher {
  asymmetricMatch(other, customTesters);
  jasmineToString(): string;
}
```

with the type hints for other and customTests still to be decided.

silkentrance:
I am closing this for now as all of the interfaces that provide a new() constructor like operation are far from what jasmine declares. I will look into that further and provide you with a new PR that will fix these issues.

PR #33823, opened 2019-03-12
Change 'export default' to 'export =', part 2

sandersn:
For packages that don't actually export a 'default' property.


sandersn:
Here are the packages with more than 1,000 download per week:

Package | Downloads
-----------|--------
memoize-one | 82,271
promise-polyfill | 40,655
workbox-sw | 26,726
xxhashjs | 17,452
parse-ms | 5,135

sandersn:
There's an odd number of changed files because sqlanywhere had an incorrect default export that was not used by the tests.

Kocal (approved):

Kocal:
But wait, given https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32526#discussion_r251679879, we can't use `export default` in our `.d.ts` if the library does not use `export default`.

For example, the [rollup-plugin-buble](https://github.com/rollup/rollup-plugin-buble/blob/master/src/index.js#L4) use `export default`, so we should keep `export default` in the [`.d.ts`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33823/files#diff-635319ca5cd9c6e3971bd2cd3cc5fe40), no? Or I am still missing something? :thinking:

pkeuter:
@sandersn This will break sqlanywhere for me. Could you please explain what these changes fix (especially in the case of sqlanywhere)? And how we can make the package work again?

See here for the exports: https://github.com/sqlanywhere/node-sqlanywhere/blob/master/lib/index.js

sandersn:
@Kocal 
The situation is complicated, which is why it's so easy to get wrong. Let's assume that the person using the package has `"esModuleInterop": false`, which is the legacy setting that Definitely Typed assumes *might* be the case.

In that case, the shipped code, not the source, must have a default property in order to default imports to work. Something like:

```js
module.exports["default"] = Thing
```

Here is the [code that ships from npm for rollup-plugin-buble](https://unpkg.com/rollup-plugin-buble@0.19.6/dist/rollup-plugin-buble.cjs.js):

```js
'use strict';
var buble = require('buble');
var rollupPluginutils = require('rollup-pluginutils');
function buble$1 ( options ) {
    // code goes here ...
}
module.exports = buble$1;
```

Note that there is a `rollup-plugin-buble.es.js` that ships as well -- it's basically the pre-babel source file -- but that's not what package.json points to! So, strictly speaking, `rollup-plugin-buble` has no default export. So, strictly speaking, a d.ts should not use `export default`, but `export =`.

Individual users can enable interop mode with `--esModuleInterop`, allowing them to write `import B from 'rollup-plugin-buble'`, but not everyone has it turned on. So packages on Definitely Typed are required to use `export =` for backward compatibility.

sandersn:
@pkeuter sqlanywhere doesn't have a default export [in the code as shipped](https://unpkg.com/sqlanywhere@1.0.24/lib/index):

```
$ node
> var sql = require('sqlanywhere')
undefined
> sql.default
undefined
> sql
{ createConnection: [Function: createConnection] }
```

In sqlanywhere's case, `export = { createConnection }` is equivalent to `export function createConnection`, so I just changed it to that. Can you show your usage of sqlanywhere that will break?

dfreeman:
LGTM for `core-object` (I actually think I have a branch with this same fix floating around somewhere 😳) — thanks @sandersn

types/memoize-one/index.d.ts
```diff
@@ -3,7 +3,9 @@
...
 
-declare function memoizeOne<T extends (...args: any[]) => any>(resultFn: T, isEqual?: EqualityFn): T;
-export type EqualityFn = (a: any, b: any, index: number) => boolean;
+declare function memoizeOne<T extends (...args: any[]) => any>(resultFn: T, isEqual?: memoizeOne.EqualityFn): T;
+declare namespace memoizeOne {
+    type EqualityFn = (a: any, b: any, index: number) => boolean;
+}
 
-export default memoizeOne;
+export = memoizeOne;
```
karol-majewski: `memoize-one` is distributed in a few formats: CommonJS, ESM, and UMD.

```
memoize-one.cjs.js
memoize-one.cjs.js.flow
memoize-one.esm.js
memoize-one.js
memoize-one.min.js
```

It then defines two entry points in [`package.json`](https://github.com/alexreardon/memoize-one/blob/d143f883ff81dba9c46f148d212337667b941a99/package.json):

-  "main": "dist/memoize-one.cjs.js",
-  "module": "dist/memoize-one.esm.js",

Since `memoize-one.esm.js` does include a `default export`, but it's not distributed under the `main` field, does it mean the change in this PR was justified and we should treat it as if it did not have a default export?

weswigham: It's justified. TS doesn't use the `module` field, so only `main` matters.

Kocal:
Oh I see!
Thanks for the explanation! :)

pkeuter:
@sandersn You are right, I'm sorry, I was looking incorrectly. It works fine.

types/memoize-one/index.d.ts
sandersn: It's not just TS, it's node. We're trying to match the lowest-common-denominator usage:

```
$ node
> require('memoize-one').default
```

If that's undefined, then the module doesn't have a default export.

Other loaders may behave differently, but not everyone is guaranteed to be using those loaders. If they are, then they can turn on `esModuleInterop`.

PR #33835, opened 2019-03-13
Add types to allow string array argument for hmset

OpesanyaAdebayo:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://www.npmjs.com/package/redis>>


blablapolicja (approved):

43081j (changes requested):
can you please point us to the specific part of the docs where this pattern is specified?

the docs state:

> All functions take either an args Array plus optional callback Function or a variable number of individual arguments followed by an optional callback.

which implies we need these forms:

* `(key, arg1, arg2, arg3, ..., callback)`
* `([key, arg1, arg2, arg3, ...], callback)`
* `(key, [arg1, arg2, arg3, ...], callback)`

is that right?

OpesanyaAdebayo:
Yes, you're right @43081j

43081j:
if so, the types specify that the _key_ can be array which isn't quite true, is it. its more like:

* `(key: string, ...args: Array<T | Callback<U>>)`
* `([string, ...Array<T | Callback<U>>])`
* `(key: string, Array<T | Callback<U>>)`

or something along those lines. most importantly the second one is missing

OpesanyaAdebayo:
Thank you very much.

I'll look into it and make changed as suggested.

blablapolicja:
@43081j https://github.com/NodeRedis/node_redis#sending-commands. It states that `All functions take either an args Array plus optional callback Function or a variable number of individual arguments followed by an optional callback`

blablapolicja:
Sorry, I got this wrong. Changes are needed

43081j:
```ts
    (args: [string, ...T[]], cb?: Callback<U>): R;
    (key: string, args: T[], cb?: Callback<U>): R;
```

might be all you need but im not too sure off the top of my head

OpesanyaAdebayo:
I've updated the typings and corresponding tests. Thanks for the help!

types/redis/index.d.ts
```diff
@@ -93,13 +94,15 @@ export interface OverloadedListCommand<T, U, R> {
...
 
 export interface OverloadedSetCommand<T, U, R> {
-    (key: string, arg1: T, arg2: T, arg3: T, arg4: T, arg5: T, arg6: T, cb?: Callback<U>): R;
-    (key: string, arg1: T, arg2: T, arg3: T, arg4: T, arg5: T, cb?: Callback<U>): R;
-    (key: string, arg1: T, arg2: T, arg3: T, arg4: T, cb?: Callback<U>): R;
-    (key: string, arg1: T, arg2: T, arg3: T, cb?: Callback<U>): R;
-    (key: string, arg1: T, arg2: T, cb?: Callback<U>): R;
-    (key: string, arg1: T | { [key: string]: T } | T[], cb?: Callback<U>): R;
-    (key: string, ...args: Array<T | Callback<U>>): R;
+    (key: string | string[], arg1: T, arg2: T, arg3: T, arg4: T, arg5: T, arg6: T, cb?: Callback<U>): R;
```
43081j: do you need this `string[]` anymore? it should be accounted for by the two new types i think

OpesanyaAdebayo: Oops. I missed that. I'll fix it and push right away. Thank you.

OpesanyaAdebayo: Updated. Kindly review. Thank you.

43081j (approved):

43081j:
one of your build failures is because these two can be unified:

```ts
    (key: string, arg1: T | { [key: string]: T } | T[], cb?: Callback<U>): R;
    (key: string, args: T[], cb?: Callback<U>): R;
```

It looks like we had already accounted for the form: `(string, [arg1, arg2, ...], cb)` in that.

You can remove the bottom one (`(key: string, args: T[], cb?: Callback<U>): R;`) i think.

OpesanyaAdebayo:
I've removed the duplicate typing. I really hope nothing breaks this time. So much for getting simpler 😂

OpesanyaAdebayo:
Build failed again 😫 Should I did what is recommended by CI: `Fix with a comment '// TypeScript Version: 3.0' just under the header.`?

This keeps getting more interesting. 😰

SimonSchick:
@OpesanyaAdebayo you'll have to create a specific folder for 3.x, see node typings as an example.

43081j:
I suspect maybe spread types in tuples weren't supported pre-3.0?

i.e. `[TFoo, ...TBar[]]`

we could bump the version but it'd need a breaking change release. so the alternative is to add a 3.x folder and use type versions.

an example of this at work is the sinon package:

* [package.json](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/be491c99abc767dbd387ffb9426bf68bd67d3785/types/sinon/package.json#L4)
* [tsconfig](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/729bada5c3700e852c9d50464f159a27e457d8ff/types/sinon/ts3.1/tsconfig.json#L14) needs to include a relative type root
* the newer of the two versions goes in a subdirectory (`3.0` in this case, as it includes all versions beyond that too such as 3.1)

OpesanyaAdebayo:
Finally got the build to pass 😄

PR #33844, opened 2019-03-13
ember-data: In DS.PromiseObject forward the passed in type to sub-objects so .content property gets correct type

erkie:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://api.emberjs.com/ember-data/3.2/classes/DS.PromiseObject

The resolved promise is the result of the `.content` property, which itself is used for proxying.

### DIFF
--- a/types/ember-data/test/promises.ts
+++ b/types/ember-data/test/promises.ts
@@ -1,0 +1,20 @@
+import DS from 'ember-data';
+import { assertType } from "./lib/assert";
+
+declare const store: DS.Store;
+
+class Person extends DS.Model {
+    firstName = DS.attr();
+}
+
+declare module 'ember-data/types/registries/model' {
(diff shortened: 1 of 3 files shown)


types/ember-data/test/promises.ts
```diff
@@ -0,0 +1,20 @@
...
+
+declare module 'ember-data/types/registries/model' {
+    export default interface ModelRegistry {
+        person: Person;
+    }
+}
+
+const promiseFind = store.findRecord("person", 1);
+
+assertType<Person | undefined>(promiseFind.content);
```
mike-north:
```suggestion
promiseFind.content; // $ExpectType Person | undefined
```

mike-north (changes requested):

mike-north (approved):
Thanks for your PR! LGTM

PR #33859, opened 2019-03-13
Update fluent definitions for v0.11

JamesNimlos:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/projectfluent/fluent.js/blob/master/fluent/src/bundle.js#L68-L85
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. // done previously

### Note
There is a recursive typedef in the dependent package `fluent-react` which is broken for `typescript@next` in master as well and prevents passing tests. See this reported issue [TypeScript#30188](https://github.com/Microsoft/TypeScript/issues/30188). I'm not sure what the solution for that is, but am willing to fix it
```
Error: /Users/jamesnimlos/dev/DefinitelyTyped/types/fluent-react/index.d.ts:101:49
ERROR: 101:49  expect  TypeScript@next compile error: 
Type instantiation is excessively deep and possibly infinite.
```

### EDIT
I've removed the circular reference in the dependent package that was causing the error and the tests for that package still pass. This is not a package I am using so I don't have good insight as to whether this is a good change or not, but due to the automated systems I need to make it in order to get a real person to review. As a precaution I've updated the noted TypeScript version of that typedef to the TS version where the breaking change was added.

As a critique, I'm surprised this breaking change is in a minor version instead of mandating a major TS version like semver dictates.

Recommended Action: I'd suggest creating a group of maintainers who are willing to inspect errors like this to be called upon when new contributors make PRs where they don't understand the next step to remedy errors. This is definitely an uncomfortable situation for new contributors and feels a bit like gatekeeping?

### DIFF
--- a/types/fluent-react/index.d.ts
+++ b/types/fluent-react/index.d.ts
@@ -76,7 +76,7 @@ export type Matching<InjectedProps, DecorationTargetProps> = {
  */
 export type Shared<
     InjectedProps,
-    DecorationTargetProps extends Shared<InjectedProps, DecorationTargetProps>
+    DecorationTargetProps
     > = {
         [P in Extract<keyof InjectedProps, keyof DecorationTargetProps>]?: InjectedProps[P] extends DecorationTargetProps[P] ? DecorationTargetProps[P] : never;
     };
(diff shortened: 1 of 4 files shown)


JamesNimlos:
I don't exactly know how to respond to this since I don't have much familiarity with the broken dependency and the break isn't due to my changes but rather that TypeScript 3.4 has a breaking change disallowing circular dependencies

PR #33881, opened 2019-03-14
Add compatibility for Waterline v3

makenoisewacamayo:
Add compatibility to waterline v3

Package modified types/waterline Add new types for v3 of waterline

#### Change list
* ConfigV3
* Add registerModel method `registerModel(collection: CollectionClassV3): void`
* CollectionClassV3
* CollectionDefinitionV3
* AttributesV3
* AttributeTypeV3
* BaseAttributeV3<T>
* new interface AutoMigration
* new interface AttributeValidationsV3
* keep old interfaces to maintain compatibility with v2

i added definition for waterline v3 keep consistency with v2 types.

### DIFF
--- a/types/waterline/index.d.ts
+++ b/types/waterline/index.d.ts
@@ -11,9 +11,6 @@ declare namespace Waterline {
     type Adapter = Object;
     type Connection = {
         adapter: string;
-    }
-    type ConnectionV3 = {
-        adapter: string;
         url?: string,
         host?: string,
         port?: number,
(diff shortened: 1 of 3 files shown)


types/waterline/index.d.ts
```diff
@@ -12,21 +12,37 @@ declare namespace Waterline {
     type Connection = {
         adapter: string;
     }
+    type ConnectionV3 = {
```
rbuckton: I couldn't tell from the documentation, but is it necessary to have definitions for the existing and the "V3" versions? Would it not make more sense to just use the updated definitions where there is sufficient overlap with the existing types? For example, `ConnectionV3` is assignment-compatible with `Connection` because all of the additional properties are optional.

makenoisewacamayo: Well v2 and v3 are not compatible at all, i can use the existing definition but my point was to note the differences.  i will update the existing types, keep in mind the type array is valid in version 2 but not in version 3.  

The actual version of waterline was build to make ease migrate from different database vendor, so the types for define tables are reduced to most compatible "string", "number", "boolean", "json" and "ref" plus definition to make 1-n and n-m relations

makenoisewacamayo: I update the existing types.

DanielRosenwasser:
I think the appropriate change would be to create a `v2` for the old declarations, and just make v3.x the primary API surface area.

makenoisewacamayo:
> I think the appropriate change would be to create a `v2` for the old declarations, and just make v3.x the primary API surface area.
Hmm, could be, Don't be sure if this break existing exiting types ? What do you think @rbuckton

uniqueiniquity:
@makenoisewacamayo I can step in while @rbuckton is out; that would be the recommended approach.
By moving the existing types to a v2 folder, you can still update them and publish updates as @types/waterline v2.X, while publishing the new API as @types/waterline v3.X.

uniqueiniquity (changes requested):
Marking as changes requested until existing types moved to [`v2`](https://github.com/DefinitelyTyped/DefinitelyTyped#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package) folder.

makenoisewacamayo:
> @makenoisewacamayo I can step in while @rbuckton is out; that would be the recommended approach.
> By moving the existing types to a v2 folder, you can still update them and publish updates as @types/waterline v2.X, while publishing the new API as @types/waterline v3.X.

Thanks i do before notice any wacamayo sound

makenoisewacamayo:
> @makenoisewacamayo I haven't seen anything from you in a while and this PR currently has problems that prevent it from being merged. The PR will be closed tomorrow if there aren't new commits to fix the issues.

Sorry i been busy please give to monday to make to amend thr PR

PR #33882, opened 2019-03-14
knex column builder defaultTo() accepts null

panzi:
knex's `ColumnCompiler.defaultTo()` method also accepts `null` values.

### DIFF
--- a/types/knex/index.d.ts
+++ b/types/knex/index.d.ts
@@ -491,7 +491,7 @@ declare namespace Knex {
         references(columnName: string): ReferencingColumnBuilder;
         onDelete(command: string): ColumnBuilder;
         onUpdate(command: string): ColumnBuilder;
-        defaultTo(value: Value): ColumnBuilder;
+        defaultTo(value: Value | null): ColumnBuilder;
         unsigned(): ColumnBuilder;
         notNullable(): ColumnBuilder;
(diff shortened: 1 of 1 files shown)


mastermatt:
@panzi can you add and fill out [the template](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/.github/PULL_REQUEST_TEMPLATE.md) for this PR please. It will have you include links to docs and/or code backing up the change.
Also add a call in the test file that takes a `null`.

panzi:
Sorry, got work stress and then I'll be away for 1.5 weeks. I just noticed this while working and wanted to let you know. Have no time for more. But I can link you to that: https://github.com/tgriesser/knex/blob/11fdc0c90afbcf8011aea5a2493458dfda3052f6/src/schema/columncompiler.js#L150

mastermatt (changes requested):
I don't think this PR should be merged.
- It seems like `null` may not actually work as a value https://github.com/tgriesser/knex/issues/3092
- As of last Oct, Knex bundles its own types https://github.com/tgriesser/knex/pull/2845

mastermatt:
I've gone ahead and begun the process of removing `@types/knex`. https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33889

PR #33884, opened 2019-03-14
Update createConnection in http2.d.ts

AlexOwl:
[NodeJS calls `createConnection` with 2 arguments: `authority` and `options`](https://github.com/nodejs/node/blob/107c95da0dd8cb4c9024d762b61b62630bfb7bba/lib/internal/http2/core.js#L2778)

[Documentation says the same (`createConnection` param)](https://nodejs.org/api/http2.html#http2_http2_connect_authority_options_listener)

[There are no tests for such case in tests folder](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/test/http2.ts)


Flarna:
I think the same is applicable for NodeJs 8 and 10. Could you update these also?

> There are no tests for such case in tests folder

Well, seems to be a good idea then to add one :o).

AlexOwl:
Updated

Flarna (approved):

BrunoScheufler (approved):

SimonSchick (approved):

sandersn:
@AlexOwl now that the unrelated lints are fixed, you'll need to merge from master.

Flarna:
@AlexOwl CI fail is not caused by your changes, seems some new check has been introduced recently.
Fix should be easy: Just replace the first line in `types/node/v10/index.d.ts` with `// Type definitions for non-npm package Node.js 10.14` (see `types/node/index.d.ts` as reference)

AlexOwl:
I will create new pull request

sandersn:
Actually, the new lint rule was not supposed to be running on old versions of the package, and I just shipped a fix. It would still be a good idea to change the header, but it's not needed to pass CI anymore.

AlexOwl:
Moved to [#34197](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34197)

PR #33889, opened 2019-03-14
Remove @types/knex.

mastermatt:
Self bundles types since [0.16.0](https://github.com/tgriesser/knex/pull/2845)


mastermatt:
I don't understand the Travis error
> AssertionError [ERR_ASSERTION]: The entry for @types/knex in notNeededPackages.json has
"libraryName": "Knex", but there is no npm package with this name.
Unneeded packages have to be replaced with a package on npm.

The addition looks like
```json
{
    "libraryName": "Knex",
    "typingsPackageName": "knex",
    "sourceRepoURL": "https://github.com/tgriesser/knex",
    "asOfVersion": "0.16.0"
}
```
and the README says 
> libraryName: Descriptive name of the library, e.g. "Angular 2" instead of "angular2". (If ommitted, will be identical to "typingsPackageName".)

So should `libraryName` be a descriptive name or the package name on npmjs.com?
Maybe @sandersn can shed some light?

sandersn:
That error is new code. Maybe it's wrong, or maybe the README needs to be changed? I'll go check.

sandersn:
So. The README needs to be updated, because I changed the meaning of `libraryName` when I added deprecated-package checking. It is the name of the NPM package that replaces `@types/knex`. That means I should probably change the name of the property. 

The right solution could be to another property, but I don't think there's much value in having a descriptive name. The only thing it's used for is in the deprecation message, which definitely needs to point to the npm source package.

For this PR, that means `libraryName: "knex"` should be correct.

mastermatt:
Tavis is now failing because there are other type packages that import/require `knex` and the compiler doesn't know how to get types from the lib itself.
I haven't been able to figure out the right solution to this.

sandersn:
`bookshelf`, for example, needs to add a `package.json` that refers to knex. Something like this (copied from ui-box):

```json
{
  "private": true,
  "dependencies": {
    "csstype": "^2.6.0"
  }
}
```

You will also have to add knex to https://github.com/Microsoft/types-publisher/blob/master/dependenciesWhitelist.txt. I'll merge that PR and publish a new version of types-publisher, which is the package that DT uses for CI checks.

...And I'll go update the README with these instructions. The current ones don't cover the case that removed packages have dependents at all.

PR #33917, opened 2019-03-15
[react] Allow children inside React.memo

nikolay-borzov:
Fixes the following issue
```
const Select = React.memo<{ id: string }>(props => 
  (<select id={props.id}>{props.children}</select>)
)

<Select id="foo">
  <option value="dog">Dog</option>
  <option value="cat">Cat</option>
</Select>
```

> Type '{ children: string; id: 'foo'; }' is not assignable to type 'IntrinsicAttributes & { id: string; }'.
  Property 'children' does not exist on type 'IntrinsicAttributes & { id: string; }'

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -752,9 +752,9 @@ declare namespace React {
     }
 
     function memo<P extends object>(
-        Component: SFC<P>,
+        Component: FunctionComponent<P>,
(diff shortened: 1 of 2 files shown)


nikolay-borzov:
Log says

> The job exceeded the maximum time limit for jobs, and has been terminated.

ferdaber:
Not a good idea. You should be typing the props interface of your wrapped component properly. See the discussion here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33602

PR #33920, opened 2019-03-15
[duplexify] add 'default' to DuplexifyConstructor

DominicKramer:
In its current form, using `duplexify` in Typescript apps
with a tsconfig that has esModuleInterop set to `true` where the
code has a class that extends `Duplexify` causes a compliation
error.

This is because if esModuleInterop is set to `true`, the
`Duplexify` constructor is accessed through the synthesized
default export of `duplexify`.

By having the `DuplexifyConstructor` have a `default` property
that is itself a `DuplexifyConstrutor`, users can use
`duplexify` with esModuleInterop set to `true` or `false`.

See issue
https://github.com/googleapis/nodejs-logging-bunyan/issues/241
for more context.

### DIFF
--- a/types/duplexify/index.d.ts
+++ b/types/duplexify/index.d.ts
@@ -14,6 +14,7 @@ interface DuplexifyConstructor {
   new (writable?: stream.Writable, readable?: stream.Readable, streamOptions?: stream.DuplexOptions): duplexify.Duplexify;
 
   obj(writable?: stream.Writable, readable?: stream.Readable, streamOptions?: stream.DuplexOptions): duplexify.Duplexify;
+  default: DuplexifyConstructor;
 }
 declare var duplexify: DuplexifyConstructor;
 declare namespace duplexify {


strax (approved):

weswigham (changes requested):
The library [has no member named default](https://github.com/mafintosh/duplexify/blob/master/index.js). If you expect your loader makes the module available as the default, you _must_ use `allowSyntheticDefaultImports` or `esModuleInterop`.

DominicKramer:
@weswigham The problem is that the these types don't appear to allow one to use `esModuleInterop` set to true.

With this `tsconfig.json` file:
```json
{
  "compilerOptions": {
    "rootDir": ".",
    "outDir": "build",
    "lib": ["es2016"],
    "module": "commonjs",
    "esModuleInterop": true
  },
  "include": [
    "src/*.ts"
  ]
}
```
Both 
```js
import {Duplexify} from 'duplexify';

class MyClass extends Duplexify {}
```
and 
```js
import * as Duplexify from 'duplexify';

class MyClass extends Duplexify {}
```
fail to compile.

The first has error:
```
src/index.ts:5:23 - error TS2689: Cannot extend an interface 'Duplexify'. Did you mean 'implements'?

5 class MyClass extends Duplexify {}
```
and the second has error
```
src/index.ts:4:23 - error TS2507: Type '{ default: DuplexifyConstructor; obj(writable?: Writable, readable?: Readable, streamOptions?: DuplexOptions): Duplexify; }' is not a constructor function type.

4 class MyClass extends Duplexify {}
```

weswigham:
With `esModuleInterop` set, you should be using `import Duplexify from 'duplexify';`

PR #33921, opened 2019-03-15
[@types/htmlparser2] Update typings for Htmlparser2 (and related modules) to match updates

johandavidson:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.npmjs.com/package/htmlparser2, https://github.com/fb55/htmlparser2#readme


types/domhandler/domhandler-tests.ts
```diff
@@ -1,14 +1,14 @@
...
     	console.error('There has been an error...');
     else
         console.log(dom);
 });
 handler.ontext = (data: string) => { console.log(data); };
 handler.onreset = () => { console.log('We have a reset.'); };
-handler.onerror = (error: Error) => { console.error(Error); };
-handler.onopentag = (name: string, attribs) => { console.log(name, attribs); };
+handler.onerror = (error: Error) => { console.error(error); };
+handler.onopentag = (name: string, attribs: { [s: string]: string }) => { console.log(name, attribs); };
```
LinusU: _might_ be cleaner with `Record<string, string>` here, it will resolve to the same thing though...

types/htmlparser2/htmlparser2-tests.ts
```diff
@@ -1,23 +1,30 @@
 /**
  * Created by staticfunction on 8/4/14.
  */
-import htmlparser = require("htmlparser2");
+import * as htmlparser from 'htmlparser2';
```
LinusU: I think that `import htmlparser = require("htmlparser2");` is the _correct_ thing to do, since it's an Node.js package with `module.exports = ...` in their source

types/htmlparser2/index.d.ts
```diff
@@ -1,30 +1,20 @@
...
-    oncommentend?: () => void;
-    oncdatastart?: () => void;
-    oncdataend?: () => void;
-    onerror?: (error: Error) => void;
-    onreset?: () => void;
-    onend?: () => void;
-}
-
-export interface Options {
+export interface ParserOptions {
```
LinusU: This will be a breaking change for anyone importing `Options`, but maybe that's okay...

We could add a `@deprecated` `export type Options = ParserOptions` though

types/htmlparser2/index.d.ts
```diff
@@ -63,31 +53,49 @@ export interface Options {
...
 export declare class WritableStream extends Writable {
-    constructor(handler: Handler, options?: Options);
+    constructor(handler: DomHandler);
+    constructor(handler: DomHandler, options: ParserOptions);
 }
 
 export declare class Parser {
-    constructor(handler: Handler, options?: Options);
+    constructor(handler: DomHandler);
+    constructor(handler: DomHandler, options: ParserOptions);
```
LinusU: Splitting this up into two functions won't allow `new Parser(handler, undefined)`, was that intentional?

types/htmlparser2/index.d.ts
```diff
@@ -1,30 +1,20 @@
...
-    ontext?: (text: string) => void;
-    onclosetag?: (text: string) => void;
-    onprocessinginstruction?: (name: string, data: string) => void;
-    oncomment?: (data: string) => void;
-    oncommentend?: () => void;
-    oncdatastart?: () => void;
-    oncdataend?: () => void;
-    onerror?: (error: Error) => void;
-    onreset?: () => void;
-    onend?: () => void;
```
LinusU: The `DomHandler` interface has all function as mandatory, whereas the previous `Handler` interface had each function as optional. I think that they are optional in the library, so we should probably reflect that in the typings...

types/domhandler/domhandler-tests.ts
johandavidson: Yes, it would look nicer but would also require TypeScript 2.1. Maybe we can stick with this for now?

types/htmlparser2/index.d.ts
johandavidson: Yes, this is wrong. I'll change!

types/htmlparser2/index.d.ts
johandavidson: Yes, I guess my inexperience working with this shines through. The @ deprecated decorator doesn't work though. Would you consider `/* @deprecated */ export type Options = ParserOptions` or maybe stick with the old naming? (I guess this will be braking changes anyway since the old handler and the new DomHandler isn't the same. I just wanted to clearly separate the DomHandlerOptions from the options for the parser)

types/htmlparser2/index.d.ts
johandavidson: No, those functions are added to `DomHandler` in the code implementation (just as they are in the `Parser`, see for example the `Parser`s `write` method). They are not optional, they are there. Maybe the old `Handler` interface reflected how `htmlparser2` was implemented before the handler was extracted to its own module?

types/htmlparser2/htmlparser2-tests.ts
johandavidson: You are right. Otherwise it would require the code to be compiled to CommonJS. I'll change.

types/domhandler/domhandler-tests.ts
WillGibson: TypeScript 2.1 have been out over two years now, I think it would be perfectly acceptable to have it as a minimum requirement.

Out of interest, do the tests run with TypeScript 2 and 3?

types/sanitize-html/index.d.ts
```diff
@@ -6,9 +6,10 @@
 //                 Rinze de Laat <https://github.com/biermeester>
 //                 Will Gibson <https://github.com/WillGibson>
 //                 A penguin <https://github.com/sirMerr>
+//                 Johan Davidsson <https://github.com/johandavidson>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-import {Options} from "htmlparser2";
+import {ParserOptions} from "htmlparser2";
```
WillGibson: As @LinusU  commented at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33921/files#r266197648, this could be a breaking change. I'm fine with it from a personal perspective, but I'm not certain what the implications would be for everyone using `sanitize-html`, especially if their test coverage is not up to scratch.

I can't do it right now, but will try and run these changes through our tests in the codebase where we use  `sanitize-html`.

johandavidson: Ok, lets go the safer way and use the deprecated Options. Then we won't need to change those other modules. I'll change.

johandavidson: About TypeScript 2.1: If we add the requirement then also all modules referencing this module needs to add the same requirement. I think that would be too much of a change right now.

types/domhandler/domhandler-tests.ts
johandavidson: If we add the TypeScript 2.1 requirement then all modules referencing this module also needs to add the same requirement. I think that might be too much of a change right now.

Eyas:
Hey @johandavidson -- this doesn't seem to match the latest htmlparser2 as I read it, in particular:

https://github.com/fb55/htmlparser2/blob/master/lib/Parser.js#L216-L222

1. With this PR, htmlparser2 takes a DomHandler from the 'domhandler' package. But a DomHandler has a bunch of functions, whereas the htmlparser2 handler has a bunch of optional? functions.

2. htmlparser2 passes a tagName to onclosetag, but DomHandler's onclosetag takes no parameters.

3. there's also domhandler being a class rather than just an object literal

Happy to send a follow-up fix but I'm not sure: (a) what the motivation of changing htmlparser2 to use the DomHandler package was initially? and (b) if the correct "fix" is just to revert that part of the PR or propagate a fix to domhandler.

johandavidson:
Hi @Eyas ,
Happy to help if possible!
1. htmlparser2 seems to have been updated since the original d.ts was created. htmlparser2 is not having its own handler anymore but instead is importing a DomHandler (https://github.com/fb55/htmlparser2/blob/master/lib/index.js#L2)
2. Domhandler actually doesn't have a parameter on the onclosetag function (https://github.com/fb55/domhandler/blob/master/index.js#L58)
3. Not sure why it is wrong?

The motivation to change htmlparser2 to use DomHandler instead is briefly described [here](https://www.npmjs.com/package/htmlparser2)

> The DomHandler, while still bundled with this module, was moved to its own module. Have a look at it for further information.

I hope this information is helpful. Have a nice day!

PR #33922, opened 2019-03-15
Fix return types on several column() functions.

CNBoland:
- Fix return type for [column().index()](https://datatables.net/reference/api/column().index()).
- Correct return type of [column().footer()](https://datatables.net/reference/api/column().footer()).
- Correct return type of [column().nodes()](https://datatables.net/reference/api/column().nodes()). 

Please fill in this template.

- [ x ] Provide a URL to documentation or source code which provides context for the suggested changes: Relevant documentation URLs included in commit and PR comments.
- [ x ] Increase the version number in the header if appropriate.
- [ x ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

### DIFF
--- a/types/datatables.net/index.d.ts
+++ b/types/datatables.net/index.d.ts
@@ -704,7 +704,7 @@ declare namespace DataTables {
         /**
          * Get the footer th / td cell for the selected column.
          */
-        footer(): any;
+        footer(): Node;
 
         /**
          * Get the header th / td cell for a column.
@@ -766,12 +766,12 @@ declare namespace DataTables {
          *
          * @param t Specify if you want to get the column data index (default) or the visible index (visible).
          */
-        index(t?: string): Api;
+        index(t?: string): number;
 
         /**
          * Obtain the th / td nodes for the selected column
          */
-        nodes(): Api[];
+        nodes(): Api;
     }
 
     interface ColumnsMethodsModel {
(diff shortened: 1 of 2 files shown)


types/datatables.net/index.d.ts
```diff
@@ -704,7 +704,7 @@ declare namespace DataTables {
         /**
          * Get the footer th / td cell for the selected column.
          */
-        footer(): any;
+        footer(): Node;
```
weswigham: You seem to cast this often - if it really returns the footer th/td cell, doesn't that make it always an `HTMLElement`, and not just a `Node`?

CNBoland: I don't disagree with you, but this fits the [documentation](https://datatables.net/reference/api/column().footer()), both in the JSDoc comments, as well as the function return type. Seeing the return type of `any` caught my attention, so I looked it up.

weswigham: The usage of [`node`](https://datatables.net/reference/type/node) in their docs seems to signify only html elements.

> Where a parameter is shown as accepting a node type, or a method returning a node type, it indicates that an object is expect / returned which refers to a DOM element.

weswigham: Minimally, the docs say that for here specifically,

> This may be a th or td element depending on the HTML for your table.

which means it is _always_ an HTMLElement and that `Node` is needlessly general.

CNBoland: I see your point and agree. I made changes accordingly. Both `header()` and `footer()` functions return `HTMLElement` now. Thank your for the clarification.

PR #33925, opened 2019-03-15
next: Add GetInitialProps type

ClassicDarkChocolate:
Add type GetInitialProps<IP, C>. Can be use like the example:

```js
interface IProps extends IInitialProps {}

class Page extends React.Component<IProps, IState> {
    public static getInitialProps: GetInitialProps<
        IInitialProps,
        NextContext
    > = (ctx) => {
        return {};
    };
}
```

---

Please fill in this template.

### DIFF
--- a/types/next/error.d.ts
+++ b/types/next/error.d.ts
@@ -10,5 +10,5 @@ export interface DefaultErrorIProps {
 }
 
 export default class Error<P = {}> extends React.Component<P & DefaultErrorIProps> {
-    static getInitialProps: GetInitialProps<NextContext, DefaultErrorIProps>;
+    static getInitialProps: GetInitialProps<DefaultErrorIProps, NextContext>;
 }
--- a/types/next/error.d.ts
+++ b/types/next/error.d.ts
@@ -1,5 +1,5 @@
 import * as React from "react";
-import { NextContext } from ".";
+import { GetInitialProps, NextContext } from ".";
 
 /**
  * Initial props returned from base Error class.
@@ -10,5 +10,5 @@ export interface DefaultErrorIProps {
 }
 
 export default class Error<P = {}> extends React.Component<P & DefaultErrorIProps> {
-    static getInitialProps(context: NextContext): Promise<DefaultErrorIProps> | DefaultErrorIProps;
+    static getInitialProps: GetInitialProps<NextContext, DefaultErrorIProps>;
 }
--- a/types/next/index.d.ts
+++ b/types/next/index.d.ts
@@ -14,8 +14,8 @@
 
 import * as http from "http";
 import * as url from "url";
-import { Server as NextServer, ServerOptions as NextServerOptions, RenderOptions } from 'next-server';
-import { NextConfig as NextServerConfig } from 'next-server/next-config';
+import { Server as NextServer, ServerOptions as NextServerOptions, RenderOptions } from "next-server";
+import { NextConfig as NextServerConfig } from "next-server/next-config";
 import { Response as NodeResponse } from "node-fetch";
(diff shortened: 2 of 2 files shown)


types/next/index.d.ts
```diff
@@ -131,8 +131,16 @@ declare namespace next {
...
+        getInitialProps?: GetInitialProps<IP, C>;
     }
+
+    /**
+     * Next.js getInitialProps type.
+     *
+     * @template IP Initial props returned from getInitialProps and passed to the component.
+     * @template C Context passed to getInitialProps.
+     */
+    type GetInitialProps<IP, C> = (ctx: C) => Promise<IP> | IP;
```
resir014: I think this `C` should be defaulted to `NextContext`. I have a plan to extend all context objects from this base interface and it makes sense to default to the base interface.

types/next/error.d.ts
```diff
@@ -10,5 +10,5 @@ export interface DefaultErrorIProps {
 }
 
 export default class Error<P = {}> extends React.Component<P & DefaultErrorIProps> {
-    static getInitialProps(context: NextContext): Promise<DefaultErrorIProps> | DefaultErrorIProps;
+    static getInitialProps: GetInitialProps<NextContext, DefaultErrorIProps>;
```
resir014: Did you mean for this to go the other way around? (`GetInitialProps<IP, C>`)

resir014 (changes requested):
Hey, thanks! Some questions/changes before this is good to go.

resir014:
Also, tests are failing in the `next/error` tests:

```
Error in next
Error: /home/travis/build/DefinitelyTyped/DefinitelyTyped/types/next/test/next-error-tests.tsx:10:7
ERROR: 10:7  expect  TypeScript@next compile error: 
Class static side 'typeof MyError' incorrectly extends base class static side 'typeof Error'.
  Types of property 'getInitialProps' are incompatible.
    Type '() => { statusCode: number; foo: string; }' is not assignable to type 'GetInitialProps<NextContext<Record<string, string | string[] | undefined>>, DefaultErrorIProps>'.
      Type '{ statusCode: number; foo: string; }' is not assignable to type 'NextContext<Record<string, string | string[] | undefined>> | Promise<NextContext<Record<string, string | string[] | undefined>>>'.
        Type '{ statusCode: number; foo: string; }' is missing the following properties from type 'Promise<NextContext<Record<string, string | string[] | undefined>>>': then, catch, [Symbol.toStringTag], finally
```

types/next/error.d.ts
ClassicDarkChocolate: Oops, I should not have done this kind of mistake. Fixed.

types/next/index.d.ts
ClassicDarkChocolate: I think this would fail on App.getInitialProps (_app.tsx)

resir014: @ClassicDarkChocolate You might be able to fix it by having `NextAppContext` extend `NextContext`, which is what I've been planning to do at some point.

types/next/error.d.ts
resir014: It's fine! And thanks for fixing.

types/next/index.d.ts
ClassicDarkChocolate: @resir014 I think they're not compatible and have different structure.

resir014: I see. Let's keep it this way for now.

resir014 (approved):
LGTM 👍 Other contributors can also add their feedbacks before we put this in.

scottdj92 (approved):

PR #33931, opened 2019-03-16
node-forge: fixed local jsbn to use actual jsbn types to fix BigInteger toString options.

mharj:
This removes local jsbn.BigInteger types and uses jsbn module types directly.
⚠️ Please check that this is correctly added as I'm still quite new to modify those types.⚠️

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/andyperlitch/jsbn/blob/master/index.js#L177

### DIFF
--- a/types/node-forge/index.d.ts
+++ b/types/node-forge/index.d.ts
@@ -29,7 +29,7 @@ declare module "node-forge" {
             data: number[];
             t: number;
             s: number;
-            toString(): string;
+            toString(b?: number): string;
         }
     }
 
--- a/types/node-forge/index.d.ts
+++ b/types/node-forge/index.d.ts
@@ -14,7 +14,7 @@
 // TypeScript Version: 2.6
 
 /// <reference types="node" />
-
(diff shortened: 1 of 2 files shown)


types/node-forge/index.d.ts
```diff
@@ -14,7 +14,7 @@
 // TypeScript Version: 2.6
 
 /// <reference types="node" />
-
+import * as jsbn from "jsbn";
```
timhwang21: We need to add a dependency on `@types/jsbn` right?

https://github.com/DefinitelyTyped/DefinitelyTyped#i-notice-some-packages-having-a-packagejson-here

andersk:
Forge does not use the `jsbn` module, it has its own copy—are you sure we should share the types?

mharj:
> Forge does not use the `jsbn` library, it has its own copy—are you sure we should share the types?

Oh, my initial commit was just fix to "toString()" as it's missing radix option, but I though from namespace that it's same module instead of copy.

PR #33950, opened 2019-03-17
feat(passport-saml): Make attributeName unknown

josecolella:
Please fill in this template.

- [X ] Test the change in your own code. (Compile and run.)
- [ Z] Add or edit tests to reflect the change. (Run with `npm test`.)
- [Z ] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/passport-saml/index.d.ts
+++ b/types/passport-saml/index.d.ts
@@ -3,8 +3,9 @@
 // Definitions by: Chris Barth <https://github.com/cjbarth>
 //                 Damian Assennato <https://github.com/dassennato>
 //                 Karol Samborski <https://github.com/ksamborski>
+//                 Jose Colella <https://github.com/josecolella>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+// TypeScript Version: 3.0
 
 import passport = require('passport');
 import express = require('express');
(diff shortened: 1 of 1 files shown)


types/passport-saml/index.d.ts
```diff
@@ -97,5 +98,5 @@ export type Profile = {
 	getAssertion(): object;  // get the assertion XML parsed as a JavaScript object
 	getSamlResponseXml(): string; // get the raw SAML response XML
 } & {
-	[attributeName: string]: string;  // arbitrary `AttributeValue`s
+	[attributeName: string]: unknown;  // arbitrary `AttributeValue`s
```
cjbarth: I don't believe this is correct. When the XML is parsed, the result is a `string`. I suppose it could be a `number`, but it certainly isn't `unknown`.

cjbarth (changes requested):

types/passport-saml/index.d.ts
josecolella: @cjbarth If you want to expand the profile object, the current type forces the user to only have string values. For example, my Profile object has a `roles` attribute which is an array of string.

josecolella: @cjbarth Why would we force the string value for all attributeNames. `unknown` still allows for typing enforcement and allows for expansion of profile object.

cjbarth: That is a good use-case, we probably need to update the docs then: https://github.com/bergie/passport-saml#configure-strategy-for-multiple-providers

cjbarth: Could you make a PR to update the docs?

josecolella: @cjbarth https://github.com/bergie/passport-saml/pull/363. PR created.

cjbarth (approved):

PR #33965, opened 2019-03-17
[node] fix static-this lint issues

Flarna:

Fix lint issues because use of `this` in static context. Most likely an update of tslint caused them to appear without an actual change in code. 

Failure details (see also #33884):
```
Error in node
Error: /home/travis/build/DefinitelyTyped/DefinitelyTyped/types/node/test/util.ts:92:42
ERROR: 92:42  static-this  Use the parent class name instead of `this` when in a `static` context.
ERROR: 93:43  static-this  Use the parent class name instead of `this` when in a `static` context.
ERROR: 94:44  static-this  Use the parent class name instead of `this` when in a `static` context.
ERROR: 95:45  static-this  Use the parent class name instead of `this` when in a `static` context.
ERROR: 96:49  static-this  Use the parent class name instead of `this` when in a `static` context.
ERROR: 97:50  static-this  Use the parent class name instead of `this` when in a `static` context.
ERROR: 98:51  static-this  Use the parent class name instead of `this` when in a `static` context.
ERROR: 99:52  static-this  Use the parent class name instead of `this` when in a `static` context.
```

### DIFF
--- a/types/node/test/util.ts
+++ b/types/node/test/util.ts
@@ -89,14 +89,14 @@ import { readFile } from 'fs';
         }
 
         static test(): void {
-            const cfn = util.callbackify(this.fn);
-            const cfnE = util.callbackify(this.fnE);
-            const cfnT1 = util.callbackify(this.fnT1);
-            const cfnT1E = util.callbackify(this.fnT1E);
-            const cfnTResult = util.callbackify(this.fnTResult);
-            const cfnTResultE = util.callbackify(this.fnTResultE);
-            const cfnT1TResult = util.callbackify(this.fnT1TResult);
-            const cfnT1TResultE = util.callbackify(this.fnT1TResultE);
+            const cfn = util.callbackify(callbackifyTest.fn);
+            const cfnE = util.callbackify(callbackifyTest.fnE);
+            const cfnT1 = util.callbackify(callbackifyTest.fnT1);
+            const cfnT1E = util.callbackify(callbackifyTest.fnT1E);
+            const cfnTResult = util.callbackify(callbackifyTest.fnTResult);
+            const cfnTResultE = util.callbackify(callbackifyTest.fnTResultE);
+            const cfnT1TResult = util.callbackify(callbackifyTest.fnT1TResult);
+            const cfnT1TResultE = util.callbackify(callbackifyTest.fnT1TResultE);
 
             cfn((err: NodeJS.ErrnoException, ...args: string[]) => assert(err === null && args.length === 1 && args[0] === undefined));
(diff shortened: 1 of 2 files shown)


Flarna:
@sandersn I wonder why these fails are not visible in the nightly runs at https://typescript.visualstudio.com/TypeScript/_build?definitionId=15

j-oliveras (approved):

SimonSchick:
@Flarna isn't this rule kind of counter productive? If this was enforced in an actual code base overridden static calls wouldn't work anymore.

Flarna:
@SimonSchick It's only in tests here so no impact to any real world project.

This rule was added in tslint 5.14.0 and as dtslint uses the `all` config it got automatically enabled for all tests in this repo.

I expect the rule is there because the typescript/javascript solution to deal with overriding static methods differs from a lot other languages and therefore a lot people get confused so better avoid this pattern.

BrunoScheufler (approved):

r3nya (approved):

weswigham:
Re: the nightly run. The nightly run only runs the `ExpectType` rule because we're only looking to see if ts itself broke any types.

Flarna:
@weswigham Would be nice to have a nightly run which checks lint also. Similar as typescript also tslint gets updated without a change in this repo.

weswigham:
We actually were having difficulty running a nightly that runs all rules - it takes too long (IIRC more than the few hours a container is allotted)

sandersn:
The nightly run actually runs lints as well, since about a week ago. I decided that the value for keeping DT clean was worth the work we on the TS team will have to keep it clean for regression testing. It runs in 1h48 instead of 1h24, which is well below the 6 hour limit for now.

But that means node should have failed along with everything else. I don't know why it didn't.

```
node OK
node/v0 OK
node/v4 OK
node/v6 OK
node/v10 OK
node/v7 OK
node/v8 OK
node-7z OK
node/v9 OK
```

Flarna:
That's strange, it failed for node v8 and v9 but not for v10 and latest... And since 15.3. even v8 and v9 don't fail anymore...
Seems I fixed this only partial, will take a look into v8/v9 and create a PR if needed. Or is it planned to disable this lint rule globally?

Flarna:
Found the reason why v8/v9 was no longer failing: #33848 
Best guess why v10/latest fails is because tests have been slitted into multiple files.

sandersn:
I certainly find static-this unnecessarily confusing, so I'll keep the rule unless there's a good scenario that's prevented by it. However, when I have time I'm going to stop using `tslint:all` as the basis for Definitely Typed, replacing it with whatever rules it evaluates to right now.

PR #33982, opened 2019-03-18
Fix return type of `Document.model(str)`, and fix #33824.

### LINKED ISSUE #33824: [mongoose] mongoose.Model<...> does not satisfy the constraint Document
**Version:**
- `@types/mongoose`: 4.7.44

**Code**

```ts
interface Foo extends mongoose.Document {
  bar: number;
}

type Entry = mongoose.Model<Foo>;
```

**Expected behavior:**
It should be fine.

**Actual behavior:**
I get the error `Type 'Foo' does not satisfy the constraint 'Document'.`

With version 4.7.43 the code works.

- [x] I tried using the `"@types/mongoose@4.7.44` package and had problems.
  - @simonxca 
  - @horiuchi 
  - @lukasz-zak 
  - @murbanowicz

jwalton:
- [x] URL to documentation or source code which provides context for the suggested changes: https://mongoosejs.com/docs/4.x/docs/api.html#model_Model-model

The `Document.model(name)` function and `Model.model(name)` function return *another* model, so the type of that other model may not be the same as the type of the current model.

This also fixes #33824.

### DIFF
--- a/types/mongoose/index.d.ts
+++ b/types/mongoose/index.d.ts
@@ -1063,8 +1063,8 @@ declare module "mongoose" {
     /** defaults to "__v" */
     versionKey?: string | boolean;
     /**
-     * By default, Mongoose will automatically 
-     * select() any populated paths. 
+     * By default, Mongoose will automatically
+     * select() any populated paths.
      * To opt out, set selectPopulatedPaths to false.
      */
     selectPopulatedPaths?: boolean;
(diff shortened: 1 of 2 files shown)


murbanowicz (approved):
Looks good to me. Thank you. I did not find this side effect when introducing my improvement.

u84six:
I'm still getting this error using version 5.5.41. Is this still pending?

itsmoirob:
@u84six A work around is to extend your interface with the mongoose.Document class. IE

```import { Document } from 'mongoose';

export interface Cat extends Document {
    name: string;
}```

PR #33988, opened 2019-03-18
Updated react-native-canvas

hmajid2301:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/Image, https://developer.mozilla.org/en-US/docs/Web/API/ImageData/ImageData
https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement


types/react-native-canvas/index.d.ts
```diff
@@ -176,8 +176,8 @@ export class Image {
 export class ImageData {
     constructor(canvas: Canvas, data: number[], height: number, width: number);
     readonly data: number[];
-    readonly height: number;
-    readonly width: number;
+    readonly height: number | undefined;
```
sheetalkamat: Please provide documentation why these can be undefined. If these can be undefined, does it mean constructor can be invoked without height and width being passed or could it pass undefined (optional or required parameter accepting undefined)

types/react-native-canvas/index.d.ts
```diff
@@ -176,8 +176,8 @@ export class Image {
...
+    readonly width: number | undefined;
```
sheetalkamat: Please provide documentation why these can be undefined. If these can be undefined, does it mean constructor can be invoked without height and width being passed or could it pass undefined (optional or required parameter accepting undefined)

sheetalkamat (changes requested):

types/react-native-canvas/index.d.ts
hmajid2301: I've added links to documentation for the changes. Thanks

PR #33989, opened 2019-03-19
[@types/redux-seamless-immutable] Add definitions

SoaresMG:

If adding a new definition:

Package: [redux-seamless-immutable](https://github.com/eadmundo/redux-seamless-immutable)

### DIFF
--- a/types/redux-seamless-immutable/index.d.ts
+++ b/types/redux-seamless-immutable/index.d.ts
@@ -1,0 +1,16 @@
+// Type definitions for redux-seamless-immutable 0.4
+// Project: https://github.com/eadmundo/redux-seamless-immutable
+// Definitions by: Leandro Soares <https://github.com/SoaresMG>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.3
+
+import { Reducer, Action } from "redux";
+import { Immutable } from "seamless-immutable";
+
+export interface SeamlessReducers {
+    [reducerName: string]: Reducer<any, any>;
+}
+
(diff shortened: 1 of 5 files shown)


types/redux-seamless-immutable/index.d.ts
```diff
@@ -0,0 +1,18 @@
+// Type definitions for 0.4.0
+// Project: https://github.com/eadmundo/redux-seamless-immutable
+// Definitions by: Leandro Soares <https://github.com/SoaresMG>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.3
+
+declare module "redux-seamless-immutable" {
```
sheetalkamat: Do not declare module here. Just take everything here outside the module declaration.

types/redux-seamless-immutable/tslint.json
```diff
@@ -0,0 +1,79 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
```
sheetalkamat: New package cannot override any rules. Use only extends rule and remove everything else and fix errors if any.

sheetalkamat (changes requested):

types/redux-seamless-immutable/index.d.ts
SoaresMG: Btw how does it work if there are no modules declared? I saw some packages doing it.

sheetalkamat: Your `index.d.ts` is defining shape of `redux-seamless-immutable` module

sheetalkamat (approved):

PR #33195, opened 2019-02-19
Node: specify `UnhandledRejectionListener` `reason` param type

OliverJAsh:
5th attempt at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33189

PR #33997, opened 2019-03-19
Node: tighten URL query type

OliverJAsh:
See the test I added which was previously failing.

Note:
- I had to use a faux `unknown` type since the types don't currently support TS 3.0. I noticed types can be specified as augmentations for TS 3.0 support, but I don't think it's possible to override an object's property in this way.
- I had to add a TS version comment due to weird lint errors in TS 2.0.

---

Please fill in this template.


types/node/node-tests.ts
```diff
@@ -505,6 +505,12 @@ import Module = require("module");
     {
         const path: url.URL = url.pathToFileURL('file://test');
     }
+
+    {
+        const urlObjectQuery: url.UrlObject['query'] = {};
+        // $ExpectError
+        const urlQuery: url.Url['query'] = urlObjectQuery;
+    }
```
OliverJAsh: For some reason `$ExpectError` doesn't seem to be working here:

```
$ npm run lint node

> definitely-typed@0.0.3 lint /Users/oliverjash/Development/DefinitelyTyped
> dtslint types "node"

../node-tests.ts(513,15): error TS2322: Type 'string | { [key: string]: {} | null | undefined; } | null | undefined' is not assignable to type 'string | ParsedUrlQuery | null | undefined'.
  Type '{ [key: string]: {} | null | undefined; }' is not assignable to type 'string | ParsedUrlQuery | null | undefined'.
    Type '{ [key: string]: {} | null | undefined; }' is not assignable to type 'string'.

    at /Users/oliverjash/Development/DefinitelyTyped/node_modules/dtslint/bin/index.js:210:19
    at fulfilled (/Users/oliverjash/Development/DefinitelyTyped/node_modules/dtslint/bin/index.js:5:58)
    at <anonymous>
```

types/node/url.d.ts
```diff
@@ -17,7 +17,7 @@ declare module "url" {
     // Input to `url.format`
     interface UrlObject extends UrlObjectCommon {
         port?: string | number;
-        query?: string | null | { [key: string]: any };
```
OliverJAsh: To summarise the problem and fix:

``` ts
type ParsedUrlQuery = { [key: string]: string }

// Before
{
    declare const object: { [key: string]: any };
    // No error! :-(
    const parsedUrlQuery: ParsedUrlQuery = object;
}

// After
{
    declare const object: { [key: string]: unknown };
    // Error! :-(
    const parsedUrlQuery: ParsedUrlQuery = object;
}
```

types/node/index.d.ts
```diff
@@ -38,6 +38,7 @@
 //                 Jordi Oliveras Rovira <https://github.com/j-oliveras>
 //                 Thanik Bhongbhibhat <https://github.com/bhongy>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.1
```
OliverJAsh: The build now errors with:

> Error: accepts depends on node but has a lower required TypeScript version.

What to do?

The reason I had to add this was because without it, `npm run lint node` resulted in lots of weird errors. 🤷‍♂️ E.g.:

```
Error: /Users/oliverjash/Development/DefinitelyTyped/types/node/node-tests.ts:510:15
ERROR: 510:15  expect  Compile error in typescript@2.0 but not in typescript@2.1.
Fix with a comment '// TypeScript Version: 2.1' just under the header.
Type 'string' is not assignable to type 'UrlObject[]'.
ERROR: 510:15  expect  Compile error in typescript@2.0 but not in typescript@2.1.
Fix with a comment '// TypeScript Version: 2.1' just under the header.
Cannot redeclare block-scoped variable 'urlObjectQuery'.
…
```

These seem like invalid errors, and I'm not sure how to workaround this.

types/node/querystring.d.ts
```diff
@@ -10,7 +10,11 @@ declare module "querystring" {
 
     interface ParsedUrlQuery { [key: string]: string | string[]; }
 
-    function stringify(obj?: {}, sep?: string, eq?: string, options?: StringifyOptions): string;
+    // When these types support TypeScript 3.0+, we can replace this with `unknown`.
+    type PoorMansUnknown = {} | null | undefined;
+    interface ParsedUrlQueryInput { [key: string]: PoorMansUnknown; }
```
OliverJAsh: Alternatively we could do `{ [key: string]: string | string[] | number | number[] | boolean | boolean[]; }`, since these are all the types `querystring.stringify` and `url.format` will implicitly convert to a `string`:

``` sh-session
$ node
> require('querystring').stringify({ string: 'string', number: 1, boolean: true, numbers: [1,2], booleans: [true, false], strings: ['string1', 'string2'] })
'string=string&number=1&boolean=true&numbers=1&numbers=2&booleans=true&booleans=false&strings=string1&strings=string2'
```

IIUC, other value types like objects will be omitted:

``` sh-session
$ node
> require('querystring').stringify({ object: { foo: 1 } })
'object='
```

Currently with `unknown` we allow these types to be passed in—after all it doesn't generate a runtime error, so it is "safe".

https://devdocs.io/node/querystring#querystring_querystring_stringify_obj_sep_eq_options

types/node/index.d.ts
SimonSchick: We cannot change the min version here as too many packages rely on it.

SimonSchick (changes requested):

types/node/index.d.ts
SimonSchick: Unless you are willing to update all dependant packages.

types/node/querystring.d.ts
SimonSchick: I agree here, the `PoorMansUnknown` is a huge hack (as suggested by the name itself), we should wait for node typings to move beyond TS 3.x instead.

SimonSchick (changes requested):

types/node/querystring.d.ts
OliverJAsh: By agreeing, are you saying to replace `PoorMansUnknown` with `string | string[] | number | number[] | boolean | boolean[]`?

FWIW, I don't think it's a huge hack—there are no downsides to using it, other than the lack of a name. Even TypeScript itself uses it when it can't infer a generic.

I think we should consider the desired behaviour and then use that to pick between the two options here.

types/node/index.d.ts
OliverJAsh: Can you advise on the 2.0 lint errors?

Flarna: TS 2.1 has introduced keyof and Lookup Types which is needed for e.g. `url.UrlObject['query']`. You could use
```
const urlObjectQuery: string | null | querystring.ParsedUrlQueryInput = {};
// $ExpectError
const urlQuery: string | null | querystring.ParsedUrlQuery = urlObjectQuery;
```

OliverJAsh: Thanks @Flarna. It doesn't test quite the same thing, but it will do!

types/node/querystring.d.ts
Flarna: Someone may a "query builder class" which works with `PoorMansUnknown` but not with `string | string[] | number | number[] | boolean | boolean[]`:
```
class MyQuery {
    val: string;
    constructor(a: number) {
         this.val = `number${a}`;
    }
}
```

OliverJAsh:
Ping @SimonSchick or someone else who can approve this

SimonSchick:
Sorry I'm too biased here, I won't approve this with the `PoorMansUnknown` in it, we should aim to push the 3.x typings instead.

Please ping someone else if you wish this to be approved, feel free to ignore my rejection.

OliverJAsh:
Can we get a consensus on this from other Node maintainers?

Options:

1. Ignore this bug for now, push to TS 3 typings, and then make this change afterwards (when we'll be able to use `unknown` instead of `PoorMansUnknown`)
2. Start benefitting from this change today, and continue to push for TS 3 typings

My two cents: 2 is the best option, given that `PoorMansUnknown` isn't that much of a hack/workaround: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33997#discussion_r266882699.

Flarna:
I think waiting for TS 3.0 as the minimum typescript version is waste of time. There were several attempts to raise this version but even raising it to 2.1 (`keyof`, `Partial`) was not possible.

Another option to avoid the `PoorMansUnknown` would be to keep the current implementation for <3.2 and use an `unknown` based for >=3.2. See #34016 where I added a real type guard for >=3.2 and a function return a boolean for <3.2.

But I have no strong opinion here and I'm also fine with the solution as as. Maybe `PoorMansUnknown` should be inlined to `ParsedUrlQueryInput` (not sure if the `| unknown` is needed) to avoid that it spreads to other locations.

OliverJAsh:
Note we're also using the poor man's `unknown` trick elsewhere in the Node typings: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33195/files#diff-a2f9a5377787f7084c7f52b20c0108cfR682.

> keep the current implementation for <3.2 and use an `unknown` based for >=3.2.

Hmm, interesting. However I'm not sure this gains us anything? Either way we're going to have `PoorMansUnknown` in the typings. 🤔 

> Maybe `PoorMansUnknown` should be inlined to `ParsedUrlQueryInput`

I'm happy to do that! I was trying to make the code self documenting so it would be clear that this needs to be changed to `unknown` when the TS upgrade eventually happens—but I realise this means it could leak into other usages. I'll inline it and add a comment instead.

Flarna:
> Note we're also using the `PoorMansUnknown` trick elsewhere

Yes, and I can remember that there were quite some discussions about it at the time you raised this PR :o)

> Hmm, interesting. However I'm not sure this gains us anything? Either way we're going to have `PoorMansUnknown` in the typings. 🤔

With current I mean that one before your PR.

Flarna (approved):

BrunoScheufler (approved):

OliverJAsh:
@tsd-bot @sandersn This one seems to have got stuck?

OliverJAsh:
@RyanCavanaugh @sandersn This one seems to have got stuck?

PR #33998, opened 2019-03-19
Initial definition for @wordpress/jest-console

mistic100:
Please fill in this template.

If adding a new definition:
- [ ] Create it with `dts-gen --dt`, not by basing it on an existing project. (**manually created**)


mistic100:
The CI reports "Expected file 'jest-console-tests.ts' to be named wordpress__jest-console-tests.ts" be the file is already correctly named, looks like it does not pick my last commit.

types/wordpress__jest-console/index.d.ts
```diff
@@ -0,0 +1,51 @@
...
+    interface Matchers<R> {
+        /**
+         * Ensure that `console.error` function was called.
+         */
+        toHaveErrored(): R;
+
+        /**
+         * Ensure that `console.error` function was called with specific arguments.
+         */
+        toHaveErroredWith(...args: any[]): R;
```
sheetalkamat: I think the arguments are suppose to be `string`. Does this method require at least one argument?

sheetalkamat (changes requested):

types/wordpress__jest-console/index.d.ts
mistic100: You can do `console.error(5)` if you want.

The code does not make any assumption on the number of arguments passed to the "With" methods https://github.com/WordPress/gutenberg/blob/master/packages/jest-console/src/matchers.js#L42  

I guess it's worth enforcing at least one argument. Toughts ?

PR #34010, opened 2019-03-19
Fix glob statCache to use fs.Stats

styfle:
Please fill in this template.

- [x] Provide URL to documentation which provides context for the suggested change: [`statCache` option](https://github.com/isaacs/node-glob/blob/master/README.md#options)

### DIFF
--- a/types/glob/index.d.ts
+++ b/types/glob/index.d.ts
@@ -4,6 +4,7 @@
 //                 voy <https://github.com/voy>
 //                 Klaus Meinhardt <https://github.com/ajafff>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.9
 
 /// <reference types="node" />
 
--- a/types/glob/index.d.ts
+++ b/types/glob/index.d.ts
@@ -7,6 +7,9 @@
 
 /// <reference types="node" />
 
+import * as fs from "fs";
+import { Stats } = fs;
+
 import events = require("events");
 import minimatch = require("minimatch");
 
@@ -34,7 +37,7 @@ declare namespace G {
         silent?: boolean;
         strict?: boolean;
         cache?: { [path: string]: boolean | 'DIR' | 'FILE' | ReadonlyArray<string> };
-        statCache?: { [path: string]: false | { isDirectory(): boolean} | undefined };
+        statCache?: { [path: string]: false | Stats | undefined };
         symlinks?: { [path: string]: boolean | undefined };
         realpathCache?: { [path: string]: string };
         sync?: boolean;
--- a/types/glob/index.d.ts
+++ b/types/glob/index.d.ts
@@ -8,7 +8,7 @@
 /// <reference types="node" />
 
 import * as fs from "fs";
-import { Stats } = fs;
+import Stats = fs.Stats;
 
 import events = require("events");
 import minimatch = require("minimatch");
--- a/types/glob/index.d.ts
+++ b/types/glob/index.d.ts
@@ -74,7 +74,7 @@ declare namespace G {
         options: IOptions;
         aborted: boolean;
         cache: { [path: string]: boolean | 'DIR' | 'FILE' | ReadonlyArray<string> };
-        statCache: { [path: string]: false | { isDirectory(): boolean; } | undefined };
+        statCache: { [path: string]: false | Stats | undefined };
         symlinks: { [path: string]: boolean | undefined };
         realpathCache: { [path: string]: string };
         found: string[];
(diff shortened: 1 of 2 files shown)


types/glob/index.d.ts
```diff
@@ -4,9 +4,13 @@
 //                 voy <https://github.com/voy>
 //                 Klaus Meinhardt <https://github.com/ajafff>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.9
```
ajafff: why is this change necessary?

types/glob/index.d.ts
```diff
@@ -34,7 +38,7 @@ declare namespace G {
         silent?: boolean;
         strict?: boolean;
         cache?: { [path: string]: boolean | 'DIR' | 'FILE' | ReadonlyArray<string> };
-        statCache?: { [path: string]: false | { isDirectory(): boolean} | undefined };
+        statCache?: { [path: string]: false | Stats | undefined };
```
ajafff: I intentionally changed it to this type. I use this option as input to mock certain files. Since the implementation only uses `isDirectory` I don't need to provide a complete implementation of `Stats`.

I guess your use case is the exact opposite and you want to reuse the statCache in your app to avoid calling `fs.stat` again?

ajafff (changes requested):
This is kind of a breaking change (at least for me).
It would help to know your use case and reasoning for this change.

types/glob/index.d.ts
styfle: > I guess your use case is the exact opposite and you want to reuse the statCache in your app to avoid calling fs.stat again?

Correct 👍

types/glob/index.d.ts
styfle: I'm not sure. I believe this happened when importing Stat 🤔 

Take a look at the previous CI: https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/508680421

styfle: @ajafff I just changed the import so it works with all TS Versions again 👍

styfle:
> It would help to know your use case and reasoning for this change.

My use case is almost identical to the test I added.

I am reading from the `statCache` so this PR is fixing the type declaration to be accurate by adding the missing methods/properties such as `isFile()`, `mode`, etc.

styfle:
@vvakame Any thoughts on this? As far as I can tell, this is the correct type.

weswigham:
@styfle could you not 
```ts
	const pattern = "test/**";
    const options: glob.IOptions & { statCache: {[idx: string]: fs.Stats} } = { statCache: {}, stat: true, dot: true };
	console.log(pattern);

	const mg = new Glob(pattern, options, (er, matches) => {
		if (er) {
			console.error(er);
			return;
		}
        const file1 = matches[0];
        const stat = options.statCache[file1];
        if (stat && stat.isFile()) {
            console.log("is file");
        } else if (stat && stat.isDirectory()) {
            console.log("is directory");
        }
	});
	console.log("after");
```
? That should work just fine, and the library _really_ doesn't need a full stats object provided to it.

styfle:
Hi @weswigham 

I am wrapping another function that does the following:

```ts
export function globHelper(opts: string | glob.IOptions) {
  let options: glob.IOptions;
  if (typeof opts === 'string') {
    options = { cwd: opts };
  } else {
    options = opts;
  }

  options.symlinks = {};
  options.statCache = {};
  options.stat = true;
  options.dot = true;
  // ...
}
```

Sure it doesn't _need_ to have anything. I can use `as`, `any`, even `!` to force it to work the way that I need.

In fact, that is exactly what is happening now as you can see here: https://github.com/zeit/now-builders/blob/a56b32f670627d29a700675ed232cc66a335944e/packages/now-build-utils/src/fs/glob.ts#L49

This file was upgraded to typescript from plain javascript a year or so ago.

The reason why I submitted this PR last year was for correctness and for others who run into the same problem when using the `glob` package, one of the [most depended on packages](https://www.npmjs.com/browse/depended) in the world.

sandersn:
@vvakame @voy @ajafff @weswigham can you review this again? I didn't see an explicit signoff from anybody, which a critical package like `glob` needs.

weswigham:
I didn't wanna signoff on the change since, strictly speaking, you don't need the change to accomplish what the OP wants (you can use a more specific options type locally) and the additional data isn't required by the library.

sheetalkamat (changes requested):
Its not clear if this change is needed.

styfle:
Hi @sheetalkamat 

What changes are you requesting?

sheetalkamat:
Needs signoff from @vvakame @voy @ajafff

styfle:
It sounds like no one wants this fix so I'll just close it and use `any` 🤷‍♂️

PR #34011, opened 2019-03-19
add typings for pino isLevelEnabled() method

schmod:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/pinojs/pino/blob/40d25dace51de27c26b6b67bfc612aa98a5263fb/docs/api.md#loggerislevelenabledlevel

### DIFF
--- a/types/pino/index.d.ts
+++ b/types/pino/index.d.ts
@@ -444,6 +444,11 @@ declare namespace P {
          * Flushes the content of the buffer in extreme mode. It has no effect if extreme mode is not enabled.
          */
         flush(): void;
+
+        /**
+         * A utility method for determining if a given log level will write to the destination.
+         */
+        isLevelEnabled(level: LevelWithSilent | string): boolean;
     }
 
     type LevelChangeEventListener = (lvl: LevelWithSilent | string, val: number, prevLvl: LevelWithSilent | string, prevVal: number) => void;
--- a/types/pino/pino-tests.ts
+++ b/types/pino/pino-tests.ts
@@ -91,6 +91,7 @@ logstderr.error('on stderr instead of stdout');
 log.useLevelLabels = true;
 log.info('lol');
 log.level === 'info';
+const isEnabled: boolean = log.isLevelEnabled('info');
 
 const extremeDest = pino.extreme();
 const logExtreme = pino(extremeDest);


alferpal:
LGTM

types/pino/index.d.ts
```diff
@@ -444,6 +444,11 @@ declare namespace P {
          * Flushes the content of the buffer in extreme mode. It has no effect if extreme mode is not enabled.
          */
         flush(): void;
+
+        /**
+         * A utility method for determining if a given log level will write to the destination.
+         */
+        isLevelEnabled(level: LevelWithSilent | string): boolean;
```
screendriver: Why a generic string? If you define `LevelWithSilent | string` you could get rid of `LevelWithSilent` because every string is valid and I guess it should only be allowed to set allowed levels. So I would delete `string`

screendriver (changes requested):

types/pino/index.d.ts
screendriver: Ok, I saw that you can set a level with LevelWithSilent | string but in my opinion this is wrong as well. If you define union for string literals and a generic string the string literal makes no sense anymore.

schmod: My best guess is that this could potentially provide hints to documentation-generators and IDE autocompletions.

This convention is followed in 4 other places in this definitions file, so I'd be a little hesitant to depart from it here without a good reason.

BendingBender: Well, actually it does. You can use the `LiteralUnion` type from https://github.com/sindresorhus/type-fest to get autocompletion for such union types in an IDE.

BendingBender: It's actually more of a bug in TypeScript that it doesn't work.

BendingBender (approved):

types/pino/index.d.ts
screendriver: All right, good to know :+1:

schmod: @screendriver Are you comfortable approving this PR?

screendriver (approved):

types/pino/index.d.ts
screendriver: I thought it would be enough that @BendingBender approved it. I approved it also now.

PR #34017, opened 2019-03-20
[@types/mongoose] Document should be exported as a class instead of an interface

ajsharp:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/ajsharp/typescript-interface-bug/blob/master/src/Person.ts#L15

* * *

Mongoose's `Document` class is an exported module from the [original package](https://github.com/Automattic/mongoose/blob/master/lib/document.js#L3180), but is only declared as an interface here. This is a problem because typescript strips interfaces from the compiled code.

This PR simply declares and exports the existing `Document` interface as a class, using typescript [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html).

The main problem I ran into was trying to use the `Document` interface in typescript code via `if (this instanceof Document) { /** ... */ }`. But Typescript strips out interfaces in compiled code, so this is not possible. I realize that I can do the same with the exported `Model` class, but as `Document` is an exported class from the original code, I _should_ be able to do that here too.

There seems to also be a related bug in typescript (see https://github.com/Microsoft/TypeScript/issues/29743) where certain kinds of interface declarations don't cause the compiler to throw the `'Document' only refers to a type, but is being used as a value here` error.

### DIFF
--- a/types/mongoose/index.d.ts
+++ b/types/mongoose/index.d.ts
@@ -3336,6 +3336,8 @@ declare module "mongoose" {
     __v?: number;
   }
 
+  export class Document {}
+
   interface SaveOptions {
     safe?: boolean | WriteConcern;
     validateBeforeSave?: boolean;


jendrikw (changes requested):
Is is possible to put the methods inside the class? I feel like that would be cleaner, but I'm not sure if the semantics are the same.

jendrikw:
I'll accept this PR now, so it can be merged.

jendrikw (approved):

uniqueiniquity (changes requested):
I agree with @jendrikw; does it make sense to just change the existing interface to a class, rather than adding this extra declaration?

ajsharp:
@jendrikw I'm not sure what you mean. I didn't add any methods, effectively just aliased an interface as a class.

@uniqueiniquity Sure, I'd support that, though, not sure what sort of backwards compatibility issues that might introduce.

uniqueiniquity:
@ajsharp I checked with @DanielRosenwasser and I believe it would be the same as far as the existing usage.

PR #34025, opened 2019-03-20
Removing react-bootstrap

aaronbeall:
Per https://github.com/react-bootstrap/react-bootstrap/issues/3489

I ran `npm run not-needed -- react-bootstrap 1.0.0-beta.0 https://github.com/react-bootstrap/react-bootstrap "React Bootstrap"`

Please fill in this template.

If removing a declaration:


vitosamson:
How does this work for versions < 1.0.0? What happens if a correction needs to be made to the typings for previous versions?

aaronbeall:
@vitosamson 
> How does this work for versions < 1.0.0? 
Pre 1.0 versions are still available on NPM, basically this stubs out >=1.0. 
> What happens if a correction needs to be made to the typings for previous versions?
That's a good question, I don't know if it's possible. I tried to research this. It isn't really addressed in the DT docs: https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package
@RyanCavanaugh

vitosamson:
Is it possible to just update `notNeededPackages.json` for >= 1.0.0 while still leaving the old code, and allowing new typings to be published for < 1.0.0?

aaronbeall:
That would make sense... I'm not sure if the publisher supports that. I couldn't find any example that maintained old versions in DT.

collingreen:
As far as I know, react-bootstrap is going to maintain both the <1.0 and >=1.0 forks for bootstrap 3 and 4, respectively. The >=1.0 version includes its own types, which is why I assume this PR exists. However, we can't just remove all the types since the <1.0 branch will continue development and, as @vitosamson pointed out, we need to maintain the ability to update/fix things over time.

aaronbeall:
@collingreen I agree, I'll close this PR because I don't think outright removing `@types/react-bootstrap` seems right at this point. However we still have the problem that `npm install react-bootstrap @types/react-bootstrap` results in an invalid state that isn't immediately obvious. It seems we need to publish a `1.0` that is a stub, while leaving open the possibility to publish `0.x` definitions. I haven't had any luck finding info on what the protocol for that would be or if the DT types publisher can support that.

@RyanCavanaugh is it possible to do what @vitosamson mentioned:

> Is it possible to just update `notNeededPackages.json` for >= 1.0.0 while still leaving the old code, and allowing new typings to be published for < 1.0.0?

collingreen:
That's certainly true and I've run into that exact problem with several other libraries recently while upgrading a large codebase. I don't know much about the underlying DT code but it would be great if it could support that kind of thing.

RyanCavanaugh:
> Is it possible to just update notNeededPackages.json for >= 1.0.0 while still leaving the old code, and allowing new typings to be published for < 1.0.0?

I think the best we can do without adding new features to types-publisher is to have `react-bootstrap/v1` with an empty `.d.ts`. Alternatively we could add support for a custom message in the not-needed-package text to say "If you meant to get this for pre-1.0, npm install @types/whatever@thething` instead. Thoughts?

aaronbeall:
> I think the best we can do without adding new features to types-publisher is to have `react-bootstrap/v1` with an empty `.d.ts`. 

Hm! So that led me to a part of the FAQ I had missed: [If a library is updated to a new major version with breaking changes, how should I update its type declaration package?](https://github.com/DefinitelyTyped/DefinitelyTyped#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package)

This seems pretty close to what we want. Can't we add a `/v0` dir folder and move all the old (current) definitions in there, and put a stub in the root for Version 1.0.0?

I guess the issue is that the types publisher doesn't actually publish the `/v0` dirs to npm, so consumers to get the latest 0.x definition would actually have to install `1.x` and set a path mapping to `/v0`? Or does the types publisher continue to publish the based on the `index.d.ts` version in the sub-dirs?

aaronbeall:
Superseded by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/43756

PR #34026, opened 2019-03-20
Add FullscreenControl component type definition added in react-map-gl 4.1

sandersiim:
Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://uber.github.io/react-map-gl/#/Documentation/api-reference/fullscreen-control

### DIFF
--- a/types/react-map-gl/index.d.ts
+++ b/types/react-map-gl/index.d.ts
@@ -1,7 +1,8 @@
-// Type definitions for react-map-gl 4.0
+// Type definitions for react-map-gl 4.1
 // Project: https://github.com/uber/react-map-gl#readme
 // Definitions by: Robert Imig <https://github.com/rimig>
 //                 Fabio Berta <https://github.com/fnberta>
+//                 Sander Siim <https://github.com/sandersiim>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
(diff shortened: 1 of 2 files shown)


types/react-map-gl/index.d.ts
```diff
@@ -309,6 +310,13 @@ export interface NavigationControlProps extends BaseControlProps {
 
 export class NavigationControl extends BaseControl<NavigationControlProps> {}
 
+export interface FullscreenControlProps extends BaseControlProps {
+    className?: string;
+    container?: object | null;
```
macobo: Is this maybe `container?: HTMLElement`? Not sure why null is accepted here.

sandersiim: I followed the propTypes used in the original implementation: https://github.com/uber/react-map-gl/blob/4.1-release/src/components/fullscreen-control.js
Notice that the defaultProp value is `null`

macobo: Ah, interesting. In that case I think `HTMLElement | null` is still the most correct thing here, especially since prop-types package does not provide html element prop type, but we can get that with ts.

sandersiim: Allright, seems reasonable

sandersiim: HTMLElement matches with mapbox-gl-js definition also: https://docs.mapbox.com/mapbox-gl-js/api/#fullscreencontrol so I'll update it.

types/react-map-gl/index.d.ts
```diff
@@ -1,7 +1,8 @@
-// Type definitions for react-map-gl 4.0
+// Type definitions for react-map-gl 4.1
 // Project: https://github.com/uber/react-map-gl#readme
 // Definitions by: Robert Imig <https://github.com/rimig>
 //                 Fabio Berta <https://github.com/fnberta>
+//                 Sander Siim <https://github.com/sandersiim>
```
fnberta: This will have the effect that you will get notified on every issue and PR related to react-map-gl types. Just something to be aware of.

fnberta (approved):

PR #34029, opened 2019-03-20
ndn-js: new typings

yoursunny:
ndn-js library contains many functions but this typings only cover its public API, as documented on [NDN Common Client Libraries API](https://named-data.net/doc/ndn-ccl-api/) page. Also, I omitted some KeyChain-related APIs (declared as `any`) because they are rarely used.

If adding a new definition:


types/ndn-js/transport.d.ts
```diff
@@ -0,0 +1,45 @@
+// tslint:disable-next-line no-empty-interface
```
sheetalkamat: Do not disable these tslint rules.

sheetalkamat (changes requested):
Please use dts gen to create the new package so we have correct headers and documentation. Please refer to https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package

types/ndn-js/delegation-set.d.ts
```diff
@@ -0,0 +1 @@
+export type DelegationSet = any;
```
sheetalkamat: This seems incorrect. https://github.com/named-data/ndn-js/blob/master/js/delegation-set.js#L51 shows its more like a class. Definitely shouldn't be any

sheetalkamat (changes requested):

types/ndn-js/delegation-set.d.ts
yoursunny: Many classes in ndn-js are rarely used so I declared them as `any`.
However, `DelegationSet` isn't one of them, so it now has full declaration, and merged into `interest.d.ts`.

sheetalkamat: If the classes are not used you should still declare them as classes without declaring properties and methods in them for better shape than using any

types/ndn-js/name.d.ts
```diff
@@ -0,0 +1,89 @@
...
+    }
+}
+
+export class Name {
+    constructor(components?: ReadonlyArray<Name.Component|Uint8Array>);
+    constructor(name: Name|string);
+
+    append(value: ReadonlyArray<number>|ArrayBuffer|Uint8Array|string|Blob, type?: ComponentType, otherTypeCode?: number): Name;
+    append(component: Name.Component): Name;
+    // tslint:disable-next-line unified-signatures
```
sheetalkamat: Do not disable this rule, instead fix the error

types/ndn-js/key-chain.d.ts
```diff
@@ -0,0 +1,43 @@
...
+    setDefaultKey(identity: PibIdentity, key: PibKey, onComplete?: () => any, onError?: (err: any) => any): void;
+
+    addCertificate(key: PibKey, certificate: CertificateV2, onComplete?: () => any, onError?: (err: any) => any): void;
+    deleteCertificate(key: PibKey, certificateName: Name, onComplete?: () => any, onError?: (err: any) => any): void;
+    setDefaultCertificate(key: PibKey, certificate: CertificateV2, onComplete?: () => any, onError?: (err: any) => any): void;
+
+    sign(data: Data, params: SigningInfo, onComplete?: (data: Data) => any, onError?: (err: any) => any): void;
+    sign(interest: Interest, params: SigningInfo, onComplete?: (interest: Interest) => any, onError?: (err: any) => any): void;
+    signWithSha256(data: Data): void;
+    // tslint:disable-next-line unified-signatures
```
sheetalkamat: Please fix this lint rule instead of disabling it

types/ndn-js/key-chain.d.ts
```diff
@@ -0,0 +1,43 @@
...
+    signWithSha256(interest: Interest): void;
+}
+
+// no declaration because these types are rarely used
+export type CertificateV2 = any;
```
sheetalkamat: Instead of making them any declare them as class without declaring methods and properties

types/ndn-js/transport.d.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+}
+
+export class WebSocketTransport extends Transport {
+}
+
+export namespace WebSocketTransport {
+    class ConnectionInfo extends TransportConnectionInfo {
+        // tslint:disable-next-line unified-signatures
+        constructor(host: string, port: number);
```
sheetalkamat: Do not disable the rule, fix error instead

types/ndn-js/face.d.ts
```diff
@@ -0,0 +1,61 @@
...
+    removePendingInterest(id: number): void;
+    removeRegisteredPrefix(id: number): void;
+    send(encoding: Blob|Buffer): void;
+    setCommandCertificateName(certificateName: Name): void;
+    setCommandSigningInfo(keyChain: KeyChain, certificateName: Name): void;
+    setInterestFilter(filter: InterestFilter|Name, onInterest: OnInterestCallback): number;
+    unsetInterestFilter(id: number): void;
+}
+
+export type ForwardingFlags = any; // no declaration because this type is rarely used
```
sheetalkamat: Please declare as class (its ok to not write all properties and methods) as that's better than any

sheetalkamat (changes requested):

sheetalkamat (approved):

PR #34034, opened 2019-03-20
Added the definition file for the 'child-process-promise' package

TheDSCPL:
The package does not already provide its own types and cannot have its `.d.ts` files generated via `--declaration`


types/child-process-promise/index.d.ts
```diff
@@ -0,0 +1,107 @@
...
+    ForkOptions,
+    SpawnOptions
+} from 'child_process';
+
+/**
+ * Simple wrapper around the child_process module that makes use of promises
+ */
+
+// stop exporting everything by default
+export {};
```
sheetalkamat: Don't need this. Please remove it

sheetalkamat (changes requested):

types/child-process-promise/index.d.ts
TheDSCPL: I was isolating implementation details because I didn't think named exports were needed for those return types because the important part comes from the 'child_process' package. Nevertheless, I've removed the `export {};`, its comment and exported the interfaces that were being so.

TheDSCPL:
@sheetalkamat This bot is really awesome! It moderates the PRs so nicely! Great work on creating it

sheetalkamat (approved):

sheetalkamat:
@TheDSCPL Thanks, the work on bot is done by @RyanCavanaugh @weswigham @sandersn and indeed great work.

PR #34038, opened 2019-03-20
[remarkable] Improve Rules types

lsagetlethias:
Improve types for better plugin development.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes:  https://github.com/jonschlinkert/remarkable

Those changes are made to have better completion during plugin development.

### DIFF
--- a/types/remarkable/index.d.ts
+++ b/types/remarkable/index.d.ts
@@ -2,8 +2,11 @@
 // Project: https://github.com/jonschlinkert/remarkable
 // Definitions by: makepost <https://github.com/makepost>
 //                 Richard Lea <https://github.com/chigix>
+//                 Lilian Saget-Lethias <https://github.com/bios21>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
 
 import Remarkable = require("./lib");
 
 export = Remarkable;
+export as namespace Remarkable;
--- a/types/remarkable/lib/index.d.ts
+++ b/types/remarkable/lib/index.d.ts
@@ -259,7 +259,7 @@ declare namespace Remarkable {
         silent: boolean
     ) => boolean;
 
-    type Rule<T extends TagToken = TagToken, R extends string = string> = <T>(
+    type Rule<T extends TagToken = TagToken, R extends string = string> = (
         /**
          * The list of tokens currently being processed.
          */
--- a/types/remarkable/lib/index.d.ts
+++ b/types/remarkable/lib/index.d.ts
@@ -147,7 +147,7 @@ declare namespace Remarkable {
         parser: ParserBlock;
         options: Options;
         env: Env;
-        tokens: [ContentToken];
+        tokens: BlockContentToken[];
         bMarks: number[];
         eMarks: number[];
         tShift: number[];
@@ -175,7 +175,7 @@ declare namespace Remarkable {
         src: string;
         env: Env;
         parser: ParserInline;
-        tokens: [ContentToken];
+        tokens: ContentToken[];
         pos: number;
         posMax: number;
         level: number;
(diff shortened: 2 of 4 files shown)


types/remarkable/index.d.ts
```diff
@@ -2,8 +2,11 @@
...
 // Definitions by: makepost <https://github.com/makepost>
 //                 Richard Lea <https://github.com/chigix>
+//                 Lilian Saget-Lethias <https://github.com/bios21>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
 
 import Remarkable = require("./lib");
 
 export = Remarkable;
+export as namespace Remarkable;
```
chigix: Why create another export? If namespace export is better, would not it be necessary to remove the previous direct export? What's more, the import usage in test file is not revised for this newly added namespace export.

types/remarkable/tsconfig.json
```diff
@@ -7,7 +7,7 @@
         "noImplicitAny": true,
         "noImplicitThis": true,
         "strictNullChecks": true,
-        "strictFunctionTypes": true,
+        "strictFunctionTypes": false,
```
chigix: why strict need to be off?

types/remarkable/lib/index.d.ts
```diff
@@ -390,81 +401,430 @@ declare namespace Remarkable {
...
         block: boolean;
     }
 
-    interface MiscTokenProps {
-        /**
-         * Image alt.
-         */
-        alt?: string;
+    // ---------------
+    // Specific Block Tokens
```
chigix: Wow, you have added most rule token interfaces 👍 Could attach some links to related resources about the tokens and provide a few small tests?

chigix:
I have commented my questions and if possible, could you tell us much more detailed thinking or some related reasons about the 2 revises? I've only got the point about the rules tokens expanding.

chigix (changes requested):
I have commented my questions and if possible, could you tell us much more detailed thinking or some related reasons about the 2 revises? I've only got the point about the rules tokens expanding.

types/remarkable/index.d.ts
bios21: Like React types for example, it helps when various kind of imports are made + it does not have conflict with the previous declaration.

Because of that, test should not be changed and already cover this change because the behavior should remain the same. (useful when using `esModuleInterop` in tsconfig options for example).

types/remarkable/tsconfig.json
bios21: This strict option need to be removed to ensure a non breaking change.

In `type Rule`, I've changed the typing to rely on generics but also made sure that it's still remain usable without it.

types/remarkable/lib/index.d.ts
bios21: IMHO, it's still very perfectible.

For source I've basically read the source code: https://github.com/jonschlinkert/remarkable/tree/master/lib/rules_block when the token stack is updated.

I will add tests for that asap.

chigix (approved):
👏

types/remarkable/lib/index.d.ts
```diff
@@ -374,97 +385,446 @@ declare namespace Remarkable {
...
          */
-        type: string;
+        block?: boolean;
+    }
 
+    // ---------------
+    // Specific Block Tokens
+    // ---------------
+
+    interface BlockquoteToken extends TagToken {}
```
armanio123: Why use an empty interface? Ins't better instead for the interfaces to extend `TagToken` so that `BlockquoteToken` serves no purpose?

Same question for a couple more empty interfaces.

bios21: I kept like this for functional purposes. I wanted for every interfaces to reflect its source to prevent any evolution.  
I agree with your proposal ; I just suggested a costless preventing.

armanio123: Sounds good, thanks for clarifying.

PR #32663, opened 2019-01-31
[@types/google-apps-script] Remove the unnecessary console declaration

DominikPalo:
See: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/32585

@motemen, @grant

PR #34042, opened 2019-03-20
Revert "[@types/google-apps-script] Remove the unnecessary console declaration"

grant:
Reverts DefinitelyTyped/DefinitelyTyped#32663.

`console` is actually needed.
We do not have `console from `@types/node` since Node is not the runtime we should be using.


grant:
You may merge this without a review.

PR #34044, opened 2019-03-20
[@types/request] HTTP Auth username and password can be null when sending a token

nwehner:
Documentation for the HTTP Authentication may be found at https://www.npmjs.com/package/request#http-authentication

Note that the example states:
`request.get('http://some.server.com/').auth(null, null, true, 'bearerToken');`

The first two parameters in .auth are username and password, which were previously only allowed to be of type string. Now these parameters are allowed to be set to null.

### DIFF
--- a/types/request/index.d.ts
+++ b/types/request/index.d.ts
@@ -227,7 +227,7 @@ declare namespace request {
         json(val: any): Request;
         aws(opts: AWSOptions, now?: boolean): Request;
         hawk(opts: HawkOptions): void;
-        auth(username: string, password: string, sendImmediately?: boolean, bearer?: string): Request;
+        auth(username: string | null, password: string | null, sendImmediately?: boolean, bearer?: string): Request;
         oauth(oauth: OAuthOptions): Request;
         jar(jar: CookieJar): Request;
 
--- a/types/request/index.d.ts
+++ b/types/request/index.d.ts
@@ -227,7 +227,8 @@ declare namespace request {
         json(val: any): Request;
         aws(opts: AWSOptions, now?: boolean): Request;
         hawk(opts: HawkOptions): void;
-        auth(username: string | null, password: string | null, sendImmediately?: boolean, bearer?: string): Request;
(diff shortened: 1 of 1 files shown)


types/request/index.d.ts
```diff
@@ -227,7 +227,7 @@ declare namespace request {
         json(val: any): Request;
         aws(opts: AWSOptions, now?: boolean): Request;
         hawk(opts: HawkOptions): void;
-        auth(username: string, password: string, sendImmediately?: boolean, bearer?: string): Request;
+        auth(username: string | null, password: string | null, sendImmediately?: boolean, bearer?: string): Request;
```
mastermatt: Does it make more sense to overload this method with the two forms?
```ts
auth(username: string, password: string, sendImmediately?: boolean): Request;
auth(username: null, password: null, sendImmediately: boolean, bearer: string): Request;
```
This would make a call like `...auth('bob', 'password', true, 'sometoken')` show an error.
The [function def is here](https://github.com/request/request/blob/master/request.js#L1338) which just proxies to [this function](https://github.com/request/request/blob/master/lib/auth.js#L126).

nwehner: Totally! Yeah, that makes a lot more sense. One definitely wouldn't want to send both a username:pass and a token at the same time.

rbuckton: @nwehner: Do you plan to make this change or should I merge this PR?

nwehner: Just edited the typings to include both options for 'auth'

DanielRosenwasser:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f6a9514a75c282328e8ad78de7a39075e0e6faee/types/request/request-tests.ts#L427

has been broken by the change

mastermatt:
yep, that test was demonstrating a technically possible, but incorrect option to call the method with.
Since there is already an example of a basic auth further up the file, that line should just be replaced with:
```ts
request.get('http://some.server.com/').auth(null, null, true, 'bearerToken');
```
@nwehner

PR #34045, opened 2019-03-20
react-maskedinput: validate should return a boolean

smeevil:
the validations are done by regex tests, which return a boolean.

from the example in the docs :
`validate(char) { return /\w/.test(char ) },`

Is this observation correct ?
otherwise it might change to return : string | boolean

### DIFF
--- a/types/react-maskedinput/index.d.ts
+++ b/types/react-maskedinput/index.d.ts
@@ -9,7 +9,7 @@
 import * as React from "react";
 
 export interface FormatCharacter {
-    validate(char: string): string;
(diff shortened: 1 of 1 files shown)


smeevil:
Sorry, I should have made an issue out of this, as the change seems to have a lot more impact than I anticipated. It seems the `formatCharacters` has multiple responsibilities, one is to validate characters with a regex pattern test, which returns a boolean, and an other is to transform characters, which can return a string. I'm not sure what the correct corse is the take now to reflect the correct types...

PR #34046, opened 2019-03-20
Add @google-cloud/text-to-speech definitions

bensalilijames:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/google-cloud__text-to-speech/index.d.ts
+++ b/types/google-cloud__text-to-speech/index.d.ts
@@ -1,0 +1,123 @@
+// Type definitions for google-cloud__text-to-speech 0.5
+// Project: https://github.com/googleapis/nodejs-text-to-speech
+// Definitions by: Ben James <https://github.com/benhjames>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+// TypeScript Version: 2.8
+
+/// <reference types="node" />
+
+export type GoogleError = any;
+
+export type APICallback<T = any> = (
+    err: GoogleError | null,
+    response?: T
+) => void;
+
+export interface PromiseLike<T> extends Promise<T> {
+    /**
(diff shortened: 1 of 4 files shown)


types/google-cloud__text-to-speech/index.d.ts
```diff
@@ -0,0 +1,123 @@
...
+    listVoices: MethodOverload<ListVoicesRequest, ListVoicesResponse>;
+    synthesizeSpeech: MethodOverload<
+        SynthesizeSpeechRequest,
+        SynthesizeSpeechResponse
+    >;
+}
+
+declare const TextToSpeech: { TextToSpeechClient: typeof TextToSpeechClient };
+
+export default TextToSpeech;
```
sheetalkamat: This needs `export =` instead of `export default`. Refer to https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default

sheetalkamat (changes requested):

types/google-cloud__text-to-speech/index.d.ts
benhjames: I can't seem to get any form of `export =` with namespaces to work (with varying problems, including [this one](https://github.com/Microsoft/dtslint/blob/master/docs/export-just-namespace.md)). If you could give any further guidance as to what else I need to do, that would be great! 👍 

(btw - [the package does adhere](https://github.com/googleapis/nodejs-text-to-speech/blob/master/src/index.js#L87) to the default export syntax, though it's faked using the older module system, does this make a difference?)

Thanks!

sheetalkamat: Sorry about that. I missed seeing that it does have default property so this should be good. Thanks

PR #33201, opened 2019-02-19
Infer parameter and return types for sinon.spy()

felixfbecker:
This allows spys to be passed to APIs that are being tested while still fulfilling the types and supporting type inference. Example:

```ts
interface Service { method(): boolean }
function myFunction(factory: () => Service) {}

it('should work', () => {
  const spy = sinon.spy(() => ({
    method: () => true
  }))
  myFunction(spy) // type checked
  spy.returnValues // correctly inferred as Service[]
})
```

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://sinonjs.org/releases/v7.2.4/spies/
- [x] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #34048, opened 2019-03-20
sinon: make calledWith partial

### LINKED ISSUE #33839: sinon: `calledWith` should be partial
At some point we broke `calledWith` in the `SinonSpyCallApi` interface.

```ts
/**
 * Returns true if spy was called at least once with the provided arguments.
 * Can be used for partial matching, Sinon only checks the provided arguments against actual arguments,
 * so a call that received the provided arguments (in the same spots) and possibly others as well will return true.
 * @param args
 */
calledWith(...args: MatchArguments<TArgs>): boolean;
```

As you can see from the comment right there:

> Can be used for partial matching, Sinon only checks the provided arguments against actual arguments,
> so a call that received the provided arguments (in the same spots) and possibly others as well will return true.

Thus, it should instead be:

```ts
calledWith(...args: Partial<MatchArguments<TArgs>>): boolean;
```

or something like that.

if i get time i can pr, otherwise cc usual people:

@JoshuaKGoldberg @SimonSchick @nicojs

43081j:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://sinonjs.org/releases/v7.3.0/spies/

---

Fixes #33839

Summary of the mass of posts below & the changes here:

* `SinonInspectable` isn't a real thing, sinon has no such interface/concept, it has been removed from the types
* `SinonSpy` when used in spread has been changed to use `any` as `TArgs` (rather than `any[]`, the default) to allow having a collection of spies (trade-off we must make to avoid introducing a workaround interface like the whole `SinonInspectable` thing did)
* `calledWith` is partial, so can accept _some_ of the arguments rather than requiring them all
* the inspectable changes introduced a breaking change we overlooked (a big one which caused thousands of type errors in my case), removing it and breaking whoever uses it is a safer play than leaving everyone else broken

### DIFF
--- a/types/sinon/ts3.1/index.d.ts
+++ b/types/sinon/ts3.1/index.d.ts
@@ -29,7 +29,7 @@ declare namespace Sinon {
          * so a call that received the provided arguments (in the same spots) and possibly others as well will return true.
          * @param args
          */
-        calledWith(...args: MatchArguments<TArgs>): boolean;
+        calledWith(...args: Partial<MatchArguments<TArgs>>): boolean;
         /**
          * Returns true if spy was called at least once with the provided arguments and no others.
          */
--- a/types/sinon/ts3.1/index.d.ts
+++ b/types/sinon/ts3.1/index.d.ts
@@ -135,16 +135,21 @@ declare namespace Sinon {
          */
         lastArg: any;
 
+        /* tslint:disable:no-unnecessary-generics */
+
         /**
          * Returns true if the spy call occurred before another spy call.
          * @param call
+         *
          */
-        calledBefore(call: SinonSpyCall): boolean;
+        calledBefore<TSpyArgs extends any[]>(call: SinonSpyCall<TSpyArgs>): boolean;
         /**
          * Returns true if the spy call occurred after another spy call.
          * @param call
          */
-        calledAfter(call: SinonSpyCall): boolean;
+        calledAfter<TSpyArgs extends any[]>(call: SinonSpyCall<TSpyArgs>): boolean;
+
+        /* tslint:enable:no-unnecessary-generics */
     }
 
     interface SinonSpy<TArgs extends any[] = any[], TReturnValue = any>
@@ -214,26 +219,32 @@ declare namespace Sinon {
 
         // Methods
         (...args: TArgs): TReturnValue;
+
+        /* tslint:disable:no-unnecessary-generics */
+
         /**
          * Returns true if the spy was called before @param anotherSpy
          * @param anotherSpy
          */
-        calledBefore(anotherSpy: SinonSpy): boolean;
+        calledBefore<TSpyArgs extends any[]>(anotherSpy: SinonSpy<TSpyArgs>): boolean;
         /**
          * Returns true if the spy was called after @param anotherSpy
          * @param anotherSpy
          */
-        calledAfter(anotherSpy: SinonSpy): boolean;
+        calledAfter<TSpyArgs extends any[]>(anotherSpy: SinonSpy<TSpyArgs>): boolean;
         /**
          * Returns true if spy was called before @param anotherSpy, and no spy calls occurred between spy and @param anotherSpy.
          * @param anotherSpy
          */
-        calledImmediatelyBefore(anotherSpy: SinonSpy): boolean;
+        calledImmediatelyBefore<TSpyArgs extends any[]>(anotherSpy: SinonSpy<TSpyArgs>): boolean;
         /**
          * Returns true if spy was called after @param anotherSpy, and no spy calls occurred between @param anotherSpy and spy.
          * @param anotherSpy
          */
-        calledImmediatelyAfter(anotherSpy: SinonSpy): boolean;
+        calledImmediatelyAfter<TSpyArgs extends any[]>(anotherSpy: SinonSpy<TSpyArgs>): boolean;
+
+        /* tslint:enable:no-unnecessary-generics */
+
         /**
          * Creates a spy that only records calls when the received arguments match those passed to withArgs.
          * This is useful to be more expressive in your assertions, where you can access the spy with the same call.
@@ -628,12 +639,14 @@ declare namespace Sinon {
     }
 
     interface SinonStubStatic {
-        // Disable rule so assignment to typed stub works (see examples in tests).
+        /* tslint:disable:no-unnecessary-generics */
+
         /**
          * Creates an anonymous stub function
          */
-        // tslint:disable-next-line no-unnecessary-generics
-        <TArgs extends any[]= any[], R = any>(): SinonStub<TArgs, R>;
+        <TArgs extends any[] = any[], R = any>(): SinonStub<TArgs, R>;
+
+        /* tslint:enable:no-unnecessary-generics */
 
         /**
          * Stubs all the object’s methods.
@@ -1153,36 +1166,38 @@ declare namespace Sinon {
         pass(assertion: any): void; // Overridable
 
         // Methods
+        /* tslint:disable:no-unnecessary-generics */
+
         /**
          * Passes if spy was never called
          * @param spy
          */
-        notCalled(spy: SinonSpy): void;
+        notCalled<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
         /**
          * Passes if spy was called at least once.
          */
-        called(spy: SinonSpy): void;
+        called<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
         /**
          * Passes if spy was called once and only once.
          */
-        calledOnce(spy: SinonSpy): void;
+        calledOnce<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
         /**
          * Passes if spy was called exactly twice.
          */
-        calledTwice(spy: SinonSpy): void;
+        calledTwice<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
         /**
          * Passes if spy was called exactly three times.
          */
-        calledThrice(spy: SinonSpy): void;
+        calledThrice<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
         /**
          * Passes if spy was called exactly num times.
          */
-        callCount(spy: SinonSpy, count: number): void;
+        callCount<TArgs extends any[]>(spy: SinonSpy<TArgs>, count: number): void;
         /**
          * Passes if provided spies were called in the specified order.
          * @param spies
          */
-        callOrder(...spies: SinonSpy<any>[]): void;
+        callOrder(...spies: Array<SinonSpy<any>>): void;
         /**
          * Passes if spy was ever called with obj as its this value.
          * It’s possible to assert on a dedicated spy call: sinon.assert.calledOn(spy.firstCall, arg1, arg2, ...);.
@@ -1191,7 +1206,10 @@ declare namespace Sinon {
         /**
          * Passes if spy was always called with obj as its this value.
          */
-        alwaysCalledOn(spy: SinonSpy, obj: any): void;
+        alwaysCalledOn<TArgs extends any[]>(spy: SinonSpy<TArgs>, obj: any): void;
+
+        /* tslint:enable:no-unnecessary-generics */
+
         /**
          * Passes if spy was called with the provided arguments.
          * It’s possible to assert on a dedicated spy call: sinon.assert.calledWith(spy.firstCall, arg1, arg2, ...);.
@@ -1268,18 +1286,24 @@ declare namespace Sinon {
          * It’s possible to assert on a dedicated spy call: sinon.assert.threw(spy.thirdCall, exception);.
          */
         threw<TArgs extends any[]>(spyOrSpyCall: SinonSpy<TArgs> | SinonSpyCall<TArgs>, exception: any): void;
+
+        /* tslint:disable:no-unnecessary-generics */
+
         /**
          * Like threw, only required for all calls to the spy.
          */
-        alwaysThrew(spy: SinonSpy): void;
+        alwaysThrew<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
         /**
          * Like threw, only required for all calls to the spy.
          */
-        alwaysThrew(spy: SinonSpy, exception: string): void;
+        alwaysThrew<TArgs extends any[]>(spy: SinonSpy<TArgs>, exception: string): void;
         /**
          * Like threw, only required for all calls to the spy.
          */
-        alwaysThrew(spy: SinonSpy, exception: any): void;
+        alwaysThrew<TArgs extends any[]>(spy: SinonSpy<TArgs>, exception: any): void;
+
+        /* tslint:enable:no-unnecessary-generics */
+
         /**
          * Uses sinon.match to test if the arguments can be considered a match.
          */
--- a/types/sinon/ts3.1/index.d.ts
+++ b/types/sinon/ts3.1/index.d.ts
@@ -111,8 +111,7 @@ declare namespace Sinon {
         yieldToOn(property: string, obj: any, ...args: any[]): void;
     }
 
-    interface SinonSpyCall<TArgs extends any[] = any[], TReturnValue = any>
-        extends SinonSpyCallApi<TArgs, TReturnValue> {
+    interface SinonSpyCallLike {
         /**
          * The call’s this value.
          */
@@ -124,7 +123,7 @@ declare namespace Sinon {
         /**
          * Return value.
          */
-        returnValue: TReturnValue;
+        returnValue: any;
         /**
          * This property is a convenience for a call’s callback.
          * When the last argument in a call is a Function, then callback will reference that. Otherwise it will be undefined.
@@ -139,44 +138,23 @@ declare namespace Sinon {
          * Returns true if the spy call occurred before another spy call.
          * @param call
          */
-        calledBefore(call: SinonSpyCall): boolean;
+        calledBefore(call: SinonSpyCallLike): boolean;
         /**
          * Returns true if the spy call occurred after another spy call.
          * @param call
          */
-        calledAfter(call: SinonSpyCall): boolean;
+        calledAfter(call: SinonSpyCallLike): boolean;
     }
 
-    /**
-     * A test spy is a function that records arguments, return value,
-     * the value of this and exception thrown (if any) for all its calls.
-     */
-    interface SinonSpy<TArgs extends any[] = any[], TReturnValue = any> extends SinonInspectable<TArgs, TReturnValue> {
-        // Methods
-        (...args: TArgs): TReturnValue;
-
+    interface SinonSpyCall<TArgs extends any[] = any[], TReturnValue = any>
+        extends SinonSpyCallLike, SinonSpyCallApi<TArgs, TReturnValue> {
         /**
-         * Creates a spy that only records calls when the received arguments match those passed to withArgs.
-         * This is useful to be more expressive in your assertions, where you can access the spy with the same call.
-         * @param args Expected args
+         * @inheritdoc
          */
-        withArgs(...args: MatchArguments<TArgs>): SinonSpy<TArgs, TReturnValue>;
-
-        /**
-         * Set the displayName of the spy or stub.
-         * @param name
-         */
-        named(name: string): SinonSpy<TArgs, TReturnValue>;
+        returnValue: TReturnValue;
     }
 
-    /**
-     * The part of the spy API that allows inspecting the calls made on a spy.
-     */
-    interface SinonInspectable<TArgs extends any[] = any[], TReturnValue = any>
-        extends Pick<
-                SinonSpyCallApi<TArgs, TReturnValue>,
-                Exclude<keyof SinonSpyCallApi<TArgs, TReturnValue>, 'args'>
-            > {
+    interface SinonSpyLike {
         // Properties
         /**
          * The number of recorded calls.
@@ -205,19 +183,19 @@ declare namespace Sinon {
         /**
          * The first call
          */
-        firstCall: SinonSpyCall<TArgs, TReturnValue>;
+        firstCall: SinonSpyCallLike;
         /**
          * The second call
          */
-        secondCall: SinonSpyCall<TArgs, TReturnValue>;
+        secondCall: SinonSpyCallLike;
         /**
          * The third call
          */
-        thirdCall: SinonSpyCall<TArgs, TReturnValue>;
+        thirdCall: SinonSpyCallLike;
         /**
          * The last call
          */
-        lastCall: SinonSpyCall<TArgs, TReturnValue>;
+        lastCall: SinonSpyCallLike;
         /**
          * Array of this objects, spy.thisValues[0] is the this object for the first call.
          */
@@ -225,7 +203,7 @@ declare namespace Sinon {
         /**
          * Array of arguments received, spy.args[0] is an array of arguments received in the first call.
          */
-        args: TArgs[];
+        args: any[];
         /**
          * Array of exception objects thrown, spy.exceptions[0] is the exception thrown by the first call.
          * If the call did not throw an error, the value at the call’s location in .exceptions will be undefined.
@@ -235,27 +213,36 @@ declare namespace Sinon {
          * Array of return values, spy.returnValues[0] is the return value of the first call.
          * If the call did not explicitly return a value, the value at the call’s location in .returnValues will be undefined.
          */
-        returnValues: TReturnValue[];
+        returnValues: any[];
+
+        // Methods
+        (...args: any[]): any;
         /**
          * Returns true if the spy was called before @param anotherSpy
          * @param anotherSpy
          */
-        calledBefore(anotherSpy: SinonInspectable): boolean;
+        calledBefore(anotherSpy: SinonSpyLike): boolean;
         /**
          * Returns true if the spy was called after @param anotherSpy
          * @param anotherSpy
          */
-        calledAfter(anotherSpy: SinonInspectable): boolean;
+        calledAfter(anotherSpy: SinonSpyLike): boolean;
         /**
          * Returns true if spy was called before @param anotherSpy, and no spy calls occurred between spy and @param anotherSpy.
          * @param anotherSpy
          */
-        calledImmediatelyBefore(anotherSpy: SinonInspectable): boolean;
+        calledImmediatelyBefore(anotherSpy: SinonSpyLike): boolean;
         /**
          * Returns true if spy was called after @param anotherSpy, and no spy calls occurred between @param anotherSpy and spy.
          * @param anotherSpy
          */
-        calledImmediatelyAfter(anotherSpy: SinonInspectable): boolean;
+        calledImmediatelyAfter(anotherSpy: SinonSpyLike): boolean;
+        /**
+         * Creates a spy that only records calls when the received arguments match those passed to withArgs.
+         * This is useful to be more expressive in your assertions, where you can access the spy with the same call.
+         * @param args Expected args
+         */
+        withArgs(...args: MatchArguments<any[]>): SinonSpyLike;
         /**
          * Returns true if the spy was always called with @param obj as this.
          * @param obj
@@ -264,29 +251,29 @@ declare namespace Sinon {
         /**
          * Returns true if spy was always called with the provided arguments (and possibly others).
          */
-        alwaysCalledWith(...args: MatchArguments<TArgs>): boolean;
+        alwaysCalledWith(...args: MatchArguments<any[]>): boolean;
         /**
          * Returns true if spy was always called with the exact provided arguments.
          * @param args
          */
-        alwaysCalledWithExactly(...args: MatchArguments<TArgs>): boolean;
+        alwaysCalledWithExactly(...args: MatchArguments<any[]>): boolean;
         /**
          * Returns true if spy was always called with matching arguments (and possibly others).
          * This behaves the same as spy.alwaysCalledWith(sinon.match(arg1), sinon.match(arg2), ...).
          * @param args
          */
-        alwaysCalledWithMatch(...args: TArgs): boolean;
+        alwaysCalledWithMatch(...args: any[]): boolean;
         /**
          * Returns true if the spy/stub was never called with the provided arguments.
          * @param args
          */
-        neverCalledWith(...args: MatchArguments<TArgs>): boolean;
+        neverCalledWith(...args: MatchArguments<any[]>): boolean;
         /**
          * Returns true if the spy/stub was never called with matching arguments.
          * This behaves the same as spy.neverCalledWith(sinon.match(arg1), sinon.match(arg2), ...).
          * @param args
          */
-        neverCalledWithMatch(...args: TArgs): boolean;
+        neverCalledWithMatch(...args: any[]): boolean;
         /**
          * Returns true if spy always threw an exception.
          */
@@ -309,17 +296,22 @@ declare namespace Sinon {
          * If the stub was never called with a function argument, yield throws an error.
          * Returns an Array with all callbacks return values in the order they were called, if no error is thrown.
          */
-        invokeCallback(...args: TArgs): void;
+        invokeCallback(...args: any[]): void;
+        /**
+         * Set the displayName of the spy or stub.
+         * @param name
+         */
+        named(name: string): SinonSpyLike;
         /**
          * Returns the nth call.
          * Accessing individual calls helps with more detailed behavior verification when the spy is called more than once.
          * @param n
          */
-        getCall(n: number): SinonSpyCall<TArgs, TReturnValue>;
+        getCall(n: number): SinonSpyCallLike;
         /**
          * Returns an Array of all calls recorded by the spy.
          */
-        getCalls(): Array<SinonSpyCall<TArgs, TReturnValue>>;
+        getCalls(): SinonSpyCallLike[];
         /**
          * Resets the state of a spy.
          */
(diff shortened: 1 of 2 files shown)


types/sinon/ts3.1/sinon-tests.ts
```diff
@@ -82,6 +82,7 @@ function testSandbox() {
     const stubInstance = sb.createStubInstance(cls);
     const privateFooStubbedInstance = sb.createStubInstance(PrivateFoo);
     stubInstance.foo.calledWith('foo', 1);
+    stubInstance.foo.calledWith('foo');
```
SimonSchick: Won't still allow stuff like `.calledWith('foo', undefined)` which would be invalid?

43081j: a partial doesn't include extra values.

as in, `Partial<[number, number]>` would not match `[1, 3, 4]`. if your inferred arguments allow for an `undefined` in the second position, sure.

SimonSchick (approved):

43081j:
fyi build fails because the recent change to add nested matchers results in an inferred assertion type of `any`.

looking into it

43081j:
@felixfbecker do you mind explaining your inspectable interface that was added not so long ago?

did you introduce it so you could pass an array of spies? or was it for some other reason?

basically it is causing trouble because TS3.1 can't infer the types correctly whereas 3.2 can. 

surely, if it was to solve the "array of spies" problem, `SinonInspectable` should be a non-generic base interface all spies implement (i.e. a `SpyLike` type) so you can correctly pass an array of them without caring what the generic types are.

43081j:
alright here's a proposed solution (latest commit).

* reverted `SinonInspectable` because im not sure what you were trying to solve with that
* introduced `SinonSpyLike` and `SinonSpyCallLike` which are non-generic bases of each interface
* changed each method which doesn't care about the _type of_ spy to take the non-generic bases instead
* changed any occurrences of arrays of spies to be arrays of the bases

In strongly typed languages you'd usually do something like this (which isn't possible in TS):

```ts
interface Foo {
  prop1: string;
  prop2: any;
}

interface Foo<T> extends Foo {
  prop2: T;
}
```

to get around the fact we can't quite do this, i named it instead `SpyLike` rather than it being a non-generic `Spy`.

felixfbecker:
The `SinonInspectable` was added because the assert functions do not need to use the call signature of the spy (it only asserts the recorded calls, never calls the spy). If the call signature was in there, passing a concrete `SinonSpy<TArgs, TReturn>` (where `TArgs` and `TReturn` are not `any`) would (rightfully) throw a type error (it was _not_ for passing an array of spies). This solution was chosen as the best after discussion on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33201. The PR at some point had just widened the constraint to `any` everywhere, [but you (rightfully) said that's not the nicest solution](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33201#discussion_r258425195). It seems this PR basically now reverts to that. Except I think it's not needed to have a separate interface for that, you could just make the type parameters `any`.

I don't really understand why `SinonInspectable` causes a problem with making `calledWith` partial. Could you give an example?

43081j:
Except you can't. A `Spy<any>` isn't the same as a `Spy<somethingElse>`. these are two distinct types.

This is exactly why in most strongly typed languages, you have a non-generic base with generic sub classes. The same reason the array has problems, too.

My change to make `calledWith` accept a partial is unrelated to the rest of the changes. The other changes were because the typescript versions DT builds against have presumably changed since the last PR was merged, so the untouched sinon types fail in 3.2 but not 3.3.

They fail because the recursive match arguments type results in `any` being inferred, allowing spies to be called with anything. 

While 3.3 and above seem to infer this correctly, it is still wrong. Better to rely less on an edge case of inference and instead just write the types a bit better.

Additionally, the `SinonInspectable` types fail in one of the versions too for a similar reason:

https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/510373241?utm_source=github_status&utm_medium=notification

43081j:
In your PR that introduced the inspectable stuff, things seemed fine until you wanted an array of spies.

You then introduced the inspectable stuff. So I suspect the only reason you really needed any of this was so you can have a mixed array of generics, in which case it was the wrong way to do it and im surprised it worked at all (i guess the `any` turned type checking off so it 'worked').

felixfbecker:
> Except you can't. A Spy<any> isn't the same as a Spy<somethingElse>. these are two distinct types.

`Spy<any>` is assignable to `Spy<somethingElse>` and `Spy<somethingElse>` is assignable to `Spy<any>`, which I think is all you need, unless I am misunderstanding you.
A `Spy<any>` should behave the exact same in every observable way as a non-generic `Spy` with all generics replaced with `any` (since `any` is assignable from and to anything). 

> I still don't fully understand your train of thought. SinonInspectable is its self generic.. so it will have many of the same problems SinonSpy has, surely.

No, because the only problem arose with the function signature, because to check the assignability of a function signature the parameters need to be assignable with inverse variance. Which is why a `SinonSpy<[P1, P2], R>` (rightfully) would not be assignable to a `SinonSpy<any[]>` (when passing it to `sinon.assert`), but a `SinonInspectable<[P1, P2], R>` is. As long as the function signature is left out, the tuple is assignable to the array `any[]`, but as soon as the function signature is in there, the array `any[]` also needs to be assignable to the tuple `[P1, P2]`, which it isn't of course, because it is potentially missing the first and second index. The solution for this was to express the fact that `assert` doesn't need to call the spy and therefor the compiler doesn't need to enforce that the function signature is compatible.

> In your PR that introduced the inspectable stuff, things seemed fine until you wanted an array of spies.

No, actually the tuple/`any[]` assignability problem came before that (check the PR convo/commit history if you're curious). There were three solutions:
1. Widen `any[]` to `any` (since `any` is assignable to and from anything including tuples, while `any[]` is not)
2. Use generic function parameters for each `assert` function (doesn't work for the array of spies)
3. Add `SinonInspectable` without the function signature (works with array of spies)

You had argued to _not_ go with 1. but with 2., but that didn't work because of arrays of spies (basically you would need something like variadic type parameters). I had preferred 3. as the best solution even before seeing the array of spies problem, and eventually decided to go with it. Given that you had argued _against_ 1. (using `any`) it confuses me that you now propose to add a complete interface where everything generic is `any`.

I still don't understand where `SinonInspectable` is a problem, could you give a piece of code that's problematic?

43081j:
It is important to realise here that a generic type is distinct from another. 

In general theory, `Foo<A>` and `Foo<B>` are two distinct types. Similarly, `Foo<any>` is too, it is not the same as the other two even if `A` would satisfy `any`. The two types are distinct. 

I suspect typescript treats this distinction loosely in that `any` pretty much turns off type checking rather than being a parent/base type of `A`. This leads you to believe the two types are equal and assignable to each other but, without that looseness, they would not be. 

So this may seem pedantic of me, because the compiler tolerates your vagueness/looseness rather than treating it as an error. 

However, the build was failing when I added my change as I already mentioned because one of the typescript versions was stricter with your looseness than the others, presumably. 

Either way, the only reason your solution worked is because typescript is lenient around that area. So I proposed the change to do it correctly and strictly instead, as a suggestion. 

I introduced a base interface which isn't generic and so any members of type `T` previously would now be any. The generic interface then inherits this and specifies the type for each of these members. 

You should have done 2 in your list. That was the most correct solution and is what I did here. But then of course you can't have an array of generics (again because they are not the same) so you need a non generic base interface they all implement. That is how you would usually solve such a problem in most typed languages.

felixfbecker:
Yeah, that's because `any` is not just a "base/parent type", it's _both_ the top and bottom type. `unknown` is the top type that you seem to think `any` should be (all other types are subtypes of it), and `never` is the bottom type (it's a subtype of everything). `any` is both, so an interface with all generics replaced with `any` behaves exactly the same as an instance of the type with `any` as type argument. That's not a "looseness" in TypeScript that may be "fixed" in a future version, it's by design (it's not a bug, it's a feature). So there isn't a point in avoiding to rely on it, especially if that results in more code - it wouldn't be idiomatic TypeScript, no matter if other typed languages require that.

43081j:
You seem to be forgetting that your changes caused my build to fail. 

Also, the correct way to pass an array of spies is definitely for all of them to implement the same interface somewhere down the line. If we had that, we wouldn't need your workaround interface and could go back to the simple solution of having type parameters (option 2).

felixfbecker:
I am not forgetting it, I just don't know why either. So if you think the `SinonInspectable` is not wrong in principle it would be nice to have a small code sample that shows the problem.

I'm fine with switching to 2 (although it would be a breaking change to remove the interface, so we need to consider keeping it around). But I believe 2. doesn't require a non-generic interface (which adds code and complexity), I believe you can get the same with just using `SinonSpy<any, any>`. I still don't think of 3 as a workaround, it's accurately describing the requirements of the functions.

felixfbecker (changes requested):
Formally putting this down to make sure we don't accidentally merge until the discussion is resolved

43081j:
3 is most definitely a work around as far as I'm concerned. An interface which exists as one in the source package has been sliced apart to get around the fact that it doesn't implement a non generic base. 

The root problem here is that two spy types don't implement a common interface. The sole problem remaining if rolling your interface split back is the fact you (correctly) can't have an array of generics. So they must implement a non generic to avoid yet another hack (any, to turn off type checking). 

I do think it is wrong in principle because it is meant to be one interface. We should use generic inference so we can pass them to methods then figure out the array problem. 

Being that the interface you added was very recent and in the common use case isn't used by consumers, I'd just go with a breaking change. Until the ongoing discussion around versioning is resolved, we can't bump out of sync with sinon's version.

felixfbecker:
I think you are too focused on one solution (base interface) with putting too much weight on your main argument that it is done this way in different typed languages. We need to argue from what makes the most _sense_. The lacking base interface is not the root problem, it's a proposed solution (saying it is the root problem is saying "the root problem is we don't have solution X", which means you will obviously arrive at the conclusion that we need solution X). The interface was not split because we don't have a common interface, it was split to make the type required by the assert functions represent exactly the properties that the assert functions require (saying it was split because we don't have a common interface is saying "solution Y was implemented because solution X was not implemented", which is a non-argument). You also said "it is meant to be one interface" without providing an argument why.

Fact is that TypeScript is different from other typed languages, not just with having `any`, but also being structurally typed. So TypeScript doesn't care if an interface explicitly `implements` another base interface, whatever type you specify, it will simply check if every property on the required type is assignable from every property of the passed type. So a having a base interface that is extended or just using a type instance of the generic interface has no impact if the type instances have no observable difference (I don't see one, but I'm open to be convinced otherwise with an example where they differ in behaviour).

You said you want to avoid the "hack" of using `any`, but your PR uses `any` in the non-generic interface for all generic properties. I don't understand how that is not "hacky"/"turning off type checking"/better than just passing `any` as a type parameter? Did you intend to use the hybrid-top-bottom type `any`, or did you mean to use the top type `unknown`?

43081j:
The assert functions expect a spy. To say they take "part of a spy" is incorrect. The source package decides this, not us. There is no such thing as an "inspectable". There is a spy and a stub (which by the the source both inherit the same base). 

Given this, we should not have an arbitrary extra interface. 

If you went with option two when you tried to solve the problem you had, your only remaining problem would most definitely be that you can't have an array of generic spies. Thus it is the sole problem because the previous types only needed changing a little to solve everything else you had trouble with. 

Given that the only remaining problem, before your changes, was that we can't pass such an array, we should find a solution to that method alone. I doubt that solution is to split the interface up and I'm not requiring it be to have a base interface, it is just a suggestion because it is what you would do in almost all typed languages with generics. 

My pr uses any because I assumed we supported pre-unknown versions of typescript and didn't want to bump the required version. So yes it should indeed by unknown.

I implemented a base interface which all distinct generic spy interfaces implement. This means anywhere which would otherwise be a type parameter should instead be the base type of all things (so unknown is fine but any will do for now). Then the generic interfaces correctly implement it and define the narrowed type.

Again the sole problem here is definitely that you can't have an array of distinct generic spy types because they are not the same and have no common parent. A common solution in most languages to this common problem is to have such types implement a common interface which isn't generic.

43081j:
@sandersn would you mind taking a look at this?

its the problem i had in microsoft/dtslint#224 in that tuples won't satisfy the base constraint.

meanwhile dtslint won't allow me to specify a single use TArgs so neither way works.

43081j:
@sandersn @sheetalkamat can one of you take a look at this?

its becoming pretty aged but the problem still needs resolving and a cleanup of the now-cluttered interfaces would be nice.

i've bumped into this problem (in the build) a lot, too, so whatever happens with this PR it would still be of great use to understand the general solution to it (single-use type params for handling tuples).

sandersn:
@43081j simple suggestion before I finish reading the entire history: disable the lint rule that forbids single-use type parameters. tslint lets you disable rules for a single line.

43081j:
@sandersn if you don't mind that i will. i just thought it would be a bit hacky as i wasn't sure if the rule is in fact right and im doing something wrong.

ill update

types/sinon/ts3.1/index.d.ts
```diff
@@ -29,7 +31,7 @@ declare namespace Sinon {
          * so a call that received the provided arguments (in the same spots) and possibly others as well will return true.
          * @param args
          */
-        calledWith(...args: MatchArguments<TArgs>): boolean;
+        calledWith(...args: Partial<MatchArguments<TArgs>>): boolean;
```
rbuckton: Unfortunately this does not work quite as intended. I assume your intention is that you can undersupply `TArgs`, i.e.:

```ts
let spy: SinonSpy<[number, string, boolean], void>;
spy.calledWith(1, "a", true); // fully supplied
spy.calledWith(1, "a"); // undersupplied
spy.calledWith(1); // undersupplied
spy.calledWith(); // undersupplied
```

The problem is that `Partial<[number, string, boolean]>` results in `[number?, string?, boolean?]`. While this *does* allow you to undersupply arguments, keep in mind that 
the types in the tuple are actually expanded to `[(number | undefined)?, (string | undefined)?, (boolean | undefined)?]`, meaning that the following are *also* legal:

```ts
spy.calledWith(undefined, undefined, undefined); // fully supplied but invalid
spy.calledWith(undefined, undefined); // undersupplied but invalid
spy.calledWith(undefined); // undersupplied but invalid
```

Unfortunately, what you probably wanted was `[] | [number] | [number, string] | [number, string, boolean]`. Unfortunately, the type system today does not have a way
to express this. It could be that this is an acceptable risk with regards to the soundness of `calledWith`. It *is* important to note that sinon does distinguish between
`spy.calledWith(1, undefined)` and `spy.calledWith(1)` since it uses `arguments.length` to handle undersupplied arguments, so `spy.calledWith(undefined)` will probably just
return `false`.

types/sinon/ts3.1/index.d.ts
```diff
@@ -133,44 +135,24 @@ declare namespace Sinon {
...
 
+        /* tslint:disable:no-unnecessary-generics */
+
         /**
          * Returns true if the spy call occurred before another spy call.
          * @param call
+         *
          */
-        calledBefore(call: SinonSpyCall): boolean;
+        calledBefore<TSpyArgs extends any[]>(call: SinonSpyCall<TSpyArgs>): boolean;
```
rbuckton:
```suggestion
        calledBefore(call: SinonSpyCall<any>): boolean;
```

You can just pass `any` instead of `any[]` (or the default, which is also `any[]`) and then you wouldn't need to disable the tslint rule.

types/sinon/ts3.1/index.d.ts
```diff
@@ -133,44 +135,24 @@ declare namespace Sinon {
...
         /**
          * Returns true if the spy call occurred after another spy call.
          * @param call
          */
-        calledAfter(call: SinonSpyCall): boolean;
-    }
+        calledAfter<TSpyArgs extends any[]>(call: SinonSpyCall<TSpyArgs>): boolean;
```
rbuckton:
```suggestion
        calledAfter(call: SinonSpyCall<any>): boolean;
```
(see above)

types/sinon/ts3.1/index.d.ts
```diff
@@ -234,26 +216,41 @@ declare namespace Sinon {
...
+        (...args: TArgs): TReturnValue;
+
+        /* tslint:disable:no-unnecessary-generics */
+
         /**
          * Returns true if the spy was called before @param anotherSpy
          * @param anotherSpy
          */
-        calledBefore(anotherSpy: SinonInspectable): boolean;
+        calledBefore<TSpyArgs extends any[]>(anotherSpy: SinonSpy<TSpyArgs>): boolean;
```
rbuckton:
```suggestion
        calledBefore(anotherSpy: SinonSpy<any>): boolean;
```
(see above)

types/sinon/ts3.1/index.d.ts
```diff
@@ -234,26 +216,41 @@ declare namespace Sinon {
...
         /**
          * Returns true if the spy was called after @param anotherSpy
          * @param anotherSpy
          */
-        calledAfter(anotherSpy: SinonInspectable): boolean;
+        calledAfter<TSpyArgs extends any[]>(anotherSpy: SinonSpy<TSpyArgs>): boolean;
```
rbuckton:
```suggestion
        calledAfter(anotherSpy: SinonSpy<any>): boolean;
```
(see above)

types/sinon/ts3.1/index.d.ts
```diff
@@ -234,26 +216,41 @@ declare namespace Sinon {
...
         /**
          * Returns true if spy was called before @param anotherSpy, and no spy calls occurred between spy and @param anotherSpy.
          * @param anotherSpy
          */
-        calledImmediatelyBefore(anotherSpy: SinonInspectable): boolean;
+        calledImmediatelyBefore<TSpyArgs extends any[]>(anotherSpy: SinonSpy<TSpyArgs>): boolean;
```
rbuckton:
```suggestion
        calledImmediatelyBefore(anotherSpy: SinonSpy<any>): boolean;
```
(see above)

types/sinon/ts3.1/index.d.ts
```diff
@@ -234,26 +216,41 @@ declare namespace Sinon {
...
         /**
          * Returns true if spy was called after @param anotherSpy, and no spy calls occurred between @param anotherSpy and spy.
          * @param anotherSpy
          */
-        calledImmediatelyAfter(anotherSpy: SinonInspectable): boolean;
+        calledImmediatelyAfter<TSpyArgs extends any[]>(anotherSpy: SinonSpy<TSpyArgs>): boolean;
```
rbuckton:
```suggestion
        calledImmediatelyAfter(anotherSpy: SinonSpy<any>): boolean;
```
(see above)

types/sinon/ts3.1/index.d.ts
```diff
@@ -234,26 +216,41 @@ declare namespace Sinon {
          * If the call did not explicitly return a value, the value at the call’s location in .returnValues will be undefined.
          */
         returnValues: TReturnValue[];
+
+        // Methods
+        (...args: TArgs): TReturnValue;
+
+        /* tslint:disable:no-unnecessary-generics */
```
rbuckton:
```suggestion
```

Not needed if you make the changes recommended below

types/sinon/ts3.1/index.d.ts
```diff
@@ -234,26 +216,41 @@ declare namespace Sinon {
...
-        calledImmediatelyBefore(anotherSpy: SinonInspectable): boolean;
+        calledImmediatelyBefore<TSpyArgs extends any[]>(anotherSpy: SinonSpy<TSpyArgs>): boolean;
         /**
          * Returns true if spy was called after @param anotherSpy, and no spy calls occurred between @param anotherSpy and spy.
          * @param anotherSpy
          */
-        calledImmediatelyAfter(anotherSpy: SinonInspectable): boolean;
+        calledImmediatelyAfter<TSpyArgs extends any[]>(anotherSpy: SinonSpy<TSpyArgs>): boolean;
+
+        /* tslint:enable:no-unnecessary-generics */
```
rbuckton:
```suggestion
```
Not needed if you make the changes recommended above.

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
 
         // Methods
+        /* tslint:disable:no-unnecessary-generics */
+
         /**
          * Passes if spy was never called
          * @param spy
          */
-        notCalled(spy: SinonInspectable): void;
+        notCalled<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
```
rbuckton:
```suggestion
        notCalled(spy: SinonSpy<any>): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
         /**
          * Passes if spy was called at least once.
          */
-        called(spy: SinonInspectable): void;
+        called<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
```
rbuckton:
```suggestion
        called(spy: SinonSpy<any>): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
         /**
          * Passes if spy was called once and only once.
          */
-        calledOnce(spy: SinonInspectable): void;
+        calledOnce<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
```
rbuckton:
```suggestion
        calledOnce(spy: SinonSpy<any>): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
         /**
          * Passes if spy was called exactly twice.
          */
-        calledTwice(spy: SinonInspectable): void;
+        calledTwice<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
```
rbuckton:
```suggestion
        calledTwice(spy: SinonSpy<any>): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
         pass(assertion: any): void; // Overridable
 
         // Methods
+        /* tslint:disable:no-unnecessary-generics */
```
rbuckton:
```suggestion
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
         /**
          * Passes if spy was called exactly twice.
          */
-        calledTwice(spy: SinonInspectable): void;
+        calledTwice<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
         /**
          * Passes if spy was called exactly three times.
          */
-        calledThrice(spy: SinonInspectable): void;
+        calledThrice<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
```
rbuckton:
```suggestion
        calledThrice(spy: SinonSpy<any>): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
         /**
          * Passes if spy was called exactly num times.
          */
-        callCount(spy: SinonInspectable, count: number): void;
+        callCount<TArgs extends any[]>(spy: SinonSpy<TArgs>, count: number): void;
```
rbuckton:
```suggestion
        callCount(spy: SinonSpy<any>, count: number): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
          * @param spies
          */
-        callOrder(...spies: SinonInspectable[]): void;
+        callOrder(...spies: Array<SinonSpy<any>>): void;
         /**
          * Passes if spy was ever called with obj as its this value.
          * It’s possible to assert on a dedicated spy call: sinon.assert.calledOn(spy.firstCall, arg1, arg2, ...);.
          */
-        calledOn(spyOrSpyCall: SinonInspectable | SinonSpyCall, obj: any): void;
+        calledOn<TArgs extends any[]>(spyOrSpyCall: SinonSpy<TArgs> | SinonSpyCall<TArgs>, obj: any): void;
```
rbuckton:
```suggestion
        calledOn(spyOrSpyCall: SinonSpy<any> | SinonSpyCall<any>, obj: any): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
         /**
          * Passes if spy was always called with obj as its this value.
          */
-        alwaysCalledOn(spy: SinonInspectable, obj: any): void;
+        alwaysCalledOn<TArgs extends any[]>(spy: SinonSpy<TArgs>, obj: any): void;
```
rbuckton:
```suggestion
        alwaysCalledOn(spy: SinonSpy<any>, obj: any): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
+
+        /* tslint:enable:no-unnecessary-generics */
```
rbuckton:
```suggestion
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
+        threw<TArgs extends any[]>(spyOrSpyCall: SinonSpy<TArgs> | SinonSpyCall<TArgs>, exception: string): void;
         /**
          * Passes if spy threw the given exception.
          * The exception is a String denoting its type.
          * It’s possible to assert on a dedicated spy call: sinon.assert.threw(spy.thirdCall, exception);.
          */
-        threw(spyOrSpyCall: SinonInspectable | SinonSpyCall, exception: any): void;
+        threw<TArgs extends any[]>(spyOrSpyCall: SinonSpy<TArgs> | SinonSpyCall<TArgs>, exception: any): void;
+
+        /* tslint:disable:no-unnecessary-generics */
```
rbuckton:
```suggestion
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
+
         /**
          * Like threw, only required for all calls to the spy.
          */
-        alwaysThrew(spy: SinonInspectable): void;
+        alwaysThrew<TArgs extends any[]>(spy: SinonSpy<TArgs>): void;
```
rbuckton:
```suggestion
        alwaysThrew(spy: SinonSpy<any>): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
         /**
          * Like threw, only required for all calls to the spy.
          */
-        alwaysThrew(spy: SinonInspectable, exception: string): void;
+        alwaysThrew<TArgs extends any[]>(spy: SinonSpy<TArgs>, exception: string): void;
```
rbuckton:
```suggestion
        alwaysThrew(spy: SinonSpy<any>, exception: string): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
         /**
          * Like threw, only required for all calls to the spy.
          */
-        alwaysThrew(spy: SinonInspectable, exception: any): void;
+        alwaysThrew<TArgs extends any[]>(spy: SinonSpy<TArgs>, exception: any): void;
```
rbuckton:
```suggestion
        alwaysThrew(spy: SinonSpy<any>, exception: any): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
+
+        /* tslint:enable:no-unnecessary-generics */
```
rbuckton:
```suggestion
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
          */
-        neverCalledWithMatch<TArgs extends any[]>(spy: SinonInspectable<TArgs>, ...args: TArgs): void;
+        neverCalledWithMatch<TArgs extends any[]>(spy: SinonSpy<TArgs>, ...args: TArgs): void;
         /**
          * Passes if spy was called with the new operator.
          * It’s possible to assert on a dedicated spy call: sinon.assert.calledWithNew(spy.secondCall, arg1, arg2, ...);.
          * @param spyOrSpyCall
          */
-        calledWithNew(spyOrSpyCall: SinonInspectable | SinonSpyCall): void;
+        calledWithNew<TArgs extends any[]>(spyOrSpyCall: SinonSpy<TArgs> | SinonSpyCall<TArgs>): void;
```
rbuckton:
```suggestion
        calledWithNew(spyOrSpyCall: SinonSpy<any> | SinonSpyCall<any>): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
         /**
          * Passes if spy threw any exception.
          */
-        threw(spyOrSpyCall: SinonInspectable | SinonSpyCall): void;
+        threw<TArgs extends any[]>(spyOrSpyCall: SinonSpy<TArgs> | SinonSpyCall<TArgs>): void;
```
rbuckton:
```suggestion
        threw(spyOrSpyCall: SinonSpy<any> | SinonSpyCall<any>): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
         /**
          * Passes if spy threw the given exception.
          * The exception is an actual object.
          * It’s possible to assert on a dedicated spy call: sinon.assert.threw(spy.thirdCall, exception);.
          */
-        threw(spyOrSpyCall: SinonInspectable | SinonSpyCall, exception: string): void;
+        threw<TArgs extends any[]>(spyOrSpyCall: SinonSpy<TArgs> | SinonSpyCall<TArgs>, exception: string): void;
```
rbuckton:
```suggestion
        threw(spyOrSpyCall: SinonSpy<any> | SinonSpyCall<any>, exception: string): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -1162,133 +1166,144 @@ declare namespace Sinon {
...
         /**
          * Passes if spy threw the given exception.
          * The exception is a String denoting its type.
          * It’s possible to assert on a dedicated spy call: sinon.assert.threw(spy.thirdCall, exception);.
          */
-        threw(spyOrSpyCall: SinonInspectable | SinonSpyCall, exception: any): void;
+        threw<TArgs extends any[]>(spyOrSpyCall: SinonSpy<TArgs> | SinonSpyCall<TArgs>, exception: any): void;
```
rbuckton:
```suggestion
        threw(spyOrSpyCall: SinonSpy<any> | SinonSpyCall<any>, exception: any): void;
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -133,44 +135,24 @@ declare namespace Sinon {
...
-         * @param args Expected args
-         */
-        withArgs(...args: MatchArguments<TArgs>): SinonSpy<TArgs, TReturnValue>;
-
-        /**
-         * Set the displayName of the spy or stub.
-         * @param name
-         */
-        named(name: string): SinonSpy<TArgs, TReturnValue>;
+        /* tslint:enable:no-unnecessary-generics */
```
rbuckton:
```suggestion
```

types/sinon/ts3.1/index.d.ts
```diff
@@ -133,44 +135,24 @@ declare namespace Sinon {
          */
         lastArg: any;
 
+        /* tslint:disable:no-unnecessary-generics */
```
rbuckton:
```suggestion
```

rbuckton (changes requested):
The overall problem that `SinonInspectable` attempted to solve was one of variance. By having both a call signature of `TArgs` as well as various methods that both accept *and* return something with `TArgs`, the type for `TArgs` would become invariant in `SinonSpy`. When we measure `TArgs` as invariant, a `SinonSpy` would only become assignable to another `SynonSpy` where `TArgs` is assignable in *both* directions. Unfortunately `any[]` (the default for `TArgs`) is not assignable to any tuple.

However, `any` (rather than `any[]`) is assignable in both directions, therefore `SinonInspectable` isn't precisely necessary, but rather any instance of `SinonSpy` where we don't care about the actual type of `TArgs` we can replace with `any` (rather than `any[]`), and assignability should work correctly.

types/sinon/ts3.1/index.d.ts
43081j: i was aware of this but it seems until we have a way of representing it in the type system, its the closest solution and is still better than what we had (as what we had is broken, in that valid sinon code will not compile).

if you agree then ill leave it as it is here, its an unfortunate compromise but again is better than what we have currently IMO

43081j:
@rbuckton this is excellent, much appreciated taking the time to look through. i've made changes you suggested

makes a lot more sense now that we should be using `any` on the way in as it doesn't make a difference to the return type, but can still use `TArgs` in places where we want to return a spy of the same type.

this has sat around for a long time and is why many people still haven't updated their types so it will be great to finally get it agreed on

felixfbecker:
Thanks @rbuckton that is 100% accurate, I was unable to get the reason for `SinonInspectable` across. I am fine with `any` too, my initial PR that added `SinonInspectable` actually used `any` first for that exact reason but I changed it because of PR feedback.

43081j:
It was already put across and well understood but was an unnecessary interface which didn't exist in sinon its self and was mostly to work around the array of generics in the end. there were various reasons the original attempt of using `any` wasn't quite right from what I remember.

master currently has a rather bad breaking change preventing an update because of this workaround interface so if we're all happy with the new implementation that is great.

rbuckton (approved):

PR #34058, opened 2019-03-21
[scroll-to-element] types for scroll-to-element

kvashonin:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/scroll-to-element/index.d.ts
+++ b/types/scroll-to-element/index.d.ts
@@ -3,8 +3,8 @@
 // Definitions by: Kirill Kvashonin <https://github.com/kirillurgant>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export interface Options {
-    offset?: number;
+interface Options {
+    offset: number;
     align?: 'top' | 'middle' | 'bottom';
     ease?: string;
     duration?: number;
@@ -12,4 +12,4 @@ export interface Options {
 
 declare function scrollToElement(selector: string | HTMLElement | Element, options?: Options): void;
 
-export default scrollToElement;
+export = scrollToElement;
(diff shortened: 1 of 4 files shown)


types/scroll-to-element/index.d.ts
```diff
@@ -0,0 +1,15 @@
...
+export interface Options {
+    offset?: number;
+    align?: 'top' | 'middle' | 'bottom';
+    ease?: string;
+    duration?: number;
+}
+
+declare function scrollToElement(selector: string | HTMLElement | Element, options?: Options): void;
+
+export default scrollToElement;
```
sheetalkamat: This needs `export =` instead of `export default`. Refer to https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default

types/scroll-to-element/index.d.ts
```diff
@@ -0,0 +1,15 @@
+// Type definitions for scroll-to-element 2.0
+// Project: https://github.com/willhoag/scroll-to-element
+// Definitions by: Kirill Kvashonin <https://github.com/kirillurgant>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export interface Options {
+    offset?: number;
```
sheetalkamat: This doesn't say it defaults to something so should not be optional property. https://github.com/willhoag/scroll-to-element#offset--number

sheetalkamat (changes requested):

types/scroll-to-element/tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6",
+            "dom"
+        ],
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "strictNullChecks": false,
```
sheetalkamat: This needs to be true for any new package

sheetalkamat (changes requested):

types/scroll-to-element/tsconfig.json
kirillurgant: thank you )

sheetalkamat (approved):

PR #34064, opened 2019-03-21
[ramda] Use user type guards for filter and reject

guillaumewuip:
The PR add an optional return type to ramda `filter`.

The use case is quite common and is tested [here](https://github.com/guillaumewuip/DefinitelyTyped/blob/6db36294609da72eb4752ffc80ce207d3cb360c3/types/ramda/ramda-tests.ts#L775-L798) : 

```typescript
  type Prop= 'A' | 'B';

  interface Type<T extends Prop> {
    prop: T;
  }

  const a: Type<'A'> = {
    prop: 'A',
  };

  const b: Type<'B'> = {
    prop: 'B',
  };

  type Things = Type<"A"> | Type<"B">;

  const things: Things[] = [a, b]; // Things are Type<'A'> or Type<'B'>

  const isA = (t: Things): t is Type<"A"> => t.prop === 'A';

  // as we are keeping only Type<'A'> here so we want results to be Type<'A'>[] and not Things[]
  const results: Array<Type<'A'>> = R.filter(
    isA,
    things
  );
```

Lodash types allow us to do the same with their `filter` :
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a228c252e2eaa784e046f65c8a81c1ca3bf32f28/types/lodash/common/collection.d.ts#L225-L228

It's important to note that it's breaking for people using `Kind extends 'array'` and `Kind extends 'object'` generics types on filter now as the parameter moves to the 3rd position. I haven't found a way to fix that.

---

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: (documented in PR description)

### DIFF
--- a/types/ramda/index.d.ts
+++ b/types/ramda/index.d.ts
@@ -600,12 +600,28 @@ declare namespace R {
         set<T, U>(str: string, obj: T): U;
     }
 
-    interface Filter {
+    interface BasicFilter {
         <T>(fn: (value: T) => boolean): FilterOnceApplied<T>;
         <T, Kind extends 'array'>(fn: (value: T) => boolean): (list: ReadonlyArray<T>) => T[];
         <T, Kind extends 'object'>(fn: (value: T) => boolean): (list: Dictionary<T>) => Dictionary<T>;
         <T>(fn: (value: T) => boolean, list: ReadonlyArray<T>): T[];
         <T>(fn: (value: T) => boolean, obj: Dictionary<T>): Dictionary<T>;
+    }
+
+    interface Filter extends BasicFilter {
+      <T, R extends T>(fn: (value: T) => value is R): FilterOnceApplied<R>;
+      <T, R extends T, Kind extends 'array'>(fn: (value: T) => value is R): (list: ReadonlyArray<T>) => R[];
+      <T, R extends T, Kind extends 'object'>(fn: (value: T) => value is R): (list: Dictionary<T>) => Dictionary<R>;
+      <T, R extends T>(fn: (value: T) => value is R, list: ReadonlyArray<T>): R[];
+      <T, R extends T>(fn: (value: T) => value is R, obj: Dictionary<T>): Dictionary<R>;
+    }
+
+    interface Reject extends BasicFilter {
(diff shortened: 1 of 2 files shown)


Miaxos (approved):

googol:
The idea of allowing type narrowing is good, but I don't agree with the way it is implemented here. In my opinion this implementation falls into the documented bad practice of a disguised type assertion (see [common mistakes in the definitely typed readme, sixth bullet](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes))

I feel a better way would be to follow the same pattern as is used for the typings of the native filter: https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es5.d.ts#L1161.
The idea is to add an overload that takes a type guard instead of a basic predicate. 
That way, the return type is grounded in something received as a parameter, it is more flexible and the type parameters will be automatically infered.

guillaumewuip:
Hello @googol :) Thanks for your review.
I completely agree with your suggestion. I have updated the PR in this way.

I now have a problem with `addIndex` that seems to absolutely want to use the special reduce case for filter.

```ts
    function lastTwo(val: number, idx: number, list: number[]) {
        return list.length - idx <= 2;
    }

    const filterIndexed = R.addIndex(R.filter);
    
    filterIndexed(lastTwo, [8, 6, 7, 5, 3, 0, 9]); // => [0, 9]
    //              ^                  
    // [tsserver 2345] [E] Argument of type '(val: number, idx: number, list: number[]) => boolean' is not assignable to parameter of type 'Placeholder | ((acc: {}, item: {}, idx: number, list?: {}[] | undefined) => {})'.
    // Type '(val: number, idx: number, list: number[]) => boolean' is not assignable to type '(acc: {}, item: {}, idx: number, list?: {}[] | undefined) => {}'. ...

    const lastTwoFn = filterIndexed(lastTwo);
    lastTwoFn([8, 6, 7, 5, 3, 0, 9]);

```


After an hour on this I still haven't find a way to fix that. Maybe you see what's going on ?

googol:
I can't figure out what the problem is here. 
Could you do a rebase on top of master (or merge master to this branch) and push so that the PR stays alive?

@tycho01 Sorry to ping you but could you take a look? Maybe you see the problem immediately

PR #34065, opened 2019-03-21
JQuery/v2: Improve typings of value parameter JQueryDeferred

### LINKED ISSUE #34022: Can the value argument of the doneFilter of JQueryGenericPromise.then() be undefined?
- [x ] I tried using the `@types/xxxx` package and had problems.
- [x ] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x ] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  

  - Authors: 
@borisyankov
@choffmeister
@Steve-Fenton
@Diullei
@tasoili
@jasons-novaleaf
@seanski
@Guuz
@ksummerlin
@basarat
@nwolverson
@derekcicerone
@AndrewGaspar
@seikichi
@benjaminjackman
@s093294
@JoshStrobl
@johnnyreilly 
@DickvdBrink
@King2500
@leonard-thieu

For jquery v2, is it correct that the value argument of the doneFilter of deferred.then can be undefined?

```
/**
 * Allows jQuery Promises to interop with non-jQuery promises
 */
interface JQueryGenericPromise<T> {
    /**
     * Add handlers to be called when the Deferred object is resolved, rejected, or still in progress.
     * 
     * @param doneFilter A function that is called when the Deferred is resolved.
     * @param failFilter An optional function that is called when the Deferred is rejected.
     */
    then<U>(doneFilter: (value?: T, ...values: any[]) => U | JQueryPromise<U>, failFilter?: (...reasons: any[]) => any, progressFilter?: (...progression: any[]) => any): JQueryPromise<U>;

    /**
     * Add handlers to be called when the Deferred object is resolved, rejected, or still in progress.
     * 
     * @param doneFilter A function that is called when the Deferred is resolved.
     * @param failFilter An optional function that is called when the Deferred is rejected.
     */
    then(doneFilter: (value?: T, ...values: any[]) => void, failFilter?: (...reasons: any[]) => any, progressFilter?: (...progression: any[]) => any): JQueryPromise<void>;
}
```

The same for interface **JQueryDeferred<T>** where the 'value' argument of **resolve** and **resolveWith** is marked as optional.

I was hoping I could remove the question marks after the 'value' argument to fix half of the strict null check errors in my projects at once :)

I've put together a few sample to work with on stackblitz using jquery v2:
[https://stackblitz.com/edit/typescript-wwcsmm](https://stackblitz.com/edit/typescript-wwcsmm)

So, are those question marks correct?

thdk:
const d = $.Deferred<number>().resolve(3).then(result => {...});
Type of 'result' should not be 'number | undefined' but should be 'number'.

const d3 = $.Deferred<number | undefined>().resolve(3).then(result =>
{...});
Type of 'result' should be 'number | undefined'

Fixes DefinitelyTyped/DefinitelyTyped#34022

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://api.jquery.com/jquery.deferred/

### DIFF
--- a/types/jquery/v2/index.d.ts
+++ b/types/jquery/v2/index.d.ts
@@ -37,7 +37,7 @@ MERCHANTABLITY OR NON-INFRINGEMENT.
 See the Apache Version 2.0 License for specific language governing permissions
 and limitations under the License.
 ***************************************************************************** */
-
+// TypeScript Version: 3.2
 
 /**
  * Interface for the AJAX setting that will configure the AJAX request
@@ -321,7 +321,7 @@ interface JQueryGenericPromise<T> {
      * @param failFilter An optional function that is called when the Deferred is rejected.
      * @see {@link https://api.jquery.com/deferred.then/#deferred-then-doneFilter-failFilter-progressFilter}
      */
-    then<U>(doneFilter: (value?: T, ...values: any[]) => U|JQueryPromise<U>, failFilter?: (...reasons: any[]) => any, progressFilter?: (...progression: any[]) => any): JQueryPromise<U>;
+    then<U>(doneFilter: (value: T, ...values: any[]) => U|JQueryPromise<U>, failFilter?: (...reasons: any[]) => any, progressFilter?: (...progression: any[]) => any): JQueryPromise<U>;
 
     /**
      * Add handlers to be called when the Deferred object is resolved, rejected, or still in progress.
@@ -330,7 +330,7 @@ interface JQueryGenericPromise<T> {
      * @param failFilter An optional function that is called when the Deferred is rejected.
      * @see {@link https://api.jquery.com/deferred.then/#deferred-then-doneFilter-failFilter-progressFilter}
      */
-    then(doneFilter: (value?: T, ...values: any[]) => void, failFilter?: (...reasons: any[]) => any, progressFilter?: (...progression: any[]) => any): JQueryPromise<void>;
+    then(doneFilter: (value: T, ...values: any[]) => void, failFilter?: (...reasons: any[]) => any, progressFilter?: (...progression: any[]) => any): JQueryPromise<void>;
 }
 
 /**
@@ -482,7 +482,7 @@ interface JQueryDeferred<T> extends JQueryGenericPromise<T> {
      * @param args Optional subsequent arguments that are passed to the doneCallbacks.
      * @see {@link https://api.jquery.com/deferred.resolve/}
      */
-    resolve(value?: T, ...args: any[]): JQueryDeferred<T>;
+    resolve(value: T, ...args: any[]): JQueryDeferred<T>;
 
     /**
      * Resolve a Deferred object and call any doneCallbacks with the given context and args.
--- a/types/jquery/v2/index.d.ts
+++ b/types/jquery/v2/index.d.ts
@@ -37,6 +37,7 @@ MERCHANTABLITY OR NON-INFRINGEMENT.
 See the Apache Version 2.0 License for specific language governing permissions
 and limitations under the License.
 ***************************************************************************** */
+// TypeScript Version: 3.2
 
 /**
  * Interface for the AJAX setting that will configure the AJAX request
--- a/types/jquery/v2/index.d.ts
+++ b/types/jquery/v2/index.d.ts
@@ -37,7 +37,6 @@ MERCHANTABLITY OR NON-INFRINGEMENT.
 See the Apache Version 2.0 License for specific language governing permissions
 and limitations under the License.
 ***************************************************************************** */
-// TypeScript Version: 3.2
 
 /**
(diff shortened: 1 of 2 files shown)


types/jquery/v2/index.d.ts
```diff
@@ -321,7 +321,7 @@ interface JQueryGenericPromise<T> {
      * @param failFilter An optional function that is called when the Deferred is rejected.
      * @see {@link https://api.jquery.com/deferred.then/#deferred-then-doneFilter-failFilter-progressFilter}
      */
-    then<U>(doneFilter: (value?: T, ...values: any[]) => U|JQueryPromise<U>, failFilter?: (...reasons: any[]) => any, progressFilter?: (...progression: any[]) => any): JQueryPromise<U>;
+    then<U>(doneFilter: (value: T, ...values: any[]) => U|JQueryPromise<U>, failFilter?: (...reasons: any[]) => any, progressFilter?: (...progression: any[]) => any): JQueryPromise<U>;
```
seanski: What happens if we want to resolve a promise without a value, i.e. `.then(() => ...)`?

thdk: Then you should create the JQueryDeferred with 'void' as type argument: 
`$.Deferred<void>().resolve() `

seanski: Got it - that makes sense. FWIW, [Angular uses the void type parameter](https://stackoverflow.com/a/54143110/367611) a lot, apparently, for that use case in RxJS.

rbuckton: The optionality of `value` does not affect your ability to define an arrow function that ignores the parameter:

```ts
function f(cb: (x: number) => void) {}
f(() => {}); // not an error
```

It's not strictly necessary to type the deferred as `void` for that to work.

types/jquery/v2/index.d.ts
```diff
@@ -37,7 +37,7 @@ MERCHANTABLITY OR NON-INFRINGEMENT.
 See the Apache Version 2.0 License for specific language governing permissions
 and limitations under the License.
 ***************************************************************************** */
-
+// TypeScript Version: 3.2
```
rbuckton: I don't see any reason this comment needs to be added to begin with, as none of these changes depend on TS 3.2. I would recommend removing this line.

rbuckton (changes requested):

types/jquery/v2/index.d.ts
thdk: @rbuckton I'm sorry, I don't see what change do you want to see on this pull request. Do you mean in `jquery-tests.ts`?

```
 var promise = $.Deferred<void>().promise();
....
  promise = promise.then(function () { });
```

If I don't use void as generic type here, I cannot assign without error to the promise variable:

Type 'JQueryPromise<void>' is not assignable to type 'JQueryPromise<{}>'

types/jquery/v2/index.d.ts
thdk:
```
Error in jquery/v2
Error: /home/travis/build/DefinitelyTyped/DefinitelyTyped/types/jquery/v2/jquery-tests.ts:3544:14
ERROR: 3544:14  expect  Compile error in typescript@3.1 but not in typescript@3.2.
Fix with a comment '// TypeScript Version: 3.2' just under the header.
Expected at least 1 arguments, but got 0.
```

Ah I thought somebody told me to add it there... So, should I add it again? Or is there another workaround?

rbuckton: If the only error is coming from `jquery-tests.ts`, perhaps you are using a feature in the test file that requires 3.2, even though the definitions themselves do not?

weswigham: IIRC, the only thing that could cause this difference would be the change that "void parameters are considered optional". Considering that `void` is a core conceit of this PR, that may be intentional; however I can't see us bumping the minimum required TS version for something as widely used as `jquery` this high quite yet.

I imagine this can be "fixed" in old TS versions by adding a noop type parameter to the zero-argument overload of `$.Deferred`, this way is can be selected on older TS versions when `void` (or anything else) is supplied as a type argument.

weswigham (changes requested):

types/jquery/v2/index.d.ts
thdk: Do you mean something like this:

```
Deferred(): JQueryDeferred<JQueryStatic["noop"]>;
Deferred<T>(beforeStart?: (deferred: JQueryDeferred<T>) => any): JQueryDeferred<T>;
```

weswigham: By "noop type parameter", I mean a normal type parameter on the signature itself (like the second overload) which is simply unused. (With accompanying lint suppression and ideally a default)

PR #34067, opened 2019-03-21
Clone type definitions from @types/autobahn to @types/autobahn-browser

papaiatis:
Please fill in this template.

If adding a new definition:


papaiatis:
So I cloned the type defs because there are two npm packages for the same library: `@types/autobahn` and `@types/autobahn-browser`. The latter works in the browser, the former works in NodeJS. They both based on the same source code but only the latter has a published type definition, and I'm using the latter one. Tested it on local setup and worked fine.

types/autobahn-browser/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6",
+            "dom"
+        ],
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "strictNullChecks": false,
```
sheetalkamat: `strictNullChecks` needs to be true for any new package.

types/autobahn-browser/tslint.json
```diff
@@ -0,0 +1,78 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
```
sheetalkamat: Please remove all these rules and fix errors if any as new packages cannot have anything in their tslint otherthan extending the dtslint json file

types/autobahn-browser/index.d.ts
```diff
@@ -0,0 +1,299 @@
+// Type definitions for AutobahnJS Browser Build 18.10
+// Project: https://crossbar.io/autobahn/, https://github.com/crossbario/autobahn-js-browser
+// Definitions by: Elad Zelingher <https://github.com/darkl>, Andy Hawkins <https://github.com/a904guy>, Wladimir Totino <https://github.com/valepu>, Mathias Teier <https://github.com/glenroy37>, Fran Rodriguez <https://github.com/spcfran>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/// <reference types="when" />
+
+export = autobahn;
```
sheetalkamat: You don't want `export =` here instead you want define them in global since these would need to be available only in global.

sheetalkamat (changes requested):

sheetalkamat:
@sandersn to comment on if this duplicating of types is correct approach. Do we instead import autobahn and make  it available in global in this package?

sandersn:
Yeah, I think that

```ts
import * as autobahn from 'autobahn';
declare global {
  declare const autobahn: typeof autobahn;
}
```

should do the right thing.

papaiatis:
@sandersn I'm sorry but, to be honest, I don't really understand what it exactly should be then. 
I just wanted to use the type defs published in `@types/autobahn` for the package `autobahn-browser`. They should be compatible with each other.

sheetalkamat:
@papaiatis in the `index.d.ts` you have written as part of this change, instead of copying types, you would want to write what @sandersn is suggesting and the types should be available.

types/autobahn-browser/autobahn-browser-tests.ts
```diff
@@ -0,0 +1,76 @@
+import autobahn = require("autobahn-browser");
```
sandersn: Since this is the browser version of autobahn, you should *not* have an import statement, and the tests should still work.

papaiatis:
> Yeah, I think that
> 
> ```ts
> import * as autobahn from 'autobahn';
> declare global {
>   declare const autobahn: typeof autobahn;
> }
> ```
> 
> should do the right thing.

I'm getting two errors/warnings in this line: `declare const autobahn: typeof autobahn;`
`A 'declare' modifier cannot be used in an already ambient context.ts(1038)`
and
`'autobahn' is referenced directly or indirectly in its own type annotation.ts(2502)
`

sandersn:
Hmm, try:

```ts
import autobahn = require('autobahn')
export = autobahn
export as namespace autobahn;
```

Note that this is still not quite correct; autobahn-browser can't be used with `import = require('autobahn-browser')`, but that definition does allow it.

types/autobahn-browser/tslint.json
papaiatis: I've got a lot of errors after I removed all the rules. Most of them are `interface name must not have an "I" prefix`. That's how it was by default. I did not make it like that so not going to fix it.

types/autobahn-browser/index.d.ts
papaiatis: I'm sorry but I don't understand this.

papaiatis:
So fixing all these rules and errors and making it up to the standards is just giving me too much trouble. 
I just wanted to have a working autocomplete for the `autobahn-browser` package based on the original `@types/autobahn` def. I'm not the author of either packages. Never written any type defs so far.
Sorry, but I would rather close this PR if I'm the one who must fix this.

PR #34071, opened 2019-03-21
[kintone/kintone-utility] Add a new type definition

shuuji3:
Hello, DefinitelyTyped team 

I created a new type definition for the `kintone-utility` package.

## Template

  - **Note:** Yes, I read the advice but the definition has many `object` in it because the original code only promises the object type. It's because the nature of the target library which is a wrapper for RESTful API of the web service mainly returning the result as a JSON object changing depended on context.

If adding a new definition:
  - **Note:** Because the original js code has JSDoc, I used [tsd-jsdoc](https://github.com/englercj/tsd-jsdoc) to create the base and modified the d.ts code.

## Additional Information for a reviewer

Though I typed `Promise<object>` in many functions, JSDoc of codes in kintone/kintone-utility does not have this type. Although JSDoc tell us that `kintone.api` returns `object` not `Promise`, I think this is wrong. Because every function of the type result in `sendRequest` function in `src/js/rest/common/sendRequest.js` and this function returns from either of `kintone.api` or `useXMLHttp`. `useXMLHttp` returns `new Promise()`. And another library published by the official developer has the type definition <https://github.com/kintone/dts-gen/blob/master/kintone.d.ts> which indicate the return value of `kintone.api` is `Promise<any>`. Also from my experience for a year in this kintone platform, this function should return `Promise<any>` and the official documentation says that, too.

### DIFF
--- a/types/kintone-utility/index.d.ts
+++ b/types/kintone-utility/index.d.ts
@@ -1,0 +1,112 @@
+// Type definitions for kintone-utility 0.4
+// Project: https://github.com/kintone/kintoneUtility
+// Definitions by: TAKAHASHI Shuuji <https://github.com/shuuji3>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped/types/kintone-utility
+// TypeScript Version: 2.2
+
+export namespace rest {
+    interface File {
+        type: string;
+        url?: string;
+        file?: { fileKey: string };
+    }
+
+    function deleteAllRecords(params: { app: number; ids: ReadonlyArray<number>; isGuest?: boolean }): object;
+
+    function deleteAllRecordsByQuery(params: { app: number; query?: string; isGuest?: boolean }): object;
+
+    function deleteRecords(params: { app: number; ids: ReadonlyArray<number>; revisions?: ReadonlyArray<number>; isGuest?: boolean }): object;
+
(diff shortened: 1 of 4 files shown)


types/kintone-utility/index.d.ts
```diff
@@ -0,0 +1,112 @@
+// Type definitions for kintone-utility 0.4
```
sheetalkamat: You need to mention that `Type definitions for non-npm package kintone-utility 0.4`

types/kintone-utility/index.d.ts
```diff
@@ -0,0 +1,112 @@
...
+// Project: https://github.com/kintone/kintoneUtility
+// Definitions by: TAKAHASHI Shuuji <https://github.com/shuuji3>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped/types/kintone-utility
+// TypeScript Version: 2.2
+
+export namespace rest {
```
sheetalkamat: https://github.com/kintone/kintoneUtility/blob/master/src/js/main.js#L73 shows the library is exported as default and added to global. You want to make the shape correct. 

sheetalkamat: Also need test for used as global test

sheetalkamat (changes requested):

PR #34077, opened 2019-03-21
Add linedelimiters property to Hunk in "diff" package 

iansan5653:
Relevant code in source module: https://github.com/kpdecker/jsdiff/blob/2c1a29c0309fef22219bcf2cbb475f20508a955b/src/patch/parse.js#L84-L102

### DIFF
--- a/types/diff/index.d.ts
+++ b/types/diff/index.d.ts
@@ -142,6 +142,7 @@ export interface Hunk {
     newStart: number;
     newLines: number;
     lines: string[];
+    linedelimiters?: string[];
 }
 
 export interface BestPath {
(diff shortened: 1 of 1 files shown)


types/diff/index.d.ts
```diff
@@ -142,6 +142,7 @@ export interface Hunk {
     newStart: number;
     newLines: number;
     lines: string[];
+    linedelimiters?: string[];
```
BendingBender: This shouldn't be optional.

BendingBender (changes requested):

types/diff/index.d.ts
iansan5653: Ok, will fix. I originally made it optional for backwards compatibility, but I just realized that the property in question was added three years ago anyway.

iansan5653: Fixed with 3a4a4bb.

szdc (approved):

BendingBender (approved):

PR #33308, opened 2019-02-22
[@types/three] SkinnedMesh: Added definitions and missing property "isSkinnedMesh" and updated constructor.

Eketol:
- Added descriptions from ThreeJS documentation.
- Added missing "isSkinnedMesh:boolean".
- Updated constructor:
  -- Removed useVertexTexture.
  -- From ThreeJS r99 SkinnedMesh no longer uses Geometry, but BufferGeometry (https://github.com/mrdoob/three.js/wiki/Migration-Guide#r98--r99).
  -- Note I haven't changed the "material" types in the constructor because there is an inconsistency between their API docs and their definitions files. I'm not sure if it doesn't support "material[]" or if it is an error in the docs, because Mesh seems to accept "material | material[]".

API Docs: https://threejs.org/docs/#api/en/objects/SkinnedMesh
Source file: https://github.com/mrdoob/three.js/blob/master/src/objects/SkinnedMesh.js
Definitions: https://github.com/mrdoob/three.js/blob/master/src/objects/SkinnedMesh.d.ts

PR #34103, opened 2019-03-22
SkinnedMesh: Added definitions and missing property "isSkinnedMesh" and updated constructor.

Eketol:
This is a new pull request from #33308
- Added descriptions from ThreeJS documentation.
- Added missing "isSkinnedMesh:boolean".
- Updated constructor:
-- Removed useVertexTexture.
-- From ThreeJS r99 SkinnedMesh no longer uses Geometry, but BufferGeometry (https://github.com/mrdoob/three.js/wiki/Migration-Guide#r98--r99).
-- Note I haven't changed the "material" types in the constructor because there is an inconsistency between their API docs and their definitions files. I'm not sure if it doesn't support "material[]" or if it is an error in the docs, because Mesh seems to accept "material | material[]".

API Docs: https://threejs.org/docs/#api/en/objects/SkinnedMesh
Source file: https://github.com/mrdoob/three.js/blob/master/src/objects/SkinnedMesh.js
Definitions: https://github.com/mrdoob/three.js/blob/master/src/objects/SkinnedMesh.d.ts


Eketol:
Sorry, last two commits meant to be in a different PR. I'm a noob with GitHub. I though I could select what commits I want to make a PR for.

types/three/three-core.d.ts
```diff
@@ -5389,17 +5389,72 @@ export class Skeleton {
...
+/**
+ * A mesh that has a Skeleton with bones that can then be used to animate the vertices of the geometry. The material must support skinning and have skinning enabled.
+ * @see MeshStandardMaterial.skinning
+ * @source https://github.com/mrdoob/three.js/blob/master/src/objects/SkinnedMesh.js
+ */
 export class SkinnedMesh extends Mesh {
-    constructor(geometry?: Geometry | BufferGeometry, material?: Material | Material[], useVertexTexture?: boolean);
 
+    /**
+     * @param  {BufferGeometry} geometry? an instance of {@link BufferGeometry}.
```
uniqueiniquity: Given that you are making a change that reflects `three.js@r99`, you should update the version in the first line of `index.d.ts` to 0.99.

types/three/three-core.d.ts
```diff
@@ -5389,17 +5389,72 @@ export class Skeleton {
...
+    /**
+     * The base matrix that is used for resetting the bound bone transforms.
+     */
     bindMatrixInverse: Matrix4;
+
+    /**
+     * Used to check whether this or derived classes are skinned meshes. Default is true.
+     * You should not change this, as it used internally for optimisation.
+     */
+    isSkinnedMesh: boolean;
```
uniqueiniquity: I'm not sure it's a good idea to add this to the .d.ts file, given their guidance to not change it.
I think it would be better to leave this off so that consumers of the library don't accidentally think it's ok to change, but if they do need to change it, they can use a cast and document why they need to change it. What do you think?

uniqueiniquity (changes requested):

types/three/three-core.d.ts
Eketol: Thanks for reviewing, I don´t know if all the changes already made in that library reflects r99. Maybe it is a task for a reviewer to do that.

types/three/three-core.d.ts
Eketol: If you don´t add that property, the compiler will complain because the library checks that property.

PR #34104, opened 2019-03-22
[bull] add function overloading

lenovouser:

Before this `job` was always `any` in the following example:

```ts
queue.process(async job => {
    //
});
```

### DIFF
--- a/types/bull/index.d.ts
+++ b/types/bull/index.d.ts
@@ -13,6 +13,7 @@
 //                 Dan Manastireanu <https://github.com/danmana>
 //                 Kjell-Morten Bratsberg Thorsen <https://github.com/kjellmorten>
 //                 Christian D. <https://github.com/pc-jedi>
+//                 Silas Rech <https://github.com/lenovouser>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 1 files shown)


types/bull/index.d.ts
```diff
@@ -414,7 +417,9 @@ declare namespace Bull {
      * If the promise is rejected, the error will be passed as a second argument to the "failed" event.
      * If it is resolved, its value will be the "completed" event's second argument.
      */
-    process(callback: ((job: Job<T>, done: DoneCallback) => void) | ((job: Job<T>) => Promise<any>) | string): void;
+    process(callback: (job: Job<T>, done: DoneCallback) => void): void;
+    process(callback: (job: Job<T>) => Promise<void>): void;
+    process(callback: string): void;
```
koblas: IMHO - there should be a type definition for the job callbacks.  Since we're pretty repetitive with this and it's a standard pattern in bull.

lenovouser: Done 👍🏻

aleccool213 (approved):

types/bull/index.d.ts
```diff
@@ -397,6 +401,8 @@ declare namespace Bull {
      */
     isReady(): Promise<this>;
 
+    /* tslint:disable:unified-signatures */
```
bondz: This is disable for the entire file and not re-enabled after.

PR #34107, opened 2019-03-22
[@types/electron-store] Set ElectronStore to be default export

riain0:
Please fill in this template.

Context below.

Context for changes:

Allows for importing Store in ES6 style and shows typings for functions and properties in text editors/IDEs.

Currently, if imported in ES6 style additions need to be made in tsconfigs and all properties and functions have 'any' types.

### DIFF
--- a/types/electron-store/index.d.ts
+++ b/types/electron-store/index.d.ts
@@ -111,4 +111,4 @@ declare class ElectronStore<T = {}> implements Iterable<[string, JSONValue]> {
     [Symbol.iterator](): Iterator<[string, JSONValue]>;
 }
 
-export = ElectronStore;
+export default ElectronStore;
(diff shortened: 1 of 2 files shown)


types/electron-store/index.d.ts
```diff
@@ -111,4 +111,4 @@ declare class ElectronStore<T = {}> implements Iterable<[string, JSONValue]> {
     [Symbol.iterator](): Iterator<[string, JSONValue]>;
 }
 
-export = ElectronStore;
+export default ElectronStore;
```
rbuckton: The previous definition was correct based on the actual implementation as seen here: https://unpkg.com/electron-store@3.0.0/index.js.

uniqueiniquity:
As Ron mentioned, it seems like the [underlying library](https://unpkg.com/electron-store@3.1.0/index.js) is not using `export default`, so this .d.ts file should not either. Instead, you should follow the appropriate template from [the TypeScript handbook](http://www.typescriptlang.org/docs/handbook/declaration-files/templates.html). Note that if you want to use `import name from 'package';` in your consuming code, you can use the `esModuleInterop` compiler setting. I'm going to close this PR, but please let me know if you have any questions!

PR #34119, opened 2019-03-22
[airtable] new package - airtable npm module (Airtable JS API)

bvalosek:
This PR adds typings to the [airtable](https://github.com/airtable/airtable.js/) npm module.

A maintainer of that repo (@EvanHahn) [explicitly requested to go the DefinitelyTyped route rather than have types in the repo](https://github.com/Airtable/airtable.js/issues/34#issuecomment-468039966)

Currently this only represents a small subset of the Airtable API.

If adding a new definition:

### DIFF
--- a/types/airtable/index.d.ts
+++ b/types/airtable/index.d.ts
@@ -4,53 +4,59 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.2
 
-export default class Airtable {
-    base(appId: string): Base;
-}
+export = Airtable;
 
-export interface Base {
-    (tableName: string): Table<{}>;
-}
+declare global {
+    class Airtable {
+        base(appId: string): Airtable.Base;
+    }
 
-export interface FieldSet {
-    [ key: string ]: undefined | string | ReadonlyArray<Attachment>;
-}
+    namespace Airtable {
+        interface FieldSet {
+            [ key: string ]: undefined | string | ReadonlyArray<Attachment>;
+        }
 
-export interface Table<TFields extends FieldSet> {
-    select(opt?: SelectOptions): Query<TFields>;
-}
+        interface Base {
+            (tableName: string): Table<{}>;
+        }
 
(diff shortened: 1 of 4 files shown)


bvalosek:
@EvanHahn pinging you per this discussion:

https://github.com/Airtable/airtable.js/issues/34

Currently only a small subset of the API is typed (just what I am currently using in a project). If this looks like its on the right track I'll expand on it.

types/airtable/index.d.ts
```diff
@@ -0,0 +1,56 @@
+// Type definitions for airtable 0.5
+// Project: https://github.com/airtable/airtable.js
+// Definitions by: Brandon Valosek <https://github.com/bvalosek>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
+export default class Airtable {
```
sheetalkamat: There is no default in this library. 
This needs `export =` instead of `export default`. Refer to https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default

types/airtable/index.d.ts
```diff
@@ -0,0 +1,56 @@
...
+    firstPage(): Promise<Response<TFields>>;
+}
+
+export type Response<TFields> = ReadonlyArray<Row<TFields>>;
+
+export interface Row<TFields> {
+    id: string;
+    fields: TFields;
+}
+
```
sheetalkamat: Because this package can be used in browser as well you want to define it as global as well. 

sheetalkamat (changes requested):

bvalosek:
thanks for the links / info @sheetalkamat , reading through that info I primarily used these examples

* https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-modifying-module-d-ts.html
* https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-class-d-ts.html

I think my changes address your issues?

PR #34132, opened 2019-03-23
[sharedb] Update type signatures for Backend #use and #addProjection

ericyhwang:
This makes narrowing type updates to a few methods, to better match the current JS implementations. I'm one of the maintainers for ShareDB, so I know the code relatively well.

Note - This is not intended to be an exhaustive update to the types. I haven't looked closely at most of the existing types.

## Changes

* `Backend#use(action, listener)` now gets rich type information on the listener's function params based on the `action`, by using the same pattern as `addEventListener` in TypeScript's built-in DOM type definitions.
* This also tweaks some method param types on the base sharedb `Backend` class, to better match the JS.
* The definitions now have a minimum of TypeScript 2.1 in order to use `keyof`.
* I've removed two types from the top-level `sharedb` namespace. The tests didn't specifically reference those types. If this is a concern, I can look at adding them back in.
  * `UseAction`, which was a union of strings, replaced by a generic type param `<A extends keyof middleware.ActionContextMap>`. This could be kept via a `keyof`.
  * `UseCallback`, a function type definition, which is replaced by an inline type that uses the generic `<A>` type mentioned above. This would be trickier to keep, due to the callback now referencing `<A>`.
* I changed the `Snapshot` type from `number` to an interface that matches [the source's Snapshot class](https://github.com/share/sharedb/blob/v1.0.0-beta.20/lib/snapshot.js).

----

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [Implementation of Backend#use](https://github.com/share/sharedb/blob/v1.0.0-beta.20/lib/backend.js#L192) and [Definition of the middleware actions](https://github.com/share/sharedb/blob/v1.0.0-beta.20/lib/backend.js#L55)
- [x] Increase the version number in the header if appropriate. (N/A? Definitions are still for sharedb@1.x-beta)
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. (Already present)

### DIFF
--- a/types/sharedb/index.d.ts
+++ b/types/sharedb/index.d.ts
@@ -1,7 +1,9 @@
 // Type definitions for sharedb 1.0
 // Project: https://github.com/share/sharedb
 // Definitions by: Steve Oney <https://github.com/soney>
+//                 Eric Hwang <https://github.com/ericyhwang>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.1
 
 /// <reference path="lib/sharedb.d.ts" />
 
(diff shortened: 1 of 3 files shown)


uniqueiniquity:

PR #34137, opened 2019-03-23
[changelog-parser] Add typing for changelog-parser

azerella:
@ungoldman Adding some type defs for `changelog-parser` ❤️ 

Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/changelog-parser/index.d.ts
+++ b/types/changelog-parser/index.d.ts
@@ -6,15 +6,19 @@
 
 interface Options {
     /**
-     * Path to changelog file
+     * Path to changelog file.
      */
     filePath: string;
     /**
-     * Changelog file string to parse
+     * Removes the markdown markup from the changelog entries by default.
+     * You can change its value to false to keep the markdown.
      */
     removeMarkdown: boolean;
 }
 
+/**
+ * Change log parser for node.
+ */
 declare function parseChangelog(options: Options|string, callback?: (result: string, error?: string) => void): Promise<object>;
 
 export = parseChangelog;
--- a/types/changelog-parser/index.d.ts
+++ b/types/changelog-parser/index.d.ts
@@ -1,0 +1,20 @@
+// Type definitions for changelog-parser 2.7
+// Project: https://github.com/hypermodules/changelog-parser
+// Definitions by:  Adam Zerella <https://github.com/adamzerella>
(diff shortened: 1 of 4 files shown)


types/changelog-parser/index.d.ts
```diff
@@ -0,0 +1,20 @@
...
+     * Path to changelog file
+     */
+    filePath: string;
+    /**
+     * Changelog file string to parse
+     */
+    removeMarkdown: boolean;
+}
+
+declare function parseChangelog(options: Options|string, callback?: (result: string, error?: string) => void): Promise<object>;
```
43081j: this should probably be a `Partial<Options>` as you don't need to pass both options (you can pass `{ filePath: 'foo' }` for example).

adamzerella: @43081j Good pickup, I've updated the types based on your feedback 😄

types/changelog-parser/index.d.ts
```diff
@@ -0,0 +1,24 @@
...
+     * Removes the markdown markup from the changelog entries by default.
+     * You can change its value to false to keep the markdown.
+     */
+    removeMarkdown: boolean;
+}
+
+/**
+ * Change log parser for node.
+ */
+declare function parseChangelog(options: Partial<Options>|string, callback?: (result: object, error: string|null) => void): Promise<object>;
```
43081j: last thing, is this the right way around? the callback. it seems to be:

`(err, result) => void` in tests
`(result, err) => void` here

need switching around maybe?

adamzerella: Updated test arguments, fixed

43081j: I'm super sorry to comment on this again but just to double check and be 100% certain, are we certain the error comes second?

https://github.com/hypermodules/changelog-parser/blob/master/README.md#callback

This seems to suggest a callback is indeed what the tests had originally: `(err: string|null, result: object) => void`

ungoldman: Hi @43081j, I am the author of `changelog-parser`, you are correct, the callback is error first.

43081j: yep though it looks like its merged already so i suppose a further PR will be needed to correct it. 

thanks for taking the time to confirm

PR #34142, opened 2019-03-23
[three] Fixed InstancedBufferAttribute definition

nsunderland-cognite:
This fixes an incorrect parameter name in the constructor and a missing parameter. See https://github.com/mrdoob/three.js/pull/15957


rbuckton (approved):

rbuckton:
Thanks for the contribution!

PR #34146, opened 2019-03-23
Add types for remove-files-webpack-plugin

srgykuz:
Hello.

Here is a types for `remove-files-webpack-plugin`.
GitHub – https://github.com/Amaimersion/remove-files-webpack-plugin
NPM – https://www.npmjs.com/package/remove-files-webpack-plugin

# Template

If adding a new definition:
  - **Note**: current JSDoc types right in the plugin file cannot be treated as a providing of it's own types. I created it only for local usage and, probably, not all editors or IDE's can treat it as a truly type.

### DIFF
--- a/types/remove-files-webpack-plugin/index.d.ts
+++ b/types/remove-files-webpack-plugin/index.d.ts
@@ -1,0 +1,110 @@
+// Type definitions for remove-files-webpack-plugin 1.0
+// Project: https://github.com/Amaimersion/remove-files-webpack-plugin
+// Definitions by: Sergey Kuznetsov <https://github.com/Amaimersion>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
+
+
+/**
+ * A plugin for webpack which removes files and folders before and after compilation.
+ */
+declare class RemovePlugin {
+    constructor(parameters: PluginParameters);
+}
+
+
+/**
+ * A parameters for plugin.
+ */
+interface PluginParameters {
+    /**
+     * Removing before compilation.
+     */
+    before: RemoveParameters;
+
+    /**
+     * Removing after compilation.
+     */
+    after: RemoveParameters;
+}
+
+
+/**
+ * A parameters for removing.
+ */
+interface RemoveParameters {
+    /**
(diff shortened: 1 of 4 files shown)


Amaimersion:
Is `tsconfig.json` file really necessary? Seems like others types from `node_modules/@types` directory doesn't have that file.

Amaimersion:
Sorry for that. I misunderstood some concepts about this extra files. Now is everything alright (i think so).

types/remove-files-webpack-plugin/index.d.ts
```diff
@@ -0,0 +1,106 @@
+// Type definitions for remove-files-webpack-plugin 1.0
+// Project: https://github.com/Amaimersion/remove-files-webpack-plugin/blob/master/README.md
+// Definitions by: Sergey Kuznetsov <https://github.com/Amaimersion>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
+
+/**
+ * A plugin for webpack which removes files and folders before and after compilation.
+ */
+// tslint:disable-next-line:no-unnecessary-class
```
sheetalkamat: Please remove the disable rule and fix errors

sheetalkamat (changes requested):

types/remove-files-webpack-plugin/index.d.ts
Amaimersion: @sheetalkamat There is no errors actually. `RemovePlugin` is a class, not a function. Unfortunately, i cannot override default tslint config. Default tslint config don't allow only constructor classes. In fact, it is incorrect. 

So, if i remove this line, then i need to either change a class to function, that is incorrect at all, or add to class some unnecessary method that doesn't part of my plugin. You can convinced of it – https://github.com/Amaimersion/remove-files-webpack-plugin/blob/master/src/plugin.js#L80

Also, you, in README, allow to add some disable rules for linter – https://github.com/DefinitelyTyped/DefinitelyTyped#lint

So, it is not against the rules, it is not against the code. That line should exists.

Amaimersion:
There are no problems. It is an incorrectness from @sheetalkamat. I will recreate it later.

PR #34148, opened 2019-03-24
[lyricist] added definitions and tests

DadUndead:
- [✅] Use a meaningful title for the pull request. Include the name of the package modified.
- [✅] Test the change in your own code. (Compile and run.)
- [✅] Add or edit tests to reflect the change. (Run with `npm test`.)
- [✅] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [✅] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [✅] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [✅] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration`
- [✅] If this is for an NPM package, match the name. If not, do not conflict with the name of an NPM package.
- [✅] Create it with `dts-gen --dt`, not by basing it on an existing project.
- [✅] `tslint.json` should be present, and `tsconfig.json` should have `noImplicitAny`, `noImplicitThis`, `strictNullChecks`, and `strictFunctionTypes` set to `true`.


types/lyricist/index.d.ts
```diff
@@ -0,0 +1,333 @@
+// Type definitions for lyricist 2.2.2
+// Project: https://github.com/scf4/lyricist, https://www.npmjs.com/package/lyricist
+// Definitions by: Aleksei Klimenko <https://github.com/DadUndead>
+// Definitions: https://github.com/DefinitelyTyped/
+// TypeScript Version: 2.7
+
+declare module 'lyricist';
+
+export default class Lyricist {
```
sheetalkamat: This needs `export =` instead of `export default`. Refer to https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default

types/lyricist/tslint.json
```diff
@@ -0,0 +1,79 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
```
sheetalkamat: Remove all the rules here and fix errors, new packages cannot have anything except extends rule in tslint json file

sheetalkamat (changes requested):

types/musicmatch/tslint.json
```diff
@@ -0,0 +1,79 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "adjacent-overload-signatures": false,
```
DanielRosenwasser: Please re-enable all lint rules.

DadUndead: Thank you for review! I removed unnecessary files from that branch

PR #34149, opened 2019-03-24
Add types for license-checker-webpack-plugin

joelspadin:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/license-checker-webpack-plugin/index.d.ts
+++ b/types/license-checker-webpack-plugin/index.d.ts
@@ -1,0 +1,73 @@
+// Type definitions for license-checker-webpack-plugin 0.0
+// Project: https://github.com/Microsoft/license-checker-webpack-plugin#readme
+// Definitions by: Joel Spadin <https://github.com/ChaosinaCan>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
+
+import { Plugin } from 'webpack';
+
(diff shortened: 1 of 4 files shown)


plantain-00 (approved):

types/license-checker-webpack-plugin/index.d.ts
```diff
@@ -0,0 +1,73 @@
+// Type definitions for license-checker-webpack-plugin 0.0
+// Project: https://github.com/Microsoft/license-checker-webpack-plugin#readme
+// Definitions by: Joel Spadin <https://github.com/ChaosinaCan>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
+
+import { Plugin } from 'webpack';
+
+declare class LicenseCheckerWebpackPlugin extends Plugin {
+    constructor(options: LicenseCheckerWebpackPlugin.Options);
```
43081j: I would use a `Partial<LicenseCheckerWebpackPlugin.Options>` here. A fully-optional interface is a bit of an anti-pattern as it will overmatch (e.g. an object which has none of the properties in `Options` but has extras, will match).

sheetalkamat: That would be better but not required hence merging.

ChaosinaCan: Thanks for the suggestion. I've implemented it here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34282

PR #34151, opened 2019-03-24
[seedrandom] State should be serializable, and removing incorrect callbacks

EugeneZ:
`seedrandom`'s state function does not return a function. It returns an object intended to be opaque, which I don't know how to describe in TS. It's serializable (unlike a function), so I made it just be an empty object. It's properties are unpredictable, but always of a similar shape:

`{ [x]: number, [y]: number, [z]: number[] }`

x,y, and z vary from implementation, and users are not intended to know what they are. The important thing is that it can be passed back to the state option when making a new instance of `seedrandom`, and that it be serializable (which the previous implementation was not).

In addition, `seedrandom`'s alternate algorithms do not take a third "callback" argument, so I removed it.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.npmjs.com/package/seedrandom (note that the state is not described here, which is why the lib was working before, but the source can be seed here: http://davidbau.com/encode/seedrandom.js )

### DIFF
--- a/types/seedrandom/index.d.ts
+++ b/types/seedrandom/index.d.ts
@@ -1,6 +1,6 @@
 // Type definitions for seedrandom 2.4.2
 // Project: https://github.com/davidbau/seedrandom
-// Definitions by: Kern Handa <https://github.com/kernhanda>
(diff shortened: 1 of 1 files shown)


kernhanda (approved):

rbuckton:
Thanks for the contribution!

PR #34155, opened 2019-03-24
[cacheable-request] Include type alias for cache adapter

paulmelnikow:
This allows easily declaring a parameter that will be passed into the
constructor.

Ref https://github.com/sindresorhus/got/pull/760#discussion_r268277506

Please fill in this template.

Did not add a test because existing tests seem to cover this behavior.

Lint is failing, but I don't understand why:

```
> definitely-typed@0.0.3 lint /Users/pnm/code/DefinitelyTyped
> dtslint types "cacheable-request"

/Users/pnm/code/DefinitelyTyped/node_modules/.bin/dtslint: line 1: use strict: command not found
/Users/pnm/code/DefinitelyTyped/node_modules/.bin/dtslint: line 2: syntax error near unexpected token `('
/Users/pnm/code/DefinitelyTyped/node_modules/.bin/dtslint: line 2: `var __awaiter = (this && this.__awaiter) || function (thisArg, _arguments, P, generator) {'
```

Tests will not run on my machine either:

```
index.d.ts:17:1 - error TS2309: An export assignment cannot be used in a module with other exported elements.

17 export = CacheableRequest;
   ~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/types/cacheable-request/index.d.ts
+++ b/types/cacheable-request/index.d.ts
@@ -1,6 +1,7 @@
 // Type definitions for cacheable-request 6.0
 // Project: https://github.com/lukechilds/cacheable-request#readme
 // Definitions by: BendingBender <https://github.com/BendingBender>
+//                 Paul Melnikow <https://github.com/paulmelnikow>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
@@ -19,8 +20,10 @@ declare const CacheableRequest: CacheableRequest;
 
 type RequestFn = typeof request;
 
+export type StorageAdapter = Store<any>
+
 interface CacheableRequest {
-    new (requestFn: RequestFn, storageAdapter?: string | Store<any>): (
+    new (requestFn: RequestFn, storageAdapter?: string | StorageAdapter): (
         opts: string | URL | (RequestOptions & CacheSemanticsOptions),
         cb?: (response: ServerResponse | ResponseLike) => void
     ) => CacheableRequest.Emitter;
--- a/types/cacheable-request/index.d.ts
+++ b/types/cacheable-request/index.d.ts
@@ -20,10 +20,8 @@ declare const CacheableRequest: CacheableRequest;
 
 type RequestFn = typeof request;
 
-export type StorageAdapter = Store<any>;
-
 interface CacheableRequest {
-    new (requestFn: RequestFn, storageAdapter?: string | StorageAdapter): (
+    new (requestFn: RequestFn, storageAdapter?: string | CacheableRequest.StorageAdapter): (
         opts: string | URL | (RequestOptions & CacheSemanticsOptions),
         cb?: (response: ServerResponse | ResponseLike) => void
     ) => CacheableRequest.Emitter;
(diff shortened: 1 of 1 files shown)


paulmelnikow:
I fixed the style error. The other error looks like a real error, though it's not related to my changes and I don't know the correct way to fix it. I'd appreciate any advice on that!

```
ERROR: 17:1   expect     TypeScript@next compile error: 
An export assignment cannot be used in a module with other exported elements.
```

types/cacheable-request/index.d.ts
```diff
@@ -19,8 +20,10 @@ declare const CacheableRequest: CacheableRequest;
 
 type RequestFn = typeof request;
 
+export type StorageAdapter = Store<any>;
```
BendingBender: Better:
```ts
export {Store} from 'keyv';
```

types/cacheable-request/index.d.ts
```diff
@@ -19,8 +20,10 @@ declare const CacheableRequest: CacheableRequest;
...
+
 interface CacheableRequest {
-    new (requestFn: RequestFn, storageAdapter?: string | Store<any>): (
+    new (requestFn: RequestFn, storageAdapter?: string | StorageAdapter): (
```
BendingBender: Revert this change.

BendingBender (changes requested):

types/cacheable-request/index.d.ts
paulmelnikow: Ah, thanks for explaining that!

I moved it into the namespace, though I left the templated type in there to keep this streamlined, since as far as I can tell, it's not changeable by the caller.

types/cacheable-request/index.d.ts
```diff
@@ -33,6 +31,8 @@ interface CacheableRequest {
 }
 
 declare namespace CacheableRequest {
+    type StorageAdapter = Store<any>;
```
BendingBender: Please don't hard-code the `any` type parameter in the `StorageAdapter` type to allow users of this type to use this parameter to correctly align `get()` and `set()` signatures of the `StorageAdapter` implementation:

```suggestion
    type StorageAdapter<TValue> = Store<TValue>;
```

Don't forget to set this parameter to `any` in the CacheableRequest constructor argument.

BendingBender (changes requested):

types/cacheable-request/index.d.ts
paulmelnikow: Isn't the value type determined by the library? It's not as if the caller can meaningfully declare any type they choose; for their type to be helpful it needs to conform to the library's usage:

```js
{
  cachePolicy: response.cachePolicy.toObject(),
  url: response.url,
  statusCode: response.fromCache ? revalidate.statusCode : response.statusCode,
  body
}
```

If we don't use `Store<any>` it seems like it would be better to add an explicit type for the cache value.

BendingBender: If you would add an explicit type for the cache value, that would be great! Otherwise, stick with what you've done so far.

BendingBender (approved):

types/cacheable-request/index.d.ts
paulmelnikow: Thanks, I'm happy to address that in a follow-on.

PR #34160, opened 2019-03-24
[nodemailer] add messageId() method to MimeNode

d-fischer:
The [example transport plugin](https://nodemailer.com/plugins/create/#transport-send-mail-callback) currently fails to compile in TypeScript because of the missing `messageId()` method on `MimeNode`.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodemailer.com/plugins/create/#transport-send-mail-callback


dex4er:
Original doc says "Returns Message-Id value. If it does not exist, then creates one"

Beside that: 👍

dex4er (changes requested):
Original jdoc:

```js
/**
  * Returns Message-Id value. If it does not exist, then creates one
  *
  * @return {String} Message-Id value
  */
```

dex4er (approved):

PR #34179, opened 2019-03-25
[node-cron] Define timezone as a string literal

RMHonor:
Please fill in this template.

### DIFF
--- a/types/node-cron/package.json
+++ b/types/node-cron/package.json
@@ -1,0 +1,6 @@
+{
+  "private": true,
+  "dependencies": {
+      "@types/tz-offset": "0.0.0"
+  }
+}
--- a/types/node-cron/package.json
+++ b/types/node-cron/package.json
@@ -1,6 +1,0 @@
-{
-  "private": true,
-  "dependencies": {
-      "@types/tz-offset": "0.0.0"
-  }
-}
--- a/types/node-cron/index.d.ts
+++ b/types/node-cron/index.d.ts
@@ -5,22 +5,7 @@
 //                 Richard Honor <https://github.com/RMHonor>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export type Timezone = 'Etc/GMT+12' | 'Pacific/Pago_Pago' | 'Pacific/Midway' | 'Pacific/Honolulu' | 'America/Juneau' | 'America/Los_Angeles' | 'America/Tijuana' | 'America/Phoenix' |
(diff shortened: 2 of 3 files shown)


burtek:
I'm not sure that relying on another package's data is a good idea here. Wouldn't it be safer to create typings for `tz-offset` and depend on them here? At least when `tz-offset` data changes, one can change it in `@types/tz-offset` and it would be handled here too...

RMHonor:
Whilst I agree, `tz-offset` is not a package published on NPM. Is it normal to have type libraries published available for libraries which aren't available through standard means?

Happy to make a start on `tz-offset` anyway

burtek:
> Whilst I agree, `tz-offset` is not a package published on NPM

https://www.npmjs.com/package/tz-offset 🤔

RMHonor:
Blind as a bat 😆

burtek:
Happens to me too, np :D

RMHonor:
#34209 may be of interest to you!

burtek:
@RMHonor approved that one :)

burtek (changes requested):
Make it depend on `@types/tz-offset` as per previous discussion

burtek:
@RMHonor any plans for making this happen? :)

RMHonor:
PR updated to use `tz-offset`

burtek (changes requested):
Shouldn't `@types/tz-offset` be added to dependencies in `package.json` now? [It won't be included by default](https://github.com/DefinitelyTyped/DefinitelyTyped#i-notice-some-packages-having-a-packagejson-here).

```json
{
    "private": true,
    "dependencies": {
        "@types/tz-offset": "0.0.0"
    }
}

**Comment by RMHonor** [comment]
Ah, of course. Pushed a new `package.json`

**Review by burtek — APPROVED** [review]

File: types/node-cron/package.json [hunk 1]
```diff
@@ -0,0 +1,6 @@
+{
+  "private": true,
+  "dependencies": {
+      "tz-offset": "0.0.0"
```
burtek: `@types/tz-offset`

burtek (changes requested):

types/node-cron/package.json
RMHonor: Updated

burtek:
Actually my bad 🤦‍♂️ it must have changed some time ago
image

RMHonor:
Ah, that's clever, so DefinitelyTyped is able to look at the required imports in `.d.ts` files and adds them as dependencies?

Updated to remove the `package.json`

burtek:
> DefinitelyTyped is able to look at the required imports in `.d.ts` files and adds them as dependencies?

looks like this.

burtek (approved):

burtek:
@RMHonor it does work ;)
https://www.npmjs.com/package/@types/node-cron?activeTab=dependencies

PR #34195, opened 2019-03-25
added CheckBox to react-native

franzmoro:
Please fill in this template.

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for react-native 0.57
+// Type definitions for react-native 0.59
 // Project: https://github.com/facebook/react-native
 // Definitions by: Eloy Durán <https://github.com/alloy>
 //                 HuHuanming <https://github.com/huhuanming>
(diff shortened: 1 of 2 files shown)


types/react-native/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for react-native 0.57
+// Type definitions for react-native 0.59
```
alloy: Was this API only added in 0.59? If not, then there’s no reason to bump this now and exclude this update from people on older RN versions.

franzmoro: It was not immediately clear to me, when it would be "appropriate" to update the version, nor that by updating I would be excluding anyone from this update.

```

- Provide a URL to documentation or source code which provides context for the suggested changes: <>
- Increase the version number in the header if appropriate.
```

The API was added long ago, so I'll take this out.

alloy: People can install whatever version they want, but that is kind of how what the versions are trying to convey. Thanks!

PR #34199, opened 2019-03-25
[ember__routing] Add types from Router Service RFC

tansongyang:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/ember__routing/-private/route-info.d.ts
+++ b/types/ember__routing/-private/route-info.d.ts
@@ -4,7 +4,7 @@
  * It is read-only and internally immutable.
  * It is also not observable, because a `Transition` instance is never changed after creation.
  */
-export default class RouteInfo {
+export default interface RouteInfo {
     /**
      * A reference to the childe route's `RouteInfo`.
      * This can be used to traverse downward to the leafmost `RouteInfo`.
@@ -40,8 +40,10 @@ export default class RouteInfo {
      */
     readonly queryParams: { [key: string]: string | undefined };
     /**
-     * This method is similar to the `find()` method defined in ECMAScript 2015.
+     * Allows you to traverse through the linked list of `RouteInfo`s from the topmost to leafmost.
+     * Returns the first `RouteInfo` in the linked list for which the callback returns true.
+     *
      * @param callback the callback to execute.
      */
-    find(callback: (item: RouteInfo) => boolean): RouteInfo;
+    find(callback: (item: RouteInfo) => boolean): RouteInfo | undefined;
 }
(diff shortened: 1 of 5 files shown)


types/ember__routing/test/router.ts
```diff
@@ -52,5 +52,15 @@ const RouterServiceConsumer = Service.extend({
...
+    onAndRouteInfo() {
+        const router = get(this, 'router');
+        router
+            .on('routeWillChange', transition => {
+                const to = transition.to;
+            })
+            .on('routeDidChange', transition => {
+                const from = transition.from;
+            });
+    },
```
mike-north: Needs some tests to protect us from regression around the new `RouteInfo` type

types/ember__routing/-private/route-info.d.ts
```diff
@@ -0,0 +1,47 @@
+// https://api.emberjs.com/ember/3.6/classes/RouteInfo
+/**
+ * A `RouteInfo` is an object that contains metadata about a specific route within a `Transition`.
+ * It is read-only and internally immutable.
+ * It is also not observable, because a `Transition` instance is never changed after creation.
+ */
+export default class RouteInfo {
```
mike-north: Is there a good reason to make this a class instead of an interface?

mike-north (changes requested):
Overall this looks excellent!
- definitely needs some tests around `RouteInfo` to catch breakage when we refactor
- unless consumers will work with `RouterInfo` as both a value and a type, please make it an interface instead of a class

types/ember__routing/test/router.ts
tansongyang: Thanks for the suggestion. Will do that.

types/ember__routing/-private/route-info.d.ts
tansongyang: I just made it a class because that's how it's defined in the Ember docs. I can definitely change that to an interface.

types/ember__routing/test/router.ts
```diff
@@ -58,9 +59,31 @@ const RouterServiceConsumer = Service.extend({
         router
             .on('routeWillChange', transition => {
                 const to = transition.to;
+                assertType<RouteInfo | null>(to.child);
```
mike-north: Can we change these to use [dtslint](https://github.com/Microsoft/dtslint) tests? assertType can let inappropriately wide types slip through (i.e., if `to.child` is of type `any`, this assertion will pass).

Example
```ts
to.child; // $ExpectType RouteInfo | null
```

mike-north (changes requested):

types/ember__routing/test/router.ts
tansongyang: Sure thing. Didn't know that was an option.

mike-north (approved):
Looks great! thanks @tansongyang!

mike-north:
@RyanCavanaugh just a heads up - I think the bot should have removed the "Revision needed" tag here

rbuckton:
Thanks for the contribution!

PR #34200, opened 2019-03-25
[toastr] support strictNullChecks and optional parameters as undefined.

matthias-ccri:
This PR updates the **toastr** types to support the `strictNullChecks` compiler flag.
In particular, when `strictNullChecks` is enabled in a dependent project, you previously couldn't specify toastr options without passing a header. I.e. this would produce an error:

```ts
toastr.success('message', undefined, {/* options */}) // Error with strictNullChecks
```

Detailed explanation of changes:
- Removed the overloads in `ToastrDisplayMethod` and replaced them with optional parameters. This aligns with the [typescript best practices](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#use-optional-parameters).
    - Note: though the first parameter (`message`) can technically be `undefined` without error (showing a blank toast), I chose to not allow `undefined` as a value for it, since passing `undefined` would almost always be a developer error.
- Similar to above, used optional parameters in `clear()`.
    - Allowed the first argument (`toast`) to be undefined even if the second argument (`clearOptions`) is passed. This seems counterintuitive, but it's the most permissive, allowing the `toast` argument to be passed without type narrowing to remove `undefined` from the type. Besides, passing `undefined, options` doesn't produce any errors. [clearToast](https://github.com/CodeSeven/toastr/blob/50092cc604850a16c985520b63df184d3e0b4086/toastr.js#L133). 

In the tests:
- Added more examples from the code on the Readme. 
- Asserted that the toastr display methods return a `JQuery` object.
- Cast JQuery `val()` return values as specific types, in order to comply with `strictNullChecks`, since `val()`'s type says it may return `undefined`.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/CodeSeven/toastr/tree/50092cc604850a16c985520b63df184d3e0b4086
https://github.com/CodeSeven/toastr/blob/50092cc604850a16c985520b63df184d3e0b4086/demo.html  
https://github.com/CodeSeven/toastr/blob/50092cc604850a16c985520b63df184d3e0b4086/toastr.js#L352  

### DIFF
--- a/types/toastr/index.d.ts
+++ b/types/toastr/index.d.ts
@@ -225,23 +225,10 @@ interface ToastrDisplayMethod {
 	 * Create a toast
 	 *
 	 * @param message Message to display in toast
-	 */
-	(message: string): JQuery;
-	/**
-	 * Create a toast
-	 *
-	 * @param message Message to display in toast
-	 * @param title Title to display on toast
-	 */
-	(message: string, title: string): JQuery;
-	/**
-	 * Create a toast
-	 *
-	 * @param message Message to display in toast
 	 * @param title Title to display on toast
 	 * @param overrides Option values for toast
 	 */
-	(message: string, title: string, overrides: ToastrOptions): JQuery;
+	(message: string, title?: string, overrides?: ToastrOptions): JQuery;
 }
 
 type ToastrType = 'error'|'info'|'success'|'warning';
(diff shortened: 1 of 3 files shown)


types/toastr/index.d.ts
```diff
@@ -373,9 +350,7 @@ interface Toastr {
 	 *
 	 * @param callback The function which will be passed the event details.
 	 */
-	subscribe: (callback: (response: ToastrResponse) => any) => void;
-
-	[key: string]: any;
```
uniqueiniquity: Is there a particular reason why you removed this? I don't think you mentioned it in your PR description, but I might be missing something.

uniqueiniquity (changes requested):

types/toastr/index.d.ts
matthiasdailey-ccri: Hi, thanks for taking the time to review this. 
Yes, I intentionally removed the any-indexer. These any-indexers suppress errors that would normally appear when you attempt to use properties and methods that don't exist. They're only really needed (as a crutch) if the interface is incomplete, and this interface is complete. See [the toastr.js](https://github.com/CodeSeven/toastr/blob/50092cc604850a16c985520b63df184d3e0b4086/toastr.js#L27-L37)

uniqueiniquity: Great, just wanted to check. Thanks!

uniqueiniquity (approved):

PR #34201, opened 2019-03-25
feat(prismjs): Update `@types/prism` to Prism 1.16

ExE-Boss:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/PrismJS/prism/pull/1782

---

Note that the previous type definition was incorrect or straight up wrong in some cases.

---

review?(@sandersn)

/cc @RunDevelopment

### DIFF
--- a/types/prismjs/index.d.ts
+++ b/types/prismjs/index.d.ts
@@ -6,6 +6,7 @@
 //                 Andre Wiggins <https://github.com/andrewiggins>
 //                 Michał Miszczyszyn <https://github.com/mmiszy>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
 
 export as namespace Prism;
 export const languages: Languages;
@@ -413,7 +414,7 @@ export class Token {
 	 */
 	greedy: boolean;
 
-	// tslint:disable-next-line:no-redundant-jsdoc-2
+	// tslint:disable:no-redundant-jsdoc-2
 	/**
 	 * Converts the given token or token stream to an HTML representation.
 	 *
--- a/types/prismjs/index.d.ts
+++ b/types/prismjs/index.d.ts
@@ -1,228 +1,432 @@
-// Type definitions for prism 1.9
+// Type definitions for prismjs 1.16
 // Project: http://prismjs.com/, https://github.com/leaverou/prism
-// Definitions by: Erik Lieben <https://github.com/eriklieben>
+// Definitions by: Michael Schmidt <https://github.com/RunDevelopment>
+//                 ExE Boss <https://github.com/ExE-Boss>
+//                 Erik Lieben <https://github.com/eriklieben>
 //                 Andre Wiggins <https://github.com/andrewiggins>
 //                 Michał Miszczyszyn <https://github.com/mmiszy>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 export as namespace Prism;
-
-export const util: Util;
 export const languages: Languages;
-export const plugins: any;
-export const hooks: Hooks;
+export const plugins: Record<string, any>;
 
 /**
- * Used to highlight all elements on current website. Calls Prism.highlightAllUnder on `document`.
+ * A function which will be invoked after an element was successfully highlighted.
  *
- * @see highlightAllUnder
- * @param async Whether to use Web Workers to improve performance and avoid blocking the UI when highlighting
- * very large chunks of code. False by default.
- * @param callback An optional callback to be invoked after the highlighting is done. Mostly useful when async
- * is true, since in that case, the highlighting is done asynchronously.
+ * @param element The element successfully highlighted.
  */
-export function highlightAll(async?: boolean, callback?: (element: Element) => void): void;
+export type HighlightCallback = (element: Element) => void;
 
 /**
- * This is the most high-level function in Prism’s API. It fetches all the elements inside `container` that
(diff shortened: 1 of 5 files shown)


types/prismjs/index.d.ts
```diff
@@ -1,228 +1,435 @@
...
+	 * Converts the given token or token stream to an HTML representation.
+	 *
+	 * The following hooks will be run:
+	 * 1. `wrap`: On each {@link Prism.Token}.
+	 *
+	 * @param token The token or token stream to be converted.
+	 * @param language The name of current language.
+	 * @param [parent] The parent token stream, if any.
+	 * @return The HTML representation of the token or token stream.
+	 * @private
```
ExE-Boss: The `Token.stringify(…)` method is private according to @RunDevelopment in https://github.com/PrismJS/prism/pull/1782#discussion_r261895197.

types/prismjs/index.d.ts
```diff
@@ -1,228 +1,435 @@
...
+	 */
+	static stringify(
```
ExE-Boss: The `Token.stringify(…)` method is private according to @RunDevelopment in https://github.com/PrismJS/prism/pull/1782#discussion_r261895197.

RunDevelopment: I marked it as private because it can only work correctly if the tokens passed to it are already encoded via `Prism.util.encode`.
Put simply: It's a deadly trap from anyone trying to use it and it should only be used by the library anyway, so I went with private.

typeofweb (approved):

types/prismjs/index.d.ts
mmiszy: In this case, maybe `Prism.util.encode` should return a different type which this function will accept instead of `TokenNode`?

types/prismjs/index.d.ts
```diff
@@ -1,228 +1,435 @@
...
-    /** Determine the type of the object */
-    type(o: any): string;
-
-    /** Get the unique id of this object or give it one if it does not have one */
-    objId(obj: any): Identifier;
-
-    /** Deep clone a language definition (e.g. to extend it) */
-    clone(o: LanguageDefinition): LanguageDefinition;
+export type GrammarValue = RegExp | TokenObject | Array<RegExp | TokenObject>;
+export type Grammar = GrammarRest & Record<string, GrammarValue>;
```
RunDevelopment: Doesn't this mean that the type of `Grammar#rest` will be `GrammarValue & Grammar` which is `never`?

I'm asking because I had a similar problem a while ago and I have yet to find a solution.

types/prismjs/index.d.ts
```diff
@@ -1,228 +1,435 @@
...
+export interface GrammarRest {
+	keyword?: GrammarValue;
+	number?: GrammarValue;
```
RunDevelopment: Can't all of those be removed?

types/prismjs/index.d.ts
```diff
@@ -1,228 +1,435 @@
...
+	 */
+	function run<K extends keyof HookEnvironmentMap>(
+		name: K,
+		env: HookEnvironmentMap[K]
+	): void;
+	function run(name: string, env: Environment): void;
 }
 
-export type TokenNode = Token | string | Array<Token | string>;
+export type TokenNode = string | Token | Array<string | Token>;
```
RunDevelopment: Wouldn't `TokenStream` be a better name?

types/prismjs/index.d.ts
ExE-Boss: Probably.

I kept them to ease IntelliSense in IDEs.

types/prismjs/index.d.ts
ExE-Boss: That wouldn’t work, since TypeScript uses structural equality when checking if two types match, and since the result of `Prism.util.encode("<x>")` is `"&lt;x&gt;"`, which is still `string`, it will match `TokenNode`.

RunDevelopment: Something in the direction of `encode(tokens: TokenNode): EncodedTokenNode`, right?

Well, I don't know, if it's worth it as I plan to change this in the future anyway.

types/prismjs/index.d.ts
ExE-Boss: No, this is different.

Merging two types using `&` when one is defined using an index signature will cause the definite one to always override the index signature.

Result: `Grammar#rest` has the type of `Grammar`.

types/prismjs/index.d.ts
ExE-Boss: Probably.

You should probably add it to `prism-core.doc.js` in https://github.com/PrismJS/prism/pull/1782.

types/prismjs/index.d.ts
RunDevelopment: That's not a bad idea, but in that case, should `class-name` and `comment` also be present?

RunDevelopment (approved):

types/prismjs/index.d.ts
mmiszy: There are certain ways to enforce nominal typing in TypeScript.

But even if you keep structural typing, having a different name for this type to differentiate between them would be a good idea IMHO.

ExE-Boss: Except that type aliases are expanded into what they are aliases of in IDEs, so it wouldn’t really be clear from hovering over the function in **VS Code**.

rbuckton:
Thanks for the contribution!

PR #34218, opened 2019-03-26
fix(sharp): add missing linear, recomb, modulate types

makepanic:
Please fill in this template.

- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present). (via `node node_modules/.bin/dtslint types sharp` because https://github.com/Microsoft/dtslint/issues/219)

* `linear` [docs](https://github.com/lovell/sharp/blob/master/docs/api-operation.md#linear) , [allowing null as first param](https://github.com/lovell/sharp/blob/1999c7103ccc293d4414e1d2c3726fb601079e2d/test/unit/linear.js#L61)
* `recomb` [docs](https://github.com/lovell/sharp/blob/master/docs/api-operation.md#recomb)
* `modulate` [docs](https://github.com/lovell/sharp/blob/master/docs/api-operation.md#modulate)

Note: `modulate` is part of the next sharp release and not yet released. I'm not sure if we should already ship it in the type definition, see https://github.com/lovell/sharp/commit/fc178de309d8e4055e86ffe1bbcfcf40754c885b#diff-343723a0cb380f664aedf0dfed5c1c2f

### DIFF
--- a/types/sharp/index.d.ts
+++ b/types/sharp/index.d.ts
@@ -424,7 +424,15 @@ declare namespace sharp {
          * @throws {Error} Invalid parameters
          * @returns A sharp instance that can be used to chain operations
          */
-        linear(a?: number, b?: number): Sharp;
+        linear(a?: number | null, b?: number): Sharp;
+
+        /**
+         * Recomb the image with the specified matrix.
+         * @param inputMatrix 3x3 Recombination matrix
+         * @throws {Error} Invalid parameters
+         * @returns A sharp instance that can be used to chain operations
+         */
+        recomb(inputMatrix: Matrix3x3): Sharp;
 
         //#endregion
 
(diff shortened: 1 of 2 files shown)


BTOdell:
We should probably wait until `modulate` is released before it's included in the type definitions.
Do you have an ETA on when that release will drop?

makepanic:
I could also split this PR to one which adds `recomb` and the `linear` adjustments and create a new one for `modulate`.


> We should probably wait until `modulate` is released before it's included in the type definitions.
> Do you have an ETA on when that release will drop?

I think only @lovell knows.

BTOdell:
I'd be happy to sign off on a pull request that only includes the definitions for the released features.

lovell:
The `modulate` operation will be in the forthcoming v0.22.1 so it would probably be best to remove it from this PR as is is not yet "missing". Hope this helps.

makepanic:
I've updated the PR by removing `modulate` types.

BTOdell (approved):

rbuckton:
Thanks for the contribution!

PR #34220, opened 2019-03-26
Add support for matrix property in TransformsStyle

CapsAdmin:
Although there is no reference to the matrix property within the transform style on the wiki (https://facebook.github.io/react-native/docs/transforms) there is mention of transformMatrix which has been deprecated which tells you to use the transform property instead. Perhaps that's an issue with the documentation.

We have currently defined transformMatrix but not within the transform property.

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -666,7 +666,11 @@ type GeolocationError = {
     POSITION_UNAVAILABLE: number;
     TIMEOUT: number;
 };
-
+   
+interface MatrixTransform {
+    matrix: Array<number>;
+}
+    
 interface PerpectiveTransform {
     perspective: number;
 }
@@ -717,6 +721,7 @@ interface SkewYTransform {
 
 export interface TransformsStyle {
     transform?: (
+        | MatrixTransform
         | PerpectiveTransform
         | RotateTransform
         | RotateXTransform
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -666,11 +666,11 @@ type GeolocationError = {
     POSITION_UNAVAILABLE: number;
     TIMEOUT: number;
 };
-   
+
 interface MatrixTransform {
     matrix: Array<number>;
 }
-    
+
 interface PerpectiveTransform {
     perspective: number;
 }


alloy:
If this is deprecated, I wonder if it makes sense to leave a comment calling that out, for when a next person in the future feels like cleaning up the file?

I don’t think we particularly need test coverage for a deprecated property, but if there is exist coverage for other props and a quick copy-paste brings this up to par then that would be cool too.

types/react-native/index.d.ts
```diff
@@ -667,6 +667,10 @@ type GeolocationError = {
     TIMEOUT: number;
 };
 
+interface MatrixTransform {
+    matrix: Array<number>;
```
uniqueiniquity: I can't find any source that shows that `matrix` is allowable at runtime.
Could you link to either the source that is handling this, or an example of usage? Thanks!

uniqueiniquity: For example, [this file](https://github.com/facebook/react-native/blob/master/ReactCommon/fabric/graphics/Transform.h) seems to define the API for allowable options, and it doesn't look like `matrix` would be allowed. Please let me know if I'm missing something, though.

uniqueiniquity (changes requested):

types/react-native/index.d.ts
CapsAdmin: It's being used here:
https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/processTransform.js#L48-L50

uniqueiniquity: Great, thanks! 
Could you add a test for this? I don't know react-nnative particularly well, but it looks like [this file](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/73b16bf4233ed1ffe686186727e7d18d4124eecd/types/react-native/test/animated.tsx) would be a good place.

CapsAdmin: We are just checking if the type system checks out, right? And not actually running any code. (I don't see anything being used anywhere)

animated.tsx seems to test the animation library which also happens to involve some transforms. I think a better place is index.tsx with a new component. 

I might as well add all the other transform properties as they aren't being tested for as well.

alloy:
@CapsAdmin I’d still love to see this PR brought to completion by adding some tests and then we can re-open it.

PR #34221, opened 2019-03-26
bn.js: Adds Uint8Array as a valid constructor argument.

MicahZoltu:
Only thing I'm uncertain on is whether `Uint8Array` is safe to include in definitions yet.  It has been around for years and bn.js supports it, though [indirectly](https://github.com/indutny/bn.js/blob/master/lib/bn.js#L144).

```
> new BigNumber(new Uint8Array([0x40, 0x20])).toString(16)
'4020'
```

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/indutny/bn.js/blob/master/lib/bn.js#L144>

### DIFF
--- a/types/bn.js/bn.js-tests.ts
+++ b/types/bn.js/bn.js-tests.ts
@@ -16,3 +16,8 @@ const red = bn.toRed(ctx);
 const newRed = red.redAdd(new BN(1));
 newRed.cmp(bn);
 newRed.fromRed();
+
+const expected = new BN(0x4020);
+const actualArray = new BN([0x40, 0x20]);
+const actualUint8Array =  new BN(new Uint8Array([0x40, 0x20]));
+const actualString = new BN('0x4020');
--- a/types/bn.js/index.d.ts
+++ b/types/bn.js/index.d.ts
@@ -25,12 +25,12 @@ interface ReductionContext {
 
 declare class BN {
     constructor(
-        number: number | string | number[] | Buffer | BN,
+        number: number | string | number[] | Uint8Array | Buffer | BN,
         base?: number | 'hex',
         endian?: Endianness
     );
     constructor(
-        number: number | string | number[] | Buffer | BN,
+        number: number | string | number[] | Uint8Array | Buffer | BN,
         endian?: Endianness
     )
 


types/bn.js/bn.js-tests.ts
```diff
@@ -16,3 +16,11 @@ const red = bn.toRed(ctx);
 const newRed = red.redAdd(new BN(1));
 newRed.cmp(bn);
 newRed.fromRed();
+
+const expected = new BN(0x4020);
+const actualArray = new BN([0x40, 0x20]);
+const actualUint8Array =  new BN(new Uint8Array([0x40, 0x20]));
+const actualString = new BN('0x4020');
+if (!actualArray.eq(expected)) throw new Error(`${actualArray} does not equal ${expected}`);
```
LogvinovLeon: Correct me if I'm wrong, not no one ever runs those files. They are tests for the types, not the behaviour and therefore any value equality checks are useless.

LogvinovLeon (changes requested):

types/bn.js/bn.js-tests.ts
MicahZoltu: I believe the npm run test command will execute them?  I put the test in to ensure that `BN` would actually accept a Uint8Array as a parameter and process it properly.  It is easy to author a type definition that doesn't align with reality, and this test ensures that the type definition aligns with reality.

LogvinovLeon: Theck the main README for this repo.
```
foo-tests.ts This contains sample code which tests the typings. This code does not run, but it is type-checked.
```
I'd like to have those run, but I think that in the current situation - it's dead code and it's misleading.

MicahZoltu: A convincing argument.  Fixed.

MicahZoltu:
Addressed comments @typescript-bot.

LogvinovLeon (approved):

MicahZoltu:
Ironically, in the intervening days since authoring this I learned that JS now has a stage 3 proposal (implemented in Chrome Live, NodeJS 10+ and Firefox behind flag) for a native `bigint` type, so I have stopped using this library.  :P

PR #34222, opened 2019-03-26
underscore: Allow to pass _.List<T> to OOP wrapper

soon:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://underscorejs.org/#oop
- [ ] Increase the version number in the header if appropriate **(this isn't a feature introduced in 1.9 so I am not sure if version should be increased)**.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }` **(tslint.json already exists)**.

-----

It is possible to pass first argument to underscore rather than to a function (so called OOP style).

However, the following code fails to compile according to the current d.ts (live js demo: https://codepen.io/awesoon/pen/RdOMYa?editors=1112):

```
const data = [
  {
    id: 1,
    text: 'This'
  },
  {
    id: 2,
    text: 'works'
  },
  {
    id: 3,
    text: 'fine'
  }
];

const groups = _(data).groupBy(x => x.id);
const values = _(groups).values().map(xs => _(xs).map(x => x.text))
console.log(values);
```

Error message is: `Property 'text' does not exist on type 'List<{ id: number; text: string; }>'`

See simplified test case in the test file changes.

This MR fixes the error above by allowing to pass objects of type `_.List<T>` to underscore

### DIFF
--- a/types/underscore/index.d.ts
+++ b/types/underscore/index.d.ts
@@ -105,6 +105,7 @@ declare module _ {
         * @param key First argument to Underscore object functions.
         **/
         <T>(value: _.Dictionary<T>): Underscore<T>;
+        <T>(value: _.List<T>): Underscore<T>;
         <T>(value: Array<T>): Underscore<T>;
         <T>(value: T): Underscore<T>;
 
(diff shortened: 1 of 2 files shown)


uniqueiniquity:
Seems reasonable to me.

PR #34223, opened 2019-03-26
[node] crypto.Verify and Signer should be classes

### LINKED ISSUE #34128: [@types/node] crypto.Verify should be a class, not an interface.
- [x] I tried using the `@types/node` package and had problems.
  - Authors: @jkomyno @a-tarasyuk  @j-oliveras @bhongy  (there are a bunch so I chose first and last two) @SimonSchick (and an active contributor to the file from git blame. 

---
Verify is a class based on reading [docs](https://nodejs.org/api/crypto.html#crypto_class_verify) and [code](https://github.com/nodejs/node/blob/f1056542f0cf2520e122d93afbaada6b6599e49b/lib/crypto.js#L143) but crypto.d.ts defines Verify only as a  [interface](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/crypto.d.ts#L279).

```typescript
import { createVerify, Verify }from 'crypto'

if (createVerify('SHA256') instanceof Verify) {
  console.log('true')
}
```

This will run and print `true` but tsc will error

> 'Verify' only refers to a type, but is being used as a value

The fix might be as simple as adding the class to crypto.d.ts

```
class Verify implements Verify {}
```

but I've never written a d.ts file so I wanted advice before a pull request.

Flarna:
https://nodejs.org/api/crypto.html#crypto_class_verify
https://nodejs.org/api/crypto.html#crypto_class_sign
https://nodejs.org/api/crypto.html#crypto_verify_verify_object_signature_signatureencoding

Besides converting `Verify` and `Signer` to classes with private constructor I made also `signature_format` in `Verify.verify` optional otherwise the sample in node docs won't work.

fixes: #34128
fyi @everett1992

### DIFF
--- a/types/node/crypto.d.ts
+++ b/types/node/crypto.d.ts
@@ -268,21 +268,23 @@ declare module "crypto" {
 
     type KeyLike = string | Buffer | KeyObject;
 
-    interface Signer extends NodeJS.WritableStream {
+    class Signer extends stream.Writable {
         update(data: BinaryLike): Signer;
         update(data: string, input_encoding: Utf8AsciiLatin1Encoding): Signer;
         sign(private_key: SignPrivateKeyInput | KeyLike): Buffer;
         sign(private_key: SignPrivateKeyInput | KeyLike, output_format: HexBase64Latin1Encoding): string;
+        private constructor();
     }
 
     function createVerify(algorith: string, options?: stream.WritableOptions): Verify;
-    interface Verify extends NodeJS.WritableStream {
+    class Verify extends stream.Writable {
         update(data: BinaryLike): Verify;
         update(data: string, input_encoding: Utf8AsciiLatin1Encoding): Verify;
         verify(object: Object | KeyLike, signature: Binary): boolean;
-        verify(object: Object | KeyLike, signature: string, signature_format: HexBase64Latin1Encoding): boolean;
+        verify(object: Object | KeyLike, signature: string, signature_format?: HexBase64Latin1Encoding): boolean;
         // https://nodejs.org/api/crypto.html#crypto_verifier_verify_object_signature_signature_format
         // The signature field accepts a TypedArray type, but it is only available starting ES2017
+        private constructor();
     }
     function createDiffieHellman(prime_length: number, generator?: number | Binary): DiffieHellman;
     function createDiffieHellman(prime: Binary): DiffieHellman;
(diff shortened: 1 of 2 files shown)


types/node/crypto.d.ts
```diff
@@ -268,21 +268,23 @@ declare module "crypto" {
 
     type KeyLike = string | Buffer | KeyObject;
 
-    interface Signer extends NodeJS.WritableStream {
+    class Signer extends stream.Writable {
         update(data: BinaryLike): Signer;
         update(data: string, input_encoding: Utf8AsciiLatin1Encoding): Signer;
         sign(private_key: SignPrivateKeyInput | KeyLike): Buffer;
         sign(private_key: SignPrivateKeyInput | KeyLike, output_format: HexBase64Latin1Encoding): string;
+        private constructor();
```
SimonSchick: Nit: Please keep ctor ontop, even if it's private.

SimonSchick (approved):
Could you also address the other classes?
Cipher, HMac etc.

BrunoScheufler (approved):

types/node/test/crypto.ts
```diff
@@ -486,6 +486,41 @@ import { promisify } from 'util';
...
+    sign.write('some data to sign');
+    sign.end();
+    const signature: string = sign.sign(privateKey, 'hex');
+
+    const verify: crypto.Verify = crypto.createVerify('SHA256');
+    verify.write('some data to sign');
+    verify.end();
+    verify.verify(publicKey, signature);    // $ExpectType boolean
+
+    verify instanceof crypto.Verify;        // $ExpectType boolean
```
SimonSchick: The type checks here seem a little redundant, TS would error if this didn't work 😛

SimonSchick (approved):
Ty

rbuckton:
Thanks for the contribution!

types/node/test/crypto.ts
everett1992: Line 504 is significant. As far as I can tell this is the only line in the test file that will error if `Verify` is changed from a class to a `interface` or `type`.

Flarna: Yep, therefore I added it. The rest is just because there were no tests for these API in place.
But the `// $ExpectType boolean` is redundant and therefore I removed it.

PR #34233, opened 2019-03-26
add context to react relay

itajaja:
see https://github.com/facebook/relay/blob/master/packages/react-relay/index.js#L13

should I add tests?

### DIFF
--- a/types/react-relay/index.d.ts
+++ b/types/react-relay/index.d.ts
@@ -17,6 +17,8 @@
 export {
     commitLocalUpdate,
     commitRelayModernMutation as commitMutation,
+    CRelayContext,
+    Environment,
     fetchRelayModernQuery as fetchQuery,
     GraphQLTaggedNode,
     requestRelaySubscription as requestSubscription,
@@ -191,3 +193,9 @@ export function createRefetchContainer<P>(
     fragmentSpec: RelayRuntimeTypes.GraphQLTaggedNode | GeneratedNodeMap,
     taggedNode: RelayRuntimeTypes.GraphQLTaggedNode
 ): RelayContainer<P>;
+
+// ~~~~~~~~~~~~~~~~~~~~~
+// Context
+// ~~~~~~~~~~~~~~~~~~~~~
+
+export const ReactRelayContext: Context<CRelayContext<Environment>>;
--- a/types/react-relay/index.d.ts
+++ b/types/react-relay/index.d.ts
@@ -17,8 +17,6 @@
 export {
     commitLocalUpdate,
     commitRelayModernMutation as commitMutation,
-    CRelayContext,
-    Environment,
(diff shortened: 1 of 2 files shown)


types/react-relay/index.d.ts
```diff
@@ -17,6 +17,8 @@
 export {
     commitLocalUpdate,
     commitRelayModernMutation as commitMutation,
+    CRelayContext,
+    Environment,
```
alloy: These are exports, not imports, hence the CI failure.

alloy (changes requested):
Thanks!

Yeah, a test would be great 👍

types/react-relay/index.d.ts
itajaja: yes lol. I tried to cowboy my way from the github editor :P

alloy:
Will you be adding a test?

itajaja:
what do you think? is it needed? I have never done "typing" tests so your call

alloy:
Yes, I would like new additions to have test coverage. A test is really just a snippet of legit usage of the types.

types/react-relay/test/react-relay-tests.tsx
```diff
@@ -501,3 +502,11 @@ requestSubscription(
         },
     }
 );
+
+
+// ~~~~~~~~~~~~~~~~~~~~~
+// Context
+// ~~~~~~~~~~~~~~~~~~~~~
+
+ReactRelayContext.Consumer.prototype
```
itajaja: making sure that ReactRelayContext and it has the expected properties

types/react-relay/test/react-relay-tests.tsx
```diff
@@ -501,3 +502,10 @@ requestSubscription(
         },
     }
 );
+
+// ~~~~~~~~~~~~~~~~~~~~~
+// Context
+// ~~~~~~~~~~~~~~~~~~~~~
+
+ReactRelayContext.Consumer.prototype;
+ReactRelayContext.Provider.prototype;
```
alloy: I actually meant a snippet of actual API usage, but at a glance I only see `createFragmentSpecResolver` as a potential option and it doesn’t seem trivial to use that right now, so let’s just go with this for now 👍

alloy (approved):

types/react-relay/test/react-relay-tests.tsx
itajaja: yeah. I mean, I didn't actually specify the Context type or any other type in this PR, that already exists and is hopefully tested in relay-runtime. i just added a `const` in the module, so this tests that the `const` is indeed there 😅

alloy: Right right, I just try to identify opportunities for backfilling some of the debt we have as well ;)

PR #34241, opened 2019-03-26
Update tmp to 0.1

pluma:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/raszi/node-tmp/blob/master/lib/tmp.js

-----

This adjust the typings for version 0.1.0 and changes the export to use a TS CommonJS export object instead of ES6 module export syntax.

### DIFF
--- a/types/tmp/index.d.ts
+++ b/types/tmp/index.d.ts
@@ -1,43 +1,70 @@
-// Type definitions for tmp 0.0
+// Type definitions for tmp 0.1
 // Project: http://github.com/raszi/node-tmp
-// Definitions by: Jared Klopper <https://github.com/optical>, Gyusun Yeom <https://github.com/Perlmint>
+// Definitions by: Jared Klopper <https://github.com/optical>
+//                 Gyusun Yeom <https://github.com/Perlmint>
+//                 Alan Plum <https://github.com/pluma>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export interface Options extends SimpleOptions {
-    mode?: number;
-    discardDescriptor?: boolean;
-    detachDescriptor?: boolean;
-}
-
-export interface SimpleOptions {
+interface TmpNameOptions {
     prefix?: string;
     postfix?: string;
     template?: string;
     dir?: string;
     tries?: number;
+}
+
+interface FileOptions extends TmpNameOptions {
+    mode?: number;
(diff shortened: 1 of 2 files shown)


pluma:
This corrects some interfaces that were wonky before:

* synchronous dir results no longer have an "fd" attribute (which doesn't exist for dirs)
* removeCallback/name used everywhere instead of cleanupCallback/path for consistency
* the Options for files and dirs are actually different

pluma:
Pinging @optical and @Perlmint because reformatting seems to have confused the bot.

optical:
Hey @pluma. Looks like quite a big change and from what I can see it looks good, but unfortunately I haven't used `tmp` in years now and its been awhile since I wrote any meaningful type definitions or contributed to definitely typed. So I'll probably defer approving this to someone else more qualified

pluma:
After some testing I've realised the CommonJS export style actually doesn't make all that much sense as the module doesn't really export anything but an object of functions. Sorry about that.

pluma:
@optical that's fine.

From what I can see not much has actually changed in 0.1 although it's described as a breaking release. The existing typings already seem to be a bit off for 0.0 as the API documentation is a bit misleading. I dug through the actual implementation of 0.1 to figure out which options are actually supported by each method as it doesn't make sense to list things they don't actually use -- also directories don't return file descriptors, so that one was also wrong.

Most of the remaining changes are for consistency between names in the sync and async APIs (erring on the side of property names) and auto-formatting because I forgot to turn off prettier (with the default settings so hopefully this isn't considered gratuitous).

Perlmint (approved):
I just skimed this PR with code mentioned, It looks good.

mceachen:
@pluma thanks for doing this refactor, but this PR requires all nontrivial consumers of `tmp` to change their code (replacing `Options` to some sub-variant). It would have been nice to try to maintain existing interface names. 

Again, thanks for doing this spring cleaning.

pluma:
Sorry about that. At least 0.1 is a semver breaking release though, so hopefully this shouldn't catch anyone out of the blue if they're using semver ranges.

pluma:
FWIW I think the only avoidable case was the rename from SimpleOptions to TmpNameOptions. All the other name changes were the result of disambiguations (i.e. one type being replaced by several distinct but mutually incompatible ones).

PR #34249, opened 2019-03-26
[facepaint] Add type definitions

antonsamper:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/facepaint/index.d.ts
+++ b/types/facepaint/index.d.ts
@@ -1,0 +1,34 @@
+// Type definitions for facepaint 1.2
+// Project: https://github.com/emotion-js/facepaint
+// Definitions by: Ciarán Curley <https://github.com/DogPawHat>
+//                 Anton Samper Rivaya <https://github.com/antonsamper>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare module 'facepaint' {
+    type BaseArg = object | object[];
+    type Arg = BaseArg | BaseArg[];
+
+    type Selector = string;
+
+    interface DynamicStyle {
+        [key: string]: {
+            [key: string]: string | number;
+        };
+    }
+
+    interface DynamicStyleFunction {
+        (...args: Arg[]): DynamicStyle;
+    }
+
+    interface FacepaintOptions {
+        literal?: boolean;
+        overlap?: boolean;
+    }
+
+    function facepaint(
+        breakpoints: Selector[],
+        options?: FacepaintOptions
+    ): DynamicStyleFunction;
+
+    export default facepaint;
+}
--- a/types/facepaint/index.d.ts
+++ b/types/facepaint/index.d.ts
@@ -3,6 +3,7 @@
 // Definitions by: Ciarán Curley <https://github.com/DogPawHat>
 //                 Anton Samper Rivaya <https://github.com/antonsamper>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
 
 declare module 'facepaint' {
     type BaseArg = object | object[];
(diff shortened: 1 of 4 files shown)


antonsamper:
Fixes https://github.com/emotion-js/facepaint/issues/18

types/facepaint/index.d.ts
```diff
@@ -0,0 +1,35 @@
+// Type definitions for facepaint 1.2
+// Project: https://github.com/emotion-js/facepaint
+// Definitions by: Ciarán Curley <https://github.com/DogPawHat>
+//                 Anton Samper Rivaya <https://github.com/antonsamper>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
+declare module 'facepaint' {
```
plantain-00: You can just remove this line and `}`

types/facepaint/tslint.json
```diff
@@ -0,0 +1,7 @@
+{
+    "extends": "dtslint/dt.json",
```
plantain-00: This file should be only `"extends": "dtslint/dt.json"`

plantain-00 (changes requested):

antonsamper:
@plantain-00 When I make the changes you have suggested, I get the following error:

```
ERROR: 39:1  npm-naming  The types for facepaint specify 'export default' but the source does not mention 'default' anywhere. See: https://github.com/Microsoft/dtslint/blob/master/docs/npm-naming.md
```

Should I add `"npm-naming": false` to `tslint.json`?

plantain-00:
@antonsamper I cannot see how your error happens, can you push your commit to see how CI complains?

plantain-00:
It seems it checked the js file in the npm package to have text `export default`, but the package export commonjs release by default.
Maybe you should disable it for now @antonsamper .
@sandersn

types/facepaint/index.d.ts
```diff
@@ -0,0 +1,31 @@
+// Type definitions for facepaint 1.2
+// Project: https://github.com/emotion-js/facepaint
+// Definitions by: Ciarán Curley <https://github.com/DogPawHat>
+//                 Anton Samper Rivaya <https://github.com/antonsamper>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
+export type BaseArg = object | object[];
```
sandersn: The shipped code of `facepaint` exports a single value, a function, and has no `default` property [1], which is required by the module spec. So you need to match that with an `export =` instead of `export default`. It also means that the exported types will need to be put into a namespace:

```ts
declare namespace facepaint {
    export type BaseArg ...
    // others ...
}
declare function facepaint(breakpoints: facepaint.Selector[], options?: facepaint.Options): facepaint.DynamicStyleFunction;
export = facepaint;
```

And will need to be imported like so:

```ts
import facepaint = require('facepaint');
```

[1]

I tried this on npm.runkit.com:

```js
var f = require('facepaint')
console.log(f.default)
// prints undefined
```

sandersn (changes requested):
Need to use `export =` not `export default`.

sandersn (approved):

antonsamper:
Thanks @sandersn @plantain-00  for your feedback

types/facepaint/index.d.ts
plantain-00: The package can be used by `import default` actually:
https://github.com/emotion-js/facepaint/blob/master/src/index.js#L2
https://github.com/emotion-js/facepaint/blob/master/package.json#L6

sandersn: Only the "module" property points to a version with a default. `"main": "dist/index.cjs.js"` does not, which means that the following code would be broken without --esModuleInterop:

```ts
// @Filename: myfile.ts
import x from 'facepaint';
console.log(x)
```

becomes

```js
// @Filename: myfile.js
var x = require('facepaint').default
console.log(x)
```

And

```sh
$ node myfile.js
undefined
```

PR #33996, opened 2019-03-19
[react] Add unstable_Profiler

eps1lon:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).
  
If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: 
  - [Profiler usage](https://github.com/bvaughn/rfcs/blob/profiler/text/0000-profiler.md)
  - [onRender types](https://github.com/bvaughn/rfcs/blob/profiler/text/0000-profiler.md#detailed-design)
- [ ] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #34253, opened 2019-03-27
[react] RE: Add unstable_Profiler

eps1lon:
Copy of #33996

Any issue I can track about the state of this repo? Most of my PRs are now handled by a bot only and humans can only give advise that's essentially ignored. I do not understand how it is acceptable that the bot can close a PR that is accepted by an owner.

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -377,7 +377,7 @@ declare namespace React {
     interface ProfilerProps {
         children?: ReactNode;
         id: string;
-        onRender: OnRenderCallback;
+        onRender: ProfilerOnRenderCallback;
     }
 
     const unstable_Profiler: ExoticComponent<ProfilerProps>;
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -22,6 +22,7 @@
 //                 Paul Sherman <https://github.com/pshrmn>
 //                 Saransh Kataria <https://github.com/saranshkataria>
 //                 Kanitkorn Sujautra <https://github.com/lukyth>
+//                 Sebastian Silbermann <https://github.com/eps1lon>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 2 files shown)


ferdaber (approved):

ferdaber:
@eps1lon if local tests pass just cc @sandersn so he can merge manually.

eps1lon:
/cc @sandersn @Jessidhia

CI failure is unrelated. Could any human stop the bot please?

cscleison:
@eps1lon is there any workaround for typing `Set`? I need to target `es5`. I know I can just do `"lib": "es6"` but that implies the es6 apis are available at runtime (for ie11 I would have to add polyfills myself because `lib` it's just to tell typescript to not complain).

how to reproduce:
```
{
  "name": "test",
  "devDependencies": {
    "@types/react": "^16.9.1",
    "typescript": "^3.5.3"
  },
  "dependencies": {
    "react": "^16.9.0"
  }
}
```
```
{
  "compilerOptions": {
    "target": "es5",
    "strict": true,
    "jsx": "react",
    "outDir": "dist"
  },
  "include": ["./*.tsx"]
}
```
```

const Foo = () => <div>bar</div>;

export { Foo };
```

then run `npx tsc`

cscleison:
my workaround in the end was adding this, which i think is acceptable because `Map` and `Set` is partially supported by IE

`"lib": ["es5", "dom", "es2015.promise", "es2015.collection"]`

eps1lon:
@cscleison Be aware that [you always have to polyfill `Map` and `Set` when using react](https://reactjs.org/docs/javascript-environment-requirements.html). IE11 does not support all methods properly so you still need add polyfills for those missing methods.

types/react/index.d.ts
```diff
@@ -356,6 +366,26 @@ declare namespace React {
     const Suspense: ExoticComponent<SuspenseProps>;
     const version: string;
 
+    /**
+     * {@link https://github.com/bvaughn/rfcs/blob/profiler/text/0000-profiler.md#detailed-design | API}
```
intgr, line 370: FYI this is a broken link.

intgr: Opened a new PR #52735

PR #34256, opened 2019-03-27
[Tern] export constructor option interfaces

nkappler:
tern plugins may add custom constructor options.
to properly extend these interfaces they need to be exported.


rbuckton (approved):

rbuckton:
Thanks for the contribution!

PR #34265, opened 2019-03-27
Add typings for cuint

ltetzlaff:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/cuint/index.d.ts
+++ b/types/cuint/index.d.ts
@@ -1,0 +1,34 @@
+// Type definitions for cuint 0.2
+// Project: https://github.com/pierrec/js-cuint
+// Definitions by:  Lukas Tetzlaff <https://github.com/ltetzlaff>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export type UIntFactory<T extends UInt> = (
+    ...numbers: Array<number | string>
+) => T;
+
+export const UINT64: UIntFactory<UInt64>;
+export const UINT32: UIntFactory<UInt32>;
+
+export class UInt {
+    protected constructor()
+
+    clone(): this;
+    add(x: this): this;
+    subtract(x: this): this;
+    multiply(x: this): this;
+    xor(x: this): this;
+    rotl(n: number): this;
+    shiftRight(n: number): this;
+
+    fromNumber(n: number): this;
+    fromBits(...bits: number[]): this;
+
+    _low: number;
+    _high: number;
+
+    toString(base?: number): string;
+}
+
+export class UInt64 extends UInt {}
+export class UInt32 extends UInt {}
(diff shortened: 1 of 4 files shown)


types/cuint/index.d.ts
```diff
@@ -0,0 +1,34 @@
...
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export type UIntFactory<T extends UInt> = (
+    ...numbers: Array<number | string>
+) => T;
+
+export const UINT64: UIntFactory<UInt64>;
+export const UINT32: UIntFactory<UInt32>;
+
+export class UInt {
```
rbuckton: `UInt` is declared here as both a type and a value because it is defined as a `class`, however the value-side isn't actually an export you can reference.

For these definitions, you might be better with something like:

```ts
export interface Uint {
    // ... every member currently in `UInt` except `constructor` ...
}

export interface UintConstructor {
  // Uint called as a function:
  (value: number): Uint;
  (low: number, high: number): Uint;
  (text: string, radix?: number): Uint;

  // Uint called as a constructor:
  new (value: number): Uint;
  new (low: number, high: number): Uint;
  new (text: string, radix?: number): Uint;

  prototype: Uint;
}

export const UINT64: UintConstructor;
export const UINT32: UintConstructor;
```

types/cuint/index.d.ts
```diff
@@ -0,0 +1,34 @@
...
+    fromNumber(n: number): this;
+    fromBits(...bits: number[]): this;
+
+    _low: number;
+    _high: number;
+
+    toString(base?: number): string;
+}
+
+export class UInt64 extends UInt {}
```
rbuckton: `UInt64` is declared here as both a type and a value because it is defined as a `class`, however the value-side isn't actually an export you can reference.

types/cuint/index.d.ts
```diff
@@ -0,0 +1,34 @@
...
+export class UInt32 extends UInt {}
```
rbuckton: `UInt32` is declared here as both a type and a value because it is defined as a `class`, however the value-side isn't actually an export you can reference.

rbuckton (changes requested):

types/cuint/index.d.ts
ltetzlaff: Cool, thx for the comment 👍

I maintained your suggestions from above while keeping 64-bit and 32-bit separate through generics to not make them interchangeable when using the lib.

ltetzlaff:
Using interfaces didn't really align well with the idea of two identical but different types, for the sake of simplicity it might make more sense to just take `Uint`. I did add `tslint:disable-next-line:unified-signatures` to allow these semantically distinct usage cases (I did not design the API in the original module 🙃)

DanielRosenwasser:
Everything good @rbuckton?

PR #34268, opened 2019-03-27
[sinon] Correct comparable type

JimiC:
Typescript complains for functions `calledBefore`, `calledAfter`, `calledImmediatelyBefore` and `calledImmediatelyAfter` because it's trying to compare a `SinonInspectable` to a `SinonSpy`. Correct comparison should be done on the same type (i.e. `SinonInspectable`).

Please fill in this template.

### DIFF
--- a/types/sinon/ts3.1/index.d.ts
+++ b/types/sinon/ts3.1/index.d.ts
@@ -238,22 +238,22 @@ declare namespace Sinon {
          * Returns true if the spy was called before @param anotherSpy
          * @param anotherSpy
          */
-        calledBefore(anotherSpy: SinonSpy): boolean;
+        calledBefore(anotherSpy: SinonInspectable): boolean;
         /**
          * Returns true if the spy was called after @param anotherSpy
          * @param anotherSpy
          */
-        calledAfter(anotherSpy: SinonSpy): boolean;
+        calledAfter(anotherSpy: SinonInspectable): boolean;
         /**
          * Returns true if spy was called before @param anotherSpy, and no spy calls occurred between spy and @param anotherSpy.
          * @param anotherSpy
          */
-        calledImmediatelyBefore(anotherSpy: SinonSpy): boolean;
+        calledImmediatelyBefore(anotherSpy: SinonInspectable): boolean;
         /**
          * Returns true if spy was called after @param anotherSpy, and no spy calls occurred between @param anotherSpy and spy.
          * @param anotherSpy
          */
-        calledImmediatelyAfter(anotherSpy: SinonSpy): boolean;
+        calledImmediatelyAfter(anotherSpy: SinonInspectable): boolean;
         /**
          * Returns true if the spy was always called with @param obj as this.
          * @param obj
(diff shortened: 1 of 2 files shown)


nicojs (changes requested):
Thanks @JimiC 

Could you add a test for this? I guess this enables something that wasn't possible before. By adding a test, you'll help prevent regression later on.

SimonSchick (approved):

JimiC:
Let me explain the use case I faced and why I'm proposing this change.

In a project that targets `ts v3.3.4000` I have code that looks like this:
```
    const spy1 = sinon.spy(Class1, 'method1');
    const spy2 = sinon.spy(Class1, 'method2');
    expect(spy1.calledBefore(spy2)).to.be.true;
```
Now `ts` infers:
- `spy1` as `sinon.SinonStub<[type1, (type2 | undefined)?], void>`
- `spy2` as `sinon.SinonStub<[type1, type2], void>`

but when you hover over `spy1` in line 3 (the `expect` one) `ts` infers `spy1` as `sinon.SinonInspectable<[type1, (type2 | undefined)?], void>` (although `SinonStub` extends `SinonInspectable`) in order to call the `calledBefore` function and presents a red squiggly under `spy2` which now is infered as `sinon.SinonStub<any[], any>`. `ts` complains that `sinon.SinonStub<[type1, type2], void>` can't be assigned to `sinon.SinonStub<any[], any>`.

If I cast `spy2` to `SinonInspectable` or if I switch the parameter type of `calledBefore` to `SinonInspectable`, `ts` is happy.

I tested that this proposed change does not affect if you use:
```
    const spy1 = sinon.spy(); // sinon.SinonStub<any[], any>
    const spy2 = sinon.spy(); //`sinon.SinonStub<any[], any>
    expect(spy1.calledBefore(spy2)).to.be.true;
```

If you look in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sinon/index.d.ts you will notice that `calledBefore` is a member of `SinonSpy` and thus having the parameter type of `calledBefore` as `SinonSpy` makes perfect sense. But in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sinon/ts3.1/index.d.ts `calledBefore` is a member of `SinonInspectable` and thus the parameter type of `calledBefore` and the other related functions owe to be of the same type i.e. `SinonInspectable` .

A test case that demonstrates this could be added for `methodSpy` and `methodSpy2` (https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sinon/ts3.1/sinon-tests.ts#L390) like 
```
    methodSpy.calledBefore(methodSpy2);
    methodSpy.calledAfter(methodSpy2);
    methodSpy.calledImmediatelyBefore(methodSpy2);
    methodSpy.calledImmediatelyAfter(methodSpy2);
```

Because I'm not curently on my development workstation, I'll have to came back to you for that later.

nicojs (approved):
Thanks for the clarification and tests. Looks awesome now 👍

rbuckton:
Thanks for the contribution!

PR #34271, opened 2019-03-27
[btoa] Accept Buffer

nikolay-borzov:
According to [the source code](https://git.coolaj86.com/coolaj86/btoa.js/src/branch/master/index.js) `btoa` function accepts both `string` and `Buffer`

### DIFF
--- a/types/btoa/btoa-tests.ts
+++ b/types/btoa/btoa-tests.ts
@@ -1,3 +1,5 @@
 import btoa = require('btoa');
 
 btoa('foo');
+
+btoa(Buffer.from('foo'));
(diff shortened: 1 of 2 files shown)


types/btoa/btoa-tests.ts
```diff
@@ -1,3 +1,6 @@
 import btoa = require('btoa');
+import fs = require('fs');
 
 btoa('foo');
+
+btoa(fs.readFileSync('foo.css'));
```
bricka: Might be simpler to just use `Buffer.from` instead of going through `fs`.

bricka (approved):

types/btoa/btoa-tests.ts
nikolay-borzov: Right, replaced by `Buffer.from('foo')`

bricka (approved):

rbuckton:
Thanks for the contribution!

Issue #24788, opened 2018-04-07
Cannot find module 'csstype'


liketurbo:
Hi, I am trying to use react with typescript, I installed (`react`, `react-dom`), (`@types/react`, `@types/react-dom`), in attempt to launch project I get error `ERROR in [at-loader] ./node_modules/@types/react/index.d.ts:55:22` `Cannot find module 'csstype'`. I tried `npm i csstype` in `@types/react` and in root of project. Also despite to error, in module `@types/react` exists folder node_modules/csstype, so I don't even know why this error happens.

***tsconfig.json***
```
{
    "compilerOptions": {
        "outDir": "./build/", // path to output directory
        "strictNullChecks": true, // enable strict null checks as a best practice
        "module": "es6", // specify module code generation
        "jsx": "react", // use typescript to transpile jsx to js
        "target": "es5", // specify ECMAScript target version
        "allowJs": true, // allow a partial TypeScript and JavaScript codebase
        
        "noImplicitAny": true,
        "noImplicitReturns": true,
        "noUnusedLocals": true,
        "noUnusedParameters": true,
    },
    "exclude": [
        "node_modules"
    ],
    "include": [
        "./source/**/*.tsx",
        "./source/**/*.ts"
    ]
}
```   
**package.json**
```
{
  "name": "suggestion-app",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "scripts": {
    "webpack:server": "babel-node node_modules/.bin/webpack-dev-server --mode=development"
  },
  "dependencies": {
    "react": "^16.3.1",
    "react-dom": "^16.3.1"
  },
  "devDependencies": {
    "@types/react": "^16.3.5",
    "@types/react-dom": "^16.0.4",
    "babel-cli": "^6.26.0",
    "babel-preset-env": "^1.6.1",
    "babel-preset-stage-3": "^6.24.1",
    "caniuse-lite": "^1.0.30000697",
    "clean-webpack-plugin": "^0.1.19",
    "css-loader": "^0.28.11",
    "eslint": "^4.19.1",
    "eslint-config-airbnb": "^16.1.0",
    "eslint-plugin-import": "^2.10.0",
    "eslint-plugin-jsx-a11y": "^6.0.3",
    "eslint-plugin-react": "^7.7.0",
    "file-loader": "^1.1.11",
    "glob": "^7.1.2",
    "html-webpack-plugin": "^3.2.0",
[...]

PR #34276, opened 2019-03-27
[graphql-react] Add initial type definitions

mike-marcacci:
Please fill in this template.

If adding a new definition:


mike-marcacci:
Wow, so... now that the azure checks are passing, I have a bit of feedback for DT. This is my first time contributing here, and oh boy is it painful. Writing the type definitions were trivial, but the rest is not straightforward, even when reading all the documentation.

What's most frustrating is that the tooling doesn't work consistently:
- The `dts-gen --dt` tool generates a `tsconfig.json` that's invalid according to the linter rules
- Running `npm test` NEVER works locally, and tries to test every repo
- Running the linter doesn't always work locally, because it checks files of types that are imported... which can have different lint configs
- Some types won't import correctly (see the tests in this PR... Koa doesn't have a default export? Yes it does!)
- You can't use JSX in a `.ts` file, but you can't rename your test file to `.tsx`, or add an additional txt file to reference in your test file
- `dtslint` is overly-opinionated in ways that actually limit the ability to correctly express types (I'm thinking of `no-unnecessary-generics` in particular.)

Overall, DT is fantastic to consume types from (and great job on making that happen), but unbelievably frustrating to contribute types to. This is too bad, because I'm sure there are others like me who have complete, high-quality definitions sitting in our project repos, which we'd gladly contribute if the process didn't take several hours.

rbuckton:
cc: @sandersn since he's been working on some of the DT tooling recently.

LeartS:
I don't find the situation as bad as @mike-marcacci , but I also encountered part of the problems he highlighted and I share the sentiment.

types/graphql-react/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+        "strictNullChecks": true,
+        "strictFunctionTypes": true,
+        "baseUrl": "../",
+        "typeRoots": [
+            "../"
+        ],
+        "types": [],
+        "noEmit": true,
+        "forceConsistentCasingInFileNames": true,
+        "jsx": "react"
```
uniqueiniquity: In order to use the default import of React in the test file, I added `esModuleInterop: true` here.
Because the type declaration file for `react` doesn't have a default export, you need this flag if you want to use `import React from "react"`. I'm not exactly sure why the authors of the `react` d.ts chose not to provide a default export, but there seems to be some oddness going on with whether or not the module actually has a default property at runtime or not. (cc @weswigham for thoughts).

types/graphql-react/graphql-react-tests.ts
```diff
@@ -0,0 +1,79 @@
...
+      query($id: !String) {
+        node(id: $id) {
+          id
+        }
+      }
+    `
+  }
+});
+
+/* This SHOULD be the test, but the DefinitelyTyped tooling is so difficult to
```
uniqueiniquity: I was able to rename this file to `graphql-react-tests.tsx`, comment out the top portion and uncomment the code below, and it passed the linter.

types/graphql-react/tsconfig.json
```diff
@@ -0,0 +1,25 @@
...
+    },
+    "files": [
+        "index.d.ts",
+        "graphql-react-tests.ts"
```
uniqueiniquity: I changed this line to `"graphql-react-tests.tsx"`.

uniqueiniquity (changes requested):
As I mentioned in your other PR, I'm very sorry that we didn't provide clearer instructions and caused you to spend several hours working on this.

I was able to make some local changes to your PR and get it to pass the linter fine. I'll detail each inline.
Let me know if when you make these changes locally, `npm run lint graphql-react` still finds issues, or if you have any questions in general.

Please keep the feedback coming. :)

EDIT: one more thought - I can certainly commit the changes I made, but I wouldn't want to commit onto your PR without your permission.

types/graphql-react/tsconfig.json
mike-marcacci: This is a great tip, and helped me fix an issue in [a different PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34277).

However, my challenge here is a bit different. When I run `npm run lint graphql-react`, I get the following in the console:

```
> definitely-typed@0.0.3 lint /Users/mike/Code/DefinitelyTyped
> dtslint types "graphql-react"
















../react/index.d.ts(30,22): error TS2307: Cannot find module 'csstype'.

    at /Users/mike/Code/DefinitelyTyped/node_modules/dtslint/bin/index.js:189:19
    at fulfilled (/Users/mike/Code/DefinitelyTyped/node_modules/dtslint/bin/index.js:5:58)
```

I'm on mac with node v11.12.0 and a fresh `node_modules`. Installing the module globally has no effect, and there is no `@types/csstype`. I would have dug into the react type definitions more, but this seems not to be a problem in CI. Of course, this leads to that painfully slow feedback loop: change, commit, push, wait, repeat.

types/graphql-react/tsconfig.json
mike-marcacci: 🤦🏻‍♂️that's the part I was missing... I had tried creating both files, and simply importing the `.tsx` from the `.ts`. This, though, caused the linter to fail because it thought it was an "unnecessary file."

types/graphql-react/tsconfig.json
mike-marcacci: It seems like I'm not the only one with [this issue](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/24788), so I'll go ahead and report it over there.

mike-marcacci: Well, I figured it out, and posted an issue in [dtslint](https://github.com/Microsoft/dtslint/issues/226)

uniqueiniquity (approved):

PR #34277, opened 2019-03-27
[graphql-api-koa] Add initial type definitions

mike-marcacci:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/graphql-api-koa/graphql-api-koa-tests.ts
+++ b/types/graphql-api-koa/graphql-api-koa-tests.ts
@@ -1,0 +1,21 @@
+import Koa from "koa";
+import { GraphQLSchema, GraphQLObjectType, GraphQLString } from "graphql";
+import { errorHandler, execute } from "graphql-api-koa";
+
+const app = new Koa()
+  .use(errorHandler())
+  .use(
+    execute({ schema: new GraphQLSchema({
+      query: new GraphQLObjectType({
+        name: "Query",
+        fields: () => ({
+          hello: {
+            type: GraphQLString,
+            resolve() {
+              return "world";
+            }
+          }
+        })
+      })
+    })
+  }));
--- a/types/graphql-api-koa/graphql-api-koa-tests.ts
+++ b/types/graphql-api-koa/graphql-api-koa-tests.ts
@@ -1,3 +1,27 @@
+import { GraphQLSchema, GraphQLObjectType, GraphQLString } from "graphql";
+import { execute } from "graphql-api-koa";
+
+execute({
+  schema: new GraphQLSchema({
+    query: new GraphQLObjectType({
+      name: "Query",
+      fields: () => ({
+        hello: {
+          type: GraphQLString,
+          resolve() {
+            return "world";
+          }
+        }
+      })
+    })
+  })
+});
+
+/* This is the thest that SHOULD exist here... but for whatever reason, the
+   DefinitelyTyped repo always errors with:
+
+   Module '"/Users/mike/Code/DefinitelyTyped/types/koa/index"' has no default export.
+
 import Koa from "koa";
 import { GraphQLSchema, GraphQLObjectType, GraphQLString } from "graphql";
 import { errorHandler, execute } from "graphql-api-koa";
@@ -19,3 +43,4 @@ const app = new Koa()
       })
     })
   }));
+*/
--- a/types/graphql-api-koa/index.d.ts
+++ b/types/graphql-api-koa/index.d.ts
@@ -7,20 +7,20 @@
 import { GraphQLSchema } from "graphql";
 import { Middleware, ParameterizedContext } from "koa";
 
-interface ExecuteOptions {
+export interface ExecuteOptions {
   schema?: GraphQLSchema;
   rootValue?: any;
   contextValue?: any;
   fieldResolver?: any;
 }
 
-export const errorHandler: <StateT = any, CustomT = {}>() => Middleware<
+export function errorHandler<StateT = any, CustomT = {}>(): Middleware<
   StateT,
   CustomT
 >;
 
-export const execute: <StateT = any, CustomT = {}>(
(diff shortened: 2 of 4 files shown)


mike-marcacci:
Wow, so... now that the azure checks are passing, I have a bit of feedback for DT. This is my first time contributing here, and oh boy is it painful. Writing the type definitions were trivial, but the rest is not straightforward, even when reading all the documentation.

What's most frustrating is that the tooling doesn't work consistently:
- The `dts-gen --dt` tool generates a `tsconfig.json` that's invalid according to the linter rules
- Running `npm test` NEVER works locally, and tries to test every repo
- Running the linter doesn't always work locally, because it checks files of types that are imported... which can have different lint configs
- Some types won't import correctly (see the tests in this PR... Koa doesn't have a default export? Yes it does!)
- You can't use JSX in a `.ts` file, but you can't rename your test file to `.tsx` (see [this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34276))

Overall, DT is fantastic to consume types from (and great job on making that happen), but unbelievably frustrating to contribute types to. This is too bad, because I'm sure there are others like me who have complete, high-quality definitions sitting in our project repos, which we'd gladly contribute if the process didn't take several hours.

uniqueiniquity:
@mike-marcacci Thank you so much for the feedback! I really appreciate that you took the time to do that.

Let me see if I can address each of your points in order. I'm also tagging @sandersn, who has been focusing on improvements to DT.

- If dts-gen is doing that, we should definitely change that :)
- You certainly should be able to test locally. My usual recommendation is to run `npm run lint <package name>` as mentioned in the README. Clearly, we should make sure the README makes that a lot clearer so contributors like yourself don't get misled about the best way to test a contribution.
- Running the linter as mentioned in the previous bullet should do the right thing. If you can easily repro a package that seems to be giving the linter trouble, could you try it with that npm script and let me know what the repro is?
- I'm a little confused about your point about `koa` having a default export. Could you elaborate on this point? I may be missing something, so I want to make sure I understand before going into details.
- You should be able to use a TSX file for your tests. See [`react-adal`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-adal/react-adal-tests.tsx) as just one example.
- I agree that `no-unnecessary-generics` causes a lot of trouble for contributors. We can certainly either revisit having that rule enabled by default, improve the precision of the rule, or improve the documentation to explain exactly why we have it enabled (I personally find the existing documentation for the rule to be lacking). Were there other lint rules that you found to be excessive?

Please, please let me know if you have any other thoughts. We really want the contribution process to be as painless as possible, and knowing what exactly you ran into will greatly help us in improving towards that goal. Thanks again!

uniqueiniquity:
Also, congratulations on your first DefinitelyTyped contribution!

mike-marcacci:
Hey @uniqueiniquity! Thanks so much for the thoughtful comments and the merge! I'll address your notes specifically:

- The tsconfig generated by `dts-gen` is missing the `"strictFunctionTypes": true` config, which is required by the lint rules, which I don't fully understand given that [it's in the codebase](https://github.com/Microsoft/dts-gen/blob/0ca14f04ba9c4921dfd1d11e360e54f3b9c4c391/lib/definitely-typed.ts#L103)
- Based on your comment, I suspect we aren't supposed to actually try to run `npm test` locally. If that's true, we should [probably change that in the PR template](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34426).
- My [other PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34276) (sorry for the cross-post, btw) gave me more serious trouble, which I'll address over there.
- I just fixed the Koa import issue in [this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34427) by adding `esModuleInterop`... perhaps this should be enabled by default in the generated `tsconfig.json`?
- I'll elaborate on my jsx issue in my other PR.
- In general, the lint rules were a bit opinionated (not having redundant `public` in front of methods, etc), but I totally respect the desire to have a consistent codebase. I was very hesitant to change or disable any of these rules, figuring that they wouldn't be there if the maintainers didn't care about them... but by the time I got to wrestling `no-unnecessary-generics` (which actually effects the way exported types can be used) I almost gave up and left a working PR with a failing build to rot and get closed by the bot.

I sincerely hope this helps, and doesn't come across as too ranty – I'm totally blown away by the expressiveness and community momentum that TypeScript has gained in the past few years. You all are doing a great job, and I figured I'd share my first-time feedback while I had it (you're only a first-timer once, after all).

Cheers!

sandersn:
Thanks again for the detailed feedback (and bug report on dtslint).

* dts-gen is something we haven't had time to touch in a while and *definitely* needs some work.
* `npm test` *is* the intended way to test, especially for packages with lots of dependents, since it mimics what CI will do, for example `npm install` in packages that need it. `npm run lint your-package-name` only checks the current package. However, any errors in local `npm test` runs should be treated as bugs. Thanks for filing the one to track `npm install` of dependencies.
* esModuleInterop is not a general solution since it requires consumers of the package to have it set too. For certain communities, react for example, this is a given. Koa might be the same? I'm not sure.
* I don't understand the jsx issue but it sounds like @uniqueiniquity does so I'll ask him about it in person.
* The lint rules are ... somewhat stale opinions at this point. In particular, we use `tslint:all` and then selectively disable rules. I don't like this approach and intend to change it during the eslint switch (whenever that happens -- after Typescript switches, I think). However, I do know that `no-unnecessary-generics` is a favourite of @RyanCavanaugh so it's not going away. [I tried to explain the reasoning here](https://github.com/RyanCavanaugh/TNTSHB/blob/master/intros/TypeScript%20for%20the%20Haskell%20or%20ML%20Programmer.md#type-parameters).

Thanks for sticking with it. DT has perennial growing pains and we're trying to get through the latest round right now.

PR #34280, opened 2019-03-28
[nodegit] Update blob.d.ts

saoudrizwan:
Fixes Blob definitions according to https://www.nodegit.org/api/blob/


dolanmiu (approved):

rbuckton:
Thanks for the contribution!

saoudrizwan:
@rbuckton @dolanmiu Glad to contribute! Thank you for all your guys' contributions too! I plan on using nodegit pretty extensively, so I will make any fixes as I see them along the way.

dolanmiu:
Sure! sounds great

PR #34284, opened 2019-03-28
Fix type of ChildProcess.channel

KSXGitHub:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/api/child_process.html#child_process_subprocess_channel

---

From [the documentation](https://nodejs.org/api/child_process.html#child_process_subprocess_channel):

> If no IPC channel currently exists, this property is `undefined`.

### DIFF
--- a/types/node/child_process.d.ts
+++ b/types/node/child_process.d.ts
@@ -7,7 +7,7 @@ declare module "child_process" {
         stdin: Writable | null;
         stdout: Readable | null;
         stderr: Readable | null;
-        readonly channel?: Pipe | null;
+        readonly channel?: Pipe;
         readonly stdio: [
             Writable | null, // stdin
             Readable | null, // stdout


types/node/child_process.d.ts
```diff
@@ -7,7 +7,7 @@ declare module "child_process" {
         stdin: Writable | null;
         stdout: Readable | null;
         stderr: Readable | null;
-        readonly channel?: Pipe | null;
+        readonly channel?: Pipe;
```
Flarna: I tried following in repl:
```
> let a = cp.spawn("dir", { stdio: [null, null, null, 'ipc'] })
undefined
> a.channel
null
```
So it seems `null` happens.

KSXGitHub: @Flarna So it is the Node.js's docs that is insufficient. I'm closing this.

Flarna: Yes, seems so. `null` and `undefined` are often intermixed as people tend to use `== null` or `== undefined` checks which end up in the same.

PR #34285, opened 2019-03-28
Add overloads to child_process.spawn for better type inference

KSXGitHub:
Without `options.stdio`, `cp.{stdin,stdout,stderr}` should not be `null`.

I make this change because I don't want to check for `null` when I am sure `stdout` is not `null`.

---

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/api/child_process.html#child_process_subprocess_stdout

> If the child was spawned with `stdio[n]` set to anything other than `'pipe'`, then this will be `null`.

### DIFF
--- a/types/node/child_process.d.ts
+++ b/types/node/child_process.d.ts
@@ -79,7 +79,7 @@ declare module "child_process" {
     }
 
     // return this object when stdio option is undefined or not specified
-    interface ChildProcessWithoutNullStreams {
+    interface ChildProcessWithoutNullStreams extends ChildProcess {
         stdin: Writable;
         stdout: Readable;
         stderr: Readable;
--- a/types/node/child_process.d.ts
+++ b/types/node/child_process.d.ts
@@ -78,6 +78,20 @@ declare module "child_process" {
         prependOnceListener(event: "message", listener: (message: any, sendHandle: net.Socket | net.Server) => void): this;
     }
 
+    // return this object when stdio option is undefined or not specified
+    interface ChildProcessWithoutNullStreams {
+        stdin: Writable;
+        stdout: Readable;
+        stderr: Readable;
+        readonly stdio: [
+            Writable, // stdin
+            Readable, // stdout
+            Readable, // stderr
+            Readable | Writable | null | undefined, // extra, no modification
+            Readable | Writable | null | undefined // extra, no modification
+        ]
+    }
+
     interface MessageOptions {
         keepOpen?: boolean;
     }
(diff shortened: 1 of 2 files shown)


types/node/child_process.d.ts
```diff
@@ -78,6 +78,20 @@ declare module "child_process" {
         prependOnceListener(event: "message", listener: (message: any, sendHandle: net.Socket | net.Server) => void): this;
     }
 
+    // return this object when stdio option is undefined or not specified
+    interface ChildProcessWithoutNullStreams {
```
Flarna: Properties like `killed`, `pid` are missing here. I think some common part should be used or this type only specializes the stdio relevant part of `ChildProcess`.

types/node/child_process.d.ts
```diff
@@ -110,7 +124,13 @@ declare module "child_process" {
         windowsVerbatimArguments?: boolean;
     }
 
+    interface SpawnOptionsWithoutStdio extends SpawnOptions {
+        stdio?: 'pipe' | Array<undefined | 'pipe'>;
+    }
+
+    function spawn(command: string, options?: SpawnOptionsWithoutStdio): ChildProcessWithoutNullStreams;
```
Flarna: I think at least one of the two `spawn` overloads should have the second argument mandatory. Same applies to the other variant two lines below.

types/node/child_process.d.ts
```diff
@@ -110,7 +124,13 @@ declare module "child_process" {
         windowsVerbatimArguments?: boolean;
     }
 
+    interface SpawnOptionsWithoutStdio extends SpawnOptions {
+        stdio?: 'pipe' | Array<undefined | 'pipe'>;
```
Flarna: I think `null` is also allowed as member in the array.

Flarna (changes requested):

types/node/child_process.d.ts
KSXGitHub: Is `null` included in the documentation though? I ask this because as long as something is not in the documentation, Node.js developers are free to change the behavior without concern of breaking changes.

Flarna: see https://nodejs.org/dist/latest-v11.x/docs/api/child_process.html#child_process_options_stdio
`null, undefined - Use default value. For stdio fds 0, 1, and 2 (in other words, stdin, stdout, and stderr) a pipe is created. For fd 3 and up, the default is 'ignore'.` 

like in a lot cases `null` and `undefined` are interchangeable here.

KSXGitHub: OK. I'll fix this. But I have another question: Since `options.stdio` itself can also be `null`, should I also modify `SpawnOptions.stdio`'s type from `StdioOptions` to `StdioOptions | null`?

KSXGitHub:
@Flarna I've made changes as you requested. Can you take a look?

types/node/child_process.d.ts
Flarna: I don't think it's needed to allow `options.stdio === null` as it's not documented.
Actual implementation in node does `options.stdio || 'pipe'` so actually everything falsy like `false`, `0` or `''` ends up in `pipe` but I don't think we should allow this in the type definitions except if there is really a good reason to do this.

types/node/child_process.d.ts
KSXGitHub: It was meant to `extends ChildProcess` but I forgot.

KSXGitHub: Fixed with https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34285/commits/0237165eb143dbddfcc45a403df34cf8cc010a33

Flarna (approved):

BrunoScheufler (approved):

PR #34289, opened 2019-03-28
[slate] support marks in schema

ivan-ha:
Support `marks` in slate schema
https://docs.slatejs.org/slate-core/schema#rule-properties

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://docs.slatejs.org/slate-core/schema#rule-properties>>

### DIFF
--- a/types/slate/index.d.ts
+++ b/types/slate/index.d.ts
@@ -39,6 +39,9 @@ export interface Rules {
     first?: ObjectAndType | ObjectAndType[];
     isVoid?: boolean;
     last?: ObjectAndType | ObjectAndType[];
+    marks?: Array<{
+        type: string | ((type: string) => boolean)
+    }>;
     nodes?: Array<{
         min?: number;
         max?: number;
--- a/types/slate/slate-tests.ts
+++ b/types/slate/slate-tests.ts
@@ -74,10 +74,12 @@ const schema: SchemaProperties = {
                 }
             }
         },
+        marks: [{ type: 'bold' }, { type: t => ['bold', 'underline'].indexOf(t) !== -1 }]
     },
     blocks: {
         image: {
             isVoid: true,
+            marks: [{ type: 'bold' }, { type: t => ['bold', 'underline'].indexOf(t) !== -1 }]
         },
     },
 };


sgreav:
Sorry I missed this, I was on vacation! Reviewing now

types/slate/index.d.ts
```diff
@@ -39,6 +39,7 @@ export interface Rules {
     first?: ObjectAndType | ObjectAndType[];
     isVoid?: boolean;
     last?: ObjectAndType | ObjectAndType[];
+    marks?: object[];
```
sgreav: Use the `Mark` type defined in this file

`marks?: Mark[];`

hanna-greaves (changes requested):
Thanks for the contribution @ivan-ha : ) I've just got one change to request

Sorry again for the delay in reviewing!

ivan-ha:
no worries @sgreav , I was on vacation too 😄 

thanks for review, fixed accordingly!

ivan-ha:
@sgreav CI failed after I changed to `marks?: Mark[]`

I think it's because my test is plain js array of object, without those `toJSON` etc stuff as declared in `Mark`

>```
>{
>  marks: [{ type: 'bold' }, { type: 'italic' }]
>}
>```
And according to [the doc](https://docs.slatejs.org/slate-core/schema#marks), isn't `marks` takes only an array of object with a `type` property?

benjiro:
@ivan-ha
Thanks for your contribution, since this is using SlateJS uses immutableJs you'll need to call the create method on the class to return a new Mark object.

> Mark.createList([{ type: 'bold' }, { type: 'italic' }])

ivan-ha:
hi @benjiro, got your point, however seems there is no `createList` defined in the `Mark` class, is it missed?

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0bb3d1863d47bf0c6c6f39e5000a400f33a10f04/types/slate/index.d.ts#L532-L546

Screenshot 2019-04-09 at 10 54 11

jackall3n:
Hey, I think createSet is the way forwards, as I expect the Marks should be unique.

benjiro:
@ivan-ha Looks like createSet is what you want, which takes an array marks. Give that a shot

ivan-ha:
ya @jackall3n @benjiro, I give that a shot earlier today too, however encounter another error

Screenshot 2019-04-09 at 17 21 20

ivan-ha:
@benjiro yes, CI passed after I change the `Rules` interface to `Immutable.Set<Mark>`, thanks

however I still got a doubt, so now in `schema.marks`, one needs to put `Mark.createSet([{ type: 'bold' }, { type: 'italic' }])` for instance, this seems deviate from [the doc](https://docs.slatejs.org/slate-core/schema#marks), any idea?

benjiro:
From my understand that is due to the type system, the current documentation is written for a pure JS implementation which would dynamically cast that as a immutable set.

isubasti:
@benjiro the schema's rules for mark accept `array` not `set` so the changes using `Mark.createSet` is wrong as it doesn't match what's actually implemented inside slate

types/slate/index.d.ts
```diff
@@ -39,6 +39,7 @@ export interface Rules {
     first?: ObjectAndType | ObjectAndType[];
     isVoid?: boolean;
     last?: ObjectAndType | ObjectAndType[];
+    marks?: Immutable.Set<Mark>;
```
isubasti: this should be an Array<{type: string}> instead of `Immutable.Set<Mark>;`

isubasti (changes requested):

types/slate/index.d.ts
isubasti: actually it should be `Array<{ type: ((type: string) => boolean) | string }>` as it now receives a function as well

ivan-ha: right @isubasti , I think the function support is from this PR https://github.com/ianstormtaylor/slate/pull/2151

ivan-ha:
thanks @isubasti , updated the def and test, please check again!

isubasti (approved):
LGTM

benjiro (approved):
@isubasti Cheers thanks for clearing that up.

ivan-ha:
@sgreav can you update the review status? thanks!

ivan-ha:
This PR was approved by 2 `slate` contributor already, however still blocked in the `revision needed` column therefore cannot move on, since @sgreav requested change before.

@RyanCavanaugh @uniqueiniquity @sheetalkamat could you help to review it? Thanks!

hanna-greaves (approved):

sgreav:
I'm really sorry, I hadn't realised I was blocking this

ivan-ha:
it's ok, no worries @sgreav 😄

PR #34296, opened 2019-03-28
semver: add missing properties and functions, correct some parameter types

ajafff:

I initially only wanted to fix `Comparator.value` and add the `minVersion` function.
I fixed some parameter and return types to match the actual runtime behavior.
I changed mutable Arrays to ReadonlyArray where appropriate.
I also changed `loose?: boolean` to `looseOrOptions?: boolean | Options`:

> All methods and classes take a final options object argument. All options in this object are false by default. The options supported are:
>
>* `loose` Be more forgiving about not-quite-valid semver strings. (Any resulting output will always be 100% strict compliant, of course.) For backwards compatibility reasons, if the options argument is a boolean value instead of an object, it is interpreted to be the loose param.
> * `includePrerelease` Set to suppress the default behavior of excluding prerelease tagged versions from ranges unless they are explicitly opted into.

### DIFF
--- a/types/semver/index.d.ts
+++ b/types/semver/index.d.ts
@@ -11,7 +11,7 @@ export type ReleaseType = "major" | "premajor" | "minor" | "preminor" | "patch" 
 
 export interface Options {
     loose?: boolean;
-    includePrerelease?: boolean
+    includePrerelease?: boolean;
 }
 
 /**
@@ -45,7 +45,7 @@ export function patch(v: string | SemVer, optionsOrLoose?: boolean | Options): n
 /**
  * Returns an array of prerelease components, or null if none exist.
  */
-export function prerelease(v: string | SemVer, optionsOrLoose?: boolean | Options): string[] | null;
+export function prerelease(v: string | SemVer, optionsOrLoose?: boolean | Options): ReadonlyArray<string> | null;
 
 // Comparison
 /**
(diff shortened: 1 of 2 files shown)


BendingBender (changes requested):
Please update the version number to 6.0

ajafff:
@BendingBender I updated the version to v5.7. `minVersion` was introduced in that release.
v6.0 didn't actually contain any API changes so there's no need to bump the major version here

BendingBender (changes requested):
Why not 6.0?

ajafff:
Typically when introducing a new major version the old version is copied into a separate folder so it can still receive bug fixes and enhancements.
When making changes to types of the latest major version they are usually backported to all major version they apply to.
Since 5.7 and 6.0 are API-wise identical I don't see a reason for this overhead.

In addition projects that depend on `@types/semver@^5.x.x` automatically benefit from this change whereas a new major version requires them to update their dependencies.

BendingBender:
This rule usually only applies to projects that develop both major versions in parallel for a while. It doesn't make sense to do this for every project, especially where you don't expect any changes for the old major, like in this case.

If you want to have definitions for the older major, you just change the version specifier in your `package.json`, after all. The old version doesn't disappear from npm after a new major version is released.

ajafff:
Alright. I changed the version to v6.0 as suggested

BendingBender (approved):

BendingBender:
Thank you!

PR #34297, opened 2019-03-28
Cleanup 20190328

sandersn:
The biggest change is a fix to protractor-browser-logs dependency:

Its dependency on protractor was "latest" &mdash; 6.0 &mdash; which doesn't work because protractor has a dependency on selenium-webdriver. In the Definitely Typed repo, selenium-webdriver is treated as a dependency on @types/selenium-webdriver because d.ts overrides .js.  But selenium-webdriver is at 3.0 while @types/selenium-webdriver is at 4.0, meaning that protractor's imports are broken for people (like DT) who have @types/selenium-webdriver installed.

There are three ways to fix this:

1. Update @types/selenium-webdriver to 4.0. This is a big task.
2. Fix protractor-browser-log's protractor dependency to 5.4.2, right before protractor updated to the latest version of selenium-webdriver.  This way the installed version of selenium-webdriver matches the version of @types/selenium-webdriver in DT.
3. ???? Other tsconfig.json/package.jon wizardry? I don't think this is
a worthwhile path to explore.

### DIFF
--- a/types/glob-parent/index.d.ts
+++ b/types/glob-parent/index.d.ts
@@ -1,5 +1,5 @@
 // Type definitions for glob-parent 3.1
-// Project: https://github.com/es128/glob-parent
+// Project: https://github.com/gulpjs/glob-parent
 // Definitions by: mrmlnc <https://github.com/mrmlnc>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
--- a/types/is-glob/index.d.ts
+++ b/types/is-glob/index.d.ts
@@ -1,5 +1,5 @@
 // Type definitions for is-glob 4.0
-// Project: https://github.com/jonschlinkert/is-glob
+// Project: https://github.com/micromatch/is-glob
 // Definitions by: mrmlnc <https://github.com/mrmlnc>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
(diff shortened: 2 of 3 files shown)


rokadias (approved):
I wish there was a version change in protractor browser logs that went with the protractor versioning. Otherwise I would think there could be some conflicts of someone who is using 6.0 protractor and doesn't include @types/selenium-webdriver since the reliance on @types/selenium-webdriver would no longer matter. It would be better to just have a particular version of protractor-browser-logs to be able to pin to when you're using @types/selenium-webdriver 3.0.

Similar to branching a version off for v2 of selenium-webdriver: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/selenium-webdriver

However I don't see any versioning here to help support which version should go with webdriver 3.0, and there's been inactivity for years so unlikely that it may cause conflicts in the future. Which is why I'm approving the PR. https://github.com/wix/protractor-browser-logs

sandersn:
@rokadias Yeah this package only gets 142 downloads / week so I don't think many people will be affected. That's why I didn't try to upgrade selenium-webdriver's types to 4.0

cnishina:
Root discussion: https://github.com/angular/protractor/issues/5192

I'll have a DefinitelyTyped PR for selenium-webdriver 4 sometime tomorrow.

PR #34299, opened 2019-03-28
Adds types for xmldom-reborn

aneeven:
Please fill in this template.

If adding a new definition:


types/xmldom-reborn/index.d.ts
```diff
@@ -0,0 +1,50 @@
+// Type definitions for xmldom-reborn 1.0.5
+// Project: https://github.com/MrDatastorage/xmldom-reborn
+// Definitions by: Qubo <https://github.com/tkqubo>
+//                 Andreas Neeven <https://github.com/aneeven>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+
+declare namespace xmldom {
```
rbuckton: Is there any particular reason you are defining this in a namespace and using `export =`? You could just remove the namespace and export everything at the top level.

types/xmldom-reborn/index.d.ts
```diff
@@ -0,0 +1,50 @@
+// Type definitions for xmldom-reborn 1.0.5
```
rbuckton: For a new definition you should only include the MAJOR.MINOR part of the package version in the header so that the types-publisher can properly version the types package.

rbuckton (changes requested):

types/xmldom-reborn/index.d.ts
aneeven: No particular reason, but I hope it's fine now. We're in a bit of a hurry over here, we need those types to swap an old package due to licensing issues.

types/xmldom-reborn/tslint.json
```diff
@@ -0,0 +1,80 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "adjacent-overload-signatures": false,
```
DanielRosenwasser: Can you please re-enable all these rules and ensure that the package contents conform?

aneeven:
Any updates?

types/xmldom-reborn/index.d.ts
```diff
@@ -0,0 +1,56 @@
...
+
+declare const DOMParser: DOMParserStatic;
+declare const XMLSerializer: XMLSerializerStatic;
+declare const DOMImplementation: DOMImplementationStatic;
+
+interface DOMImplementationStatic {
+    new(): DOMImplementation;
+}
+
+interface DOMParser {
```
sheetalkamat: From the example usage here it seems like this needs to be class instead of interface.
https://github.com/MrDatastorage/xmldom-reborn#example

sheetalkamat: What I meant is why not declare this as class intead of mixing  interface, class and statics as another interface

sheetalkamat (changes requested):

PR #34322, opened 2019-03-29
Add string type to valid data for PubNub decrypt

FlorianDr:
**Question:**
I actually need some input here. The flow types in the repo are using `object` and allow on options parameter. The API docs use `string` as first parameter and don't expose options object. What should I do? 
Here are links to both: 
* https://www.pubnub.com/docs/nodejs-javascript/api-reference-misc#decrypt
* https://github.com/pubnub/javascript/blob/master/src/core/components/cryptography/index.js#L108

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.pubnub.com/docs/nodejs-javascript/api-reference-misc#decrypt

### DIFF
--- a/types/pubnub/index.d.ts
+++ b/types/pubnub/index.d.ts
@@ -116,7 +116,7 @@ declare class Pubnub {
   ): any;
 
   decrypt(
-    data: object,
+    data: string | object,
     customCipherKey?: string,
     options?: Pubnub.CryptoParameters
   ): any;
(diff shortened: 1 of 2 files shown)


FlorianDr:
FYI please check question above.

ChristianBoehlke (approved):
It seems like the method can only accept a base64-encoded string (if no `customDecrypt` option is specified). So this change looks good to me.

rbuckton:
Thanks for the contribution!

PR #34325, opened 2019-03-29
Update for Microsoft/TypeScript#30568

sandersn:
This PR changes some types in TS 3.5.


sandersn:
1. I forgot that 3.4 and below will still have the old types. I'll convert the ExpectType into a type that both types can be assigned to.
2. `npm run test` is running out of memory, I think when trying to test both lodash and react at the same time. Does this mean anyone with react and lodash together are going to fail to build?

ferdaber (approved):

sandersn:
Build was clean locally (after adding `--max_old_space_size=16384`) so I'm going to merge since this is changes only tests.

Issue #33690, opened 2019-03-07
[@types/react-redux] 'hoist-non-react-statics' has no exported member 'NonReactStatics'


stewartrule:
- [x] I tried using the `@types/react-redux` package and had problems.
- [x] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.

@jamesreggio @JounQin

updating from `@types/react-redux 7.0.1` to `@types/react-redux 7.0.2` gives the following error:

```
'/node_modules/hoist-non-react-statics' has no exported member 'NonReactStatics'.

47 import { NonReactStatics } from 'hoist-non-react-statics';
```
seems like it was introduced here: https://github.com/DefinitelyTyped/DefinitelyTyped/commit/8b1beff944f6c7bf913b6fcee31fb5f7129064a7

jamesreggio:
Yikes. I introduced a dependency upon `@types/hoist-non-react-statics` in that change, but I didn't add it as a dependency. This problem is, I'm not sure where to declare it as a dependency, since the types only depend upon the types.

@JounQin, can you help me to understand how to fix this. Do we need to add an `///<reference` or add something to the `package.json`?

As a temporary workaround, you can `npm install --dev @types/hoist-non-react-statics` to your project.

darrencrossley:
I could be wrong but I think the problem is maybe more simple,

    import { NonReactStatics } from 'hoist-non-react-statics';

should be

    import NonReactStatics from 'hoist-non-react-statics';

downgrading to @types/react-redux 7.0.1 is a quick fix until this is fixed.

JustFly1984:
I got this issue today too. Downgrading to 7.0.1 helped

bordi:
Same here.

surgeboris:
> Yikes. I introduced a dependency upon `@types/hoist-non-react-statics` in that change, but I didn't add it as a dependency

DefinitelyTyped automatically added ` @types/hoist-non-react-statics` as a dependency to `@types/react-redux`, but (apparently) that wasn't enough for your typings to work.

> As a temporary workaround, you can `npm install --dev @types/hoist-non-react-statics` to your project.

[...]

PR #34329, opened 2019-03-29
react-redux dependencies: Semver is a lie and the caret is a bad idea :heart:

weswigham:
Over in https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33690 people are still having issues with `npm` not installing the required versions of `react-redux`'s dependencies into existing projects - at this point, I believe it's because the caret isn't a strong enough constraint to force a version greater than or equal to `3.3` when an older version in the same major range is installed. In any case, an explicit `>=` requirement should do nicely.

### DIFF
--- a/types/react-redux/package.json
+++ b/types/react-redux/package.json
@@ -1,7 +1,7 @@
 {
     "private": true,
     "dependencies": {
-        "@types/hoist-non-react-statics": "^3.3.0",
+        "@types/hoist-non-react-statics": ">=3.3.0",
         "redux": "^4.0.0"
     }
 }


weswigham:
cc @sandersn who also has some background here

sandersn:
I don't think that's how caret works, though:

```
✔ test  [master L|✚ 11…4]$ npm install semver
+ semver@6.0.0
✔ test  [master L|✚ 11…4]$ node
> var s = require('semver')
> s.satisfies('3.2.3', '^3.3.0')
false
> s.satisfies('3.2.3', '>=3.3.0')
false 
```

Of course npm may treat the output of semver specially, or may not even *use* the semver package despite [referring to it in the documentation.](https://docs.npmjs.com/misc/semver)

weswigham:
It could be that the one user in the thread was installing an older version of `react-redux` itself? IDK. I guess we'll just sit on it, then.

PR #34331, opened 2019-03-29
Fix Slick.EventHandler

matthias-ccri:
It looks like the existing types for `Slick.EventHandler` were incorrect. They don't match the code: https://github.com/mleibman/SlickGrid/blob/bf4705a96c40fea088216034def4d0256a335e65/slick.core.js#L133

Added tests.
Also enabled strictNullChecks and resolved the resulting errors.

### DIFF
--- a/types/slickgrid/index.d.ts
+++ b/types/slickgrid/index.d.ts
@@ -93,16 +93,16 @@ declare namespace Slick {
 		* @method subscribe
 		* @param fn {Function} Event handler.
 		*/
-		public subscribe(fn: (e: EventData, data: T) => any): void;
-		public subscribe(fn: (e: DOMEvent, data: T) => any): void;
+		public subscribe(fn: (e: EventData, data: T) => void): void;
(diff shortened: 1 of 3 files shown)


types/slickgrid/index.d.ts
```diff
@@ -93,16 +93,16 @@ declare namespace Slick {
 		* @method subscribe
 		* @param fn {Function} Event handler.
 		*/
-		public subscribe(fn: (e: EventData, data: T) => any): void;
-		public subscribe(fn: (e: DOMEvent, data: T) => any): void;
+		public subscribe(fn: (e: EventData, data: T) => void): void;
+		public subscribe(fn: (e: DOMEvent, data: T) => void): void;
```
matthiasdailey-ccri: Typing a callback's return value as `void` is the best practice: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#use-optional-parameters

PR #34335, opened 2019-03-29
react-redux currently relies on a TS bug (Microsoft/TypeScript#30634) to sucessfully typecheck

weswigham:
This fixes that. The constraint on `Shared`'s second type parameter's constraint seems to be completely unneeded, so I just.... removed it.


weswigham:
Since CI is green, I'm just going to merge this, as this shouldn't break anyone AFAIK - the constraint actually meant approximately nothing in current TS (maybe any TS?) since it was circular, so it was _effectively_ unconstrained already.

PR #34336, opened 2019-03-29
Change some libaries to rely less on implicit {} in test output

weswigham:
In anticipation of https://github.com/Microsoft/TypeScript/pull/30637, I've changed the few tests that expressly capture the inference-failure type to not do so, and, in some cases (specifically in `postman-collection`), changed the type parameter to explicitly extend `{}` because the non-null/non-undefined-ness under `strict` was both preferable and desired.

### DIFF
--- a/types/ember/test/ember-module-tests.ts
+++ b/types/ember/test/ember-module-tests.ts
@@ -2,8 +2,12 @@ import Ember from 'ember';
 
 // $
 Ember.$; // $ExpectType JQueryStatic
+
+const top = (<T>(x?: T): T => x!)();
+type Top = typeof top;
+declare function expectTypeNativeArrayTop(x: Ember.NativeArray<Top>): void;
 // A
-Ember.A(); // $ExpectType NativeArray<{}>
+expectTypeNativeArrayTop(Ember.A());
 Ember.A([1, 2]); // $ExpectType NativeArray<number>
 // addListener
 Ember.addListener({ a: 'foo' }, 'a', {}, () => {});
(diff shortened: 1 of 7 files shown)


weswigham:
Hope yall are OK with these small changes to ensure we don't break DT when we publish a nightly with the change in the OP. I'm going to merge this, so lmk if anything about it irks you, and I'll do my best to refactor with the change in mind or explain why I'm testing for the inference failure type this way~ ❤️

types/ember/test/ember-module-tests.ts
```diff
@@ -2,8 +2,12 @@ import Ember from 'ember';
 
 // $
 Ember.$; // $ExpectType JQueryStatic
+
+const top = (<T>(x?: T): T => x!)();
+type Top = typeof top;
+declare function expectTypeNativeArrayTop(x: Ember.NativeArray<Top>): void;
 // A
-Ember.A(); // $ExpectType NativeArray<{}>
+expectTypeNativeArrayTop(Ember.A());
```
mike-north: @weswigham - Isn't this a less-strict test than what it replaces? i.e., if a regression were introduced where `Ember.A()`'s return type were inappropriately widened to `any`, the old test would have caught it and the new one wouldn't.

mike-north: Happy to fix this myself in a follow-on PR btw. I just don't know if loosening the test was part of the adjustment necessary to work well with Microsoft/TypeScript#30637

weswigham: The loosening isn't required for anything - I was just trying to keep the updated tests simple, esp. since I believe the only way to do the exact test is via conditional types, which not every package in the PR can use (due to being targeted at older TS versions).

PR #34340, opened 2019-03-30
Add annotations to ramda tests which contain type inference failures …

weswigham:
…which trigger future `unknown` warnings, in preparation for https://github.com/Microsoft/TypeScript/pull/30637. For `unknown`, TS warns on operations like `>`, which TS does not currently for `{}` (for some reason, we probably should). The code _probably_ should have always had an annotation, since there's nowhere other than the annotation to infer the type from.

### DIFF
--- a/types/ramda/ramda-tests.ts
+++ b/types/ramda/ramda-tests.ts
@@ -1975,9 +1975,9 @@ class Rectangle {
 };
 
 () => {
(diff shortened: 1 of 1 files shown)


googol (approved):
Great, this seems like a good thing!

I've had to use tslint before to ensure I don't get those inferred empty objects in my ramda functions, having a built-in check for this is great!

PR #34347, opened 2019-03-30
[@types/ramda] fix returning value for split to match the expected value for fromEntries func

ArtemKorchunov:
Problem:
Type 'string[]' in returning value of `split('=')` is missing the following properties from type 'KeyValuePair<number, {}>': 0, 1 that is an argument for fromPairs function

When trying to
```
const queryString = "?item=2&page=10&total=90";

compose(
  fromPairs,
  map(split("=")),
  split("&"),
  tail
);
```

- [ ] Add or edit tests to reflect the change. (Run with npm test.)

- [x] Follow the advice from the readme.

- [x] Avoid common mistakes.

### DIFF
--- a/types/ramda/index.d.ts
+++ b/types/ramda/index.d.ts
@@ -2712,8 +2712,8 @@ declare namespace R {
          * Splits a string into an array of strings based on the given
          * separator.
          */
-        split<T>(sep: string | RegExp): (str: string) => T;
-        split<T>(sep: string | RegExp, str: string): T;
+        split<T = string[]>(sep: string | RegExp): (str: string) => T;
+        split<T = string[]>(sep: string | RegExp, str: string): T;
 
         /**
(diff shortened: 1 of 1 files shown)


43081j:
This doesn't seem quite right. ramda's split function always returns a string array. it might make more sense to change the `fromPairs` function to accept arrays or have type parameters.

googol:
There isn't anything to be done on the `fromPairs` side either, since it requires a tuple of exactly two elements and those elements are already type parameters on it.

The thing to be done for this problem is wrapping `split` in your custom code in a way that verifies that it returns two elements in the array, and typing the result accordingly.

DanielRosenwasser:
Unfortunately I don't think this is the right fix - you really don't know how many elements any given call to `split` is going to return in the first place. In your example, it'll work on `queryString`, but it's not guaranteed to work on any arbitrary input.

PR #34356, opened 2019-03-31
[storybook-readme] Migrate to v5

lonyele:
Please fill in this template.

Hi, maintainer of storybook-readme versioned up to v5 as storybook released v5. This storybook-readme v5 has a new code with a backward compatibility with previous versions. So I added a new types for v5 and left all previous types with a comment. I happened to add and update this type definitions, I still am not super confident myself, so always welcome any help. Here is the link for that original issue https://github.com/tuchk4/storybook-readme/issues/124


types/storybook-readme/index.d.ts
```diff
@@ -16,10 +16,12 @@ export type DecoratorPattern = (
 
 export type HOCPattern = (story: RenderFunction) => Renderable | null;
 
+export type MakeDecoratorResult = (...args: any) => any;
+
 // Types added for v5
-export type addReadme = (decorator: DecoratorPattern) => React.ReactNode;
-export type addFooter = (md: string) => void;
-export type addHeader = (md: string) => void;
+export const addReadme: MakeDecoratorResult
```
lonyele: I'm not really sure If the type should be like this.  On storybook-readme code,  `const addReadme = makeDecorator({ ... }) `  So It is the return value of `makeDecorater` and It is from https://github.com/storybooks/storybook/blob/next/lib/addons/src/make-decorator.ts#L38   with the type of return value is `type MakeDecoratorResult = (...args: any) => any;`

uniqueiniquity: That seems like the best you can do for now. I would suggest asking storybook to export `MakeDecoratorResult`, so you could import that here and stay in sync with future changes.

PR #34382, opened 2019-04-01
fix(node-fetch): HeadersInit type

JasonLi914:
Please fill in this template.

https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L18260
https://github.com/bitinn/node-fetch/blob/master/src/headers.js#L96 https://developer.mozilla.org/en-US/docs/Web/API/Headers/Headers

### DIFF
--- a/types/node-fetch/index.d.ts
+++ b/types/node-fetch/index.d.ts
@@ -1,10 +1,11 @@
-// Type definitions for node-fetch 2.1
+// Type definitions for node-fetch 2.3
 // Project: https://github.com/bitinn/node-fetch
 // Definitions by: Torsten Werner <https://github.com/torstenwerner>
 //                 Niklas Lindgren <https://github.com/nikcorg>
 //                 Vinay Bedre <https://github.com/vinaybedre>
 //                 Antonio Román <https://github.com/kyranet>
 //                 Andrew Leedham <https://github.com/AndrewLeedham>
+//                 Jason Li <https://github.com/JasonLi914>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 /// <reference types="node" />
@@ -37,7 +38,7 @@ export class Request extends Body {
 export interface RequestInit {
     // whatwg/fetch standard options
     body?: BodyInit;
-    headers?: HeaderInit | { [index: string]: string };
+    headers?: HeadersInit;
     method?: string;
     redirect?: RequestRedirect;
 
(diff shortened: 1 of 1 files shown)


types/node-fetch/index.d.ts
```diff
@@ -165,12 +166,12 @@ export type ResponseType =
...
 
 export interface ResponseInit {
-    headers?: HeaderInit;
+    headers?: HeadersInit;
     status: number;
     statusText?: string;
 }
 
-export type HeaderInit = Headers | string[];
+export type HeadersInit = Headers | {[key: string]: string};
```
nikcorg:
```suggestion
export type HeadersInit = Headers | string[][] | { [key: string]: string };
```

nikcorg (changes requested):
Nice change. It caught a bug about the previous `HeaderInit` having an erronoeus type in the `HeaderInit` union previously. It shouldn't be omitted, though, as [it is supported by node-fetch](https://github.com/bitinn/node-fetch/blob/master/src/headers.js#L78-L86).

types/node-fetch/index.d.ts
JasonLi914: Ah yeah, I missed that code block in node-fetch initially which is why I removed it. Added it back  as per your suggestions, thanks for the catch! Also, I made this a breaking change, but not sure if it is such since it is a fix?

nikcorg: @JasonLi914 The version number in these types should reflect which version of node-fetch they represent. The DefinitelyTyped release team handle any other versioning.

types/node-fetch/index.d.ts
```diff
@@ -1,10 +1,11 @@
-// Type definitions for node-fetch 2.1
+// Type definitions for node-fetch 3.0
```
nikcorg:
```suggestion
// Type definitions for node-fetch 2.3
```

nikcorg (changes requested):

types/node-fetch/index.d.ts
nikcorg: Thanks for mentioning the version earlier, I totally omitted to check this change.

JasonLi914:
@nikcorg - I made the changes you suggested. Let me know if there is anything else, thanks!

nikcorg (approved):
LGTM

thumbs-up-green

jwalton:
Umm... Is there a reason we renamed `HeaderInit` to `HeadersInit`, breaking all packages which depend on this?  Could we maybe put an `export type HeaderInit = HeadersInit;` so we don't break backwards compatibility?

JasonLi914:
@jwalton - The reason I named it as such is because that's how it is represented in `lib DOM` and so wanted to make it consistent. But I do agree with adding `HeaderInit` to not break existing consumers. I can spin up another PR in a sec to include this change.

jwalton:
You are a scholar and and a gentleman, sir.  :)

PR #34394, opened 2019-04-02
@types/xml2js change the type of the Error object

redlickigrzegorz:
Please fill in this template.

I didn't create an issue for this problem but in this pull request, I changed the type for the `err` parameter which was treated as `any` but it should be always the `Error` object instead.

I found this problem during using the [util.promisify](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original). I got the error about missing third argument: `TS2554: Expected 3 arguments, but got 2.`

The types provided for `util` module treated this method as the method with 3 arguments instead of 2 arguments and the callback.

With this change, everything is working correctly.


claasahl (approved):
LGTM

thanks for your contribution, @redlickigrzegorz

PR #34400, opened 2019-04-02
Ember: fix type for `Route#controllerFor`.

chriskrycho:
Previously, `Route#controllerFor` was defined as taking a name from the controller registry, but it should accept the name of a route matching either the current route or one of its parents in the route hierarchy. We cannot (currently) constrain it to match the route hierarchy, so simply using `string` is the best option we have and users can then narrow it to a specific `Controller` subclass.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://api.emberjs.com/ember/3.8/classes/Route/methods/controllerFor?anchor=controllerFor>
- [ ] ~Increase the version number in the header if appropriate.~

### DIFF
--- a/types/ember__routing/test/route.ts
+++ b/types/ember__routing/test/route.ts
@@ -3,7 +3,6 @@ import Array from '@ember/array';
 import EmberObject from '@ember/object';
 import Controller from '@ember/controller';
 import Transition from '@ember/routing/-private/transition';
-import { assertType } from './lib/assert';
 
 class Post extends EmberObject {}
 
--- a/types/ember__routing/test/route.ts
+++ b/types/ember__routing/test/route.ts
@@ -3,6 +3,7 @@ import Array from '@ember/array';
 import EmberObject from '@ember/object';
 import Controller from '@ember/controller';
 import Transition from '@ember/routing/-private/transition';
+import { assertType } from './lib/assert';
 
 class Post extends EmberObject {}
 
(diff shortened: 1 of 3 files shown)


mike-north (changes requested):
in addition to addressing the build failures (should be detectable locally w/ `yarn lint ember__routing`), please add negative and positive test(s)

chriskrycho:
Gah, stupid lint issue. I though I'd caught everything. And yeah, I'll add tests as well.

chriskrycho:
@mike-north, any thoughts on further tests to write here? I'm happy to add one or more further, but none occurred to me. (That said: it has been a loooong day in a variety of ways, so I would not put missing an obvious case past myself.)

types/ember__routing/test/route.ts
```diff
@@ -3,6 +3,7 @@ import Array from '@ember/array';
 import EmberObject from '@ember/object';
 import Controller from '@ember/controller';
 import Transition from '@ember/routing/-private/transition';
+import { assertType } from './lib/assert';
```
mike-north: I don't think this import is needed

mike-north (changes requested):
What I’d look for in a test is something that fails before your change and passes after it.

Example: using `controllerFor` to retrieve a controller NOT present in the controller registry.

types/ember__routing/test/route.ts
chriskrycho: Fixed in the commit I'm about to push. Thanks!

chriskrycho:
@mike-north the reason I was struggling to come up with *additional* tests, I realize, is because I already have that: [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34400/files#diff-af94b119f2eef7558364c47b0943b198R115). We should be good to merge.

mike-north (approved):

PR #34402, opened 2019-04-02
xrm: Set removePreSearch handler to match addPreSearch

derekfinlinson:
Found I was unable to use removePreSearch since the expected handler type didn't match addPreSearch.


daryllabar (approved):
Makes sense!

DanielRosenwasser:

derekfinlinson:
@DanielRosenwasser thanks for the heads up. I updated my config.

PR #34405, opened 2019-04-02
React-Big-Calendar: improve typing for Views, specify type for Event,…

aaltepet:
1. Views type should have view names be optional
2. Event properties are known, and the Event type now defines them
3. The BigCalendarProps no longer needs to `extend React.Props`

   1. Docs shows `views` prop has not containing all view types: https://intljusticemission.github.io/react-big-calendar/examples/index.html#prop-views
   2. Event type is defined in the docs here: https://intljusticemission.github.io/react-big-calendar/examples/index.html#prop-events
   3. React type definition stating `interface Props` is deprecated: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1185

### DIFF
--- a/types/react-big-calendar/index.d.ts
+++ b/types/react-big-calendar/index.d.ts
@@ -20,15 +20,21 @@ export type stringOrDate = string | Date;
 export type ViewKey = 'MONTH' | 'WEEK' | 'WORK_WEEK' | 'DAY' | 'AGENDA';
 export type View = 'month' | 'week' | 'work_week' | 'day' | 'agenda';
 export type Views = View[] | {
-    day: boolean | React.SFC | React.Component,
-    agenda: boolean | React.SFC | React.Component,
-    month: boolean | React.SFC | React.Component,
-    week: boolean | React.SFC | React.Component,
-    work_week: boolean | React.SFC | React.Component
+    work_week?: boolean | React.SFC | React.Component,
+    day?: boolean | React.SFC | React.Component,
+    agenda?: boolean | React.SFC | React.Component,
+    month?: boolean | React.SFC | React.Component,
+    week?: boolean | React.SFC | React.Component
 };
 export type Navigate = 'PREV' | 'NEXT' | 'TODAY' | 'DATE';
 
-export type Event = object;
+export interface Event {
+    allDay?: boolean;
+    title: string;
+    start: Date;
+    end: Date;
+    resource?: any;
+}
 export interface DateRange {
     start: Date;
     end: Date;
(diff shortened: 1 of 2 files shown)


types/react-big-calendar/react-big-calendar-tests.tsx
```diff
@@ -85,6 +85,26 @@ class CalendarResource {
...
 
+// overriding 'views' props
+{
+    const DaySFC: React.SFC = () => null;
+    class MonthView extends React.Component {
+        render() {
+            return null;
+        }
+    }
+    // supplying object to 'views' prop with only some of the supported views
```
eps1lon: But then your defining a value for every view. It's not actually testing optionality of `views`

types/react-big-calendar/index.d.ts
```diff
@@ -230,8 +236,8 @@ export class DateLocalizer {
     format(value: FormatInput, format: string, culture: Culture): string;
 }
 
-export interface BigCalendarProps<TEvent extends Event = Event, TResource extends object = object> extends React.Props<BigCalendar<TEvent, TResource>> {
```
eps1lon: Permalink: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6f7a141fb85d645316c66eb228f158a3f24513c9/types/react/index.d.ts#L1228

eps1lon (changes requested):

types/react-big-calendar/react-big-calendar-tests.tsx
aaltepet: @eps1lon - This case is testing that some `views` properties are optional, i.e. not all of them need to be supplied.  I've updated the comment to say so.

I also added a new test case demonstrating the optionality of `views`

aaltepet:
I've updated the PR with the requested changes.  Thanks!

types/react-big-calendar/react-big-calendar-tests.tsx
eps1lon: Sure but that wasn't the case at the time I made the comment. Now it's doing what the comment it says it does.

eps1lon (approved):

types/react-big-calendar/index.d.ts
```diff
@@ -230,8 +236,8 @@ export class DateLocalizer {
...
-    localizer: DateLocalizer;
+export interface BigCalendarProps<TEvent extends Event = Event, TResource extends object = object> {
+    localizer?: DateLocalizer;
```
TomasHubelbauer: Localizer doesn't seem to be optional:
https://github.com/intljusticemission/react-big-calendar/blob/master/src/Calendar.js#L57

TomasHubelbauer (changes requested):

types/react-big-calendar/index.d.ts
siavelis: You're right! 
In the same file the `localizer` is passed in `mergeWithDefaults()` [here].(https://github.com/intljusticemission/react-big-calendar/blob/master/src/Calendar.js#L785) This function does not seem to handle undefined instances (check [here](https://github.com/intljusticemission/react-big-calendar/blob/master/src/localizer.js#L46)

PR #34408, opened 2019-04-02
Add markdown-to-jsx typings

ecraig12345:
Add typings for [markdown-to-jsx](https://www.npmjs.com/package/markdown-to-jsx).

If adding a new definition:


types/markdown-to-jsx/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "no-null-undefined-union": false
```
uniqueiniquity: This rule shouldn't be causing any errors, since it shouldn't be enabled for `dtslint`. Do you get any errors when running `dtslint` by invoking `npm run lint markdown-to-jsx` from the root of the `DefinitelyTyped` repo?

uniqueiniquity (changes requested):

types/markdown-to-jsx/tslint.json
ecraig12345: React.ReactNode and multiple other types from React include both null and undefined. I could disable the rule on individual lines instead if that's better? (the error occurs in five places currently)

uniqueiniquity (approved):

types/markdown-to-jsx/tslint.json
uniqueiniquity: Works for me. I think you might be running with a different version of TSLint locally than what we expect, and as a result you're getting errors from that rule. Doesn't hurt to disable the rule, though, so this will work fine.

sunknudsen:
@ecraig12345 Thanks for adding the type definitions of [markdown-to-jsx](https://www.npmjs.com/package/markdown-to-jsx)

Experiencing issues using these type definitions with more elaborate overrides.

According to you, would replacing

```typescript
export type ComponentOverride = string | React.ComponentClass | React.SFC | {
    component: string | React.ComponentClass | React.SFC;
    props?: any;
};
```

by

```typescript
export type ComponentOverride =
    | string
    | React.ComponentClass<any, any>
    | React.SFC<any>
    | {
          component: string | React.ComponentClass<any, any> | React.SFC<any>;
          props?: any;
      };
```

be OK? I am working on a PR.

Also, do you know why I am getting the following error when running `npm run lint markdown-to-jsx`?

```shell
../react/index.d.ts(34,22): error TS2307: Cannot find module 'csstype'.
```

uniqueiniquity:
@sunknudsen For the latter issue, I recommend running `npm install` in the `types/react` folder - it relies on a dependency that's outside of DefinitelyTyped.

sunknudsen:
@uniqueiniquity That worked thanks!

PR #34416, opened 2019-04-02
[next-server] Change children type to ReactNode

osdiab:
Pretty sure this should be `ReactNode`, not `ReactElement`, as it's defined on the standard anchor tag itself.

Based on how ReactRouter just inherits from `React.AnchorHTMLAttributes`:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-router-dom/index.d.ts#L47

Which ultimately inherits from `React.DOMAttributes`:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1213

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: see above
- [N/A] Increase the version number in the header if appropriate.
- [N/A] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

### DIFF
--- a/types/next-server/link.d.ts
+++ b/types/next-server/link.d.ts
@@ -11,7 +11,7 @@ export interface LinkProps {
     href?: string | UrlLike;
     as?: string | UrlLike;
     passHref?: boolean;
-    children: React.ReactElement;
+    children: React.ReactNode;
 }
 
 export default class Link extends React.Component<LinkProps> {}


osdiab:
checking the source code to see how Link is defined - if it passes through props for anchor tags to the underlying link component it'd be better to just inherit directly, since there's other attributes that would be nice to have access to.

osdiab:
hmmm, doesn't seem to based on my eyeballing the client side link impl, though i didn't look that close:

https://github.com/zeit/next.js/blob/canary/packages/next/client/link.js

osdiab:
Oh nm i didn't understand how link works. Closing

PR #34424, opened 2019-04-03
poi: add poi types

bolasblack:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/poi/index.d.ts
+++ b/types/poi/index.d.ts
@@ -1,0 +1,223 @@
+// Type definitions for poi 12.5
+// Project: https://github.com/egoist/poi/
+// Definitions by: c4605 <https://github.com/bolasblack>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import { ICompiler, Configuration as WebpackConfig } from "webpack";
+import { Configuration as WebpackDevServerConfig } from "webpack-dev-server";
+import { Options as HtmlWebpackPluginOptions } from "html-webpack-plugin";
+import * as WebpackChainConfig from "webpack-chain";
+import CAC from "cac/types/CAC";
+
+export type Mode = "production" | "development" | "test";
+
+export interface Opts {
+    type: string;
+    mode: Mode;
+    [k: string]: any;
+}
+
+/**
+ * https://poi.js.org/config.html
+ */
+export interface Config {
+    entry?: Config.Entry;
+    output?: Config.Output;
+    pages?: Config.Pages;
+    babel?: Config.Babel;
+    css?: Config.Css;
+    assets?: Config.Assets;
+    envs?: Config.Envs;
+    constants?: Config.Constants;
+    chainWebpack?: Config.ChainWebpack;
+    configureWebpack?: Config.ConfigureWebpack;
+    publicFolder?: Config.PublicFolder;
+    devServer?: Config.DevServer;
+    plugins?: Config.Plugins;
+}
+
+export namespace Config {
+    type Entry = WebpackConfig["entry"];
+
+    interface Output {
+        dir?: string;
+        clean?: boolean;
+        format?: "iife" | "cjs" | "umd";
+        moduleName?: string;
+        sourceMap?: boolean;
+        minimize?: boolean;
+        publicUrl?: string;
+        fileNames?: Output.FileNames;
+        target?:
+            | "web"
+            | "electron-renderer"
+            | "electron-main"
+            | "node"
+            | "node-webkit"
+            | "async-node"
+            | "webworker";
+        html?: Output.Html;
+    }
+    namespace Output {
+        interface FileNames {
+            js?: string;
+            css?: string;
+            font?: string;
+            image?: string;
+        }
+
+        interface Html {
+            title?: string;
+            filename?: string;
+            template?: string;
+            inject?: boolean;
+        }
+    }
+
+    interface Pages {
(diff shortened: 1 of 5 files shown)


types/poi/index.d.ts
```diff
@@ -0,0 +1,223 @@
...
+        publicUrl?: string;
+        fileNames?: Output.FileNames;
+        target?:
+            | "web"
+            | "electron-renderer"
+            | "electron-main"
+            | "node"
+            | "node-webkit"
+            | "async-node"
+            | "webworker";
```
egoist: Poi also accepts `electron`: https://github.com/egoist/poi/blob/bba820b16c7b160fcaa5802598dde1971854ad52/core/poi/lib/utils/validateConfig.js#L18

types/poi/index.d.ts
```diff
@@ -0,0 +1,223 @@
...
+
+/**
+ * https://poi.js.org/api.html#constructor-argv
+ */
+export class Poi {
+    mode: Mode;
+
+    cli: CAC;
+
+    command: CAC["command"];
```
egoist:
```suggestion
   /** The current running command */
    command: CAC["command"];
```

types/poi/index.d.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    config: Config;
+
+    isProd: boolean;
+
+    configLoader: ConfigLoader;
+
+    constructor(argv?: string[]);
+
+    hasDependency(name: string): boolean;
```
egoist:
```suggestion
   /** Check if a package is included in the `dependencies` or `devDependencies` field your `package.json` 
    hasDependency(name: string): boolean;
```

egoist:
This is a more detailed list of all the options: https://github.com/egoist/poi/blob/master/core/poi/lib/utils/validateConfig.js

types/poi/index.d.ts
```diff
@@ -0,0 +1,223 @@
...
+
+    interface DevServer {
+        host?: string;
+        port?: string | number;
+        hot?: boolean;
+        hotOnly?: string;
+        hotEntries?: string[];
+        historyApiFallback?: WebpackDevServerConfig["historyApiFallback"];
+        open?: boolean;
+        proxy?: WebpackDevServerConfig["proxy"];
```
egoist: `proxy` can be a `string`

bolasblack:
> This is a more detailed list of all the options: [egoist/poi:core/poi/lib/utils/validateConfig.js@`master`](https://github.com/egoist/poi/blob/master/core/poi/lib/utils/validateConfig.js)

Thanks, I will update the Config definition after reading the code

types/poi/index.d.ts
bolasblack: Hi @egoist, I found proxy also support function, but I can not find the function signature both in [poi documention](https://poi.js.org/config.html#devserver-proxy) or [webpack dev server documention](https://webpack.js.org/configuration/dev-server/#devserverproxy)

Can you tell me where introduced the definition of this function?

bolasblack:
Finally...It's weird, when I execute `npm run lint poi`, I get error:

```
../uglify-js/index.d.ts(9,30): error TS7016: Could not find a declaration file for module 'source-map'. '/Users/c4605/workSpace/DefinitelyTyped/node_modules/source-map/source-map.js' implicitly has an 'any' type.
  Try `npm install @types/source-map` if it exists or add a new declaration (.d.ts) file containing `declare module 'source-map';`
../webpack/index.d.ts(30,30): error TS7016: Could not find a declaration file for module 'source-map'. '/Users/c4605/workSpace/DefinitelyTyped/node_modules/source-map/source-map.js' implicitly has an 'any' type.
  Try `npm install @types/source-map` if it exists or add a new declaration (.d.ts) file containing `declare module 'source-map';`

    at /Users/c4605/workSpace/DefinitelyTyped/node_modules/dtslint/bin/index.js:189:19
    at fulfilled (/Users/c4605/workSpace/DefinitelyTyped/node_modules/dtslint/bin/index.js:5:58)
```

uniqueiniquity (changes requested):
It appears that the `poi` package exports a class directly, rather than a module containing the class `Poi`.
Therefore, you should use the [constructable module template](http://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-class-d-ts.html) so that the library can be imported without runtime failures.

bolasblack:
Hi @uniqueiniquity, thanks for your reply, do you mean I'd better write all the `export` in a `namespace`, then export this namespace at the end?

uniqueiniquity:
@bolasblack Yes, that's correct.
My suggestion would be to do something like the following:
```ts
declare class PoiCore {
...
}

declare namespace PoiCore {
  ....
  export interface Babel {
    ...
  }
  ...
}

export = PoiCore;
```

types/poi/package.json
```diff
@@ -0,0 +1,7 @@
+{
+    "private": true,
+    "dependencies": {
+        "cac": "~6.4.2",
```
bolasblack: Hi @egoist, I found cac@6.5 used the default export of `'event'`, it will make tsc report an error: [`Module '"events"' has no default export.`](https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/519946409#L578)

bolasblack:
Hi @uniqueiniquity, I modified the code according to your suggestion.

uniqueiniquity (approved):

PR #34426, opened 2019-04-03
Run lint instead of test

mike-marcacci:
According to [this comment](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34277#issuecomment-479279885) I suspect we aren't supposed to actually run `npm test`, which makes sense given how difficult it is to get working locally.


mike-marcacci:
My assumption here was incorrect; closing.

PR #34435, opened 2019-04-03
Updating firefox-webext-browser to v67.0b7

jsmnbom:
*Had no tests before*
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://gist.github.com/jsmnbom/b53891f23e61b635ea811dc7a83db005


types/firefox-webext-browser/index.d.ts
```diff
@@ -697,6 +700,8 @@ declare namespace browser._manifest {
 
     type _MatchPattern = "<all_urls>";
 
+    type _PersistentBackgroundProperty = "undefined" | "undefined";
```
DanielRosenwasser: Uh, why is this duplicated?

jsmnbom: Not only was that duplicated, it was also totally wrong.. The generator really didn't like the new manifest it seems. Should be rolled back to a version that actually makes sense now.

PR #34454, opened 2019-04-03
Move global declarations into a namespace for 'amap-js-api'.

DanielRosenwasser:
These types pollute the global namespace, and conflict with the upcoming `Omit` type helper coming to TypeScript 3.5.

Related to https://github.com/Microsoft/TypeScript/pull/30552


breeze9527 (approved):

DanielRosenwasser:
Thanks for the review!

PR #34455, opened 2019-04-03
[electron-window-state] Electron should not be a dependency 

timfish:
Please fill in this template.

With electron as a full dependency of `@types/electron-window-state`, the latest version of electron is downloaded into the dependency tree, even if you're not using this version.

### DIFF
--- a/types/electron-window-state/package.json
+++ b/types/electron-window-state/package.json
@@ -1,6 +1,6 @@
 {
     "private": true,
-    "dependencies": {
+    "devDependencies": {
         "electron": "*"
     }
 }


timfish:
I'm not sure how to fix this because it looks like `devDependencies` are not allowed. 

However, electron definitely shouldn't be a full dependency of any `@types` package. Currently you can end up with multiple versions of electron in you dependency tree.

timfish:
Also @andy-ms as it looks like you've worked on this package!

rhysd:
You should add `electron` as dev dependency of your project, not here because whether depending on `electron` using npm is up to the project.

timfish:
Electron is already a dependency of `@types/electron-window-state`. 

https://www.npmjs.com/package/@types/electron-window-state?activeTab=dependencies

DanielRosenwasser:
Technically this **should** work the same as `@types/node` itself. It really feels like it should be a `devDependency` which the end-user needs installed. @sandersn, any clues?

rhysd:
I thought someone may not have `electron` dependency in `package.json` in favor of `electron-builder`. `electron-builder` [does not depend on `electron`](https://www.npmjs.com/package/electron-builder) because it downloads Electron binary for each platform by itself.

timfish:
If you don't depend on an Electron version, when packaging with electron-builder it downloads and uses the latest release. This is almost always not want you want as it will likely break when major versions are released.

The electron package uses electron-download to download the binaries in postinstall.

The Electron npm module is not shipped in node_modules of a packaged app so the recommendation from the electron maintainers is to add it to your project as a devDependency. Electron-builder will pick this up and stop packaging your app with the latest.

rhysd (approved):

timfish:
@DanielRosenwasser what was the resolution for this? 

Does the typings check/test tooling need to be changed to accept `devDependencies` with a whitelist?

PR #33437, opened 2019-02-27
aws-lambda Authorizer response context only allows booleans, numbers and strings

richardcornelissen:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<url here>>
- [x] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #34456, opened 2019-04-03
Revert #33437 "aws-lambda Authorizer response context only allows booleans, …

simonbuchan:
…numbers and strings"

This reverts #33437 / commit eed96d4372ed55c81eb7e7708c5f3ea02c579ddd, as requested by @penguinbroker and @kossmoboleat (at least)


simonbuchan:
@richardcornelissen, let me know if you would prefer to remain in Authors - I was unsure about that one.

PR #34459, opened 2019-04-04
Type Definitions for package project-name

ajilo297:
Please fill in this template.

If adding a new definition:


types/project-name/index.d.ts
```diff
@@ -0,0 +1,10 @@
+// Type definitions for project-name
+// Project: https://github.com/jonschlinkert/project-name
+// Definitions by: Ajil Oommen <https://github.com/ajilo297/>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/**
+ * Type Definition for project-name
+ * @param cwd The Base Current Working Directory of project 
+ */
+export default function (cwd: string): string;
```
DanielRosenwasser: This package was declared using `module.exports = ...` or some variant. It should be declared with an `export = ...` like one of the following:

### Variables

```ts
declare var xyz: SomeType;

export = xyz;
```

### Classes

```ts
class Foo {
  // ...
}

export = Foo
```

### Functions

```ts
function foo(): any;

export = foo;
```

```ts
import foo = require("foo");
```

DanielRosenwasser (changes requested):

types/project-name/index.d.ts
ajilo297: I have made the changes

types/project-name/index.d.ts
```diff
@@ -0,0 +1,11 @@
...
+// Project: https://github.com/jonschlinkert/project-name
+// Definitions by: Ajil Oommen <https://github.com/ajilo297/>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export = index;
+/**
+ * Type Definition for project-name
+ * @param cwd The Base Current Working Directory of project
+ */
+declare function index(cwd: string): string;
```
DanielRosenwasser: `cwd` should be optional as per the docs.

types/project-name/tslint.json
```diff
@@ -0,0 +1,80 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "adjacent-overload-signatures": false,
```
DanielRosenwasser: Please remove all these toggled lint rules.

DanielRosenwasser (changes requested):

types/project-name/tslint.json
ajilo297: Fixed.

DanielRosenwasser:
Thank you!

PR #34460, opened 2019-04-04
[hummus-recipe] added a namespace for types and interfaces, fixed a typo and a wrong type

she11sh0cked:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/chunyenHuang/hummusRecipe/blob/master/lib/index.js


types/hummus-recipe/hummus-recipe-tests.ts
```diff
@@ -1,4 +1,4 @@
-import { Recipe } from 'hummus-recipe';
+import * as Recipe from "hummus-recipe";
```
DanielRosenwasser: Since this is a callable or constructable entity, it isn't legal in ES2015 to import and call/construct this with a namespace import like `import * as foo from 'foo'`, so could you please convert this to either

```ts
import foo = require('foo')
```

or the following with `esModuleInterop` turned on

```ts
import foo from 'foo';
```

DanielRosenwasser (changes requested):

types/hummus-recipe/hummus-recipe-tests.ts
erikberressem: Thank you for pointing it out

DanielRosenwasser (approved):

erikberressem:
@DanielRosenwasser Do i still need to do something to get it merged?

DanielRosenwasser:
Thanks!

PR #34461, opened 2019-04-04
Update Redirect to use RouteComponentProps

lifeiscontent:
Update Redirect to use RouteComponentProps

Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/reach__router/index.d.ts
+++ b/types/reach__router/index.d.ts
@@ -77,7 +77,7 @@ export interface RedirectProps<TState> {
     replace?: boolean;
 }
 
-export class Redirect<TState> extends React.Component<RedirectProps<TState>> { }
+export class Redirect<TState> extends React.Component<RouteComponentProps<RedirectProps<TState>>> { }
 
 export interface MatchProps<TParams> {
     path: string;


kingdaro:
[I'm reading through the docs](https://reach.tech/router/api/Redirect), and I don't see where it allows using normal route component props. The only ones I see listed are those that the type definitions currently define. Could you point to where it does? Or was there a non-documented change?

itsMapleLeaf (changes requested):
[I'm reading through the docs](https://reach.tech/router/api/Redirect), and I don't see where it allows using normal route component props. The only ones I see listed are those that the type definitions currently define. Could you point to where it does? Or was there an undocumented change?

lifeiscontent:
@kingdaro any child of `Router` is considered a `Route` component, so now that I've type it out, my changes don't make sense. However another question now enters my head, Can children be typed?

itsMapleLeaf (approved):
Children can't be typed, so this does look like the only real solution to that. I'll approve this

lifeiscontent:
@kingdaro I don't think this should of been merged, Redirect components aren't always direct children of Router so now this change makes the incorrect assumption that it is the case.

I think although not ideal, it'd be better in user land to create another function like this:

```ts
const RedirectRoute: React.FC<RouteComponentProps & RedirectProps> (props) => <Redirect {...props} />
```

PR #34476, opened 2019-04-04
Added the optional parameter to the save function.

FrankBrullo:
Please fill in this template.

If adding a new definition:

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://rawgit.com/MrRio/jsPDF/master/docs/jspdf.js.html#line4688>>

If removing a declaration:

### DIFF
--- a/types/jspdf/index.d.ts
+++ b/types/jspdf/index.d.ts
@@ -186,7 +186,7 @@ declare module 'jspdf' {
             stroke():void;
             fill():void;
             translate(x:number, y:number):void;
-            measureText(text:string):number;
+            measureText(text:string):{ width: number };
         };
 
         // jsPDF plugin: fromHTML
--- a/types/jspdf/index.d.ts
+++ b/types/jspdf/index.d.ts
@@ -1,8 +1,9 @@
-// Type definitions for jsPDF 1.2
+// Type definitions for jsPDF 1.3
 // Project: https://github.com/MrRio/jsPDF
 // Definitions by: Amber Schühmacher <https://github.com/amberjs>
 //                 Kevin Gonnord <https://github.com/lleios>
 //                 Jackie Weng <https://github.com/jemerald>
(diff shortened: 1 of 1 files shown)


types/jspdf/index.d.ts
```diff
@@ -76,7 +77,7 @@ declare module 'jspdf' {
         setLineCap(style:string|number):jsPDF;
         setLineJoin(style:string|number):jsPDF;
         output(type?:string, options?:any):any;
-        save(filename:string):jsPDF;
+        save(filename:string, options?: {returnPromise: boolean}): Promise<any>;
```
jemerald: This may cause problem for consumers that rely on the old return type.

Consider use function overload?

```suggestion
        save(filename:string):jsPDF;
        save(filename:string, options: {returnPromise: boolean}): Promise<any>;
```

FrankBrullo: The documentation is wrong.
Looking at the plain vanilla javascript code in the library the save function never return an instance of jsPDF.
Here is the link with the code: https://rawgit.com/MrRio/jsPDF/master/docs/jspdf.js.html#line4706
Inside the if statement "if (options.returnPromise === false)" the code doesn't return any value.
So the returned value is either undefined or a promise.
Also, I am returning a just promise instead of "Promise<any> | undefined" so that when the user invoke save with the returnPromise set to true he can do a concatenation like this:
doc.save('myFile.pdf', {returnPromise: true}).then(()=>{ ... });
instead of checking if it is undefined.

I hope it clarifies my choices.

Thanks

jemerald: I see... Maybe this then?

```suggestion
        save(filename:string): void;
        save(filename:string, options: {returnPromise: boolean}): Promise<any>;
```

FrankBrullo: Ok, that's fine.
I have applied your suggestion.

Thanks!

jemerald (approved):

PR #34479, opened 2019-04-04
[@hapi/*] new namespace for hapi modules

lenovouser:


lenovouser:
@SimonSchick Hi, do you mind checking if I missed anything? I didn't change any linting, types etc. as I don't think any enhancement to these areas should be part of this switch. I updated the version header according to their current version of the corresponding package on npm.

lenovouser:
@rafaelsouzaf @saboya I've seen you two pretty active with hapi-related types so maybe you could to a quick check too.

No linting / types etc. though, I've already seen a few things I want to fix (`dtslint` everywhere, remove `require()`'s) but I want to do that in different PRs.

This is just the move `module => @hapi/module`

saboya (approved):
Seems pretty straightforward.

SimonSchick:
@lenovouser the changes look sane, do we know yet if the non-namespace packages are deprecated and if we should potentially deprecate them?

lenovouser:
@SimonSchick I didn't in this change because I also don't know yet, but I think they will be deprecated on npm (the actual modules) at some point, yeah.

I think we should leave them in anyways. In case we need to fix types for the old module versions in the future. Some even have nested older versions in them. I don't think we should remove that. Or what are your thoughts?

I am going to ask Eran what he intends to do with the old modules over Slack.

PR #34480, opened 2019-04-04
[chrome] Add missing type to chrome.runtime

mionim:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [type-RequestUpdateCheckStatus](https://developers.chrome.com/extensions/runtime#type-RequestUpdateCheckStatus)

### DIFF
--- a/types/chrome/index.d.ts
+++ b/types/chrome/index.d.ts
@@ -5242,6 +5242,9 @@ declare namespace chrome.runtime {
         version: string;
     }
 
+    /** Result of the update check. */
+    export type RequestUpdateCheckStatus = 'throttled' | 'no_update' | 'update_available';
+
     export interface PortDisconnectEvent extends chrome.events.Event<(port: Port) => void> { }
 
     export interface PortMessageEvent extends chrome.events.Event<(message: any, port: Port) => void> { }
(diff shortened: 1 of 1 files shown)


types/chrome/index.d.ts
```diff
@@ -5242,6 +5242,9 @@ declare namespace chrome.runtime {
         version: string;
     }
 
+    /** Result of the update check. */
+    export type RequestUpdateCheckStatus = 'throttled' | 'no_update' | 'update_available';
```
matcarlson:
```suggestion
    export interface RequestUpdateCheckStatus = 'throttled' | 'no_update' | 'update_available';
```

matcarlson (changes requested):
Could you modify `type` to `interface` to match the rest of the file?

Checked source, LGTM otherwise.

types/chrome/index.d.ts
MiOnim: The suggested change results in syntax error. This is not an interface. A variable with this new type has to be a `string`, not an `object`.

Also, there are 4 more instances of `type` in this file.

matcarlson: Makes sense. Thanks for the clarification.

matcarlson (approved):

uniqueiniquity:

PR #34482, opened 2019-04-04
Add type for new "ref" proptype in airbnb-prop-type

ahuth:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/airbnb/prop-types/pull/54

### DIFF
--- a/types/airbnb-prop-types/index.d.ts
+++ b/types/airbnb-prop-types/index.d.ts
@@ -5,7 +5,6 @@
 // TypeScript Version: 3.0
 
 import * as PropTypes from 'prop-types';
-import * as React from 'react';
 
 export interface ReactComponentLike {
     setState(...args: any[]): any;
@@ -24,6 +23,12 @@ export interface ReactClassComponentLike {
 export type ReactFunctionComponentLike = (...args: any[]) => PropTypes.ReactNodeLike;
 
 export type ReactTypeLike = string | ReactClassComponentLike | ReactFunctionComponentLike;
+
+export interface ReactRefLike<T> {
+    readonly current: T | null;
+}
+
+export type ReactLegacyRefLike<T> = (instance: T | null) => void | ReactRefLike<T> | null;
 
 export interface Specifier<T = any> {
     max?: number;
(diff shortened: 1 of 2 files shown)


types/airbnb-prop-types/index.d.ts
```diff
@@ -1,10 +1,11 @@
-// Type definitions for airbnb-prop-types 2.11
+// Type definitions for airbnb-prop-types 2.13
 // Project: https://github.com/airbnb/prop-types
 // Definitions by: Miles Johnson <https://github.com/milesj>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
 import * as PropTypes from 'prop-types';
+import * as React from 'react';
```
milesj: We purposefully avoided coupling the `@types/react` package (as seems by all the interfaces below). Can you add a `RefLike` instead?

ahuth: Sure

types/airbnb-prop-types/index.d.ts
```diff
@@ -24,6 +24,12 @@ export type ReactFunctionComponentLike = (...args: any[]) => PropTypes.ReactNode
 
 export type ReactTypeLike = string | ReactClassComponentLike | ReactFunctionComponentLike;
 
+export interface ReactRefLike<T> {
+    readonly current: T | null;
+}
+
+export type ReactLegacyRefLike<T> = (instance: T | null) => void | ReactRefLike<T> | null;
```
ahuth: Not totally sure we want `| null` or not. Included it because React's definition for ref does.

ahuth:
Thanks for looking at this @milesj. Can you take a look at this again, please?

types/airbnb-prop-types/index.d.ts
ahuth: Meh, decided to remove it

types/airbnb-prop-types/index.d.ts
```diff
@@ -24,6 +24,12 @@ export type ReactFunctionComponentLike = (...args: any[]) => PropTypes.ReactNode
 
 export type ReactTypeLike = string | ReactClassComponentLike | ReactFunctionComponentLike;
 
+export interface ReactRefLike<T> {
+    readonly current: T | null;
+}
+
+export type ReactLegacyRefLike<T> = (instance: T | null) => void | ReactRefLike<T>;
```
milesj: I think you want `((instance: T | null) => void) | ReactRefLike<T>`, which is a union of either. Currently it's a return type of either.

ahuth: Yeah, you're right. Nice catch.

I avoided this while testing because the order was flipped

ahuth:
Alright @milesj, I think this is sort of, kind of, probably, maybe right now

milesj (approved):
Looks good, thanks!

PR #34484, opened 2019-04-04
[ldapjs] Add client.search() callback types

pmoleri:

  - Documentation: [client.search()](http://ldapjs.org/client.html#search)
  - Source Code:
    - [Client](https://github.com/joyent/node-ldapjs/blob/v1.0.1/lib/client/client.js)
    - [Messages](https://github.com/joyent/node-ldapjs/tree/v1.0.1/lib/messages): LDAPResult, LDAPMessage, SearchEntry
    - [Attribute](https://github.com/joyent/node-ldapjs/blob/v1.0.1/lib/attribute.js)

### DIFF
--- a/types/ldapjs/index.d.ts
+++ b/types/ldapjs/index.d.ts
@@ -77,7 +77,7 @@ export var Change: {
 
 export type SearchReference = any;
 
-export class SearchCallbackResponse extends EventEmitter {
+export interface SearchCallbackResponse extends EventEmitter {
     on(event: string | symbol, listener: (...args: any[]) => void): this;
     on(event: "searchEntry", listener: (entry: SearchEntry) => void): this;
     on(event: "searchReference", listener: (referral: SearchReference) => void): this;
@@ -293,7 +293,7 @@ export class ApproximateFilter extends Filter {
 	constructor(options: { attribute: string, value: string })
 }
 
-export class AttributeJson {
+export interface AttributeJson {
     type: string;
     vals: string[];
 }
--- a/types/ldapjs/index.d.ts
+++ b/types/ldapjs/index.d.ts
@@ -1,6 +1,6 @@
 // Type definitions for ldapjs 1.0
 // Project: http://ldapjs.org
-// Definitions by: Charles Villemure <https://github.com/cvillemure>, Peter Kooijmans <https://github.com/peterkooijmans>
+// Definitions by: Charles Villemure <https://github.com/cvillemure>, Peter Kooijmans <https://github.com/peterkooijmans>, Pablo Moleri <https://github.com/pmoleri>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 /// <reference types="node" />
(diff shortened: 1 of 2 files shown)


types/ldapjs/index.d.ts
```diff
@@ -75,8 +75,19 @@ export var Change: {
 	new(change: Change): Change;
 }
 
+export type SearchReference = any;
+
+export class SearchCallbackResponse extends EventEmitter {
+    on(event: string | symbol, listener: (...args: any[]) => void): this;
```
uniqueiniquity: Please move this to the bottom of the list of overloads for `on`; otherwise, this case will always be chosen.

types/ldapjs/index.d.ts
```diff
@@ -75,8 +75,19 @@ export var Change: {
 	new(change: Change): Change;
 }
 
+export type SearchReference = any;
+
+export class SearchCallbackResponse extends EventEmitter {
```
uniqueiniquity: I would either make this an interface or not export it; otherwise, someone might import `SearchCallbackResponse` from the library and expect it to exist at runtime.

types/ldapjs/index.d.ts
```diff
@@ -281,3 +292,94 @@ export class NotFilter extends Filter {
 export class ApproximateFilter extends Filter {
 	constructor(options: { attribute: string, value: string })
 }
+
+export class AttributeJson {
```
uniqueiniquity: Same here; it appears that there is no class called `AttributeJson` that's actually exposed from the library at runtime.

uniqueiniquity (changes requested):

types/ldapjs/index.d.ts
pmoleri: I don't know exactly how this works, that was my impression too before looking at [node typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/fcf24503eb7bfebeb6279e2a5b341f078c73df87/types/node/v8/index.d.ts#L5115) which I finally used as reference.
Also tests work just fine matching with the correct overload.

types/ldapjs/index.d.ts
pmoleri: Good catch, this was a mistake.

pmoleri:
@uniqueiniquity thanks for your review, I corrected the things you mentioned. There's still a pending comment, because I'm not sure it's a real issue. If it is an issue I would like to understand why it's not an issue in node type declarations.

types/ldapjs/index.d.ts
uniqueiniquity: I just got a chance to ask around. I was incorrect in stating that this case will always be chosen.
It's true that for most overloads, the first will be chosen, but there's actually special handling for this particular case (i.e. string literals vs. just `string`) from before string literal types were widely usable to effectively "move the more specific ones to the top". As a result, the way the node typings are written are an artifact of that older handling. So this will work fine, but it's more consistent style with all overloads to move it to the bottom. Would you like to move the general ones to the bottom? If not, I'll go ahead and merge.

pmoleri: Ok, I'm glad to do it if now is the more consistent way. I tried to be consistent with node typings :)

uniqueiniquity (approved):

types/ldapjs/index.d.ts
uniqueiniquity: Thank you!

PR #34496, opened 2019-04-05
Update index.ts

BMNTeam:
There is was a bug in Content interface definition where instead of having Style interface for style property 'string' was defined. Also pageSize accept custom page format with width and height, so it would be better to allow this in interface instead Enums only.

Please fill in this template.

If adding a new definition:
- [X ] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration`

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://pdfmake.github.io/docs/document-definition-object/page/>>


evolkmann (approved):

evolkmann:
Hi @BMNTeam,
your changes are good!
The tests will have to be adapted to pass CI.

PR #30255, opened 2018-11-03
[react-redux] implemented proper typings for thunk action creators in TDispatchProps

surgeboris:
- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/reduxjs/redux-thunk#composition
- [x] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #34498, opened 2019-04-05
react-redux: correct `mapDispatchToProps` param type

### LINKED ISSUE #34493: react-redux: dispatch props are not omitted
Using all the latest versions:

``` json
  "dependencies": {
    "@types/react": "^16.8.12",
    "@types/react-dom": "^16.8.3",
    "@types/react-redux": "^7.0.6",
    "typescript": "^3.4.1"
  }
```

``` tsx
import { connect, MapDispatchToProps, MapStateToProps } from 'react-redux';
import { ActionCreator } from 'redux';

type State = {};

type OwnProps = {
    ownProp: number;
};

type StateProps = {
    stateProp: number;
};

type MyAction = ActionCreator<{}>;
type DispatchProps = {
    myAction: MyAction;
};

type Props = OwnProps & StateProps & DispatchProps;

declare const MyComponent: React.FC<Props>;

declare const mapStateToProps: MapStateToProps<StateProps, OwnProps, State>;
declare const mapDispatchToProps: MapDispatchToProps<DispatchProps, OwnProps>;

const enhance = connect(
    mapStateToProps,
    mapDispatchToProps,
);
const MyComponent2 = enhance(MyComponent);

// Unexpected type error
// Property 'myAction' is missing in type '{ ownProp: number; }' but required in type 'Readonly<Pick<Props, "myAction" | "ownProp"> & OwnProps>'.
const el1 = <MyComponent2 ownProp={1} />;
```

---

  - Authors: @tkqubo @kenzierocks @clayne11 @tansongyang @nicholasboll @mdibyo @kallikrein @val1984 @jrakotoharisoa @apapirovski @surgeboris @soerenbf

OliverJAsh:
Fixes https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34493

---

Please fill in this template.


types/react-redux/react-redux-tests.tsx
```diff
@@ -93,6 +91,16 @@ function MapDispatch() {
 
     const verifyNull = <TestNull foo='bar' />;
 
+    // We deliberately cast the right-hand side to any because…
+    const mapDispatchToProps2: MapDispatchToProps<DispatchProps, OwnProps> = {} as any;
```
OliverJAsh: I need to finish writing this comment, but truth is, I don't understand why.

If 2 values have the same type annotation but different values on the right-hand side (on none at all if it's a `declare`), TS seems to behave differently. This is essentially the behaviour we're trying to test for here. I don't know why TS does this, nor what it's called. Can anyone enlighten me? Then I will finish writing my comment about why this test is needed in addition to the one above.

``` ts
type MyUnion<T> = (() => T) | T;

const value1: MyUnion<{ foo: number }> = { foo: 1 };
// No type error
value1.foo;

const value2: MyUnion<{ foo: number }> = 1 as any;
// Type error
value2.foo;

declare const value3: MyUnion<{ foo: number }>
// Type error
value3.foo;
```

types/react-redux/index.d.ts
```diff
@@ -171,7 +171,7 @@ export interface Connect {
 
     <no_state = {}, TDispatchProps = {}, TOwnProps = {}>(
         mapStateToProps: null | undefined,
-        mapDispatchToProps: TDispatchProps,
+        mapDispatchToProps: MapDispatchToPropsParam<TDispatchProps, TOwnProps>,
```
OliverJAsh: Without this, `connect` would use the whole `MapDispatchToProps` (_union_) for the `TDispatchProps` (object) generic.

uniqueiniquity (changes requested):
@Jessidhia @surgeboris @soerenbf @mDibyo I'd very much appreciate a review from one of you or another owner. I don't actually know enough about `react-redux` to be sure about this change and it seems pretty impactful.

@OliverJAsh if no one reviews in a few days (let's say next Wednesday) feel free to tag me and I'll take another look.

OliverJAsh:
> if no one reviews in a few days (let's say next Wednesday) feel free to tag me and I'll take another look.

Will wait until Wednesday

OliverJAsh:
> Will wait until Wednesday

@uniqueiniquity It's Wednesday! ⏰

Pinging other maintainers for a review: @tkqubo @kenzierocks @clayne11 @tansongyang @NicholasBoll @mDibyo @Kallikrein @val1984 @jrakotoharisoa @apapirovski @surgeboris @soerenbf

types/react-redux/index.d.ts
uniqueiniquity: I took some time and I think I understand your change now.

One question: why is `ResolveThunks` currently used only in the "object" case?
Is there any potential negative effect of using that in the union case (i.e. is it a problem if the type system can't tell whether `mapDispatchToProps` is an object or not, but at runtime it actually is a function, but the `ResolveThunks` forced something else to be a certain way elsewhere)?

uniqueiniquity (changes requested):

types/react-redux/index.d.ts
OliverJAsh: I'm not sure I can answer that question. I did a `git blame` to find where this was originally added, and I came across https://github.com/DefinitelyTyped/DefinitelyTyped/pull/30255, however I couldn't find any discussion in there about this. /cc @surgeboris who implemented that PR.

uniqueiniquity: Note that my point is not that I think it should be used in the non-object case, but rather that I can't tell from documentation whether you would or wouldn't want to. I think that this change would, when providing a function as `mapDispatchToProps`, first try checking without resolving thunks and then try without it.

OliverJAsh: I'm not quite sure I follow. Can you elaborate more on what changes you think might need to be made, or what I might need to investigate before we can get this PR approved? Sorry if I'm being slow here.

uniqueiniquity: I looked at it some more. I think what I worried about was a case like the following:
```ts
const simpleAction = (payload: boolean) => ({
        type: 'SIMPLE_ACTION',
        payload,
    });

    interface OwnProps {
        foo: string;
    }
    interface TestComponentProps extends OwnProps {
        simpleAction: typeof simpleAction;
        simpleAction2: typeof simpleAction;
    }
    class TestComponent extends React.Component<TestComponentProps> { }

    const mapStateToProps = ({ foo }: { foo: string }) => ({ foo })

    const mapDispatchToProps2 = (dispatch: Dispatch) => {
        return {
            simpleAction: (payload: boolean) => dispatch(simpleAction(payload)),
            simpleAction2: (payload: boolean) => () => dispatch(simpleAction(payload))
        };
    };

    const Test1 = connect(mapStateToProps, mapDispatchToProps)(TestComponent);
    const verify = <div>
        <Test1 foo='bar' />
    </div>;
```
I was under the impression that the type checker would first pick the `MapDispatchToPropsNonObject` overload and fail to check (because of the type of `simpleAction2`), _and then go back_ and pick the second overload with the new change and succeed because of `ResolveThunks` thinking that the value provided for `simpleAction2` is a thunk. Turns out that's not how the overload resolution works - there's no backtracking like I expected. Do you want to add this as an `// $ExpectError` test? if not, I'll go ahead and merge :) Thanks!

uniqueiniquity (approved):

types/react-redux/index.d.ts
OliverJAsh: > Do you want to add this as an `// $ExpectError` test? if not, I'll go ahead and merge :) Thanks!

Thanks for clarifying. Unfortunately I still don't understand, so I'm hesitant to add the test. Would you like me to add it anyway? Or perhaps that test could be added in a separate PR, if it's not related to this change?

uniqueiniquity: That's fine, I'll consider adding it separately.
Thanks for the contribution; sorry I delayed it so long!

PR #34503, opened 2019-04-05
Add type definitions for array-binarysearch.closest

hvoecking:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/array-binarysearch.closest/index.d.ts
+++ b/types/array-binarysearch.closest/index.d.ts
@@ -1,0 +1,13 @@
+// Type definitions for array-binarysearch.closest 0.2
+// Project: https://www.npmjs.com/package/array-binarysearch.closest
+// Definitions by: Heye Vöcking <https://github.com/hvoecking>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export function binarySearch<T>(
+    arr: ReadonlyArray<T>,
+    val: T,
+    fn?: null | ((itm: T, val: T, m: any, arr: ReadonlyArray<T>) => number),
+    ths?: any,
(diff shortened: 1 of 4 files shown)


types/array-binarysearch.closest/index.d.ts
```diff
@@ -0,0 +1,13 @@
+// Type definitions for array-binarysearch.closest 0.2
+// Project: https://www.npmjs.com/package/array-binarysearch.closest
+// Definitions by: Heye Vöcking <https://github.com/hvoecking>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export function binarySearch<T>(
```
DanielRosenwasser: This code won't work - the original package is declared with `module.exports = `, so this package should use `export = binarySearch`.

types/array-binarysearch.closest/index.d.ts
```diff
@@ -0,0 +1,13 @@
...
+    arr: ReadonlyArray<T>,
+    val: T,
+    fn?: null | ((itm: T, val: T, m: any, arr: ReadonlyArray<T>) => number),
```
DanielRosenwasser: Technically this function should only be generic when `fn` is provided. Otherwise it really only safely works on `ReadonlyArray<string>`s and `ReadonlyArray<number>`s.  I'd probably make overloads to model this.

DanielRosenwasser (changes requested):

hvoecking:
@DanielRosenwasser thank you for the review. I addressed your comments with a fixup commit

DanielRosenwasser:
Great job! Thank you!

PR #34507, opened 2019-04-05
Created new type definition for save-csv library.

FrankBrullo:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/save-csv/index.d.ts
+++ b/types/save-csv/index.d.ts
@@ -1,16 +1,17 @@
-// Type definitions for save-csv 4.1.0
+// Type definitions for save-csv 4.1
 // Project: https://github.com/silverwind/save-csv
 // Definitions by: Frank Brullo <https://github.com/FrankBrullo>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.1
 
-interface SaveCsvProps {
-    filename: string,
-    sep?: string,
-    eol?: string,
-    quote?: string,
-    bom?: boolean,
-    mime?: string,
-    formatter?: (value: string | JSON) => void
+export interface SaveCsvProps {
+    filename: string;
+    sep?: string;
+    eol?: string;
+    quote?: string;
+    bom?: boolean;
+    mime?: string;
+    formatter?: (value: string | JSON) => void;
 }
-export default function saveCsv (array: ReadonlyArray<any>, options?: SaveCsvProps): void;
+
+export function saveCsv(array: ReadonlyArray<any>, options?: SaveCsvProps): void;
--- a/types/save-csv/index.d.ts
+++ b/types/save-csv/index.d.ts
@@ -1,0 +1,16 @@
+// Type definitions for save-csv 4.1.0
+// Project: https://github.com/silverwind/save-csv
+// Definitions by: Frank Brullo <https://github.com/FrankBrullo>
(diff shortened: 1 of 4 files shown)


types/save-csv/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+    filename: string;
+    sep?: string;
+    eol?: string;
+    quote?: string;
+    bom?: boolean;
+    mime?: string;
+    formatter?: (value: string | JSON) => void;
+}
+
+export default function saveCsv(array: ReadonlyArray<any>, options?: SaveCsvProps): void;
```
DanielRosenwasser: This package was declared using `module.exports = ...` or some variant. It should be declared with an `export = ...` like one of the following:

### Variables

```ts
declare var xyz: SomeType;

export = xyz;
```

### Classes

```ts
class Foo {
  // ...
}

export = Foo
```

### Functions

```ts
function foo(): any;

export = foo;
```

```ts
import foo = require("foo");
```

DanielRosenwasser (changes requested):

types/save-csv/index.d.ts
FrankBrullo: Thank you @DanielRosenwasser . The documentation was very helpful.
I tried before like you said but it didn't work (probably because I didn't know the `esModuleInterop` switch.
At the end of the day, I used `export default` because it was working out-of-the-box with a default project created with create-react.
If I apply your changes, how can I make sure that anybody else will know that they have to use such that switch once they download the type definition?

DanielRosenwasser: I think that's up to the TypeScript project - we should provide a better error message in those cases. I just opened up https://github.com/Microsoft/TypeScript/issues/30837

types/save-csv/save-csv-tests.ts
```diff
@@ -0,0 +1,6 @@
+import saveCsv from 'save-csv';
+
+saveCsv([
+    {a: 1, b: 2},
+    {a: 3, b: 4},
+  ]);
```
DanielRosenwasser: Can you remove the stray spaces?

types/save-csv/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "npm-naming": false
```
DanielRosenwasser: Why is this still disabled?

DanielRosenwasser (changes requested):

types/save-csv/index.d.ts
FrankBrullo: @DanielRosenwasser I am not really happy to implement a solution that is syntactically perfect but requires manual fiddling from the user once they install the type.
I would rather prefer a less perfect solution but which works in any case.

Anyway if this is how you want it for the sake of moving this forward I will do it.

types/save-csv/save-csv-tests.ts
FrankBrullo: Done

types/save-csv/tslint.json
FrankBrullo: Because I was waiting for your answer. Now it shouldn't be there anymore.

FrankBrullo (commented):
Applied all the changes

DanielRosenwasser:
Thanks for your patience and for this PR!

PR #34520, opened 2019-04-05
[scheduler] Add tracing module

eps1lon:
Please fill in this template.

  - type definitions are extracted from the flow types in [scheduler/src/Tracing.js](https://github.com/facebook/react/tree/v16.8.0/packages/scheduler/src/Tracing.js) and [scheduler/src/TracingSubscriptions.js](https://github.com/facebook/react/tree/v16.8.0/packages/scheduler/src/TracingSubscriptions.js) which are bundled into `scheduler/tracing`
  - tests are losely based on [Tracing-test.internal.js](https://github.com/facebook/react/blob/v16.8.0/packages/scheduler/src/__tests__/Tracing-test.internal.js)

If removing a declaration:

If you only use this module in `development` you can narrow the types with [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) and 

```ts
import { Build } from 'scheduler/tracing';

declare module 'scheduler/tracing' {
  interface Build { type: 'development'; }
}
```


types/scheduler/tracing.d.ts
```diff
@@ -0,0 +1,93 @@
...
+ * The callback is immediately returned if the enableSchedulerTracing is disabled.
+ * It is unclear for which bundles this is the case.
+ *
+ * @param callback
+ * @param threadID
+ */
+export function unstable_wrap<T extends (...args: any[]) => any>(
+  callback: T,
+  threadID?: number
+): WrappedFunction<T>;
```
Methuselah96: Since the callback can be immediately return, snouldn't the return type be `WrappedFunction<T> | T`?

Methuselah96 (changes requested):

types/scheduler/tracing.d.ts
eps1lon: Sounds right. I had this expressed in the first iteration but removed it for some reason.

eps1lon: You're correct. This is the case if the `enableSchedulerTracing` feature flag is disabled which happens at build time. This affects other variables as well.

The best thing would be to type all bundles: `development`, `production` and `profiling`. I'll have some ideas for it. 

What do you think of assuming the development build for now?

eps1lon:
Added the following to the issue description:

If you only use this module in development you can narrow the types with [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) and

```ts
import { Build } from 'scheduler/tracing';

declare module 'scheduler/tracing' {
  interface Build { type: 'development'; }
}
```

types/scheduler/tracing.d.ts
```diff
@@ -0,0 +1,131 @@
...
+}
+
+export interface SubscriberRef {
+  current: Subscriber | null;
+}
+
+export const __interactionsRef: IfSchedulerTracing<InteractionsRef, null>;
+export const __subscriberRef: IfSchedulerTracing<SubscriberRef, null>;
+
+export function unstable_clear<T>(callback: () => T): T;
```
Methuselah96: Sorry, one last question. Since this function has a try/catch block around the callback and doesn't return anything if it catches an error, should the return type be `T | undefined`? This is also true of a couple other functions.

Methuselah96 (changes requested):

types/scheduler/tracing.d.ts
eps1lon: It has a try-finally. The error isn't caught. I had to get used to this as well. I had a bunch of `| undefined` types in place until I realized that they are never caught. `finally` is just a cleanup construct.

Methuselah96 (approved):

PR #34526, opened 2019-04-06
create types for robust-point-in-polygon

matthewboman:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/robust-point-in-polygon/index.d.ts
+++ b/types/robust-point-in-polygon/index.d.ts
@@ -3,6 +3,8 @@
 // Definitions by: Crash Springfield <https://github.com/crashspringfield>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export type Point = [number, number];
+type Point = [number, number];
 
-export function robustPointInPolygon(vs: Point[], point: Point): number;
+function robustPointInPolygon(vs: Point[], point: Point): number;
+
+export = robustPointInPolygon;
--- a/types/robust-point-in-polygon/index.d.ts
+++ b/types/robust-point-in-polygon/index.d.ts
@@ -1,0 +1,8 @@
+// Type definitions for robust-point-in-polygon 1.0
+// Project: https://github.com/mikolalysenko/robust-point-in-polygon
+// Definitions by: Crash Springfield <https://github.com/crashspringfield>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export type Point = [number, number];
+
+export function robustPointInPolygon(vs: Point[], point: Point): number;
--- a/types/robust-point-in-polygon/index.d.ts
+++ b/types/robust-point-in-polygon/index.d.ts
@@ -5,6 +5,9 @@
 
 type Point = [number, number];
 
-function robustPointInPolygon(vs: Point[], point: Point): number;
+declare function robustPointInPolygon(vs: Point[], point: Point): number;
 
-export = robustPointInPolygon;
+export {
+  Point,
+  robustPointInPolygon
+};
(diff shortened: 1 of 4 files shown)


types/robust-point-in-polygon/robust-point-in-polygon-tests.ts
```diff
@@ -0,0 +1,4 @@
+import assert = require('assert');
+import inPolygon = require('robust-point-in-polygon');
+
+assert.equal(typeof inPolygon, 'function');
```
DanielRosenwasser: This isn't test that's appropriate for DefinitelyTyped. Just write code for how a user is supposed to use this library.

types/robust-point-in-polygon/index.d.ts
```diff
@@ -0,0 +1,8 @@
+// Type definitions for robust-point-in-polygon 1.0
+// Project: https://github.com/mikolalysenko/robust-point-in-polygon
+// Definitions by: Crash Springfield <https://github.com/crashspringfield>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export type Point = [number, number];
+
+export function robustPointInPolygon(vs: Point[], point: Point): number;
```
DanielRosenwasser: This package was declared using `module.exports = ...` or some variant. It should be declared with an `export = ...` like one of the following:

### Variables

```ts
declare var xyz: SomeType;

export = xyz;
```

### Classes

```ts
class Foo {
  // ...
}

export = Foo
```

### Functions

```ts
function foo(): any;

export = foo;
```

You can then import it with

```ts
import foo = require("foo");
```

DanielRosenwasser (changes requested):

types/robust-point-in-polygon/robust-point-in-polygon-tests.ts
```diff
@@ -0,0 +1,5 @@
+import * as robustPointInPolygon from 'robust-point-in-polygon');
```
DanielRosenwasser: Since this is a callable or constructable entity, it isn't legal in ES2015 to import and call/construct this with a namespace import like `import * as foo from 'foo'`, so could you please convert this to either

```ts
import foo = require('foo')
```

or the following with `esModuleInterop` turned on

```ts
import foo from 'foo';
```

types/robust-point-in-polygon/robust-point-in-polygon-tests.ts
crashspringfield: I've updated the tests

types/robust-point-in-polygon/index.d.ts
crashspringfield: I've updated the export to 
```
export {
  Point,
  robustPointInPolygon
};
```

types/robust-point-in-polygon/index.d.ts
```diff
@@ -0,0 +1,13 @@
...
+// Definitions by: Crash Springfield <https://github.com/crashspringfield>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+type Point = [number, number];
+
+declare function robustPointInPolygon(vs: Point[], point: Point): number;
+
+export {
+  Point,
+  robustPointInPolygon
```
DanielRosenwasser: This is not the same thing as what the package does, and won't work at runtime. You really want something closer to

```ts
type Point = [number, number];

declare function robustPointInPolygon(vs: Point[], point: Point): number;
export = robustPointInPolygon;
```

or

```ts
declare namespace robustPointInPolygon {
    type Point = [number, number];
}

declare function robustPointInPolygon(vs: Point[], point: Point): number;
export = robustPointInPolygon;
```

if exporting the `Point` type is that useful.

Then you can use `import robustPointInPolygon = require("robust-point-in-polygon");`

DanielRosenwasser (changes requested):

types/robust-point-in-polygon/index.d.ts
crashspringfield: I've gone ahead and updated it with your suggestions, and instead of importing/exporting the Point type, just re-declared it in the test so the test would run. Thanks for the help!

PR #34527, opened 2019-04-06
pageRows should be a DerrivedDataObject, not a number

cpepin:
- [X] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/tannerlinsley/react-table/blob/v6/src/index.js>

As you can see from source, pageRows is actually just a subset of the resolved data.  I came across this when writing a custom table, leveraging the table state for paginated results.


Gelio (approved):
Sorry for the late reply. The code change looks good to me 👍 Nice catch

uniqueiniquity:

PR #34535, opened 2019-04-07
Event specific arg types

joeflateau:
Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://developers.google.com/doubleclick-gpt/reference#googletag.events.SlotRenderEndedEvent

If removing a declaration:

### DIFF
--- a/types/doubleclick-gpt/index.d.ts
+++ b/types/doubleclick-gpt/index.d.ts
@@ -2,6 +2,7 @@
 // Project: https://developers.google.com/doubleclick-gpt/reference
 // Definitions by: John Wright <https://github.com/johngeorgewright>
 //                 Steven Joyce <https://github.com/steven-joyce>
+//                 Joe Flateau <https://github.com/joeflateau>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.2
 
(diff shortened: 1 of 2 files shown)


types/doubleclick-gpt/index.d.ts
```diff
@@ -155,32 +164,23 @@ declare namespace googletag {
     }
 
     namespace events {
-        interface ImpressionViewableEvent {
-            serviceName: string;
-            slot: Slot;
-        }
-
-        interface SlotOnloadEvent {
```
johngeorgewright: Removing these would be a breaking change to anyone relying on them. According to the [docs](googletageventsimpressionviewableevent) these simply extend `googletag.events.Event`... might be worth keeping them as empty extensions.

johngeorgewright (changes requested):

types/doubleclick-gpt/index.d.ts
joeflateau: I did initially have them as empty extensions but tslint complained about them. I’ll re add and address the linter by disabling the rule. Good call.

johngeorgewright: Ah, yes. You can either turn off the linter there, or create "types":

```typescript
type SlotOnloadEvent = Event;
```

johngeorgewright (approved):

PR #34549, opened 2019-04-08
[multy]: TS definitions for multy

dolezel:


plantain-00 (approved):

types/multy/index.d.ts
```diff
@@ -0,0 +1,20 @@
...
+
+declare module "koa" {
+    interface Request {
+        body: any;
+    }
+}
+
+declare function multy(opts?: busboy.BusboyConfig): Koa.Middleware;
+
+declare namespace multy {}
```
DanielRosenwasser: This is an inappropriate hack. [Please read more here from the DefinitelyTyped README](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/fbba9f55beab742b9d1c0fa9ef714e0771ad9d05#should-i-add-an-empty-namespace-to-a-package-that-doesnt-export-a-module-to-use-es6-style-imports)

DanielRosenwasser (changes requested):

types/multy/index.d.ts
dolezel: I added alias for options, so user can use type directly from this package.

PR #34583, opened 2019-04-09
[@types/auth0-js] add missing error_description field

marknelissen:
Auth0-js is not consistent in the use of errorDescription within error objects returned to the callback method, not even within the same method (eg. WebAuth.checkSession). Depending on the error (and thus the place where it is constructed) the description in contained in either errorDescription or error_description.

On top of this, the official documentation snippets make use of error_description. It is unclear which is the "official" one, but technically, both are possible. So the typing should reflect this.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/auth0/auth0.js/blob/master/src/web-auth/index.js>>

### DIFF
--- a/types/auth0-js/index.d.ts
+++ b/types/auth0-js/index.d.ts
@@ -1,9 +1,10 @@
-// Type definitions for Auth0.js 8.11
+// Type definitions for Auth0.js 9.10
 // Project: https://github.com/auth0/auth0.js
 // Definitions by: Adrian Chia <https://github.com/adrianchia>
 //                 Matt Durrant <https://github.com/mdurrant>
 //                 Peter Blazejewicz <https://github.com/peterblazejewicz>
 //                 Bartosz Kotrys <https://github.com/bkotrys>
+//                 Mark Nelissen <https://github.com/marknelissen>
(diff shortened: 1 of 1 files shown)


adrianchia (approved):

uniqueiniquity (changes requested):
Since this is a popular package, [perhaps it would be best to create a `v8` directory](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/a26cc7b6f60de02a152168a21a9159a856bf581b#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package).

adrianchia:
I believe versions before v9 are [deprecated](https://auth0.com/docs/libraries/auth0js/v9/migration-guide#should-i-migrate-to-v9-) now. We could probably verify what is missing.

uniqueiniquity (approved):

marknelissen:
This change is also compatible with v8, since the behaviour I described at the start of the pull request is also present in v8 https://github.com/auth0/auth0.js/blob/v8.12.3/src/web-auth/index.js . It is thus not a breaking change.

PR #34585, opened 2019-04-09
fix slonik query type

mmkal:
also add some generics and new options on client configuration

@sebald @gajus


sebald:
@mmkal Awesome! Just two little remarks:

Are these still for v15.4? Otherwise it should be update in the `index.d.ts` header.

Also, could you add yourself to the `Definitions by:`? This way you can also review the typings 🙂

sebald (approved):
Thanks!

uniqueiniquity:

PR #34587, opened 2019-04-09
Remove sequelize types

samalexander:
Removing types for the sequelize library, which now ships official typings.

If removing a declaration:


samalexander:
Requested sequelize be added to the types-publisher dependencies whitelist:
https://github.com/Microsoft/types-publisher/pull/604

todd:
The types that ship with Sequelize are for v5+ only. The types in this repository are for v4. Please do not remove these declaration files.

samalexander:
@todd the types packages would still exist for v4. I assume you want to keep them around in case of issues with the definitions that require patching? If that is the case, can we warn the user to use the official types for v5+, which would help avoid confusion?

todd:
@samalexander Yeah, agreed that we need a way to inform users about the distinction between declaration files. Open to suggestions. I'm thinking at minimum we add a `postinstall` hook that spits out some text about the difference in versions. The header comments in the declaration files themselves list what version the types are supposed to be for, but could probably be more specific. It might also be worth adding documentation directly to Sequelize about what to use for specific versions.

Issue #25202, opened 2018-04-21
Cannot use UserAgentParser as class type from "ua-parser-js"


JoshuaKGoldberg:
If you know how to fix the issue, make a pull request instead.

- [x] I tried using the `@types/ua-parser-js` package and had problems.
- [x] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @superduper @legendecas @MeLlamoPablo>

```typescript
import UserAgentParser = require("ua-parser-js");

const parser: UserAgentParser = new UserAgentParser();
//            ~~~~~~~~~~~~~~~
```

```
Cannot find name 'UserAgentParser'.
```

MeLlamoPablo:
This could be fixed by changing:

```diff
declare module "ua-parser-js" {
+    export class UAParser {
-    class UAParser {
        static VERSION: string;
        static BROWSER: IUAParser.BROWSER;
        static CPU: IUAParser.CPU;
        static DEVICE: IUAParser.DEVICE;
        static ENGINE: IUAParser.ENGINE;
        static OS: IUAParser.OS;

        /* ... */
    }
-   const exported: typeof UAParser & { UAParser: typeof UAParser };
-   export = exported
}
```

However it would invalidate the following import syntaxes:

```typescript
import * as UAParser from "ua-parser-js"
import UAParser = require("ua-parser-js")
```

making this one be the only one valid:

```typescript
import {UAParser} from "ua-parser-js"
```

I don't think this is a big deal at all because that's the standard ES6/TypeScript import syntax. @superduper @legendecas any thoughts?

JoshuaKGoldberg:
Interestingly enough, it looks like the library supports both syntaxes, but only mentions the old-style `import UAParser = require("ua-parser-js")` in the README. 

https://github.com/faisalman/ua-parser-js/blob/dfb8e758d107655897761115c4b1780d0d989e3c/src/ua-parser.js#L1009

https://github.com/faisalman/ua-parser-js/blob/dfb8e758d107655897761115c4b1780d0d989e3c/src/ua-parser.js#L1040

[...]

PR #34594, opened 2019-04-09
[@types/ua-parser-js] issue #25202 fix

KseniaKapnik:
updating export declaration to be the standard ES6/TypeScript
@MeLlamoPablo

### DIFF
--- a/types/ua-parser-js/index.d.ts
+++ b/types/ua-parser-js/index.d.ts
@@ -139,7 +139,7 @@ declare namespace IUAParser {
 }
 
 declare module "ua-parser-js" {
-    class UAParser {
+    export class UAParser {
         static VERSION: string;
         static BROWSER: IUAParser.BROWSER;
         static CPU: IUAParser.CPU;
@@ -192,6 +192,4 @@ declare module "ua-parser-js" {
          */
         getResult(): IUAParser.IResult;
     }
-    const exported: typeof UAParser & { UAParser: typeof UAParser };
-    export = exported
 }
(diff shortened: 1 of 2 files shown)


MeLlamoPablo (approved):

types/ua-parser-js/index.d.ts
```diff
@@ -139,7 +139,7 @@ declare namespace IUAParser {
 }
 
 declare module "ua-parser-js" {
-    class UAParser {
+    export class UAParser {
```
uniqueiniquity: This doesn't seem right to me.
In particular, the module _is_ the class (i.e. it's a constructable module) as shown in [the source](https://unpkg.com/ua-parser-js@0.7.19/src/ua-parser.js).
However, this change makes the type say that the module is an object with a property `UAParser`, which will fail at runtime.

If you want to use ES6 imports (i.e. `import UAParser from 'ua-parser-js';`), you can use the `--esModuleInterop` compiler setting. This doesn't seem right to me.
In particular, the module _is_ the class (i.e. it's a constructable module).
However, this change makes the type say that the module is an object with a property `UAParser`, which will fail at runtime.

If you want to use ES6 imports (i.e. `import UAParser from 'ua-parser-js';`), you can use the `--esModuleInterop` compiler setting.

uniqueiniquity (changes requested):

types/ua-parser-js/index.d.ts
MeLlamoPablo: It doesn't fail at runtime though. There is this line in the source: `exports.UAParser = UAParser;`, so the module is the class, but also contains the class, which makes this syntax valid.

If you want, you can take a look at #25202 for more context on this PR.

uniqueiniquity: Oops, my mistake! I missed that piece.

uniqueiniquity (approved):

uniqueiniquity:

PR #34600, opened 2019-04-09
Added react-native-dotenv

hmajid2301:
Please fill in this template.

If adding a new definition:


hmajid2301:
I am a bit stuck with one, the way the library works is you don't know the names of the imports 
`import { API_KEY, ANOTHER_CONFIG } from 'react-native-dotenv'`, I am unsure how to define it so that the library can export something with any name (as a type string). Currently I have only got this to work 
by using `env` then calling it like `env.API_KEY`. So we don't want to hardcode `API_KEY` or `CONFIG_OPTION` in the interface in the `index.d.ts` file. 

Thanks

types/react-native-dotenv/index.d.ts
```diff
@@ -0,0 +1,13 @@
...
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+interface Env {
+    [name: string]: string;
+}
+
+declare const env: Env;
+
+export { env };
```
uniqueiniquity: Instead of this export, you should do `export = env;`.
This way the import can match what will exist at runtime.
Unfortunately, TS won't let you do what's presented in the docs `import { ANY_STRING_HERE } from "react-native-dotenv";` - there's no way to say a module has any possible export. But you will be able to do `import env = require("react-dotnet-env");` or `import env from "react-dotnet-env"` with the `--esModuleInterop` flag.

uniqueiniquity (changes requested):

hmajid2301:
Added Fixes

uniqueiniquity (approved):

PR #34601, opened 2019-04-09
[@types/react-select] Make SingleValue children prop accept children of type ReactNode

yaroslavgribov:
I think it makes sense for SingleValue component to accept children of type `ReactNode`, in consistency with other custom components accepting children of type `ReactNode`, not just string.
I wonder if such a change deserves a version bump, if it does - I'll be happy to do so.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <[Control.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7c33a958277e53aa335dcddec28ebfe6502638ca/types/react-select/lib/components/Control.d.ts#L17) [Option.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7c33a958277e53aa335dcddec28ebfe6502638ca/types/react-select/lib/components/Option.d.ts#L25)>

### DIFF
--- a/types/react-select/lib/components/SingleValue.d.ts
+++ b/types/react-select/lib/components/SingleValue.d.ts
@@ -1,4 +1,4 @@
-import { ComponentType } from 'react';
+import { ComponentType, ReactNode } from 'react';
 import { colors, spacing } from '../theme';
 import { CommonProps } from '../types';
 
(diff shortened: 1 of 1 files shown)


claasahl (approved):
Thanks for your contribution, @yaroslavgribov

PR #34602, opened 2019-04-09
add plaid-link defs

afholderman:
Please fill in this template.

If adding a new definition:
- [X ] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration`
- [ X] If this is for an NPM package, match the name. If not, do not conflict with the name of an NPM package.
- [x ] Create it with `dts-gen --dt`, not by basing it on an existing project.
- [x ] `tslint.json` should be present, and `tsconfig.json` should have `noImplicitAny`, `noImplicitThis`, `strictNullChecks`, and `strictFunctionTypes` set to `true`.

Parameter reference in docs: https://plaid.com/docs/#parameter-reference


types/plaid-link/index.d.ts
```diff
@@ -0,0 +1,117 @@
+// Type definitions for non-npm package plaid-link-browser 2.0
+// Project: https://github.com/plaid/link (Does not have to be to GitHub, but prefer linking to a source code repository rather than to a project website.)
+// Definitions by: Aaron Holderman <https://github.com/afholderman>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped/
+// TypeScript Version: 2.4
+
+/*~ If this module is a UMD module that exposes a global variable 'myLib' when
```
DanielRosenwasser: Delete these comments.

DanielRosenwasser: These comments should be removed

types/plaid-link/index.d.ts
```diff
@@ -0,0 +1,117 @@
...
+        Plaid: {
+            create(params: Plaid.CreateConfig): Plaid.LinkHandler;
+            open(): void;
+            exit(force?: boolean): void;
+        };
+        linkHandler: Plaid.LinkHandler;
+    }
+}
+
+export namespace Plaid {
```
DanielRosenwasser: Since you're making a global value declaration, why not just make `Plaid` a global namespace that exports 3 functions?

types/plaid-link/index.d.ts
```diff
@@ -0,0 +1,117 @@
...
+        verification_status: VerificationStatus;
+    }
+    interface Error {
+        display_message: string;
+        error_code: string;
+        error_message: string;
+        error_type: string;
+    }
+    enum EventName {
+        ERROR = "ERROR",
```
DanielRosenwasser: The library doesn't actually export an enum at runtime.

DanielRosenwasser (changes requested):

types/plaid-link/index.d.ts
afholderman: Is it preferable in this case then to use a string union type similar to VerificationStatus above?

types/plaid-link/index.d.ts
afholderman: I'm sorry I don't understand this request, do you have an example of what you're looking for?

DanielRosenwasser: Sure - would it make sense to get rid of the `declare global` block and write the namespace as a global with 3 functions?

```ts
namespace Plaid {
    export function create(params: Plaid.CreateConfig): Plaid.LinkHandler;
    export function open(): void;
    export function exit(force?: boolean): void;

    // ...

    type EventName =
        | "ERROR"
        // ...
}
```

types/plaid-link/index.d.ts
DanielRosenwasser: Yup, I think so.

DanielRosenwasser (changes requested):
I left some new comments

types/plaid-link/index.d.ts
afholderman: I tested this and it does not work as expected. Plaid-link loads via a script and is made available on the window object. Without out it in the declare global Window TS throws: Property 'Plaid' does not exist on type 'Window'.

afholderman: Here is a sandbox with an example implementation: https://codesandbox.io/s/8k2qll1900?fontsize=14&module=%2Fsrc%2FPlaidComponent%2FPlaidComponent.tsx

afholderman: Hey @DanielRosenwasser if you get a chance mind looking at the above?

afholderman:
@DanielRosenwasser would you mind taking a look at this again?

DanielRosenwasser:
Thanks!

PR #34614, opened 2019-04-10
Added construct and function signature for ShellString

msanguineti:
I have added a construct and function signature to ShellString in order to do something like:

```js
var foo = new ShellString('hello');
```

This is the example given in the [shelljs](https://github.com/shelljs/shelljs#shellstringstr) docs.

Previously I had to add `//@ts-ignore` to this line in order to use the types in my scripts.

Also, the function signature allows for:

```js
var foo = ShellString('hello')
```

Both function and construct have overloaded methods that accept string arrays (`[ ]` and `...`), e.g.:

```js
var foo = new ShellString('hello', 'world')
var bar = ShellString(['hello', 'world'])
```

Please fill in this template.

https://github.com/shelljs/shelljs#shellstringstr


types/shelljs/index.d.ts
```diff
@@ -965,6 +966,53 @@ export type ShellString = string & ShellReturnValue;
...
+	 *
+	 * as per example in shelljs docs:
+	 * https://github.com/shelljs/shelljs#shellstringstr
+	 *
+	 * @param value 	The string value to wrap.
+	 * @return				A string-like object with special methods.
+	 */
+	new(value: string): ShellString;
+	new(value: string[]): ShellArray;
+	new(...value: string[]): ShellArray;
```
gkalpak: This signature seems invalid :confused:

The docs only show the `new (value: string)` signature, although they do mention wrapping an array and exposing `stdout` and `code` (without however explaining that these can be passed as arguments).

So, it is not clear what can be considered public API. It might be that people are only expected to use the first arg (and the other two are only used by shelljs internally) - not sure.

So, we should either only have signatures for one argument (`string` or `string[]`) or all args (based on [the code][1]):

```ts
(stdout: string, stderr?: string, code?: number): ShellString;
(stdout: string[], stderr?: string, code?: number): ShellString;
```

WDYT?

[1]: https://github.com/shelljs/shelljs/blob/9237f66c5/src/common.js#L147

gkalpak (changes requested):

types/shelljs/index.d.ts
msanguineti: Initially, I went for the full blown `ShellString(stdout, stderr, code)` but I believe that this is meant only for internal use. They are setting those after wrapping so that you can check what went wrong. 

However, the signatures I have included do follow the docs. Maybe, I should remove the function signature since the docs only show the `new` construct, although I am actually using these signatures extensively in scripts. It is really nice to do something like:

```js
ShellString(JSON.stringify(my_object, null, 2)).to('./file.json')
```

But I swear I saw somewhere 'official' where they where using the function signature... Cannot find it anymore, I will try to dig deeper.

M.

gkalpak: I don't mind the function signature :grin:

So, if we only go for the documented one-arg signature, what is this one meant for: `new(...value: string[]): ShellArray;`? 😕 
I think it should be removed. Do I miss something?

msanguineti: You're probably right. I'll remove the `...` parameters and then we're just left with single parameter signatures for both construct and function.

I'll get to it sometime in the next few hours.

msanguineti:
I have accidentally rebased the branch... No coffee and late nights...

Shall I close this request and make another 'clean' one?

PR #34615, opened 2019-04-10
Adding StripeOption "betas" which allows use of Stripe's Checkout Beta

itsonlycode:
Stripe has made a new checkout api that can be used if you include  betas: ['checkout_beta_4']  in StripeOptions.
 https://stripe.com/docs/payments/checkout

### DIFF
--- a/types/stripe-v3/index.d.ts
+++ b/types/stripe-v3/index.d.ts
@@ -31,6 +31,7 @@ declare namespace stripe {
 
     interface StripeOptions {
       stripeAccount?: string;
+      betas?: string[];
     }
 
     interface TokenOptions {
--- a/types/stripe-v3/stripe-v3-tests.ts
+++ b/types/stripe-v3/stripe-v3-tests.ts
@@ -6,6 +6,7 @@ declare function it(desc: string, fn: () => void): void;
 describe("Stripe", () => {
     it("should excercise all Stripe API", () => {
         const stripe = Stripe('public-key');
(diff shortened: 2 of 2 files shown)


types/stripe-v3/index.d.ts
```diff
@@ -31,6 +31,7 @@ declare namespace stripe {
 
     interface StripeOptions {
       stripeAccount?: string;
+      betas?: string[];
```
marlosin: I couldn't find docs about this property in the page you mentioned. `StripeOptions` is used as a parameter of `StripeStatic` (which is documented here https://stripe.com/docs/stripe-js/reference#stripe-function)

itsonlycode: It is found under Checkout Client Quickstart and then step 3 ('write it yourself')
https://stripe.com/docs/payments/checkout/client#integrate

itsonlycode: Had a conversation with stripe developers on freenode #stripe and it IS part of the api as it stands even though it is not yet documented on https://stripe.com/docs/stripe-js/reference#stripe-function

marlosin (approved):

uniqueiniquity (changes requested):
Since this is a popular package, could you add a brief test to avoid regressions to this change in the future? Thanks!

uniqueiniquity (approved):

uniqueiniquity:

PR #34616, opened 2019-04-10
Fix the onRangeChange event types in react-big-calendar

TomasHubelbauer:
The event argument differs based on the calendar view. It would probably be better to make the calendar props a discriminated union keyed by the view so that this event could have either of the two signatures depending on the view, but that's a lot bigger change which I am not willing to dedicate time to so in the interim I'd like to contribute a semi-fix by at least providing correct types.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: http://intljusticemission.github.io/react-big-calendar/examples/index.html#prop-onRangeChange


TomasHubelbauer:
For anyone wondering, the behavior exactly is:

| View | Type | Note |
|-|-|-|
| `day` | `Date[]` | Ar array with two elements, the date and `null` |
| `week` | `Date[]` | An array of 8 elements, 7 dates, one for each day of the week and `null` |
| `month` | `{ start: stringOrDate; end: stringOrDate }[]` | 2 elements, the object and `null` |
| `agenda` | `{ start: stringOrDate; end: stringOrDate }[]` | 2 elements, the object and `null` |

I have no idea why the `null` gets appended to all of the arrays… Also the `day` view date seems to be a day late but alas.

eps1lon (approved):
> I have no idea why the null gets appended to all of the arrays

It's a null-terminated list :smile: 

I guess we should recommend to destructure the individual dates instead of iterating over them.

TomasHubelbauer:
> I guess we should recommend to destructure the individual dates instead of iterating over them.

Hmmm, now thinking about this and making the props a union, we could get something like:

```ts
type Props =
  | { view: 'day', onRangeChange: (date: [Date]) => void; } & OtherProps
  | { view: 'week', onRangeChange: (dates: [Date, Date, Date, Date, Date, Date, Date]) => void; } & OtherProps
  | { view: 'month', onRangeChange: (range: [{ start: stringOrDate, end: stringOrDate }]) => void; } & OtherProps
  | { view: 'agenda', onRangeChange: (date: [{ start: stringOrDate, end: stringOrDate }]) => void; } & OtherProps
```

This could be _really_ nice, what do you think about it @eps1lon? I might take some time to do this after all.

eps1lon:
> This could be really nice, what do you think about it @eps1lon? I might take some time to do this after all.

Sounds perfect. This way we could use tuples instead.

TomasHubelbauer:
I am thinking something like this:

```ts
export type BigCalendarProps<TEvent extends Event = Event, TResource extends object = object> = React.Props<BigCalendar<TEvent, TResource>> & {
    localizer: DateLocalizer;

    date?: stringOrDate;
    getNow?: () => Date;
    events?: TEvent[];
    onNavigate?: (newDate: Date, view: View, action: Navigate) => void;
    onView?: (view: View) => void;
    onDrillDown?: (date: Date, view: View) => void;
    onSelectSlot?: (slotInfo: { start: stringOrDate, end: stringOrDate, slots: Date[] | string[], action: 'select' | 'click' | 'doubleClick' }) => void;
    onDoubleClickEvent?: (event: TEvent, e: React.SyntheticEvent<HTMLElement>) => void;
    onSelectEvent?: (event: TEvent, e: React.SyntheticEvent<HTMLElement>) => void;
    onSelecting?: (range: { start: stringOrDate, end: stringOrDate }) => boolean | undefined | null;
    selected?: any;
    views?: Views;
    drilldownView?: View | null;
    getDrilldownView?: ((targetDate: Date, currentViewName: View, configuredViewNames: View[]) => void) | null;
    length?: number;
    toolbar?: boolean;
    popup?: boolean;
    popupOffset?: number | { x: number, y: number };
    selectable?: boolean | 'ignoreEvents';
    longPressThreshold?: number;
    step?: number;
    timeslots?: number;
    rtl?: boolean;
    eventPropGetter?: EventPropGetter<TEvent>;
    slotPropGetter?: SlotPropGetter;
    dayPropGetter?: DayPropGetter;
    showMultiDayTimes?: boolean;
    min?: stringOrDate;
    max?: stringOrDate;
    scrollToTime?: Date;
    culture?: string;
    formats?: Formats;
    components?: Components<TEvent>;
    messages?: Messages;
    titleAccessor?: keyof TEvent | ((event: TEvent) => string);
    allDayAccessor?: keyof TEvent | ((event: TEvent) => boolean);
    startAccessor?: keyof TEvent | ((event: TEvent) => Date);
    endAccessor?: keyof TEvent | ((event: TEvent) => Date);
    resourceAccessor?: keyof TEvent | ((event: TEvent) => any);
    resources?: TResource[];
    resourceIdAccessor?: keyof TResource | ((resource: TResource) => any);
    resourceTitleAccessor?: keyof TResource | ((resource: TResource) => string);
    defaultView?: View;
    defaultDate?: Date;
    className?: string;
    elementProps?: React.HTMLAttributes<HTMLElement>;
} & (
    | { view: 'day'; onRangeChange?: (date: [Date, null]) => void; }
    | { view: 'week'; onRangeChange?: (date: [Date, Date, Date, Date, Date, Date, Date, null]) => void; }
    | { view: 'work_week'; onRangeChange?: (date: [Date, Date, Date, Date, Date, null]) => void; }
    | { view: 'month'; onRangeChange?: (range: [{ start: stringOrDate; end: stringOrDate; }, null]) => void; }
    | { view: 'agenda'; onRangeChange?: (range: [{ start: stringOrDate; end: stringOrDate; }, null]) => void; }
);
```

I ran into a problem though - the `onRangeChange` callback is not dependent on `view` only but also `defaultView`. So the union with have to essentially have two discriminators.

- Either you added view and then the `onRangeChange` callback has one of 5 forms per the value
- Or you added defaultView and then the `onRangeChange` callback has one of 5 forms per the value
- Or neither and then week is assumed (default prop) and `onRangeChange` has only 1 possible form

I would know how to fix this if it was only a single prop, but with two I am not sure how to go about it, ideas?

eps1lon:
Ah no that doesn't work. `onRangeChange` needs to process all possible inputs since `view` can change. There are basically `>|views|!` possible combinations of argument types `onRangeChange` can have (and the you need to consider uncontrolled vs controlled behavior). I would stick with the initial approach.

TomasHubelbauer:
Are there any shortcomings to this?
```ts
& (
        | { view: 'day'; defaultView?: undefined; onRangeChange?: (date: [Date, null]) => void; }
        | { view: 'week'; defaultView?: undefined; onRangeChange?: (date: [Date, Date, Date, Date, Date, Date, Date, null]) => void; }
        | { view: 'work_week'; defaultView?: undefined; onRangeChange?: (date: [Date, Date, Date, Date, Date, null]) => void; }
        | { view: 'month'; defaultView?: undefined; onRangeChange?: (range: [{ start: stringOrDate; end: stringOrDate; }, null]) => void; }
        | { view: 'agenda'; defaultView?: undefined; onRangeChange?: (range: [{ start: stringOrDate; end: stringOrDate; }, null]) => void; }
        | { defaultView: 'day'; view?: undefined; onRangeChange?: (date: [Date, null]) => void; }
        | { defaultView: 'week'; view?: undefined; onRangeChange?: (date: [Date, Date, Date, Date, Date, Date, Date, null]) => void; }
        | { defaultView: 'work_week'; view?: undefined; onRangeChange?: (date: [Date, Date, Date, Date, Date, null]) => void; }
        | { defaultView: 'month'; view?: undefined; onRangeChange?: (range: [{ start: stringOrDate; end: stringOrDate; }, null]) => void; }
        | { defaultView: 'agenda'; view?: undefined; onRangeChange?: (range: [{ start: stringOrDate; end: stringOrDate; }, null]) => void; }
        | { view?: undefined; defaultView?: undefined; onRangeChange?: (date: [Date, Date, Date, Date, Date, Date, Date, null]) => void; }
    );
```

I gave it more thought and I came up with this which I think covers all cases.

We could go as far as to say `onRangeChange?: ([sunday, monday, tuesday, wednesday, thursday, friday, saturday]: [Date, Date, Date, Date, Date, Date, Date, null]) => void;` instead of `onRangeChange?: (date: [Date, Date, Date, Date, Date, Date, Date, null]) => void;` I think because the calendar always uses Sunday as the first day of the week in this callback I think. Would need to verify that.

TomasHubelbauer:
Actually just as I asked it I realized, yes, there are. The user would always have to use a constant string for these values - so like you said - controlled vs. uncontrolled etc. Okay, ditching this idea. Will fix Travis and stick with the current diff.

eps1lon:
@TomasHubelbauer Do you want to add yourself to the header? Looks like you made plenty of contributions to these types already. Would fastrack your patches if the PR is not reviewed by the other owners.

TomasHubelbauer:
Good idea, added myself.

PR #34621, opened 2019-04-10
loadable__server - add attributes to the get functions

luisherranz:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.smooth-code.com/open-source/loadable-components/docs/api-loadable-server/

---- 

`@loadable/server` accepts an optional object with attributes in their `getScriptTags`, `getScriptElements`, `getLinkTags`, `getLinkElements`, `getStyleTags` and `getStyleElements` now:
https://www.smooth-code.com/open-source/loadable-components/docs/api-loadable-server/

### DIFF
--- a/types/loadable__server/index.d.ts
+++ b/types/loadable__server/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for @loadable/server 5.8.0
+// Type definitions for @loadable/server 5.8
 // Project: https://github.com/smooth-code/loadable-components
 // Definitions by: Martynas Kadiša <https://github.com/martynaskadisa>
 //                 Luis Herranz <https://github.com/luisherranz>
--- a/types/loadable__server/index.d.ts
+++ b/types/loadable__server/index.d.ts
@@ -29,6 +29,24 @@ export type ChunkExtractorOptions = {
  });
 
 /**
+ * Chunk that is received by the AttrFn function.
+ */
+export interface Chunk {
+	chunk: string;
+	filename: string;
+	linkType: string;
+	path: string;
+	scriptType: string;
+	type: string;
+	url: string;
+}
+
+/**
+ * Used to insert attributes in the get functions.
+ */
+export type AttrFn = (chunk: Chunk) => {};
+
+/**
  * Used to collect chunks server-side and get them as script tags or script elements
  */
 export class ChunkExtractor {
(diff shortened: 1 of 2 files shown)


types/loadable__server/index.d.ts
```diff
@@ -51,32 +51,32 @@ export class ChunkExtractor {
 	/**
 	 * Get scripts as a string of `<script>` tags
 	 */
-	getScriptTags(): string;
+	getScriptTags(attributes?: { [key: string]: string }): string;
```
martynaskadisa: It looks like it not only can accept an object, but also a function as documented here:
https://www.smooth-code.com/open-source/loadable-components/docs/api-loadable-server/#chunkextractorgetscripttags

Also, specifying that it's of type `{ [key: string]: string }` seems incorrect. I don't see where it enforces that property value should be a `string`. I think this can be just `object`.

martynaskadisa (changes requested):

types/loadable__server/index.d.ts
luisherranz: Hey @martynaskadisa, thanks for reviewing the PR.

> It looks like it not only can accept an object, but also a function as documented here

Yeah, I saw that but the fact is... the docs are wrong, the code is not there. Those functions don't accept an `attrFn` yet. For that reason I didn't add it. Maybe we should ping them to explain.

> Also, specifying that it's of type { [key: string]: string } seems incorrect. I don't see where it enforces that property value should be a string. I think this can be just object.

Well, those attributes are going to end up being strings in the dom, so I guess it is good to have them as strings in typescript.

cc: @jsfez @neoziro @rhy0u could you please take a look at these two issues?

martynaskadisa: > Yeah, I saw that but the fact is... the docs are wrong, the code is not there. Those functions don't accept an attrFn yet. For that reason I didn't add it. Maybe we should ping them to explain.

Ahh, I see. It seems this behaviour has only landed in master and has not been released yet https://github.com/smooth-code/loadable-components/pull/277.

luisherranz: Oh, good catch! Let's wait until they release it then.

Regarding the other one, let's see if they can chime in here as well.

neoziro: https://github.com/smooth-code/loadable-components/releases/tag/v5.8.0

I trust you guys for typings!

luisherranz: Ok, thanks @neoziro!

@martynaskadisa I've made another push with `{}` instead of `{ [key: string]: string }` and `attrFn`.

martynaskadisa: @luisherranz Awesome!

types/loadable__server/index.d.ts
```diff
@@ -1,6 +1,7 @@
-// Type definitions for @loadable/server 5.2
+// Type definitions for @loadable/server 5.3
```
martynaskadisa: Could you also change this to the newest version of the library which is 5.8?
https://github.com/smooth-code/loadable-components/releases/tag/v5.8.0

martynaskadisa (changes requested):
Thanks for the contribution, the only small part that should be updated is the version for which these typings are for. Other than that, looks good for me, good work 😄

types/loadable__server/index.d.ts
luisherranz: Is it `5.8.0` or `5.8`?

luisherranz: `5.8` :)

martynaskadisa (approved):

uniqueiniquity:

PR #34636, opened 2019-04-10
Add type definitions for package ringbufferjs

amchelle:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/ringbufferjs/index.d.ts
+++ b/types/ringbufferjs/index.d.ts
@@ -1,0 +1,26 @@
+// Type definitions for ringbufferjs 1.1
+// Project: https://github.com/janogonzalez/ringbufferjs#readme
+// Definitions by: Amchelle Clendenin <https://github.com/amchelle>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export = RingBuffer;
+
+declare class RingBuffer<T> {
+    _elements: T[];
+    _first: number;
+    _last: number;
+    _size: number;
+    _evictedCb: (element: T) => any;
+
+    constructor(capacity: number, evictedCB?: (element: T) => any);
+
+    capacity(): number;
+    isEmpty(): boolean;
+    isFull(): boolean;
+    peek(): T;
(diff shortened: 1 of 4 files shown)


types/ringbufferjs/index.d.ts
```diff
@@ -0,0 +1,24 @@
+// Type definitions for ringbufferjs 1.1
+// Project: https://github.com/janogonzalez/ringbufferjs#readme
+// Definitions by: Amchelle Clendenin <https://github.com/amchelle>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export default class RingBuffer<T> {
```
uniqueiniquity: It seems like the [underlying library](https://unpkg.com/ringbufferjs@1.1.0/index.js) is not using `export default`, so this .d.ts file should not either. Instead, you should follow the appropriate template from [the TypeScript handbook](http://www.typescriptlang.org/docs/handbook/declaration-files/templates.html). Note that if you want to use `import name from 'package';` in your consuming code, you can use the `esModuleInterop` compiler setting.

uniqueiniquity (changes requested):

types/ringbufferjs/index.d.ts
amchelle: Ok thanks for the tip! Let me take a look

amchelle: I removed the `default export`, please let me know if there is anything else

uniqueiniquity: I think you actually want to use [this template](http://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-class-d-ts.html).

The way you have the declaration file right now, and shown in the tests, will fail at runtime.

amchelle: Thank you for pointing me to the right template! I think I found the required reading on `esModuleInterop`. Third time's the charm

uniqueiniquity (approved):

uniqueiniquity:

PR #34637, opened 2019-04-10
react-sketchapp: added types for Svg

saschazar21:
This PR adds types for react-sketchapp's [Svg component](https://airbnb.io/react-sketchapp/docs/API.html#svg).

The types were added according the folder structure, so it's possible to directly import each `Svg` component directly from its source file, e.g:

```jsx
import { Svg } from 'react-sketchapp';
import Circle from 'react-sketchapp/lib/components/Svg/Circle';

export const MyCircleComponent = (props: YourProps) => (
  <Svg viewBox="0 0 32 32">
    <Circle cx="16" cy="16" r="12" />
  </Svg>
);
```

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [React Native's Svg documentation](https://github.com/react-native-community/react-native-svg#supported-elements) - together with react-sketchapp's source code.
- [] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.


types/react-sketchapp/index.d.ts
```diff
@@ -1,10 +1,10 @@
-// Type definitions for react-sketchapp 0.13
+// Type definitions for react-sketchapp 0.15
 // Project: https://github.com/airbnb/react-sketchapp
 // Definitions by: Rico Kahler <https://github.com/ricokahler>
 //                 DomiR <https://github.com/DomiR>
 //                 Sascha Zarhuber <https://github.com/saschazar21>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.3
```
uniqueiniquity: What change caused the minimum compatible version to become 3.3?
When testing locally, it seems like 2.8 still works fine.

types/react-sketchapp/tsconfig.json
```diff
@@ -1,25 +1,19 @@
 {
     "compilerOptions": {
+        "allowSyntheticDefaultImports": true,
```
uniqueiniquity: Also, what change made this change necessary? It also seems to work fine without this locally.
This can cause issues, since consumers of this library may not have the flag enabled (i.e. the compiler option only applies within this package, not to consumers of it).

uniqueiniquity (changes requested):

types/react-sketchapp/index.d.ts
saschazar21: You're completely right, that was a leftover from local tests with another module.

fixed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34637/commits/dcceb8d800c34f28c81d0684d73b55d9afcc6634

types/react-sketchapp/tsconfig.json
saschazar21: Same here, had to use it for another module.

fixed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34637/commits/dcceb8d800c34f28c81d0684d73b55d9afcc6634

types/react-sketchapp/tsconfig.json
```diff
@@ -1,6 +1,5 @@
 {
     "compilerOptions": {
-        "allowSyntheticDefaultImports": true,
```
ricokahler: Does this still work? If I remember correctly, if you remove this option then you'd have to change the imports to `import * as React from 'react';` but i'm not entirely sure. Can you confirm that this works?

ricokahler: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34637/files#diff-9f40a6b1b02ac40eb8462ac5e4bfbba9R9 e.g.

saschazar21: Yes, that's why I added it in the first place, but then Travis threw an error, because of the `no-import-default-of-export-equals` rule, which is also referring to the `allowSyntheticDefaultImports` compile option.

[See docs](https://github.com/Microsoft/dtslint/blob/master/docs/no-import-default-of-export-equals.md)

uniqueiniquity (approved):

PR #34641, opened 2019-04-10
typescript@next cleanup: Deprecate obsolete DOM packages

sandersn:
Remove 3 packages that are no longer needed in TS 3.5 because of DOM updates. The packages will remain on npm, but their newest versions will be deprecated.

I had to point the three packages to Typescript as their replacement, because all 3 are non-npm packages since they were representing browser standards.


magcius:
... But TypeScript 3.5 is not out yet? Was this landed prematurely?

PR #34647, opened 2019-04-11
[mysql] Improve function documentation

alexjurkiewicz:
Please fill in this template.

- [x] Add or edit tests to reflect the change. (Run with `npm test`.) **No test changes**
```
ERROR: 216:8  unified-signatures  This overload and the one on line 208 can be combined into one signature taking `'connection' | 'acquire'`.
ERROR: 224:8  unified-signatures  This overload and the one on line 208 can be combined into one signature taking `'connection' | 'release'`.
ERROR: 224:8  unified-signatures  This overload and the one on line 216 can be combined into one signature taking `'acquire' | 'release'`.
```
**These have been split up to allow different function docs.** Is it OK to add ignore declaration for these?

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/mysqljs/mysql
- [x] Increase the version number in the header if appropriate. **No type definition changes.**
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. **Is already done**

### DIFF
--- a/types/mysql/index.d.ts
+++ b/types/mysql/index.d.ts
@@ -143,15 +143,28 @@ export interface Connection extends EscapeFunctions {
 
     on(ev: 'drain' | 'connect', callback: () => void): Connection;
 
+    /**
+     * Set handler to be run when the connection is closed.
+     */
     on(ev: 'end', callback: (err?: MysqlError) => void): Connection;
 
     on(ev: 'fields', callback: (fields: any[]) => void): Connection;
 
-    on(ev: 'error', callback: (err: MysqlError) => void): Connection;
+    /**
+     * Set handler to be run when a a fatal error occurs.
+     */
+    on(ev: 'error', callback: (err: MysqlError) => void): Pool;
 
-    on(ev: 'enqueue', callback: (...args: any[]) => void): Connection;
+    /**
+     * Set handler to be run when a callback has been queued to wait for an
+     * available connection.
+     */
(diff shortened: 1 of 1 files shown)


alexjurkiewicz:
I added a commit to ignore the lint errors. Still not sure if it's the right approach though.

types/mysql/index.d.ts
```diff
@@ -535,37 +667,97 @@ export interface MysqlError extends Error {
 }
 
 export const enum Types {
```
kpping: On this `Enum: Types`, I think we should leave it as is until we can precisely document them. What do you think ?

kpping (commented):
- I think the return type (in the comments) is incorrect.
- For `Enum: Types` part, I think leave it as is might be better. What do you think ?
- I think `tslint:disable-next-line:unified-signatures` is acceptable in this case.

Thank you for your contribution.

types/mysql/index.d.ts
alexjurkiewicz: that's fine with me

alexjurkiewicz:
Thank you for the review :tada: . I removed the Enum type conversion as you suggested, and fixed the three errors.

kpping (approved):
Thank you for your contribution. 🎉

PranavSenthilnathan:

PR #34648, opened 2019-04-11
[natural] updated classifiers classification definition

ybentz:
This PR adds a more accurate definition for the Bayesian and logistic regression classifiers' classification entry, it's currently defined as `string[]` but it's actually an array of a specific object type `{ label: string, value: number }`.

Side note: this is my first time contributing to DT so I hope that I updated the tests in a satisfying manner.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/NaturalNode/natural#bayesian-and-logistic-regression
- [ ] If you are making substantial changes, consider adding a tslint.json containing { "extends": "dtslint/dt.json" }.

### DIFF
--- a/types/natural/index.d.ts
+++ b/types/natural/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for Natural v0.6.3
+// Type definitions for Natural 0.6
 // Project: https://github.com/NaturalNode/natural
 // Definitions by: Dylan R. E. Moonfire <https://github.com/dmoonfire>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
--- a/types/natural/index.d.ts
+++ b/types/natural/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for Natural 0.2.2
+// Type definitions for Natural v0.6.3
 // Project: https://github.com/NaturalNode/natural
 // Definitions by: Dylan R. E. Moonfire <https://github.com/dmoonfire>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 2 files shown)


dmoonfire (approved):
Looks good, thank you.

types/natural/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for Natural 0.2.2
+// Type definitions for Natural v0.6.3
```
uniqueiniquity: This appears to be a malformed header. Could you remove the line in the `tslint.json` file in this folder that says `dt-header: false` and fix any errors that come up? You will want this first line to say `Type definitions for natural 0.6`. You are welcome to remove any other such line from the `tslint.json` file, though that might be a lot of work depending on what issues are being ignored. Thanks!

uniqueiniquity (changes requested):

types/natural/index.d.ts
ybentz: @uniqueiniquity thanks for the feedback! I pushed a fix for the header error you mentioned + another error that came up for headers in the test file.

1 thing to note is that when editing this specific header I was following the instructions [here](http://definitelytyped.org/guides/contributing.html) which seem official but I guess are outdated. The error message does send you to the right place ([here](https://github.com/Microsoft/dtslint/blob/master/docs/dt-header.md)) but I didn't see a link to this page anywhere in the contribution guides which I tried to follow as best I could, so for future new contributors this should probably be updated in the docs.

uniqueiniquity: Thanks for pointing that out! I've never actually seen the page that you linked, but I'll try to get that updated.

uniqueiniquity (approved):

types/natural/index.d.ts
ybentz: Of course.
I'm wondering how this didn't come up yet, the first 4 results when googling "definitelytyped contribute" point to this guide.
Looking at http://definitelytyped.org seems like it's not very up to date in general, not sure who's maintaining it but the readme for its [repo](https://github.com/DefinitelyTyped/definitelytyped.github.io) also has broken links to the guides under "Edit online".

PR #34653, opened 2019-04-11
orientjs update definition file and update generic in orientjs statements

saeedtabrizi:


saeedtabrizi:
Great , after some tries to pass testing and linting the new definition file is Ready and be OK.
Thanks

PR #34657, opened 2019-04-11
auth0: Expose more of the Job resource in management client

pacey:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://auth0.com/docs/api/management/v2#!/Jobs/get_jobs_by_id

### DIFF
--- a/types/auth0/index.d.ts
+++ b/types/auth0/index.d.ts
@@ -867,8 +867,8 @@ export class ManagementClient {
 
 
   // Jobs
-    getJob<T extends Job>(params: ObjectWithId): Promise<T>;
-    getJob<T extends Job>(params: ObjectWithId, cb?: (err: Error, data: T) => void): void;
+    getJob(params: ObjectWithId): Promise<Job>;
+    getJob(params: ObjectWithId, cb?: (err: Error, data: Job) => void): void;
 
     importUsers(data: ImportUsersOptions): Promise<ImportUsersJob>;
     importUsers(data: ImportUsersOptions, cb?: (err: Error, data: ImportUsersJob) => void): void;
--- a/types/auth0/index.d.ts
+++ b/types/auth0/index.d.ts
@@ -2,7 +2,7 @@
 // Project: https://github.com/auth0/node-auth0
 // Definitions by: Seth Westphal <https://github.com/westy92>
 //                 Ian Howe <https://github.com/ianhowe76>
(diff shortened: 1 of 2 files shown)


pacey:
The version in the definition header seems to have fallen out of sync. There is 2.9.3 in Git, but 2.9.10 is published to NPM. I can update this to 2.9.11 if required?

types/auth0/index.d.ts
```diff
@@ -509,13 +509,64 @@ export interface StatsParams {
   to: string;
 }
 
+export type Job = ImportUsersJob | ExportUsersJob | VerificationEmailJob;
+
+export type JobFormat = 'csv' | 'json';
+
+export type JobStatus = 'pending' | 'completed';
```
pacey: I would assume there is another status for "running" or "in_progress", but I've never seen it and I can't find anything in the Auth0 docs to suggest it exists/what the value is

pacey: I found "processing"and "failed" mentioned here https://community.auth0.com/t/management-api-we-need-better-docs/18910/2

armanio123:
> The version in the definition header seems to have fallen out of sync. There is 2.9.3 in Git, but 2.9.10 is published to NPM. I can update this to 2.9.11 if required?

DefinitelyTyped uses the patch version itself. It will auto-increment the patch version for every update of the same minor version, so it's impossible to have types that target only a patch version. They must work for all patches of a minor version.

types/auth0/index.d.ts
```diff
@@ -509,13 +509,64 @@ export interface StatsParams {
...
+
 export interface ImportUsersOptions {
-  connection_id: string;
-  users: string;
+    users: string;
+    connection_id: string;
+    upsert?: boolean;
+    external_id?: string;
+    send_completion_email?: boolean;
 }
```
armanio123: Changing the interface name from `UserIdParams` to `VerificationEmailOptions` will break users using this type. I don't think that's an intended behavior. I suggest changing back the name unless you feel this is a different interface that needs to be fixed.

armanio123 (changes requested):

pacey:
Thanks for having a look @armanio123, I'll have a look when I get into work

types/auth0/index.d.ts
pacey: The name change has been reverted

armanio123 (approved):

PR #34670, opened 2019-04-11
Webpack v3 test and docs use options as a string, yet don't declare it in the types

weswigham:
Confirmed in the [v3 docs](https://github.com/webpack/webpack.js.org/blob/v3.11.0/src/content/configuration/module.md#useentry).

Found while working on https://github.com/Microsoft/TypeScript/pull/30853.


johnnyreilly:
Thanks!

PR #34672, opened 2019-04-11
[expo] re-export types from RN SVG

SimenB:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/react-native-community/react-native-svg/blob/3e4505caa484188eb4dcc94e57011be9bdd0ce94/index.d.ts

`react-native-svg` comes with its own extensive typings, and the one currently in DT has a bunch of holes (I've sent 2 PRs recently, and my project still has a bunch of other errors). Re-using their typings seems ideal (and fixes all of the type errors in the SVGs in my project).

I'm not sure if the lost props exports are wanted?

### DIFF
--- a/types/expo/expo-tests.tsx
+++ b/types/expo/expo-tests.tsx
@@ -595,7 +595,7 @@ const allSecureStoreKeychainAccessibleValues: number[] = [
         </Svg.Defs>
         <Svg.G transform="translate(0, 0)" y={20}>
             <Svg.Text fill="blue" transform={{ translateX: 0, translateY: 0 }}>
-                <Svg.TextPath href="#path" startOffset="-10%">
+                <Svg.TextPath href="#path" startOffset="-10%" midLine="smooth">
                     We go up and down,
                     <Svg.TSpan fill="red" dy="5,5,5">then up again</Svg.TSpan>
                 </Svg.TextPath>
@@ -609,7 +609,7 @@ const allSecureStoreKeychainAccessibleValues: number[] = [
         </Svg.G>
         <Svg.Use href="#shape" transform="translate(0, 0)" x="20" y="0" />
         <Svg.Use href="#shape" transform={{ translateX: 0, translateY: 0 }} x="20" y="0" width="20" height="20"/>
-        <Svg.Symbol id="symbol" viewBox="0 0 150 110" width="100" height="50">
+        <Svg.Symbol id="symbol" viewBox="0 0 150 110">
             <Svg.Circle cx="50" cy="50" r="40" strokeWidth="8" stroke="red" fill="red"/>
             <Svg.Circle cx="90" cy="60" r="40" strokeWidth="8" stroke="green" fill="white"/>
         </Svg.Symbol>
--- a/types/expo/index.d.ts
+++ b/types/expo/index.d.ts
@@ -2610,146 +2610,30 @@ export namespace SQLite {
 /**
  * Svg
  */
-export interface SvgCommonProps {
-    fill?: string;
-    fillOpacity?: number | string;
-    fillRule?: 'nonzero' | 'evenodd';
-    opacity?: number | string;
-    stroke?: string;
-    strokeWidth?: number | string;
-    strokeOpacity?: number | string;
-    strokeLinecap?: string;
-    strokeLinejoin?: string;
-    strokeDasharray?: any[];
-    strokeDashoffset?: any;
-    transform?: string | object;
-    x?: number | string;
-    y?: number | string;
-    rotate?: number | string;
-    rotation?: number | string;
-    scale?: number | string;
-    origin?: number | string;
-    originX?: number | string;
-    originY?: number | string;
-    id?: string;
-    disabled?: boolean;
-    onPress?: () => any;
-    onPressIn?: () => any;
-    onPressOut?: () => any;
-    onLongPress?: () => any;
-    delayPressIn?: number;
-    delayPressOut?: number;
-    delayLongPress?: number;
-}
+import * as RNSvg from 'react-native-svg';
 
-export interface SvgRectProps extends SvgCommonProps {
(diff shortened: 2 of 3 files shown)


types/expo/expo-tests.tsx
```diff
@@ -609,7 +609,7 @@ const allSecureStoreKeychainAccessibleValues: number[] = [
         </Svg.G>
         <Svg.Use href="#shape" transform="translate(0, 0)" x="20" y="0" />
         <Svg.Use href="#shape" transform={{ translateX: 0, translateY: 0 }} x="20" y="0" width="20" height="20"/>
-        <Svg.Symbol id="symbol" viewBox="0 0 150 110" width="100" height="50">
```
SimenB: not sure about this one. `symbol` seems to support `width` and `height`: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/symbol

/cc @msand

msand: @SimenB It seems the geometry properties x, y, width and height were added in Svg 2.0: https://svgwg.org/svg2-draft/struct.html#SymbolElement
react-native-svg doesn't have support for it yet at least, might be better to wait until it's implemented before adding it to the types. As otherwise people might get surprised when the types suggest it's available, but it has no visible impact. And, if it gets implemented later on, then they would likely get a different output once they upgrade.

SimenB: That makes sense, thanks!

armanio123:
@SimenB @msand Based on your comments, I'll wait for merging. Let me know when this is ready.

types/expo/expo-tests.tsx
janaagaard75: @msand: So SimonB did the right thing, no? And so we should give this a thumbs up, correct? Or did I miss something?

janaagaard75 (approved):

types/expo/expo-tests.tsx
SimenB: Yeah, this is resolved, I just kept the conversation around instead of hitting "resolve" so reviewers could see this

armanio123:
Thanks for confirming. Merging now.

ankhzet:
Shouldn't it point to a specific version of `react-native-svg`? As of now, `"react-native-svg": "*"` require results in '9.4.0' of react-native-svg, while expo 32.0.6 (latest available) points to '8.0.10', which causes duplication and (AFAIKT) double registration of the module ("Tried to register two views with the same name RNSVGRect" error in the console).

Fixed by adding strict resolution to package.json, but this should not be required from consumer of this package, imo:
```
  "resolutions": {
    "react-native-svg": "8.0.10"
  }
```

jaulz:
@ankhzet good catch! Feel free to open a PR and fix it.

burtek:
@ankhzet true! Actually I was also thinking about peerDependency, not sure how that's gonna work...|

Anyways, would you create a PR?

SimenB:
It's just a type import, why should it say `Tried to register two views with the same name RNSVGRect`? This code should never be actually loaded, it's just used for type checking.

However, I do agree it should match https://github.com/expo/expo/blob/13bf3ff0ad308aab33fee4f0cf0367422664d5f1/packages/expo/package.json#L143, I just don't understand why it causes runtime issues

ankhzet:
@SimenB , i'm not sure what kind of wizardry it is either.
Could it be caused by other module (namely, `react-native-qrcode-svg`), that is peer-dependant on yet another version of `"react-native-svg": "^6.5.2"`, loading the 'react-native-svg' module, but for some reason - wrong version of it? F.e. the '9.4.0' version (is mentioned in yarn.lock and resolved from @types/expo) is loaded by `react-native-qrcode-svg`, while '8.0.10' (resolved from 'expo') is loaded by `expo` itself, thus - duplication?

Also, react-native-maps also got duplicated (but doesn't cause conflicts, probably, because in runtime it is used only by `expo`):
```
yarn why react-native-maps
=> Found "react-native-maps@0.24.2"
info Reasons this module exists
   - "@types#expo" depends on it
   - Hoisted from "@types#expo#react-native-maps"
info Number of shared dependencies: 0
=> Found "expo#react-native-maps@0.22.1"
info This module exists because "expo" depends on it.
info Number of shared dependencies: 0
```

ankhzet:
@jaulz, @burtek tldr: I don't feel qualified for making a PR a.t.m. 
I'm not sure I can dedicate sufficient amount of time for making a PR properly. Knowing myself, I would eventually dig in on the best practices for creating typedef repos and that may/would delay PR infinitely with my current lack of free time %).

@burtek, your 'peerDependency' idea seems reasonable to me, but that also raises the question, should or should not 'react-native-maps' and others also be peer dependencies, and should or shouldn't `expo` also be a peer dependency, and to which version it is better to lock to, etc.

jaulz:
@ankhzet just change it [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/expo/package.json) and follow the respective descriptions in the issue template. It shouldn't take more than 10 minutes and it's a good way to give something back to the community.

ankhzet:
@jaulz "DefinitelyTyped/types/expo/package.json should not include field peerDependencies"
Related issue: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/20290

Also, don't see how dependencies can be locked on 'per-typedef-package-version' basis.
Looks like this requires some more research.

P.S. @jaulz, and I believed it would not take more than 10 minutes... xD

jaulz:
@ankhzet only dependencies are allowed (which is also what we want here, since we talking about types only) and you can refer to an older version like this: 
- https://github.com/DefinitelyTyped/DefinitelyTyped#i-notice-some-packages-having-a-packagejson-here
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/pikaday/package.json

PR #34678, opened 2019-04-12
[NodeJS] Support 'typesVersions' in node/v6 through node/v9

rbuckton:
Similar to the recent change to split NodeJS v11 and v10 to support `"typesVersions"`, this does the same for NodeJS v9 and v8. This change will be necessary to continue to support Node v8 and v9 once https://github.com/Microsoft/TypeScript/pull/30790 has merged.


rbuckton:
@sandersn: Any idea what's going on [here](https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/519043301#L1244)? types-publisher is complaining about a package called `aframe` being added to `notNeededPackages.json`, but I don't see that package listed.

SimonSchick:
@rbuckton with the size of node typings growing continuously, are there any plans to purge the EOL versions <= 6 (end of april) or even move them into a standard TS lib?

sandersn:
@rbuckton The diff with master is quite complicated in that run. aframe shows up as partly deleted, which is what causes the error, but three (which aframe depends on) shows up as *added*. This makes me think that the branch must be fairly old (pre-March 27). I think that the notNeededPackages code I added recently gets confused when a popular package, with dependents, gets removed, causing dependents to add a package.json. 

I would merge from master as a workaround.

rbuckton:
@sandersn: Thanks, rebasing with latest master seems to have fixed the issue.

rbuckton:
Since the bot things this is a new definition, I'll notify the reviewers manually:

---
🔔 @Microsoft @DefinitelyTyped @jkomyno @a-tarasyuk @alvis @r3nya @btoueg @brunoscheufler @smac89 @tellnes @touffy @DeividasBakanas @eyqs @Flarna @Hannes-Magnusson-CK @KSXGitHub @hoo29 @kjin @ajafff @islishude @mwiktorczyk @matthieusieben @mohsen1 @n-e @octo-sniffle @parambirs @eps1lon @SimonSchick @ThomasdenH @WilcoBakker @wwwy3y3 @ZaneHannanAU @jeremiergz @samuela @kuehlein @j-oliveras @bhongy - please [review this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34678/files) in the next few days. Be sure to explicitly select **`Approve`** or **`Request Changes`** in the GitHub UI.

Flarna:
Node.Js 9 is EOL since 2018-06-30, long before Node.Js 6 (2019-04-30). Therefore I don't think changes v9 are needed.
We should consider to remove EOL versions from the repo instead to avoid references to it and people modifying it. Or at least mark them as frozen.

SimonSchick:
@Flarna I think we should remove them, the node team doesn't maintain them and neither should we, people can always augment their typings.

rbuckton:
The split of Node v8 and Node v9 typings has less to do with maintainability, and more to do with supporting the `"typesVersions"` redirection so that an upcoming change in the TypeScript compiler will not result in unresolvable build breaks.

This has more to do with ensuring that the next release of TypeScript does not result in major breaks for anyone still using Node v9, as module augmentations *cannot* address the issue this is meant to solve. An augmentation cannot address the "Duplicate identifier" error that would result from having conflicting `interface IteratorResult<T>` and `type IteratorResult<T>` definitions, which is why we are employing `"typesVersions"` here. Even if we were to remove the definitions for Node v9, the existing previously published definitions would still be available and would run into this issue. I discussed this with @RyanCavanaugh and we believe this is a fix we should still take despite the EOL status of Node v9.

SimonSchick:
@rbuckton I think our comments were a little off-topic and we weren't arguing the actual changes made here but rather making a note that we should remove them to avoid future churn :)

rbuckton:
Hopefully the changes we've made to the Node definitions for `"typesVersions"` are comprehensive enough to avoid collisions with ECMAScript standard library definitions for the foreseeable future.

Flarna:
But why not v6 and v4 then? It's more likely that users are still on v6/v4 then on v9 as they were LTS versions.

rbuckton:
@Flarna: v6 and v4 didn't show up with errors in our DT tests for the compiler, although v7 did. I will probably need to consider including the same changes to support v7 as well.

@weswigham: It seems our dt test coverage doesn't stretch back to v6, v4, or v0. Is this intentional?

@RyanCavanaugh, @DanielRosenwasser: Any thoughts on how far back we should go to address this issue?

rbuckton:
After some review:
* `node/v0` succeeds because it does not contain a forward references for `Symbol` or `IteratorResult`.
* `node/v4` succeeds because it does not contain a forward references for `Symbol` or `IteratorResult`.
* `node/v6` succeeds because it specifies `"lib": ["es5"]` in its `tsconfig.json`. However, it would fail if referenced in a project that targets `es2015` or later (or references `es2015` libs) as it contains forward references for `Symbol` and `IteratorResult`.
* `node/v7` through `node/v9` fail because they specify `"lib": ["es6"]` in its tsconfig.json and contain forward references for `Symbol` and `IteratorResult`.

It seems for proper backwards compatibility, we need to correct `node/v6` through `node/v9`.

rbuckton:
I've split up node/v7 and node/v6 as well to align with the other node versions. This restructuring is not necessary for node/v4 and node/v0 at this time.

rbuckton:
The split seems to be causing an OOM error in the travis push build. This may be due to the larger number of files, so I am going to revert the split but preserve the `"typesVersions"` support to see if that addresses the issue for now.

package.json
```diff
@@ -18,7 +18,7 @@
         "compile-scripts": "tsc -p scripts",
         "not-needed": "node scripts/not-needed.js",
         "update-codeowners": "node scripts/update-codeowners.js",
-        "test": "node node_modules/types-publisher/bin/tester/test.js --run-from-definitely-typed",
+        "test": "node --max_old_space_size=4096 node_modules/types-publisher/bin/tester/test.js --run-from-definitely-typed",
```
Flarna: Be careful with such changes. This gives 4GB to each node process but the travis CI machines most likely don't have that much. A while ago I reduced the number of parallel processes as the sum of all processes causes OOM.

Flarna:
I think it's fine to keep the one file solution. There are not much changes expected so doing them in two files (most likely copy paste) should be no issue.
Splitting the files was good for reusing but review/search is harder now.

package.json
rbuckton: I just made a [change](https://github.com/Microsoft/types-publisher/pull/607) to make this unnecessary and will be reverting this change.

rbuckton:
@Flarna or others, can you take a look? I'd like to merge this if possible as changes continue to be made to the v6 and v8 definitions.

types/node/v6/index.d.ts
```diff
@@ -9,4457 +9,40 @@
...
-}
+// NOTE: TypeScript version-specific augmentations can be found in the following paths:
+//          - ~/base.d.ts         - Shared definitions common to all TypeScript versions
+//          - ~/index.d.ts        - Definitions specific to TypeScript 2.1
+//          - ~/ts3.1/index.d.ts  - Definitions specific to TypeScript 3.1
 
-interface Error {
-    stack?: string;
-}
+// NOTE: Augmentations for TypeScript 3.1 and later should use individual files for overrides
```
Flarna: actually 3.2

types/node/v6/tsconfig.json
```diff
@@ -2,7 +2,7 @@
     "compilerOptions": {
         "module": "commonjs",
         "lib": [
-            "es5"
+            "es6"
```
Flarna: why this change? I can remember about some complains that node typings don't work with lib es5, usually because of missing forward declarations. if es6 is used here this is not detected by tests

types/node/v6/index.d.ts
```diff
@@ -9,4457 +9,40 @@
...
 interface Iterator<T> {
     next(value?: any): IteratorResult<T>;
 }
-
+interface IterableIterator<T> { }
 interface IteratorResult<T> { }
 interface AsyncIterableIterator<T> {}
 interface SymbolConstructor {
     readonly iterator: symbol;
-    readonly asyncIterator: symbol;
```
Flarna: have you removed `asyncIterator` per accident here? For whatever reason it's only present in v6 so maybe it's anyway wrong.

Flarna (approved):
I haven't looked on every single line but it seems correct to me

types/node/v6/index.d.ts
rbuckton: I removed it intentionally. It seemed out of place and I verified that it does not, in fact, exist in Node v6.

sharikovvladislav:
Isn't this needed in `@types/node@10.x.x`?

Flarna:
@sharikovvladislav Try latest node 10 type definitions.

sharikovvladislav:
Latest for 10 is `10.14.22`. It it ok if I update `10.9.x` to `10.14.22`?

Flarna:
At least the split for typescript versions is included there.

sharikovvladislav:
Thank you. That helped.

PR #34680, opened 2019-04-12
Use secureOptions in request tests, not securityOptions

weswigham:
Found while working on https://github.com/Microsoft/TypeScript/pull/30853. "securityOptions" is fictional - there's no documentation reference of it anywhere.

### DIFF
--- a/types/request-promise-native/request-promise-native-tests.ts
+++ b/types/request-promise-native/request-promise-native-tests.ts
@@ -1,6 +1,7 @@
 import rpn = require('request-promise-native');
 import * as errors from 'request-promise-native/errors';
 import * as path from 'path';
+import constants = require('constants');
 
 rpn('http://www.google.com')
   .then(console.dir)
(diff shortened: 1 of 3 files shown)


mastermatt:
This set correctly in [`@types/node`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/tls.d.ts#L351). Do we know why this didn't error before? 

[https.AgentOptions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/https.d.ts#L14) -> [tls.ConnectionOptions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/tls.d.ts#L261) -> [tls.SecureContextOptions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/tls.d.ts#L351)

[Node docs](https://nodejs.org/api/https.html#https_https_request_url_options_callback)

weswigham:
Because TS didn't issue an excess property error when the target type was a union or intersection before; see linked TS issue.

mastermatt (approved):

PR #34688, opened 2019-04-12
Update sc-broker-cluster and socketcluster-server type definitions

DanielRose:

https://github.com/SocketCluster/socketcluster-server/blob/master/scserversocket.js
https://github.com/SocketCluster/sc-broker-cluster/blob/master/index.js

### DIFF
--- a/types/socketcluster-server/scserversocket.d.ts
+++ b/types/socketcluster-server/scserversocket.d.ts
@@ -29,12 +29,29 @@ declare class SCServerSocket extends Emitter {
 
     constructor(id: string, server: SCServer, socket: WebSocket);
 
+    on(event: string, listener: SCServerSocket.AnyFunction): this;
     on(event: "error", listener: (error: Error) => void): this;
     on(event: "message" | "raw", listener: (message: WebSocket.Data) => void): this;
     on(event: "connectAbort" | "disconnect" | "close", listener: (code: number, data?: any) => void): this;
     on(event: "authStateChange", listener: (stateChangeData: SCServerSocket.StateChangeData) => void): this;
     on(event: "authenticate", listener: (authToken?: SCServer.AuthToken) => void): this;
     on(event: "deauthenticate", listener: (oldToken?: SCServer.AuthToken) => void): this;
+
+    once(event: string, listener: SCServerSocket.AnyFunction): this;
+    once(event: "error", listener: (error: Error) => void): this;
+    once(event: "message" | "raw", listener: (message: WebSocket.Data) => void): this;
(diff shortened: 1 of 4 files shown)


types/socketcluster-server/scserversocket.d.ts
```diff
@@ -29,13 +29,33 @@ declare class SCServerSocket extends Emitter {
 
     constructor(id: string, server: SCServer, socket: WebSocket);
 
+    // tslint:disable-next-line ban-types
+    on(event: string, listener: Function): this;
```
uniqueiniquity: As discussed [here](https://github.com/DefinitelyTyped/DefinitelyTyped#common-mistakes), we'd strongly prefer not using the type `Function`; even (...args: any[]) => any) is preferable.

uniqueiniquity (changes requested):

uniqueiniquity:

types/socketcluster-server/scserversocket.d.ts
DanielRose: This is a copy of the definition from Emitter, which I added so the general case works (see the test). I could change it here, but even better would be if you can tell me how to write the type definition in a way that the "base" method definition works as well.

DanielRose:
> Just as a heads up, your commits don't seem to be associated with your GitHub account. While this isn't technically a problem, you might care if you want more appropriate attribution. You can either make sure your GitHub account is associated with the email address you're using for your commits, or rebase and amend your commits to fix the author name and email.
> If you don't care about any of this stuff, I can always just pull in the changes.

Ok, I'll check, but not before Monday :) Thanks!

types/socketcluster-server/scserversocket.d.ts
uniqueiniquity: The best way to think about it is that every time you see `Function`, it should really be something like `(...args: any[]) => any)`. While the former has all the same properties as the latter (and therefore has the same shape and is compatible from a type comparison perspective), it's ultimately an object type, and therefore makes type checking a little harder because it lacks things like a notion of a return type or arguments. So if you use the latter form here, it will certainly be compatible with the definition from `Emitter` and will allow any function a user ends up passing there. Would also be great to fix the definition of `Emitter`, but I don't want to cause too much scope creep for your PR.

uniqueiniquity: Note that if this feels a bit verbose, you can always introduce something like 
`type AnyFunction = (...args: any[]) => any` and use that.

uniqueiniquity (approved):

PR #34691, opened 2019-04-12
[storybook__addon-knobs] Update for v5

azmenak:
Updating for changes made in v5

https://github.com/storybooks/storybook/tree/next/addons/knobs

### DIFF
--- a/types/storybook__addon-knobs/index.d.ts
+++ b/types/storybook__addon-knobs/index.d.ts
@@ -79,7 +79,7 @@ export interface OptionsKnobOptions {
 export function optionsKnob<T>(
     label: string,
     values: {
-        [key: string]: any;
+        [key: string]: T;
     },
     defaultValue?: T,
     options?: OptionsKnobOptions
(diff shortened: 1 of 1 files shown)


joscha (commented):
I think for v5 there should be types coming with the package, are they not?

azmenak:
> I think for v5 there should be types coming with the package, are they not?

@joscha I don't think so, using `@storybook/addon-knobs@5.0.8`

I'm seeing this in VSCode

Screen Shot 2019-04-15 at 9 25 52 AM

Also checked the files installed in `node_modules`, no type definitions in there

types/storybook__addon-knobs/index.d.ts
```diff
@@ -65,6 +66,25 @@ export function array<T>(name: string, value: ReadonlyArray<T>, separator?: stri
...
+    | "check"
+    | "inline-check"
+    | "select"
+    | "multi-select";
+}
+
+export function optionsKnob<T>(
+    label: string,
+    values: {
+        [key: string]: any;
```
joscha:
```suggestion
        [key: string]: T;
```

types/storybook__addon-knobs/index.d.ts
```diff
@@ -65,6 +66,25 @@ export function array<T>(name: string, value: ReadonlyArray<T>, separator?: stri
...
+    },
+    defaultValue?: T,
```
joscha: any chance for us to make sure `defaultValue` is a value in `values`?

joscha (changes requested):
Intent LGTM, I think the values should follow the generic. The defaultValue would be an added bonus but I think it would require changes in storybook itself.

joscha (approved):

PR #34699, opened 2019-04-12
Mark parameters in knockout test as optional

weswigham:
Found while working on https://github.com/Microsoft/TypeScript/pull/30853 - the cb as written required `ComponentInfo` to be present, but the signature it's being assigned to has it as optional. Prior to the linked PR, the union with `KnockoutComponentTypes.EmptyConfig` hid this bug, as pretty much anything is assignable to the empty config.

### DIFF
--- a/types/knockout/test/index.ts
+++ b/types/knockout/test/index.ts
@@ -650,7 +650,7 @@ function test_Components() {
         ko.components.register("name", { template: "string-template", viewModel: { instance: null } });
 
         // viewModel from createViewModel factory method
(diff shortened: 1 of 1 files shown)


Retsam:
LGTM.  FYI the latest minor version of knockout packages their types, and it looks like they have the [same issue](https://github.com/knockout/knockout/blob/master/spec/types/module/test-module.ts#L691) over there.  It'd be great if you'd also open the same change as PR against their repo.

Retsam (approved):

mbest:
The problem is actually that the type definition for `createViewModel` is wrong. Knockout always passes both parameters to the function.

PR #34700, opened 2019-04-12
Remove excess proprty and TODO from inert

weswigham:
Found while working on https://github.com/Microsoft/TypeScript/pull/30853 which _will_ mark these as an error, as desired. (There's no great way to make a version-conditional `$ExpectError`, so i'm just removing the prop so DT stays clean)

### DIFF
--- a/types/hapi__inert/hapi__inert-tests.ts
+++ b/types/hapi__inert/hapi__inert-tests.ts
@@ -97,8 +97,7 @@ const provision = async () => {
                         return [''];
                     }
                     return new Error('');
-                },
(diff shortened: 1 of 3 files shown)


lenovouser:
Can you remove the property in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/345727aa17cd526b3617c1c85315f4fa4e4da7be/types/hapi__inert/hapi__inert-tests.ts#L101 too? I just copied that over today.

weswigham:
Done.

lenovouser (approved):

PR #34702, opened 2019-04-12
Remove h2o2 todo

weswigham:
Found while working on https://github.com/Microsoft/TypeScript/pull/30853 - which _does_ make this into an error, just like the TODO wants. (Without conditional $ExpectError's there's no good way to test it, though)


jasonswearingen (approved):
thank you for this, not sure how it got there.

PR #34706, opened 2019-04-12
[@types/prop-types] Add support of `as const` to oneOf

ibezkrovnyi:
### Common way before https://github.com/Microsoft/TypeScript/pull/29510 was to use `tuple` trick

```ts
export const tuple = <T extends string[]>(...args: T) => args;
const sizes = tuple('default', 'large');
export type SizeType = typeof sizes[number];

export const Button = (props: { size: SizeType }) => <div></div>;

Button.propTypes = {
  size: PropTypes.oneOf(sizes),
};
```

### After https://github.com/Microsoft/TypeScript/pull/29510 has landed we could use `as const`
```ts
const sizes = ['default', 'large'] as const;
export type SizeType = typeof sizes[number];

export const Button = (props: { size: SizeType }) => <div></div>;

Button.propTypes = {
  size: PropTypes.oneOf(sizes),   // Error: Type 'readonly ["default", "large"]' is missing the following properties from type '{}[]': pop, push, reverse, shift and 6 more. [2345]
};
```

### Some evidence
`tuple` is pretty common approach, for example, it is used in [ant.design](https://github.com/ant-design/ant-design/blob/eba49beecd9d0617b34195137ce36b5f56679a00/components/button/button.tsx#L43)

### Change
Change is to replace type `T[]` of `oneOf` parameter with `ReadonlyArray<T>`.
This change relaxes requirement for an Array.

> Also I tried `readonly T[]` and it works fine too. But this is [pretty new feature](https://github.com/Microsoft/TypeScript/pull/29435)

### Tested
This change was tested on my project and I used `as const` in part of code location, while keeping old way for another part of code locations. Both old and new ways work fine.

### DIFF
--- a/types/prop-types/index.d.ts
+++ b/types/prop-types/index.d.ts
@@ -65,7 +65,7 @@ export const element: Requireable<ReactElementLike>;
 export const symbol: Requireable<symbol>;
 export const elementType: Requireable<ReactComponentLike>;
 export function instanceOf<T>(expectedClass: new (...args: any[]) => T): Requireable<T>;
-export function oneOf<T>(types: T[]): Requireable<T>;
+export function oneOf<T>(types: ReadonlyArray<T>): Requireable<T>;
 export function oneOfType<T extends Validator<any>>(types: T[]): Requireable<NonNullable<InferType<T>>>;
 export function arrayOf<T>(type: Validator<T>): Requireable<T[]>;
 export function objectOf<T>(type: Validator<T>): Requireable<{ [K in keyof any]: T; }>;
--- a/types/prop-types/index.d.ts
+++ b/types/prop-types/index.d.ts
@@ -79,8 +79,8 @@ export function exact<P extends ValidationMap<any>>(type: P): Requireable<Requir
  * @param typeSpecs Map of name to a ReactPropType
  * @param values Runtime values that need to be type-checked
  * @param location e.g. "prop", "context", "child context"
- * @param componentName Name of the component for error messages.
- * @param getStack Returns the component stack.
+ * @param componentName Name of the component for error messages
+ * @param getStack Returns the component stack
  */
 export function checkPropTypes(typeSpecs: any, values: any, location: string, componentName: string, getStack?: () => any): void;
 


ibezkrovnyi:
Some issue with `typescript@3.5.0-dev20190412 (next)` and `recompose/recompose-test.tsx`
Reverting my change didn't help. It is probably some change introduced by new version of TypeScript compiler.

image

image

image

ibezkrovnyi:
### TypeScript
After reverting my changes (using master branch):

typescript@3.4.3: :heavy_check_mark: no issue
typescript@3.5.0-dev.20190411: :heavy_check_mark: no issue
typescript@3.5.0-dev.20190412: :heavy_multiplication_x: issue in recompose
typescript@3.5.0-dev.20190413: :heavy_multiplication_x: issue in recompose

### Issue
Issue is in `enhancer4` of `/types/recompose/recompose-tests.tsx`
Some screenshots: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34706#issuecomment-482735664

@RyanCavanaugh could you please advise what should I do - while issue with recompose isn't related to this PR, Travis CI build of this PR failed

ferdaber:
`ReadonlyArray<T>` is already a subtype of `Array<T>`, so wouldn't this change make the parameter more restrictive for those who just have regular arrays?

ibezkrovnyi:
@ferdaber 

1) `ReadonlyArray<T>` is not subtype of `Array<T>`, it has no `push`, `pop`, `shift`, etc. Inside `lib.es5.d.ts` `Array` and `ReadonlyArray` are two separate interfaces.

2) `ReadonlyArray` is less restrictive, as it doesn't require from argument to have methods `push`, `pop`, etc.

3) Changed `oneOf` was tested with both `ReadonlyArray`s and with `Array`s too and it works fine with both

> Tested<br>
This change was tested on my project and I used as const in part of code location, while keeping old way for another part of code locations. Both old and new ways work fine.

ferdaber:
Derp, I went the opposite way. Sounds good, thanks for the clarification!

ferdaber (approved):

ibezkrovnyi:
Removed non-consistent periods in JSDoc to trigger CI build

PR #34711, opened 2019-04-13
Types for npm package: react-hooks-helper

jedmundo:


.github/CODEOWNERS
```diff
@@ -4296,6 +4296,7 @@
 /types/react-highlight-words/                       @mhegazy @diogodca @kellyrmilligan
 /types/react-highlighter/                           @oizie
 /types/react-holder/                                @isman-usoh
+/types/react-hot-loader/                            @jedmundo
```
armanio123: Don't change this file as is auto-generated. Instead you can be added on the react-hot-loader "definititions by" if you'll like to be a code-owner.

types/react-hooks-helper/index.d.ts
```diff
@@ -0,0 +1,33 @@
+// Type definitions for react-hooks-helper 1.6
```
armanio123: I noticed that this package also distributes a umd module. You might want to add a namespace for that. `export as namespace myLibl;`.

Check [this](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) template for details.

armanio123 (changes requested):

.github/CODEOWNERS
jedmundo: Thanks! Done.

types/react-hooks-helper/index.d.ts
jedmundo: done!

armanio123 (approved):

PR #34714, opened 2019-04-14
Add typing definitions for bootstrap-menu

superheri:
Please fill in this template.

If adding a new definition:


superheri:
If anyone can help me with this, I had a bit of problem with exporting the main function since it's not supposed to be what is exported by default and called directly when doing require...

types/bootstrap-menu/bootstrap-menu-tests.ts
```diff
@@ -0,0 +1,12 @@
+import BootstrapMenu from "bootstrap-menu";
```
armanio123: This should be `import BootstrapMenu = require "bootstrap-menu"` instead.

types/bootstrap-menu/bootstrap-menu-tests.ts
```diff
@@ -0,0 +1,12 @@
...
+
+BootstrapMenu("#main", {
```
armanio123: The documentation shows this as a class instead. `var a = new BoostrapMenu(...)`

types/bootstrap-menu/index.d.ts
```diff
@@ -0,0 +1,114 @@
...
+export type BootstrapMenuEvent = "click" | "right-click" | "hover";
+
+export type BootstrapMenuRowElement = any;
+
+/**
+ * The function to mount the menu to an existing UI item
+ * @param selector Css selector of the element to attach the Menu to
+ * @param options Options to configure the menu
+ */
+export default function BootstrapMenu(selector: string, options: BootstrapMenuOptions): void;
```
armanio123: I believe this is incorrect as the package doesn't export a `default`. Instead you should use `export  =` syntax. `export = BootstrapMenu`.

Also is not a `function` but a class.

```ts
declare class BootstrapMenu {
    constructor(...)
}
```

Last, the rest of the exports can me added to a namespace:
```ts
declare namespace BoostrapMenu {
    export interface BootstrapMenuOptions {
        ...
    }
}
```

You can see more details on the [template documentation](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-class-d-ts.html).

armanio123 (changes requested):

types/bootstrap-menu/index.d.ts
superheri: https://github.com/dgoguerra/bootstrap-menu/blob/master/src/BootstrapMenu.js#L431 For me this looks like a default export. If I'm not doing an export default I can't do directly new BootstrapMenu with the `import BootstrapMenu = require("bootstrap-menu")`

superheri: Ok my bad I just say that doing `export =` instead of `export class` works.

superheri:
@armanio123 Is there a reason the build is failing even if I did like you told me?

armanio123:
Sorry I forgot about tslint. If you check the travis logs, you'll notice this error: 
```
ERROR: 111:9  no-unnecessary-class  Every member of this class is a constructor. Use functions instead.
```
You an either avoid this disabling on `tslint.json` the rule, or if you want to go the extra mile, I noticed the  library also exposes a couple other functions on the class liek `init`, `updatePosition`, etc. You could add at least one method and this error will go away.

It's up to you. I'll approve any of the two solutions.

superheri:
I added the `destroy` and `close` methods since it might be useful to call those to remove unused menus.

armanio123 (approved):

armanio123:
Thanks for addressing the comments.

PR #34720, opened 2019-04-14
Support priority on Mail.Options

poseidonCore:
As stated in this file:
https://github.com/nodemailer/nodemailer/blob/master/lib/mailer/mail-message.js

### DIFF
--- a/types/nodemailer/lib/mailer/index.d.ts
+++ b/types/nodemailer/lib/mailer/index.d.ts
@@ -137,8 +137,7 @@ declare namespace Mail {
         dkim?: DKIM.Options;
         /** method to normalize header keys for custom caseing */
         normalizeHeaderKey?(key: string): string;
-        
-        priority?:"high"|"normal"|"low";
+        priority?: "high"|"normal"|"low";
     }
 
     type PluginFunction = (mail: MailMessage, callback: (err?: Error | null) => void) => void;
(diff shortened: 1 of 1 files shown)


types/nodemailer/lib/mailer/index.d.ts
```diff
@@ -137,6 +137,7 @@ declare namespace Mail {
         dkim?: DKIM.Options;
         /** method to normalize header keys for custom caseing */
         normalizeHeaderKey?(key: string): string;
+        priority?: "high"|"normal"|"low";
```
bioball: Seems like this should just be `"high"` or `"low"`, based off their source code

https://github.com/nodemailer/nodemailer/blob/master/lib/mailer/mail-message.js#L225...L238

bioball (changes requested):
Didn't see a `"normal"` priority anywhere?

poseidonCore:
@biobal: ... It is the default, as in the code you quoted ...

https://github.com/nodemailer/nodemailer/blob/master/lib/mailer/mail-message.js#L236...L237

bioball:
When is the "normal" value actually used? It doesn't seem like that means anything in code.

poseidonCore:
It's used in cases where programmers are explicitly specifying the priority, rather than just assuming it.

If 'normal' is passed, then it means in the code that no priority is set in the email itself, which its way of implying 'normal' priority, as described in the code itself : `// do not add anything, since all messages are 'Normal' by default`.

The fact that the code does nothing with the value does not negate its semantic significance to the consumer of the type interface. 

It is obvious that the code accommodates this as an option, and therefore it should be an option. Otherwise, it is misleading to say that the only priorities are just  `"high"|"low"`.

bioball:
Ok, that makes sense to me

bioball (approved):
LGTM

PR #34724, opened 2019-04-14
feat: add types for throttleit

ifiokjr:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/throttleit/index.d.ts
+++ b/types/throttleit/index.d.ts
@@ -1,0 +1,18 @@
+// Type definitions for throttleit 1.0
+// Project: https://github.com/component/throttle
+// Definitions by: Ifiok Jr. <https://github.com/ifiokjr>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+type AnyFunction<R = any> = (...args: any[]) => R;
+
+/**
+ * Returns a new function that, when invoked, invokes `func` at most once per `wait` milliseconds.
+ *
+ * @param func Function to wrap.
+ * @param wait Number of milliseconds that must elapse between `func` invocations.
+ * @return A new function that wraps the `func` function passed in.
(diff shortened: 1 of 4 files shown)


types/throttleit/index.d.ts
```diff
@@ -0,0 +1,18 @@
...
+type AnyFunction<R = any> = (...args: any[]) => R;
+
+/**
+ * Returns a new function that, when invoked, invokes `func` at most once per `wait` milliseconds.
+ *
+ * @param func Function to wrap.
+ * @param wait Number of milliseconds that must elapse between `func` invocations.
+ * @return A new function that wraps the `func` function passed in.
+ */
+declare function throttle<GFunction extends AnyFunction>(fn: GFunction, wait: number): GFunction;
```
armanio123: You can use the type `Function` instead of `AnyFunction`.

armanio123 (changes requested):

types/throttleit/index.d.ts
ifiokjr: The reason I didn't do that was because of this lint error.

> Don't use 'Function' as a type. Avoid using the `Function` type. Prefer a specific function type, like `() => void` - `.tslint(ban-types)`

Should I disable that rule in the `tslint.json` file

ifiokjr:
@armanio123 see the [comment](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34724#discussion_r275836402) I made on why I didn't use `Function`?

types/throttleit/index.d.ts
armanio123: Ah my bad. I discussed about this with the team and we agree is better to avoid `Function` and use `AnyFunction` as you originally posted as `Function` comes with another set of issues. 

I'll merge this. Thanks for contributing.

armanio123 (approved):

PR #34739, opened 2019-04-15
Add type definitions for ssh2-sftp-client 2.5

Infantys:


ascariandrea (approved):

viamuli (commented):
Looks pretty good but shouldn't the version bump to 3.x instead of 2.5 since the public API change in `put` (l:27)?

Infantys:
> Looks pretty good but shouldn't the version bump to 3.x instead of 2.5 since the public API change in `put` (l:27)?

The latest version is ssh2-sftp-client 2.5.0 from Mar 8 2019 (https://github.com/jyu213/ssh2-sftp-client/releases). Changes are made for this version.

viamuli (approved):

types/ssh2-sftp-client/ssh2-sftp-client-tests.ts
```diff
@@ -1,42 +1,40 @@
-import Client = require('ssh2-sftp-client');
-import * as fs from 'fs';
+import * as Client from "ssh2-sftp-client";
```
armanio123: Could you please rollback this line? Namespaces object are not callable. See [this](https://stackoverflow.com/questions/39415661/what-does-resolves-to-a-non-module-entity-and-cannot-be-imported-using-this) post for details.

armanio123 (changes requested):

armanio123 (approved):

PR #34743, opened 2019-04-15
Added type definitions for b64-lite 1.3

JasonHK:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/b64-lite/index.d.ts
+++ b/types/b64-lite/index.d.ts
@@ -1,0 +1,12 @@
+// Type definitions for b64-lite 1.3
+// Project: https://github.com/kevlened/b64-lite#readme
+// Definitions by: JasonHK <https://github.com/JasonHK>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export as namespace b64Lite;
+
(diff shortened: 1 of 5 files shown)


types/b64-lite/index.d.ts
```diff
@@ -0,0 +1,12 @@
+// Type definitions for b64-lite 1.3
+// Project: https://github.com/kevlened/b64-lite#readme
+// Definitions by: JasonHK <https://github.com/JasonHK>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export as namespace b64Lite;
```
armanio123: I believe the namespace is incorrect as the module doesn't declare any global variables. Please remove it.

armanio123 (changes requested):

types/b64-lite/index.d.ts
JasonHK: It did provide a UMD build, `dist/b64-lite.umd.js`, that exposed `b64Lite` to the global object.

armanio123: Great, in that case merging as is. Thanks for clarifying.

PR #34745, opened 2019-04-15
[Storybook] Update handler to return any.

milesj:
Please fill in this template.

--------

Handlers may return a value, for example an `onSubmit` that returns a promise. At the moment, `undefined` causes type issues, so just accept `any`.

to: @joscha

### DIFF
--- a/types/storybook__addon-actions/index.d.ts
+++ b/types/storybook__addon-actions/index.d.ts
@@ -4,7 +4,7 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
-export type HandlerFunction = (...args: any[]) => undefined;
+export type HandlerFunction = (...args: any[]) => any;
 export type DecoratorFunction = (args: any[]) => any[];
 export interface Options {
     depth?: number;
@@ -12,6 +12,8 @@ export interface Options {
     limit?: number;
 }
 
-export function decorateAction(decorators: DecoratorFunction[]): (name: string, options?: Options) => HandlerFunction;
+export function decorateAction(
+    decorators: DecoratorFunction[],
+): (name: string, options?: Options) => HandlerFunction;
 export function configureActions(options: Options): undefined;
 export function action(name: string): HandlerFunction;
--- a/types/storybook__addon-actions/index.d.ts
+++ b/types/storybook__addon-actions/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for @storybook/addon-actions 3.4
+// Type definitions for @storybook/addon-actions 5.0
 // Project: https://github.com/storybooks/storybook, https://github.com/storybooks/storybook/tree/master/addons/actions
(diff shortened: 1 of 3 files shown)


milesj:
Bump @joscha @jicjjang

joscha:
@milesj can you link the respective doc or code piece in the storybook repo please?

milesj:
@joscha It's for the `action` function: https://github.com/storybooks/storybook/blob/master/addons/actions/src/preview/action.js#L5

joscha:
Thanks, it's referring to this part: https://github.com/storybooks/storybook/tree/master/addons/actions#action-decorators

I don't think we can make the changes exactly as you propose as the `HanderFunction` type is also used for the standard non-decorated action, and that one has a return value of `undefined`. They are probably two distinct types, maybe `HandlerFunction` and `DecoratedHandlerFunction`? Also, do you think they can be generic instead of returning `any`?
The last thing that comes to mind is that we should probably update `decorateAction` to `decorate` in v5, wdyt?

milesj:
@joscha Sounds good. I added a new type that uses generic inference. Verified in the test and it returned `boolean`.

joscha:
Thanks for the changes. I am a bit confused as I expected the types to be swapped. The simple `action('bla')` is the one that would still return `undefined`, so it would be the `HandlerFunction` and the decorated one is the one that would produce the different return value that gets passed on, e.g. the `ActionFunction` - what am I missing?

milesj:
@joscha

If `onFoo` is typed as `() => number` and in your story, if you do `onFoo={action('onFoo')}`, this causes a TS type error as `onFoo` expects a return of `number`, *not* `undefined`.

joscha:
> if you do `onFoo={action('onFoo')}`, this causes a TS type error as `onFoo` expects a return of `number`, _not_ `undefined`.

yes, rightfully so, because a simple `action('onFoo')` would actually return `undefined`, not a number. Just because we update the types wouldn't mean the runtime changes, so if your `onFoo` wanted to use an `action`, you'd need to `onFoo={() => { action('onFoo')(); return 123; }}`. Changing the signature to `=> any` only means that we are lying to the type system?

joscha:
I think I got confused about the intent of this PR, sorry about that. If the idea is to change the signature so TS would (incorrectly) accept an `action` function as any function that is not a handler (onclick, etc.) with a `() => undefined` signature, then I think this PR should not be merged.

milesj:
@joscha You're completely right. It's been a long day, I was thinking about this in reverse, hah. I think a good fix would be upstream in storybook, possible something like `action('onFoo', 123)`, where you can define the expected return.

I'll close this for now.

joscha:
> possible something like `action('onFoo', 123)`, where you can define the expected return.

that would actually be really cool and then we can infer the return from the type passed as second param. It would work more like a mock function with a predetermined return value.

milesj:
@joscha Yeah exactly, was thinking the same. I'll tinker around with it tomorrow.

PR #34763, opened 2019-04-16
node-red: Fix RED.nodes.getNode return type

bernardobelchior:
Update getNode return type, since the 'id' may not map to a valid node. In this case, node-red will return null.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/node-red/node-red/blob/0.20.5/packages/node_modules/%40node-red/runtime/lib/nodes/flows/index.js#L208

### DIFF
--- a/types/node-red/index.d.ts
+++ b/types/node-red/index.d.ts
@@ -1,9 +1,10 @@
-// Type definitions for node-red 0.17
+// Type definitions for node-red 0.20
 // Project: http://nodered.org
 // Definitions by: Anders E. Andersen <https://github.com/andersea>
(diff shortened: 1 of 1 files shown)


bernardobelchior:
I'm not sure if/how I should write tests for this changes. 
Would love some points in the right direction :smile:

andersea (approved):
Thanks for the help!

andersea:
@bernardobelchior - It probabably isn't possible to test this on actual code. If you are sure this is allowed, then all you can do, is to mock a function that uses the api and verify that it compiles.

PR #34764, opened 2019-04-16
[FIX] Updated some type declaration for tabulator-tables

simusr2:
Added missing some enum values to predefined types (according to tabulator documentation/examples)


simusr2:
Fixed build errors. Edited format type declaration.

PR #34766, opened 2019-04-16
[echarts] fix data prop for markPoint

benleilei:
series-bar.markPoint.data is array, not is object


PranavSenthilnathan:
In the future please use the template for when submitting PRs. It helps us review your fix quicker and helps you catch errors. 

qq1037305420:
In addition, excepting series-bar, other series such as series-line which have the filed "markPoint.data" also have their types wrong.

PR #34768, opened 2019-04-16
Update index.d.ts

Kamill90:
Added export for type audioOnly 

If adding a new definition:


vintage (approved):
Looks good :+1:

PR #34769, opened 2019-04-16
Types for node-mailjet npm module.

Nikola-Andreev:
- [ x Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).

If adding a new definition:


types/node-mailjet/index.d.ts
```diff
@@ -0,0 +1,285 @@
+// Type definitions for node-mailjet 3.3
+// Project: https://github.com/mailjet/mailjet-apiv3-nodejs
+// Definitions by: Nikola Andreev <https://github.com/Nikola-Andreev>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
+export as namespace MailJet;
```
armanio123: I believe this library doesn't expose a UMD module. If that's the case, then this line is incorrect and should be removed.

armanio123 (changes requested):

types/node-mailjet/index.d.ts
Nikola-Andreev: Thanks @armanio123 , i removed the main namespace

armanio123 (approved):

PR #34777, opened 2019-04-16
[wegame-api] API improvment and bug fix for onXXXTouch APIs.

jcyuan:
Please fill in this template.

[T] Use a meaningful title for the pull request. Include the name of the package modified.
[T] Test the change in your own code. (Compile and run.)
[T] Add or edit tests to reflect the change. (Run with npm test.)
[T] Follow the advice from the readme.
[T] Avoid common mistakes.
[T] Run npm run lint package-name (or tsc if no tslint.json is present).

[T] Provide a URL to documentation or source code which provides context for the suggested changes:  https://developers.weixin.qq.com/minigame/dev/api/
[T] Increase the version number in the header if appropriate.
[T] If you are making substantial changes, consider adding a tslint.json containing { "extends": "dtslint/dt.json" }.

main changes:
1, fixed bug: the members of the argument object of the callback function for touch event (e.g.: onTouchMove onTouchDown etc) are incorrect.

2, add onTap etc members for API wx.createXXXButton.

3, improve: change the array members of the callback function as read only array for some APIs.

4, change version number to 2.6.5

5, add `non-npm package` to the dt-header to match the new rule.


types/wegame-api/tslint.json
```diff
@@ -1,3 +1,6 @@
 {
-    "extends": "dtslint/dt.json"
+    "extends": "dtslint/dt.json",
+    "rules": {
+    	"dt-header": false
```
armanio123: Please don't disable any rules. I suggest instead just using `2.6`, without the patch.

armanio123 (changes requested):

types/wegame-api/tslint.json
jcyuan: thanks for reviewing on this PR @armanio123 , actually I found chai-jquery uses this way to modify the dt-header: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34112, so I did this to keep the version number newest.
this definition is for a middleware API of a software, no js lib, and the version number changes very often, like 1-2 week a release 2.6.6 2.6.7 2.6.8, so I have to change my lib that often, so I think it's better to keep the minor versoin number to avoid confuse.

sandersn: Disabling dt-header will just hide the problem, unfortunately. Definitely Typed uses the patch version itself [1], so it's impossible to have types that target only a patch version. They must work for all patches of a minor version. 

Does wegame-api really add or break things in patch versions? You might have to just add things from newer patch versions and expect that users who are on older patch versions will not be warned if they incorrectly use these new properties.

[1] Definitely Typed will ignore the patch version you provide and, starting from 0, increment the patch version for every update of the same minor version.

jcyuan: @sandersn oh! so actually the version in .d.ts is the version for the definition but not the JS Lib version?

armanio123: That's correct the version on the d.ts. is the version for the definition.

jcyuan: I got it, I'm sorry for the confusion, will do the changes

jcyuan:
all resolved, please re-check again, thanks @armanio123

armanio123 (approved):

armanio123:
Thanks for fixing it.

PR #34779, opened 2019-04-16
chrome typings: added chrome.serial bindings

ekinsol:

Implemented all missing chrome.serial bindings:

https://developer.chrome.com/apps/serial


matcarlson (approved):
Thanks for adding this namespace.

PR #34781, opened 2019-04-16
[FIX] Updated some type declaration for tabulator-tables

simusr2:
Added missing some enum values to predefined types (according to tabulator documentation/examples)

### DIFF
--- a/types/tabulator-tables/index.d.ts
+++ b/types/tabulator-tables/index.d.ts
@@ -122,7 +122,7 @@ declare namespace Tabulator {
         /**  Setting this option to true will cause Tabulator to create a list of page size options, that are multiples of the current page size. In the example below, the list will have the values of 5, 10, 15 and 20.
 
     When using the page size selector like this, if you use the setPageSize function to set the page size to a value not in the list, the list will be regenerated using the new page size as the starting valuer	*/
-        paginationSizeSelector?: true | false | number[];
+        paginationSizeSelector?: true | number[];
         /**  By default the pagination controls are added to the footer of the table. If you wish the controls to be created in another element pass a DOM node or a CSS selector for that element to the paginationElement option.*/
         paginationElement?: HTMLElement | "string";
         /** Lookup list to link expected data feilds from the server to their function	* default* {
@@ -764,7 +764,7 @@ declare namespace Tabulator {
     interface ColumnDefinition extends ColumnLayout, CellCallbacks {
         // Layout
         /** sets the text alignment for this column */
-        align?: "left" | "center" | "right" | string; // Align?
+        align?: ColumnDefinitionAlign; // Align?
         /** sets the minimum width of this column, this should be set in pixels (this takes priority over the global option of columnMinWidth) */
         minWidth?: number;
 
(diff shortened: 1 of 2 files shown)


types/tabulator-tables/index.d.ts
```diff
@@ -764,7 +764,7 @@ declare namespace Tabulator {
     interface ColumnDefinition extends ColumnLayout, CellCallbacks {
         // Layout
         /** sets the text alignment for this column */
-        align?: "left" | "center" | "right"; // Align?
+        align?: "left" | "center" | "right" | string; // Align?
```
Jojoshua: Documentation only states left,center,right.
http://tabulator.info/docs/4.2/columns#definition

types/tabulator-tables/index.d.ts
```diff
@@ -1376,6 +1376,7 @@ declare class Tabulator {
     browserSlow: boolean;
     modules: any;
     options: Tabulator.Options;
+    element: any;
```
Jojoshua: element is not part of the Tabulator class. It is a parameter that can be passed in but not saved to this

 `var Tabulator = function(element, options){`

types/tabulator-tables/index.d.ts
```diff
@@ -1003,7 +1003,7 @@ You can pass an optional additional property with sorter, sorterParams that shou
     type CustomAccessorParams = {} | ((value: any, data: any, type: "data" | "download" | "clipboard", column?: ColumnComponent) => any);
     type ColumnCalc = "avg" | "max" | "min" | "sum" | "concat" | "count" | ((values: any[], data: any[], calcParams: {}) => number);
     type ColumnCalcParams = (values: any, data: any) => any;
-    type Formatter = "plaintext" | "textarea" | "html" | "money" | "image" | "datetime" | "datetimediff" | "link" | "tickCross" | "color" | "star" | "traffic" | "progress" | "lookup" | "buttonTick" | "buttonCross" | "rownum" | "handle" | ((cell: CellComponent, formatterParams: {}, onRendered: EmptyCallback) => string | HTMLElement);
+    type Formatter = ((cell: CellComponent, formatterParams: {}, onRendered: EmptyCallback) => string) | string | "plaintext" | "textarea" | "html" | "money" | "image" | "datetime" | "datetimediff" | "link" | "tickCross" | "color" | "star" | "traffic" | "progress" | "lookup" | "buttonTick" | "buttonCross" | "rownum" | "handle" | ((cell: CellComponent, formatterParams: {}, onRendered: EmptyCallback) => string | HTMLElement);
```
Jojoshua: The custom formatter at the end handles the one you copied on the front. The custom formatter should return either a string or html element per the docs.

http://tabulator.info/docs/4.2/format#format-builtin

Also, I don't see where you can add a plain string as a Formatter option. This negates the type benefits as well. I would undo this one too.

types/tabulator-tables/index.d.ts
```diff
@@ -122,7 +122,7 @@ declare namespace Tabulator {
         /**  Setting this option to true will cause Tabulator to create a list of page size options, that are multiples of the current page size. In the example below, the list will have the values of 5, 10, 15 and 20.
 
     When using the page size selector like this, if you use the setPageSize function to set the page size to a value not in the list, the list will be regenerated using the new page size as the starting valuer	*/
-        paginationSizeSelector?: true | number[];
+        paginationSizeSelector?: true | false | number[];
```
Jojoshua: The docs don't say anything about setting this to false. It is default false and redundant to set.

http://tabulator.info/docs/4.2/release#page

types/tabulator-tables/tabulator-tables-tests.ts
```diff
@@ -120,7 +120,7 @@ colDef.formatterParams = {
     huge: true
 };
 // Custom Formatter
-colDef.formatter = (cell: Tabulator.CellComponent, formatterParams: {}, onRendered) => {
+colDef.formatter = (cell: Tabulator.CellComponent, formatterParams: {}, onRendered: Tabulator.EmptyCallback) => {
```
Jojoshua: Typescript already infers this as Tabulator.EmptyCallback.

Jojoshua (changes requested):

types/tabulator-tables/index.d.ts
simusr2: Thanks for all your replies.
We did this particular edit because we have to get the id of a tabulator javascript object. 
Using table.element.id this can be done, but if element is not declared we can't use it. 
table.element returns the html object of the table.
Do you know how to get the id?
Thanks

types/tabulator-tables/index.d.ts
simusr2: I undestand what you say but this row is uncorrect. 
I will do another commit only for this. 
Thanks

types/tabulator-tables/index.d.ts
simusr2: From https://github.com/olifolkerd/tabulator/blob/master/src/js/core.js :
```
Tabulator.prototype.initializeElement = function(element){
	if(element instanceof HTMLElement){
		**this.element** = element;
```

I think **this.element** is part of tabulator class.

Jojoshua: I gotcha now. I'm good with it.

Jojoshua (approved):

PR #34784, opened 2019-04-16
Fix errors from Microsoft/TypeScript#30769

sandersn:
Microsoft/TypeScript#30769 is now in `typescript@next` and finds a couple of bugs in tests.

active-libreoffice's test code was wrong, and easy to fix because it was just using the wrong type. 

react-bootstrip-table's test code is harder to fix, since the ref's value is known to be a string, but all of `newRow`'s properties are numbers except one, so the code is guaranteed to be wrong for all but one property. 

For now I just put a cast in, but production code should probably also add a type check beforehand: `if (typeof newRow[column.field] === 'string')`. TS doesn't narrow on computed element access so the cast is still necessary.

### DIFF
--- a/types/activex-libreoffice/activex-libreoffice-tests.ts
+++ b/types/activex-libreoffice/activex-libreoffice-tests.ts
@@ -129,7 +129,7 @@
 
     function createStruct<K extends keyof LibreOffice.StructNameMap>(strTypeName: K): LibreOffice.StructNameMap[K] {
         const classSize = coreReflection.forName(strTypeName);
-        const aStruct: [LibreOffice.StructNameMap[K]] = [] as any;
+        const aStruct: [LibreOffice.InstantiableNameMap[K]] = [] as any;
         classSize.createObject(aStruct);
         return aStruct[0];
     }
--- a/types/react-bootstrap-table/react-bootstrap-table-tests.tsx
+++ b/types/react-bootstrap-table/react-bootstrap-table-tests.tsx
@@ -779,7 +779,7 @@ class MyCustomBody extends React.Component<MyCustomBodyProps> implements ModalBo
   getFieldValue() {
     const newRow: Partial<Product> = {};
     this.props.columns.forEach((column, i) => {
-      newRow[column.field] = (this.refs[column.field] as HTMLInputElement).value;
+      newRow[column.field] = (this.refs[column.field] as HTMLInputElement).value as number & string;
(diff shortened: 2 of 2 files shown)


types/react-bootstrap-table/react-bootstrap-table-tests.tsx
```diff
@@ -779,7 +779,7 @@ class MyCustomBody extends React.Component<MyCustomBodyProps> implements ModalBo
   getFieldValue() {
     const newRow: Partial<Product> = {};
     this.props.columns.forEach((column, i) => {
-      newRow[column.field] = (this.refs[column.field] as HTMLInputElement).value;
+      newRow[column.field] = (this.refs[column.field] as HTMLInputElement).value as number & string;
```
dawnmist: @sandersn Is it actually possible for a value to be of type number and type string at the same time? I would have thought `number | string` to be more appropriate.

sandersn: Yep, it's definitely impossible! This cast is just to tell the compiler we know what we're doing -- specifically, because `HTMLInputElement.value : string`, we are asserting that `newRow[column.field]` can safely be assigned `string` even though the compiler conservatively guesses its type to be `string & number` (the intersection of types of all properties of `newRow`). This isn't generally true, so in real code we'd want to guard with `if (typeof newRow[column.field] === typeof 'string')`

types/activex-libreoffice/activex-libreoffice-tests.ts
```diff
@@ -129,7 +129,7 @@
 
     function createStruct<K extends keyof LibreOffice.StructNameMap>(strTypeName: K): LibreOffice.StructNameMap[K] {
```
zspitz: @sandersn This would be better as:
```
function createStruct<K extends keyof LibreOffice.StructNameMap>(strTypeName: K): LibreOffice.StructNameMap[K] {
    const reflected = coreReflection.forName(strTypeName) as com.sun.star.reflection.XIdlClass<LibreOffice.StructNameMap[K]>;
    const aStruct: [LibreOffice.StructNameMap[K]] = [] as any;
    reflected.createObject(aStruct);
    return aStruct[0];
}
```

sandersn: Opened #34814 to do this.

Issue #34588, opened 2019-04-09
[@types/react] Type instantiation is excessively deep and possibly infinite


diegohaz:
#34386 possibly broke one of my components with this error:
```
TS2589: Type instantiation is excessively deep and possibly infinite.
```
My component has already complex types. My guess is that this change introduced deeper levels of type instantiation and, combined with mine, reached the typescript compiler limit.

Reproduction:
```sh
git clone https://github.com/reakit/reakit
cd reakit
git checkout cf7426f0cb35c2ed4a02308cc426d247b3145ad2
yarn
yarn add -WD @types/react@16.8.12
yarn type-check # no errors with @types/react@16.8.12
yarn add -WD @types/react@16.8.13
yarn type-check # Found 3 errors
```

- [x] I tried using the `@types/xxxx` package and had problems.
- [x] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @johnnyreilly @bbenezech @pzavolinsky @digiguru @ericanderson @tkrotoff @DovydasNavickas @onigoetz @theruther4d @guilhermehubner @ferdaber @jrakotoharisoa @pascaloliv @hotell @franklixuefei @Jessidhia @pshrmn @saranshkataria @lukyth @eps1lon

diegohaz:
I solved it by reducing the complexity of my types (basically removing one level in the path array), but I'm not sure this is the best solution, so this issue may be still valid.

ferdaber:
Can you give a snippet or post a repo that has your types? It's not possible to debug with just this information.

orta:
Hi thread, we're moving DefinitelyTyped to use [GitHub Discussions](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/53377) for conversations the `@types` modules in DefinitelyTyped.

To help with the transition, we're closing all issues which haven't had activity in the last 6 months, which includes this issue. If you think closing this issue is a mistake, please pop into the [TypeScript Community Discord](https://discord.gg/typescript) and mention the issue in the `definitely-typed` channel.

PR #34386, opened 2019-04-02
[React] Fix optionality of inferred props

matt-tingen:
Please fill in this template.

-   [x] Use a meaningful title for the pull request. Include the name of the package modified.
-   [X] Test the change in your own code. (Compile and run.)
-   [X] Add or edit tests to reflect the change. (Run with `npm test`.)
-   [X] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
-   [X] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
-   [X] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:

-   [x] Provide a URL to documentation or source code which provides context for the suggested changes: Fixes #33742
-   [ ] ~~Increase the version number in the header if appropriate.~~
-   [ ] ~~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.~~

This change ensures that the optional modifier (`?`) is copied when using inferred properties from `propTypes`.

This reopens #34053 as I was able to get the test suite to complete with `--max_old_space_size` (
[test run log](https://github.com/DefinitelyTyped/DefinitelyTyped/files/3032292/test.log)). The only errors are `no-null-undefined-union` lint errors. This is a [new rule](https://github.com/palantir/tslint/pull/4589) and is [likely not finalized](https://github.com/palantir/tslint/pull/4589#issuecomment-477758158). Addressing the lint errors seems orthogonal to this change, but let me know if we need to do so first.

PR #34789, opened 2019-04-17
[React] Preserve optionality of props in union types

### LINKED ISSUE #34589: [@types/react] latest 16.18.13 makes optional prop as required
as discuss of #34386

Here is the reproduce code. It works on 16.18.12 but failed on 16.18.13:
```tsx
import * as PropTypes from 'prop-types';

export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
export const tuple = <T extends string[]>(...args: T) => args;

const ButtonTypes = tuple('default', 'primary', 'ghost', 'dashed', 'danger');
export type ButtonType = (typeof ButtonTypes)[number];
const ButtonShapes = tuple('circle', 'circle-outline', 'round');
export type ButtonShape = (typeof ButtonShapes)[number];
const ButtonSizes = tuple('large', 'default', 'small');
export type ButtonSize = (typeof ButtonSizes)[number];
const ButtonHTMLTypes = tuple('submit', 'button', 'reset');
export type ButtonHTMLType = (typeof ButtonHTMLTypes)[number];

export type AnchorButtonProps = {
  href: string;
  target?: string;
  onClick?: React.MouseEventHandler<HTMLAnchorElement>;
} &
  Omit<React.AnchorHTMLAttributes<HTMLAnchorElement>, 'type'>;

export type NativeButtonProps = {
  htmlType?: ButtonHTMLType;
  onClick?: React.MouseEventHandler<HTMLButtonElement>;
} &
  Omit<React.ButtonHTMLAttributes<HTMLButtonElement>, 'type'>;

export type ButtonProps = AnchorButtonProps | NativeButtonProps;

class Button extends React.Component<ButtonProps, any> {
  static propTypes = {
    type: PropTypes.string,
    shape: PropTypes.oneOf(ButtonShapes),
    size: PropTypes.oneOf(ButtonSizes),
    htmlType: PropTypes.oneOf(ButtonHTMLTypes),
    onClick: PropTypes.func,
    loading: PropTypes.oneOfType([PropTypes.bool, PropTypes.object]),
    className: PropTypes.string,
    icon: PropTypes.string,
    block: PropTypes.bool,
  };

  render() {
    return null;
  }
}

// Follow code will get ts error that require props which is optional
export default () => (
  <Button>test</Button>
);
```

ref issue: ant-design/ant-design#15930

@matt-tingen @Jessidhia @ferdaber

matt-tingen:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: Fixes #34589
- [x] ~Increase the version number in the header if appropriate.~
- [x] ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.~

Test results: [test.log](https://github.com/DefinitelyTyped/DefinitelyTyped/files/3087672/test.log)
There are two test failures: one each in `react-bootstrap-table` and `recompose`. I confirmed that they are pre-existing, presumably due to nightly TS changes. Please let me know if these need to be addressed in this PR.

This addresses an issue introduced in #34386 where optional props in union types were reverted to being required.

Thanks for the pointer on `DistributivePick`, @ferdaber; it works great!
This approach does further deepen the definition which does not bode well for #34588. However, I would consider this fix a priority over the deepness issue since the latter issue was resolved in user-land. When I get some more time, I'm going to look into alternate solutions that would help to keep `MergePropTypes` more shallow.

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -2760,6 +2760,7 @@ declare namespace React {
     }
 }
 
+type DistributivePick<T, K extends keyof T> = T extends any ? Pick<T, K> : never;
 // naked 'any' type in a conditional type will short circuit and union both the then/else branches
 // so boolean is only resolved for T = any
 type IsExactlyAny<T> = boolean extends (T extends never ? true : false) ? true : false;
@@ -2774,11 +2775,11 @@ type MergePropTypes<P, T> =
         // If declared props have indexed properties, ignore inferred props entirely as keyof gets widened
         string extends keyof P ? P :
             // Prefer declared types which are not exactly any
-            & Pick<P, NotExactlyAnyPropertyKeys<P>>
+            & DistributivePick<P, NotExactlyAnyPropertyKeys<P>>
             // For props which are exactly any, use the type inferred from propTypes if present
-            & Pick<T, Exclude<keyof T, NotExactlyAnyPropertyKeys<P>>>
+            & DistributivePick<T, Exclude<keyof T, NotExactlyAnyPropertyKeys<P>>>
             // Keep leftover props not specified in propTypes
-            & Pick<P, Exclude<keyof P, keyof T>>;
+            & DistributivePick<P, Exclude<keyof P, keyof T>>;
 
 // Any prop that has a default prop becomes optional, but its type is unchanged
 // Undeclared default props are augmented into the resulting allowable attributes
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -2760,7 +2760,6 @@ declare namespace React {
     }
 }
 
-type DistributivePick<T, K extends keyof T> = T extends any ? Pick<T, K> : never;
 // naked 'any' type in a conditional type will short circuit and union both the then/else branches
 // so boolean is only resolved for T = any
 type IsExactlyAny<T> = boolean extends (T extends never ? true : false) ? true : false;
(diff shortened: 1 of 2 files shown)


types/react/test/managedAttributes.tsx
```diff
@@ -272,3 +272,5 @@ interface LeaveMeAloneDtslint { foo: string; }
 // const weakComponentIndexedTest2: JSX.LibraryManagedAttributes<{ propTypes: typeof weakComponentPropTypes }, IndexedComponentProps> = { foo: '' };
 // const weakComponentIndexedTest3: JSX.LibraryManagedAttributes<{ propTypes: typeof weakComponentPropTypes }, WeakIndexedComponentProps> = { foo: '' };
 // const weakComponentIndexedTest4: JSX.LibraryManagedAttributes<{ propTypes: typeof weakComponentPropTypes }, WeakIndexedComponentProps> = { foo: 4 };
+
+// const optionalUnionPropTest: JSX.LibraryManagedAttributes<{ propTypes: {} }, { optional?: string } | { optional?: number }> = {};
```
matt-tingen: Having `propTypes` defined is needed in order to fall into `MergePropTypes` from `ReactManagedAttributes`.

types/react/index.d.ts
```diff
@@ -2774,11 +2775,11 @@ type MergePropTypes<P, T> =
         // If declared props have indexed properties, ignore inferred props entirely as keyof gets widened
         string extends keyof P ? P :
             // Prefer declared types which are not exactly any
-            & Pick<P, NotExactlyAnyPropertyKeys<P>>
+            & DistributivePick<P, NotExactlyAnyPropertyKeys<P>>
```
ferdaber: You might not need to have distributive pick peppered over if you just wrap the entire `MergePropTypes` type with a conditional:
```ts
type MergePropTypes<P, T> = P extends any ? /* ... the rest of the stuff */ : never
```

ferdaber (commented):
You might be able to save some lines of code just by applying the conditional wrap around `MergePropTypes`?

types/react/index.d.ts
matt-tingen: Hmm, I'll give it a try, but I think we would have to do `P extends any ? T extends any ? /* ... */ : never : never` since we `Pick` from both `P` and `T`. If that is the case, I feel like three `DistributivePick`s is easier to read than two ternaries.

ferdaber: Ah you’re right. I thought we were only distributing against P. I do wonder though if T will ever result in a union since it’s only ever applying against the type of `propTypes`

ferdaber (approved):
The changes look good if they fix the issues reported. Thanks for doing this!

types/react/index.d.ts
matt-tingen: Good point. I suppose it's technically possible for `T` to be a union, but probably not worth supporting. I'll update it and confirm that works.

matt-tingen:
I have made the requested change. Here is an updated 
[test.log](https://github.com/DefinitelyTyped/DefinitelyTyped/files/3091134/test.log). The `recompose` issue appears to have been resolved in the latest TS nightly, but the rest is the same.

matt-tingen:
@Jessidhia Is this good to merge?

PR #34798, opened 2019-04-17
elasticsearch: add include_type_name for the relevant Indices APIs

PAStheLoD:
This parameter was introduced around 6.7. It's a requirement for writing cross-version compatible code targeting 6.7.x and 7.x

https://www.elastic.co/guide/en/elasticsearch/reference/master/removal-of-types.html

### DIFF
--- a/types/elasticsearch/index.d.ts
+++ b/types/elasticsearch/index.d.ts
@@ -1121,6 +1121,7 @@ export interface IndicesCloseParams extends GenericParams {
 }
 
 export interface IndicesCreateParams extends GenericParams {
+    includeTypeName?: boolean;
     waitForActiveShards?: string;
     timeout?: TimeSpan;
     masterTimeout?: TimeSpan;
@@ -1199,6 +1200,7 @@ export interface IndicesForcemergeParams extends GenericParams {
 }
 
 export interface IndicesGetParams extends GenericParams {
+    includeTypeName?: boolean;
     local?: boolean;
     ignoreUnavailable?: boolean;
     allowNoIndices?: boolean;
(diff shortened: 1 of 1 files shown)


pushplay:
Unfortunately the Elasticsearch definition file is still on 5.0.  https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/elasticsearch/index.d.ts#L1  To start supporting a new major version there's a whole process to be followed: https://github.com/DefinitelyTyped/DefinitelyTyped#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package
If you're willing to do that work I think a lot of people would appreciate it.

PAStheLoD:
First of all, thanks for the quick reply!

Ah, that's unfortunate, as nowadays Elastic co. is working on a new JS client (the regular one's repo has already been renamed as "legacy"), which has TS support, thus making any large investment into the maintenance of an 5.0 only declaration package sort of pointless and a bit of a dilemma. (Because the new client supports anything from 5.0 and up, with a very similar public API, but uses the snake_case convention for TS definitions to make migration a guaranteed chore.)

I haven't found any word from elastic regarding the future of the legacy client, but they stopped updating that repo back in early February.

All in all, I'll close this PR, as it is incorrect, and if migrating to the new client turns out to be unfeasible, maybe doing a hacky major release thing could be viable. (As in, copy the current 5.x stuff into a dir, fix the tsconfig paths, add these includeTypeName params and call it 6.x compatible - as otherwise for simple operations it works - then gradually clean it up as problems arise.)

pushplay:
Thanks for the info about the new client.  I haven't needed to work on Elasticsearch for a while so I'm a bit out of the loop there.  Totally agree that putting effort into the declaration file of a deprecated client doesn't make sense.

PR #34800, opened 2019-04-17
[express-graphql] Add options introduced in v0.8.0

ChristianBoehlke:
Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [v0.8.0 Changelog](https://github.com/graphql/express-graphql/releases/tag/v0.8.0) and [Flow Types](https://github.com/graphql/express-graphql/blob/cf69ec6158b9e183b562938780d9210ee6133810/src/index.js#L83-L110)

If removing a declaration:

### DIFF
--- a/types/express-graphql/index.d.ts
+++ b/types/express-graphql/index.d.ts
@@ -57,11 +57,13 @@ declare namespace graphqlHTTP {
          * An optional function which will be used to validate instead of default `validate`
          * from `graphql-js`.
          */
-        customValidateFn?: (
-            schema: GraphQLSchema,
-            documentAST: DocumentNode,
-            rules?: ReadonlyArray<any>,
-        ) => ReadonlyArray<GraphQLError> | null;
+        customValidateFn?:
+            | ((
+                  schema: GraphQLSchema,
+                  documentAST: DocumentNode,
+                  rules: ReadonlyArray<any>,
+              ) => ReadonlyArray<GraphQLError>)
+            | null;
 
         /**
          * An optional function which will be used to execute instead of default `execute`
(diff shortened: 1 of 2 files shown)


types/express-graphql/index.d.ts
```diff
@@ -46,11 +53,36 @@ declare namespace graphqlHTTP {
...
 
+        /**
+         * An optional function which will be used to validate instead of default `validate`
+         * from `graphql-js`.
+         */
+        customValidateFn?: (
+            schema: GraphQLSchema,
+            documentAST: DocumentNode,
+            rules?: ReadonlyArray<any>,
+        ) => ReadonlyArray<GraphQLError> | null;
```
PranavSenthilnathan: The `|` has precedence over `=>` so this definition indicates that the return type could be null rather than the function being null (which is what you want based on the flow types).

types/express-graphql/index.d.ts
```diff
@@ -46,11 +53,36 @@ declare namespace graphqlHTTP {
...
         pretty?: boolean | null;
 
+        /**
+         * An optional function which will be used to validate instead of default `validate`
+         * from `graphql-js`.
+         */
+        customValidateFn?: (
+            schema: GraphQLSchema,
+            documentAST: DocumentNode,
+            rules?: ReadonlyArray<any>,
```
PranavSenthilnathan: As discussed [here](http://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html), it's not recommended to have optional parameters in callbacks.

Also, the flow types indicate the parameter is not optional (maybe for the same reason as our recommend above).

PranavSenthilnathan (changes requested):

ChristianBoehlke:
Hey @PranavSenthilnathan, thank you a lot for your feedback! I have changed the type definitions as you have suggested.

PR #34804, opened 2019-04-17
[@types/got] Fix for extended got instance when options are used

wingsbob:
Please fill in this template.

This is a fix for this scenario:

```typescript
import * as got from 'got';

const client = got.extend({ json: true, baseUrl: 'https://domain.com' });

client('/example', { query: { foo: 'bar' } }); // Errors, as the options don't include `json: true`
```


types/got/index.d.ts
```diff
@@ -90,11 +90,13 @@ declare namespace got {
 
     interface GotJSONFn {
         (url: GotUrl): GotPromise<any>;
+        (url: GotUrl, options: GotOptions<string | null>): GotPromise<any>;
         (url: GotUrl, options: GotJSONOptions): GotPromise<any>;
```
ikokostya: Looks like overload with `GotJSONOptions` is redundant?

ikokostya: Nevermind.

I think correct type here is `Partial<GotJSONOptions>`.

wingsbob: Yeah, that's a bit smarter 😄

ikokostya (approved):

PR #34806, opened 2019-04-17
Add some missing Cesium type definitions

mwijngaard:
Add cartographicLimitRectangle to Globe class, and add sampleTerrainMostDetailed function on Cesium namespace.


mwijngaard:
Thanks, @weswigham !

PR #26094, opened 2018-05-29
[reactstrap] Replacing components with generics.

Sprogel:
- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: *N/A* - no functional code change
- [x] Increase the version number in the header if appropriate. *N/A*
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. N/A

PR #34809, opened 2019-04-17
Make reactstrap compile faster

sandersn:
This reverts the strictness part of #26094 without changing its
interface. It retains

1. string index signatures, so that arbitrary props can be passed in.
2. extending from React.Componenent, so that individual types can be subclassed.
3. Type parameters on the individual types.

However, string index signatures can no longer be avoided, and type parameters are ignored. This preserves interface compatibility -- existing code will not break from this change -- but disables excess prop checking. It's possible that a breaking change that returns the type package to its old strict behaviour can be made at the same time as a breaking change of reactstrap itself.

This change improves performance dramatically, and allows linting to succeed without running out of memory. This will unblock changes to `@types/react`, since reactstrap is tested as part of react's CI tests.

```
== Before (3.3.4000) ==
Files:              98
Lines:           63288
Nodes:          201816
Identifiers:     63784
Symbols:        961453
Types:          493230
Memory used:   601673K
I/O Read time:   0.02s
Parse time:      0.39s
Program time:    0.45s
Bind time:       0.22s
Check time:     10.03s
Total time:     10.70s

=== After 2 (3.3.4000) ===
Files:              98
Lines:           63255
Nodes:          200300
Identifiers:     63132
Symbols:        178691
Types:           80350
Memory used:   179229K
I/O Read time:   0.01s
Parse time:      0.41s
Program time:    0.47s
Bind time:       0.23s
Check time:      4.60s
Total time:      5.29s
```

After the first commit, the number of types is just over 337,000. On `typescript@next`, Before had 1.5 million types, and After had 107,000 types.


sandersn:
I'm going to merge this to unblock react. Please comment if you want, though.

types/reactstrap/lib/Breadcrumb.d.ts
```diff
@@ -1,13 +1,14 @@
...
   tag?: string;
   className?: string;
   listTag?: string;
   listClassName?: string;
   cssModule?: CSSModule;
-} & T;
+}
 
-declare class Breadcrumb<T = {[key: string]: any}> extends React.Component<BreadcrumbProps<T>> {}
+declare class Breadcrumb<T = {[key: string]: any}> extends React.Component<BreadcrumbProps> {}
```
FaithForHumans: I was on the side of not having the strict type option from the beginning, so this change is good with me. The original PR was intended to split the gap between two other contributors.

That being said, would something like
```typescript
import { CSSModule } from '../index';

export interface BreadcrumbProps extends React.HTMLAttributes<HTMLElement> {
  tag?: string;
  className?: string;
  listTag?: string;
  listClassName?: string;
  cssModule?: CSSModule;
}

declare class Breadcrumb<T = {[key: string]: any}> extends React.Component<BreadcrumbProps & T> {}
export default Breadcrumb;
```
still have the performance hit? It still cuts down on the number of types since we could keep the `index.html` from this merge, and if someone wanted the strictness, they could use the generic.

I'm not sure how you got your timing numbers, otherwise I would test that change myself.

sandersn: Yes, it would still be slow. :crying_cat_face:  50% of the slowdown is `BreadcrumpProps & T`, wherever you happen to put the intersection. The other 30% is the re-exports in index.d.ts, to make the default strict.

FaithForHumans: I didn't realize the intersections had such a performance impact. Thanks for the explanation!

sandersn: They're a multiplier for other types that are already slow. React is actually the main culprit here; each subclass of React.Component creates multiple thousands of types.

I was trying to get DT back on its feet with this PR since the crash was blocking React PRs. React needs a more considered look at performance soon, though.

mtsmfm:
I've found this PR breaks existing codes when we have our own custom component with `defaultProps`.

```tsx
import { Button, ButtonProps } from "reactstrap";

type Props = {
  message: string;
} & ButtonProps;

class MyButton extends React.Component<Props> {
  public static defaultProps = {
    message: "hi"
  };

  public render() {
    const { message, ...props } = this.props;
    return <Button {...props}>{message}</Button>;
  }
}

const App = () => {
  return (
    <>
      /*
  Overload 1 of 2, '(props: Readonly<Props>): MyButton', gave the following error.
    Property 'message' is missing in type '{}' but required in type 'Readonly<Props>'.
  Overload 2 of 2, '(props: Props, context?: any): MyButton', gave the following error.
    Property 'message' is missing in type '{}' but required in type 'Readonly<Props>'.
*/
      <MyButton />
    </>
  );
};
```

Here's the example:

https://codesandbox.io/s/typesreactstrap-with-defaultprops-lkru4

mtsmfm:
@sandersn Is this intentional?

sandersn:
@mtsmfm 

1. It is not intentional.
2. The PR should not have made any props *more* required than before.
3. This PR resulted in `@types/reacstrap@8.0.1`. Is your code broken on 8.0.1 and working on 8.0.0?

I'll defer to the `@types/reactstrap` contributors since they know more than me. I maintain Definitely Typed, but I don't know the specifics of any library. When I made this change I just avoided breaking tests while improving the compile time.

FaithForHumans:
I think there might be a typo in your example you gave.

> ```tsx
> type Props = {
>   message: string;
> } & ButtonProps;

You've got message marked as required here, but you're omitting it later on. I see you've got a default prop for it, are you sure you didn't mean to mark `message` optional?

If this worked at some point, it makes me think there was a `react` types change with the way `defaultProps` worked and auto marked a prop as optional if it existed in `defaultProps`. But I'll agree with sandersn, I don't think this is due to a `@types/reactstrap` change.

mtsmfm:
Here's the working example with @types/reactstrap 8.0.0

https://codesandbox.io/s/typesreactstrap-800-with-defaultprops-5h36u

mtsmfm:
> 3\. This PR resulted in `@types/reacstrap@8.0.1`. Is your code broken on 8.0.1 and working on 8.0.0?
> 

Yes.

@types/reactstrap 8.0.1 https://codesandbox.io/s/typesreactstrap-with-defaultprops-lkru4
@types/reactstrap 8.0.0 https://codesandbox.io/s/typesreactstrap-800-with-defaultprops-5h36u

> If this worked at some point, it makes me think there was a react types change with the way defaultProps worked and auto marked a prop as optional if it existed in defaultProps. I don't think this is due to a @types/reactstrap change.

Hmm, in the examples, there's no difference other than @types/reactstrap's version.

mtsmfm:
Previously, type parameter `T` for `ButtonProps` was `{}` by default but it now has `[key: string]: any`.

```tsx
export type ButtonProps<T = {}> = React.ButtonHTMLAttributes<HTMLButtonElement> & {
  ...
} & T
```

```tsx
export interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
  [key: string]: any;
  ...
}
```

It seems index signature breaks `defaultProps`.

https://codesandbox.io/s/defaultprops-jnecm

mtsmfm:
@FaithForHumans

> are you sure you didn't mean to mark `message` optional?

Yes, it should be marked as optional by `defaultProps`.

https://codesandbox.io/s/defaultprops-s5jg2

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-0.html#support-for-defaultprops-in-jsx

sandersn:
Huh. @weswigham is that a known limitation of our defaultProps support, or a bug?

weswigham:
We don't handle defaultProps, per sey, the react.d.ts does. I imagine some kind of handling could be added into it, if need be.

PR #34811, opened 2019-04-17
@types/react-native add support for onFocus/onBlur and Platform.isTV

tykus160:
Adding onFocus/onBlur for TouchableWithoutFeedback and Platform.isTV boolean.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://facebook.github.io/react-native/docs/touchablewithoutfeedback#onfocus
https://facebook.github.io/react-native/docs/touchablewithoutfeedback#onblur
https://facebook.github.io/react-native/docs/building-for-apple-tv

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -1172,11 +1172,14 @@ export type ReturnKeyTypeAndroid = "none" | "previous";
 export type ReturnKeyTypeIOS = "default" | "google" | "join" | "route" | "yahoo" | "emergency-call";
 export type ReturnKeyTypeOptions = ReturnKeyType | ReturnKeyTypeAndroid | ReturnKeyTypeIOS;
 
+export interface TargetedEvent {
+    target: number;
+}
+
 /**
  * @see TextInputProps.onFocus
  */
-export interface TextInputFocusEventData {
-    target: number;
+export interface TextInputFocusEventData extends TargetedEvent {
     text: string;
     eventCount: number;
 }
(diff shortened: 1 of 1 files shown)


saranshkataria (approved):

alloy (commented):
Why are we using `any` here `NativeSyntheticEvent<any>`, is there nothing more specific?

tykus160:
This event is defined in React Native as follows:
```
type TargetEvent = SyntheticEvent<
  $ReadOnly<{|
    target: number,
  |}>,
>;
```
And to be honest I'm not sure how to properly translate it into typescript.

types/react-native/index.d.ts
```diff
@@ -5169,6 +5169,20 @@ export interface TouchableWithoutFeedbackProps extends TouchableWithoutFeedbackP
      */
     hitSlop?: Insets;
 
+    /**
+     * When `accessible` is true (which is the default) this may be called when
+     * the OS-specific concept of "blur" occurs, meaning the element lost focus.
+     * Some platforms may not have the concept of blur.
+     */
+    onBlur?: (e: NativeSyntheticEvent<any>) => void;
```
alloy:
```suggestion
    onBlur?: (e: NativeSyntheticEvent<{ target: number }>) => void;
```

types/react-native/index.d.ts
```diff
@@ -5169,6 +5169,20 @@ export interface TouchableWithoutFeedbackProps extends TouchableWithoutFeedbackP
...
+
+    /**
+     * When `accessible` is true (which is the default) this may be called when
+     * the OS-specific concept of "focus" occurs. Some platforms may not have
+     * the concept of focus.
+     */
+    onFocus?: (e: NativeSyntheticEvent<any>) => void;
```
alloy:
```suggestion
    onFocus?: (e: NativeSyntheticEvent<{ target: number }>) => void;
```

alloy:
I see, thanks for digging that up! I’ve left some suggestions based on that, which you can commit if you agree.

It seems like we could refactor some of our event data interfaces to all extend a base `TargetedEvent` interface that has just `target: number`, but that doesn’t need to happen now if you don’t feel comfortable enough yet with TS.

PR #34818, opened 2019-04-18
[@types/redis] ++ add missing type Buffer

janerikmai:
When ClientOpts.return_buffers is set to **true** => Commands.**get** and Commands.**GET** can be **string** or **Buffer**.

This PR fixing the bug that it can only be **string** right now.

### DIFF
--- a/types/redis/index.d.ts
+++ b/types/redis/index.d.ts
@@ -439,8 +439,8 @@ export interface Commands<R> {
     /**
      * Get the value of a key.
      */
-    get(key: string, cb?: Callback<string>): R;
-    GET(key: string, cb?: Callback<string>): R;
+    get(key: string, cb?: Callback<string | Buffer>): R;
+    GET(key: string, cb?: Callback<string | Buffer>): R;
 
     /**
      * Returns the bit value at offset in the string value stored at key.
--- a/types/redis/redis-tests.ts
+++ b/types/redis/redis-tests.ts
@@ -6,11 +6,12 @@ const num = 0;
 const str = 'any string';
 const err: Error = new Error();
 const args: any[] = [];
-const resCallback: (err: Error | null, res: any) => void = () => {};
-const numCallback: (err: Error | null, res: number) => void = () => {};
-const strCallback: (err: Error | null, res: string) => void = () => {};
-const okCallback: (err: Error | null, res: 'OK') => void = () => {};
-const messageHandler: (channel: string, message: any) => void = () => {};
+const resCallback: (err: Error | null, res: any) => void = () => { };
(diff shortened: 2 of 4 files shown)


ferrantejake (changes requested):
These changes do not fully encompass the full extent of the behavior anticipated from assigning the `return_buffers` flag. E.g. the response from setting a key/value pair will be a Buffer. Also style changes, albeit for consistency, should be in a separate pull request. Additionally, the response from a get command will _only_ be a Buffer with the `return_buffers` flag set. It cannot be both.

See documentation for reference on `return_buffers`:
[https://github.com/NodeRedis/node_redis#options-object-properties](https://github.com/NodeRedis/node_redis#options-object-properties)

janerikmai:
> These changes do not fully encompass the full extent of the behavior anticipated from assigning the `return_buffers` flag. E.g. the response from setting a key/value pair will be a Buffer. Also style changes, albeit for consistency, should be in a separate pull request. Additionally, the response from a get command will _only_ be a Buffer with the `return_buffers` flag set. It cannot be both.
> 
> See documentation for reference on `return_buffers`:
> https://github.com/NodeRedis/node_redis#options-object-properties

But so next to the style stuff, the get/GET can be _only_ string **or** _only_ Buffer, right? Or how to write a fixed typing depending on a flag?

ferrantejake:
No fix is necessary for the return  type. Because the return type is dependent on the client's configuration at runtime there is no way to describe this at compile type with types. My comment on the return type was to address the original PR description (for clarification) which says:

> When ClientOpts.return_buffers is set to **true** => Commands.**get** and Commands.**GET** can be **string** or **Buffer**.

The difference between the above and the actual behavior is that it is there is only one return type at any given point in time, as opposed to either type when the `return_buffers` flag is set.

As for what needs to be reviewed: the GET command is not the only command which is affected by this flag, per the documentation:

> If set to true, then all replies will be sent to callbacks as Buffers instead of Strings.

This means this affects any command which returns a value, e.g. SET. It is by no means required to do so, but when adding to a library ideally single issues are taken care of in one foul swoop instead of across PRs. In other words this PR would address the library as a whole, as opposed to just the GET command. This differentiates half-done libs from complete ones. Take the time to add the rest of the affected types to address the full coverage of the changes you want to address and I'd be happy to :+1: your PR (:

PranavSenthilnathan:
@janerikmai Are you planning on addressing the comments? No rush, just wanted to know if not so I can close out the PR in that case.

janerikmai:
@PranavSenthilnathan you can close, I will address at later time in some days and do a new PR
@ferrantejake got it :D howdy...

PR #34821, opened 2019-04-18
atom: update to version 1.36

lierdakil:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://atom.io/docs/api/v1.36.0

Long story short, a few updates. Note though that there are a couple things we missed a while ago (pre 1.31). Also, one minor tweak is that `TextEditor.getLastCursor()` can return `undefined` for e.g. readonly editors. Other changes should be self-explanatory.

### DIFF
--- a/types/atom/atom-tests.ts
+++ b/types/atom/atom-tests.ts
@@ -17,6 +17,7 @@ declare const mouseEvent: MouseEvent;
 
 declare let buffer: Atom.TextBuffer;
 declare const color: Atom.Color;
+declare let maybeCursor: Atom.Cursor | undefined;
 declare let cursor: Atom.Cursor;
 declare let cursors: Atom.Cursor[];
 declare let decoration: Atom.Decoration;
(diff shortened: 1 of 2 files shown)


GlenCFL:
LGTM
`TextEditor::getLastCursor` being able to return undefined doesn't seem intended, but I think it can after looking through their code (located [here](https://github.com/atom/atom/blob/v1.36.0/src/text-editor.js#L2823)).

lierdakil:
@GlenCFL, `getLastCursor` returning `undefined` was something I've personally run into at some point, but tbh I don't exactly remember under what circumstances, and looking through commit history, this is indeed probably unintended. Perhaps filing a bug report on Atom repo would've been a better option now that I think about it.

lierdakil:
Yep, looking at the code, `getLastCursor` shouldn't return `undefined` ever. Couldn't reproduce it either on 1.36. So my bad. Will whip up an amendment shortly.

PR #34829, opened 2019-04-18
prettier: support parsers included as plugins

ChristianMurphy:
as documented in https://prettier.io/docs/en/browser.html#commonjs

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://prettier.io/docs/en/browser.html#commonjs

### DIFF
--- a/types/prettier/index.d.ts
+++ b/types/prettier/index.d.ts
@@ -132,9 +132,9 @@ export interface ParserOptions extends RequiredOptions {
 }
 
 export interface Plugin {
-    languages: SupportLanguage[];
-    parsers: { [parserName: string]: Parser };
-    printers: { [astFormat: string]: Printer };
+    languages?: SupportLanguage[];
+    parsers?: { [parserName: string]: Parser };
+    printers?: { [astFormat: string]: Printer };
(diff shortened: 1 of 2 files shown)


types/prettier/index.d.ts
```diff
@@ -139,6 +139,12 @@ export interface Plugin {
     defaultOptions?: Partial<RequiredOptions>;
 }
 
+export interface BrowserParserPlugin {
```
ikatyang: The browser plugin is actually the same as the Node.js plugin (`Plugin`), though it seems we did not describe it well:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/03186dbc08372aa1ca9689147386523588be6efd/types/prettier/index.d.ts#L134-L140

Their fields should be all optional.

ikatyang (changes requested):

types/prettier/index.d.ts
ChristianMurphy: Thanks @ikatyang! Updated :ballot_box_with_check:

ikatyang (approved):

PR #34831, opened 2019-04-18
Fix incorrect specification of `readPreference` and correct some `MongoClientOptions`

jloveridge:
Update `MongoClientOptions` properties to make them more correct.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://mongodb.github.io/node-mongodb-native/3.1/api/MongoClient.html#.connect and https://mongodb.github.io/node-mongodb-native/3.1/api/GridFSBucket.html

### DIFF
--- a/types/mongodb/index.d.ts
+++ b/types/mongodb/index.d.ts
@@ -22,6 +22,7 @@
 //                 Florian Richter <https://github.com/floric>
 //                 Erik Christensen <https://github.com/erikc5000>
 //                 Nick Zahn <https://github.com/Manc>
+//                 Jarom Loveridge <https://github.com/jloveridge>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
@@ -187,7 +188,7 @@ export interface SessionOptions {
 export interface TransactionOptions {
     readConcern?: ReadConcern;
     writeConcern?: WriteConcern;
-    readPreference?: any;
+    readPreference?: ReadPreference | string;
 }
 
 export interface MongoClientCommonOption {
@@ -235,7 +236,8 @@ export interface MongoClientOptions extends
     ReplSetOptions,
     SocketOptions,
     SSLOptions,
-    HighAvailabilityOptions {
+    HighAvailabilityOptions,
+    WriteConcern {
 
     /**
      * The logging level (error/warn/info/debug)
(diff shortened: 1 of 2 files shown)


types/mongodb/index.d.ts
```diff
@@ -277,6 +278,19 @@ export interface MongoClientOptions extends
      */
     useNewUrlParser?: boolean;
 
+    /** The write concern */
+    w?: number | string;
+    /** write concern timeout */
+    wtimeout?: number;
+    /** Specify a journal write concern (default: false) */
+    j?: boolean;
```
CaselIT: It's best to make the interface extend `WriteConcern`, so we avoid duplicating `w`, `wtimeout` and `j` parameters

CaselIT (changes requested):

types/mongodb/index.d.ts
jloveridge: I have updated the PR as you suggested. Since you felt it was best to extend `WriteConcern` for the client options I also did the same for `GridFSBucketWriteStreamOptions` which was also defining these same properties. And while the documentation for that (https://mongodb.github.io/node-mongodb-native/3.1/api/GridFSBucketWriteStream.html) indicates that `j` is a `number` instead of a `boolean` like it is everywhere else that documentation appears to be incorrect since the driver ends up calling into: `chunks.insertOne` passing in the options after transforming them through `getWriteOptions`. MongoDB's documentation for `collection.insertOne` (https://docs.mongodb.com/manual/reference/method/db.collection.insertOne/index.html) which should handle options the same way here as in any other place.

CaselIT (approved):

types/mongodb/index.d.ts
CaselIT: I did not suggest to `GridFSBucketWriteStreamOptions` since you had not changed it. But I agree that it is almost certainly an error with the docs

jloveridge: I realize you didn't suggest making the change to `GridFSBucketWriteStreamOptions`. I felt for the sake of consistency it was best to make the change there for the same reason as your suggestion related to the changes to `MongoClientOptions`. Just wish the docs were accurate. 

On a side note, if the MongoDB team ever authors their module in TypeScript, as they indicated they would be doing two years ago at MongoDB World, that would be awesome!

CaselIT: Or they should just migrate the code to typescript :)

PR #34839, opened 2019-04-18
[glob-watcher] Add glob-watcher@5.0.2 definition

TeamworkGuy2:
Please fill in this template.

Based off a combination of
 - [glob-watcher index.js](https://github.com/gulpjs/glob-watcher/blob/5681c11c80589e05387ffcd0b654ddc4bbfa70d5/index.js)
 - [chokidar types .d.ts](https://github.com/paulmillr/chokidar/blob/5e09c85ea04ffb7fd9cdb829892540e93ce63553/types/index.d.ts)
 - [gulp types .d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4f6bd1655718efee8a58936bb537848170cd6a93/types/gulp/index.d.ts)

### DIFF
--- a/types/glob-watcher/glob-watcher-tests.ts
+++ b/types/glob-watcher/glob-watcher-tests.ts
@@ -1,0 +1,60 @@
+import * as fs from 'fs';
+import * as globWatcher from 'glob-watcher';
+import * as Undertaker from 'undertaker';
+
+function test() {
+    const path = "";
+    const globs = ["", ""];
+
+    const opts1 = {
+    };
+
+    const opts2 = {
+        delay: 500,
+        events: ["add", "change", "unlink"],
+        depth: 1
+    };
+
+    optionsTypeTest(opts1);
+    optionsTypeTest(opts2);
+
+    const taskFunc: Undertaker.TaskFunction = () => "";
+
+    const cb = () => {
+        console.log("watch change!");
+    };
+
+    let cnt = 0;
+    const pCall = () => {
+        return new Promise<string>((resolve, reject) => resolve((++cnt).toString()));
+    };
+
+    const res = [
+        globWatcher(path),
+        globWatcher(globs),
+        globWatcher(path, opts1),
+        globWatcher(globs, opts2),
+        globWatcher(path, cb),
+        globWatcher(globs, cb),
+        globWatcher(path, opts1, cb),
+        globWatcher(globs, opts2, cb),
+        globWatcher(path, pCall),
+        globWatcher(globs, pCall),
+        globWatcher(path, opts2, pCall),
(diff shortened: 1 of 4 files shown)


types/glob-watcher/glob-watcher-tests.ts
```diff
@@ -0,0 +1,60 @@
+import * as fs from 'fs';
+import * as globWatcher from 'glob-watcher';
```
armanio123: namespaces are not collable. Instead use
```suggestion
import globWatcher = require('glob-watcher');
```

types/glob-watcher/index.d.ts
```diff
@@ -0,0 +1,167 @@
...
+
+        /**
+         * File size polling interval.
+         */
+        pollInterval?: number;
+    }
+}
+
+export = GlobWatcher;
+export as namespace GlobWatcher;
```
armanio123: I believe this is incorrect. The package doesn't provide a umd module or uses the global scope.

armanio123 (changes requested):

types/glob-watcher/glob-watcher-tests.ts
TeamworkGuy2: It's callable because the namespace shares the same name with the two function definition exports at the top of index.d.ts, line 16 and line 24.
The package being modeled is a single function export.

types/glob-watcher/index.d.ts
TeamworkGuy2: True, but I'd like to access the exported types, specifically line 58 of the test file: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34839/files#diff-2be02a26dd97ea05d84195b762a71911R58
How can I make `WatchOptions` accessible?

types/glob-watcher/glob-watcher-tests.ts
armanio123: Yeah I noticed that but it's a bad practice and comes with a couple of problems. You can read about it on the definitively typed [FAQ here](https://github.com/DefinitelyTyped/DefinitelyTyped#should-i-add-an-empty-namespace-to-a-package-that-doesnt-export-a-module-to-use-es6-style-imports). Also there's a very complete post explaining a little bit more about this [here](https://stackoverflow.com/questions/39415661/what-does-resolves-to-a-non-module-entity-and-cannot-be-imported-using-this).

types/glob-watcher/index.d.ts
armanio123: You should be fine by removing this line and using `import =` syntax on the test file.

armanio123 (approved):

PR #34844, opened 2019-04-19
Add typings for node-insights

wrumsby:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/node-insights/index.d.ts
+++ b/types/node-insights/index.d.ts
@@ -1,0 +1,85 @@
+// Type definitions for node-insights 1.1
+// Project: https://github.com/flightstats/node-insights#readme
+// Definitions by: Walter Rumsby <https://github.com/wrumsby>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
+export interface Config {
+    insertKey: string;
+    queryKey: string;
+    accountId: string;
+    timerInterval?: number;
+    maxPending?: number;
+    defaultEventType?: string;
+    enabled?: boolean;
+    gzip?: boolean;
+}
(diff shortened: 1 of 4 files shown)


jloveridge:
Not sure if this will present a problem to anyone but the `node-insights` module does not actually have an export called `default` as is implied in these types. It, as many other npm modules, simply exports the function directly. How would these typings affect someone regardless of their use of the `esModuleInterop` setting in `tsconfig.json`?

types/node-insights/index.d.ts
```diff
@@ -0,0 +1,85 @@
...
+    timerInterval?: number;
+    maxPending?: number;
+    defaultEventType?: string;
+    enabled?: boolean;
+    gzip?: boolean;
+}
+
+export type CallbackFunc = (err: Error | null | undefined, responseBody: string) => void;
+
+export default class Insights {
```
armanio123: As @jloveridge has mentioend, the `default` should be removed if there's no default export.

types/node-insights/node-insights-tests.ts
```diff
@@ -0,0 +1,25 @@
+import Insights from 'node-insights';
```
armanio123: Instead use `import x = require('module')` syntax. You can still use this style byt the `--allowSyntheticDefaultImports` or `--esModuleInterop` should be turned on instead.

armanio123 (changes requested):
Check this [documentation](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require) for details about the default export.

wrumsby:
@armanio123 changes made as requested.

armanio123 (approved):

PR #34845, opened 2019-04-19
Add new typing for electron-clipboard-extended

Jungwoo-An:
Please fill in this template.

If adding a new definition:

Thanks.

### DIFF
--- a/types/electron-clipboard-extended/index.d.ts
+++ b/types/electron-clipboard-extended/index.d.ts
@@ -7,7 +7,7 @@
 import * as Electron from 'electron';
 
 interface ClipboardExtended extends Electron.Clipboard {
-    off: (event: string) => void;
+    off: (event: string, listener?: () => void) => void;
     startWatching: () => void;
     stopWatching: () => void;
 }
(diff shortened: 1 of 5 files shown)


types/electron-clipboard-extended/index.d.ts
```diff
@@ -0,0 +1,17 @@
+// Type definitions for electron-clipboard-extended 1.x
+// Project: https://github.com/arjun-g/electron-clipboard-extended
+// Definitions by: Jungwoo An <https://github.com/Jungwoo-An>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.3
+
+import * as Electron from 'electron';
+
+interface ClipboardExtended extends Electron.Clipboard {
+    off: (event: string) => void;
```
armanio123: The optional `listener` parameter is missing. I don't know much about the library but by looking at the examples seems to be something like this:
```suggestion
    off: (event: string, listener?: () => void) => void;
```

armanio123 (changes requested):

Jungwoo-An:
@armanio123 Thanks.

armanio123 (approved):

PR #34846, opened 2019-04-19
[secure-password] Add definition

jloveridge:
Please fill in this template.

If adding a new definition:

Attempting to run `dtslint secure-password` always resulted in:
`Duplicate identifier 'prototype'.`

I was unable to resolve it so resorted to an older style module definition format that seems to work just fine.

### DIFF
--- a/types/secure-password/secure-password-tests.ts
+++ b/types/secure-password/secure-password-tests.ts
@@ -1,0 +1,14 @@
+import SecurePassword = require('secure-password');
+
+const sp = new SecurePassword();
+const spWithOpts = new SecurePassword({memlimit: 10, opslimit: 10});
+
+const clearTextBuf = Buffer.from('testing');
+
+const hashPromise = sp.hash(clearTextBuf).then((hash) => hash);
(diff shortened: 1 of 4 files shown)


jloveridge:
The build failure appeared to be due to a timeout issue. Added some comments to the function definitions and pushed up again, this time it passed the build so I don't really think it was anything I could have avoided.

types/secure-password/secure-password-tests.ts
```diff
@@ -0,0 +1,15 @@
+import * as SecurePassword from 'secure-password';
```
armanio123: namespaces are not callable, instead use: 
```suggestion
import SecurePassword = require('secure-password')
```

types/secure-password/tslint.json
```diff
@@ -0,0 +1,81 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "adjacent-overload-signatures": false,
```
armanio123: Please don't disable any rules unless there's a specific reason for it.

armanio123 (changes requested):

PR #34853, opened 2019-04-19
[@types/nodekell] Add type definition for nodekell

syrflover:
Please fill in this template.

If adding a new definition:


types/nodekell/tslint.json
```diff
@@ -0,0 +1,11 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "no-void-expression": false,
```
armanio123: Could you enable all rules or is there a reason on why they need to be disabled?

armanio123 (changes requested):

types/nodekell/tslint.json
syrflover: i'm use $ExpectType void in test

- no void expression
  - [fnothing](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34853/files#diff-3938434a7f31e1805d36128f4bf1c928R96)
  - [then return void](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34853/files#diff-3938434a7f31e1805d36128f4bf1c928R1423)
  - [then return promise void](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34853/files#diff-3938434a7f31e1805d36128f4bf1c928R1446)
  - [sleep](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34853/files#diff-3938434a7f31e1805d36128f4bf1c928R2250)

unified signatures force me use union type. so i'm set false.

max line length & array type makes hard to see the code.

if you talk to me "enable rules", I'll try enable rules.

syrflover: I don't want upload this type.

PR #24509, opened 2018-03-25
Add new RenderProps-style Context from React 16.3

scally:
React 16.3, the latest version, has a new recommended API for Context, with Context, Provider, and Consumer typings.

These changes are merged in this PR to React: 
https://github.com/facebook/react/pull/11818

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/reactjs/rfcs/blob/master/text/0002-new-version-of-context.md
- [x] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #34854, opened 2019-04-19
[@types/react] React.creatContext should receive a optional value

wszgxa:
Please fill in this template.

DOC: https://reactjs.org/docs/context.html#reactcreatecontext

And there is an [article](https://medium.com/@thehappybug/using-react-context-in-a-typescript-app-c4ef7504c858) with this issue, use a tricky way to avoid this problem.

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -336,7 +336,7 @@ declare namespace React {
         displayName?: string;
     }
     function createContext<T>(
-        defaultValue: T,
+        defaultValue?: T,
         calculateChangedBits?: (prev: T, next: T) => number
     ): Context<T>;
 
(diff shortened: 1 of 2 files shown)


eps1lon:
Isn't the problem with this that I can create a context with `const Context = createContext<SomeValue>()` and then `useContext(Context)` would return `SomeValue` instead of `SomeValue | undefined`? That doesn't sound desirable. I'd rather have opt-out of strict types instead of opt-in to strict types.

ferdaber:
@eps1lon 's reasoning is correct. We chose to explicitly not include `defaultValue` as optional even when considering the docs. The React team also does not consider `defaultValue` to be itself optional. Let me see if I can pull up the old PR...

ferdaber:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/24509#issuecomment-382213106

PR #34855, opened 2019-04-19
specified options for integer, sentence, word, email, url

Hakier:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://chancejs.com/ (Not new version, just those options were generic and I specified them according to docs)

### DIFF
--- a/types/chance/index.d.ts
+++ b/types/chance/index.d.ts
@@ -4,6 +4,7 @@
 //                 Brice BERNARD <https://github.com/brikou>
 //                 Carlos Sanchez <https://github.com/cafesanu>
 //                 Colby M. White <https://github.com/colbywhite>
+//                 Hakier <https://github.com/hakier>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 declare namespace Chance {
@@ -28,16 +29,16 @@ declare namespace Chance {
         bool(opts?: Options): boolean;
         character(opts?: Options): string;
         floating(opts?: Options): number;
-        integer(opts?: Options): number;
+        integer(opts?: {min: number, max: number}): number;
         letter(opts?: Options): string;
         natural(opts?: Options): number;
         string(opts?: Options): string;
 
         // Text
         paragraph(opts?: Options): string;
-        sentence(opts?: Options): string;
+        sentence(opts?: {words: number}): string;
         syllable(opts?: Options): string;
-        word(opts?: Options): string;
+        word(opts?: {syllables?: number, length?: number}): string;
 
         // Person
         age(opts?: Options): number;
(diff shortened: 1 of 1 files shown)


types/chance/index.d.ts
```diff
@@ -28,16 +29,16 @@ declare namespace Chance {
         bool(opts?: Options): boolean;
         character(opts?: Options): string;
         floating(opts?: Options): number;
-        integer(opts?: Options): number;
+        integer(opts?: {min: number, max: number}): number;
```
colbywhite: my reading of this is that, as you've typed it, you can't provide a min _without_ a max, which isn't the case. (the doc doesn't indicate it, but if you look at the source code or run it yourself, you'll see that you _can_ provide one without the other.)

you'll need to make the min/max optional like the other functions have.

```suggestion
        integer(opts?: {min?: number, max?: number}): number;
```

types/chance/index.d.ts
```diff
@@ -70,7 +71,7 @@ declare namespace Chance {
         color(opts?: Options): string;
         company(): string;
         domain(opts?: Options): string;
-        email(opts?: Options): string;
+        email(opts?: {domain: string}): string;
```
colbywhite: technically, doing the following is accepted in chance.

```javascript
new Chance().email({})
```

it's another slight of the doc that it's not clearly conveyed, but this is allowed. the way this is typed wouldn't allow for that, right? it's a weird edge case, but should be supported.

```suggestion
        email(opts?: {domain?: string}): string;
```

colbywhite (changes requested):
thanks for typing the params!! that's been a pet peeve of mine for a while and have been wanting to take a stab at it.

this j/ needs a couple small tweaks before it's 100% accurate.

> Add or edit tests to reflect the change. (Run with `npm test`.)

this little bit from the template is referring to, in this case, `chance-tests.ts`. none of the functions you've tweaked have tests for them in that file. doing so will allow you to show the edge cases that aren't allowed by they way it's currently typed. just add some usages of these methods in that file to ensure it compiles as you expect.

EDIT:

> Use a meaningful title for the pull request. Include the name of the package modified.

based on this tidbit from the template, your PR title should include the name `chance`. i used the angular convention and prepended a _fix(chance):_ to my PR. looking at the other PRs, there doesn't appear to be a definitive way to include the name, but it should still be included. I imagine this affects the changelog.

types/chance/index.d.ts
colbywhite: You know, now that I think about it a bit, the following is also technically allowed by chance.

```javascript
new Chance().email({a: 1})
```

obviously, that's a bad usage though so we shouldn't support that in the typing. so should we allow the empty object? the way you're typing the methods that have multiple options though, you are allowing the empty object for them. you're way prevents the weird empty object for methods with a _single_ param, which i'm starting to think is better. but my way makes it consistent across methods. not sure which is better. 🤔

PR #34861, opened 2019-04-19
Added explicit options to 'markdownlint'.

DanielRosenwasser:


types/markdownlint/index.d.ts
```diff
@@ -18,7 +18,233 @@ declare namespace markdownlint {
...
+
+        /**
+         * This rule is triggered when list items are not indented by the configured number of spaces.
+         * @default 2
+         */
+        "ul-indent"?: boolean | {
+            indent?: number;
+        };
+
+        "trailing-spaces"?: boolean | {
```
ark120202: I think it should be
```suggestion
        "no-trailing-spaces"?: boolean | {
```

ark120202 (approved):

PranavSenthilnathan:
Thanks for using the template!

PR #34890, opened 2019-04-21
Introduce type definitions for jest-expect-message

mike-d-davydov:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/jest-expect-message/index.d.ts
+++ b/types/jest-expect-message/index.d.ts
@@ -6,12 +6,8 @@
 
 /// <reference types="jest"/>
 
-export {};
-
-declare global {
-	namespace jest {
-        interface Expect {
-			<T = any>(actual: T, message: string): Matchers<T>;
-		}
+declare namespace jest {
+    interface Expect {
+        <T = any>(actual: T, message: string): Matchers<T>;
     }
 }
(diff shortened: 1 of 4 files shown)


plantain-00 (approved):

types/jest-expect-message/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+// Project: https://github.com/mattphillips/jest-expect-message#readme
+// Definitions by: Mike Davydov <https://github.com/mike-d-davydov>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.4
+
+/// <reference types="jest"/>
+
+export {};
+
+declare global {
```
PranavSenthilnathan: Can you try removing lines 9 and 11? By default if you have the remainder of the code, it will find the `jest` namespace in the global scope and augment it.

PranavSenthilnathan (changes requested):

types/jest-expect-message/index.d.ts
mike-d-davydov: Unfortunately, I can not, because it is exactly what I am trying to do: jest-expect message is a package that auguments jest `expect` by adding extra argument, it is the whole point. If you look at installation instructions for that package, they tell people to include jest-expect as Jest setup script. So if person uses that package, they are supposed to have an augumented "expect"

mike-d-davydov: Thanks! Looks like it works.

mike-d-davydov:
@PranavSenthilnathan , please kindly review.

PranavSenthilnathan (approved):

PranavSenthilnathan:

PR #33478, opened 2019-02-28
feat: Add `tape‑async`

ExE-Boss:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If adding a new definition:
- [x] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration`
- [x] If this is for an NPM package, match the name. If not, do not conflict with the name of an NPM package.
- [x] Create it with `dts-gen --dt`, not by basing it on an existing project.
- [x] `tslint.json` should be present, and `tsconfig.json` should have `noImplicitAny`, `noImplicitThis`, `strictNullChecks`, and `strictFunctionTypes` set to `true`.

---

[`tape‑async`](https://github.com/parro-it/tape-async) is a thin wrapper around `tape` that re‑exports all `tape` exports and allows the `TestCase` to return a `Promise` or be a generator function.

---

Thanks to @sandersn for helping me fix the build in https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33491#issuecomment-469433347.

PR #34892, opened 2019-04-21
fix(ssri): Return `Promise` type by default

ExE-Boss:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/zkat/ssri/blob/v6.0.1/README.md#api>

---

Since I’ve completely rewritten the tests to use `$ExpectType`, I’ve ended up reformatting them to use tabs.

I’ve also added union returning overloads for `parse(…)` and `fromHex(…)` until Microsoft/TypeScript#14107 is resolved.

### DIFF
--- a/types/ssri/index.d.ts
+++ b/types/ssri/index.d.ts
@@ -8,141 +8,174 @@ import { Utf8AsciiLatin1Encoding, Hash as CryptoHash } from "crypto";
 import { Readable, Transform } from "stream";
 
 export interface HashLike {
-    algorithm: string;
-    digest: string;
-    options?: string[];
+	algorithm: string;
+	digest: string;
+	options?: string[];
 }
 
 export interface IntegrityLike {
-    [algorithm: string]: HashLike[];
+	[algorithm: string]: HashLike[];
 }
 
 export type IntegrityMap = Integrity & IntegrityLike;
 
 export class Hash implements HashLike {
-    constructor(hash: string, opts?: { strict?: boolean });
+	constructor(hash: string, opts?: { strict?: boolean });
 
-    source: string;
-    algorithm: string;
-    digest: string;
-    options?: string[];
-    isHash: boolean;
+	source: string;
+	algorithm: string;
+	digest: string;
+	options?: string[];
+	isHash: boolean;
 
-    hexDigest(): string;
+	hexDigest(): string;
 
-    toJSON(): string;
+	toJSON(): string;
 
-    toString(opts?: { strict?: boolean }): string;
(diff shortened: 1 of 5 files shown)


huan086:
Could you keep the changes to the minimum by using 4 spaces instead of tabs? Majority of the types uses spaces. I found tabs in 22% of the packages, others uses spaces.

For comma dangle, I think it makes sense only for multi-line arguments. Changes on the same line will always show up in diff even with comma dangle.

ExE-Boss:
You can use the `?w=1` query parameter to hide whitespace‑only changes in the diff: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34892/files?w=1

huan086 (approved):

types/ssri/index.d.ts
```diff
@@ -1,126 +1,181 @@
...
+): IntegrityMap | Hash;
 
 export function stringify(
-    obj: string | IntegrityLike | HashLike,
-    opts?: { strict?: boolean, sep?: string }
+	obj: string | IntegrityLike | HashLike,
+	opts?: { strict?: boolean, sep?: string, },
 ): string;
 
 export function fromHex(
```
PranavSenthilnathan: Another option is to just have one function like:
```
export function fromHex<T extends boolean = false>(
  hexDigest: string, 
  algorithm: string, 
  opts?: { 
    single?: T, 
    string?: boolean, 
    options?: ReadonlyArray<string>
  }
): T extends true ? Hash : IntegrityMap;
```

PranavSenthilnathan (changes requested):

types/ssri/index.d.ts
ExE-Boss: That would bump the required TypeScript version to 2.8.

types/ssri/tslint.json
```diff
@@ -1 +1,9 @@
-{ "extends": "dtslint/dt.json" }
+{
+	"extends": "dtslint/dt.json",
+	"rules": {
```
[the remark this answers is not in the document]
ExE-Boss: @sandersn has approved me doing this in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33478#issuecomment-470728638.

types/ssri/index.d.ts
PranavSenthilnathan: The TypeScript version should really be specified in the top of the d.ts file if people are going to be relying on a fixed version.

PranavSenthilnathan (approved):

PR #34899, opened 2019-04-21
[@types/react] Improve ComponentProps handling of class components

s-h-a-d-o-w:
Wrapping `P` in `ReactManagedAttributes` generates a type where `defaultProps` of a class component are taken into consideration.

Please fill in this template.

The code that made me discover this problem is closed source.
But in addition to the included test, I can share this piece of dummy code that does not work with the current types:
```ts
type Props = {
  foo: () => number;
  bar: number
};
class MyComponent extends React.Component<Props> {
  static defaultProps = {
    foo: () => 123,
  };

  render() {
    return <div>{this.props.foo() + this.props.bar}</div>;
  }
}

type MyComponentProps = React.ComponentProps<typeof MyComponent>;
// TS2741: Property 'foo' is missing in type '{ bar: number; }' but required in type 'Props'.
const props: MyComponentProps = { bar: 123 };
```

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -765,7 +765,7 @@ declare namespace React {
      */
     type ComponentProps<T extends keyof JSX.IntrinsicElements | JSXElementConstructor<any>> =
         T extends JSXElementConstructor<infer P>
-            ? P
+            ? ReactManagedAttributes<T, P>
             : T extends keyof JSX.IntrinsicElements
                 ? JSX.IntrinsicElements[T]
                 : {};
--- a/types/react/test/ComponentProps.tsx
+++ b/types/react/test/ComponentProps.tsx
@@ -1,0 +1,43 @@
+import React = require("react");
+
+// Function/class components
+// ---------------------------------------
+interface FunctionComponentProps {
+    foo?: number;
+    bar: number;
+}
+const FunctionComponent: React.FC<FunctionComponentProps> = ({
+    foo = 123,
+    bar
+}) => {
+    return <div>{foo + bar}</div>;
+};
+
+interface ClassComponentProps {
+    foo: number;
+    bar: number;
+}
+class ClassComponent extends React.Component<ClassComponentProps> {
+    static defaultProps = {
+        foo: 123,
+    };
+
+    render() {
+        return <div>{this.props.foo + this.props.bar}</div>;
+    }
+}
+
+const fcProps: React.ComponentProps<typeof FunctionComponent> = { bar: 123 };
+const ccProps: React.ComponentProps<typeof ClassComponent> = { bar: 123 };
+
+// Intrinsic elements
+// ---------------------------------------
+type ImgProps = React.ComponentProps<'img'>;
+// $ExpectType "async" | "auto" | "sync" | undefined
+type ImgPropsDecoding = ImgProps['decoding'];
+type ImgPropsWithRef = React.ComponentPropsWithRef<'img'>;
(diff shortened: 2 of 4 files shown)


ferdaber:
Hmm I'm not sure that this is a great idea. This isn't really the intended case for `ComponentProps`, it's supposed to grab the internal, or inner props of a component, not its API. Maybe instead of changing the current type, you could create another type like `ApparentProps` or `OuterComponentProps`?

s-h-a-d-o-w:
Hm, then I suppose I should submit issues about [this](https://github.com/sw-yx/react-typescript-cheatsheet#the-types-i-need-werent-exported) and [this](https://github.com/piotrwitek/react-redux-typescript-guide#reactcomponentpropstypeof-component). Because what they are describing is getting the API and they should be notified that that's not what `ComponentProps` is intended to do.

Out of curiosity - what is the use case for getting the internal props?

ferdaber:
Consider the use case of `React.memo` where it is used. It should be targeting the inner props because we want to ensure that within the implementation of a component, a prop that has a default is always going to be defined in the comparison function.

s-h-a-d-o-w:
Ah... I see, thanks!

Naming... hm... while I appreciate that you want my input, I suppose in the end, this should be decided by you and/or other core contributors because I think chances are that people will end up using this particular type a **lot** once they realize that they don't have to (or maybe even shouldn't) export prop types by default.

My two cents on this topic are - personally, when names are conflicting, I prefer to use the more concise one for the thing that is user-facing (or at least likely to be used more often by users). But I guess to preserve backwards compatibility, that's not an option here. From your suggestions, `Outer...` seems the better one to me but I also think that something like `Defaulted...`, `Defaultized...` or `...WithDefaults` might convey better what the difference to regular `ComponentProps` is. Of course a counter argument to this would be that the naming should not be related to implementation details.

eps1lon:
I'm honestly a bit confused about this too. If `ComponentProps` is not used to describe the API it shouldn't be exported. 

It seems like this only works the way it does currently to support React.memo which is even more reason to use a specialized helper. I don't think there is other user facing code that is executed after default props are resolved?

ferdaber:
A lot of these were never intended to be exported. As maintainers, we've been a bit cavalier on putting things inside the `React` namespace, and since the entire export is the namespace all of these are accessible, but some types are were never really intended to be public API, this one included.

eps1lon:
Memo and defaultProps won't matter soon anyway because:

> We'd add a warning in createElement if something that doesn't have .prototype.isReactComponent on it uses defaultProps. This includes other special components like forwardRef and memo.

-- https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#deprecate-defaultprops-on-function-components

s-h-a-d-o-w:
Interesting. But not everybody has migrated their code bases to use hooks yet (and maybe some just don't want to), so doesn't it still make sense to support class components?

eps1lon:
I know `React.memo` supports class components but why not just use `React.PureComponent` or `React.Component` + sCU in those cases?

I was referring to `ComponentProps` not resolving defaultProps to have non-nullable types in the comparison function. Once defaultProps is no longer resolved for `React.memo` `ComponentProps` can resolve `defaultProps` safely without breaking `React.memo` because `defaultProps` will only be used on class components.

s-h-a-d-o-w:
For some reason, I thought you referred to memo and defaultProps as separate things. Never mind me. :)

PR #34900, opened 2019-04-22
Deprecate types for packages that provide their own types

BendingBender:
If removing a declaration:


YashdalfTheGray (commented):
My approval (for right now) is limited to `is-docker` because that's the only one I have context over. I'll come back in a few days if this moves nowhere to check others and help where I can.

Maistho (approved):
I've looked through each repository linked in the changes from `notNeededPackages.json` and concluded that each repository has typings in the root folder.

PR #34907, opened 2019-04-22
fix(treeify): types

yxliang01:
- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/notatestuser/treeify . Note: the documentation didn't specify that the `hideFunctions` argument for method `asTree` is optional, it is however optional. The documentation doesn't also explicitly state that it supports `Array` object, it instead just use placeholder `obj` which can be an `Array`.
- [X] Tested locally

Changes:
- Fixed previously only `object` can be supplied as argument for printing
- Made `TreeObject` attributes `readonly`
- Fixed `asTree` `hideFunctions` argument that it should be optional

### DIFF
--- a/types/treeify/index.d.ts
+++ b/types/treeify/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for treeify 1.1.0
+// Type definitions for treeify 1.1
 // Project: https://github.com/notatestuser/treeify
 // Definitions by: Mike North <https://github.com/mike-north>
 //                 Xiao Liang <https://github.com/yxliang01>
--- a/types/treeify/index.d.ts
+++ b/types/treeify/index.d.ts
@@ -1,7 +1,7 @@
 // Type definitions for treeify 1.1
 // Project: https://github.com/notatestuser/treeify
 // Definitions by: Mike North <https://github.com/mike-north>
-// Definitions by: Xiao Liang <https://github.com/yxliang01>
+//                 Xiao Liang <https://github.com/yxliang01>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.1
 
--- a/types/treeify/index.d.ts
+++ b/types/treeify/index.d.ts
@@ -1,27 +1,29 @@
-// Type definitions for treeify 1.0
+// Type definitions for treeify 1.1.0
 // Project: https://github.com/notatestuser/treeify
 // Definitions by: Mike North <https://github.com/mike-north>
+// Definitions by: Xiao Liang <https://github.com/yxliang01>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.1
 
 export interface TreeObject {
-    [k: string]: TreeValue;
+    readonly [k: string]: TreeValue;
 }
-export type TreeValue = string | TreeObject;
+export type TreeValue = string | PrintObject;
+export type PrintObject = TreeObject | ReadonlyArray<any>;
 
 export function asTree(
-    treeObj: TreeObject,
+    printObj: PrintObject,
     showValues: boolean,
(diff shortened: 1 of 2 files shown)


yxliang01:
Hmm how do I locate where the error is? The CI logs don't really point that out.

mike-north (changes requested):
Thanks for your contribution @yxliang!

Could you fill out the pull request template for this change? I need some of that information in order to review.
- Links to where I can find public API documentation to support each substantive change you're proposing (function arg renaming isn't something I'm too worried about)
- Whether or not you've tested this in a real app or not

mike-north:
> Hmm how do I locate where the error is? The CI logs don't really point that out.

[The error](https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/523034980#L471) relates to the DT header parser having trouble with the changes you've made to the header comments at the top, since they [only expect to find a major and minor package version](https://github.com/Microsoft/definitelytyped-header-parser/blob/f5c301b88d72b93eadd0d2a46e52985eac5d94ec/index.ts#L75-L76)

yxliang01:
Will modify the issue soon.

mike-north (changes requested):
Not all review feedback has been acted on yet

yxliang01:
@mike-north Modified PR description as per your comment https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34907#pullrequestreview-229373213 .

mike-north:
> Note: the documentation didn't specify that the hideFunctions argument for method asTree is optional, it is however optional. 

> The documentation doesn't also explicitly state that it supports Array object, it instead just use placeholder obj which can be an Array.

@yxliang01 How do we know whether these things just happen to work for now, or whether they're part of the public API that the maintainers intend to support?

Are there tests that assert these cases work properly now, and will continue to work with future changes?

yxliang01:
@mike-north Nope. No test supporting the type changes in this PR. We know it works for now by trying. No way to guarantee it will be working in the future versions as the inability to predict the future nor controlling `treeify`'s future. As suggested above, the documentation didn't mention what the object to be passed to be like, that includes it doesn't mention passing an array is supported. 

Things to note though:
1. In one example provided in documentation, `treeify.asTree` is called without supplying the `hideFunctions` value.
2. The type file header specifies the targeted version. Why do we need to assure about the future versions?

yxliang01:
I am currently awaiting for @mike-north 's response.

mike-north:
> Nope. No test supporting the type changes in this PR. We know it works for now by trying. No way to guarantee it will be working in the future versions as the inability to predict the future nor controlling treeify's future. As suggested above, the documentation didn't mention what the object to be passed to be like, that includes it doesn't mention passing an array is supported.

These are all indications that these changes are not necessarily in alignment with the project's public API

> In one example provided in documentation, treeify.asTree is called without supplying the hideFunctions value.

However, the documentation at the bottom doesn't suggest that `hideFunctions` is optional. We have conflicting signals here and need to straighten that out before we can be confident about the library's public API

> The type file header specifies the targeted version. Why do we need to assure about the future versions?.

Without knowing whether something is part of the public API, we don't know whether breaking/removing it would be accompanied by a [major version bump](https://docs.npmjs.com/about-semantic-versioning)

> What is the difference between an object between an Array? At the end of the day, an Array is an object? In that sense, a parameter of type object should accept any JS value that is not undefined nor null? (correct me if wrong)

I think you're referring to your use of `ReadonlyArray<any>` here?

The goal of these `@types/*` packages is to represent the public API of each library as accurately as possible. This does not mean that we should type according to what _happens to work_, but instead _what the authors/maintainers intend to officially support_. I don't see a readonly `any[]` as being a good representation of the public API of this library, as it allows for consumers to assume things that simply aren't going to happen (i.e., an array of `null`, an array of `true`s, etc...)

Issue #30740, opened 2018-11-21
[@types/react-redux] 6.0.10 fails to resolve thunks


rhys-vdw:
- [x] I tried using the `@types/react-redux` package and had problems.
- [x] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @surgeboris

Hi @surgeboris, I see that your recent changes intend is to "properly support thunks". However they seem to have had the exact opposite effect in our code base!

This code works in 6.0.9:

```ts
export const OrderForm = connect(
  (state: State) => ({
    order: getOrder(state),
    initialOrder: Order.createEmpty(),
    isConfirming: isConfirming(state),
    externalQuote: getExternalQuote(state),
    panelQuote: getPanelQuote(state),
  }),
  {
    onCancel: cancelOrder,
    fetchExternalQuote,
    fetchPanelQuote,
    fetchQuoteQueued,
    confirmOrder,
    createOrder,
  }
```

The new error in 6.0.10:

```
Argument of type 'typeof OrderForm' is not assignable to parameter of type 'ComponentType<Matching<{ order: Readonly<PermittedOrderParams>; initialOrder: PermittedOrderParams; isConfirming: boolean; externalQuote: Readonly<ThirdPartyQuote>; panelQuote: Readonly<PanelQuote>; } & { ...; }, Props>>'.
  Type 'typeof OrderForm' is not assignable to type 'ComponentClass<Matching<{ order: Readonly<PermittedOrderParams>; initialOrder: PermittedOrderParams; isConfirming: boolean; externalQuote: Readonly<ThirdPartyQuote>; panelQuote: Readonly<PanelQuote>; } & { ...; }, Props>, any>'.
    Types of parameters 'props' and 'props' are incompatible.
      Type 'Matching<{ order: Readonly<PermittedOrderParams>; initialOrder: PermittedOrderParams; isConfirming: boolean; externalQuote: Readonly<ThirdPartyQuote>; panelQuote: Readonly<PanelQuote>; } & { ...; }, Props>' is not assignable to type 'Readonly<Props>'.
        Types of property 'confirmOrder' are incompatible.
[...]

PR #34913, opened 2019-04-22
react-redux: add types for hooks which will be added in version 7.1

MrWolfZ:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/reduxjs/react-redux/tree/v7-hooks-alpha/src/hooks

I am the author of the react-redux alpha hooks implementation (which I prototyped in TypeScript). These type definitions are mostly taken from my own source code.

### DIFF
--- a/types/react-redux/index.d.ts
+++ b/types/react-redux/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for react-redux 7.0
+// Type definitions for react-redux 7.1
 // Project: https://github.com/reduxjs/react-redux
 // Definitions by: Qubo <https://github.com/tkqubo>,
 //                 Kenzie Togami <https://github.com/kenzierocks>,
@@ -12,6 +12,7 @@
 //                 Anatoli Papirovski <https://github.com/apapirovski>
 //                 Boris Sergeyev <https://github.com/surgeboris>
 //                 Søren Bruus Frank <https://github.com/soerenbf>
+//                 Jonathan Ziller <https://github.com/mrwolfz>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
@@ -38,6 +39,7 @@ import {
 import {
     Action,
     ActionCreator,
+    ActionCreatorsMapObject,
     AnyAction,
     Dispatch,
     Store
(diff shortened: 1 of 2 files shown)


types/react-redux/index.d.ts
```diff
@@ -445,3 +447,240 @@ export const ReactReduxContext: Context<ReactReduxContextValue>;
...
+ *   return (
+ *     <div>
+ *       <span>{value}</span>
+ *       <button onClick={increaseCounterByOne}>Increase counter by 1</button>
+ *     </div>
+ *   )
+ * }
+ */
+export function useActions<
+    T extends [ActionCreator<A>, ActionCreator<A>, ActionCreator<A>, ActionCreator<A>, ActionCreator<A>, ActionCreator<A>, ActionCreator<A>, ActionCreator<A>, ActionCreator<A>],
```
threehams: I'm still running tests locally, but we can likely replace all these overloads with a single `T extends ReadonlyArray<readonly ActionCreator<A>>`. That prevents the action creator tuple from getting widened to an array.

types/react-redux/index.d.ts
```diff
@@ -445,3 +447,240 @@ export const ReactReduxContext: Context<ReactReduxContextValue>;
...
+    A = AnyAction
```
thebrodmann: I think there is no need for the `A = AnyAction`. Actually, this generic type looks useless. Also, it causes problems with some Redux middlewares like redux-thunk.

the-dr-lazy (changes requested):

types/react-redux/index.d.ts
thebrodmann: Typescript couldn't infer the type of data in every index without tuples **for now**. 

```ts
// With action creator tuples
useAction([a, b]) //=> [A, B]

// Without action creator tuples
useAction([a. b]) //=> (A | B)[]
```

threehams: Looks like it only works properly on TypeScript 3.4 and above. Too bad, but it's a good option for a couple months from now...

thebrodmann: Is it possible to release a major version that has a peer dependency on typescript@^3.4 now?

MrWolfZ: The compatibility with TypeScript 3.0 is exactly the reason why I typed it this way. I think these typings are just bearable enough that I wouldn't want to bump the peer dependency just for this. In a couple of months it's an easy fix.

types/react-redux/index.d.ts
MrWolfZ: The reason I did it this way is to be similar to how `bindActionCreators` from redux is [typed](https://github.com/reduxjs/redux/blob/da3d89fc229843c6514d29fbb204fc073359fd54/index.d.ts#L469). I don't think it hurts to have that extra generic parameter since in pretty much all cases you are not going to specify any of the parameters anyways. However, as you say, it can easily be replaced with `any` or `AnyAction`, so if you still think I should remove it, I will do so.

thebrodmann: Let's have a minimal implementation for it without extra generics. Also, it is important to don't restrict action creator return type to `AnyAction` for compatibility with thunks or other similar middlewares.

thebrodmann: Please see [MrWolfZ/DefinitelyTyped#1](https://github.com/MrWolfZ/DefinitelyTyped/pull/1).

MrWolfZ: I have merged the PR.

the-dr-lazy (approved):

types/react-redux/index.d.ts
threehams: I think we need feedback from more people on this. Redux bound actions are confusing to type (I would say nearly impossible) since the return value is determined by the middleware you're running. We can't solve every case of this, but in the most basic case of this, setting this to `any` will cause a lot of errors since we can't do conditional checking between action creators (which return an object) and thunks (which return whatever the thunk returns).

The existing typings have built-in support for `redux-thunk`, for instance, which is missing from these types:
```
TActionCreator extends (
    ...args: infer TParams
  ) => (...args: any[]) => infer TReturn
    ? (...args: TParams) => TReturn
    : TActionCreator
```

@surgeboris  and @rhys-vdw had to resolve a middleware-related bug/conflict in https://github.com/DefinitelyTyped/DefinitelyTyped/issues/30740, so they may have a specific interest here.

threehams: I'll work on a PR tonight to allow any function as an action, but still support thunk return values. That should cover the simplest cases. It shouldn't be anywhere near as difficult as `connect()` has been.

I'll also add some tests so we're not guessing anymore.

types/react-redux/index.d.ts
```diff
@@ -445,3 +447,231 @@ export const ReactReduxContext: Context<ReactReduxContextValue>;
...
+export function useActions<
+    T extends [ActionCreator<any>, ActionCreator<any>, ActionCreator<any>, ActionCreator<any>, ActionCreator<any>, ActionCreator<any>]
+>(actions: T, deps?: ReadonlyArray<any>): T;
+export function useActions<T extends [ActionCreator<any>, ActionCreator<any>, ActionCreator<any>, ActionCreator<any>, ActionCreator<any>]>(actions: T, deps?: ReadonlyArray<any>): T;
+export function useActions<T extends [ActionCreator<any>, ActionCreator<any>, ActionCreator<any>, ActionCreator<any>]>(actions: T, deps?: ReadonlyArray<any>): T;
+export function useActions<T extends [ActionCreator<any>, ActionCreator<any>, ActionCreator<any>]>(actions: T, deps?: ReadonlyArray<any>): T;
+export function useActions<T extends [ActionCreator<any>, ActionCreator<any>]>(actions: T, deps?: ReadonlyArray<any>): T;
+export function useActions<T extends [ActionCreator<any>]>(actions: T, deps?: ReadonlyArray<any>): T;
+export function useActions<T extends ReadonlyArray<ActionCreator<any>>>(actions: T, deps?: ReadonlyArray<any>): T;
+export function useActions<T extends ActionCreatorsMapObject>(actions: T, deps?: ReadonlyArray<any>): T;
```
threehams: The object form has the same issue of not supporting thunks (or any random middleware-allowed action creator) as the other forms. I can take care of that tonight along with the other thunk changes.

threehams (changes requested):

types/react-redux/index.d.ts
rhys-vdw: Hey @threehams, just responding to the mention. Don't think I've got anything of value to add here as I haven't really dug into hooks yet, but thanks for the heads up. I'll keep an eye on the issue nonetheless.

threehams:
Sorry, I have no idea how to do a pull request into your fork (doesn't seem to let me compare?)

This commit supports thunk return values, and adds test coverage for all hooks:
https://github.com/threehams/DefinitelyTyped/commit/297a2e379b2473d1ba96c08ee84953f14d0313bb

MrWolfZ:
@threehams thanks, I have tried manually merging your changes into my fork. However, I get errors during linting:

```
ERROR: 1338:15  expect                   Compile error in typescript@3.0 but not in typescript@3.1.
Fix with a comment '// TypeScript Version: 3.1' just under the header.
Type 'ResolveArrayThunks<[(selected: boolean) => { type: string; payload: boolean; }, ({ items }: { items: string[]; }) => { type: string; payload: { items: string[]; }; }, () => (dispatch: Dispatch<AnyAction>) => { type: string; payload: never[]; }]>' is not an array type.
ERROR: 1354:15  expect                   Compile error in typescript@3.0 but not in typescript@3.1.
Fix with a comment '// TypeScript Version: 3.1' just under the header.
Type 'ResolveArrayThunks<[(selected: boolean) => { type: string; payload: boolean; }]>' is not an array type.
ERROR: 1462:13  expect                   Compile error in typescript@3.0 but not in typescript@3.1.
Fix with a comment '// TypeScript Version: 3.1' just under the header.
Type 'ResolveArrayThunks<[(selected: boolean) => { type: string; payload: boolean; }, ({ items }: { items: string[]; }) => { type: string; payload: { items: string[]; }; }, () => (dispatch: Dispatch<AnyAction>) => { type: string; payload: never[]; }]>' is not an array type.
```

I think the issue is with mapped types and tuples in 3.0. Therefore, we can either bump the TS version to 3.1, or we need to create a version of `ResolveTupleThunks` for each overload.

MrWolfZ:
Oh, and another thing I noticed is that `useDispatch` doesn't allow dispatching thunks, but that because the typing of `Dispatch` itself only allows actions to be dispatched. Any idea of how to work around this?

thebrodmann:
> Any idea of how to work around this?

I think this should be defined in redux-thunk type definitions by overloading the `Dispatch` interface. Am I right?

threehams:
Let's bump the version to 3.1. It was released six months ago, so that's 3.5 years in frontend years.

redux-thunk dispatch... There's an open discussion in redux-thunk about typings for it, since they're in the repo. For consistency we should probably make it work just like `useActions` and `useRedux` (assume redux-thunk is enabled), but I'll invite them for input.

the-dr-lazy (approved):

MrWolfZ:
I rebased the PR on latest master. But even without it as you can see the build is failing due to a dependency from mirrorx on react-redux requiring TypeScript 3.0. We could of course bump their TypeScript version to 3.1 as well, but I am not sure if it will end with that one dependency. Should we instead just make the types compatible with 3.0 after all?

threehams:
What's your timeline on merging? I'd like to take another pass on making these types more easily extensible for other middleware.

One thing I wasn't able to figure out: how to have everything trace back to Redux's `Dispatch` type. I don't see how the Redux `Dispatch` interface can be extended or overloaded to support more complex middleware, given its single generic argument:

```
export interface Dispatch<A extends Action = AnyAction> {
  <T extends A>(action: T): T
}
```

Middleware allows `object | (...args[]) => whatever random thing`. Best I can think of is to handle thunks internally to these typings, but that doesn't cover sagas, or observables, or whatever could exist out there.

thebrodmann:
> I don't see how the Redux Dispatch interface can be extended or overloaded to support more complex middleware

I think the best way to have typings for middlewares is that every middleware overload it's needed Redux and react-redux interfaces internally.

I can't figure out why redux-thunk doesn't overload the `Dispatch` interface in the following way:

```ts
declare module "redux" {
  export interface Dispatch<A extends Action = AnyAction> {
    <TAction extends A>(thunk: (dispatch: Dispatch) => TAction): TAction;
  }
}
```

There is a working example of the above code on [codesandbox](https://codesandbox.io/s/18n0molz34).

Although in the past I said that it would be great to support thunks in the hooks typings but now I don't think that handling any middlewares typings in react-redux be a good idea.

Alternatively, **we can export some interfaces like `Dispatch` in Redux that could be overloaded by middlewares upon their needs**.

> that doesn't cover sagas, or observables

Sagas or observables don't need to be dispatched. So they don't need any changes in types of hooks or `Dispatch`.

threehams:
I've been trying to find some way of avoiding the second case, where every middleware library contains overloads for every overload in `useActions`. 

To generalize ResolveThunks to work with any middleware, we just need to convert an input action creator (which can be anything valid for overloaded Dispatch) into a function with the same signature, but the return type of the result of running that action through `Dispatch`:

```typescript
ReturnType<Dispatch<ReturnType<TActionCreator>>>
```

I really wish that worked, but I'm hitting the limits of what I understand about declaration merging:
https://codesandbox.io/s/2wz16qzpky

In this example, `store.dispatch` works just fine, as it gets the overloaded `Dispatch` signature. However, the `Dispatch` type imported from `"redux"` only contains the base interface. If we had some way of collecting together all the overloads added by all middleware, this would just work, and middleware would only need to over overload `Dispatch`. This would make all those libraries less coupled to `react-redux`'s definitions.

keithlayne:
A new alpha was published today, minus `useRedux`.  @MrWolfZ I submitted MrWolfZ/DefinitelyTyped#2 but that was mostly me figuring out how to make a PR from one fork to another :)  It's just a bunch of deletions and a fixed typo.

keithlayne:
So another [alpha release](https://github.com/reduxjs/react-redux/releases/tag/v7.1.0-alpha.4) has come out, removing `useActions` in favor of `useDispatch`.  Hopefully this simplifies things for the typings, but I can imagine this moving some ot the typing problems toward the end user.

MrWolfZ:
@keithlayne yeah, I'll hold off removing the typings for `useActions` until it is clear that the hook is gone for good (I am lobbying for it to be added back again)

keithlayne:
Yeah, that's my bad, I should have checked the thread before posting here.  I saw that whole conversation afterward.

threehams:
Looks like I was completely wrong about dispatch being any better for middleware typing; when we `import { Dispatch } from "redux"` it's still not the type extended by redux-thunk, for instance. `bindActionCreators` + `dispatch` was working by pure coincidence as `dispatch` was becoming `any` in my local project.

Same issue applies as before: solve extensibility of `useDispatch()` and we solve `useActions()`.

threehams:
This PR changes `useDispatch` to be based off Redux's `Dispatch` type.
https://github.com/MrWolfZ/DefinitelyTyped/pull/3

redux-thunk PR which adds the Dispatch overload:
https://github.com/reduxjs/redux-thunk/pull/247

the-dr-lazy (approved):

jamsch:
Looks like with the alpha release of v7.1.0-alpha.4, [`useActions`](https://react-redux.js.org/next/api/hooks#removed-useactions) will be gone for good. Not sure if it will come back.

threehams:
A note on my PRs: I'm pretty discouraged by the direction the hooks API is going (in addition to coordination difficulties across middleware libraries) and am going to be stepping back from this. I would like to help make these typings usable out of the box with thunks at the least, but I don't think attempting to have PRs in 2+ libraries is a manageable solution. We might want to just throw thunk support into the typings with a note about how it's not ideal.

adberard:
How can we use this pull request in a react-redux 7.1.0-alpha.* project ?

So far I am manually patching ./node_modules/@types/react-redux with that repo types on each install.

I tryed using directly the`"@types/react-redux": "DefinitelyTyped/DefinitelyTyped#34913/head"` in package.json, is downloading the whole definitelyTyped into  ./node_modules/@types/react-redux and ts is still not aware of the types declarations

malj:
@adberard Create a `react-redux.d.ts` file, add its parent folder to [`compilerOptions.typeRoots`](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types) in `tsconfig.json`, and paste the additions from the pull request you want to use:

```typescript
import { Action, Dispatch, AnyAction } from 'redux'

declare module 'react-redux'
{
    export function useDispatch<A extends Action = AnyAction>(): Dispatch<A>
    export function useSelector<TState, TSelected>(selector: (state: TState) => TSelected, deps?: ReadonlyArray<any>): TSelected
    // ... other additions from the pull request
}
```

MrWolfZ:
I have updated the typings for the latest API changes. I have also added an overload for `useDispatch` that allows at least a somewhat reasonable approach to get correct typings (see [here](https://codesandbox.io/s/reactredux-usedispatch-redux-thunk-8u77p) for a usage example).

adberard:
> I have updated the typings for the latest API changes. I have also added an overload for `useDispatch` that allows at least a somewhat reasonable approach to get correct typings (see [here](https://codesandbox.io/s/reactredux-usedispatch-redux-thunk-8u77p) for a usage example).

the codesandbox produce an error stating
 ```
Invariant Violation
Invalid hook call. Hooks can only be called inside of the body of a function component.
```

**Comment by MrWolfZ** [comment]
@adberard yeah, that is intentional. This example is not supposed to show working code, but _compiling_ code. Therefore please also ignore the eslint parsing error.

**Comment by Cretezy** [comment]
Is there a reason why `deps` was removed from `useSelector` 2 days ago? It seems to still be part of the API: https://react-redux.js.org/next/api/hooks#useselector

**Comment by jamsch** [comment]
> Is there a reason why `deps` was removed from `useSelector` 2 days ago? It seems to still be part of the API: https://react-redux.js.org/next/api/hooks#useselector

Looks like it was removed as of 7.1.0-alpha.5
https://github.com/reduxjs/react-redux/commit/a787aeeb5ab064e66ff168eedbb962d3397b378d

**Comment by Cretezy** [comment]
Ah you're totally right. They didn't publish that version in that case, since docs (and possibly npm) aren't updated

**Comment by markerikson** [comment]
@threehams : just out of curiosity, can you clarify your comment on "discouraged by the direction the hooks API is going" ?

**Comment by threehams** [comment]
I'll post in the main feedback channel later tonight, and sorry for the offhand remark in here. I probably would have left that out if I were in a better general state of mind when I left that comment.

**Comment by markerikson** [comment]
@threehams : sure, please leave a comment in the hooks feedback issue, or ping me directly over in Reactiflux (`@acemarke`).

File: types/react-redux/index.d.ts [hunk 4]
```diff
@@ -445,3 +466,91 @@ export const ReactReduxContext: Context<ReactReduxContextValue>;
  * multiple actions dispatched outside of React only result in a single render update.
  */
 export function batch(cb: () => void): void;
+
+/**
+ * Compares two arbitrary values for shallow equality. Object values are compared based on their keys, i.e. they must
+ * have the same keys and for each key the value must be equal according to the `Object.is()` algorithm. Non-object
```
markerikson: We're not actually using `Object.is()` or its equivalent right now - it's just a straight-up reference `===`.

MrWolfZ: @markerikson Are you maybe confusing this with the `useSelector` hook? The `shallowEqual` function is using `Object.is()` as you can see [here](https://github.com/reduxjs/react-redux/blob/v7-hooks-alpha/src/utils/shallowEqual.js).

markerikson: Whoops, yeah :)

renchap:
Similar to how you can define you own `useDispatch` hook when you use `redux-thunk` to have proper typing, I think it would be useful to have an easy and documented way to have your own `useSelector` that you can export along your `store` and custom `useDispatch`, so you do not need to retype the `state` arg each time you use `useSelector`.

For example:

```typescript
// Current version
import { useSelector } from "react-redux"
import { RootState } from "store"

const currentUser = useSelector((state: RootState) => state.currentUser)

// With proposal
import { useSelector } from "store"

const currentUser = useSelector(state => state.currentUser)
```

You can currently achieve this with something like:

```typescript
import { useSelector as useSelectorBase } from "react-redux"

export const useSelector = <TSelected>(
  selector: (state: RootState) => TSelected,
  equalityFn?: (left: TSelected, right: TSelected) => boolean,
): TSelected => useSelectorOrig(selector, equalityFn)
```

which is a bit verbose because you need to repeat the `TSelected` template parameter. Would it be possible to have a more concise (and update-proof) way of typing your own `useSelector`?

renchap:
@MrWolfZ Yes I has something like this in mind. For me there are 2 goals:
- Have this documented somewhere, so people know they can use this pattern of custom hooks with pre-typed state (similar to `useDispatch`)
- Be less dependent of the types implementation, so if the typings change you dont need to update your custom hook definition

Your proposal seems to solve both items :)

MrWolfZ:
I've added the `TypedUseSelectorHook` interface to the type definitions. I've also added a paragraph to the `useSelector` comment that mentions this new interface. I am not sure if there is a better way of making people aware of its existence.

RIP21:
@markerikson hey :) 7.1 got released, having this merged and published would be awesome :P

markerikson:
@RIP21 	: we don't control the typings at all :)  that's up to the community.

RIP21:
@markerikson Aha :P Gotcha. Saw you on reviewers list, thought that you are related to release somehow :)

Reviewers pleeeeease, make this one happen :)

markerikson (approved):

PR #34917, opened 2019-04-22
[next] Custom request shape for the server

osdiab:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [According to this doc](https://github.com/zeit/next.js/tree/canary/packages/next#fetching-data-and-component-lifecycle), on the server the HTTP request object is exposed in `ctx`; with a custom Express server, it is idiomatic for middleware to modify the Request object, so it follows that this can have custom fields.
- **N/A** Increase the version number in the header if appropriate.
- **N/A** If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

Shoutout to @tinaroh for pointing out these types were missing.

### DIFF
--- a/types/next/app.d.ts
+++ b/types/next/app.d.ts
@@ -13,10 +13,13 @@ export type AppComponentContext = NextAppContext;
  *
  * @template Q Query object schema.
  */
-export interface NextAppContext<Q extends DefaultQuery = DefaultQuery> {
-    Component: NextComponentType<any, any, NextContext<Q>>;
+export interface NextAppContext<
+    Q extends DefaultQuery = DefaultQuery,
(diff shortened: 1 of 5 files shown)


osdiab:
An alternative approach would be to use the Express typings for this, which rely on global ambient declarations to replace the `Express.Request` object definition, but I think that's kinda ugly and also ties the typings to express more tightly.

resir014 (approved):
👍 I think it's good that we leave it like this. Next.js is agnostic when it comes to server framework choices so IMO tying the request types to Express might not be a good idea.

scottdj92 (approved):

PR #34935, opened 2019-04-23
[@types/promise-retry] Export a namespace

kraenhansen:
When exporting a function using `export = promiseRetry` the module should have a namespace of the same name.

See https://github.com/Microsoft/TypeScript/issues/5073

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/Microsoft/TypeScript/issues/5073

### DIFF
--- a/types/promise-retry/index.d.ts
+++ b/types/promise-retry/index.d.ts
@@ -24,4 +24,6 @@ type RetryableFn<ResolutionType> = ((retry: (error: any) => never, attempt: numb
  */
 declare function promiseRetry<ResolutionType>(retryableFn: RetryableFn<ResolutionType>, options?: OperationOptions): Promise<ResolutionType>;
 declare function promiseRetry<ResolutionType>(options: OperationOptions, retryableFn: RetryableFn<ResolutionType>): Promise<ResolutionType>;
+// Exporting a namespace too
+namespace promiseRetry {}
 export = promiseRetry;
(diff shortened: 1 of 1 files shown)


shirakaba (approved):
Looks good. Thanks for providing the TypeScript issue reference.

types/promise-retry/index.d.ts
```diff
@@ -24,4 +24,6 @@ type RetryableFn<ResolutionType> = ((retry: (error: any) => never, attempt: numb
  */
 declare function promiseRetry<ResolutionType>(retryableFn: RetryableFn<ResolutionType>, options?: OperationOptions): Promise<ResolutionType>;
 declare function promiseRetry<ResolutionType>(options: OperationOptions, retryableFn: RetryableFn<ResolutionType>): Promise<ResolutionType>;
+// Exporting a namespace too
+declare namespace promiseRetry {}
```
PranavSenthilnathan: Could you provide the use case for this (playground link or, preferably, tests)? Consumers of this library should use the `require` syntax for importing or have the `esModuleInterop` flag + `import x from 'promise-retry'` syntax. This namespace change shouldn't be needed if the module is consumed the right way.

PranavSenthilnathan (changes requested):

types/promise-retry/index.d.ts
kraenhansen: I believe there are legitimate reasons that consumers of this package don't want to import using commonjs `require` nor enable the global `esModuleInterop` configuration option, just to consume this package. For one thing, this package claims to support TypeScript 2.3, the `esModuleInterop` was first [introduced with TypeScript 2.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-7.html) and the `esModuleInterop` flag is not enabled by default.

PR #34941, opened 2019-04-23
Update arangodb@3.5

pluma:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/arangodb/arangodb/blob/3.5/CHANGELOG

-----

`npm test` seems to fail because of xmlbuilder (unrelated).

### DIFF
--- a/types/arangodb/index.d.ts
+++ b/types/arangodb/index.d.ts
@@ -625,6 +625,7 @@ declare namespace ArangoDB {
             withData?: boolean
         ): CollectionChecksum;
         count(): number;
+        documentId(documentKey: string): string;
         drop(options?: { isSystem?: boolean }): void;
         figures(): CollectionFigures;
         load(): void;
--- a/types/arangodb/index.d.ts
+++ b/types/arangodb/index.d.ts
@@ -514,6 +514,7 @@ declare namespace ArangoDB {
 
     interface Index<T extends object = any> {
         id: string;
+        name: string;
         type: IndexType;
         fields: Array<keyof T | string>;
         sparse: boolean;
@@ -656,6 +657,7 @@ declare namespace ArangoDB {
         ): Array<Document<T>>;
         exists(name: string): boolean;
         firstExample(example: Partial<Document<T>>): Document<T> | null;
+        getResponsibleShard(document: DocumentLike): string;
         insert(data: DocumentData<T>, options?: InsertOptions): InsertResult<T>;
         insert(
             array: ReadonlyArray<DocumentData<T>>,
--- a/types/arangodb/index.d.ts
+++ b/types/arangodb/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for non-npm package ArangoDB 3.4
+// Type definitions for non-npm package ArangoDB 3.5
 // Project: https://github.com/arangodb/arangodb
 // Definitions by: Alan Plum <https://github.com/pluma>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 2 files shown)


types/arangodb/index.d.ts
```diff
@@ -951,8 +952,9 @@ declare namespace ArangoDB {
 
         // AQL
         _createStatement(query: Query | string): Statement;
+        _query(query: Query, options?: QueryOptions): Cursor;
         _query(
-            query: Query | string,
+            query: string,
```
pluma: Note: the old typings were actually incorrect: `db._query` wouldn't take any extra arguments if the first argument is a `Query` object. ArangoDB 3.5 adds support for passing a `Query` object with `QueryOptions`.

types/arangodb/index.d.ts
```diff
@@ -1163,6 +1165,10 @@ declare namespace Foxx {
     interface Request {
         arangoUser: string | null;
         arangoVersion: number;
+        auth: null | {
+            bearer?: string;
+            basic?: { username?: string; password?: string };
+        };
```
pluma: These are strictly mutually exclusive but `{bearer?: string} | {basic?: ...}` has unintuitive DX. I tried using `{bearer?: string, basic: never} | {bearer: never, basic?: ...}` but that didn't seem to guarantee exclusivity either.

PranavSenthilnathan: A union of two object types represents a type with all the properties of the first object type or all the properties of the second object type. However, this means that an object with all the properties of both objects satisfies this also.

In the definition with `never`, the error you're seeing is due to making the `never` typed property required instead of optional. If you change it a little, then it will work: [playground link](https://www.typescriptlang.org/play/index.html#src=type%20t%20%3D%20null%20%7C%20%7B%20bearer%3A%20string%2C%20basic%3F%3A%20never%20%7D%20%7C%20%7B%20bearer%3F%3A%20never%2C%20basic%3A%20%7B%20username%3F%3A%20string%2C%20password%3F%3A%20string%20%7D%20%7D%3B%0D%0A%0D%0Alet%20a%3A%20t%20%3D%20null%3B%20%2F%2F%20OK%0D%0Alet%20b%3A%20t%20%3D%20%7B%20bearer%3A%20%22b%22%20%7D%3B%20%2F%2FOK%0D%0Alet%20c%3A%20t%20%3D%20%7B%20basic%3A%20%7B%7D%20%7D%3B%20%2F%2F%20OK%0D%0Alet%20d%3A%20t%20%3D%20%7B%20bearer%3A%20%22d%22%2C%20basic%3A%20%7B%7D%20%7D%3B%20%2F%2F%20ERROR).

pluma: @PranavSenthilnathan This is a return value though. The logic I was trying to express is "if bearer is present, basic is undefined; if basic is present, bearer is undefined" and "if basic.username is undefined, so is basic.password".

Typing this as `{ bearer: string } | { basic: { username?: string, password?: string } }` means I can't even check for existence of either property, which is awkward for users.

Typing this with `never` means the following still works:

```js
if (auth.bearer) {
  doSomething(auth.basic); // this property shouldn't be valid here
}
```

So it's effectively indistinguishable from the much simpler typing I've ended up with.

PranavSenthilnathan: If `doSomething` is typed to not accept `undefined` then you would get an error instead. I see what you're saying though - even though the property will be strongly typed as `unknown`, the property access will not be an error as it would be for accessing nonexistent properties on normal objects

PR #34942, opened 2019-04-23
[@ember/array] Fix incorrect typing of Enumerable.sortBy

nickschot:
`sortBy` allows you to pass multiple properties to sort on. This is documented here: https://api.emberjs.com/ember/3.9/classes/EmberArray/methods/sortBy?anchor=sortBy


nickschot:
@mike-north

mike-north (changes requested):
Thanks @nickshot!

This seems to align with Ember's public API. We just need some tests for this change, and then should be good to go!

PR #34946, opened 2019-04-23
Updated react-native-dotenv

hmajid2301:
Updated react-native-dotenv tried to use syntax like this #34600. However the library throws an
error https://github.com/zetachang/react-native-dotenv/issues/5 as discussed in this issue.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/zetachang/react-native-dotenv/issues/5


types/react-native-dotenv/index.d.ts
```diff
@@ -4,9 +4,5 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
-interface Env {
-    [name: string]: string;
-}
-
-declare const env: Env;
-export = env;
+declare const _a: any;
```
PranavSenthilnathan: I'm not quite sure what this change is intending to solve. Could you add a test case to give an example of the problem and how this fixes it?

PranavSenthilnathan (changes requested):

types/react-native-dotenv/index.d.ts
hmajid2301: In  #34600, it was recommended to import the library `import env from "react-native-dotenv"`.
However if you try this it throw an error, as per this issue  zetachang/react-native-dotenv#5. 
So thats what I'm trying to resolve.

hmajid2301:
Changes made.

hmajid2301:
@PranavSenthilnathan not sure if you saw my comments.

In #34600, it was recommended to import the library import env from "react-native-dotenv".
However if you try this it throw an error, as per this issue zetachang/react-native-dotenv#5.
So thats what I'm trying to resolve.

PranavSenthilnathan:
I think there are two parts to what you want to solve.

1) Allowing arbitrary property access for a module. This isn't possible: https://stackoverflow.com/a/31729755

2) Disallow importing of the entire module (to force users to extract the properties in the import statement). Again, this isn't possible since the import syntax always allows importing the whole module.

I'm not sure how this plugin works internally but as I understand it does some rewriting of the import statement itself, so that might be the mechanism it uses to give an error when people import the wrong things. But from the original JavaScript/TypeScript language syntax, this isn't something that's supported.

@sandersn

hmajid2301:
To be honest I just wanted to hide the compiler warning by having some definitions for the library.

chrisdrackett:
@hmajid2301 have you gotten around the "@types/react-native-dotenv"' has no exported member" issues somehow? are you just adding ignores?

hmajid2301:
When I was using this package I was using Expo, since then I've ejected and I am using `react-native-config`.

PR #34948, opened 2019-04-23
[react-map-gl] Type additions and Edits.

pcwa-ahendricks:
Added string union type for width & height.Added string union type for width & height. Added MapLoadEvent interface for use with onLoad function.

### DIFF
--- a/types/react-map-gl/index.d.ts
+++ b/types/react-map-gl/index.d.ts
@@ -41,7 +41,7 @@ export interface MapboxProps extends Partial<ViewState> {
     mapOptions?: object;
     mapStyle?: string | object;
     visible?: boolean;
-    onLoad?: () => void;
+    onLoad?: (event: MapLoadEvent) => void;
     onError?: (e: MapError) => void;
     reuseMap?: boolean;
     width: number | string;
@@ -97,8 +97,8 @@ export interface MapControllerOptions {
 }
 
 export interface ViewportProps {
-    width: number;
-    height: number;
+    width: number | string;
+    height: number | string;
     latitude: number;
     longitude: number;
     zoom: number;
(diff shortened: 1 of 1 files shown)


pcwa-ahendricks (commented):
Reviewed.

types/react-map-gl/index.d.ts
```diff
@@ -97,8 +97,8 @@ export interface MapControllerOptions {
 }
 
 export interface ViewportProps {
-    width: number;
-    height: number;
+    width: number | string;
+    height: number | string;
```
fnberta: I don't think this is correct, these should be numbers. Take a look at https://github.com/uber/react-map-gl/blob/master/src/utils/map-state.js

fnberta (changes requested):
Thanks for this contribution. The `onLoad` change looks good but I think the union of `number | string` is not correct. Happy to be convinced otherwise if I'm wrong.

types/react-map-gl/index.d.ts
wheredoesyourmindgo: Thanks for the review @fnberta 

See [This example](https://github.com/uber/react-map-gl/blob/master/examples/exhibit-browserify/app.js) or (these prop-types](https://github.com/uber/react-map-gl/blob/f2bc61b6184d99c3ffa7267373b16e380dd600ba/src/components/static-map.js#L56). Using a string allows for setting width and height to '100vw' and '100vh' respectively.

fnberta: Yes, but these examples are covered by the `MapboxProps` definition which includes:
```
width: number | string;
height: number | string;
```
Or am I missing something?

pcwa-ahendricks: I see. Thanks for the clarification.

fnberta:
I would still like to get the `onLoad` change merged. If you agree, please re-open.

PR #34950, opened 2019-04-23
[react-bootstrap-typeahead] Fix HighlighterProps to reflect how they're used

nwalters512:
Please fill in this template.

## Explanation

In the source for the `Highlighter` component, there's an explicit check to see if the `search` prop is falsey. This contradicts the existing types, which requires that the `search` prop is provided. This would also produce type errors with [the given sample code](https://github.com/ericgio/react-bootstrap-typeahead/blob/master/docs/API.md#search-string-required):

```jsx
<Typeahead
  ...
  renderMenuItemChildren={(option, props, idx) => (
    <Highlighter search={props.text}>
      {option[props.labelKey]}
    </Highlighter>
  )}
/>
```

In this case, `props.text` is typed as `string | undefined`, which causes a type error. So an alternative fix would be to change the typings for `MenuProps.text`, but for me it was easier to verify that `Highligher` handles null/undefined well than it was to reason about when `MenuProps.text` might or might not be undefined.

### DIFF
--- a/types/react-bootstrap-typeahead/index.d.ts
+++ b/types/react-bootstrap-typeahead/index.d.ts
@@ -334,7 +334,7 @@ export class TypeaheadInputMulti<T extends TypeaheadModel> extends React.Compone
 --------------------------------------------------------------------------- */
 export interface HighligherProps {
     children: React.ReactNode;
-    search: string;
+    search?: string;
 }
 
 export class Highlighter extends React.PureComponent<HighligherProps> { }


nwalters512:
As a followup for the reviewers when they get here: `onKeyDown` is typed with `Event` - should that be changed to `KeyboardEvent`? Similarly for `onClick`, etc.

dalevfenton:
@nwalters512 I'm conflicted since the react-bootstrap-typeahead docs you linked to explicitly say that search is a required value, but the source does indeed check if it's falsey and falls back if it is so I'm inclined to accept this PR. You may want to put in a PR to fix the docs in that repository as well.

I would like to see test case as well added to the tests.

Do you know if the `prop` parameter passed to `renderMenuItemChildren` is correctly typed as `string|undefined` or if it's always a `string`? If that is typed incorrectly it would be nice to fix that for completeness as well.

If you want to tighten up the typings on the event handlers I think that would be a welcome change.

nwalters512:
Based on [`typeaheadContainer`](https://github.com/ericgio/react-bootstrap-typeahead/blob/308df6c7454e3fa460938c7993cb05921c5c0b33/src/containers/typeaheadContainer.js), it seems like `text` is always actually a string (and is only ever empty instead of `undefined`. I'm not sure if that prop is touched elsewhere though - I wonder if there was a good reason that the type is `string | undefined` now.

I'll work on event handler typings in another PR.

dalevfenton (approved):

PranavSenthilnathan:

PR #34953, opened 2019-04-24
Add data property to Insights

wrumsby:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/flightstats/node-insights/blob/master/index.js#L49

### DIFF
--- a/types/node-insights/index.d.ts
+++ b/types/node-insights/index.d.ts
@@ -22,6 +22,7 @@ declare class Insights {
     static collectorBaseURL: string;
     static queryBaseURL: string;
 
+    data: Array<Record<string, string | number | boolean>>;
     enabled: boolean;
     queryKey: string;
     insertKey: string;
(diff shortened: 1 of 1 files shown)


types/node-insights/index.d.ts
```diff
@@ -22,6 +22,7 @@ declare class Insights {
     static collectorBaseURL: string;
     static queryBaseURL: string;
 
+    data: Array<Record<string, any>>;
```
PranavSenthilnathan: I think the intention of library was to make this a 'private' property since it doesn't go through `Object.defineProperty` explicitly like other properties.

types/node-insights/index.d.ts
```diff
@@ -35,7 +36,7 @@ declare class Insights {
      * Add insights data to the queue.
      * It is sent when the queue reaches a max size or a period of time has elapsed
      */
-    add(data: object, eventType?: string): void;
+    add(data: Record<string, any>, eventType?: string): void;
```
PranavSenthilnathan: The docs on npmjs.com seem to indicate that data can in fact be an arbitrary object and isn't restricted to `Record<string, any>`.

PranavSenthilnathan (changes requested):

wrumsby:
@PranavSenthilnathan I agree with your comments here. Feels like this change is unnecessary, so I'll close this PR. Thanks for looking at it.

PR #34955, opened 2019-04-24
Update stream.d.ts to match spec

lifeiscontent:
fix declaration for Writable to match spec: https://nodejs.org/api/stream.html#stream_writable_end_chunk_encoding_callback

Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/node/node-tests.ts
+++ b/types/node/node-tests.ts
@@ -482,7 +482,12 @@ import Module = require("module");
     https.globalAgent.options.ca = [];
 
     {
-        function reqListener(req: http.IncomingMessage, res: http.ServerResponse): void {}
+        function reqListener(req: http.IncomingMessage, res: http.ServerResponse): void {
+            res.end();
+            res.end('foo');
+            res.end('foo', () => undefined);
+            res.end('foo', 'utf8', () => undefined);
+        }
 
         class MyIncomingMessage extends http.IncomingMessage {
             foo: number;
--- a/types/node/stream.d.ts
+++ b/types/node/stream.d.ts
@@ -121,7 +121,7 @@ declare module "stream" {
             write(chunk: any, cb?: (error: Error | null | undefined) => void): boolean;
             write(chunk: any, encoding: string, cb?: (error: Error | null | undefined) => void): boolean;
             setDefaultEncoding(encoding: string): this;
-            end(chunk: any): void;
+            end(chunk?: any): void;
             end(chunk: any, cb?: () => void): void;
             end(chunk: any, encoding: string, cb?: () => void): void;
             cork(): void;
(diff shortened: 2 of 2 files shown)


Flarna (changes requested):
Why this change? Which use case is now possible which was not before? Could you add a test for this case please?

lifeiscontent:
@Flarna the declaration was wrong as noted in the link I specified above.

when you call .end() you can either pass nothing, or a chunk which is written to the end of the stream

Flarna:
@lifeiscontent It's also allowed to pass only a callback.

lifeiscontent:
@Flarna so does that mean that this signature supports all 3 args to to be the first?

```
writable.end([chunk][, encoding][, callback])
```

I just want to make sure I am understanding this correctly, once I understand, I'll update the definition to match the spec.

Flarna:
@lifeiscontent `encoding` makes only sense if `chunk` is present but it's optional.
So the following variants are allowed to my understanding (and already supported by current definition):
```
end()
end(cb)
end(chunk)
end(chunk, encoding)
end(chunk, cb)
end(chunk, encoding, cb)
```

lifeiscontent:
@Flarna so how would I type this?

```
  res.write = (...args) => {
    if (typeof args[0] === 'string') {
      resBody += args[0];
    }
    return oldWrite.apply(res, args);
  };
```

because currently this is whats causing the error for me

Flarna:
@lifeiscontent Don't know if I understand your problem but `res.write = (...args: any[]) => {` silences typescript.
The append to `resBody` is protected by a typeguard so it's fine.
The type of `res.write()` should not change because of the assignment of a function accepting more inputs so users of `res.write()` should be fine.
As you don't tinker with `args[]` in your function `oldWrite.apply` should be also ok even types are lost here so no help of type system. Don't know if this can be improved.

lifeiscontent:
@Flarna I’m building a testing utility which intercepts write and end for res so I can check the body content in tests

Flarna:
@lifeiscontent But what is the actual problem? Are you unsure which argument you should check or how to write this down in typescript?

j-oliveras:
Reading node [source code](https://github.com/nodejs/node/blob/8dae89b396df64e6a9e44cb94efe27809a5a3d89/lib/_http_outgoing.js#L671-L679) the current types are valid as @Flarna said. Encoding are only used if chunk is defined.

lifeiscontent:
@Flarna how to write it in typescript, the oldWrite.apply thinks its first argument should be a function

Flarna:
@lifeiscontent What's wrong with`res.write = (...args: any[]) => {`?

PR #34970, opened 2019-04-24
[node-gravatar] add string enums and profile_url()

AprilArcus:
Please fill in this template.

> I am documenting the following additional features:
> - `size` should be of type `number`. Interpolating a non-integer value here will cause gravatar's API to silently replace the `size` parameter with a default value.
> - various options are redefined from strings to unions of string literals
> - the `format` option is only meaningful on the `profile_url()` function. This function is added to the definition and the type of its options bag is updated accordingly.
> no version number changes
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`

### DIFF
--- a/types/gravatar/index.d.ts
+++ b/types/gravatar/index.d.ts
@@ -1,23 +1,35 @@
 // Type definitions for gravatar v1.8.0
 // Project: https://github.com/emerleite/node-gravatar
 // Definitions by: Denis Sokolov <https://github.com/denis-sokolov>
+//                 April Arcus <https://github.com/aprilarcus>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-declare namespace GravatarModule {
-  type Options = {
-    d?: string
-    default?: string
-    f?: string
-    forcedefault?: string
-    format?: string
-    protocol?: string
-    r?: string
-    rating?: string
-    s?: string
-    size?: string
+declare module 'gravatar' {
+
+  type URL = string
+
+  type Default = URL | '404' | 'mp' | 'identicon' | 'monsterid' | 'wavatar' | 'retro' | 'robohash' | 'blank'
+
+  type Rating = 'g' | 'pg' | 'r' | 'x'
+
+  interface Options {
+    size?: number,
+    s?: number,
+    default?: Default,
+    d?: Default,
+    rating?: Rating,
+    r?: Rating,
+    forcedefault?: 'y',
+    f?: 'y',
+    protocol?: boolean | 'http' | 'https',
+    cdn?: string
   }
 
-  function url(email: string, options?: Options, forceProtocol?: boolean): string;
-}
+  interface ProfileOptions extends Options {
+    format?: 'json' | 'xml' | 'php' | 'vcf' | 'qr'
+  }
 
-export = GravatarModule;
+  export function url (email: string, options?: Options, https?: boolean): string
+  export function profile_url (email: string, options?: ProfileOptions, https?: boolean): string
+
+}
--- a/types/gravatar/index.d.ts
+++ b/types/gravatar/index.d.ts
@@ -12,9 +12,11 @@ declare module 'gravatar' {
 
   type Rating = 'g' | 'pg' | 'r' | 'x'
 
+  type Size = string | number
+
   interface Options {
-    size?: number,
-    s?: number,
+    size?: Size,
+    s?: Size,
     default?: Default,
     d?: Default,
     rating?: Rating,
--- a/types/gravatar/gravatar-tests.ts
+++ b/types/gravatar/gravatar-tests.ts
@@ -3,12 +3,12 @@
 import gravatar = require('gravatar');
 
 gravatar.url("email@example.com");
-gravatar.url("email@example.com", { s: 200, r: "pg", d: "404" });
-gravatar.url("email@example.com", { size: 200, r: "pg", d: "404" });
-gravatar.url("email@example.com", { s: 200 });
+gravatar.url("email@example.com", { s: "200", r: "pg", d: "404" });
+gravatar.url("email@example.com", { size: "200", r: "pg", d: "404" });
+gravatar.url("email@example.com", { s: "200" });
 gravatar.url("email@example.com", { default: "404" });
-gravatar.url("email@example.com", { s: 200, rating: "pg", d: "404" }, true);
-gravatar.url("email@example.com", { s: 200, r: "pg", default: "404" }, false);
+gravatar.url("email@example.com", { s: "200", rating: "pg", d: "404" }, true);
+gravatar.url("email@example.com", { s: "200", r: "pg", default: "404" }, false);
 gravatar.url("email@example.com", { d: "404" }, false);
 gravatar.url("email@example.com", { forcedefault: "y" }, false);
 gravatar.url("email@example.com", { f: "y" });
--- a/types/gravatar/gravatar-tests.ts
+++ b/types/gravatar/gravatar-tests.ts
@@ -3,14 +3,14 @@
 import gravatar = require('gravatar');
 
 gravatar.url("email@example.com");
-gravatar.url("email@example.com", { s: "200", r: "pg", d: "404" });
-gravatar.url("email@example.com", { size: "200", r: "pg", d: "404" });
-gravatar.url("email@example.com", { s: "200" });
+gravatar.url("email@example.com", { s: 200, r: "pg", d: "404" });
+gravatar.url("email@example.com", { size: 200, r: "pg", d: "404" });
+gravatar.url("email@example.com", { s: 200 });
 gravatar.url("email@example.com", { default: "404" });
-gravatar.url("email@example.com", { s: "200", rating: "pg", d: "404" }, true);
-gravatar.url("email@example.com", { s: "200", r: "pg", default: "404" }, false);
+gravatar.url("email@example.com", { s: 200, rating: "pg", d: "404" }, true);
+gravatar.url("email@example.com", { s: 200, r: "pg", default: "404" }, false);
 gravatar.url("email@example.com", { d: "404" }, false);
 gravatar.url("email@example.com", { forcedefault: "y" }, false);
 gravatar.url("email@example.com", { f: "y" });
 gravatar.url("email@example.com", { protocol: "https" });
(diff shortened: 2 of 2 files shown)


types/gravatar/index.d.ts
```diff
@@ -1,23 +1,35 @@
...
+    r?: Rating,
+    forcedefault?: 'y',
+    f?: 'y',
+    protocol?: boolean | 'http' | 'https',
+    cdn?: string
   }
 
-  function url(email: string, options?: Options, forceProtocol?: boolean): string;
-}
+  interface ProfileOptions extends Options {
```
denis-sokolov: You don't need to change type to an interface to provide extend functionality. With type arithmetic everything can be expressed, too. I don't care much about type vs interface, but I got surprised as to why the type was switched to an interface, seemed unrelated to the PR.

```suggestion
  type ProfileOptions = Options & {
```

types/gravatar/index.d.ts
```diff
@@ -1,23 +1,35 @@
...
-    default?: string
-    f?: string
-    forcedefault?: string
-    format?: string
-    protocol?: string
-    r?: string
-    rating?: string
-    s?: string
-    size?: string
+declare module 'gravatar' {
```
denis-sokolov: Great modernization change!

types/gravatar/index.d.ts
```diff
@@ -1,23 +1,35 @@
...
+
+  type URL = string
+
+  type Default = URL | '404' | 'mp' | 'identicon' | 'monsterid' | 'wavatar' | 'retro' | 'robohash' | 'blank'
+
+  type Rating = 'g' | 'pg' | 'r' | 'x'
+
+  interface Options {
+    size?: number,
```
denis-sokolov: Of course the size is supposed to be a number in the end, but we're writing typings for the node-gravatar, and that library expects a string size. If we provide it a number, at best things will happen to work due to JavaScript type casting, and at worst we will violate some expectations by the library author. Perhaps they decide to call `.toUppercase`, breaking our code.

Screenshot from docs for node-gravatar:

image

denis-sokolov (changes requested):
Enums are great here, profile_url is great. I disagree with the change for `size` and, seeing how it's not the focus of the PR, perhaps it's not a big deal to take it out?

types/gravatar/index.d.ts
AprilArcus: Sure, we could do that, but I think `interface ... extends` reads more clearly, and the semantics are identical for practical purposes.

types/gravatar/index.d.ts
AprilArcus: The contract is a little bit safer if we expose a Numeric type, for the reasons I outlined above:

> size should be of type number. Interpolating a non-integer value here will cause gravatar's API to silently replace the size parameter with a default value.

There's no conceivable reason for the author to call ".toUppercase()" on a value that only has correct behavior when it happens to be an integer or a string-encoded integer. Exposing this as a `number` will catch some errors, such as naively passing a CSS length like `32px`.

types/gravatar/index.d.ts
denis-sokolov: Considering how they are identical except for a personal matter of preference, wouldn't it make sense to keep whatever convention is already used in the code? :-)

To be clear, I could not care less about syntax conventions in this file, so the debate is purely theoretical for me.

AprilArcus: If you think `interface ... extends` is aesthetically offensive, I'll change it, but I repeat that I find it more readable. Interfaces are good descriptors for option bags; intersection types are more powerful and in this case, unnecessarily so. Consider the [law of least power](https://en.wikipedia.org/wiki/Rule_of_least_power).

types/gravatar/index.d.ts
denis-sokolov: Apologies, I did not see your original description, the PRs in this repo have a lot of boilerplate in the description.

Of course I agree with you that this is how the library should behave.

We disagree in what the 3rd-party typings should do. From where I stand, we’re augmenting existing npm modules by providing a programmatic description of their documentation, no more. The module has a very clear established expectation for this parameter to be a string, and the current typings convey this expectation to the TypeScript compiler.

Your proposed change would be amazing as a pull request to node-gravatar module, but not here.

AprilArcus: How about `string | number`?

types/gravatar/index.d.ts
denis-sokolov: I don’t think the `interface` keyword is aesthetically offensive. I do believe that type arithmetic reads more clearly.

But as I’ve said before, I don’t care if you rewrite the entire file with an ASCII graphic of a dinosaur. :-)

types/gravatar/index.d.ts
denis-sokolov: This is still not what the node-gravatar library claims to accept in their documentation. Either my value to maintain types consistent with documentation is valid, or it isn’t. ¯\_(ツ)_/¯

AprilArcus: I admit that the developer's own docs use string values here, so `number` is not correct.

I believe that `number | string` is the way to go here. It is bad ergonomics to require a developer to string-coerce a value that only has meaning when it's an integer.

denis-sokolov: I agree with you on the ergonomics 100%. And the fact that this should be the number in the first place. I support you 100% to go to node-gravatar, make a PR and make that library better.

How is it more beneficial to make this change only for TypeScript types, bringing inconsistency between the types and the implementation?

AprilArcus: Exactly what change would one even make to `node-gravatar` to make it explicitly support a numeric type here? Nothing implies that it can't be; the value gets sent to `querystring.stringify()` which gracefully handles numeric inputs and returns a string.

AprilArcus: In fact, passing a numeric value produces a working result, despite the fact that typescript currently improperly flags as a type violation with the existing definitions.

denis-sokolov: That's where the difficulty lies, indeed. Based on the implementation alone, it's ambiguous what types the library supports. But even if it wasn't, the implementation might be buggy. The documentation stores author's intent (when the documentation exists, that is).

As is, the documentation mentions that the values expected are "query string options". Although not 100%  certain, we know that HTTP query string options can only be strings, so this is a reason to believe the author only expects strings.

image

This is later reaffirmed in the usage examples, not one of which includes size being a number.

image

So in my personal judgement, the node-gravatar library's intent is to facilitate constructing a string url from string parts. I may be wrong, but I don't want to rely on the automatic type conversion when I use TypeScript specifically to get away from it! :-)

With typings as is, we know that we are definitely using it as expected, at the cost of a bit more boilerplate of type conversions. 🤷‍♂️

AprilArcus: Consider this example, where I perform arithmetic on the `size` value and wish to pass it in to `gravatar.url()` as-is.

```typescript
const devicePixelRatios = [1, 2, 3]

const Gravatar: React.FunctionComponent<
  { email: string, size: number | null } & JSX.IntrinsicElements['img']
> = React.memo(
  ({ email, size, ...props }) => {
    const src = gravatar.url(email, { protocol: 'https', default: '404', size })
    const srcSet = devicePixelRatios.map(
      (ratio) =>
        `${gravatar.url(email, { protocol: 'https', default: '404', size: size * ratio })} ${ratio}x`
    ).join(', ')
    return <img {...props} src={src} srcSet={srcSet} />
  }
)
```

what benefit, in type safety or developer ergonomics or strict adherence to the existing implementation would be obtained from forcing me to pass this value as `size: (size * ratio) as string`?

denis-sokolov: A measly one, but the `as` keyword in TypeScript is an indicator of a potential bug, so in a pull request review I would notice it immediately as a danger and ask whether we are sure there is no run-time bug here and whether we could encode this assurance programatically.

In fact, we can, and I suggest you use not the `as` keyword, but convert types manually:

```ts
size: String(size * ratio)
```

Now the strict typechecker is happy, but, more importantly, the run-time conversion is guaranteed to be working. Now we don't need to rely on the implementation details of node-gravatar, but, instead, clearly perform the conversion on our terms.

AprilArcus: And how might `querystring.stringify()` be re-implemented such that it would be clear to you that it intends to take numbers as well as strings? You've suggested I contribute that work elsewhere, but in fact the nature of JavaScript is such that it can't be done.

Calling `String(size * ratio)` here and pushing the type checking in to runtime is an actual waste of my customer's CPU time and battery power.

We don't live in a perfect world and I maintain that this is the correct tradeoff. In any event, you now have my work and my opinions; I leave the final disposition of the matter to you and your team.

AprilArcus: By the way, the `querystring.stringify()` library used by the author of `node-gravatar` does in fact handle numeric and boolean inputs explicitly: https://github.com/Gozala/querystring/blob/master/encode.js#L25-L36

AprilArcus: By the way, node's `querystring` library function does in fact explicitly handle a variety of types: https://nodejs.org/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options

denis-sokolov: It does not matter what the querystring library does, it's not a part of the public API of node-gravatar. The change I've suggested to make would be to the documentation of node-gravatar to mention the ability to pass a number as size.

To be clear, I am not part of the DefinitelyTyped team, I am a 3rd-party contributor. The bot on the project decides to assign the original author of some particular typings for review, hoping they’ll know more context, which I do in this case.

orta:
I'm not quite sure where we stand on this PR, is it ready to be merged?

denis-sokolov:
Back in April we had a disagreement with the author of the PR. I believe the pull request misrepresents the intent of the authors of node-gravatar in question and may lead to unexpected bugs if node-gravatar happens to release a version they consider to be backwards-compatible.

Nothing has changed since April, we still don’t have a consensus. To be clear, I am not a member of DefinitelyTyped, so you can override my objections, no problem!

peterblazejewicz:
fyi: missing exposed `profile_url` was added as part of #42843

AprilArcus:
Sounds good, closing this.

PR #34995, opened 2019-04-25
Cleanup 20190425

sandersn:
1. Update project urls
1. remove jquery-notifier: types for a package that was already types for a non-npm package.
2. react-test-renderer's test incorrectly assumed that <t> was an html tag. This is now an error. I updated the test to use `<div>`, since that's a tag I've heard of before.


Haroenv (approved):
Lgtm for React InstantSearch, thanks!

PR #35003, opened 2019-04-25
Add passport AuthenticationError class declaration

AlbertoFdzM:
Please fill in this template.

Select one of these and delete the others

  - https://github.com/jaredhanson/passport/blob/master/lib/errors/authenticationerror.js
  - https://github.com/passport/www.passportjs.org/pull/51/files

### DIFF
--- a/types/passport/lib/errors/authenticationerror.d.ts
+++ b/types/passport/lib/errors/authenticationerror.d.ts
@@ -1,0 +1,5 @@
+export default class AuthenticationError extends Error {
+    name: 'AuthenticationError';
+    message: string;
+    status: number;
+}
--- a/types/passport/index.d.ts
+++ b/types/passport/index.d.ts
@@ -6,6 +6,7 @@
 //                 Tomek Łaziuk <https://github.com/tlaziuk>
 //                 Daniel Perez Alvarez <https://github.com/danielpa9708>
 //                 Kevin Stiehl <https://github.com/kstiehl>
+//                 Alberto Fernandez <https://github.com/albertofdzm>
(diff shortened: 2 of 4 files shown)


types/passport/lib/errors/authenticationerror.d.ts
```diff
@@ -0,0 +1,5 @@
+export default class AuthenticationError extends Error {
```
weswigham: This is the wrong kind of export for this [source.](https://github.com/jaredhanson/passport/blob/master/lib/errors/authenticationerror.js#L20)
It should be
```ts
class AuthenticationError extends Error {
    name: 'AuthenticationError';
    message: string;
    status: number;
}
export = AuthenticationError;
```
instead.

weswigham (changes requested):

types/passport/lib/errors/authenticationerror.d.ts
AlbertoFdzM: > This is the wrong kind of export for this [source.](https://github.com/jaredhanson/passport/blob/master/lib/errors/authenticationerror.js#L20)
> It should be
> 
> ```ts
> class AuthenticationError extends Error {
>     name: 'AuthenticationError';
>     message: string;
>     status: number;
> }
> export = AuthenticationError;
> ```
> 
> instead.

Tried it and the linter give me this:
```
ERROR: 2:38  expect  TypeScript@next compile error:
This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.
```

weswigham: Yep, as the error suggests, just set `esModuleInterop` in the test `tsconfig`.

PR #35004, opened 2019-04-26
Types for the 'mixed.notType' property from the Yup library.

rwx222:
Hi, i'm implementing the types for the 'mixed.notType' property from the yup library, which is not typed right now.
The [Yup library code](https://github.com/jquense/yup/blob/a8935b789aed67a00867b536e6684ad9c493fc53/src/locale.js#L8) and [the issue](https://github.com/jquense/yup/issues/249) that is happening.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [yup library code](https://github.com/jquense/yup/blob/a8935b789aed67a00867b536e6684ad9c493fc53/src/locale.js#L8) and [the issue](https://github.com/jquense/yup/issues/249).

### DIFF
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -369,6 +369,7 @@ export class Ref {
 // tslint:disable-next-line:no-empty-interface
 export interface Lazy extends Schema<any> {}
 
+// tslint:disable-next-line:strict-export-declare-modifiers
 interface FormatErrorParams {
   path: string;
   type: string;
@@ -376,6 +377,7 @@ interface FormatErrorParams {
   originalValue?: any;
 }
 
+// tslint:disable-next-line:strict-export-declare-modifiers
 type LocaleValue =
     | string
     | ((params: FormatErrorParams) => string);
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -374,7 +374,6 @@ interface FormatErrorParams {
   type: string;
   value?: any;
   originalValue?: any;
-  label?: string;
 }
 
 type LocaleValue =
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -50,10 +50,6 @@ export type TestOptionsMessage =
     | string
     | ((params: object & Partial<TestMessageParams>) => string);
 
-export type NotTypeOptionsMessage =
-    | ((_ref: NotTypeRef) => string)
-    | string;
-
 export interface Schema<T> {
     clone(): this;
     label(label: string): this;
@@ -80,7 +76,6 @@ export interface Schema<T> {
     typeError(message?: TestOptionsMessage): this;
     oneOf(arrayOfValues: any[], message?: TestOptionsMessage): this;
     notOneOf(arrayOfValues: any[], message?: TestOptionsMessage): this;
-    notType(message: NotTypeOptionsMessage): this;
     when(keys: string | any[], builder: WhenOptions<this>): this;
     test(
         name: string,
@@ -237,13 +232,6 @@ export interface TestContext {
     createError: (params?: { path?: string; message?: string }) => ValidationError;
 }
 
-export interface NotTypeRef {
-    path: string;
-    type: string;
-    value: string | null;
-    originalValue: string | null;
-}
-
 export interface ValidateOptions {
     /**
      * Only validate the input, and skip and coercion or transformation. Default - false
@@ -381,8 +369,20 @@ export class Ref {
 // tslint:disable-next-line:no-empty-interface
 export interface Lazy extends Schema<any> {}
 
+interface FormatErrorParams {
+  path: string;
+  type: string;
+  value?: any;
+  originalValue?: any;
+  label?: string;
+}
+
+type LocaleValue =
+    | string
+    | ((params: FormatErrorParams) => string);
+
 export interface LocaleObject {
-    mixed?: { [key in keyof MixedSchema]?: string };
+    mixed?: { [key in keyof MixedSchema]?: string; } & { notType?: LocaleValue };
     string?: { [key in keyof StringSchema]?: string };
     number?: { [key in keyof NumberSchema]?: string };
     boolean?: { [key in keyof BooleanSchema]?: string };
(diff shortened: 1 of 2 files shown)


sseppola:
Hi, thanks for your contribution.

I've not been able to confirm this behavior with the docs/code, the [yup library code you refer to](https://github.com/jquense/yup/blob/a8935b789aed67a00867b536e6684ad9c493fc53/src/locale.js#L8) only shows a function on the locale of mixed, however it seems to me like you add it to the mixed schema. I cannot find `notType` in mixed schema https://github.com/jquense/yup#mixed

I'm not saying it's wrong, but I'm not familiar with this part of yup. Can you help me understand `Mixed.notType`? Alternatively one of the other maintainers here may be able confirm this behavior.

Regarding the code, is there a reason you didn't you co-locate `NotTypeRef` with `NotTypeOptionsMessage`?

YashdalfTheGray (commented):
> Regarding the code, is there a reason you didn't you co-locate `NotTypeRef` with `NotTypeOptionsMessage`?

I'm also wondering about this. 

I did some digging and I only found one usage of `locale.notType`. It's in the `SchemaType` definition but I'm not sure how that data is flowing through yup (need to do some homework). I do think that we shouldn't put `notType` on the Schema declaration itself, we should put it on the `LocaleObject` instead.

drebits:
Hi @sseppola @YashdalfTheGray.

I did some code on [codesandbox](https://codesandbox.io/s/xr06nz0roo) so you can see the mixed.notType working.

I know this mixed.notType is not documented yet in the library, and I don't know why, but is working as you can see.

And look at this code:
code

Here, I'm translating every error from the yup library using setLocale. And the type error is handled by mixed.notType function. And this mixed.notType can be a string, a function which returns a string, or a function receiving the _ref parameter (with its properties path, type, value, originalValue) that returns a string.

Regarding this comment:
```
Regarding the code, is there a reason you didn't you co-locate NotTypeRef with NotTypeOptionsMessage?
```
I only guided myself in the code I saw (sorry I'm not an english native speaker).
Of course you can change the code in a better way, because I'm not too good at typescript.

YashdalfTheGray:
Interesting, I do see that it's implemented this way in `locale.js` in the yup codebase and I ran through your code sandbox. It looks right, I don't know what will happen if we add it to the mixed schema rather than the locale schema though. 

What I'm trying to say is, `notType` exists on mixed locale and not mixed schema so should we add the types to mixed locale rather than mixed schema? Or do you have a usecase for adding it to mixed schema?

sseppola:
After playing with the codesandbox I don't think the `notType` function should be on the mixed schema.

As [all types inherit from the mixed type](https://github.com/jquense/yup#mixed), we'd expect to find `notType` available on `Yup.number()`, however this is not the case, just try the following:
```js
import * as Yup from 'yup';
console.log(typeof Yup.number().notType) // => undefined
```

Instead `notType` is part of the setLocale options, which is demonstrated [in this codesandbox](https://codesandbox.io/s/132825vr5q).
Screen Shot 2019-04-28 at 20 05 03


`Yup.setLocale` accepts the interface `LocaleObject` defined on this line:
 https://github.com/DefinitelyTyped/DefinitelyTyped/blob/afbdb8f705c39286b9c0e040dcaf824c5b31b61e/types/yup/index.d.ts#L384

So that means we need to update the following type:
```ts
export interface LocaleObject {
  mixed?: { [key in keyof MixedSchema]?: string };
  ...
}
```
where `notType` is not one of `keys in keyof MixedSchema`.

Using the [codesandbox](https://codesandbox.io/s/132825vr5q) I was able to confirm that the values can be either a string or a function, and traced it to the following lines in yup:
- [/src/util/createValidation.js#L52](https://github.com/jquense/yup/blob/a8935b789aed67a00867b536e6684ad9c493fc53/src/util/createValidation.js#L52)
- [/src/ValidationError.js#L44](https://github.com/jquense/yup/blob/a8935b789aed67a00867b536e6684ad9c493fc53/src/ValidationError.js#L44)

So I believe locales can be typed as follows:
```ts
interface FormatErrorParams {
  path: string;
  value: any;
  originalValue: any;
  label?: string;
}

type LocaleValue = string | (params: FormatErrorParams) => string;

export interface LocaleObject {
  mixed?: {
    [key in keyof MixedSchema]?: LocaleValue;
    notType?: LocaleValue;
  };
  ...etc
}
```

@YashdalfTheGray do you agree? and if so, @drebits are you up for making this change in your PR?

drebits:
Of course @sseppola, I'll be attentive.

YashdalfTheGray:
@sseppola yep, I agree with these changes! Love the deep dive!

drebits:
Hi guys @sseppola @YashdalfTheGray, I just made the suggested changes.

Just one thing, I don't think that **label** must be a property from **FormatErrorParams**, because of [this](https://codesandbox.io/s/7zxqpq0zv1) and [this](https://github.com/jquense/yup/blob/a8935b789aed67a00867b536e6684ad9c493fc53/src/locale.js#L8).  So I did not put it.

And also I had to put this line:

```
// tslint:disable-next-line:strict-export-declare-modifiers
```

Because due to this rule I have to export everything, so let me know what do you prefer, and I'll make the change if necessary.

YashdalfTheGray:
We probably should be exporting that interface anyway since it's being used in an exported function. I know it doesn't strictly matter but it's good to be consistent.

types/yup/index.d.ts
```diff
@@ -369,8 +369,21 @@ export class Ref {
 // tslint:disable-next-line:no-empty-interface
 export interface Lazy extends Schema<any> {}
 
+// tslint:disable-next-line:strict-export-declare-modifiers
+interface FormatErrorParams {
```
YashdalfTheGray: We should just export this too.

types/yup/index.d.ts
```diff
@@ -369,8 +369,21 @@ export class Ref {
...
+  path: string;
+  type: string;
+  value?: any;
+  originalValue?: any;
+}
+
+// tslint:disable-next-line:strict-export-declare-modifiers
+type LocaleValue =
```
YashdalfTheGray: And this.

types/yup/yup-tests.ts
```diff
@@ -452,6 +453,33 @@ const validateOptions: ValidateOptions = {
     }
 };
 
+const localeNotType1: LocaleObject = {
```
YashdalfTheGray: Nice tests! 💯

YashdalfTheGray (approved):
Looks good to me!

PR #35014, opened 2019-04-26
express-http-proxy: Add limit parameter for proxy options

erik-beus:
Please fill in this template.

- [ x ] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/villadora/express-http-proxy/blob/master/lib/requestOptions.js#L90

### DIFF
--- a/types/express-http-proxy/index.d.ts
+++ b/types/express-http-proxy/index.d.ts
@@ -16,7 +16,7 @@ interface ProxyOptions {
      * format supported by `bytes`, for example `1000`, `'500kb'` or `'3mb'`.
      * See https://github.com/stream-utils/raw-body/blob/master/index.d.ts
      */
-    limit?: number | string | null;
+    limit?: number | string;
     proxyReqPathResolver?: (req: Request) => string;
     proxyReqOptDecorator?: (proxyReqOpts: RequestOptions, srcReq: Request) => RequestOptions;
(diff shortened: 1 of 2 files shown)


types/express-http-proxy/index.d.ts
```diff
@@ -11,6 +11,12 @@ import { RequestOptions, IncomingHttpHeaders, OutgoingHttpHeaders } from "http";
 import { Response } from "express-serve-static-core";
 
 interface ProxyOptions {
+    /**
+     * The byte limit of the body. This is the number of bytes or any string
+     * format supported by `bytes`, for example `1000`, `'500kb'` or `'3mb'`.
+     * See https://github.com/stream-utils/raw-body/blob/master/index.d.ts
+     */
+    limit?: number | string | null;
```
ulrichb: As limit is optional already, why `| null`?

ulrichb (changes requested):

types/express-http-proxy/index.d.ts
ulrichb: If `null` is supported please add a test case, if not please  just remove.

erik-beus: @ulrichb I added it since it was part of the type definition from the `raw-body` package. But yes, since limit is optional it makes sense to leave it out for now.
PR has been updated.

ulrichb (approved):

ulrichb:
Thanks @erik-beus !

PR #35018, opened 2019-04-26
fix: restore null union type on promisify callbacks

jstewmon:
This fixes the promisify signatures that were broken in #34770


Flarna:
Is there a reason why v8 is not corrected with this PR?

jstewmon:
I just didn't notice that they were broken in the other PR b/c the diff didn't load by default. Fixing now...

Flarna (approved):

PR #35022, opened 2019-04-26
[nock] Add matchHeader to Interceptor.

mastermatt:
This method is available on the Scope or Interceptor level, however, the type was only included on Scope before.

  - [Method implementation](https://github.com/nock/nock/blob/24e5b47425b6d21c0491bcdbe3ea0114ca61a222/lib/interceptor.js#L437-L440)
  - [Usage of values](https://github.com/nock/nock/blob/24e5b47425b6d21c0491bcdbe3ea0114ca61a222/lib/interceptor.js#L230-L245), to verify param types.


types/nock/index.d.ts
```diff
@@ -86,14 +86,15 @@ declare namespace nock {
...
-        query(params: boolean | { (querObject: any): boolean; } | any): this;
+        query(params: boolean | { (queryObject: any): boolean; } | any): this;
 
         reply(responseCode: number, body?: string | any, headers?: HttpHeaders): Scope;
         reply(responseCode: number, callback: (uri: string, body: string, cb?: ReplyCallback) => ReplyCallbackResult, headers?: HttpHeaders): Scope;
         reply(callback: (uri: string, body: string, cb?: ReplyCallback) => ReplyCallbackResult, headers?: HttpHeaders): Scope;
         replyWithError(errorMessage: string | any): Scope;
         replyWithFile(responseCode: number, fileName: string, headers?: HttpHeaders): Scope;
 
+        matchHeader(name: string, value: string | RegExp | { (value: string): boolean; }): this;
```
PranavSenthilnathan: Any reason you're using a callable object type here instead of a function type?

mastermatt: I copied the line from the other instance above, but I'd be happen to update both cases if that's a prefered syntax.
```ts
matchHeader(name: string, value: string | RegExp | ((value: string) => boolean)): this;
```

PranavSenthilnathan: No, I'm for consistency here.

PR #35023, opened 2019-04-26
Slonik 16.16

mmkal:
@gajus @sebald some more new types (I'm adding these as I need them, can't guarantee I'll open a new PR here every time the core package gets a new feature)


types/slonik/slonik-tests.ts
```diff
@@ -24,6 +24,11 @@ import {
   TypeParserType,
   UniqueIntegrityConstraintViolationError
 } from 'slonik';
+import { ArrayTokenSymbol, TupleListTokenSymbol } from 'slonik/symbols';
+
+// make sure symbols are unique
+// $ExpectError
+const badSymbolAssignment: typeof ArrayTokenSymbol = TupleListTokenSymbol;
```
mmkal: previously all the symbols were assignable to each other - I updated them from `symbol` to `unique symbol` and this is the test to make sure that sticks

PR #35031, opened 2019-04-26
[react-transition-group] add JSDoc usage comments for individual props

ferdaber:
Please fill in this template.

No types were changed. I just added new JSDoc comments copied from the official docs site for individual props.

### DIFF
--- a/types/react-transition-group/Transition.d.ts
+++ b/types/react-transition-group/Transition.d.ts
@@ -19,10 +19,12 @@ export interface TransitionActions {
      * appear only adds an additional enter transition.
      */
     appear?: boolean;
+
     /**
      * Enable or disable enter transitions.
      */
     enter?: boolean;
+
     /**
      * Enable or disable exit transitions.
      */
@@ -41,6 +43,7 @@ export interface TransitionProps extends TransitionActions {
      * Show the component; triggers the enter or exit states
      */
     in?: boolean;
+
     /**
      * By default the child component is mounted immediately along with the
      * parent Transition component. If you want to "lazy mount" the component on
@@ -49,12 +52,14 @@ export interface TransitionProps extends TransitionActions {
      * also specify `unmountOnExit`.
      */
     mountOnEnter?: boolean;
+
     /**
      * By default the child component stays mounted after it reaches the
      * 'exited' state. Set `unmountOnExit` if you'd prefer to unmount the
      * component after it finishes exiting.
      */
     unmountOnExit?: boolean;
+
     /**
      * The duration of the transition, in milliseconds. Required unless addEndListener is provided.
      *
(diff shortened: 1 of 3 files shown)


types/react-transition-group/Transition.d.ts
```diff
@@ -24,18 +37,94 @@ export type TransitionStatus =
...
+     */
     onExited?: ExitHandler;
-    [prop: string]: any;
+    [ prop: string ]: any;
+    /**
+     * A function child can be used instead of a React element. This function is
+     * called with the current transition status ('entering', 'entered',
+     * 'exiting',  'exited', 'unmounted'), which can be used to apply context
+     * specific props to a component.
+     * ```js
```
ybiquitous: Typo?

```suggestion
     * ```jsx
```

types/react-transition-group/Transition.d.ts
```diff
@@ -50,7 +139,7 @@ export interface TransitionProps extends TransitionActions {
  * It's up to you to give meaning and effect to those states. For example we can
  * add styles to a component when it enters or exits:
  *
- * ```jsx
+ * ```js
```
ybiquitous: I think `jsx` is more appropriate than `js` here.

types/react-transition-group/TransitionGroup.d.ts
```diff
@@ -29,7 +29,7 @@ declare namespace TransitionGroup {
  * automatically by the `<TransitionGroup>`. You can use _any_ `<Transition>`
  * component in a `<TransitionGroup>`, not just css.
  *
- * ```jsx
+ * ```js
```
ybiquitous: Also here.

ybiquitous (commented):
Thanks for your contribution!

I think there are no big problems. I left comments about small issues, so please confirm them. 🙏

types/react-transition-group/Transition.d.ts
ferdaber: I found it strange when I was exploring how JSDoc is actually rendered in IDEs and in the case of VSCode it doesn't actually like JSX code fences (it doesn't recognize it). It will do syntax highlighting in the helper tooltip if I leave it as JS, however (it highlights JSX syntax correctly as well). I can turn it back though if we want to be super-general and not be VSCode specific; I didn't test it with other IDEs.

ybiquitous: I'm trying VSCode 1.33.1 on my MacBook, but it seems `jsx` renders syntax highlight correctly... 🤔 

Here is the screenshot:

image

ferdaber: Yep. They both highlight the same in the actual JSDoc markup, but the highlighting does not work for JSX in the actual tooltip:

The below is using `js`
image

The below is using `jsx`
image

ybiquitous: Hmm, it looks to me that this is a VSCode's bug... 🤔 
I think there are the following reasons that `jsx` is better than `js`:

- If this tooltip behavior is a VSCode's bug, it may be fixed in the future.
- It may not work well other IDEs or JSDoc (markdown) parser.
- Developers may be confusing because the original source code uses `jsx`. For example [here](https://github.com/reactjs/react-transition-group/blob/master/src/Transition.js#L36).

ybiquitous (approved):
LGTM! Thanks! 👍

PR #35033, opened 2019-04-26
Jest: Added Generic for "toMatchObject" function

favna:
Please fill in this template.

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: **None. It is a purely internal change not based on any changes to the lib.**

### DIFF
--- a/types/jest/index.d.ts
+++ b/types/jest/index.d.ts
@@ -18,6 +18,7 @@
 //                 Sebastian Sebald <https://github.com/sebald>
 //                 Andy <https://github.com/andys8>
 //                 Antoine Brault <https://github.com/antoinebrault>
+//                 Jeroen Claassens <https://github.com/favna>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
@@ -757,8 +758,25 @@ declare namespace jest {
         toMatch(expected: string | RegExp): R;
         /**
          * Used to check that a JavaScript object matches a subset of the properties of an object
+         * 
+         * Optionally, you can provide an object to use as Generic type for the expected value.
+         * This ensures that the matching object matches the structure of the provided object-like type.
+         * 
+         * @example
+         * 
+         * type House = {
+         *   bath: boolean;
+         *   bedrooms: number;
+         *   kitchen: {
+         *     amenities: string[];
+         *     area: number;
+         *     wallColor: string;
+         *   }
+         * };
+         * 
+         * expect(desiredHouse).toMatchObject<House>(...standardHouse, kitchen: {area: 20}) // wherein standardHouse is some base object of type House
          */
-        toMatchObject(expected: {} | any[]): R;
+        toMatchObject<O extends {}>(expected: O | any[]): R;
         /**
          * This ensures that a value matches the most recent snapshot with property matchers.
          * Check out [the Snapshot Testing guide](http://facebook.github.io/jest/docs/snapshot-testing.html) for more information.
--- a/types/jest/index.d.ts
+++ b/types/jest/index.d.ts
@@ -776,7 +776,7 @@ declare namespace jest {
          *
          * expect(desiredHouse).toMatchObject<House>(...standardHouse, kitchen: {area: 20}) // wherein standardHouse is some base object of type House
          */
-        toMatchObject<O extends {}>(expected: O | any[]): R;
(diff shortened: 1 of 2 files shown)


types/jest/index.d.ts
```diff
@@ -757,8 +758,25 @@ declare namespace jest {
...
+         *     amenities: string[];
+         *     area: number;
+         *     wallColor: string;
+         *   }
+         * };
+         *
+         * expect(desiredHouse).toMatchObject<House>(...standardHouse, kitchen: {area: 20}) // wherein standardHouse is some base object of type House
          */
-        toMatchObject(expected: {} | any[]): R;
+        toMatchObject<O extends {}>(expected: O | any[]): R;
```
wsmd: `toMatchObject` can also work with arrays:

> You can also pass an array of objects, in which case the method will return true only if each object in the received array matches (in the toMatchObject sense described above) the corresponding object in the expected array.

So we could probably support arrays in the generic as well. For example, currently the following will pass without any issues.

```ts
.toMatchObject<[{ foo: string }, { bar: number }]>([{ foo: true }, { bar: () => {} }]); // no errors!
```

What do you think about something like this? This should add type safety for tuples and array types via the generic if desired.

```suggestion
        toMatchObject<E extends {} | any[]>(expected: E): R;
```

types/jest/index.d.ts
```diff
@@ -757,8 +758,25 @@ declare namespace jest {
         toMatch(expected: string | RegExp): R;
         /**
          * Used to check that a JavaScript object matches a subset of the properties of an object
+         *
+         * Optionally, you can provide an object to use as Generic type for the expected value.
+         * This ensures that the matching object matches the structure of the provided object-like type.
```
wsmd: I don't see any comments similar to this in here.

I believe the documentation throughout the file is intended to document Jest's API as stated in its docs, not on how to use the types. The type information, such as generic arguments, can still be inferred from the editor sometimes:

Screen Shot 2019-04-27 at 11 28 34 AM

wsmd (changes requested):
Thanks for the PR @Favna! This will be a helpful addition.

I left a couple of notes. Let me know what you think!

types/jest/index.d.ts
Favna: I based the including of an example on `.toHaveNthReturnedWith()` typing just above `toMatchObject()`. As for inferring from IntelliSense, the examples are in fact also included in the IntelliSense and I personally thing it's nice to have some extra documentation, especially for TypeScript structures that are considered "advanced" (being that Generics are under "advanced types" on the docs) so it helps TS starters.

GIF recording of the IntelliSense I see, with the example:

19-57-46_27-04-2019_%pn


TL;DR: I think it doesn't give any negative effects and only (if minor), positive ones. I see no reason to remove it.

types/jest/index.d.ts
Favna: That's a very good suggestion. I've made your changes as well as added a corresponding test in dbfa03d852c22dbe723f559a8029ce389278570a

types/jest/index.d.ts
wsmd: The comment added stood out a little bit to me from the rest of the documentation comments since it isn't describing Jest's functionality, but given what you said above I don't think it hurts to keep it either. Thanks for the clarification! 😄

wsmd (approved):
LGTM! 👍

types/jest/index.d.ts
```diff
@@ -757,8 +758,25 @@ declare namespace jest {
...
+         *   bath: boolean;
+         *   bedrooms: number;
+         *   kitchen: {
+         *     amenities: string[];
+         *     area: number;
+         *     wallColor: string;
+         *   }
+         * };
+         *
+         * expect(desiredHouse).toMatchObject<House>(...standardHouse, kitchen: {area: 20}) // wherein standardHouse is some base object of type House
```
jwbay: Is the intent to spread into an object here?
```suggestion
         * expect(desiredHouse).toMatchObject<House>({ ...standardHouse, kitchen: {area: 20} }) // wherein standardHouse is some base object of type House
```

Favna: This was indeed meant to be an object spread. I indeed forgot a `{`. Looks like the PR was merged already though so this'll have to be covered in a next one. That said... It's not *really* important I think so I think I'll just wait for the next PR to come up and the bot will ping me now as well so I can request the change then from whoever is editing.

PR #35034, opened 2019-04-26
fundamental-react 0.5.1

Feinbube:
Please fill in this template.

If adding a new definition:


types/fundamental-react/lib/Tabs.d.ts
```diff
@@ -0,0 +1,36 @@
...
+    /* _INTERNAL USE ONLY._ */
+    selected?: boolean;
+    /* Additional props to be spread to the tab content's <div> element. */
+    tabContentProps?: { [x: string]: any };
+    /* Localized text to display on the tab. */
+    title?: string;
+    /* _INTERNAL USE ONLY._ */
+    onClick?: (
+        event: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
+        index?: number
```
PranavSenthilnathan: As discussed [here](http://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html), it's not recommended to have optional parameters in callbacks. The [underlying library](https://github.com/SAP/fundamental-react/blob/master/src/Tabs/Tab.js#L44) seems to always call this function with an `index` argument.

types/fundamental-react/lib/ActionBar.d.ts
```diff
@@ -0,0 +1,37 @@
...
+    /* Localized text for the description. */
+    description?: string;
+    /* Additional props to be spread to the description's `<p>` element. */
+    descriptionProps?: { [x: string]: any };
+    /* Heading level. `<h1>` is reserved for the page title. */
+    headingLevel?: 1 | 2 | 3 | 4 | 5 | 6;
+    titleProps?: { [x: string]: any };
+} & { [x: string]: any };
+
+export class ActionBar extends React.Component<ActionBarProps> {
```
PranavSenthilnathan: It looks like `fundamental-react` organizes its internal exports as `default` exports and then exports them in the index.js with different names. this code exports with the final name and then the index.d.ts just exports without renaming. There's no functional difference between the two, but if you wanted you could use default exports here like the original library.

PranavSenthilnathan (changes requested):

types/fundamental-react/lib/Tabs.d.ts
Feinbube: that's great feedback! thank you :)

types/fundamental-react/lib/ActionBar.d.ts
Feinbube: Thank you for your feedback. It turned out that it did make a difference in this project. I reworked the folder structure and covered all the other ways to import classes from fundamental-react.

Feinbube:
```
import { FormSet } from "fundamental-react";
import { FormItem } from "fundamental-react/lib";
import { FormLabel } from "fundamental-react/lib/Forms";
import FormInput from "fundamental-react/lib/Forms/FormInput";
```

PranavSenthilnathan (approved):

PranavSenthilnathan:

PR #35037, opened 2019-04-27
pouchdb-live-find - type declaration files

assemblethis:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/pouchdb-live-find/index.d.ts
+++ b/types/pouchdb-live-find/index.d.ts
@@ -1,0 +1,48 @@
+// Type definitions for pouchdb-live-find 0.4
+// Project: https://github.com/colinskow/pouchdb-live-find
+// Definitions by: assemblethis <https://github.com/assemblethis>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+/// <reference types="pouchdb-core" />
+declare namespace PouchDB {
+    namespace LiveFind {
+
+        interface RequestDef<Content extends {}> extends Pick<PouchDB.Find.FindRequest<Content>, Exclude<keyof PouchDB.Find.FindRequest<Content>, 'use_index'>>
+        {
+            aggregate?: boolean; // if true outputs an aggregate list on every update event
+        }
+
+        interface PaginateOptions<Content extends {}> extends Pick<PouchDB.Find.FindRequest<Content>, 'sort' | 'skip'| 'limit'> {}
+
+        interface UpdateEvent {
+            action: 'REMOVE' | 'ADD' | 'UPDATE'
+            id: any
+            rev: any
+            doc: any
+        }
+        interface LiveFeed<Content extends {} = {}> {
+            on(event: 'ready', listener: () => void): this;
+            on(event: 'error', listener: (err: any) => void): this;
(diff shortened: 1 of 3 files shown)


types/pouchdb-live-find/index.d.ts
```diff
@@ -0,0 +1,47 @@
...
+declare namespace PouchDB {
+    namespace LiveFind {
+        interface RequestDef<Content extends {}> extends Pick<Find.FindRequest<Content>, Exclude<keyof Find.FindRequest<Content>, 'use_index'>> {
+            aggregate?: boolean; // if true outputs an aggregate list on every update event
+        }
+
+        interface PaginateOptions<Content extends {}> extends Pick<Find.FindRequest<Content>, 'sort' | 'skip'| 'limit'> {}
+
+        interface UpdateEvent {
+            action: 'REMOVE' | 'ADD' | 'UPDATE';
```
rpgeeganage: I think that declare an enum for `'REMOVE' | 'ADD' | 'UPDATE'` will be helpful.

types/pouchdb-live-find/index.d.ts
```diff
@@ -0,0 +1,47 @@
...
+// TypeScript Version: 3.0
+
+/// <reference types="pouchdb-core" />
+declare namespace PouchDB {
+    namespace LiveFind {
+        interface RequestDef<Content extends {}> extends Pick<Find.FindRequest<Content>, Exclude<keyof Find.FindRequest<Content>, 'use_index'>> {
+            aggregate?: boolean; // if true outputs an aggregate list on every update event
+        }
+
+        interface PaginateOptions<Content extends {}> extends Pick<Find.FindRequest<Content>, 'sort' | 'skip'| 'limit'> {}
```
rpgeeganage: I think that declare an enum for 'sort' | 'skip'| 'limit' will be helpful.

types/pouchdb-live-find/index.d.ts
assemblethis: From what I've read string literal types are the way to go (https://medium.com/@katbusch/typescript-enums-explained-e5f9a101afc9): 

```
Recommendation
So of the many types of enums which do I recommend using?

Trick question: none.

I recommend string literal types. It’s more TypeScripty. It’s extremely readable. You can see it in the code as is. If you print the value of a variable you won’t get a mysterious 0 or 1; you’ll get the actual string itself every time.

```

types/pouchdb-live-find/index.d.ts
assemblethis: These are the keys of the interface that I'm picking. Can you give an example of how you would handle this with an enum?

rpgeeganage: What about something like https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5344bfc80508c53a23dae37b860fb0c905ff7b24/types/pinterest-sdk/index.d.ts#L13

types/pouchdb-live-find/index.d.ts
rpgeeganage: some enum like https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5344bfc80508c53a23dae37b860fb0c905ff7b24/types/pinterest-sdk/index.d.ts#L13

PR #35038, opened 2019-04-27
fabric: event mouse props

aleksanderd:
https://github.com/fabricjs/fabric.js/blob/ce23118378fb0ebefcd9d25f3402909329691787/src/mixins/canvas_events.mixin.js#L431

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@
 *.csproj
 *.map
 *.swp
+*.code-workspace
 .DS_Store
 
 _Resharper.DefinitelyTyped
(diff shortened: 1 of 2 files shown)


BradleyHill (approved):
Looks like in fabricjs:
```
          options = {
            e: e,
            target: target,
            subTargets: targets,
            button: button || LEFT_CLICK,
            isClick: isClick || false,
            pointer: this._pointer,
            absolutePointer: this._absolutePointer,
            transform: this._currentTransform
          };
```
Could probably include subTargets to close it out.

PR #35053, opened 2019-04-28
webpack: add child context properties

LukeSheard:
Adds missing properties to a "Compiler" instance as seen in following lines of code:

https://github.com/webpack/webpack/blob/master/lib/Compiler.js#L183 
https://github.com/webpack/webpack/blob/master/lib/Compiler.js#L565 
https://github.com/webpack/webpack/blob/master/lib/Compiler.js#L122


johnnyreilly:
Thanks @lukesheard!

PR #35056, opened 2019-04-28
webpack: fix for Compiler instance 

LukeSheard:
Fix for #35053 where types were added to wrong interface.


johnnyreilly:
Thanks!

PR #35083, opened 2019-04-29
 @types/passport-http: fix unused template param #1 

ianwremmel:
It looks like this was rigged up so the appropriate callback definition could be inferred, but one detail got missed.

@krizalys
@tlaziuk
@cjbarth
@jamesadarich

Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/passport-http/index.d.ts
+++ b/types/passport-http/index.d.ts
@@ -12,7 +12,7 @@ import express = require("express");
 
 export interface BasicStrategyOptions<req extends boolean = boolean> {
     realm?: string;
-    passReqToCallback?: boolean;
+    passReqToCallback?: req;
 }
 
 export interface DigestStrategyOptions {
--- a/types/passport-http/passport-http-tests.ts
+++ b/types/passport-http/passport-http-tests.ts
@@ -25,6 +25,9 @@ function validateNonce(nonce: string) {
 function validateParams(nonce: string, cnonce: string, nc: number, opaque: string) {
 }
 
+new http.BasicStrategy({passReqToCallback: false}, (username, password, done) => {});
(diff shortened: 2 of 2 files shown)


cjbarth (commented):
I'm not sure why this change is being requested. It seems from looking at passport/index.d.ts that this indeed should be a `boolean`.

ianwremmel:
It's still a boolean, but it's the template parameter `req extends boolean = boolean`. Without this change, the constructor overloads at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/234afdc55eee47734f0801e1532eea104318eeac/types/passport-http/index.d.ts#L58-L59 have no effect. `new BasicStrategy({passReqToCallback: true}, (req, username, password, done) =>...)` produces a type error because `{passReqToCallback: true}` satisfies both overloads. 

With the change, it can see that the `true` signals that the second parameter passed to `BasicStrategy()` is definitely `BasicVerifyFunctionWithRequest` and not `BasicVerifyFunction|BasicVerifyFunctionWithRequest`.

cjbarth:
I follow your logic there, though I've never leveraged that feature of TypeScript myself. Could you write a test that validates that kind of typing works with the constructor overloads you describe?

ianwremmel:
I think so. (I think) this my first DT PR. Do I just write some example code and make sure there are no type failures?

cjbarth:
Yep. There is a test file already in the repo. You can put the code there with all the typings that you want to work and it will be preserved for future modders :)

cjbarth:
I'll have a look at this later today or tomorrow.

cjbarth (approved):

PR #32990, opened 2019-02-12
[React] Allow to use ReactElement without Props

VincentLanglet:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Most of the time, we (I) use ReactElement<any>. Passing props to this type is not really useful, so I added a default value.

PR #35107, opened 2019-04-30
[react-motion] Fix missing generic-param error on children

lukasproust:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/chenglou/react-motion/blob/master/src/Types.js#L65

### DIFF
--- a/types/react-motion/index.d.ts
+++ b/types/react-motion/index.d.ts
@@ -65,7 +65,7 @@ interface MotionProps {
      * Callback with your interpolated styles. Must return one react element to render
      * @param interpolatedStyle
      */
-    children?: (interpolatedStyle: PlainStyle) => ReactElement;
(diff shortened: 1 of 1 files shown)


jessetrinity:
The need for this change is not immediately apparent to me. Is there a test for this?

MichaelBitard:
It was based on that: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34912

> Prior to #32990, the React.ReactElement type did not have an optional first type param. Rather than forcing @storybook/addon-storyshots users to upgrade their @types/react package (which may require them to also upgrade react itself), we can switch to use JSX.Element which is and has been a simple alias for React.ReactElement<any, any>.

That's the same issue for react-motion.

I'm not sure how I could add a test on this though

PR #35114, opened 2019-04-30
[ember] all computed property descriptors are also PropertyDecorators

dfreeman:
As of [Ember 3.10](https://github.com/emberjs/ember.js/blob/v3.10.0-beta.1/CHANGELOG.md#v3100-beta1-april-02-2019), computed property descriptors are stage-1 property decorators. The declarations for items in `@ember/*` packages have been updated individually to reflect that, but downstream functions that return a `ComputedProperty<T>` (like `ember-data`'s `attr()` or custom domain-specific computed macros in individual applications) haven't reflected this new reality.

This updates the definition of `ComputedProperty` and removes the one-off `& PropertyDecorator` annotations that are no longer necessary.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/emberjs/rfcs/blob/master/text/0408-decorators.md#detailed-design

### DIFF
--- a/types/ember__object/test/octane.ts
+++ b/types/ember__object/test/octane.ts
@@ -34,6 +34,12 @@ import {
     uniqBy
 } from "@ember/object/computed";
 
+function customMacro(message: string) {
+    return computed(() => {
+        return [message, message];
+    });
+}
+
 // Native class syntax
 class Foo extends EmberObject {
     firstName: string;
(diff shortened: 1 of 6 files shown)


chriskrycho:
Also noting for posterity that the official [ember-decorators-polyfill](https://github.com/pzuraq/ember-decorators-polyfill) makes this work all the way back to late 2.x, which is actually *older* than Ember's LTS support policy, so users can opt into this at any point in our own supported 3.x matrix.

chriskrycho (approved):

types/ember__object/test/octane.ts
```diff
@@ -47,6 +53,17 @@ class Foo extends EmberObject {
         return `${this.firstName} ${this.lastName}`;
     }
 
+    @computed("firstName", "lastName") // $ExpectError
```
chriskrycho: For my own knowledge: does this test work b/c we're returning a `MethodDecorator` here? I.e. `MethodDecorator`s work on getters but not on regular properties?

dfreeman: That's right — `MethodDecorator`s receive the descriptor as an argument, while `PropertyDecorator`s (at least as far as TS's implementation is concerned) only receive the target and the key.

mike-north (approved):

PR #35117, opened 2019-04-30
Add options for ensureDir and ensureDirSync

sangdth:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/jprichardson/node-fs-extra/blob/master/docs/ensureDir-sync.md>>
- [x] Did not make substantial changes.

---

I'm new to this, please take a look and let me know if there is anything need to be done.
Thanks in advance.

### DIFF
--- a/types/fs-extra/fs-extra-tests.ts
+++ b/types/fs-extra/fs-extra-tests.ts
@@ -15,6 +15,9 @@ const modeNum = 0;
 const modeStr = "";
 const object = {};
 const errorCallback = (err: Error) => { };
+const ensureOptions: fs.EnsureOptions = {
+   mode: 0o700
+};
 const readOptions: fs.ReadOptions = {
 	reviver: {}
 };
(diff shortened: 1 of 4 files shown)


types/fs-extra/fs-extra-tests.ts
```diff
@@ -142,11 +145,11 @@ fs.writeJSON(file, object, writeOptions).then(() => {
 fs.writeJsonSync(file, object, writeOptions);
 fs.writeJSONSync(file, object, writeOptions);
 
-fs.ensureDir(path).then(() => {
+fs.ensureDir(path, ensureOptions).then(() => {
```
shiftkey: These tests should be additive, as I believe the previous examples are still valid.

types/fs-extra/fs-extra-tests.ts
```diff
@@ -15,6 +15,9 @@ const modeNum = 0;
 const modeStr = "";
 const object = {};
 const errorCallback = (err: Error) => { };
+const ensureOptions: fs.EnsureOptions = {
```
shiftkey: What about tests for providing a `number` as the second parameter?

types/fs-extra/fs-extra-tests.ts
sangdth: I agree, let me change it

types/fs-extra/fs-extra-tests.ts
sangdth: I have added the test for number.

PR #35121, opened 2019-04-30
[aws-lambda] Keys should be optional inside CloudFrontHeaders

sgarner:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: _see below_
- ~Increase the version number in the header if appropriate.~
- ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.~

----

According to the AWS documentation, it is optional to specify a `key` for each header value, but the type declaration is currently requiring this to be specified unnecessarily.

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-event-structure.html#lambda-event-structure-response

> - `key` (optional) is the case-sensitive name of the header as it appears in an HTTP request; for example, `accept` or `host`.
> - If you do not include the header key portion of the key-value pair, Lambda@Edge will automatically insert a header key using the header name that you provide. Regardless of how you've formatted the header name, the header key that is inserted automatically will be formatted with initial capitalization for each part, separated by hyphens (-).
>
>   For example, you can add a header like the following, without a header key: `'content-type': [{ value: 'text/html;charset=UTF-8' }]`. In this example, Lambda@Edge creates the following header key: `Content-Type`.

By making this property optional, one is able to write:

```ts
import { CloudFrontHeaders } from 'aws-lambda';

const headers: CloudFrontHeaders = {
  'content-type': [{ value: 'text/plain '}],
};
```

... instead of having to repeat the header name like:

```ts
import { CloudFrontHeaders } from 'aws-lambda';

const headers: CloudFrontHeaders = {
  'content-type': [{ key: 'content-type', value: 'text/plain '}],
};
```

### DIFF
--- a/types/aws-lambda/aws-lambda-tests.ts
+++ b/types/aws-lambda/aws-lambda-tests.ts
@@ -1019,6 +1019,11 @@ const cloudFrontRequestHandler: AWSLambda.CloudFrontRequestHandler = (event: AWS
 
 const cloudFrontResponseHandler: AWSLambda.CloudFrontResponseHandler = (event: AWSLambda.CloudFrontResponseEvent, context: AWSLambda.Context, cb: AWSLambda.CloudFrontResponseCallback) => { };
 
+const cloudFrontHeaders: AWSLambda.CloudFrontHeaders = {
+    'content-type': [{ value: 'text/plain' }],
+    'x-foo-bar': [{ key: 'X-Foo-Bar', value: 'example' }]
+};
+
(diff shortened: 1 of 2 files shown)


ohookins:
That's a good find, and one that will make a bunch of code simpler!
Maybe instead of modifying the existing test, you can have one additional case with just the value, and another with also the key? We should leave the empty additional headers case.

sgarner:
> Maybe instead of modifying the existing test, you can have one additional case with just the value, and another with also the key? We should leave the empty additional headers case.

Sure, I've added a separate test for this now.

ohookins (approved):

PR #35125, opened 2019-04-30
[ldapjs] created initial stubs for Server, createServer, and related objects/functions

mscottnelson:
[ldapjs] created initial stubs for Server, createServer, and related objects/functions

Initial missing types associated with Server object for ldapjs, as documented here: http://ldapjs.org/server.html


pmoleri:
Hi @mscottnelson , it seems that in this PR you removed the version number from the first line and now the bot is publishing 0.0.x versions instead of 1.0.x version. I guess that's a mistake, if that's the case could you create a new PR to fix it?

PR #35140, opened 2019-05-01
[@types/pusher-js] 4.3.0 - introduce new forceTLS & deprecate encrypted

BrendonSled:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/pusher/pusher-js/blob/v4.3.0/src/core/options.ts

### DIFF
--- a/types/pusher-js/index.d.ts
+++ b/types/pusher-js/index.d.ts
@@ -35,6 +35,11 @@ declare namespace pusher {
         /**
          * Forces the connection to use encrypted transports.
          */
+        forceTLS?: boolean;
+        /**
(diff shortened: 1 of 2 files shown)


cainlevy:
There's a brief existing test for the deprecated prop. Would you add a copy for the new one?

cainlevy (changes requested):
There's a brief existing test for the deprecated prop. Would you add a copy for the new one?

BrendonSled:
@cainlevy I've updated the tests

cainlevy (approved):

Issue #35151, opened 2019-05-02
[@types/paper] divdeAt definition is completely wrong!


luz-alphacode:
The definition of **divdeAt** function of @types/paper is wrong.
```ts
        /**
         * Divides the path on the curve at the given location into two curves, by inserting a new segment at the given location.
         * @param location - the curve location at which to divide the existing curve by inserting a new segment
         * @returns the newly inserted segment if the location is valid, {code null} otherwise
         */
        divideAt(location: CurveLocation): Path;

        /**
         * Divides the path on the curve at the given offset into two curves, by inserting a new segment at the given location.
         * @param offset - the offset at which to divide the existing curve by inserting a new segment
         * @returns the newly inserted segment if the location is valid, {code null} otherwise
         */
        divideAt(offset: number): Path;
```

As you see, even in the comment, it says the divideAt function **return the newly insert segment**, but the return type is defined as **Path** instead, which is wrong!

PR #35153, opened 2019-05-02
fix path.divideAt return definition as @types/paper

glenzli:
The divdeAt funtion shoud return segment, not a Path, even the comment says the return is a Segment.

The problem is addressed here: [https://github.com/DefinitelyTyped/DefinitelyTyped/issues/35151](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/35151)

And the official document of **divideAt** function is here.
[http://paperjs.org/reference/path/#divideat-location](http://paperjs.org/reference/path/#divideat-location)

### DIFF
--- a/types/paper/index.d.ts
+++ b/types/paper/index.d.ts
@@ -3027,20 +3027,20 @@ declare module paper {
          * Clears the path’s handles by setting their coordinates to zero, turning the path into a polygon (or a polyline if it isn’t closed).
          */
         clearHandles(): void;
-
+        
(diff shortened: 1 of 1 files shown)


jessetrinity:
@clark-stevenson @Xakaloz @SebasWebDev @rfried

Any interest in this PR?

Thanks!

clark-stevenson (approved):
Looks good to me!  I am sorry for such a long delay!

PR #35159, opened 2019-05-02
add missing properties to @types/pdfMake

sbilbault:
- modification to TDocumentDefinitions based on https://pdfmake.github.io/docs/document-definition-object/images/  and https://pdfmake.github.io/docs/document-definition-object/page/
- modification to Table based on http://pdfmake.org/playground.html example on Tables
- modification to TDocumentHeaderFooterFunction based on https://pdfmake.github.io/docs/document-definition-object/headers-footers/
- modification to Content based on https://pdfmake.github.io/docs/document-definition-object/styling/

### DIFF
--- a/types/pdfmake/index.d.ts
+++ b/types/pdfmake/index.d.ts
@@ -155,7 +155,7 @@ declare module "pdfmake/build/pdfmake" {
 
     interface Table {
         body: Content[][] | TableCell[][];
-        dontBreakRows?: boolean;
+        // dontBreakRows?: boolean;
         headerRows?: number;
         heights?: Array<string | number> | TableRowFunction;
         layout?: string | TableLayoutFunctions;
@@ -163,7 +163,7 @@ declare module "pdfmake/build/pdfmake" {
     }
 
     interface Content {
-        style?: string | string[];
+        style?: "string"; //string | string[];
         margin?: Margins;
         text?: string | string[] | Content[];
         columns?: Content[];
@@ -181,15 +181,15 @@ declare module "pdfmake/build/pdfmake" {
     }
 
     interface TDocumentDefinitions {
-        background?: any;
+        // background?: any;
         compress?: boolean;
         content: string | Content;
         defaultStyle?: Style;
         footer?: TDocumentHeaderFooterFunction;
         header?: TDocumentHeaderFooterFunction;
-        images?: any;
+        // images?: any;
         info?: TDocumentInformation;
-        pageBreakBefore?: any;
+        // pageBreakBefore?: any;
         pageMargins?: Margins;
         pageOrientation?: PageOrientation;
         pageSize?: PageSize;
--- a/types/pdfmake/index.d.ts
+++ b/types/pdfmake/index.d.ts
@@ -190,7 +190,7 @@ declare module "pdfmake/build/pdfmake" {
         images?: { [key: string]: string };
         info?: TDocumentInformation;
         pageBreakBefore?: (
-            currentNode?: any,
+            currentNode?: CurrentNode,
             followingNodesOnPage?: any,
             nodesOnNextPage?: any,
             previousNodesOnPage?: any
--- a/types/pdfmake/index.d.ts
+++ b/types/pdfmake/index.d.ts
@@ -155,7 +155,7 @@ declare module "pdfmake/build/pdfmake" {
 
     interface Table {
         body: Content[][] | TableCell[][];
-        dontBreakRows?: () => boolean;
+        dontBreakRows?: boolean;
         headerRows?: number;
         heights?: Array<string | number> | TableRowFunction;
         layout?: string | TableLayoutFunctions;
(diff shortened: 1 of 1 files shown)


types/pdfmake/index.d.ts
```diff
@@ -171,16 +181,19 @@ declare module 'pdfmake/build/pdfmake' {
     }
 
     interface TDocumentDefinitions {
-        info?: TDocumentInformation;
+        background?: any;
```
evolkmann: This should not be `any`.

types/pdfmake/index.d.ts
```diff
@@ -171,16 +181,19 @@ declare module 'pdfmake/build/pdfmake' {
...
-        footer?: TDocumentHeaderFooterFunction;
         content: string | Content;
-        styles?: Style;
-        pageSize?: PageSize;
-        pageOrientation?: PageOrientation;
-        pageMargins?: Margins;
         defaultStyle?: Style;
+        footer?: TDocumentHeaderFooterFunction;
+        header?: TDocumentHeaderFooterFunction;
+        images?: any;
```
evolkmann: This should not be `any`.

types/pdfmake/index.d.ts
```diff
@@ -171,16 +181,19 @@ declare module 'pdfmake/build/pdfmake' {
...
+        info?: TDocumentInformation;
+        pageBreakBefore?: any;
```
evolkmann: This should not be `any`.

evolkmann (changes requested):
Hi @sbilbault, some comments below:
- The tests are currently failing. The tests are depending on the sample docDefinitions in `pdfmake-tests.ts`, which will have to be adopted to the new properties
- The new properties `background`, `images` and `pageBreakBefore` should not be `any`. According to the pdfmake-docs, some simple types can be created (For example `string | () => string` for [background-layer](https://pdfmake.github.io/docs/document-definition-object/background-layer/)).
- You replaced all `'` by `"`. Is there any reason or was it done by ur editorconfig/IDE?

Best, Enzo

types/pdfmake/index.d.ts
sbilbault: type added

types/pdfmake/index.d.ts
sbilbault: type added

types/pdfmake/index.d.ts
sbilbault: type added

sbilbault (commented):
Hi @evolkmann,

Thanks for the quick review !
- The tests are now passing, error was general to all branches of DefinitelyTyped yesterday
- The new properties are now typed more accurately, instead of any.
- Replacement of all ' by "  was it done by my editorconfig/IDE, I cannot prevent it.

Best, Sébastien

types/pdfmake/index.d.ts
```diff
@@ -171,16 +181,24 @@ declare module 'pdfmake/build/pdfmake' {
     }
 
     interface TDocumentDefinitions {
-        info?: TDocumentInformation;
+        background?: (currentPage?: any, pageSize?: any) => any;
```
evolkmann: This could also be a `string` and the function returns a `string` as well. The arguments are up to the developer and irrelevant for the internals of pdfmake (as far as I understand the current docs).

```
background?: () => string | string;
```

types/pdfmake/index.d.ts
```diff
@@ -171,16 +181,24 @@ declare module 'pdfmake/build/pdfmake' {
...
-        footer?: TDocumentHeaderFooterFunction;
         content: string | Content;
-        styles?: Style;
-        pageSize?: PageSize;
-        pageOrientation?: PageOrientation;
-        pageMargins?: Margins;
         defaultStyle?: Style;
+        footer?: TDocumentHeaderFooterFunction;
+        header?: TDocumentHeaderFooterFunction;
+        images?: Record<string, string>;
```
evolkmann: Can you explain what `Record<string, string>` does here? I assume you take this from [here](https://pdfmake.github.io/docs/document-definition-object/images/). In my opinion, the following is easier to read (because I assume some people would have to look up `Record`):

```
images?: { [key: string]: string };
```

types/pdfmake/index.d.ts
```diff
@@ -171,16 +181,24 @@ declare module 'pdfmake/build/pdfmake' {
...
+        info?: TDocumentInformation;
+        pageBreakBefore?: (
+            currentNode?: any,
```
evolkmann: The type of `currentNode` is available here: https://pdfmake.github.io/docs/document-definition-object/page/

evolkmann (changes requested):
Thanks again for your efforts! :) I added some more comments.

Best, Enzo

types/pdfmake/index.d.ts
evolkmann: Do you use the `Record` type in your daily work for types like this? Just out of curiosity :D

types/pdfmake/index.d.ts
sbilbault: ok

types/pdfmake/index.d.ts
sbilbault: I do use Record, but it is exactly the same as far as I understood as the `{ [key: string]: string }`

types/pdfmake/index.d.ts
sbilbault: CurrentNode type has been added

sbilbault:
Hi @evolkmann ,

Thanks for your review! I believe this should soon be ok !

Best, Sébastien

evolkmann (approved):

PR #35160, opened 2019-05-02
 Add definition for string-to-color

Bleser:

If adding a new definition:


types/string-to-color/tslint.json
```diff
@@ -0,0 +1,80 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "adjacent-overload-signatures": false,
```
weswigham: you shouldn't be disabling all these lint rules for a new package - just `{ "extends": "dtslint/dt.json" }` should do, adn the library's types should just need to be `export default function generate(obj:any): string;`.

Bleser: I corrected

PR #35161, opened 2019-05-02
Update index.d.ts for gridstack

JohnArcher:
Added missing jsdoc documentation for addWidget().
Also allow parameter id of addWidget() to be a string.
Removed unneccessary whitespaces.

https://github.com/gridstack/gridstack.js/tree/develop/doc#addwidgetel-x-y-width-height-autoposition-minwidth-maxwidth-minheight-maxheight-id

### DIFF
--- a/types/gridstack/gridstack-tests.ts
+++ b/types/gridstack/gridstack-tests.ts
@@ -1,12 +1,5 @@
 /// <reference types="jquery"/>
 
-
-
-// Type definitions for Gridstack
-// Project: http://troolee.github.io/gridstack.js/
-// Definitions by: Pascal Senn <https://github.com/PascalSenn/>, Ricky Blankenaufulland <https://github.com/ZoolWay/>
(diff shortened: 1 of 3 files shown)


JohnArcher:
Oh boy. I have no idea why this breaks. Why does this occurs: `Error: ../node/v7/index.d.ts: Definitions must use global references to other packages, not parent ("../xxx") references.(Based on reference 'base.d.ts')`

I can't believe that it has something to do with my small changes. I didn't touch any configuration. Any help would be appreciated.

jessetrinity:
Pinging @PascalSenn @ZoolWay @Sl1MBoy

@JohnArcher thanks for your contribution! If you're feeling up for it, this might be a good time to remove `"dt-header": false,` from `tslint.json` and add a version number to the header.

JohnArcher:
@jessetrinity Thanks for your hints! I commited the changes.

PR #35165, opened 2019-05-02
Updated "delaunator" typings to version 3.0

tobiaskraus:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/mapbox/delaunator

I updated the JSDoc according the updated README from mapbox/delaunator and its datatypes.
I added Delaunator.coords. As there is no  Delaunator.Node anymore (Delaunator.hull changed its datatype), I removed the namespace and exported the class Delaunator directly. So the new import syntax is:

```ts
import { Delaunator } from 'delaunator';
```

(see delaunator-tests.ts) ... which I consider good practice.

### DIFF
--- a/types/delaunator/delaunator-tests.ts
+++ b/types/delaunator/delaunator-tests.ts
@@ -1,4 +1,4 @@
-import * as Delaunator from 'delaunator';
+import { Delaunator } from 'delaunator';
 
 // Zipped points [x0, y0, x1, y1, ...]
 const zippedPoints = [168, 180, 168, 178, 168, 179, 168, 181, 168, 183, 167, 183, 167, 184];
@@ -22,9 +22,10 @@ Delaunator.from(customPoints, point => point.x, point => point.y);
 Delaunator.from(customPoints, getX, getY);
 
 // To get the coordinates of all triangles, use:
-const triangles = d.triangles;
-const halfedges = d.halfedges;
-const hull = d.hull;
+const triangles = d.triangles; // $ExpectType Uint32Array
+const halfedges = d.halfedges; // $ExpectType Uint32Array
+const hull = d.hull; // $ExpectType Uint32Array
+const coords = d.coords; // $ExpectType ArrayLike<number> | Float64Array
 const coordinates: number[][][] = [];
 for (let i = 0; i < triangles.length; i += 3) {
     coordinates.push([
@@ -33,3 +34,16 @@ for (let i = 0; i < triangles.length; i += 3) {
         defaultPoints[triangles[i + 2]]
     ]);
 }
+
+// Or use Delaunator.coords (but coords is a flat array in the form of [x0, y0, x1, y1, ...])
+const coordinates2: number[][][] = [];
+for (let i = 0; i < triangles.length; i += 3) {
+    coordinates2.push([
+        [coords[triangles[i] * 2], coords[triangles[i] * 2 + 1]],
+        [coords[triangles[i + 1] * 2], coords[triangles[i + 1] * 2 + 1]],
+        [coords[triangles[i + 2] * 2], coords[triangles[i + 2] * 2 + 1]]
+    ]);
+}
+
+// both approaches should give the same result
+JSON.stringify(coordinates) === JSON.stringify(coordinates2);
--- a/types/delaunator/delaunator-tests.ts
+++ b/types/delaunator/delaunator-tests.ts
@@ -1,4 +1,4 @@
-import { Delaunator } from 'delaunator';
+import Delaunator from 'delaunator';
 
 // Zipped points [x0, y0, x1, y1, ...]
 const zippedPoints = [168, 180, 168, 178, 168, 179, 168, 181, 168, 183, 167, 183, 167, 184];
--- a/types/delaunator/delaunator-tests.ts
+++ b/types/delaunator/delaunator-tests.ts
@@ -13,7 +13,7 @@ interface CustomPoint {
     x: number;
     y: number;
 }
-const customPoints = [{x: 168, y: 180}, {x: 168, y: 178}, {x: 168, y: 179}, {x: 168, y: 181}, {x: 168, y: 183}, {x: 167, y: 183}, {x: 167, y: 184}];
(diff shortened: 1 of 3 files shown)


tobiaskraus:
Today, the last 12 Merge Requests failed, due to the same error.

It seems like it has nothing to do with my changes. Let's wait what happens...

BTOdell:
I looked at the current Delaunator code and it exports the class as default: `export default class ...`

But these changes are describing it as using a named export. Is this correct? I'm not sure if the DefinitelyTyped build includes running the tests against the actual downloading library (which is needed to answer this question).

types/delaunator/index.d.ts
```diff
@@ -1,17 +1,19 @@
-// Type definitions for delaunator 2.0
+// Type definitions for delaunator 3.0
 // Project: https://github.com/mapbox/delaunator#readme
 // Definitions by: Denis Carriere <https://github.com/DenisCarriere>
 //                 Bradley Odell <https://github.com/BTOdell>
+//                 Tobias Kraus <https://github.com/tobiaskraus>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-declare class Delaunator<P> {
+export class Delaunator<P> {
```
DenisCarriere: Should be `export default`

https://github.com/mapbox/delaunator/blob/master/index.js#L5

types/delaunator/delaunator-tests.ts
```diff
@@ -1,4 +1,4 @@
-import * as Delaunator from 'delaunator';
+import { Delaunator } from 'delaunator';
```
DenisCarriere: Should be

```ts
import Delaunator from 'delaunator';
```

https://github.com/mapbox/delaunator/blob/master/index.js#L5

Also in their tests:

https://github.com/mapbox/delaunator/blob/master/test.js#L3

`import { Delaunator }` would resolve `undefined`

DenisCarriere (changes requested):

types/delaunator/index.d.ts
tobiaskraus: Yes, you are right. And when I test it, in an example, my version doesn't work, but export default works.

But when I change to export default and run `npm run lint delaunator` in DefinitelyTyped, I get this error:

```
Error: C:/.../DefinitelyTyped/types/delaunator/index.d.ts:8:1
ERROR: 8:1  npm-naming  The types for delaunator specify 'export default' but the source does not mention 'default' anywhere.
```

whereas with export class, the lint didn't complain.

Do you know why? What should I do? Ignore the lint?
In the meantime I try to dig deeper by myself...

tobiaskraus: I see that `npm run lint delaunator` looks into `delaunator.min.js` of npms dalaunatator package. (Don't know why, maybe because of package.json "main" attribute) But this file is meant to be required with Node / Browserify. The ES module version lays in `delaunator/index.js`.

Should I simply try to push it, even though the lint fails, or is there anything I can do to fix this lint error message?

tobiaskraus: I tried it anyway (your mentioned changes). But apparently the test breaks, and therefore cannot be merged. The error is the same as I saw locally with `npm run lint delaunator`:
https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/528247076?utm_source=github_status&utm_medium=notification

I need help here.

DenisCarriere (approved):
👍 looks good to me

Strange that the linting picks up the minified version

types/delaunator/index.d.ts
DenisCarriere: Would need help from the DefinitelyTyped team on this one

weswigham: https://unpkg.com/delaunator@3.0.2/delaunator.js is listed as the package's primary entrypoint, which does not mention a `default`.

weswigham: (It does `return Delaunator;` in a UMD pattern, so an `export = Delaunator` would be appropriate)

DenisCarriere: That's a real tricky one, so that means the Typescript definition would be different for both NodeJS and web?

`module` resolves

```ts
export default class Delaunator {
```

https://unpkg.com/delaunator@3.0.2/index.js

DenisCarriere: Correct, the UMD pattern `export = Delaunator` is correct here.

In NodeJS, if you import via CommonJS, here's what is returned.

```js
> require('delaunator')
{ [Function: Delaunator] from: [Function: from] }
```

tobiaskraus: > That's a real tricky one, so that means the Typescript definition would be different for both NodeJS and web?

@DenisCarriere yes that's how I understand it now as well.

## Another test of mine for Browser (Parcel Bundler):

I tried to change how I understood:

*index.d.ts*
```ts
declare class Delaunator<P> {
  ...
}
export = Delaunator;
```

*delaunator-test.ts*

invalid (tsc complains):
```ts
import Delaunator from 'delaunator';
// error: Module '".../DefinitelyTyped/types/delaunator/index"' has no default export

import { Delaunator } from 'delaunator';
// error: Module '".../DefinitelyTyped/types/delaunator"' has no exported member 'Delaunator'
// error: This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.
```

from https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require I see the right way of importing a module which is exported with `export = ` syntax:
```ts
import Delaunator = require('delaunator');
```

## Result
* `npm run lint delaunator` worked
* an example project of mine with parcel-bundler:
    * `tsc` didn't complain
    * but the compiled JavaScript didn't work because `Delaunator.from` was not a function as `Delaunator` was an object: `{ default: class Delaunator }` 

## Two different definition files ?
So yes, parcel-bundler used the ES module version `delaunator/index.js` version, which requires apparantly a different TypeScript Definition file than Node.js / Browserify needs (for the UMD version)

**What to do? I'm lost. It would have been my first Open Source contribution 😞**

weswigham: @tobiaskraus use the `export =` syntax and just use `esModuleInterop` plus the default imports in your project - that should check out both at compile time and runtime.

sandersn:
@DenisCariere @BTOdell the 'export default' check is simple right now. It just uses the entrypoint as defined in package.json, which is why it uses the minified code. Then it just looks for the word 'default' in the text, which is way too generous, but catches some errors.

PR #35176, opened 2019-05-03
fix(nes): add callback parameters to onDisconnect

jaulz:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/hapijs/nes/blob/master/API.md#new-clienturl-options

### DIFF
--- a/types/nes/client.d.ts
+++ b/types/nes/client.d.ts
@@ -8,7 +8,7 @@ declare class Client {
     constructor(url: string, options?: Client.ClientOptions);
     onError: (err: any) => void;
     onConnect: () => void;
-    onDisconnect: (willReconnect: boolean, log: { code: string, explanation: string, reason: string, wasClean: boolean }) => void;
+    onDisconnect: (willReconnect: boolean, log: { code: number, explanation: string, reason: string, wasClean: boolean }) => void;
     onUpdate: (message: any) => void;
     connect(options: Client.ClientConnectOptions): Promise<any>;
     connect(): Promise<any>;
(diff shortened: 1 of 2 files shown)


saboya:
[This test](https://github.com/hapijs/nes/blob/master/test/client.js#L133-L155) leads me to believe `code` is `number` instead of `string`, but I haven't actually tested it. Are you positive that it's a string?

types/nes/client.d.ts
```diff
@@ -8,7 +8,7 @@ declare class Client {
     constructor(url: string, options?: Client.ClientOptions);
     onError: (err: any) => void;
     onConnect: () => void;
-    onDisconnect: () => void;
+    onDisconnect: (willReconnect: boolean, log: { code: string, explanation: string, reason: string, wasClean: boolean }) => void;
```
saboya: [This test](https://github.com/hapijs/nes/blob/master/test/client.js#L133-L155) leads me to believe `code` is `number` instead of `string`, but I haven't actually tested it. Are you positive that it's a string?

saboya (changes requested):

types/nes/client.d.ts
jaulz: You are right, thanks for the hint!

saboya (approved):

saboya (changes requested):
Actually, I just remembered:

This definition is deprecated and is replaced by `@types/hapi__nes` since `Hapi` moved to it's own ~namespace~scope, on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34479. The new type definition is located [here](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/hapi__nes/).

This should be left untouched and those definitions should be updated instead.

jaulz:
@saboya thanks for the hint! Here we go: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35182

PR #35187, opened 2019-05-03
[x-ray] add instance.abort parameter

leonardodino:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/matthewmueller/x-ray/pull/232

---

created in this PR: https://github.com/matthewmueller/x-ray/pull/232 and called [here](https://github.com/matthewmueller/x-ray/blob/c95ef3e2a9a260155983b5ddb0eac65267f11d49/index.js#L127)
original package repo tests: [`1`](https://github.com/matthewmueller/x-ray/blob/3c5b9e4a007937c26222c10646087b046b4aab1c/test/xray_spec.js#L338
) [`2`](https://github.com/matthewmueller/x-ray/blob/3c5b9e4a007937c26222c10646087b046b4aab1c/test/xray_spec.js#L304)

### DIFF
--- a/types/x-ray/index.d.ts
+++ b/types/x-ray/index.d.ts
@@ -2,7 +2,7 @@
 // Project: https://github.com/lapwinglabs/x-ray#readme
 // Definitions by: Matt Traynham <https://github.com/mtraynham>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.3
 
 /// <reference types="node" />
 
@@ -31,7 +31,7 @@ declare namespace XRay {
     }
     type Selector = ScalarSelector | SelectorArray;
 
-    type AbortHandler = (data: any[], url: string) => boolean;
+    type AbortHandler = (data: unknown[], url: string) => boolean;
 
     interface Instance extends XRayCrawler.Instance {
         (
--- a/types/x-ray/index.d.ts
+++ b/types/x-ray/index.d.ts
@@ -31,7 +31,7 @@ declare namespace XRay {
     }
     type Selector = ScalarSelector | SelectorArray;
 
-    type AbortHandler<T extends any> = (data: T[], url: string) => boolean;
+    type AbortHandler = (data: any[], url: string) => boolean;
 
     interface Instance extends XRayCrawler.Instance {
         (
@@ -55,8 +55,7 @@ declare namespace XRay {
     interface InstanceInvocation {
         (callback: Callback): void;
         (source: string, callback: Callback): void;
-        // tslint:disable-next-line:no-unnecessary-generics
-        abort<T>(arg: AbortHandler<T>): this;
+        abort(arg: AbortHandler): this;
         paginate(selector: Selector): this;
         limit(n: number): this;
         stream(): NodeJS.ReadStream;
--- a/types/x-ray/index.d.ts
+++ b/types/x-ray/index.d.ts
@@ -31,6 +31,8 @@ declare namespace XRay {
     }
     type Selector = ScalarSelector | SelectorArray;
 
+    type AbortHandler<T extends any> = (data: T[], url: string) => boolean;
+
     interface Instance extends XRayCrawler.Instance {
         (
             source: string,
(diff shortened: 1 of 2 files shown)


mtraynham (approved):

types/x-ray/index.d.ts
```diff
@@ -53,7 +55,8 @@ declare namespace XRay {
     interface InstanceInvocation {
         (callback: Callback): void;
         (source: string, callback: Callback): void;
-        abort(): this;
+        // tslint:disable-next-line:no-unnecessary-generics
```
weswigham: Why the suppression and why the generic `AbortHandler`? `(data: unknown[], url: string) => boolean;` seems fine?

weswigham (changes requested):

types/x-ray/index.d.ts
leonardodino: actually: because I don't know better ¯\_(ツ)_/¯
I'm pretty new to `TS`. Thanks for reviewing (:

--- 

Is that the same as just shoving `any` without generics?
It would only disable the `abort<CustomType>()` syntax, right?
I read a bit of the [linked document](https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md) and didn't understand why generics are bad in this case.

weswigham: > It would only disable the abort<CustomType>() syntax, right?

Yeah.

> Is that the same as just shoving any without generics?

Kinda. `any` is extremely permissive - it effectively disables the typechecker. `never` and `unknown` are the bottom and top types, respectively. Everything is assignable to `unknown` and `never` is assignable to anything. Based on how the callback parameter is used, one of the two is likely correct. If the contents of the `data` array are never meant to be inspected (eg, because it's always empty), then `never` would be correct. If the `data` parameter may be invoked with literally anything, then `unknown` would be correct, since it'd force people who use the `data` to narrow that broad type down to the specific instances they can actually handle before using it.

So yeah - the generic here is bad because it doesn't really do anything, since it only appears on one position and has no constraint (so it's not actually constraining anything, which is what generics are for).

leonardodino: `unknown` is a newer construct, right? `3.0`?

_sorry_: the newbie questions are coming:
- I would have to bump the TypeScript version in the file header, right?
- In that case, should I bump the package version also? Or is it ok to keep it there? (`x-ray` is still at `2.3.3`, or is that unrelated?)

---

Thanks for the help here, It's my first type-only contribution

weswigham: > unknown is a newer construct, right?

Yeah, it's a more recent addition, Prior to it's addition, `{}` was the closest type to a real top, but it didn't include `undefined` or `null` in its domain.

> I would have to bump the TypeScript version in the file header, right?

I believe so, yes.

> In that case, should I bump the package version also?

Nah. The publisher tags the last version supported by any given TS major.minor version, so the old version will still be there, tagged for older versions of TS.

leonardodino: I guess I get it now:

```typescript
type AbortHandler = (data: unknown[], url: string) => boolean;
abort((results: Array<{id: string}>) => results.some(result => result.id === 'issue_40'))
```

results in: 
```
Type 'unknown[]' is not assignable to type '{ id: string; }[]'.
  Type 'unknown' is not assignable to type '{ id: string; }'
```

But if I change the first line to: 
```typescript
type AbortHandler = (data: never[], url: string) => boolean;
```

It works!

---

Although this leads me to think `never` was a really unfortunate choice of word, at least for beginners.

Reading the `unknown[]` and making it assignable would be more intuitive.

Am I holding it wrong?

weswigham: Well, so, the usage here is a 'lil strange. Over [here](https://github.com/matthewmueller/x-ray/pull/232/files#diff-168726dbe96b3ce427e7fedce31bb0bcR126), it seems like the `abort` callback is _always_ called with the `obj`, which is whatever `obj` the stream ingested. The type you annotate on this function has absolutely no bearing on what that type is - in fact, that `obj`'s shape is completely `unknown`! _Assuming_ it's an `{id: string}[]` is _optimistic_ at best, and potentially outright wrong - it's entirely possible that the `obj` has no `id` property _at all_, at which point assuming as much would crash at runtime. Therefore, using `unknown` and treating it as a completely unknown type, then safely narrowing it to known bits, eg, with `"id" in result && result.id === 'issue_404')`, you can access the unknown bits without doing something that'll maybe throw at runtime.

leonardodino: ok, got it now. (: Thanks again @weswigham.

I'll change it to `unknown`.

The `obj` follows the format of the `Selector`, so it should be possible to make it work correctly with a single top-level generic for the instance, but that's out of scope right now for me.

leonardodino: @weswigham ready for review

weswigham (approved):

PR #35193, opened 2019-05-03
fix(node): events.once with generic sign

imcotton:
## Note:
This is following up my previously PR #35130, which would be better have had this addition.

## Please fill in this template.

## Select one of these and delete the others:

If adding a new definition:

## If changing an existing definition:

## If removing a declaration:

### DIFF
--- a/types/node/events.d.ts
+++ b/types/node/events.d.ts
@@ -2,7 +2,8 @@ declare module "events" {
     class internal extends NodeJS.EventEmitter { }
 
     namespace internal {
-        function once(emitter: EventEmitter, event: string | symbol): Promise<any[]>;
+        // tslint:disable-next-line no-unnecessary-generics
+        function once<T extends any[]>(emitter: EventEmitter, event: string | symbol): Promise<T>;
          class EventEmitter extends internal {
             /** @deprecated since v4.0.0 */
             static listenerCount(emitter: EventEmitter, event: string | symbol): number;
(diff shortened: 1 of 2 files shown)


types/node/events.d.ts
```diff
@@ -2,7 +2,8 @@ declare module "events" {
     class internal extends NodeJS.EventEmitter { }
 
     namespace internal {
-        function once(emitter: EventEmitter, event: string | symbol): Promise<any[]>;
+        // tslint:disable-next-line no-unnecessary-generics
+        function once<T extends any[]>(emitter: EventEmitter, event: string | symbol): Promise<T>;
```
SimonSchick: We shouldn't do that, a type assertion should be used instead.
See https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md
Eg. a generic should not be used if it only has 1 reference.

imcotton: Thanks for the reviewing (again), I did saw this alert from `dtslint` and then read this [comment][1] from the issue, do you think it's could be a valid case for marking as `tslint ignore`, or shall letting user to use type assertion as the only way?

[1]: https://github.com/Microsoft/dtslint/issues/76#issuecomment-364201501

imcotton: /ping @SimonSchick

SimonSchick: github must've eaten my reply.
And no you should use casting since there is only a single reference to `T.

imcotton:
Thanks for the feedback, closing the PR.

PR #35200, opened 2019-05-04
Add types for react-shadow-dom-retarget-events

remcohaszing:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/react-shadow-dom-retarget-events/index.d.ts
+++ b/types/react-shadow-dom-retarget-events/index.d.ts
@@ -1,0 +1,19 @@
+// Type definitions for react-shadow-dom-retarget-events 1.0
+// Project: https://github.com/weltn24/react-shadow-dom-retarget-events#readme
+// Definitions by: Remco Haszing <https://github.com/remcohaszing>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+export = retargetEvents;
+
+/**
+ * Fixes events for react components rendered in a `shadow dom`.
+ *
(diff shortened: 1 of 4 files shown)


types/react-shadow-dom-retarget-events/index.d.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Fixes events for react components rendered in a `shadow dom`.
+ *
+ * When you render a react component inside `shadow dom` events will not be dispatched to react. I.e. when a user clicks in your react component nothing happens. This happens (or does not happen)
+ * with any events.
+ *
+ * A bug is filed at [#10422](https://github.com/facebook/react/issues/10422).
+ */
+declare function retargetEvents(shadowRoot: ShadowRoot): void;
+
+declare namespace retargetEvents {}
```
weswigham: Delete this line, set `esModuleInterop: true` in the test tsconfig, and use default imports in the test.

weswigham (changes requested):

types/react-shadow-dom-retarget-events/index.d.ts
remcohaszing: Thank you for your feedback. I’ll gladly change this. However, I wonder why.

I submitted #35199 at around the same time as this PR. Both describe modules that export just one function that is assigned to `module.exports` in the original JavaScript code.

Why should this module be using `esModuleInterop: true`, but should `@types/inflation` not?

weswigham: `inflate` actually uses the namespace to hold a type, So it can keep the namespace. It should probably still use `esModuleInterop` and a default import for the test, but I'm usually more concerned with the type definition file correctness.

remcohaszing: Thanks for explaining! I changed the code as you requested.

PR #35201, opened 2019-05-04
[graphlib] Add a type variable for graph labels

mdlavin:
Allow strict typing of node labels


mdlavin:
If the removing of the extra whitespace is an issue, I can put up a fix to restore it

mdlavin:
The typing isn't quite right, I'm closing until I have it working better

PR #23198, opened 2018-01-25
lodash: Work with classic module resolution

ghost:
Fixes https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23164#issuecomment-360555123

With classic module resolution we need to write `/index` explicitly.

PR #35203, opened 2019-05-04
[lodash] `partition` allow setting tuple type

nikolay-borzov:
Essentially it allows specifying the tuple type by using type guard when `partition` is used to split array by item types

```
const [numbers, strings] = partition([1, 2, '3', '4'], (value): value is number => typeof value === 'number');
// or
const [numbers, strings] = partition([1, 2, '3', '4'], _.isNumber);
```

I wasn't able to run lint because is TS error

```
declare module "../index"
```

> Cannot augment module '"DefinitelyTyped/types/lodash/index.d.ts"' with value exports because it resolves to a non-module entity.ts(2649)

### DIFF
--- a/types/lodash/common/collection.d.ts
+++ b/types/lodash/common/collection.d.ts
@@ -1777,6 +1777,10 @@ declare module "../index" {
         * @param callback The function called per iteration.
         * @return Returns the array of grouped elements.
         **/
+        partition<T, U extends T>(
+            collection: List<T> | null | undefined,
+            callback: ValueIteratorTypeGuard<T, U>
+        ): [U[], Array<Exclude<T, U>>];
         partition<T>(
             collection: List<T> | null | undefined,
             callback: ValueIteratee<T>
@@ -1795,6 +1799,10 @@ declare module "../index" {
         /**
          * @see _.partition
          */
+        partition<T, U extends T>(
+            this: LoDashImplicitWrapper<List<T> | null | undefined>,
+            callback: ValueIteratorTypeGuard<T, U>
+        ): LoDashImplicitWrapper<[U[], Array<Exclude<T, U>>]>;
         partition<T>(
             this: LoDashImplicitWrapper<List<T> | null | undefined>,
             callback: ValueIteratee<T>
@@ -1813,6 +1821,10 @@ declare module "../index" {
         /**
          * @see _.partition
          */
+        partition<T, U extends T>(
+            this: LoDashExplicitWrapper<List<T> | null | undefined>,
+            callback: ValueIteratorTypeGuard<T, U>
+        ): LoDashExplicitWrapper<[U[], Array<Exclude<T, U>>]>;
         partition<T>(
             this: LoDashExplicitWrapper<List<T> | null | undefined>,
             callback: ValueIteratee<T>
--- a/types/lodash/lodash-tests.ts
+++ b/types/lodash/lodash-tests.ts
@@ -2794,6 +2794,8 @@ _.chain([1, 2, 3, 4]).unshift(5, 6); // $ExpectType LoDashExplicitWrapper<number
 
 // _.partition
 {
+    const mixedArray = [1, 2, '3', '4'];
+
     // $ExpectType [any[], any[]]
     _.partition(anything, (value) => {
         value; // $ExpectType any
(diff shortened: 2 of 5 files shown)


types/lodash/common/collection.d.ts
```diff
@@ -1781,6 +1781,10 @@ declare module "../index" {
             collection: List<T> | null | undefined,
             callback: ValueIteratee<T>
         ): [T[], T[]];
+        partition<T1 | T2>(
```
aj-r: I don't think this is valid syntax? Should be `<T1, T2>`.

types/lodash/common/collection.d.ts
```diff
@@ -1781,6 +1781,10 @@ declare module "../index" {
...
+            collection: List<T1 | T2> | null | undefined,
+            callback: ValueIteratee<T1 | T2>
```
aj-r: I think this should be something with a type guard, maybe:

`ValueIteratorTypeGuard<T1 | T2, T1>`

Or basically:

`(value: T1 | T2) => value is T1`

Also, this new overload should probably be the first overload. Otherwise TS will always choose the old overload, unless you explicitly specify the type parameters.

types/lodash/lodash-tests.ts
```diff
@@ -2809,6 +2811,8 @@ _.chain([1, 2, 3, 4]).unshift(5, 6); // $ExpectType LoDashExplicitWrapper<number
         value; // $ExpectType AbcObject
         return true;
     });
+    // $ExpectType [number[], string[]]
+    _.partition<number, string>(mixedArray, (value) => typeof value === 'number');
```
aj-r: If you use the type guard as suggested in my other comment, then you should be able to write this without explicitly specifying any generic types, e.g.

```ts
 _.partition(mixedArray, (value): value is number => typeof value === 'number');
```

or more simply:

```ts
 _.partition(mixedArray, _.isNumber);
```

types/lodash/readme.md
```diff
@@ -47,16 +47,16 @@
   - In general, no.
 - If the wrapper functions are almost copies of the original functions, shouldn't we auto-generate them like we do for `lodash/fp`?
   - Good idea! If you have time, submit a PR.
-- When I ran `npm run lint lodash`, I got an error that loks like `<--- Last few GCs --->`.
+- When I ran `npm run lint lodash`, I got an error that looks like `<--- Last few GCs --->`.
```
aj-r: Thank you for fixing these typos!

aj-r (changes requested):

aj-r:
@andy-ms Any idea what to do about the Travis error?

```
../lodash/common/collection.d.ts(2,16): error TS2649: Cannot augment module '"/home/travis/build/DefinitelyTyped/DefinitelyTyped/types/lodash/index.d.ts"' with value exports because it resolves to a non-module entity.
```

It seems to be unrelated to this PR. Also only happens on `typescript@next` according to https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23198#issuecomment-489350194.

types/lodash/common/collection.d.ts
nikolay-borzov: Right. Sorry I made this change blindly

types/lodash/common/collection.d.ts
nikolay-borzov: Will try this

types/lodash/lodash-tests.ts
nikolay-borzov: Updated and moved `partition` definition 
```
partition<T1, T2>(
  collection: List<T1 | T2> | null | undefined,
  callback: ValueIteratorTypeGuard<T1 | T2, T1>
): [T1[], T2[]];
partition<T>(
  collection: List<T> | null | undefined,
  callback: ValueIteratee<T>
): [T[], T[]];
```

```
_.partition(mixedArray, (value): value is number => typeof value === 'number');
```
> Expected type to be:
  [number[], string[]]
got:
  [number[], (string | number)[]]

TypeScript thinks that T2 is `number | string`

types/lodash/common/collection.d.ts
nikolay-borzov: Fixed

types/lodash/lodash-tests.ts
aj-r: Weird, I guess you'll need to keep the explicit `<number, string>` then.

Actually, another thing you could try:

```ts
partition<T, U extends T>(
    collection: List<T> | null | undefined,
    callback: ValueIteratorTypeGuard<T, U>
): [U[], Array<Exclude<T, U>>];
```

This should forcibly exlude `number` from the second array in the return value. It should also work better when converted into the `fp.d.ts` format. (Note that the generic types represent slightly different things now - `T` is `number|string` and `U` is `number`.)

nikolay-borzov: Looks like the alternative approach works!

aj-r (approved):

PR #35223, opened 2019-05-06
chrome: make tabs.discard arguments optional

jaens:
These arguments are both optional, see:
https://developer.chrome.com/extensions/tabs#method-discard


fredrikhr (commented):
LGTM! 👍

PR #35229, opened 2019-05-06
Added callback to fluxible rehydrate function definition

JYCabello:
The parameter exists in the library.

- [x] Add or edit tests to reflect the change. (Run with `npm test`.) => No required changes observed

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/yahoo/fluxible/blob/7ad54c83ee9694af8733918b32fc38bdec057ecf/packages/fluxible/lib/Fluxible.js#L179
- [x] Increase the version number in the header if appropriate. => It's targeting the same version of the package
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. => Already existed.

### DIFF
--- a/types/fluxible/index.d.ts
+++ b/types/fluxible/index.d.ts
@@ -90,7 +90,7 @@ export class Fluxible {
      * @param callback
      * @async Rehydration may require more asset loading or async IO calls
      */
-    rehydrate(state: any, callback: (err: Error, context: FluxibleContext) => void): void;
+    rehydrate(state: any, callback: (err: Error, context?: FluxibleContext) => void): void;
 }
 
 /**
--- a/types/fluxible/index.d.ts
+++ b/types/fluxible/index.d.ts
@@ -90,7 +90,7 @@ export class Fluxible {
      * @param callback
      * @async Rehydration may require more asset loading or async IO calls
      */
-    rehydrate(state: any, callback: (err: Error, context?: FluxibleContext) => void): void;
+    rehydrate(state: any, callback?: (err: Error, context: FluxibleContext) => void): void;
 }
 
 /**
--- a/types/fluxible/index.d.ts
+++ b/types/fluxible/index.d.ts
@@ -90,7 +90,7 @@ export class Fluxible {
      * @param callback
      * @async Rehydration may require more asset loading or async IO calls
      */
-    rehydrate(state: any): void;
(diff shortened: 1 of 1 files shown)


types/fluxible/index.d.ts
```diff
@@ -90,7 +90,7 @@ export class Fluxible {
      * @param callback
      * @async Rehydration may require more asset loading or async IO calls
      */
-    rehydrate(state: any): void;
+    rehydrate(state: any, callback: () => void): void;
```
xbIm: i think it should be callback: (err: Error, context: FluxibleContext) => void

please fix

xbIm (changes requested):
i think it should be callback: (err: Error, context: FluxibleContext) => void

please fix

types/fluxible/index.d.ts
JYCabello: You are absolutely right. I even noticed the parameter was missing because I had a function with that signature. Sorry there and thanks for catching it up.

It's fixed now.

JYCabello: And it should be optional, making it so.

xbIm: right, thanks for the fix, please fill free to commit new fixes if find any

xbIm (approved):
approved

types/fluxible/index.d.ts
```diff
@@ -90,7 +90,7 @@ export class Fluxible {
      * @param callback
      * @async Rehydration may require more asset loading or async IO calls
      */
-    rehydrate(state: any): void;
+    rehydrate(state: any, callback: (err: Error, context?: FluxibleContext) => void): void;
```
jessetrinity: Did you mean to make this optional?
http://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#optional-parameters-in-callbacks

yeraytba: No, I meant to make the callback itself optional. I'll prepare another PR.

JYCabello: You are right, I meant to make the **callback** itself optional. Updating it.

JYCabello: Fixed, thanks a lot.

xbIm (approved):
approved

PR #35234, opened 2019-05-06
fix(yargs): Command builder function can return void

ExE-Boss:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/yargs/yargs/blob/master/docs/advanced.md#describing-positional-arguments>

---

The `yargs.command(…)` builder is allowed to return void.


forivall:
Note that with this change, when the user returns `void`, they will not get the inferred arguments type from the arguments defined in the builder. We could either
* disallow your pr change, to force the user into returning the value, to include the typing, or
* you should add a doccomment to note that, when the user returns `void`, the inferred arguments type will _not_ include those typings

What do other maintainers think?

forivall (changes requested):
On the documentation for the "builder" option, please note that if void is returned, then the arguments type in the handler will not include the command-specific arguments.

forivall (approved):
Sorry for the delay

types/yargs/index.d.ts
```diff
@@ -130,12 +131,14 @@ declare namespace yargs {
          * Set `description` to false to create a hidden command. Hidden commands don't show up in the help output and aren't available for completion.
          * @param [builder] Object to give hints about the options that your command accepts.
          * Can also be a function. This function is executed with a yargs instance, and can be used to provide advanced command specific help.
+         *
+         * Note that when `void` is returned, the handler `argv` object type will not include command-specific arguments.
          * @param [handler] Function, which will be executed with the parsed `argv` object.
          */
-        command<U>(command: string | ReadonlyArray<string>, description: string, builder?: (args: Argv<T>) => Argv<U>, handler?: (args: Arguments<U>) => void): Argv<T>;
+        command<U = T>(command: string | ReadonlyArray<string>, description: string, builder?: (args: Argv<T>) => Argv<U> | void, handler?: (args: Arguments<U>) => void): Argv<T>;
```
uniqueiniquity: Instead of what you have here, I'd recommend defining the following type and using it:
```ts
type BuilderCallback<T,U> = ((args: Argv<T>) => Argv<U>) | ((args: Argv<T>) => void)
```
What you have currently doesn't really make sense - `void` isn't a type in the normal sense, but rather a way to describe that a function doesn't return.

uniqueiniquity (changes requested):

types/yargs/index.d.ts
ExE-Boss: Will that still work correctly with inferring `U`?

uniqueiniquity: I'm not sure - I'm not familiar enough with `yargs` to write a test that would demonstrate it working correctly or not.
Locally, I tried using the type i suggested inline (that is, instead of defining `BuilderCallback`) and everything works as expected, including the test you added.

ExE-Boss: It seems to be working fine even with `BuilderCallback`.

uniqueiniquity (approved):

PR #35235, opened 2019-05-06
[react-redux]: Add an explicit dependency on hoist-non-react-statics

### LINKED ISSUE #33690: [@types/react-redux] 'hoist-non-react-statics' has no exported member 'NonReactStatics'

### LINKED ISSUE #34473: [@types/react-redux] connected component is typed as any
- [X] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
  - Authors: @tkqubo @kenzierocks @clayne11 @tansongyang @NicholasBoll @mDibyo @pdeva @Kallikrein @jrakotoharisoa @apapirovski @surgeboris @soerenbf 

After last minor/patch update (@types/react-redux@7.0.6, TSC 3.4.1), `TestContainer` below is typed as `any`. `Connector` seem to be typed correctly.

```
const Test: React.SFC<{ active: boolean }> = () => <div />

const Connector = connect(
  (state: State) => ({ config: state.config })
)

const TestContainer = Connector(Test)
```

Tips? Bug?

schmod:
Will hopefully fix #33690 and fix #34473  by introducing a direct dependency on `hoist-non-react-statics` v3.3.x.

`hoist-non-react-statics` v2.5.x is still commonly-used, and shipped its own `index.d.ts` definitions.  Unfortunately, because this package is widely-used, npm/yarn will often produce a package tree that places `hoist-non-react-statics` above `@types/react-redux`, causing TS to incorrectly resolve the v2.5 definitions instead of the correct v3.3 definitions in `@types`. 

Adding an explicit dependency on the (type-less) v3.3 package appears to force TS to look in `@types` for the correct definition, when it resolves the dependency.

Please fill in this template.

- **N/A** -- this is a dependency change to deal with a module-resolution quirk.

### DIFF
--- a/types/react-redux/package.json
+++ b/types/react-redux/package.json
@@ -2,6 +2,7 @@
     "private": true,
     "dependencies": {
         "@types/hoist-non-react-statics": "^3.3.0",
+        "hoist-non-react-statics": "^3.3.0",
         "redux": "^4.0.0"
     }
 }


abirmingham:
@schmod are we sure this is the right solution? It seems a bit odd for `@types/react-redux` to depend on a typeless package. Tagging @weswigham because he was active #33690 (sorry Wesley no good deed goes unpunished I suppose).

weswigham:
It's odd, but it's to express that it's _incompatible_ with the lower version which bundles incompatible types. It's super strange, but the only other thing I could think of would be something like `"peerDependencies": { "hoist-non-react-statics": "!firstWithTypes-lastWithTypes" }`
if you catch my drift, and I'm not even sure a semver constraint like that actually can be expressed.

kenzierocks:
Reading the `semver` package (and testing on command-line) seems to indicate that `<2.2 || ^3` is valid. I looked up when the definitions were added and removed, so adding that as a `peerDependency` should do the trick.

weswigham:
If we can validate that still solves the original issue (so the types package won't be installed alongside the incompatible versions), that'd probably be ok.

PR #35236, opened 2019-05-06
[chrome] Add missing callbacks to browserAction/pageAction

tophf:
The callback parameter is supported since Chrome 67 on the methods listed below, thanks to https://crrev.com/c/979451

<details><summary>The list.</summary>

* pageAction.show
* pageAction.hide
* pageAction.setTitle
* pageAction.setPopup

* browserAction.enable
* browserAction.disable
* browserAction.setBadgeBackgroundColor
* browserAction.setBadgeText
* browserAction.setTitle
* browserAction.setPopup
</details>
Please fill in this template.

  * https://developer.chrome.com/extensions/browserAction
  * https://developer.chrome.com/extensions/pageAction
  * https://crrev.com/c/979451 - this is where the change was introduced in Chrome 67

### DIFF
--- a/types/chrome/index.d.ts
+++ b/types/chrome/index.d.ts
@@ -488,29 +488,42 @@ declare namespace chrome.browserAction {
      * Since Chrome 22.
      * Enables the browser action for a tab. By default, browser actions are enabled.
      * @param tabId The id of the tab for which you want to modify the browser action.
+     * @param callback Supported since Chrome 67
      */
-    export function enable(tabId?: number): void;
-    /** Sets the background color for the badge. */
-    export function setBadgeBackgroundColor(details: BadgeBackgroundColorDetails): void;
-    /** Sets the badge text for the browser action. The badge is displayed on top of the icon. */
-    export function setBadgeText(details: BadgeTextDetails): void;
-    /** Sets the title of the browser action. This shows up in the tooltip. */
-    export function setTitle(details: TitleDetails): void;
+    export function enable(tabId?: number, callback?: () => void): void;
+    /**
+     * Sets the background color for the badge.
+     * @param callback Supported since Chrome 67
+     */
+    export function setBadgeBackgroundColor(details: BadgeBackgroundColorDetails, callback?: () => void): void;
+    /**
+     * Sets the badge text for the browser action. The badge is displayed on top of the icon.
+     * @param callback Supported since Chrome 67
+     */
+    export function setBadgeText(details: BadgeTextDetails, callback?: () => void): void;
+    /**
+     * Sets the title of the browser action. This shows up in the tooltip.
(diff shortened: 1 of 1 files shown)


couven92:
@tophf Nice1! Could you use the text from the documentation page instead of `Supported since Chrome 67.`? Not that the documentation text is really that enlightening, but it keeps it consistent with the rest of the definitions. Maybe add `@since Since Chrome 67.` tags on the changed methods?

tophf:
>Could you use the text from the documentation page instead of Supported since Chrome 67.? Not that the documentation text is really that enlightening, but it keeps it consistent with the rest of the definitions. 

It's already inconsistent in types/chrome/index.d.ts: there are several variations of the phrase. Some of the other API definitions have just `@param callback` and no description at all. I don't know who would need that duplicated verbalization of `() => {}` (and similar self-evident parameter definitions) into English. 

>Maybe add @since Since Chrome 67. tags on the changed methods?

The methods existed in [much] older versions of Chrome so it would be misleading AFAIK because `@since` usually applies to the entire target of jsdoc, which is the API method in this case.

---

The problem I wanted to address is that specifying the callback in an older version of Chrome would throw an error and the API won't even run. I've tried to preserve the useless text plastered on other `@callback` descriptions, but the result was overly verbose.

Now that I've investigated more, I see one decent and actually meaningful description, which we should probably use instead of all the existing meaningless ones:

> Callback on success, or on failure (in which case runtime.lastError will be set).

* We could use it to replace all the meaningless callback descriptions.
* As for the methods in this PR, we could combine it like maybe "Callback on success, or on failure (in which case runtime.lastError will be set). Supported since Chrome 67"

couven92:
Hmm... yeah, true, that is annoying...Well, the callbacks are all optional, so the "Since Chrome 67" doesn't actually matter (because of backwards-compatability).

And for the `@param callback` tag, I don't really know... I usually just wrote an inline Markdown codeblock (using backtick, or triple-backtick) and used the docs verbatim, but that's not necessarily ideal in my opinion either

tophf:
If an extension specified the callback and its user runs it in an old Chrome, the extension will fail. This is a rare case though since older versions of Chrome aren't supported anyway so we could omit this 'since' notice indeed.

sandersn:
@couven92 @tophf I can't tell from reading this thread -- is this PR ready to go or is there further discussion needed to figure out how to document the callbacks?

tophf:
We've discussed the descriptions of `@callback` and I'd be fine with any decision, but my preference is to keep my current text.

sandersn:
@couven92 are you OK with the text as it currently is?

couven92:
Oh, yeah, sorry, go ahead, it's fine! 👍

PR #19356, opened 2017-08-26
[lodash] Better chaining wrappers, overload clean-up

aj-r:
Fixes #14337, Fixes #19228, Fixes #14748, Fixes #18956, Fixes #14282

Apologies for the massive PR. The fix for #14337 required me to modify the chaining wrappers for every single function, so I figured I should clean up some of the messy overloads along the way (which also fixed several other issues).

Unfortunately, this will be a breaking change in terms of what type arguments are accepted. So if anyone was explicitly passing type arguments (e.g. `_.chain([]).find<_.LoDashExplicitObjectWrapper<any>>('a')`), their code may be broken. But I think in the long run this change should be more helpful than harmful, as it removes the need to explicitly pass type arguments in almost all cases.

Summary of changes:

- The chain wrapper interfaces have been greatly simplified. Instead of many wrapper types (`LoDashImplicitArrayWrapper`, `LoDashImplicitObjectWrapper`, `LoDashImplicitNumberArrayWrapper`, etc.), there are now only 2 wrapper types: `LoDashImplicitWrapper` and `LoDashExplicitWrapper`.
  - This allows the return types to be more accurate (especially for functions like `head` and `find`).
- Cleaned up messy function overloads (removed unnecessary type arguments)
  - Made `iteratee` arguments more consistent - now they all use one of the new `Iteratee` types, unless they have a good reason not to.
- Replaced `Object` and `Function` types with more appropriate types
- Changed many `Dictionary<T>` inputs to just take plain objects, so an index signature is not required

Please fill out this template:

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://lodash.com/docs/4.17.4
- [x] Increase the version number in the header if appropriate. (not appropriate)
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #35241, opened 2019-05-06
Make lodash faster

sandersn:
This change makes lodash faster for Typescript versions 3.1 and above. 

The principle change is to undo most of #19356, which implemented the chaining methods of lodash via two interfaces. Each interface contains all the methods of lodash, and each methods has overloads for each of the types supported by lodash -- list or object, for example. The overloads are distinguished by their `this` parameter. Something like this:

```ts
interface Wrapper<T> {
  map<U, V>(this: List<U>, f: (u: U) => V): Wrapper<List<V>>;
  map<U, V>(this: Dictionary<U>, f: (u: U) => V): Wrapper<List<V>>;
}
```

This is abysmally slow. With Typescript 3.3, completions take 8 seconds or so to show up. After this PR, they take 1 second. With Typescript 3.4, both the compiler and language service often run out of memory, but this is due to 3.4's over-reliance on structural comparison. 3.5 will return performance to that of 3.3.

It creates a gigantic type with every method in lodash, and each method has many overloads, and each overload is generic. And each overload has an extra parameter, the `this` parameter, that is almost always generic, and almost always a union. In order to compare instances of lodash wrappers, each method must be instantiated, including every one of its overloads. That means each overload must be instantiated, including every one of its parameters.

Using `this` parameters to dual-dispatch at compile-time is clever, but has two usability drawbacks as well:

1. Error messages are "The this-context 'number' does not match the type of the this parameter 'List<T> | null | undefined' for 'map'.", not "'map' is not found on 'Primitive'".
2. The language service attempts to work around this by filtering methods by their this context. However, this is extremely slow.

This PR changes lodash to use an inheritance hierarchy instead. `LodashImplicitWrapper` now has five subtypes: `Collection`, `Object`, `Function`, `String` and `Primitive`. Lodash's call signature now has overloads that produce each of these wrapper types.

With an inheritance hierarchy, LodashImplicitWrapper contains the shared methods that truly apply the same way to all types. Then each subtype contains type-specific methods. Each method has many fewer overloads -- often none at all -- and is usually not generic. And the number of parameters shrinks greatly without this parameters everywhere. It also means that error messages and language service responsiveness improve.

The two drawbacks are:

1. More duplicated code, in particular methods like `concat` that apply *almost* the same way to each type, except for `Collection`. `Collection` can't unsoundly override a `concat` from `LodashImplicitWrapper`, so each subtype has to have its own `concat`.
2. For `any`, showing all methods that *might* apply can be helpful for JS users. However, I think the situation is not that bad right now, and can perhaps be improved. `any` has its own overload with `_(any)`, which currently returns `Collection<any>` but could return an intersection type.

Notes on implementation:

1. The non-chaining methods are barely changed. There are some inefficiencies there, but the scale is pretty small since `LodashStatic` is never instantiated; it's just a big collection of generic signatures.
2. The implementation uses conditional types and modern mapped-type distributivity to retain most of the chaining from the previous implementation. That is, if a List.reduce returns an object, it will be wrapped in an Object wrapper. If it returns a number, it will be wrapped in a Primitive wrapper. However, the conditional type is only used where it produces a better type &mdash; about 25 times. Most functions in lodash don't change the container type.
Using modern features makes this PR possible, but means that older versions of Typescript can't take advantage of it. I put the new code in a `ts3.1/` subdirectory so that both versions can be maintained until 2.x is old enough to drop support for.
3. The most interesting part of the code is `common/common.d.ts` and `common/seq.d.ts`, which have the implicit chain constructor and explicit chain constructor, respectively.
5. lowdb relies on lodash's chaining API, but created a copy of it using a script. The only thing the copy is used for is to distinguish the return type of `write` between the sync and async API in lowdb. Since lowdb is a rarely used package, and the return type of `write` is rarely used, I instead fudged its return type as `T & Promise<T>`, combined the sync and async APIs, and got rid of the script-generated copy.


sandersn:
Hey @typescript-bot, it's not actually new! The lodash maintainers need to take a look at it still:
@bczengel
@chrootsu
@stepancar
@aj-r
@ailrun
@e-cloud
@thorn0
@jtmthf
@DomiR

types/lodash/ts3.1/common/array.d.ts
```diff
@@ -0,0 +1,2090 @@
...
+         * Creates an array of elements split into groups the length of size. If collection can’t be split evenly, the
+         * final chunk will be the remaining elements.
+         *
+         * @param array The array to process.
+         * @param size The length of each chunk.
+         * @return Returns the new array containing chunks.
+         */
+        chunk<T>(array: List<T> | null | undefined, size?: number): T[][];
+    }
+    interface Collection<T> {
```
aj-r: I'm not a huge fan of some of these names (`Collection<T>`,  `Object<T>`) because they sound like other types. MY C#/Java brain looks at `Collection<T>` and thinks it's a lot like `List<T>` or `Array<T>`, but it's not. That said I can't think of a better name that's not painfully long.

aj-r (changes requested):
I don't have time to review the whole thing yet, but here is a preliminary review. This is really exciting because the slowness and memory issues have been a huge pain point.

I abandoned this inheritance approach because it had some critical issues at the time, especially with the explicit wrappers, but it looks like conditional types solve a lot of those issues.

types/lodash/ts3.1/common/common.d.ts
```diff
@@ -0,0 +1,281 @@
...
+        * kebabCase, last, lastIndexOf, lowerCase, lowerFirst, lt, lte, max, maxBy, mean, meanBy,
+        * min, minBy, multiply, noConflict, noop, now, nth, pad, padEnd, padStart, parseInt, pop,
+        * random, reduce, reduceRight, repeat, result, round, runInContext, sample, shift, size,
+        * snakeCase, some, sortedIndex, sortedIndexBy, sortedLastIndex, sortedLastIndexBy, startCase,
+        * startsWith, stubArray, stubFalse, stubObject, stubString, stubTrue, subtract, sum, sumBy,
+        * template, times, toFinite, toInteger, toJSON, toLength, toLower, toNumber, toSafeInteger,
+        * toString, toUpper, trim, trimEnd, trimStart, truncate, unescape, uniqueId, upperCase,
+        * upperFirst, value, and words.
+        **/
+        <TrapAny extends { __trapAny: any }>(value: TrapAny): Collection<any>;
```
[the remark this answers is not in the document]
sandersn: No, it doesn't break things. I didn't have time to check performance in the language service yet, but that is my main concern. The overloaded signatures that result from intersecting methods will be less useful than hand-written ones, but that shouldn't matter for `any`.

types/lodash/ts3.1/common/array.d.ts
sandersn: I'd like to use `List`, but got a name conflict with `type List<T> = ArrayLike<T>`. Maybe if I change it to `interface List<T> extends ArrayLike<T> { }`.

types/lodash/ts3.1/common/common.d.ts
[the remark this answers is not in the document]
sandersn: Later: Performance is fine. Explicit chains that end with `ExpChain` result in unions instead of intersections, but I think that's fine -- from looking at uses of lodash, I know that both `_.chain` and long method chains are fairly rare.

types/lodash/ts3.1/common/array.d.ts
sandersn: Later: I tried that and compilation never finished.

Other ideas are welcome!

types/lodash/ts3.1/common/array.d.ts
```diff
@@ -0,0 +1,2090 @@
...
+        /**
+         * @see _.first
+         */
+        first(): string | undefined;
+    }
+    interface StringChain {
+        /**
+         * @see _.first
+         */
+        first(): StringChain;
```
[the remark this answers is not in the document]
sandersn: Here are the methods that might return undefined (as found by a simple grep, not probably missing a few):

* first/last
* sample
* find[Last]Key
* min/max[By]
* pop/shift
* reduce[Right]
* get
* nth

So, 15 in total.
I'll introduce `StringNullableChain` and see how it affects performance. If the hit of introducing more overloads to `chain/ExpChain` and another set of methods is too great then I'll hold off on this change. I don't think losing nullability in explicit chains is *that* bad because `_.chain` is relatively rare for lodash.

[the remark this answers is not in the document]
sandersn: Actually push, pop, min, max, sample, findKey, nth and reduce already notate nullability on the collection and object methods, they just need to use ExpChain to get the correct. get does not, but I don't think it's worthwhile to introduce an CollectionNullableChain for one method.

I'll push a commit with StringNullableChain since improves nullability tracking for strings a decent amount.

aj-r:
FYI, #35203 was just merged which is probably causing the conflicts.

types/lodash/ts3.1/common/array.d.ts
aj-r: Collection is probably fine. I was thinking something like `CollectionWrapper` or `ListWrapper` (and `ObjectWrapper`, `PrimitiveWrapper`, etc) to indicate that they aren't real collecttions, they are just wrappers. I don't feel too strongly though.

types/lodash/ts3.1/common/array.d.ts
```diff
@@ -0,0 +1,2120 @@
...
+         * @param array The array to recursively flatten.
+         * @return Returns the new flattened array.
+         */
+        flattenDeep<T>(array: ListOfRecursiveArraysOrValues<T> | null | undefined): T[];
+    }
+    interface Collection<T> {
+        /**
+         * @see _.flattenDeep
+         */
+        flattenDeep(): Collection<T>;
```
aj-r: Too bad we can't really get the right type here. I feel like `T` is almost always the wrong type though. Maybe `Collection<any>` would be better? Or maybe something like what you did for `flatMapDeep`:

```ts
flattenDeep<TResult = any>(): Collection<TResult>;
```

Same below.

types/lodash/ts3.1/common/array.d.ts
```diff
@@ -0,0 +1,2120 @@
...
+        /**
+         * @see _.head
+         */
+        head(): string | undefined;
+    }
+    interface StringChain {
+        /**
+         * @see _.head
+         */
+        head(): StringChain;
```
aj-r: I think this should also be StringNullableChain? Since `_.chain("").head() === undefined`

types/lodash/ts3.1/common/common.d.ts
```diff
@@ -0,0 +1,286 @@
...
+declare module "../index" {
+    type Omit<T, K extends keyof any> = Pick<T, Exclude<keyof T, K>>;
+    type PartialObject<T> = GlobalPartial<T>;
+    type Many<T> = T | ReadonlyArray<T>;
+    type ImpChain<T> =
+        T extends { __trapAny: any } ? Collection<any> :
+        T extends null | undefined ? Primitive<T> :
+        T extends string | null | undefined ? String :
+        T extends (...args: any) => any ? Function<T> :
+        T extends List<infer U> | null | undefined ? Collection<U> :
```
aj-r: Ah, I didn't noticate before that these chain types incorporated null/undefined already. The only issue is that when you call `value()`, the null/undefined information will be lost. Maybe that's ok?

types/lodash/ts3.1/common/common.d.ts
```diff
@@ -0,0 +1,286 @@
...
+        * toString, toUpper, trim, trimEnd, trimStart, truncate, unescape, uniqueId, upperCase,
+        * upperFirst, value, and words.
+        **/
+        <TrapAny extends { __trapAny: any }>(value: TrapAny): Collection<any> & Function<any> & Object<any> & Primitive<any> & String;
+        <T extends null | undefined>(value: T): Primitive<T>;
+        (value: string | null | undefined): String;
+        <T extends (...args: any) => any>(value: T): Function<T>;
+        <T = any>(value: List<T> | null | undefined): Collection<T>;
+        <T extends object>(value: T | null | undefined): Object<T>;
+        <T>(value: T): Primitive<T>;
```
aj-r: Would it be better to combine these into a single overload:

```ts
<T>(value: T): ImpChain<T>;
```

to avoid defining all of these rules twice? It looks like this is already getting out of sync with `ImpChain`. Or would that be bad for performance?

aj-r (changes requested):

aj-r (changes requested):
Finally found time to review the whole thing :)

types/lodash/ts3.1/lodash-tests.ts
```diff
@@ -0,0 +1,7046 @@
+// tslint:disable-next-line:no-relative-import-in-test
+import fp = require("./fp");
```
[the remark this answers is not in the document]
sandersn: Module resolution:

Here's why 'lodash' works:
1. Lookup lodash from type root. 
2. Look for lodash.d.ts, but not present.
3. Look for `lodash` folder: found.
4. Look at package.json for typesVersions redirects.
5. Look in lodash/ts3.1 for index.d.ts: found.
2. Look at package.json

Here's how 'lodash/fp' fails:
1. Lookup lodash/fp from type root.
2. Look for lodash/fp.d.ts: found.

Resolution never gets to steps (3) and (4), which trigger the redirect. Instead, you get the parent lodash/fp.d.ts, not lodash/ts3.1/fp.d.ts.

This only applies to DT itself; the ts3.1 packages published on npm will only have the contents of the ts3.1 directory.

types/lodash/ts3.1/common/array.d.ts
sandersn: Well, I think the type will be a supertype of the correct type at least. Consider:

```ts
[1, [2, 3, [3, 4]], 6, 7]: (number | (number | number[])[])[]
```

flattenDeep of that should be `number[]` but `number[]` is a subtype of `(number | (number | number[])[])[]` at least, so you can manually narrow using casts probably. Type parameters are just an alternate syntax for a cast in this case, and I'd rather not encourage that. So I'll skip the type parameter here; `flatMapDeep` was an existing use of this pattern that I didn't want to break.

types/lodash/ts3.1/common/array.d.ts
sandersn: Nice catch!

types/lodash/ts3.1/common/common.d.ts
sandersn: I think it's fine to be a bit lossy, although we might re-evaluate after people have a chance to use the new code. This only makes a difference when `first/last` happens at the end of a long chain of wrapped method calls, and I don't think that happens very often.

types/lodash/ts3.1/common/common.d.ts
sandersn: Good idea. I'll test it. Conditional types also distribute across unions, so that may cause less coherent types. So I'll check for that too.

sandersn: Later: distribution makes null checking way too strict. 390 errors in the tests, all based around `Dictionary<number> | null | undefined` resulting in `Object<Dictionary<number>> | Primitive<null> | Primitive<undefined>`.

It's also 25% less efficient.

sandersn:
All right, I think this is ready to go.

aj-r (approved):

b4dnewz:
This part at top of declaration file:
```ts
declare module "lodash" {
    interface ObjectChain<T> {
        /**
         * @description Be careful: This function overwrites the whole database.
         */
        write(): T & Promise<T>;
    }
    interface PrimitiveChain<T> {
        write(): T & Promise<T>;
    }
    interface CollectionChain<T> {
        write(): ArrayLike<T> & Promise<ArrayLike<T>>;
    }
    interface FunctionChain<T> {
        write(): T & Promise<T>;
    }
    interface StringChain {
        write(): string & Promise<string>;
    }
}
```

Somehow is not augmenting the __ObjectChain__ declaration, but overriding instead, resulting in missing method declarations when using Lowdb:

image

The code that produces the error:

```js
const db = low(new FileSync("db.json"));
db.forEach((value, key) => {
// some code
})
```

When prompting for method autocomplete it suggest only __getState__, __readState__, __read__, __write__

I had to downgrade to v1.0.8 to make it work again.

sandersn:
@b4dnewz I can see all the completions (forEach, forEachRight, forIn, etc) in the Definitely Typed tests. It sounds like you don't have `@types/lodash` installed or that it's not being resolved correctly. Double-check your node_modules, then try importing lodash manually to see if it's available in your project.

b4dnewz:
I had both __@types/lodash__ and __@types/lowdb__ installed, but probably as you said lodash types wasn't resolved properly by my editor.

After deleting node_modules and reinstalling them It worked again, thanks.

Gerst20051:
I'm getting the following error starting with this version. Was `.values()` missed in this refactor?

```
file.ts:55:8 - error TS2551: Property 'values' does not exist on type 'FunctionChain<any> | ObjectChain<any> | PrimitiveChain<any> | StringChain | StringNullableChain | (CollectionChain<any> & FunctionChain<any> & ObjectChain<any> & PrimitiveChain<...> & StringChain) | CollectionChain<...> | ObjectChain<...>'. Did you mean 'value'?
  Property 'values' does not exist on type 'FunctionChain<any>'.

55       .values()
          ~~~~~~

  node_modules/@types/lodash/ts3.1/common/seq.d.ts:147:9
    147         value(): TValue;
                ~~~~~~~~~~~~~~~~
    'value' is declared here.
```

PR #35250, opened 2019-05-07
feat(chance): add option definitions for common methods

colbywhite:
Please fill in this template.

  - the documentation is not thorough on all the parameters that are possible. so source code was looked at
  - [chance.bool](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L130)
  - [chance.word](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L711)
  - [chance.string](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L386)
  - [chance.url](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L1460)
  - [chance.integer](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L267)
  - [chance.first](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L839)
  - [chance.last](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L862)
    - note that the [`lastNames` data map](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L2598) supports more nationalities than the one for `firstNames`
  - [chance.prefix/chance.name_prefix](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L1019)
  - [chance.suffix/chance.name_suffix](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L1097)
  - [chance.name](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L966)
    - note that the options for this one are passed into several others as is. this ends up making things a bit more complicated, but what is valid to `name` is the smallest subset of all the methods it passes the options to. the typing reflects that. i.e `last` takes in more nationalities than `first`, but `name` should only allow the nationalities that support `first`.
  - [chance.email](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L1377)
  - [chance.sentence](https://github.com/chancejs/chancejs/blob/1.0.18/chance.js#L652)
  - no major/minor version bump needed
  - not needed

This is a different attempt at #34855 from @Hakier. It does more methods while also using source code to inform the types since the documentation has holes in it. It also uses a [clever solution](https://stackoverflow.com/a/48244432) (i.e the `AtLeastOneKey` generic type) from @jcalz in order to prevent an empty object from being passed in as a parameter. while this is technically allowed by chancejs, it is a sign that the user is doing something incorrect and thus it is ideal for the typings to prevent that.

This also fixes an error I introduced in ##34523. The return types of `chance.locales` should've been a `string[]`, not a `string`.

### DIFF
--- a/types/chance/chance-tests.ts
+++ b/types/chance/chance-tests.ts
@@ -91,8 +91,8 @@ date = chance.date({max});
 const language: string = chance.locale();
 const region: string = chance.locale({region: true});
 
-const languages: string = chance.locales();
-const regions: string = chance.locales({region: true});
+const languages: string[] = chance.locales();
+const regions: string[] = chance.locales({region: true});
 
 let word: string = chance.word();
 word = chance.word({syllables: 10, length: 10});
--- a/types/chance/chance-tests.ts
+++ b/types/chance/chance-tests.ts
@@ -95,9 +95,10 @@ const languages: string[] = chance.locales();
 const regions: string[] = chance.locales({region: true});
 
 let word: string = chance.word();
-word = chance.word({syllables: 10, length: 10});
+word = chance.word({syllables: 10, length: 10, capitalize: true});
 word = chance.word({syllables: 10});
 word = chance.word({length: 10});
+word = chance.word({capitalize: true});
 
 let randomString: string = chance.string();
 randomString = chance.string({pool: 'abcdef', length: 10});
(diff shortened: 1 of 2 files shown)


jessetrinity:
@cbowdon @brikou @cafesanu

Any issues with this being merged? Thanks!

PR #35254, opened 2019-05-07
[@types/react] add template to IntrinsicElements

magic-akari:
Please fill in this template.

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -2593,6 +2593,7 @@ declare namespace React {
         summary: DetailedHTMLFactory<HTMLAttributes<HTMLElement>, HTMLElement>;
         sup: DetailedHTMLFactory<HTMLAttributes<HTMLElement>, HTMLElement>;
         table: DetailedHTMLFactory<TableHTMLAttributes<HTMLTableElement>, HTMLTableElement>;
+        template: DetailedHTMLFactory<HTMLAttributes<HTMLElement>, HTMLElement>;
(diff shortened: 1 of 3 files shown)


magic-akari:
> The job exceeded the maximum time limit for jobs, and has been terminated.
It looks like there is no error. But the CI runs out of time. 
What should I do?

types/react/index.d.ts
```diff
@@ -2593,6 +2593,7 @@ declare namespace React {
         summary: DetailedHTMLFactory<HTMLAttributes<HTMLElement>, HTMLElement>;
         sup: DetailedHTMLFactory<HTMLAttributes<HTMLElement>, HTMLElement>;
         table: DetailedHTMLFactory<TableHTMLAttributes<HTMLTableElement>, HTMLTableElement>;
+        template: DetailedHTMLFactory<HTMLAttributes<HTMLElement>, HTMLElement>;
```
eps1lon: Looks like [`templaate`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template) even implements [its own interface](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTemplateElement). Is this available in the dom lib?

eps1lon (commented):
Never used this element. Is it fully implemented in `react-dom`?

types/react/index.d.ts
magic-akari: > Looks like [`templaate`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/template) even implements [its own interface](https://developer.mozilla.org/en-US/docs/Web/API/HTMLTemplateElement). Is this available in the dom lib?

Yes. It's my fault to forget this. I'll correct it.

PR #35258, opened 2019-05-07
fix: PageSize to be of type string literal

iamstiil:
Updated PageSize from enum to string literals.


types/pdfmake/index.d.ts
```diff
@@ -23,56 +23,55 @@ declare module "pdfmake/build/pdfmake" {
...
+        | 'C6'
+        | 'C7'
+        | 'C8'
+        | 'C9'
+        | 'C10'
+        | 'RA1'
+        | 'RA2'
+        | 'RA3'
+        | 'RA4'
+        | 'SRA1
```
evolkmann: You missed a `'` here

evolkmann (changes requested):
Hi @iamstiil, thanks for submitting the PR. I am fine with the change, but what was the reason behind it?

types/pdfmake/index.d.ts
iamstiil: The reason is that the string values are not accessible via `pdfmake.PageSize`. So using `pdfmake.PageSize.A4` resulted in valid linting but compile errors. Using a string instead throws linting errors but does compile.

dancrumb:
I feel like this change is a) potentially a breaking change (unless you're arguing that these would never work) and b) need a test that fails with the old typing and passes with the new typing.

types/pdfmake/index.d.ts
```diff
@@ -23,56 +23,55 @@ declare module "pdfmake/build/pdfmake" {
...
+        | 'SRA2'
+        | 'SRA3'
+        | 'SRA4'
+        | 'EXECUTIVE'
+        | 'FOLIO'
+        | 'LEGAL'
+        | 'LETTER'
+        | 'TABLOID';
 
     enum PageOrientation {
```
sheetalkamat: While at it, you probably want to convert PageOrientation to type as well

sheetalkamat: While at it, you probably want to convert PageOrientation to type as well

sheetalkamat (changes requested):

PR #35261, opened 2019-05-07
feat(types): add sylvester-es6

briwa:

I am adding a new definition:

### Some notes
I'm adding types for [`sylvester-es6`](https://github.com/pithumke/sylvester-es6) by @pithumke. There is already a type for [`sylvester`](https://www.npmjs.com/package/sylvester), however this is a type for the fork that was rewritten in ES6, which is what I am preferred.

Some interfaces are still the same, so I carried out some of the [previous types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/sylvester/index.d.ts) and removed some as well (that is why I added the previous contributor in this type definition). Adding more types since the fork has more features.

However, I encountered a problem when linting:
> ERROR: 2:1  npm-naming  At least one of the project urls listed in the header, ["https://github.com/pithumke/sylvester-es6"], must match the homepage listed by npm, 'https://github.com/pithumke/sylvester'.
If your d.ts file is not for the npm package with URL https://github.com/pithumke/sylvester,
change the name by adding -browser to the end and change the first line

Weird thing is the link is correct, but it keeps throwing errors. Also, opening this link (https://github.com/pithumke/sylvester) actually redirects to https://github.com/pithumke/sylvester-es6. Anyone knows why?

EDIT: For now, I'll point the project name to the original URL before renaming (https://github.com/pithumke/sylvester), thanks @pithumke for clarifying! This looks like a bug in `dslint` tho (CMIIW)...

Any help and reviews are greatly appreciated. Thanks.


pithumke:
@briwa The redirect happens because I renamed the repository from `pithumke/sylvester` to `pithumke/sylvester-es6` after forking it. GitHub kindly keeps the old repository link alive and redirects to the renamed repository. The error thrown by your linter seems to come from me not having adapted the package file accordingly (changing the URL of `homepage` and possibly even `bugs.url`). Please give me a nudge if you need this changed - I'll then push a new release to npm (might as well revert the "Putting curly braces on the next line where ever ES6 allows it." thing while I'm at it ... not really convinced anymore about why I even bothered doing that in the first place ...).

On a side note: In your comment you said "Adding more types since the fork has more features.". I assume you are referring to the addition of `glUtils` in my fork, since that's basically everything I added in terms of functionality. I'm not involved in "DefinitelyTyped" and I'm not entirely sure how works, but maybe you should consider having a separate TS type definition for `glUtils` (if you don't already have one) to avoid redundant definitions. However, please be aware that `glUtils` was basically just a small utility snippet used in [this](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API/Tutorial/Adding_2D_content_to_a_WebGL_context) MDN tutorial and has been removed from the tutorial by now, so I'm not sure if it would really be worth it...

briwa:
@pithumke Hey, thanks for replying! Yeah, you are right, I am referring to the `glUtils`. Since `sylvester-es6` npm package actually exports the Sylvester features including the `glUtils` functions that you did, I have no choice but to type them altogether, otherwise it is exported in the same package but not typed (which is not preferred). But correct me if I'm wrong.

Also, your fork has `Polygon` and `Vertex` as well which aren't in the original Sylvester. I think it is a good addition. But nonetheless, if you're thinking of removing the utility functions, let me know, I'll update the type definition afterwards.

I'm not too sure how the linting handles redirection and for some reason not reading the latest homepage. I am thinking to create a separate issue for this.

Anyway, thanks for your fork, man! Sylvester is a great library but really outdated, having an ES6 version should be the right way to go.

PR #35271, opened 2019-05-08
Lodash: `negate`: fix usage with rest parameters

OliverJAsh:
Please fill in this template.

### DIFF
--- a/types/lodash/lodash-tests.ts
+++ b/types/lodash/lodash-tests.ts
@@ -3519,6 +3519,17 @@ fp.now(); // $ExpectType number
 
 // _.negate
 {
+    const fnWithRestParameters: (...args: number[]) => boolean = anything;
+    // Ideally these would capture all parameters, but it's not easily done
+    // without generic rest parameters, which requires TS 3.0, which these types
+    // don't currently support.
+    // Note: these tests should be tested without `strictFunctionTypes`, as
+    // overload behaviour differs in that case.
+    _.negate(fnWithRestParameters); // $ExpectType (a1: number) => boolean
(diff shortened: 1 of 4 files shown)


types/lodash/lodash-tests.ts
```diff
@@ -3519,6 +3519,17 @@ fp.now(); // $ExpectType number
...
+    // Ideally these would capture all parameters, but it's not easily done
+    // without generic rest parameters, which requires TS 3.0, which these types
+    // don't currently support.
+    // Note: these tests should be tested without `strictFunctionTypes`, as
+    // overload behaviour differs in that case.
+    _.negate(f); // $ExpectType (a1: number) => boolean
+    _(f).negate(); // $ExpectType LoDashImplicitWrapper<(a1: number) => boolean>
+    _.chain(f).negate(); // $ExpectType LoDashExplicitWrapper<(a1: number) => boolean>
+    fp.negate(f); // $ExpectType (a1: number) => boolean
+
```
OliverJAsh: Prior to the overload change made in this PR, these tests would fail when `strictFunctionTypes` was disabled.

The 0 parameter overload would be used instead.

Changing the order of the overloads helps, but it doesn't completely fix the problem.

IIUC, the only way to fix it properly is to add an overload with generic rest parameters:

``` ts
        negate<T extends any[]>(predicate: (...args: T) => any): (...args: T) => boolean;
```

However, this requires TS 3.0, which these types do not currently support.

aj-r (changes requested):
There is an open PR that is adding special lodash types for TS >= 3.0. Maybe it's best to wait until that is submitted.

OliverJAsh:
Sure. Let's track the bug here: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/35273

PR #35272, opened 2019-05-08
Avoid extends check for Memo/LazyExoticComponent

sandersn:
A subtype check involving ComponentPropsWithRef<T> is extremely expensive because it is guaranteed to be structural. This PR avoids the subtype check by using intersection instead, which is equivalent in this case.

This PR doesn't address the underlying issue that ComponentProps and PropsWithRef are extremely expensive, but the same technique might be more generally applicable for easy savings in the react ecosystem.

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -777,9 +777,9 @@ declare namespace React {
 
     // will show `Memo(${Component.displayName || Component.name})` in devtools by default,
     // but can be given its own specific name
-    interface MemoExoticComponent<T extends ComponentType<any>> extends NamedExoticComponent<ComponentPropsWithRef<T>> {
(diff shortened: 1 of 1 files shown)


sandersn:
*Really* rough performance difference from a single run of `tsc --diagnostics` is:

Measure | Before | After
---|----|----
Type count | 94,869 | 81,454
Memory | 168MB | 152MB
Time | 14 | 12

By order of stability -- time of course varies all over the place, but type count is stable between runs.

ferdaber (approved):

ferdaber:
Curious if you have any ideas to improve the efficiency of `ComponentProps`? Is it the generic type param constraint checking? Or the nested conditional type inference?

sandersn:
Pretty sure it's option 2, the nested conditionals w/`infer`. I don't have ideas on how to improve it yet, but I'll look at it again tomorrow. I just wanted to start with some easy band-aids.

PR #35278, opened 2019-05-08
[elliptic] Fixes up type definitions.

MicahZoltu:
Both `keyFromPrivate` and `keyFromPublic` work with Uint8Array.  `keyFromPublic` accepts anything with a length and an indexer, though I'm not sure how to correctly setup the types for that.  `keyFromPrivate` works with anything that `BN` works with as input, which includes `Uint8Array`s.

The only valid value for `enc` in `getPublic` and `getPrivate` is `hex`, so I narrowed the type on those to only support that instead of `string`.

The `compact` boolean on `getPublic` with an `enc` parameter set to anything other than `hex` will yield `[ this.getY().isEven() ? 0x02 : 0x03 ].concat(x)`.  I chose the string 'array' as an opinion here so at least future readers of code written to use that overload of `getPublic` can have an idea of what they are getting out when they pass something besides `'hex'` in.  (note: if nothing is a falsy value for `enc` is provided when `concat` is truthy, the function will behave the same as if no parameters were passed).

Please fill in this template.

### DIFF
--- a/types/elliptic/elliptic-tests.ts
+++ b/types/elliptic/elliptic-tests.ts
@@ -26,7 +26,7 @@ const y = pubPoint.getY();
 // 1) '04' + hex string of x + hex string of y; or
 // 2) object with two hex string properties (x and y); or
 // 3) object with two buffer properties (x and y)
-const pub = pubPoint.encode('hex');                                 // case 1
+const pub = pubPoint.encode('hex', true);                            // case 1
 const aPub = { x: x.toString('hex'), y: y.toString('hex') };         // case 2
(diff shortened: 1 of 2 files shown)


MicahZoltu:
Bleh, I guess I'll actually clone this report and make sure it all works against existing tests.

MicahZoltu:
Fixed the issues with the tests @danwbyrne and @Gilthoniel.  The library follows the "`hex` or anything" pattern in a number of places, and in the places I ran across I continued to use `array` as the alternative so when people use the library their code will be more readable, even though technically any truthy value will work.

danwbyrne (approved):

PR #35282, opened 2019-05-08
Fix types/mongoose definition

kaipenglu:
Fix func "isDeleted" name with "$" prefix added.


jackchess:
Here is doc. https://mongoosejs.com/docs/api.html#document_Document-$isDeleted

orblazer (approved):

PR #35283, opened 2019-05-08
[fs-extra] Streamline version with package version

ffflorian:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/jprichardson/node-fs-extra/blob/master/CHANGELOG.md

AFAIC there were no API changes between v5 and v7.

### DIFF
--- a/types/fs-extra/index.d.ts
+++ b/types/fs-extra/index.d.ts
@@ -1,11 +1,12 @@
-// Type definitions for fs-extra 5.1
+// Type definitions for fs-extra 7.0
 // Project: https://github.com/jprichardson/node-fs-extra
 // Definitions by: Alan Agius <https://github.com/alan-agius4>,
 //                 midknight41 <https://github.com/midknight41>,
 //                 Brendan Forster <https://github.com/shiftkey>,
 //                 Mees van Dijk <https://github.com/mees->,
-//                 Justin Rockwood <https://github.com/jrockwood>
-//                 Sang Dang <https://github.com/sangdth>
+//                 Justin Rockwood <https://github.com/jrockwood>,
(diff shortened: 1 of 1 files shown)


sandersn:
While we're waiting for the typings authors to show up, I have a couple of questions:

1. Is v5 likely to change? fs-extra is widely used, so I would guess so, except..
2. If v7 has no changes relative to v5, then why add it now, or add it at all?

Basically, it would be nice if we could just have 5 or 7 in the DT repo and not both. But fs-extra is so popular that both might well be needed.

shiftkey:
@sandersn apologies for the delay - I wasn't familiar with the way these major versions are supported to know I should weigh in.

> Is v5 likely to change? fs-extra is widely used, so I would guess so, except..

Looking at the [CHANGELOG](https://github.com/jprichardson/node-fs-extra/blob/master/CHANGELOG.md) for `fs-extra`, you'll see that it's rolling forward and generously bumping the major version whenever any sort of breaking change is introduced. So I don't expect to see a bump in `v5` that would necessitate maintaining a separate typings file for that version.

> If v7 has no changes relative to v5, then why add it now, or add it at all?

This is the part that's puzzling me, and maybe @ffflorian has some extra context here that I don't.

I'd recommend just updating the main version to align better with the `fs-extra` version, rather than maintaining v5 and v7 separately.

ffflorian:
> I'd recommend just updating the main version to align better with the `fs-extra` version, rather than maintaining v5 and v7 separately.

Sure, good idea. I just wasn't sure how to handle this properly. I'll update my PR.

shiftkey (approved):

PR #35285, opened 2019-05-08
[paper] Include Base class definition + type save getters and setters

anion155:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<[paper.js/Base.js](https://github.com/paperjs/paper.js/blob/develop/src/core/Base.js)>>

### DIFF
--- a/types/paper/index.d.ts
+++ b/types/paper/index.d.ts
@@ -1152,32 +1152,32 @@ declare module paper {
         modulo(size: number[]): Size;
         modulo(size: number): Size;
 
-		/**
-		 * Sets the size with the given width and height values.
+        /**
+         * Sets the size with the given width and height values.
          * @param width - the width
          * @param height - the height
-		 */
-		set(width: number, height: number): Size;
+         */
+        set(width: number, height: number): Size;
 
-		/**
-		 * Sets the size using the numbers in the given array as dimensions.
+        /**
+         * Sets the size using the numbers in the given array as dimensions.
          * @param array - an array of numbers
-		 */
-		set(array: number[]): Size;
+         */
+        set(array: number[]): Size;
 
-		/**
-		 * Sets the size using the properties in the given object.
+        /**
+         * Sets the size using the properties in the given object.
          * @param object - the object literal containing properies (width:10, height:10 etc)
-		 */
-		set(object: any): Size;
+         */
+        set(object: any): Size;
 
-		/**
-		 * Sets the size using the coordinates of the given Size object.
+        /**
+         * Sets the size using the coordinates of the given Size object.
          * @param size - the size to duplicate from
          */
-		set(size: Size): Size;
+        set(size: Size): Size;
 
-		/**
+        /**
          * Sets the size using the point.x and point.y values of the given Point object.
          * @param point - the point from which to create a size
          */
@@ -5495,7 +5495,7 @@ declare module paper {
 
     }
 
-	export interface Settings {
+    export interface Settings {
         /**
          * controls whether newly created items are automatically inserted into the scene graph, by adding them to project.activeLayer — default: true
          */
--- a/types/paper/index.d.ts
+++ b/types/paper/index.d.ts
@@ -31,7 +31,8 @@ declare module paper {
          * classes match. If they do not match, a newly created object is returned
          * instead.
          */
-        static importJSON<T = any>(json: string, target?: T): T;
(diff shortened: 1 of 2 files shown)


anion155:
@luz-alphacode So, maybe `color: Color; setColor(c: Color | string): void;` everywhere?

About symbols. Did not looked into this, yet. Maybe later.

luz-alphacode:
@anion155 Totally agreed. About symbol and other stuffs, they works well, just a bit confusing if you check with the document. And by the way, a getTimesWithTangent(tangent) for **Curve** class is updated lately and is missing in the definition file. Maybe you can add it as well.
```typescript
/**
 * Calculates the curve-time parameters where the curve is tangential to provided tangent. Note that tangents at the start or end are included.
 * @param tangent - the tangent to which the curve must be tangential
 * @returns at most two curve-time parameters, where the curve is tangential to the given tangent
 */
  getTimesWithTangent(tangent: Point): number[];
```

luz-alphacode:
@rfried Oh, that's a hard work! It's true that the paper.js documentation sometimes mismatch the actual code implementation. Months ago I've tried find a way to balance this but I gave up at last, now I know the reason why they are different. It's a really excellent work because once I get to know the differences, it's quiet easy to use. Thank you for what you've done for this.
Sometimes the design pattern of a js library and a ts library is so different, making a completely typescript-comforting definition file is just impossible.

rfried:
@luz-alphacode 
Thanks :) I'm not going to claim I got it all right (see my questionable use of the code SymbolItem instead of the documented PlacedSymbol) and I'm glad to see more nuance added to the definition for sure. I feel like much of the cases, such as always using a Color, are mitigated by the fact that the constructors are pretty flexible. For instance, Color's constructor takes a css string. In the end the code for setting a color is more verbose than color = 'red' but isn't terrible as color = Color('red'). That said it's a neat use of get and set as a possible improvement to that flow.

My biggest disappointment was the inability to create operator overloading.
    Point p;
    p.add(new Point(100, 100))
is less clean than the paperscript version:
     p+Point(100, 100)

The code is also pretty fluid in mixing classes and types too (which is some code-trickery that was pretty hard for me to follow in the original source).
    var point = new Point(5, 10);
    var result = point + 20; <--- this is nice and how could it be represented nicely in Typescript?
    console.log(result); // {x: 25, y: 30}

There are a lot of union types in the definition as a result of this :-p

luz-alphacode:
@rfried The typescript team claimed that they will not implement operator overloading since the typescript should keep consistency to javascript. I'm quite disagree with that, anyway, typescript is not runtime, why not add more sugars. Hoping one day they will change their mind.

anion155:
@rfried about Style. It is a class. All fields in it always presents. To do the update some props of it's, like `this.style = { prop: value }` in TS, it has to be `this.setStyle({ prop: value })`. The way you did is not TS way of doing things.

rfried:
@anion155 
The library under the typescript is super flexible and opting into typescript adds constraints. It's debatable how restricted this library should be to conform to "the TS way". Maybe Styles should have been an interface instead. 

The problem is that if you create a Style and set that style on another object then you have to make sure all the fields in the class are the way they should be. It's easy in the case where you take a style and modify a field and set it again. It's hard when you set a style from scratch and it breaks the ease-of-use that I was reading from the library's flexibility.

anion155:
@rfried There is no `set a style from scratch` in paper's objects. Almost every setter is updater. For style it is `setStyle(style) { this._style.set(style); }`, same for property descriptor `style`. Function `set` in paper's object only updates internal fields.

rfried:
@anion155 
I missed your use of Partial<Style> on the set. This does make the interface nicer as you said and allows for partial sets in the API `this.setStyle({ prop: value })` . 

can I still do this?
```
let gTextStyle: paper.Style = {
  fontFamily: "sans-serif",
  justification: "center",
  strokeColor: 'red',
  fillColor: 'green'
} as paper.Style;
```

anion155:
@rfried It will work. But your `gTextStyle` will not be `paper.Style` instance. For this case, you will probably need `new paper.Style({ fontFamily: '' })` or `gTextStyle: Partial<Style> = {}` if you do not want to store it like style instance. I'm going to do some more changes, like making some interfaces instead things like `Partial<Style>`. But a bit later

rfried:
@anion155 
nice, thanks

PR #35308, opened 2019-05-09
fix(node-pushnotifications): add return messages

jaulz:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/appfeel/node-pushnotifications/blob/master/src/push-notifications.js


endel:
Hey @jaulz, I believe the `Message.regId` shouldn't be a string. I've just upgraded my node modules and I'm having this compilation error:

```typescript
error TS2339: Property 'keys' does not exist on type 'string'.

85                     return message.regId.keys.p256dh
                                            ~~~~
```

I'm not using this extensively atm tho, so not sure if something has changed on node-pushnotifications itself, but I believe it hasn't. 

I use `regId.keys.p256dh` during checking for delivery failures:

```typescript
const results = await push.send(registrationIds, data);
results[0].message
  .filter(message => message.error)
  .map(message => {
    console.error("PUSH ERROR: " + message.error)
    return message.regId.keys.p256dh;
  });
```

Could it be that `Message.regId` is mistyped? Cheers

PR #35317, opened 2019-05-09
Add types for react-pdf

CodeDaraW:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/react-pdf/dist/Page.d.ts
+++ b/types/react-pdf/dist/Page.d.ts
@@ -2,16 +2,14 @@ import * as React from 'react';
 
 export type RenderFunction = () => JSX.Element;
 
-/* tslint:disable:ban-types */
 export interface PDFPageItem {
     _transport: object;
     commonObjs: object;
-    getAnnotations: Function;
-    getTextContent: Function;
-    getViewport: Function;
-    render: Function;
+    getAnnotations: (...args: any[]) => any;
+    getTextContent: (...args: any[]) => any;
+    getViewport: (...args: any[]) => any;
+    render: (...args: any[]) => any;
 }
-/* tslint:enable:ban-types */
 
 export interface TextLayerItemInternal {
     fontName: string;
--- a/types/react-pdf/dist/Page.d.ts
+++ b/types/react-pdf/dist/Page.d.ts
@@ -1,0 +1,191 @@
+import * as React from 'react';
+
+export type RenderFunction = () => JSX.Element;
+
+/* tslint:disable:ban-types */
+export interface PDFPageItem {
+    _transport: object;
+    commonObjs: object;
+    getAnnotations: Function;
+    getTextContent: Function;
+    getViewport: Function;
+    render: Function;
+}
+/* tslint:enable:ban-types */
+
+export interface TextLayerItemInternal {
+    fontName: string;
+    itemIndex: number;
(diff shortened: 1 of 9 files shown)


types/react-pdf/tsconfig.json
```diff
@@ -0,0 +1,30 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6",
+            "dom"
+        ],
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "strictNullChecks": false,
```
sandersn: Please set `strictNullChecks: true`. Many consumers of this package will have it on.

types/react-pdf/dist/Page.d.ts
```diff
@@ -0,0 +1,191 @@
+import * as React from 'react';
+
+export type RenderFunction = () => JSX.Element;
+
+/* tslint:disable:ban-types */
+export interface PDFPageItem {
+    _transport: object;
+    commonObjs: object;
+    getAnnotations: Function;
```
sandersn: you might consider `(...args: any[]) => any` instead of `Function` if you haven't already. It's almost the same, but doesn't accept `Function`, which might be desired.

sandersn (changes requested):
Need to fix tsconfig.json

CodeDaraW:
Hi, I have fixed it, please review the changes again.

sandersn (approved):

jmchn:
@CodeDaraW Is this definition using the webwoker though?

CodeDaraW:
@jmchn No. In my opinion, react-pdf doesn't care about the pdfjs worker. It's all about pdfjs, not react-pdf.

jmchn:
@CodeDaraW 
I am not sure I understand your answer. react-pdf provides a way to enable web worker though. In practice, it's better enable react-pdf with web worker because the perf would be really bad if the pdf is long. 
https://github.com/wojtekmaj/react-pdf#enable-pdfjs-worker

CodeDaraW:
@jmchn `react-pdf` is just a React component library, `@types/react-pdf` is just type definition library for React component. So, `@types/react-pdf` doesn't provides type definition for pdfjs worker.  

In the document, you can import `pdfjs` from `react-pdf`, but in fact `react-pdf` just export `pdfjs` from `pdfjs-dist`. So you can import `pdfjs` from `pdfjs-dist` directly and use `@types/pdfjs-dist`.

See https://github.com/wojtekmaj/react-pdf/blob/b2123e6ef3f7165656fa28f57bd2b05da4f67087/src/entry.js#L1

jmchn:
@CodeDaraW  I see what you mean.. I guess the way I tried to make it work with web worker is using react-pdf by declaring module 'react-pdf/dist/entry.webpack' and that will use the built-in entry.webpack file to create a web worker for me. I haven't spent times investigating how to create a web worker using this change/type definition.

CodeDaraW:
@jmchn Declaring module 'react-pdf/dist/entry.webpack' is a solution too.

jmchn:
@CodeDaraW  Another thing might worth noting is that
instead doing
`export default class Outline extends React.Component<Props> { }`
but do 
`export default class Outline extends React.PureComponent<Props> { }`?

Issue #34959, opened 2019-04-24
[@types/react-select] Anything can be passed to Select


szymkab:
If you know how to fix the issue, make a pull request instead.

- [x] I tried using the `@types/react-select` package and had problems.
- [x] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @claasahl @jonfreedman

Hi,

Because of line 54 at `Select.d.ts` anything can be passed to select and typescript will not show any errors.

`export type SelectComponentsProps = { [key in string]: any };`

What is the purpose of that `{ [key in string]: any };` in Select types?

artgladun:
Together with my team, we are trying to move from react-select v1 to v2; but with such `any` props it is very hard to work with and check everything manually instead to just trust linter... especially there are small differences like disabled -> isDisabled

szymkab:
The thing is that if interface extends `{ [key in string]: any };` you can pass literally everything to that component. It somehow solves an issue with having custom props in select components but it has a lot of disadvantages, e.g. you can pass everywhere  (even if you make typo typescript will not inform you) and intelisense is not working.

So, I've prepared a solution with a generic argument that can be used for custom props of select components (and is by default set to `{ [key in string]: any };`).

If you don't pass second generic argument you can have `any` props in custom component just like now:

![Zrzut ekranu 2019-04-26 o 16 17 39](https://user-images.githubusercontent.com/13819284/56814244-09505a00-683f-11e9-9e8d-b5c361e99f2f.png)

If you pass an interface as second argument you can have type on that custom prop:

![Zrzut ekranu 2019-04-26 o 16 18 27](https://user-images.githubusercontent.com/13819284/56814270-1705df80-683f-11e9-9356-34e31460f6f0.png)

agwells:
This one caused problems for me too, because I've made some wrapper components around React-Select, that override some props, and pass other props through. So I have a props definition that did that by using some helper types based on `Pick` and `Extract`. It looks something like this:

```typescript
import { Props as ReactSelectProps } from 'react-select/lib/Select';

// Omit: Create a type that has all the keys of T except those named in K
[...]

PR #35318, opened 2019-05-09
[react-select] Add generic argument for custom select components props

szymkab:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34959

@claasahl - new pull request

### DIFF
--- a/types/react-select/lib/Async.d.ts
+++ b/types/react-select/lib/Async.d.ts
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import Select, { Props as SelectProps } from './Select';
+import Select, { Props as SelectProps, DefaultSelectComponentsProps } from './Select';
 import { handleInputChange } from './utils';
 import manageState from './stateManager';
 import { OptionsType, InputActionMeta } from './types';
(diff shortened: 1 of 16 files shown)


claasahl (approved):
Thanks for your contribution, @szymkab

sandersn:
This solution is really expensive and may result in users of react-select in running out of memory, particularly if they use a lot of other react packages. Is there some other way to accomplish what you want?

Before this change, react-select takes 180MB of memory to compile, and generates 91,335 types. 
After this changes, it generates 105,571 types. Types are a decent proxy for both time and memory usage, so that means it is about 15% more expensive.

That is with typescript@next. The numbers about 3% worse with 3.4.

sandersn (changes requested):
Need to find a more efficient way to restrict usage.

szymkab:
@sandersn There is no other way to accomplish what I've done. That generic argument was missing from the beginning and issue was solved by adding `{ [key: string]: any }` but that's a workaround and nothing more.

szymkab:
@sandersn bump

sandersn:
Inside types/react-select, run `tsc --diagnostics`. Compare resource usage between master and your branch. Play around with variations to see if you can find a cheaper way to fix the problem. Memory usage doesn't vary much between compiles, and type count correlates well with both memory usage and compile time, and is stable between compiles. Those are the two measures I would focus on.

You might not be able to make this improvement to the types, unfortunately. We *really* need commonly used React components to avoid high memory usage since everybody in VS Code depends on them, and it increases batch compile times as well.

PR #35331, opened 2019-05-09
fix(stripe): Fix application_fee on IChargeCreationOptions

mdlavin:
It looks like the `application_fee` attribute might have been copied from the Charge object instead of from the create Charge documentation. The actual attribute name is `application_fee_amount` and the relevant documentation is https://stripe.com/docs/api/charges/create#create_charge-application_fee_amount

### DIFF
--- a/types/stripe/index.d.ts
+++ b/types/stripe/index.d.ts
@@ -1191,15 +1191,13 @@ declare namespace Stripe {
             currency: string;
 
             /**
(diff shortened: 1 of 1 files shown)


viralpickaxe (approved):
Thanks for this! Verified in stripe documentation https://stripe.com/docs/api/charges/create#create_charge-application_fee_amount

PR #35333, opened 2019-05-10
[node-fetch] Add signal to Request and RequestInit

hjr3:
Please fill in this template.

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/bitinn/node-fetch/blob/master/CHANGELOG.md#v230

### DIFF
--- a/types/node-fetch/index.d.ts
+++ b/types/node-fetch/index.d.ts
@@ -12,6 +12,7 @@
 
 import { Agent } from "http";
 import { URLSearchParams } from "url";
+import { AbortSignal } from 'abort-controller';
 
 export class Request extends Body {
     constructor(input: string | { href: string } | Request, init?: RequestInit);
@@ -22,6 +23,7 @@ export class Request extends Body {
     redirect: RequestRedirect;
     referrer: string;
     url: string;
+    signal?: AbortSignal;
 
     // node-fetch extensions to the whatwg/fetch spec
     agent?: Agent;
@@ -41,6 +43,7 @@ export interface RequestInit {
     headers?: HeadersInit;
     method?: string;
     redirect?: RequestRedirect;
+    signal?: AbortSignal;
 
     // node-fetch extensions
     agent?: Agent; // =null http.Agent instance, allows custom proxy, certificate etc.
--- a/types/node-fetch/index.d.ts
+++ b/types/node-fetch/index.d.ts
@@ -12,7 +12,7 @@
 
 import { Agent } from "http";
 import { URLSearchParams } from "url";
-import { AbortSignal } from 'abort-controller';
+import { AbortSignal } from "abort-controller";
 
 export class Request extends Body {
     constructor(input: string | { href: string } | Request, init?: RequestInit);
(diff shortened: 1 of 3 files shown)


hjr3:
This PR will fail testing. It is not obvious to me how to add `abort-controller` as a dependency. Or if that is not allowed, how to properly define `signal`.

hjr3:
ping @torstenwerner @nikcorg @vinaybedre @kyranet @AndrewLeedham @JasonLi914

kyranet:
You can try this: 
image

Source: https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package

hjr3:
Thanks @kyranet for pointing me in the right direction. Creating a `package.json` inside `node-fetch` seems to work. It is blocked by `abort-controller` not being in the whitelist.

I opened https://github.com/microsoft/types-publisher/pull/612 and will resume once that is merged.

types/node-fetch/index.d.ts
```diff
@@ -12,6 +12,7 @@
 
 import { Agent } from "http";
 import { URLSearchParams } from "url";
+import { AbortSignal } from 'abort-controller';
```
kyranet: Just a nit, seeing that the other imports use double quotes:
```suggestion
import { AbortSignal } from "abort-controller";
```

kyranet (changes requested):
Now seeing that the pending PR to whitelist this package has merged, let's trigger the build with a new commit!

hjr3:
@kyranet I was waiting for the change to land in the _production_ branch. Looks like that happened. There looks to be some issue with typescript 2.8 finding the type definitions for `abort-controller`.

```
Error: /Users/herman/Code/DefinitelyTyped/types/node-fetch/index.d.ts:15:29
ERROR: 15:29  expect  Compile error in typescript@2.8 but not in typescript@2.9.
Fix with a comment '// TypeScript Version: 2.9' just under the header.
Could not find a declaration file for module 'abort-controller'. '/Users/herman/Code/DefinitelyTyped/types/node-fetch/node_modules/abort-controller/dist/abort-controller.js' implicitly has an 'any' type.
  Try `npm install @types/abort-controller` if it exists or add a new declaration (.d.ts) file containing `declare module 'abort-controller';`
```

`tsc types/node-fetch/node-fetch-tests.ts` does not throw the same error.

I added `// TypeScript Version: 2.9` to the header, but that starts down a rabbit hole of changing the typescript version for`types/node-fetch`, `types/form-data`, `types/got`, `types/got/v8` and I finally hit a blocker with:

```
Error: download depends on got v8 but has a lower required TypeScript version.
```

nikcorg:
How about using an interface instead of using a dependency?

nikcorg:
The `node-fetch` package depends on `"abort-controller": "^1.1.0",` while this definition depends on `"abort-controller": "^3.0.0"`. Is that significant? The earlier version doesn't have type definitions, though.

hjr3:
@nikcorg `abort-controller` is a dev dependency https://github.com/bitinn/node-fetch/blob/master/package.json#L41

Reading the node-fetch source, there is no actual dependency on that package. https://github.com/bitinn/node-fetch/blob/bf8b4e8db350ec76dbb9236620f774fcc21b8c12/src/request.js#L36-L43

PR #35340, opened 2019-05-10
Add additional optionHeight prop type. (any) => number

JonnyPickard:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/bvaughn/react-virtualized-select#react-virtualized-select-props>>

### DIFF
--- a/types/react-virtualized-select/index.d.ts
+++ b/types/react-virtualized-select/index.d.ts
@@ -1,6 +1,7 @@
 // Type definitions for react-virtualized-select 3.0
 // Project: https://github.com/bvaughn/react-virtualized-select
 // Definitions by: Sean Kelley <https://github.com/seansfkelley>
+//               : Jonny Pickard <https://github.com/jonnypickard>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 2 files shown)


Steve-Fenton (approved):
Thank you for this pull request, especially as it contains a change to the tests.

I've checked the change against the documentation and it's all good.

types/react-virtualized-select/index.d.ts
```diff
@@ -22,9 +23,13 @@ export interface VirtualizedOptionRenderOptions<T> {
...
 }
 
+interface OptionHeightFunc {
+    (props: any): number;
+  }
+
 export interface AdditionalVirtualizedSelectProps<TValue> {
     maxHeight?: number;
-    optionHeight?: number;
+    optionHeight?: number | OptionHeightFunc;
```
seansfkelley: Per [the docs](https://github.com/bvaughn/react-virtualized-select#react-virtualized-select-props), it looks like this signature can be a bit more precise. I think it can be typed `({ option: TValue }) => number`, or at least `({ option: any }) => number`.

Also, could you just inline the definition here? The interface adds a lot of indirection.

seansfkelley (changes requested):

types/react-virtualized-select/index.d.ts
Steve-Fenton: That does look like the signature of the interface specified here. The body of OptionHeightFunc is:

        (props: any): number;

JonnyPickard: I think I purposefully chose not to do `({ option: TValue }) => number` as it limits the function to having to take an object of that shape as args. In our case, we were wanting to pass in other arguments which I think makes sense as it's our callback function we should control what we do with it as long as it returns a number. I won't be able to check my code to confirm until Monday now though as it's on my work machine.

seansfkelley: > Are you suggesting props is tightened up with { option: any} in that interface?

Yes, and also that the interface be removed because being a simple non-overloaded one-argument non-generic function, putting it in an interface seems like extra indirection for no value.

> it limits the function to having to take an object of that shape as args.

Your function can choose not to take any arguments, but if it takes one, it has to be `{ option: TValue }` as far as I can tell from the docs. How would you pass in different arguments if react-virtualized-select is the one calling your function?

PR #35343, opened 2019-05-10
Knockout.mapping jsdocs, tests and typing overhall

ltlombardi:
This took a lot of time. Knockout Mapping has a API with too much overload. It's methods accept all kind of inputs..
Changes:
- small fixes in knockout
- fix type for the mapping options.
- reduce number of method overload, better mapping all the method calling options with optional parameters.
- complete overhaul of typings
- add JSDoc to most of the methods.
- add Undocumented when method doesn't have documentation.

Reasoning:
The idea behind `type KnockoutObservableType<T>` is good, but doesn't work as intended. This will make object properties `T | KnockoutObservable<T>`, and you will have to put the correct type in each property to use them, what defeats the purpose of this typing. 
For example: 
```
let userInputJS: User = { firstName: 'foo', lastName: 'bar', age: 12 };
interface User {
    firstName: string;
    lastName: string;
    age: number;
}
let mappedUser : KnockoutObservableType<User> = mapping.fromJS(userInputJS); 
mappedUser.age(43); // gives error, because can be Number, not just KnockoutObservable<number>
(mappedUser.age as KnockoutObservable<number>)(43); // have to do something like this...
```

Besides, this gives `RangeError: Maximum call stack size exceeded` in TSC in the examples I used in the tests file, when used in  `fromJS<T>()`. It's seems to me it's a recursive type that doesn't have a end clause... Maybe this can be used with TS 2.8 and conditional typing.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://knockoutjs.com/documentation/plugins-mapping.html>>

### DIFF
--- a/types/knockout.mapping/index.d.ts
+++ b/types/knockout.mapping/index.d.ts
@@ -2,6 +2,7 @@
 // Project: https://github.com/SteveSanderson/knockout.mapping
 // Definitions by: Boris Yankov <https://github.com/borisyankov>, 
 //                 Mathias Lorenzen <https://github.com/ffMathy>
+//                 Leonardo Lombardi <https://github.com/ltlombardi>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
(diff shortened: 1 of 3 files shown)


types/knockout.mapping/index.d.ts
```diff
@@ -12,11 +13,36 @@ export as namespace mapping;
 declare var self: KnockoutMapping;
 export = self;
 
+
 declare global {
-    type KnockoutObservableType<T> = {	
-        [P in keyof T]: KnockoutObservable<KnockoutObservableType<T[P]>>|T[P];	
+    // the idea behind this is good, but doesn't work as intended. This will make object properties T | Observable<T>,
```
Igorbek: Agree. It must be possible to use conditional types here. Do you want to try them?

Igorbek (approved):
I have no much expertise in KO mapping, but it looks good to me.

PR #10831, opened 2016-08-25
keep bluebird promise type-compatible with the lib.es6.d.ts promise

austinried:
This change keeps the Bluebird promise type-compatible with the ES6 promise definition found here: https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es6.d.ts#L5505

I ran into this issue with WebStorm's version of those ES6 typings, but without the Symbol.toStringTag you get an error about the property being missing on Bluebird's promise interface when trying to mix the two.  Once it's added the issue is resolved.

PR #34805, opened 2019-04-17
[bluebird] Restore assignability to native Promises (fixes #11027)

lhecker:
## Intention

Recently Bluebird added support for `Symbol.toStringTag` in https://github.com/petkaantonov/bluebird/pull/1421.
This PR adds typings for this new "feature", in turn restores assignability of Bluebird to native Promise types and thus fixes the longstanding issue #11027.

This Bluebird feature has _not_ been released yet, but I expect this to happen relatively soon.
We'll have to decide whether we wait or not for Bluebird to release the changes in the above PR.
I've inquired @petkaantonov whether we intends to release a minor version for this [here](https://github.com/petkaantonov/bluebird/pull/1421#issuecomment-484134806).
In that case we could safely release this PR as part of seperate Bluebird 3.6 typings.
If it'll be a patch release we can merge this PR at any point in time instead.

## Code

The PR is split up in 2 commits. You should review them seperately.
The first commit only renames the Bluebird import for tests, in order for us to more easily have access to the native `Promise` class (including in the future).

cc: @tmueller

## Template

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [ ] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/petkaantonov/bluebird/pull/1421
- [x] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #35353, opened 2019-05-10
Adding [Symbol.toStringTag] to Bluebird

dancrumb:
This is a rehash of #10831, which was closed because Bluebird did not have this Symbol available back then.

Please fill in this template.

- Follows on from https://github.com/petkaantonov/bluebird/issues/1277

### DIFF
--- a/types/bluebird/index.d.ts
+++ b/types/bluebird/index.d.ts
@@ -42,7 +42,7 @@ type IterateFunction<T, R> = (item: T, index: number, arrayLength: number) => Re
 
 declare class Bluebird<R> implements PromiseLike<R>, Bluebird.Inspection<R> {
   [Symbol.toStringTag]: "Promise";
-  
+
   /**
    * Create a new promise. The passed in function will receive functions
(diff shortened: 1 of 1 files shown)


lhecker:
Unfortunately this PR is a bit incomplete and has a number of issues (I'm sorry):
* Test cases for this new functionality are missing
* The `Symbol.toStringTag` field on Bluebird promises actually returns "Object" instead of "Promise"
* The actual signature of the property is `readonly` - see [here](https://github.com/microsoft/TypeScript/blob/ae3d1d45c1f98b103adbbafb03601205a981ee9f/lib/lib.es2015.symbol.wellknown.d.ts#L169)
* This in turn requires a TypeScript version upgrade of Bluebird's typings and all of its transitive dependents
* Bluebird wasn't actually released yet with this feature unfortunately - see [here](https://github.com/petkaantonov/bluebird/pull/1421#issuecomment-484822157)

I humbly suggest to close this PR in favor of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34805.

dancrumb:
Gladly!

PR #35354, opened 2019-05-10
Fix colornames default export type definition

blturner:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/timoxley/colornames/blob/1.1.1/index.js#L24

The exported default function does not return a `Color` object, it returns the hexadecimal string value of the resolved color name. The type definition assumes it is a `Color` object.

I encountered this error when attempting to use the library in an existing typescript project using `import`.

In testing this change to the type definition, I did notice that using `require` did not cause any typescript errors:
```
const getHex = require("colornames");
```

But importing the library this way did:
```
import * as getHex from "colornames";
```

The changes to the type definition in this PR work with both of the above examples.

### DIFF
--- a/types/colornames/colornames-tests.ts
+++ b/types/colornames/colornames-tests.ts
@@ -1,13 +1,9 @@
 import colorNames = require("colornames");
 
 // $ExpectType string | undefined
-const color = colorNames("red");
+colorNames("red");
 // $ExpectType string | undefined
-color;
-// $ExpectType string | undefined
-const undefinedColor = colorNames("donkey");
-// $ExpectType string | undefined
-undefinedColor;
+colorNames("donkey");
 // $ExpectType Color
 const blue = colorNames.get("blue");
 // $ExpectType boolean | undefined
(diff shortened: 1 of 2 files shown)


jarofghosts (approved):

types/colornames/colornames-tests.ts
```diff
@@ -1,19 +1,21 @@
...
 const color = colorNames("red");
-
-// $ExpectType string
-color.value;
+// $ExpectType string | undefined
+color;
+// $ExpectType string | undefined
+const undefinedColor = colorNames("donkey");
+// $ExpectType string | undefined
+undefinedColor;
```
manuth: You may want to replace this portion of code by
```ts
// $ExpectType string | undefined
colorNames("red");
// $ExpectType string | undefined
colorNames("donkey");
```

As checking the variables a second time is not necessary.

neurolag (changes requested):
This PR looks perfect to me.
I just compared it to the API of `colornames` and it just matches perfectly.

Please consider to have a look at the comment I wrote during this review.

types/colornames/colornames-tests.ts
blturner: @manuth thanks for the feedback!

neurolag (approved):
Looks all fine to me.
Thanks for your contribution!

PR #35356, opened 2019-05-10
[@types/algoliasearch] Update type for hits to include objectID

KSaryan:
https://www.algolia.com/doc/api-client/methods/indexing/?language=javascript#the-object-id

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.algolia.com/doc/api-client/methods/indexing/?language=javascript#the-object-id

### DIFF
--- a/types/algoliasearch/index.d.ts
+++ b/types/algoliasearch/index.d.ts
@@ -981,7 +981,7 @@ declare namespace algoliasearch {
   }
   interface BrowseResponse {
     cursor?: string;
-    hits: {}[];
+    hits: {objectID: string}[];
     params: string;
     query: string;
     processingTimeMS: number;
--- a/types/algoliasearch/index.d.ts
+++ b/types/algoliasearch/index.d.ts
@@ -981,7 +981,7 @@ declare namespace algoliasearch {
   }
   interface BrowseResponse {
     cursor?: string;
-    hits: {objectID: string}[];
+    hits: {objectID: string, [attribute: string]?: unknown}[];
     params: string;
     query: string;
     processingTimeMS: number;
--- a/types/algoliasearch/index.d.ts
+++ b/types/algoliasearch/index.d.ts
@@ -981,7 +981,7 @@ declare namespace algoliasearch {
   }
   interface BrowseResponse {
     cursor?: string;
-    hits: {objectID: string; [attribute: string]?: unknown;}[];
+    hits: {objectID: string; [attribute: string]: unknown}[];
     params: string;
     query: string;
     processingTimeMS: number;
--- a/types/algoliasearch/index.d.ts
+++ b/types/algoliasearch/index.d.ts
@@ -9,7 +9,7 @@
 //                 Antoine Rousseau <https://github.com/antoinerousseau>
 //                 Luca Pasquale <https://github.com/lucapasquale>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.0
 
 type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>
 
--- a/types/algoliasearch/index.d.ts
+++ b/types/algoliasearch/index.d.ts
@@ -981,7 +981,7 @@ declare namespace algoliasearch {
   }
   interface BrowseResponse {
     cursor?: string;
-    hits: {objectID: string, [attribute: string]?: unknown}[];
+    hits: {objectID: string; [attribute: string]?: unknown}[];
     params: string;
     query: string;
     processingTimeMS: number;
--- a/types/algoliasearch/index.d.ts
+++ b/types/algoliasearch/index.d.ts
@@ -981,7 +981,7 @@ declare namespace algoliasearch {
   }
   interface BrowseResponse {
     cursor?: string;
-    hits: {objectID: string; [attribute: string]?: unknown}[];
+    hits: {objectID: string; [attribute: string]?: unknown;}[];
     params: string;
     query: string;
     processingTimeMS: number;


squirly (approved):

types/algoliasearch/index.d.ts
```diff
@@ -981,7 +981,7 @@ declare namespace algoliasearch {
   }
   interface BrowseResponse {
     cursor?: string;
-    hits: {}[];
+    hits: {objectID: string}[];
```
neryortez: That would cause errors when someone tried to access other properties. Typescript will complain that the type of `hit` does not contain more properties than `objectID`.

I think that leaving it like before is great. But if you think that having `objectID` directly mentioned, the following suggestion will make sense.
```suggestion
    hits: {objectID: string, [key: string]: any}[];
```

neryortez (changes requested):

types/algoliasearch/index.d.ts
squirly: > That would cause errors when someone tried to access other properties. Typescript will complain that the type of hit does not contain more properties than objectID.

That is currently the case for all attributes on objects of type `{}`, as in this typing.
See https://www.typescriptlang.org/play/#src=let%20hit%3A%20%7B%7D%3B%0D%0Alet%20hit2%3A%20%7B%20objectID%3A%20string%20%7D%3B%0D%0A%0D%0Ahit.objectID%3B%0D%0Ahit2.objectID%3B%0D%0A%0D%0Ahit.a%3B%0D%0Ahit2.a%3B%0D%0A

For this type, there is a subset of known fields on the object, `objectID`.
Given: `let hit: BrowseResponse['hits'][number];`, the following is always valid: `hit.objectID;`; and the typings for this library should reflect this fact.

Adding `[key: string]: any` would allow for accessing any attribute on the object, even if it does not exist causing Typescript to throw an errors. 
See https://www.typescriptlang.org/play/#src=const%20hit%3A%20%7B%20objectID%3A%20string%3B%20%5Bkey%3A%20string%5D%3A%20any%20%7D%20%3D%20%7BobjectID%3A%20'my-id'%7D%3B%0D%0A%0D%0Ahit.objectID%3B%0D%0Ahit.a.toString()%3B%0D%0A

The current fashion in TypeScript would be to use `[key: string]?: unknown`, which is no more ergonomic or clear than omitting these attributes.

Other mechanisms; such as guards, `in` + `typeof`, or a validation library (`io-ts`); should be used to verify other attributes exist and are of the correct type before accessing them. If these methods are not desirable, user's of this typing should subvert the typesystem locally in their code using `as any` or `as {[key: string]: any}`.

neryortez: You're right, my suggestion is not sufficiently secure.

A more complete way would be to create a `Hit` type as a `Union Type`:
```
type Hit<T = {}> = T & { objectID: string };
```
You can [see it working successfully here](https://www.typescriptlang.org/play/#src=type%20Hit%3CT%20%3D%20%7B%7D%3E%20%3D%20T%20%26%20%7B%20objectID%3A%20string%20%7D%3B%0D%0A%0D%0Ainterface%20MyObject%20%7B%0D%0A%20%20%20%20prop1%3A%20string%3B%0D%0A%7D%0D%0A%0D%0Alet%20hit%3A%20Hit%3CMyObject%3E%3B%0D%0Alet%20hit2%3A%20Hit%3B%0D%0A%0D%0Ahit.objectID%3B%0D%0Ahit.prop1%3B%0D%0A%0D%0A%2F%2F%20It%20still%20lets%20you%20use%20objectID%0D%0Ahit2.objectID%3B%0D%0A%2F%2F%20And%20gives%20error%20when%20props%20don't%20exist%0D%0Ahit.anotherProp%3B%0D%0Ahit2.a%3B%20%0D%0A).
But that will require the `BrowseResponse` to have a generic type:
```
interface BrowseResponse<T> {
    cursor?: string;
    hit: Hit<T>;
    // the rest of the BrowseResponse props
}
```
As well as the methods that use `BrowseResponse`.

I have made a pull request to @KSaryan fork [here](https://github.com/KSaryan/DefinitelyTyped/pull/1). That uses the `Hit` type on the search methods as well.

squirly: This creates an unbounded generic which is a similar issue to casting `as any`. Generics should be used for aiding inference not casting.

neryortez: Please watch my [pull request](https://github.com/KSaryan/DefinitelyTyped/pull/1). As it is currently working as expected on a project of mine.

It lets the user bind the generic to a type or not. Will always have the `objectID` property. And TypeScript throws error when a property doesn't exist.

squirly: I do not think this is a good approach, it is not typesafe. Instead users should validate the recieved data instead of casting using a generic. If casting is desired, it should be done explicitly:
`hit as any as T`

Casting is okay in application code where assumptions are known, but in libraries typings should make minimal assumptions and strive for correctness.

neryortez: Working on it...

neryortez: The pull request has been updated.

Working example can be found [here.](https://stackblitz.com/edit/typescript-zxbtbj)

squirly: It seems there are no further objections to this iterative change. It adds no breaking changes and does not reduce the typesafety. I feel this is mergable as is, and further changes can be discussed and made in a seperate PR.

francoischalifour: This type is too restrictive. You can take a look at the [`Hit` type](https://github.com/algolia/instantsearch.js/blob/65911504b9ae0e2596c75f0ae74133985dc79850/src/types/instantsearch.ts#L46-L71) we created inside InstantSearch.js (which relies on `algoliasearch`). You can ignore the two `__position` and `__queryID` keys which are internal to InstantSearch.

[See Algolia response reference →](https://www.algolia.com/doc/api-reference/api-methods/search/#method-response-hits)

francoischalifour (changes requested):

types/algoliasearch/index.d.ts
squirly: The proposed change is less restrictive than what is currently implemented.
If additional work is desired, is it possible to do it in a follow-up PR?

*Note on the linked code above:*
I am very against adding the following line:
```ts
[attribute: string]: any;
```
It breaks down most of the typesafety in a way that is not reversible by the developer using it.

If anything it should be:
```ts
[attribute: keyof any]?: unknown;
```
Which requires correct narrowing into the type system before a value is accessible.

Haroenv: that's a very good idea, I'd accept that @squirly,  thanks again!

PR #35361, opened 2019-05-11
[paper] Fix hitTest types, add constructors to PaperScope

dobesv:

* http://paperjs.org/reference/item/#hittest-point

> options.class: Function — only hit-test against a specific item class, or any of its sub-classes, by providing the constructor function against which an instanceof check is performed: Group, Layer, Path, CompoundPath, Shape, Raster, SymbolItem, PointText, …
>
> ...
>
> Returns: HitResult — a hit result object describing what exactly was hit or null if nothing was hit

* http://paperjs.org/reference/paperscope/

> When working with normal JavaScript code, PaperScope objects need to be manually created and handled.
> 
> Paper classes can only be accessed through PaperScope objects. Thus in PaperScript they are global, while in JavaScript, they are available on the global paper object. For JavaScript you can use paperScope.install(scope) to install the Paper classes and objects on the global scope. Note that when working with more than one scope, this still works for classes, but not for objects like paperScope.project, since they are not updated in the injected scope if scopes are switched.
> 
> The global paper object is simply a reference to the currently active PaperScope.

### DIFF
--- a/types/paper/index.d.ts
+++ b/types/paper/index.d.ts
@@ -1236,6 +1236,37 @@ declare module paper {
          */
         static get(id: string): PaperScope;
 
+        /**
+         * All the classes for paper can be constructed via the PaperScope instance, so
+         * that they are associated with that scope
+         */
+        Matrix: typeof Matrix;
+        Point: typeof Point;
+        Rectangle: typeof Rectangle;
+        Size: typeof Size;
+        PaperScope: typeof PaperScope;
+        Item: typeof Item;
+        Group: typeof Group;
+        Layer: typeof Layer;
+        Shape: typeof Shape;
+        Raster: typeof Raster;
+        PlacedSymbol: typeof PlacedSymbol;
+        PathItem: typeof PathItem;
+        Path: typeof Path;
+        CompoundPath: typeof CompoundPath;
+        Segment: typeof Segment;
+        Curve: typeof Curve;
+        CurveLocation: typeof CurveLocation;
+        Project: typeof Project;
+        Symbol: typeof Symbol;
+        Style: typeof Style;
+        Color: typeof Color;
+        Gradient: typeof Gradient;
+        GradientStop: typeof GradientStop;
+        View: typeof View;
+        Tool: typeof Tool;
+        TextItem: typeof TextItem;
+        PointText: typeof PointText;
     }
     export interface IHitTestOptions{
 
--- a/types/paper/index.d.ts
+++ b/types/paper/index.d.ts
@@ -1276,9 +1276,9 @@ declare module paper {
         tolerance?: number;
 
         /**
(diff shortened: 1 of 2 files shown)


types/paper/index.d.ts
```diff
@@ -1247,7 +1278,7 @@ declare module paper {
         /**
          * only hit-test again a certain item class and its sub-classes: Group, Layer, Path, CompoundPath, Shape, Raster, PlacedSymbol, PointText, etc.
          */
-        class?: string;
+        class?: Function;
```
sandersn: prefer `(...args: any[]) => any` to `Function` unless you want to support functions created with `new Function(...)`.

sandersn (changes requested):
Requesting changes until @rfried's question is answered. (So that the bot will tag this PR correctly.)

dobesv:
You are supposed to pass in one of the paper class objects.  See the example.  I wasn't sure how best to write that as a type, Function works.  The arrow syntax is not right.  It's supposed to be an actual class that works with instance of.  Refer to the docs, there is no mention of a string.  I don't know the history of that, it's possible it used to take a string, I don't know.

dobesv:
@rfried I think if you look at the documentation I quoted & linked in the PR description it's pretty clear what the type of the field is supposed to be currently.  I don't know if it was ever a string or if it would (still?) accept a string.

@sandersn It is supposed to be a "class object" e.g. something you can new "new" and "instanceof" on.  I wasn't sure how to write that.  I did a bit more looking around and tried something the might be better, let me know.

types/paper/index.d.ts
dobesv: Changed to `class?: { new(...args: any[]): Item; };`

sandersn:
@dobesv OK, you have exactly the right type now. Much better than `Function`.

@rfried Does that explanation make sense to you? If so, I will merge this PR.

rfried:
I'm away from a computer for a few hours. The thing I want to check is that 
I'm pretty sure I got that it was a string from the paper. Js code and not 
their documentation. I expect the reason Function worked is that the 
"classes" in javascript are implemented that way. I'm almost certain that 
"|string" would work.

dobesv:
If you look here, it appears to support a field called `type` which is a `string`.  `class` is used in the `instanceof` version of the check.

https://github.com/paperjs/paper.js/blob/develop/src/item/Item.js#L2022

```
       var checkSelf = !(options.guides && !this._guide
                || options.selected && !this.isSelected()
                // Support legacy Item#type property to match hyphenated
                // class-names.
                || options.type && options.type !== Base.hyphenate(this._class)
                || options.class && !(this instanceof options.class)),
```

rfried:
@dobesv 

Yup that's what I was thinking of and no it isn't the class as you said. Sorry for the delay.

Change looks good

rfried (approved):

sandersn (approved):

PR #35374, opened 2019-05-11
Add new types/jsviews definitions

BorisMoore:
Definitions for https://www.npmjs.com/package/jsviews
(See depends on https://www.npmjs.com/package/jsrender - see types/jsrender)

If adding a new definition:
- [x] `tslint.json` should be present, and `tsconfig.json` should have `noImplicitAny`, `noImplicitThis`, and `strictFunctionTypes` set to `true`. (However `strictNullChecks` is set to `false`)

- n/a

If removing a declaration:
- n/a

### DIFF
--- a/types/jsviews/index.d.ts
+++ b/types/jsviews/index.d.ts
@@ -1,0 +1,376 @@
+// Type definitions for JsViews 1.0
+// Project: http://www.jsviews.com/#jsviews
+// Definitions by: Boris Moore <https://github.com/borismoore>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
+
+/// <reference types="jquery" />
+/// <reference types="jsrender" />
+
+declare module 'jsviews' {
+  export = jsviews;
+}
+
+declare const jsviews: ((jquery?: JQueryStatic) => JQueryStatic) & JQueryStatic;
+
+// ********************************** JsObservable **********************************
+
+interface JQueryStatic {
+  /* $.observable() */
+  observable: JsViews.Observable;
+
+  /* $.observe() */
+  observe: JsViews.Observe;
+
+  /* $.unobserve() */
+  unobserve: JsViews.Unobserve;
+}
+
+declare namespace JsViews {
+
+interface Observable {
+  /* $.observable(array) */
+  (data: any[]): ArrayObservable;
+  (ns: string, data: any[]): ArrayObservable;
+
+  /* $.observable(object) */
+  (data: any): ObjectObservable;
+  (ns: string, data: any): ObjectObservable;
+}
+
+interface Observe {
+  /* $.observe(...) */
+  (object: any[], handler: ChangeHandler<ArrayEvtArgs>): any;
+  (object: object, path: string, handler: ChangeHandler<EvtArgs>): any;
+  (object: object, path: string, path2: string, handler: ChangeHandler<EvtArgs>): any;
+  (object: object, path: string, path2: string, path3: string, handler: ChangeHandler<EvtArgs>): any;
+  (object: object, path: string, path2: string, path3: string, path4: string, handler: ChangeHandler<EvtArgs>): any;
+  (object: object, ...restOf: any[]): any;
+
+  (ns: string, object: any[], handler: ChangeHandler<ArrayEvtArgs>): any;
+  (ns: string, object: object, path: string, handler: ChangeHandler<EvtArgs>): any;
+  (ns: string, object: object, ...restOf: any[]): any;
+}
+
+interface Unobserve {
+  /* $.unobserve(...) */
+  (object: any[], handler?: ChangeHandler<ArrayEvtArgs>): any;
+  (object: object, path?: string, handler?: ChangeHandler<EvtArgs>): any;
+  (object: object, handler: ChangeHandler<EvtArgs>): any;
+  (object: object, ...restOf: any[]): any;
+  (handler?: ChangeHandler<EvtArgs>): any;
+
+  (ns: string, object: any[], handler?: ChangeHandler<ArrayEvtArgs>): any;
+  (ns: string, object: object, path?: string, handler?: ChangeHandler<EvtArgs>): any;
+  (ns: string, object: object, handler: ChangeHandler<EvtArgs>): any;
+  (ns: string, object: object, ...restOf: any[]): any;
+  (ns: string, handler?: ChangeHandler<EvtArgs>): any;
+}
+
+interface ObjectObservable {
+  /* Observable object, returned by $.observable(object) */
+  setProperty(path: string, value: any): ObjectObservable;
+  setProperty(values: Hash<any>): ObjectObservable;
+  observeAll(handler: ChangeHandler<EvtArgs>, filter?: FilterHandler): void;
+  unobserveAll(handler?: ChangeHandler<EvtArgs>, filter?: FilterHandler): void;
+  removeProperty(path: string): ObjectObservable;
+}
+
+interface ArrayObservable {
+  /* Observable array, returned by $.observable(array) */
+  insert(index: number, data: any): ArrayObservable;
+  insert(data: any): ArrayObservable;
+  refresh(newItems: any[]): ArrayObservable;
+  remove(index?: number, numToRemove?: number): ArrayObservable;
+  move(oldIndex: number, newIndex: number, numToRemove?: number): ArrayObservable;
+  observeAll(handler: ChangeHandler<EvtArgs>, filter?: FilterHandler): void;
+  unobserveAll(handler?: ChangeHandler<EvtArgs>, filter?: FilterHandler): void;
+}
+
+interface FilterHandler {
+  /* Filter function (callback handler) passed to observeAll(...) */
+  (path: string, ob: any, parentObs: Array<{ [prop: string]: any }>): void;
+}
+
+interface ChangeHandler<T> {
+  /* Change handler function passed to $.observe() or $.observable.observeAll() */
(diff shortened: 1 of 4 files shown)


BorisMoore:
BTW - just to let you know, I am also the owner of https://www.npmjs.com/package/jsviews (see https://github.com/BorisMoore/jsviews)

types/jsviews/index.d.ts
```diff
@@ -0,0 +1,376 @@
+// Type definitions for JsViews 1.0
+// Project: http://www.jsviews.com/#jsviews
+// Definitions by: Boris Moore <https://github.com/borismoore>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
+
+/// <reference types="jquery" />
+/// <reference types="jsrender" />
+
+declare module 'jsviews' {
```
sandersn: please use the new UMD syntax:

``
export = jsviews;
export as namespace jsviews;
```

You may have to move `interface JQueryStatic` inside a `declare global` to make it properly extend jquery.
You'll also want to change the namespace name to `declare namespace jsviews` so that it merges properly with the jsviews export, so that the types are remain accessible as `jsviews.Observable` (for example).

**Review by sandersn — CHANGES_REQUESTED** [review]
Need to use new UMD syntax.

File: types/jsviews/index.d.ts (hunk 1 above)
Comment (BorisMoore) [reply types/jsviews/index.d.ts]: Hi, thanks for the suggestions. However I'm not sure how to make those changes compatible with the jsrender type definitions at types/jsrender. The JsViews namespace is [defined within the jsrender definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/jsrender/index.d.ts#L35), and is used by jsviews also. Interfaces used by jsrender are then extended in the presence of jsviews. (The JsViews platform uses - and extends - JsRender platform). So the jsviews module is not the same as the JsViews namespace - in the current definitions...

Comment (BorisMoore) [reply types/jsviews/index.d.ts]: I should perhaps clarify that the typescript definitions for JsViews and JsRender are not new - they  have been published [here](https://www.jsviews.com/#typescript), since [the v1.0.0 commit](https://github.com/BorisMoore/jsviews/issues/175). However the JsViews index.d.ts was not yet published to DefinitelyTyped (so I described it as 'new' in that context).

So the goal is of course to publish the same definitions as were already published, and avoid any breaking changes.

Both JsRender and JsViews typescript definitions have been up and running for several months, without issues. There a many different use-case scenarios, including [Browserify](https://www.jsviews.com/#node/browserify) and [Webpack](https://www.jsviews.com/#node/webpack), using modules, [node.js server scenarios](https://www.jsviews.com/#jsrnode), and regular web pages (SPA) - where JsRender and JsViews both extend `JQuery` (and `JQueryStatic`) but do not add any globals as such.

Therefore I think that the relationship between Modules and Namespaces needs to be maintained unchanged (even if the syntax were to be modified) - so that 'semantics' are preserved. And in index.d.ts, `jsviews` and `JsViews` need to be distinct from each other...

The `JsViews` namespace effectively concerns interfaces that are used in extending `JQuery` and `JQueryStatic` (at least for the scenarios where jQuery is present), and which are shared between jsrender and jsviews. Extending jQuery has been working correctly...

@sandersn: could you indicate how you feel we can best proceed? Thanks for you help on this...

**Comment by BorisMoore** [comment]
I replied [5 days ago](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35374#discussion_r284483574), within the [conversation](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35374#pullrequestreview-237415910), and again [3 days ago](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35374#discussion_r285284100), requesting clarification from @sandersn. He has not replied. I was awaiting his response. The changes he asked for seem to me to break the relationship between types/jsrender and types/jsviews, which share the JsViews namespace. All tests work with the current code, but do not seem to be able to run successfully with the change he proposed. Please see also my responses to him [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35374#discussion_r284483574) and [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35374#discussion_r285284100), for more context. 

If you can help me to figure how setting `export = jsviews; export as namespace jsviews;` can be compatible with the types/jsviews/jsviews-tests.ts  provided [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35374/files#diff-c7027060e6e636fa8a8e38a59542b954), I'd be glad to make the change. Or if you can direct me to documentation that shows  how that can be done... 

Otherwise, I believe that the [conversation](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35374#pullrequestreview-237415910) can be resolved, following my responses, and the pull request can be merged. 

Including @jessetrinity who merged [types/jsrender](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35148#event-2326009286), on which types/jsviews depends...

Please let me know if I am missing anything.

**Comment by BorisMoore** [comment]
Since @sandersn did not reply, I will take that as meaning he agrees to my response that his proposed change is not finally appropriate, given the dependency on types/jsrender, and on the existing published jsviews.d.ts [here](https://www.jsviews.com/#typescript), so I will resolve that conversation. Let me know if this does not work for you....

File: types/jsviews/index.d.ts (hunk 1 above)
Comment (sandersn) [reply types/jsviews/index.d.ts]: Still reading, but it looks like the initial solution is a module augmentation:

```ts
declare module 'jsrender' {
  declare namespace JsViews {
    // augmentations that are currently starting on line 31 go here...
  }
}
```

sandersn: If JsViews and JsRender install to node_modules and have a package.json, then you should not add types to Definitely Typed at all. Just have a "types" entry in package.json that points to the correct d.ts files.

BorisMoore: Is there any reason we can't just go with the current version, since it works correctly, and is exactly the same approach as was used and correctly merge in types/jsrender?

I understand that TypeScript evolves, and that you appreciate when the more up-to-date style is used. But the previous style works too, and has already [been published](https://www.jsviews.com/#typescript) (with download [here](https://www.jsviews.com/download/typescript/jsviews/index.d.ts)) and is being used by end users and developers... JsRender and JsViews are both used extensively by individuals and by teams... 

I really don't want to make unnecessary changes that create churn and confusion and risk causing breaking changes. To make changes for aesthetic or stylistic reasons seems out of place to me, here....

BorisMoore: It says [here](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html)

> There are two main ways you can publish your declaration files to npm:
> 
> - bundling with your npm package, or
> - publishing to the @types organization on npm.
> 
> If your package is written in TypeScript then the first approach is favored. Use the --declaration flag to generate declaration files. This way, your declarations and JavaScript will always be in sync.
> If your package is not written in TypeScript then the second is the preferred approach.

JsRender and JsViews were not written in TypeScript, and are mostly used by people coding in JavaScript, not using TypeScript. Therefore it seems to me to make more sense to follow the published guidance above, and not include index.d.ts in the package.

BorisMoore: While JsRender and JsViews can install to node-modules using npm, by far the majority of users obtain them from  [here](http://localhost:7399/INDEXNOTMINIFIED.html#download) or by CDN [here](https://cdnjs.com/libraries/jsviews) or [here](https://cdnjs.com/libraries/jsrender) and use them in web pages. They may still need to use TypeScript, but they have not installed an npm package at all, and may not even know that an npm package exists...

RyanCavanaugh:
@BorisMoore we're managing literally hundreds of PRs here and do take weekends off to see our families; declaring yourself victor by default because someone hasn't replied on your time table is not how this works.

types/jsviews/index.d.ts
BorisMoore: If I try to use your augmentation style, then I need replace both instances of `declare namespace JsViews` [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/28031887bc521a6ee517b79ee5b750262c36b4b2/types/jsviews/index.d.ts#L79) and [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/28031887bc521a6ee517b79ee5b750262c36b4b2/types/jsviews/index.d.ts#L183) by [your suggested code](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35374#discussion_r285745776) (with an additional closing  `}` [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/28031887bc521a6ee517b79ee5b750262c36b4b2/types/jsviews/index.d.ts#L148) and [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/28031887bc521a6ee517b79ee5b750262c36b4b2/types/jsviews/index.d.ts#L376)).

If I then run tsc on the test files I get dozens (maybe hundreds) of errors.

I went in to my local copy of jsrender index.d.ts, and made the same change there for declare namespace JsViews. Again, hundeds (?) of errors.

So it seems like these changes are not appropriate, and (it seems to me) not necessary.

BorisMoore:
Yes, @RyanCavanaugh, appreciated. I'm sorry but I felt stuck as the bot was saying it would close the pull request tomorrow because of inactivity on my part, and yet I had in fact been active in responding to @sandersn's proposal, and was ready to wait longer. But the bot would have then close the PR. I agree you have a huge amount of work, and I thank you for that...

BorisMoore:
I have indeed been active (see the exchanges above), but have not been able to establish how the reviewer's proposed code change can be achieved without breaking the majority of test cases.

PR #35378, opened 2019-05-12
[csv-parse] Remove package

ffflorian:
Please fill in this template.

- [x] ~~Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).~~

If removing a declaration:

See https://github.com/adaltas/node-csv-parse/blob/v1.2.2/lib/index.d.ts


ffflorian:
@wdavidw @adaltas can you maybe approve? :)

wdavidw:
I approve, we have internalized the TS definitions. Since I am here, I would enjoy a feedback on your side in term of good practices. For example, we don't have a "tsconfig.json" nor a "tslint.json", is it necessary? The concerned packages are:

* [csv](https://github.com/adaltas/node-csv)
* [csv-generate](https://github.com/adaltas/node-csv-generate)
* [csv-parse](https://github.com/adaltas/node-csv-parse)
* [csv-stringify](https://github.com/adaltas/node-csv-stringify)
* [stream-transform](https://github.com/adaltas/node-stream-transform/)

sandersn:
@wdavidw If you are writing the code in JS then you only need the tsconfig.json for the tests for your types, or you can just use command line args of course. I didn't see tests in `csv` so this may be future work for you.

Of course you can add one to get checking for your JS code as well, but `// @ts-check` works for that too.

I would not personally add tslint.json at this point. Just stick with eslint if you use that.

wdavidw:
Thank you for your feedback. Inside each package test folder, we have a single ts test file which only checks the API. Those tests are executed with Mocha along the other coffee tests.

PR #35380, opened 2019-05-12
[storybook-readme] Add new feature importing addReadme from html

lonyele:

According to https://github.com/tuchk4/storybook-readme/issues/87#issuecomment-489794946 I added a few lines on this. The actual implementation code is from https://github.com/tuchk4/storybook-readme/blob/master/packages/storybook-readme/src/html/index.js


lonyele:
hm... somehow my local setting is messed up. I couldn't run lint or tests....

types/storybook-readme/storybook-readme-tests.tsx
```diff
@@ -3,6 +3,7 @@ import { storiesOf, addDecorator } from "@storybook/react";
 import { withDocs, withReadme, doc, addReadme, configureReadme, addFooter, addHeader } from "storybook-readme";
 import Marked from "storybook-readme/components/Marked";
 import registerWithPanelTitle from "storybook-readme/registerWithPanelTitle";
+import { addReadme as addHtmlReadme } from "storybook-readme/html";
```
lonyele: I'm not sure if it's right or not. From current architecture, named export `addRadme` is imported from `storybook-readme` or `storybook-readme/html`. Problem is when I had to write a test on this. Since it is a type definition, it looks fine... or maybe not. What do you think? Should I use another test files for this?

sandersn: I don't entirely understand. Regardless, it sounds like you should be able to write either

```ts
import { addReadme } from 'storybook-readme/html'
// OR
import { addReadme } from 'storybook-readme'
```

It is fine to do that in the same test file as long as you use two different local names, which you did.

sandersn (approved):

types/storybook-readme/storybook-readme-tests.tsx
lonyele: Yeah, just explain a little bit more. That two different `addReadme` can't be used at the same time in actual usage. However when I was writing this test, I wanted to verify if newly added `addReadme` from `storybook-readme/html` is importing fine and its imported value(`addReadme`) can be an argument of `addDecorator(addHtmlReadme)`
So as you said, doing like this looks fine

PR #35385, opened 2019-05-12
[nock] Fix Interceptor.reply.

mastermatt:
I was working on removing the `no-any-union` lint override and realized
that `any` was being used in several places where a plain objected was
intended. It was also covering up that reply bodies can be Buffers or
Streams.

[Docs for specifying replies](https://github.com/nock/nock#specifying-replies)

Note: There was a bug in place. Previously, the types allowed for `reply`
to take two args, a callback and an object of headers. However, this
form is not documented and turns out to not be functional when
inspecting [the source](https://github.com/nock/nock/blob/cb56669e0bee13dfa9c668ecb3d0efb714f6240c/lib/interceptor.js#L68-L72).

I’ve confirmed with the lib maintainers that the bug was in `@types`.
https://github.com/nock/nock/issues/1513

- [x] Provide a URL to documentation or source code which provides context for the suggested changes.

### DIFF
--- a/types/nock/index.d.ts
+++ b/types/nock/index.d.ts
@@ -10,6 +10,8 @@
 
 /// <reference types="node" />
 
+import { ReadStream } from 'fs';
+import { ClientRequest } from "http";
 import { Url } from 'url';
 
 export = nock;
@@ -39,14 +41,25 @@ declare namespace nock {
 
     let back: NockBack;
 
+    interface POJO { [k: string]: any; }
+    type RequestBodyMatcher = string | Buffer | RegExp | POJO | { (body: any): boolean; };
+    interface RequestHeaderMatcher { [key: string]: string | RegExp | { (headerValue: string): boolean; }; }
+
     interface HttpHeaders { [key: string]: string | string[] | { (req: any, res: any, body: string): any; }; }
     type InterceptFunction = (
         uri: string | RegExp | { (uri: string): boolean; },
-        requestBody?: string | RegExp | { (body: any): boolean; } | any,
+        requestBody?: RequestBodyMatcher,
         interceptorOptions?: Options
     ) => Interceptor;
+
+    type ReplyBody = string | Buffer | ReadStream | POJO;
     type ReplyCallback = (err: any, result: ReplyCallbackResult) => void;
-    type ReplyCallbackResult = string | [number, string | any] | [number, string | any, HttpHeaders] | any;
+    type ReplyCallbackResult = ReplyBody | [number, ReplyBody] | [number, ReplyBody, HttpHeaders];
+    interface ReplyCallbackContext extends Interceptor {
+        req: ClientRequest & {
+            headers: { [k: string]: string }
+        };
+    }
 
     interface Scope extends NodeJS.EventEmitter {
         get: InterceptFunction;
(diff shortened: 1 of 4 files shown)


bennyn:
@mastermatt This change seems to break my code.

I was doing the following:

```typescript
nock(global.REST_URL)
.post(global.REST_RESOURCE)
.query(() => true)
.reply((uri: string, body: string) => {
  return [
    200,
    JSON.stringify({
      something: 'something'
    }),
  ];
});
```

It worked when using @types/nock v10.0.2 but with v10.0.3 I receive the following error:

> error TS2345: Argument of type '(uri: string, body: string) => (string | number)[]' is not assignable to parameter of type 'number'.

Checking [nock's specifying replies section](https://github.com/nock/nock/tree/v10.0.6#specifying-replies) it should work but with the new types I have to pass an extra response code to `reply` in order to make it work:

**Workaround**

```typescript
nock(global.REST_URL)
.post(global.REST_RESOURCE)
.query(() => true)
.reply(200, (uri: string, body: string) => {
  return [
    200,
    JSON.stringify({
      something: 'something'
    }),
  ];
});
```

Can you have a look at this?

mastermatt:
@bennyn sorry for the inconvenience. First off, there was nothing wrong with your initial version as far as the JS version of Nock is concerned. 

The change I made that is affecting you was a fix to the signature of the callback passed to `reply`. Specifically, it use to declare that the `body` argument was a string. However, this is not always true. [JSON bodies are parsed into plain objects](https://github.com/nock/nock/blob/31623fbc01a1c5a827580ee393dc3a3d1846f225/lib/request_overrider.js#L304-L314).

This is affecting you because of the way you declared the callback signature `(uri: string, body: string)`.
If you instead you write it like this, the issue will disappear.
```ts
.reply((uri, body) => {
  return [
    200,
    JSON.stringify({
      something: 'something'
    }),
  ];
})
```
Doing it this way, TS will know that `uri` is a string and `body` is either a string **or** an object.
Depending on your use case, you may or may not need to handle the variability of the type inside the callback block.

Going back through this, I did find a small bug where it thinks the `body` could also be a `Buffer`, but I don't think that will get in your way.

bennyn:
Thanks for your fast reply! I removed the type declarations from my method signature (to let TypeScript automatically resolve the types) but this gives me an error when using `JSON.parse(body)`:

> error TS2345: Argument of type 'ReplyBody' is not assignable to parameter of type 'string'. Type 'POJO' is not assignable to type 'string'.

Here is a screenshot from my application:

image

I logged `typeof body` and it is `string`. Have I found just another corner case?

mastermatt:
So in your case @bennyn, the `body` arg is a string, however, that's not always true. As such, the type of `body` is set as a union between a string and an object. 

That TS error you're getting is basically telling you to explicitly narrow down the type.

There are a couple different ways to do this.
You could support both. If the req starts sending a JSON content-type header the body will be parsed for you already.
```ts
const newOrder: NewOrder = typeof body === "string" ? JSON.parse(body) : body;
```

Another option would be to throw an error if you don't get a string
```ts
if (typeof body !== "string") {
  throw "oh no! ...";
}
```

In either case you need some sort of casting in order to have a variable with a `NewOrder` type.

_Having a generic form of `.reply` might be a nice feature._ 🤔

bennyn:
Ok, checking for `typeof body === "string"` did the trick! 🎉 

I was just confused because it was working before the typings update. Thanks for helping me with this. 👍

PR #35386, opened 2019-05-12
[hapi]: fix custom events and (partly) nested server methods

Ninerian:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://hapijs.com/api#-servermethodname-method-options

While working with the types for Hapi, I discovered two flaws, which are IMO incorrect typed and not representing the documented functionality.

## 1. Custom events

The hapi method `server.event(eventName)` allows to register a custom event type. But when I try to use this in an event handler, typescript throws an error, as it only expects the default events in the ServerEvents interface. To fix this, I added an generic handler for `on` and `once` to handle the custom events.

Docu: 
## 2. Nested server methods.

As in the documentation for [server.methods](https://hapijs.com/api#-servermethodname-method-options):

> Method names can be nested (e.g. utils.users.get) which will automatically create the full path under server.methods (e.g. accessed via server.methods.utils.users.get). 

The current type of 

```typescript
  interface Dictionary<T> {
        [key: string]: T;
    }
```

does not allow this nesting. 

Unfortunately does my change no pass the tests and so I would left this PR for discussion, how to solve the issue.

 @rafaelsouzaf, @jhsimms, @SimonSchick, @saboya

### DIFF
--- a/types/hapi/index.d.ts
+++ b/types/hapi/index.d.ts
@@ -4034,7 +4034,7 @@ export namespace Lifecycle {
 
 export namespace Util {
     interface Dictionary<T> {
-        [key: string]: T;
+        [key: string]: T | Dictionary<T>;
     }
 
     type HTTP_METHODS_PARTIAL_LOWERCASE = 'get' | 'post' | 'put' | 'patch' | 'delete' | 'options';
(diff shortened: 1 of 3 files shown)


SimonSchick (changes requested):
Going with no on this one, if you want to push custom events you want to extend the module instead via declaration merging, otherwise we lose type safety.

Ninerian:
@SimonSchick thanks for the explanation. But how could the nested methods be solved?

SimonSchick:
Nested server methods should probably NOT be solved by changing the dictionary type since that's a generic type used in many places.
You may want to change it in the required locations instead.

PR #35391, opened 2019-05-13
fix(sylvester): vector rotate should be able to take matrix as well

briwa:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [`Vector.prototype.rotate takes in Matrix as well`](https://github.com/pithumke/sylvester-es6/blob/master/src/Vector.js#L274)

I realized that when I was using the types for the [ES6 fork of sylvester]((https://github.com/pithumke/sylvester-es6)), apparently [Vector instance takes in Matrix as well](https://github.com/pithumke/sylvester-es6/blob/master/src/Vector.js#L274), not only the angle in radians. This is also the case for the [original library](https://github.com/jcoglan/sylvester/blob/master/src/vector.js#L192). I've fixed both the original and the fork.

Let me know what you think. Thanks.

### DIFF
--- a/types/sylvester-es6/index.d.ts
+++ b/types/sylvester-es6/index.d.ts
@@ -171,7 +171,7 @@ export class Vector {
      * Rotates the vector about the given object. The object should be a point if the vector is 2D,
      * and a line if it is 3D. Be careful with line directions!
      */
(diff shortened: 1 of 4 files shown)


StephaneAlie (approved):
👍

PR #35393, opened 2019-05-13
Feature(react-native): Fix Platform module definition

ovr:
Please fill in this template.

### Motivation

Platform is a global object working on top of `NativeModules.PlatformConstants`

Declaration

https://github.com/facebook/react-native/blob/0ee5f68929610106ee6864baa04ea90be0fc5160/Libraries/Utilities/Platform.android.js#L20

https://github.com/facebook/react-native/blob/0ee5f68929610106ee6864baa04ea90be0fc5160/Libraries/Utilities/Platform.ios.js#L20

I am having this code, I am interested to "dont cast Platform to PlatformIOSStatic type" because with union TS can determinate type and this will be type-guarded:

```typescript
    if (Platform.OS === 'ios') {
        if (isLandscape && !Platform.isPad) {
            return 32;
        } else {
            return 44;
        }
    } else if (Platform.OS === 'android') {
        return 56;
    } else {
        return 64;
    }
```

Inside IDE with this changes:

image

image

Thanks

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -5960,7 +5960,6 @@ export type PlatformOSType = "ios" | "android" | "macos" | "windows" | "web";
 
 interface PlatformStatic {
     isTV: boolean;
-    OS: PlatformOSType;
     Version: number | string;
 
     /**
@@ -5970,8 +5969,25 @@ interface PlatformStatic {
 }
 
 interface PlatformIOSStatic extends PlatformStatic {
+    OS: 'ios';
     isPad: boolean;
     isTVOS: boolean;
+}
+
+interface PlatformAndroidStatic extends PlatformStatic {
+    OS: 'android';
+}
+
+interface PlatformMacOSStatic extends PlatformStatic {
+    OS: 'macos';
+}
+
+interface PlatformWindowsOSStatic extends PlatformStatic {
+    OS: 'windows';
+}
+
+interface PlatformWebStatic extends PlatformStatic {
+    OS: 'web';
 }
 
 /**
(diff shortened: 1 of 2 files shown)


Skn0tt (approved):
Looks good to me.

types/react-native/test/index.tsx
```diff
@@ -909,3 +910,18 @@ const PermissionsAndroidTest = () => {
...
+            return 32;
+        } else {
+            return 44;
+        }
+    } else if (Platform.OS === 'android') {
+        return 56;
+    } else {
+        return 64;
+    }
+};
```
alloy: Can you update this test to access platform specific properties inside the android and else branches too?

types/react-native/index.d.ts
```diff
@@ -9470,8 +9486,7 @@ interface NativeModulesStatic {
  * <code>const MyModule = NativeModules.ModuleName</code>
  */
 export const NativeModules: NativeModulesStatic;
-export const Platform: PlatformStatic;
-export const PlatformIOS: PlatformIOSStatic;
```
alloy: This is going to break some code… but not sure how to best deprecate that otherwise 🤔 I guess readers will find this PR soon enough, when things break 🤷‍♂

alloy (changes requested):
Nice, I like it 👌

types/react-native/test/index.tsx
ovr: Added

types/react-native/index.d.ts
ovr: > I guess readers will find this PR soon enough, when things break

Sure, btw https://github.com/search?q=PlatformIOSStatic&type=Code

35 code results in the whole github, i think it's not used (only 0.001% of all RN users)

alloy (approved):

alloy:
Thanks!

PR #35398, opened 2019-05-13
Add types for dynamodb module

katsanva:

### DIFF
--- a/types/dynamodb/DynamoDB.d.ts
+++ b/types/dynamodb/DynamoDB.d.ts
@@ -1,0 +1,55 @@
+export interface Projection {
+    ProjectionType?: 'ALL' | 'KEYS_ONLY' | 'INCLUDE' | string;
+    NonKeyAttributes?: string[];
+}
+
+import * as stream from 'stream';
+
+export type DynamoDB = any;
+
+export type DocumentClient = any;
+
+export type binaryType =
+    | Buffer
+    | {}
+    | ArrayBuffer
+    | DataView
+    | Int8Array
+    | Uint8Array
+    | Uint8ClampedArray
+    | Int16Array
+    | Uint16Array
+    | Int32Array
+    | Uint32Array
+    | Float32Array
+    | Float64Array
+    | stream.Stream;
+
+interface StringSet {
+    type: 'String';
+    values: string[];
+}
+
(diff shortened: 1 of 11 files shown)


katsanva:
The reason why the build is failing is `aws-sdk` dependency. Any ideas on how that can be resolved?

sandersn:
aws-sdk uses lots of memory on TS3.4. I need to investigate it and fix it. I'll look at it after I've caught up with the other DT PRs.

katsanva:
@sandersn then I will bring back the `aws-sdk` dependency when memory issue will be resolved

types/dynamodb/Model.d.ts
```diff
@@ -0,0 +1,147 @@
...
+    destroy(options: any): Promise<any>;
+    toJSON(): object;
+}
+
+export class Model extends Item {
+    constructor(attrs: any);
+}
+
+export namespace Model {
+    const get: GetOperation;
```
sandersn: Aren't these consts and functions equivalent to statics on `class Model`? This is perfectly legal but it's an odd style.

Typescript doesn't support nested types in classes, so the types do need to be in a namespace.

types/dynamodb/DynamoDB.d.ts
```diff
@@ -0,0 +1,55 @@
...
+    TableName: string;
+    Key: object;
+    AttributesToGet?: string[];
+    ConsistentRead?: boolean;
+    ReturnConsumedCapacity?: string;
+    ProjectionExpression?: string;
+    ExpressionAttributeNames?: { [key: string]: string };
+}
+
+export {};
```
sandersn: Empty export lists are used to convert script files into modules, but this file is already a module because of all the other exports. I don't think it's needed.

sandersn (approved):
Clever workaround. I'll merge as-is, although I have a couple of style comments you might want to follow up on. I took your first commit and created a branch from it on DefinitelyTyped/DefinitelyTyped: fix-aws-dynamodb-crash.

I'll add commits as I try to fix the crash.

types/dynamodb/DynamoDB.d.ts
sandersn: The linter should probably forbid this actually.

PR #35399, opened 2019-05-13
[algoliasearch]: Add type for algolia places

aDEVE:
Please fill in this template.

### DIFF
--- a/types/algoliasearch/index.d.ts
+++ b/types/algoliasearch/index.d.ts
@@ -2063,7 +2063,7 @@ declare namespace algoliasearch {
          * List of associated postcodes.
          * https://community.algolia.com/places/api-clients.html#api-suggestion-postcode
          */
-        postcodes?: string[];
+        postcode?: string[];
         /**
          * Associated population.
          * https://community.algolia.com/places/api-clients.html#api-suggestion-population
--- a/types/algoliasearch/index.d.ts
+++ b/types/algoliasearch/index.d.ts
@@ -1881,40 +1881,176 @@ declare namespace algoliasearch {
   }
 }
 
-
 declare namespace algoliasearchInitPlaces {
-
+    /**
+     * Restrict the search results to a specific type.
+     * https://community.algolia.com/places/api-clients.html#api-options-type
+     */
     interface QueryInterface {
+        /**
+         * Query used to perform the search.
+         */
         query?: string;
+        /**
+         * Restrict the search results to a specific type.
+         * https://community.algolia.com/places/api-clients.html#api-options-type
+         */
         type?: "city" | "country" | "address" | "busStop" | "trainStation" | "townhall" | "airport";
+        /**
+         * Restrict the search results to a specific type.
+         * https://community.algolia.com/places/api-clients.html#api-options-type
+         */
         hitsPerPage?: number;
+        /**
+         * If specified, restrict the search results to a single language. You can pass two letters country codes (ISO 639-1).
+         * Warning: language parameter is case sensitive and should be lowercase otherwise it will fallback to default language.
+         * https://community.algolia.com/places/api-clients.html#api-options-language
+         */
         language?: string;
+        /**
+         * If specified, restrict the search results to a specific list of comma-separated countries. You can pass two letters country codes (ISO 3166-1).
+         * Default: Search on the whole planet.
+         * Warning: country codes must be lower-cased.
+         * https://community.algolia.com/places/api-clients.html#api-options-countries
+         */
         countries?: string;
+        /**
+         * Force to first search around a specific latitude longitude.
+         * The option value must be provided as a string: latitude,longitude like 12.232,23.1.
+         * The default is to search around the location of the user determined via his IP address (geoip).
+         * https://community.algolia.com/places/api-clients.html#api-options-aroundLatLng
+         */
         aroundLatLng?: string;
+        /**
+         * Whether or not to first search around the geolocation of the user found via his IP address. This is true by default.
+         * https://community.algolia.com/places/api-clients.html#api-options-aroundLatLngViaIP
+         */
         aroundLatLngViaIP?: string;
+        /**
+         * Radius in meters to search around the latitude/longitude. Otherwise a default radius is automatically computed given the area density.
+         * https://community.algolia.com/places/api-clients.html#api-options-aroundRadius
+         */
         aroundRadius?: number;
+        /**
+         * Controls whether the _rankingInfo object should be included in the hits. This defaults to false.
+         * The _rankingInfo object for a Places query is slightly different from a regular Algolia query
+         * and you can read up more about the difference and how to leverage them in our guide.
+         * https://community.algolia.com/places/api-clients.html#api-options-getRankingInfo
+         */
         getRankingInfo?: boolean;
     }
 
     interface PlaceInterface {
+        /**
+         * Endpoint to search.
+         * https://community.algolia.com/places/api-clients.html#endpoints
+         */
         search(e: QueryInterface, f: (err: any, response: ResultInterface) => void): void;
     }
 
+    /**
+     * Result of search.
+     * https://community.algolia.com/places/api-clients.html#json-answer
+     */
     interface ResultInterface {
+        /**
+         * Contains all the hits matching the query.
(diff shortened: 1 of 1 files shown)


types/algoliasearch/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for algoliasearch-client-js 3.30.1
+// Type definitions for algoliasearch-client-js 3.30.2
```
Haroenv: this version number doesn't exist

types/algoliasearch/index.d.ts
```diff
@@ -1880,9 +1881,210 @@ declare namespace algoliasearch {
...
+         */
+        getRankingInfo?: boolean;
+    }
+
+    interface PlaceInterface {
+        /**
+         * Endpoint to search.
+         * https://community.algolia.com/places/api-clients.html#endpoints
+         */
+        search(e: QueryInterface, f: (err: any, response: ResultInterface) => void): void;
```
Haroenv: there's also the Promise version

Haroenv (commented):
there's also a `places.reverse({ aroundLatLng: '' })` method, cc @JonathanMontane

I need to check, but I think it's might also be available in the lite version

types/algoliasearch/index.d.ts
aDEVE: It should be ok

Haroenv (approved):

types/algoliasearch/index.d.ts
```diff
@@ -1880,9 +1881,254 @@ declare namespace algoliasearch {
...
+            };
+            county: { default: string; [key: string]: string };
+            locale_names: { default: string[]; [key: string]: string[] };
+            postcode: highlightResultValueInterface[];
+        }
+        /**
+         * https://community.algolia.com/places/api-clients.html#api-suggestion-name
+         * List of names of the place. If no language parameter is specified, retrieves all of them.
+         */
+        locale_names: { default: string; [key: string]: string };
```
JonathanMontane: default and language fields should be `string[]` instead of `string`

types/algoliasearch/index.d.ts
```diff
@@ -1880,9 +1881,254 @@ declare namespace algoliasearch {
...
+    /**
+     * Hit of search.
+     * https://community.algolia.com/places/api-clients.html#api-suggestion-name
+     */
+    interface HitInterface {
+        /**
+         * List of associated administrative region names.
+         * https://community.algolia.com/places/api-clients.html#api-suggestion-administrative
+         */
+        administrative: string[];
```
JonathanMontane: as crazy as it sounds, administrative is `optional`

types/algoliasearch/index.d.ts
```diff
@@ -1880,9 +1881,254 @@ declare namespace algoliasearch {
...
+         */
+        query: string;
+    }
+
+
+    /**
+     * Hit of search.
+     * https://community.algolia.com/places/api-clients.html#api-suggestion-name
+     */
+    interface HitInterface {
```
JonathanMontane: the HitInterface shape is dependent on whether the language parameter was provided or not.

If a supported language was provided as a query parameter the interface looks like this:
```ts
interface LocalizedHitInterface {
  administrative?: string[];
  country: string;
  country_code: string;
  county?: string[];
  city?: string[];
  postcodes?: string[];
  population?: number;
  geoloc: ...
  
}

File: types/algoliasearch/index.d.ts [hunk 6]
```diff
@@ -1880,9 +1881,254 @@ declare namespace algoliasearch {
...
+         * Endpoint to search.
+         * https://community.algolia.com/places/api-clients.html#endpoints
+         */
+        search(e: QueryInterface, cb: (err: Error, response: ResultSearchInterface) => void): void;
+
+        /**
+         * Endpoint to search.
+         * https://community.algolia.com/places/api-clients.html#endpoints
+         */
+        search(e: QueryInterface): Promise<ResultSearchInterface>;
```
JonathanMontane: See my comments about the `language` parameter further below.

I don't know if this is valid typescript but we should probably have two separate declarations for the search method, if it's possible:

```ts
  search(e: QueryInterface): Promise<ResultSearchInterface>;
  search(e: LocalizedQueryInterface): Promise<LocalizedResultSearchInterface>;
```

JonathanMontane (changes requested):
Thanks a lot for taking the time to write this declaration file.
There a few assumptions that are incorrect, mostly due to limitations within places and early-stage design decisions that are confusing and not well known.

types/algoliasearch/index.d.ts
```diff
@@ -1880,9 +1881,254 @@ declare namespace algoliasearch {
   }
 }
 
-declare function algoliasearch(
-  applicationId: string,
-  apiKey: string,
-  options?: algoliasearch.ClientOptions
-): algoliasearch.Client;
+declare namespace algoliasearchInitPlaces {
```
keichinger: Where does that name come from? I couldn't find it in the source of `places` package (see https://github.com/algolia/places/search?q=algoliasearchInitPlaces&unscoped_q=algoliasearchInitPlaces).

I'm not really fond of that name :) If we could change it, I'd highly appreciate changing it to something like `algoliasearch.Places`. However, if it's a requirement to augment a specific JavaScript module that I can't find, then we have to roll with it I guess.

keichinger (changes requested):

types/algoliasearch/index.d.ts
aDEVE: Like Result has same structure if language is set but `hits` is changing, I did that :
```
hits: HitInterface[] | LocalizedHitInterface[];
```

types/algoliasearch/index.d.ts
aDEVE: Could you take a look, i made some changes 😁

keichinger (approved):

types/algoliasearch/index.d.ts
keichinger: Perfect! Thank you very much 🥰

types/algoliasearch/index.d.ts
```diff
@@ -1878,11 +1879,332 @@ declare namespace algoliasearch {
...
+        /**
+         * Associated population.
+         * https://community.algolia.com/places/api-clients.html#api-suggestion-population
+         */
+        population?: number;
+        /**
+         * Associated list of latitude and longitude.
+         * https://community.algolia.com/places/api-clients.html#api-suggestion-latlng
+         */
+        geoloc: { lat: number; lng: number };
```
JonathanMontane:
```suggestion
        _geoloc: { lat: number; lng: number };
```

types/algoliasearch/index.d.ts
```diff
@@ -1878,11 +1879,332 @@ declare namespace algoliasearch {
...
+        /**
+         * https://community.algolia.com/places/api-clients.html#api-suggestion-city
+         * List of the associated city names. If no language parameter is specified, retrieves all of them.
+         */
+        city?: string[];
+        /**
+         * List of associated postcodes.
+         * https://community.algolia.com/places/api-clients.html#api-suggestion-postcode
+         */
+        postcodes?: string[];
```
JonathanMontane:
```suggestion
        postcode?: string[];
```

types/algoliasearch/index.d.ts
```diff
@@ -1878,11 +1879,332 @@ declare namespace algoliasearch {
...
+        /**
+         * List of associated postcodes.
+         * https://community.algolia.com/places/api-clients.html#api-suggestion-postcode
+         */
+        postcode?: string[];
+        /**
+         * Associated list of latitude and longitude.
+         * https://community.algolia.com/places/api-clients.html#api-suggestion-latlng
+         */
+        geoloc: { lat: number; lng: number };
```
JonathanMontane:
```suggestion
        _geoloc: { lat: number; lng: number };
```

JonathanMontane (commented):
Thank you so much for taking the time to do these typings.

I have found 3 things that are incorrect, and I have provided you with suggestions to fix these details.

aDEVE:
I made these changes, thanks for provided suggestions 😀

JonathanMontane (approved):
Thanks a lot for this type definition 👍

PR #35401, opened 2019-05-13
[react-select] Fix the default exported component by using the StateManager

pcorpet:
The exported component of the ReactSelect library is actually a wrapper around the Select component, using a StateManager. See the wrapping call [here](https://github.com/JedWatson/react-select/blob/master/src/index.js#L7). The updated component accepts the same props but it now handles properly having a `select` field on which one can call methods like `focusOption`.

I think the previous version was in place to quickly get a type running but does not match the values returned by the library.

Checklist:

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/JedWatson/react-select/blob/master/src/index.js#L7>

I'm not sure about increasing the version number: it's still related to the same major/minor version of the React Select library. However it's a breaking change compare to the (broken) existing state of the types.

### DIFF
--- a/types/react-select/lib/stateManager.d.ts
+++ b/types/react-select/lib/stateManager.d.ts
@@ -4,10 +4,31 @@ import SelectBase, { Props as SelectProps } from './Select';
 import { ActionMeta, InputActionMeta, ValueType } from './types';
 
 export interface DefaultProps<OptionType> {
+  defaultInputValue: string;
+  defaultMenuIsOpen: boolean;
+  defaultValue: ValueType<OptionType>;
+}
+
+export interface Props<OptionType> {
   defaultInputValue?: string;
   defaultMenuIsOpen?: boolean;
   defaultValue?: ValueType<OptionType>;
+  inputValue?: string;
+  menuIsOpen?: boolean;
+  value?: ValueType<OptionType>;
+  onChange?: (value: ValueType<OptionType>, actionMeta: ActionMeta) => void;
 }
+
+type StateProps<T extends SelectProps<any>> = Pick<T, Exclude<keyof T,
+  | 'inputValue'
+  | 'value'
+  | 'menuIsOpen'
+  | 'onChange'
+  | 'onInputChange'
+  | 'onMenuClose'
+  | 'onMenuOpen'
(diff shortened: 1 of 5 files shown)


types/react-select/lib/stateManager.d.ts
```diff
@@ -1,25 +1,28 @@
...
 
-export class StateManager<OptionType> extends Component<Props<OptionType>, State<OptionType>> {
-  static defaultProps: Props<any>;
+type GetOptionType<T> = T extends SelectBase<infer OT> ? OT : never;
 
-  select: ElementRef<any>;
+export class StateManager<
+  OptionType = { label: string; value: string },
+  T extends SelectBase<OptionType> = SelectBase<OptionType>
+> extends Component<SelectProps<OptionType> & DefaultProps<OptionType>, State<OptionType>> {
```
claasahl: Looking at the original sources, I would recommend to reference `Props` and add the type `StateProps`.

https://github.com/JedWatson/react-select/blob/master/src/stateManager.js#L53

https://github.com/JedWatson/react-select/blob/master/src/stateManager.js#L25-L36

claasahl (changes requested):

types/react-select/lib/stateManager.d.ts
```diff
@@ -1,25 +1,28 @@
 import { Component, ComponentType, Ref as ElementRef } from 'react';
 
+import SelectBase, { Props as SelectProps } from './Select';
 import { ActionMeta, InputActionMeta, ValueType } from './types';
 
-export interface Props<OptionType> {
+export interface DefaultProps<OptionType> {
```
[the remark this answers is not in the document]
pcorpet: Done.

types/react-select/lib/stateManager.d.ts
pcorpet: Done as well.

pcorpet:
@claasahl can you take another look please?

claasahl:
@pcorpet thanks for the reminder!

claasahl (approved):
thanks for your contribution, @pcorpet

PR #35405, opened 2019-05-13
[Rebass] Update rebass base props to facilitate use of string or css function

jbarzegar:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://www.styled-components.com/docs/api#css-prop>>

Hello! I noticed rebass complained whenever a string is used inside its css prop, however as the [styled-components docs show](https://www.styled-components.com/docs/api#css-prop) you can use a string/template literal.

### DIFF
--- a/types/rebass/index.d.ts
+++ b/types/rebass/index.d.ts
@@ -18,7 +18,10 @@ type Omit<T, K> = Pick<T, Exclude<keyof T, K>>;
 
 export interface BaseProps extends React.Props<any> {
     as?: React.ReactType;
-    css?: StyledComponents.CSSObject;
+    css?:
+        | StyledComponents.CSSObject
+        | StyledComponents.BaseThemedCssFunction<any>
+        | string;
 }
 
 interface BoxKnownProps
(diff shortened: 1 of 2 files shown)


angusfretwell (approved):
Good catch!

rhysd (approved):

types/rebass/index.d.ts
```diff
@@ -18,7 +18,10 @@ type Omit<T, K> = Pick<T, Exclude<keyof T, K>>;
 
 export interface BaseProps extends React.Props<any> {
     as?: React.ReactType;
-    css?: StyledComponents.CSSObject;
+    css?:
+        | StyledComponents.CSSObject
+        | StyledComponents.BaseThemedCssFunction<any>
```
rhysd: Though it is not mandatory, adding test case for `StyledComponents.BaseThemedCssFunction` would make this patch even better.

jbarzegar: I'll add a test case right now.

jbarzegar: Ah whoops I realized that `StyledComponents.BaseThemedCssFunction` isn't returned to the `css` prop but `StyledComponents.FlattenSimpleInterpolation` is.  Will fix

jbarzegar: Fixed

rhysd (approved):
Thank you. Looks good

PR #35406, opened 2019-05-13
[Auth0] Add explicit typing support for User and App Metadata

dancrumb:
This provides users with the ability to explicitly define the typing around their User and App metadata

### DIFF
--- a/types/auth0/index.d.ts
+++ b/types/auth0/index.d.ts
@@ -3,6 +3,7 @@
 // Definitions by: Seth Westphal <https://github.com/westy92>
 //                 Ian Howe <https://github.com/ianhowe76>
 //                 Alex Bjørlig <https://github.com/dauledk>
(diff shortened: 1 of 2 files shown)


types/auth0/index.d.ts
```diff
@@ -400,8 +401,8 @@ export interface User {
   created_at?: string;
   updated_at?: string;
   identities?: Identity[];
-  app_metadata?: AppMetadata;
-  user_metadata?: UserMetadata;
+  app_metadata: A;
```
dancrumb: The removal of the `?`s here is intentional: Auth0 will always return these.

alexbjorlig (approved):
I really love your solution with `UserData<A = AppMetadata, U=UserMetadata>`.

This is an awesome change, and something that I am looking forward to integrate into our usage of Auth0.

PR #35409, opened 2019-05-13
lodash: Chain conditional types ignore null | undefined

### LINKED ISSUE #35400: [@types/lodash] Property 'get' does not exist on type 'PrimitiveChain<undefined>
  - Authors: @aj-r
 
I have pretty simple chain expression:
```jsx
chain([{ type: 'axisX', dataType: 'Date' }])
  .find({ type: 'axisX' })
  .get('dataType')
  .value()
```

with `@types/lodash` **4.14.121** it works fine, but with **4.14.126** I get a type error:
```
Property 'get' does not exist on type 'PrimitiveChain<undefined> | ObjectChain<{ type: string; dataType: string; }>'.
  Property 'get' does not exist on type 'PrimitiveChain<undefined>'.
```

So it looks like typing issue, because `get` method will exist in this chain in any case (even if I pass null as chain argument)

sandersn:
since .get chains undefined in lodash. This is mainly needed to support chains like

```ts
_.chain([{ type: 'axisX', dataType: 'Date' }])
  .find({ type: 'axisX' })
  .get('dataType')
  .value()
```

with `--strict` on. I don't like not having to check for undefined here, but it is the way that lodash operates. Any opinions from the typings authors?

Fixes #35400


types/lodash/ts3.1/common/object.d.ts
```diff
@@ -1097,6 +1097,16 @@ declare module "../index" {
          */
         get<TDefault>(path: number, defaultValue: TDefault): T | TDefault;
     }
+    interface Primitive<T> {
+        /**
+         * @see _.get
+         */
+        get(path: PropertyPath): any;
```
aj-r: I'm worried about losing type safety on intersection chain types - if any of the intersected types include `Primitive<undefined>` then the return type becomes `any`, which is unexpected. 

Maybe this should return something like:

```ts
T extends null | undefined ? undefined : any;
```

aj-r (changes requested):

aj-r:
Actually I think there is a more fundamental issue here. In general, the lodash types permit null/undefined inputs, and the return type does not depend on whether the inputs are nullable. I think we've lost this behavior in the chaining though.

We could make the chain creation functions filter out null/undefined types, but I'm not sure how to handle ExpChain and ImpChain. We could do something like `T extends null|undefined ? never : ...` which would work most of the time, but would not handle pure null types properly.

sandersn:
I'll try `never`...pure `null` and `undefined` are (1) rare, (2) not useful, (3) even rarer as the result of a chain. 

There is one example in the lodash tests, which has `_.chain({ a: undefined }).get('a')`. I think that's an acceptable inaccuracy.

types/lodash/ts3.1/common/object.d.ts
sandersn: I just deleted the Primitive.get since it's not needed to fix original bug after altering the Chain types.

types/lodash/ts3.1/common/common.d.ts
```diff
@@ -7,15 +7,15 @@ declare module "../index" {
     type Many<T> = T | ReadonlyArray<T>;
     type ImpChain<T> =
         T extends { __trapAny: any } ? Collection<any> & Function<any> & Object<any> & Primitive<any> & String :
-        T extends null | undefined ? Primitive<T> :
+        T extends null | undefined ? never :
         T extends string | null | undefined ? String :
```
aj-r: Missed this before... does the `null|undefined` on this line actually do anything?

aj-r (approved):

types/lodash/ts3.1/common/common.d.ts
sandersn: Yes, it prevents null and undefined from producing String, since `null extends string | null | undefined`.

PR #35422, opened 2019-05-13
allow null on attrs for NodeType.create methods

indfnzo:
The ProseMirror docs [recommend passing null](https://prosemirror.net/docs/ref/#model.NodeType.create) to the NodeType.create, NodeType.createChecked, and NodeType.createAndFill methods to use the default attrs of the NodeType.

This pull request makes the typings allow nulls for these methods.

@bradleyayers @davidka @timjb @neknalb @patsimm


timjb (approved):
LGTM!

PR #35425, opened 2019-05-14
[react-tracking] Export additional interfaces

damassi:
This PR exports a few additional interfaces for use in [our wrapper implementation](https://github.com/artsy/reaction/blob/master/src/Artsy/Analytics/track.ts#L3) of `react-tracking`, which were lost when we moved the types over to DefinitivelyTyped.


types/react-tracking/test/react-tracking-with-types-tests.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { Track, track as _track, TrackingProp } from 'react-tracking';
+import { Track, track as _track, TrackingProp, Options, Decorator } from 'react-tracking';
```
alloy: This doesn’t really seem necessary when they are not being used, but adding the exports doesn’t really require additional test coverage imo.

alloy:
Thanks!

PR #35452, opened 2019-05-14
Improve cloneable-readable types

vaskevich:
Please fill in this template.

If adding a new definition:

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: <https://npmjs.com/package/cloneable-readable>

If removing a declaration:

### DIFF
--- a/types/cloneable-readable/cloneable-readable-tests.ts
+++ b/types/cloneable-readable/cloneable-readable-tests.ts
@@ -1,5 +1,5 @@
-import { PassThrough } from 'stream';
-import cloneable = require('cloneable-readable');
+import { PassThrough } from "stream";
+import cloneable = require("cloneable-readable");
 
 const ps = new PassThrough(); // $ExpectType PassThrough
 const cl = cloneable(ps); // $ExpectType Cloneable<PassThrough>
@@ -9,3 +9,7 @@ process.stdin.pipe(cl).pipe(process.stdout);
 
 cloneable.isCloneable(ps); // $ExpectType boolean
 cloneable.isCloneable(cl); // $ExpectType boolean
+
+if (cloneable.isCloneable(ps)) {
+    ps; // $ExpectType PassThrough & Readable & { clone(): Cloneable<Readable>; }
+}
--- a/types/cloneable-readable/cloneable-readable-tests.ts
+++ b/types/cloneable-readable/cloneable-readable-tests.ts
@@ -1,5 +1,5 @@
-import { PassThrough } from "stream";
-import cloneable = require("cloneable-readable");
+import { PassThrough } from 'stream';
+import cloneable = require('cloneable-readable');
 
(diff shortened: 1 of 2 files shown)


types/cloneable-readable/cloneable-readable-tests.ts
```diff
@@ -1,5 +1,5 @@
-import { PassThrough } from 'stream';
-import cloneable = require('cloneable-readable');
+import { PassThrough } from "stream";
+import cloneable = require("cloneable-readable");
```
qlonik: Let's keep the quotes in the same style they were. I don't think this matters, but it adds unnecessary churn in this commit

types/cloneable-readable/index.d.ts
```diff
@@ -1,17 +1,22 @@
-// Type definitions for cloneable-readable 1.1
+// Type definitions for cloneable-readable 2.0
 // Project: https://github.com/mcollina/cloneable-readable#readme
-// Definitions by: Nikita Volodin <https://github.com/qlonik>
+// Definitions by: Nikita Volodin <https://github.com/qlonik>, Oleg Vaskevich <https://github.com/vaskevich>
```
qlonik: Can you please place your name on the next line and align with other names?

types/cloneable-readable/index.d.ts
```diff
@@ -1,17 +1,22 @@
...
 /// <reference types="node"/>
 
-import { Readable } from 'stream';
+import { Readable } from "stream";
+
+declare namespace cloneable {
+    type Cloneable<T extends Readable> = T & { clone(): Cloneable<T> };
+}
 
-type Cloneable<T> = T & { clone(): Cloneable<T> };
```
qlonik: I'm not sure of the reasoning behind the change to a namespace? It seems to complicate the declaration and make type name longer. Could you elaborate please?

qlonik (changes requested):
Thank you for pull request!

It seems that the change is about bumping version and modifying `isCloneable` to be a type-guard. Could you please reduce this commit to only include changes related to those

types/cloneable-readable/index.d.ts
vaskevich: This is to export `Cloneable` so it can be used as a type externally. Unfortunately writing `export type Cloneable ...` doesn't work due to this error:

```
An export assignment cannot be used in a module with other exported elements.
```

and adding a namespace seems to be the workaround.

types/cloneable-readable/cloneable-readable-tests.ts
vaskevich: My linter auto-corrected this, but let me revert.

types/cloneable-readable/index.d.ts
vaskevich: Sure.

qlonik (approved):

PR #35462, opened 2019-05-15
fetch-mock: allow headers[name] to be set to number

gr2m:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.13

### DIFF
--- a/types/fetch-mock/index.d.ts
+++ b/types/fetch-mock/index.d.ts
@@ -82,7 +82,7 @@ declare namespace fetchMock {
         /**
          * Set the response headers.
          */
-        headers?: { [key: string]: string };
+        headers?: { [key: string]: string | number };
 
         /**
          * If this property is present then a Promise rejected with the value
@@ -141,7 +141,7 @@ declare namespace fetchMock {
         /**
          * key/value map of headers to match
          */
-        headers?: { [key: string]: string };
+        headers?: { [key: string]: string | number };
 
         /**
          * key/value map of query strings to match, in any order


QuentinBouygues (approved):
Looks good !

merrywhether:
Wait, headers are only strings (or string[]), as here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v11/http.d.ts#L7-L8.

Since there's no `SomeHeader: "1"`, you can't differentiate between number and string.

gr2m:
Content-Lenght header is set to numbers

merrywhether:
Right, it looks like a number: `Content-Length: 451`, but because there is no `Content-Length: "451"` it's impossible for receivers to discern a type, so everything is just string.

If you're treating it as a number, it's only working because JS is type-coercing for you.

From a node server:
```
console.log(req.headers['content-length'], typeof req.headers['content-length']);
> 451 string

**Comment by gr2m** [comment]
This updated resolve my problem: https://github.com/wolfy1339/request.js/commit/6fd805e62accc649f5a77f1f1b75fa8cc255350b

**Comment by merrywhether** [comment]
Right, but that code is incorrect. The response headers should be `content-length: "0"` because that's what would be returned.

You can test this in the browser as well:
```
new Headers({ "content-length": 0 }).get("content-length")
> "0"
```

gr2m:
Ah, so you mean this one [`MockResponseObject#headers[key]`](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35462/files#diff-7c2e90d9e6e61ccd40556963293406deR85) will always be a `string`?

I can make a follow up PR to fix that

merrywhether:
Yeah, looking at a http://www.wheresrhys.co.uk/fetch-mock/#usagecustom-classes more closely, it looks like you're right that the mock function should accept numbers but will always return strings.

PR #35464, opened 2019-05-15
[log-process-errors] Add type definitions for 'log-process-errors'

JorgenVatle:
Added type definitions for [`log-process-errors`](https://github.com/ehmicky/log-process-errors) - an overhaul to Node.js' default error handling.


types/log-process-errors/index.d.ts
```diff
@@ -0,0 +1,64 @@
...
+     */
+    colors?: boolean;
+}
+
+/**
+ * Log Process Errors
+ *
+ * @link https://github.com/ehmicky/log-process-errors/blob/master/docs/API.md#logprocesserrorsoptions
+ */
+export default function logProcessErrors(options?: LogProcessErrorsOptions): () => void;
```
sandersn: I don't think `logProcessErrors` is actually a default export; 
`require('log-process-errors').default` is undefined. It should be 
```ts
declare function logProcessErrors(...
declare namespace logProcessErrors {
  export type LogLevel = ...
  export type ExceptionType = ...
  ...
}
export = logProcessErrors;
```

and then imported with `import LogProcessErrors = require('log-process-errors')

sandersn (changes requested):
Need to use commonjs export, not ES2015 default export.

types/log-process-errors/index.d.ts
JorgenVatle: Oops! I'll correct that right away. 😅

sandersn (approved):

Issue #35475, opened 2019-05-15
With NodeJS v10 and TSC v3.4.5 @types/mongo throws error with this library


riginoommen:
Line Number
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1957e4cb6a5ed54ca310f015e5a61411bb713da7/types/mongodb/index.d.ts#L36

Error Log
```bash
riginoommen-mac:portal-server riginoommen$ rm -rf dist/ && grunt build && grunt serve
Running "newer:tslint:all" (newer) task
No newer files to process.

Running "copy:build" (copy) task
Created 3 directories, copied 5 files

Running "ts:build" (ts) task
Compiling...
Using tsc v3.4.5
node_modules/@types/mongodb/index.d.ts(36,10): error TS2305: Module '"tls"' has no exported member 'checkServerIdentity'.

>> 1 non-emit-preventing type warning
>> Error: tsc return code: 2
Warning: Task "ts:build" failed. Use --force to continue.

Aborted due to warnings.
Aborted due to warnings.
riginoommen-mac:portal-server riginoommen$ node -v
v10.15.3
```

CaselIT:
Your issue seems strange.
Node v10 has that method
https://nodejs.org/docs/latest-v10.x/api/tls.html#tls_tls_checkserveridentity_hostname_cert

Maybe try to update the node typing?

riginoommen:
@CaselIT Yes i updated the node typing and this got foxed thanks for suggesting

niyazhussain:
@riginoommen  i  used npm install @types/node  -g still no luck .

riginoommen:
@niyazhussain can you share the ```package.json``` here

niyazhussain:
{
  "name": "SportyApi",
  "version": "0.1.0",
  "repository": {
    "type": "git",
  },
  "scripts": {
    "start": "npm run serve",
    "build": "npm run build-ts && npm run tslint  && npm run copy-static-assets",
    "serve": "node dist/server.js",
    "watch-node": "nodemon dist/server.js",
    "watch": "concurrently -k -p \"[{name}]\"  \"npm run watch-ts\" \"npm run watch-node\"",
    "test": "jest --forceExit --coverage --verbose",
    "build-ts": "tsc",
    "watch-ts": "tsc -w",
    "tslint": "tslint -c tslint.json -p tsconfig.json",
    "debug": "npm run build && npm run watch-debug",
[...]

PR #35478, opened 2019-05-15
#35475 - With reference to the new node version

riginoommen:
Issue - #35475


CaselIT (changes requested):
Removing the attribute is not the solution...

CaselIT:
You issue seems strange.
Node v10 has that method
https://nodejs.org/docs/latest-v10.x/api/tls.html#tls_tls_checkserveridentity_hostname_cert

Maybe try to update the node typing?

PR #35479, opened 2019-05-15
[vis] Added widthConstraint options to NodeOptions

slavede:
Please fill in this template.

Also fixed previous lint issues

http://visjs.org/docs/network/nodes.html
NodeOptions should have widthConstraint option.

### DIFF
--- a/types/vis/vis-tests.ts
+++ b/types/vis/vis-tests.ts
@@ -1,4 +1,6 @@
-import { NodeOptions } from './index';
+import { NodeOptions } from 'vis';
+import vis from 'vis';
+
 // Test DataSet constructor
 new vis.DataSet();
 new vis.DataSet({});
--- a/types/vis/vis-tests.ts
+++ b/types/vis/vis-tests.ts
@@ -1,3 +1,4 @@
+import { NodeOptions } from './index';
 // Test DataSet constructor
 new vis.DataSet();
 new vis.DataSet({});
@@ -217,3 +218,29 @@ options = {
 };
 
 network.setOptions(options);
+
+//
+// should accept different formats of widthConstraint for NodeOptions
+//
+let nodeWidthConstraintOptions: NodeOptions = {
+    widthConstraint: {
+        maximum: 100
+    }
+}
+nodeWidthConstraintOptions = {
+    widthConstraint: false
+}
+nodeWidthConstraintOptions = {
+    widthConstraint: {
+        minimum: 1,
+        maximum: 5
+    }
+}
+nodeWidthConstraintOptions = {
+    widthConstraint: {
+        minimum: 1
+    }
+}
+nodeWidthConstraintOptions = {
+    widthConstraint: 150
+}
--- a/types/vis/vis-tests.ts
+++ b/types/vis/vis-tests.ts
@@ -1,5 +1,5 @@
 import { NodeOptions } from 'vis';
-import vis from 'vis';
+import vis = require('vis');
 
 // Test DataSet constructor
 new vis.DataSet();
@@ -37,7 +37,7 @@ data.add([
 ]);
 
 // subscribe to any change in the DataSet
-data.on('*', (event, properties, senderId) => {
+data.on('*', (event: any, properties: any, senderId: any) => {
   console.log('event', event, properties);
 });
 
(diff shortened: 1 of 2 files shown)


types/vis/vis-tests.ts
```diff
@@ -34,7 +37,7 @@ data.add([
 ]);
 
 // subscribe to any change in the DataSet
-data.on('*', (event, properties, senderId) => {
+data.on('*', (event: any, properties: any, senderId: any) => {
```
adripanico: do not use any

types/vis/vis-tests.ts
```diff
@@ -54,7 +57,7 @@ console.log('item1', item1);
 
 // retrieve a filtered subset of the data
 let items = data.get({
-  filter: (item) => {
+  filter: (item: any) => {
```
adripanico: do not use any

types/vis/vis-tests.ts
```diff
@@ -132,14 +135,14 @@ const dataset = new vis.DataSet<TestData>();
...
     return (item.group === 2);
   }
 });
 
 // retrieve all items having a property balance with a value above zero
 const positiveBalance = dataset.get({
-  filter: (item) => {
+  filter: (item: any) => {
```
adripanico: do not use any

types/vis/vis-tests.ts
```diff
@@ -77,7 +80,7 @@ console.log('formatted items', items);
 data = new vis.DataSet<TestData>();
 
 // subscribe to any change in the DataSet
-data.on('*', (event, properties, senderId) => {
+data.on('*', (event: any, properties: any, senderId: any) => {
```
adripanico: do not use any

types/vis/vis-tests.ts
```diff
@@ -132,14 +135,14 @@ const dataset = new vis.DataSet<TestData>();
 
 // retrieve all items having a property group with value 2
 const group2 = dataset.get({
-  filter: (item) => {
+  filter: (item: any) => {
```
adripanico: do not use any

adripanico (changes requested):

types/vis/vis-tests.ts
slavede: This in fact is not related with this pull request, but hooks (which have rules newer than latest vis change) are failing because implicit type warning. This is just to make hooks happy, I really dont think this should be a blocker for this PR because in fact i dont know what should be there.

types/vis/vis-tests.ts
slavede: This in fact is not related with this pull request, but hooks (which have rules newer than latest vis change) are failing because implicit type warning. This is just to make hooks happy, I really dont think this should be a blocker for this PR because in fact i dont know what should be there.

types/vis/vis-tests.ts
slavede: This in fact is not related with this pull request, but hooks (which have rules newer than latest vis change) are failing because implicit type warning. This is just to make hooks happy, I really dont think this should be a blocker for this PR because in fact i dont know what should be there.

types/vis/vis-tests.ts
slavede: This in fact is not related with this pull request, but hooks (which have rules newer than latest vis change) are failing because implicit type warning. This is just to make hooks happy, I really dont think this should be a blocker for this PR because in fact i dont know what should be there.

types/vis/vis-tests.ts
slavede: This in fact is not related with this pull request, but hooks (which have rules newer than latest vis change) are failing because implicit type warning. This is just to make hooks happy, I really dont think this should be a blocker for this PR because in fact i dont know what should be there.

types/vis/vis-tests.ts
adripanico: Revert the changes then.

slavede: The thing is that then hooks will fail, and it won't be able to merge even this.

Hooks fails because you shouldn't use implicit detection. This solves it just enough to pass the hooks.

slavede: We can open a separate ticket (even without this PR coming in) that types should be handled here. Because from now on, every PR hook will fail because previously wasn't done properly, and in the meanwhile definitely typed introduced different hooks

slavede:
@adripanico can we resolve this PR then? If your comments are applied, still the build would fail. That's really separated issue which I've just applied workaround temporary so the rest of the contributors can fix other parts. This really blocks merging of really needed fix.

Issue #34080, opened 2019-03-21
@types/jasmine: False compile error when methods are overloaded


fknittel:
Since #33860 we get false compile errors when spying on overloaded methods.

Minimal example:
```TypeScript
interface I {
  f(): string;
  f(x: any): number;
}

const spyObject = jasmine.createSpyObj<I>("spyObject", ["f"]);
spyObject.f.and.returnValue("a string, erroneously required to be a number");
```
Results in:
```
TS2345: Argument of type '"a string, erroneously required to be a number"' is not assignable to parameter of type 'number'.
```

Our expectation: The accepted parameter type of `returnValue` should be `string | number`, not only `number`.

(Unfortunately the problematic method overloading (as demonstrated in the `I` example interface) occurs in external libraries we use, so we can't change that part.)

CC PR author: @kolodny

kolodny:
Hmm, this seems to be a typescript error. I've created a PoC that highlights this:

[...]

PR #35483, opened 2019-05-15
[jasmine] Allow calling spyOn for class constructors and add AnyMethods type to jasmine namespace

kolodny:
This PR includes a fix to allow calling `spyOn` with a class constructor as well as an `AnyMethods` type which allows opting out of the inferring that's done for `createSpyObj`. This should unblock #34080 for overloaded functions

- ~Provide a URL to documentation or source code which provides context for the suggested changes: <<url here>>~
- ~Increase the version number in the header if appropriate.~

### DIFF
--- a/types/jasmine/index.d.ts
+++ b/types/jasmine/index.d.ts
@@ -11,6 +11,7 @@
 //                 Yaroslav Admin <https://github.com/devoto13>
 //                 Domas Trijonis <https://github.com/fdim>
 //                 Peter Safranek <https://github.com/pe8ter>
+//                 Moshe Kolodny <https://github.com/kolodny>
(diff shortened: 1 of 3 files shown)


pe8ter:
These changes still don't work with our project. Instead of rolling forward with more changes, I'd prefer to revert `3.3.12` because if this code doesn't compile, then something's wrong:

```typescript
const kinesisSpy = spyOn(AWS, 'Kinesis');
```

kolodny:
@pe8ter There's always the option of doing `const kinesisSpy = spyOn<any>(AWS, 'Kinesis');` for that case. Either way I've included the fix for using `spyOn` with class constructors in this PR and will update the description to reflect that

PR #35489, opened 2019-05-15
Fix, add types to Inquirer.js

jednano:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/SBoudrias/Inquirer.js/pull/55>>


types/inquirer/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for Inquirer.js 6.x
+// Type definitions for Inquirer.js 6.3.x
```
jedmao: I freshly copy/pasted all the examples from [inquirer/examples](https://github.com/SBoudrias/Inquirer.js/tree/master/packages/inquirer/examples).

jedmao:
@vvakame @paulvanbrenk can either of you give this PR a look? It says it's in review, but it's not (yet). That was my own review comments.

sandersn:
Actually, the bot got pretty confused on this one. It missed the owners of all these existing packages:

console-ui: @dfreeman
inquirer-npm-name:  @manuth
inquirer: @tkQubo, @ppathan, @jouderianjr, @bang88, @bitjson, @synarque, @jrockwood, @kwkelly, @chigix,
yeoman-environment: @bolasblack
yeoman-generator: @armorik83, @janslow, @ikatyang, @tasadar2, @haggen, @chigix, @misterdev,
yeoman-test: @ikatyang

jedmao:
@sandersn does this require any more additional work on my end? I'm not understanding the bot thing. Did I break something there or is that a separate issue?

sandersn:
The bot tagged this as "new definition" because it contains a new definition, but it also edits a lot of existing packages. I manually pinged the authors of those packages so they'll have a chance to review the changes. In particular, inquirer is a very popular package so those changes need to be reviewed before merging.

jedmao:
Thanks for the clarification. For those reviewing, know that I've copied all the latest documentation into the TSDoc comments and the latest examples from Inquirer into the tests, so there should be great confidence that these changes are good.

manuth:
The changes made to the inquirer-npm-name are just fine.
As I didn't ever contribute to the type-declarations of inquirer I'll wait for the others to make a review for a day or two.

bolasblack:
The changes made to yeoman-environment are fine.

In fact, this package has not been made any real changes. (I think)

types/inquirer/index.d.ts
```diff
@@ -288,7 +358,7 @@ declare namespace inquirer {
             extra?: any;
             key?: string;
             checked?: boolean;
-            disabled?: string | (<T = Answers>(answers: T) => any);
+            disabled?: string | (<A>(answers: A) => any);
```
manuth: Please consider to add a generic parameter to `ChoiceOption`, `Choice`, `Choices` and `ChoiceType` rather than adding a generic parameter to a callbacks and constructors.
This would make it ways better compatible with `QuestionOptions<A>` and more comfortable to use, I think.

types/yeoman-generator/index.d.ts
```diff
@@ -17,14 +17,30 @@ import { Observable } from 'rxjs';
...
-    interface Question extends inquirer.Question {
+    interface GeneratorQuestion {
         /**
          * whether to store the user's previous answer
          */
         store?: boolean;
     }
-    type Questions = Question | Question[] | Observable<Question>;
     type Answers = inquirer.Answers;
+    type Question<A extends Answers = Answers> = (
```
manuth: I'd rather recommend you to use `GeneratorQuestion<A> extends inquirer.Question<A>` as this would cause the `Generator.Question`-type to look exactly the same.

neurolag (changes requested):
So 2 days passed - thats why I decided to make a review.
There is just one or two small changes I recommend you to do.

However this PR looks just fine to me. Awesome job!

types/yeoman-generator/index.d.ts
jedmao: Yes, that would be ideal; however, it's not possible because of the union type.
```
An interface can only extend an object type or intersection of object types with statically
known members.ts(2312)
```
This is doable:
```ts
type Question<A extends Answers = Answers> = inquirer.Question<A>;
```
But this is not:
```ts
interface Question<A extends Answers = Answers> extends inquirer.Question<A> {}
```

jedmao: Actually, I have an idea...

types/inquirer/index.d.ts
jedmao: A bit out of scope, but I'll do it.

jedmao:
@manuth, I've addressed your CR in 946c986. Thanks!

types/inquirer/index.d.ts
```diff
@@ -70,90 +72,158 @@ declare namespace inquirer {
...
+        transformer?(input: string, answers: A, flags: any): string;
+        /**
+         * Change the number of lines that will be rendered when using `list`, `rawList`,
+         * `expand` or `checkbox`.
+         */
+        pageSize?: number;
+    }
+
+    type Question<A extends Answers = Answers, T extends Record<string, any> = {}> = (
+        ListQuestion<A> & T |
```
manuth: Oooh I see what you planned to do
But I think another good option would be to drop the additional generic parameter and use
```ts
type GeneratorQuestion = Question<A> & {
    /** comment stuff */
    store?: boolean
}
```
in the yeoman-generator instead.
What do you think about this idea?

neurolag (changes requested):
Awesome! thanks for your rapid commit.
You might want to check out the one comment I made.

Your changes look very good to me!

types/inquirer/index.d.ts
jedmao: Yeah, I like that much better. See dc59344!

neurolag (approved):
Awesome, we're done!
Thanks for taking my advices into consideration :+1:

jedmao:
Thanks @manuth, can you please officially approve the PR so it can go to the next step?

manuth:
I just did so some minutes ago - the DT Bot is a little bit laggy sometimes 😅

neurolag (approved):
As I've had a close look through the whole changes I'm now approving these changes in place of the authors of inquirer and yeoman-generator.

Hope you guys don't mind ✨

jedmao:
@sandersn what are the next steps at this point? We don't have to wait for every single party to be tagged, do we? Some of these people were tagged just for a TS version change in their definitions.

@vvakame, you still around these days?

sandersn:
@jedmao Inquirer is the blocker here since it's the popular package and has extensive changes. I want to get sign off from one of @tkqubo, @ppathan, @jouderianjr, @bang88, @bitjson, @synarque, @jrockwood, @kwkelly, @chigix

jedmao:
@tkqubo can you take a look? Too many people were tagged and I'm afraid it's going to cause the [bystander effect](https://en.wikipedia.org/wiki/Bystander_effect).

types/inquirer/index.d.ts
```diff
@@ -9,21 +9,23 @@
...
 
 declare namespace inquirer {
     type Prompts = { [name: string]: prompts.Base };
-    type ChoiceType = string | objects.ChoiceOption | objects.Separator;
-    type Questions<T = Answers> =
-        | Question<T>
-        | ReadonlyArray<Question<T>>
-        | Observable<Question<T>>;
+    type ChoiceType<A> = string | objects.ChoiceOption<A> | objects.Separator;
+    type Questions<A extends Answers = Answers> =
```
chigix: Why not type it as `Questions<A extends Answers>` directly?

types/inquirer/index.d.ts
```diff
@@ -70,90 +71,134 @@ declare namespace inquirer {
...
-         * Default value(s) to use if nothing is entered, or a function that returns the default value(s).
-         * If defined as a function, the first parameter will be the current inquirer session answers.
+         * Receive the current user answers hash and should return `true` or `false` depending
+         * on whether or not this question should be asked. The value can also be a simple boolean.
          */
-        default?: any | ((answers: T) => any) | ((answers: T) => Promise<any>);
+        when?: boolean | ((answers: A) => boolean | Promise<boolean>);
+    }
+
+    interface QuestionOptions<A> {
```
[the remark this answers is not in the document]
chigix: Although specific type string is defined and given in specific Question interfaces, still not understand why not state `type` property existing in the generic `QuestionCommon` interface

types/inquirer/index.d.ts
jedmao: Just wanted to provide a default so you can say `Questions` w/o having to write `Questions<A>`, but I see now how that doesn't make a lot of sense, because nobody would have empty answers, right? Is that your logic?

jedmao: Anyway, the default of `= Answers` was pre-existing, so maybe we'd do well to leave it alone?

types/inquirer/index.d.ts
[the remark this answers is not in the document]
jedmao: The `QuestionCommon` interface is only for props that are shared between ALL question types. This `QuestionOptions` interface is only meant to be picked from, (e.g., `Pick<T, 'choices'>`). The reason I wrote it this way is so that these props only have to be written once, but they can be picked for relevant question types that need them, complete with the TSDoc comments.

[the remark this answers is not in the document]
jedmao: The props in this interface are used across more than one question type, but not all question types. Does that make sense?

types/inquirer/index.d.ts
manuth: Using `Answers` as generic parameter by default makes it easier do write "quick'n'dirty" code - it might also be helpful if you extend an untyped module which asks questions using inquirer.

You might want to fetch the answers even if you don't really know what answers you might expect.

types/inquirer/index.d.ts
```diff
@@ -70,90 +72,158 @@ declare namespace inquirer {
...
-         *      <li>confirm</li>
-         *      <li>list</li>
-         *      <li>rawlist</li>
-         *      <li>password</li>
-         * </ul>
-         * @defaults: 'input'
+         * The name to use when storing the answer in the answers hash.
+         * If the name contains periods, it will define a path in the answers hash.
          */
-        type?: string;
```
chigix: I'm sorry for previous comment [at a wrong position](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35489/files#r285837466). Actually, what I did not understand is why not keep `type: string`  as a placeholder here and then the specific `type`s given in each detailed question interfaces could be viewed as override(or implement) this `type` property. Because `type` property is exactly shared by all question types excepting the values are various but all are strings.

manuth: In inquirer you can optionally add and use custom question-types.
Having this in mind a type-property in the base class seems reasonable to me

What do you guys think?

types/inquirer/index.d.ts
chigix: I got it perfectly now. Thanks sincerely @jedmao @manuth

types/inquirer/index.d.ts
jedmao: I think if you have custom question types then you should define them in a type, just as yeoman-generator ads a `store` prop. If you want something more loose than that, then you can define a looser type too. I'll give this some more thought and look for documentation, because it's nowhere in the examples.

RyanCavanaugh:
@chigix @jedmao do you both think this is ready to merge?

types/inquirer/index.d.ts
jedmao: @manuth it looks like someone tried to introduce custom prompt types, but the issue and PR were closed. See https://github.com/SBoudrias/Inquirer.js/pull/238

jedmao:
@RyanCavanaugh I'm good on my end!

types/inquirer/index.d.ts
manuth: Sorry - my bad
What I was talking about was custom prompts:
https://github.com/SBoudrias/Inquirer.js/blob/master/README.md#inquirerregisterpromptname-prompt

People can create their own prompts and they are accessible by setting their `name` in a question's `type`.

jedmao: Also, fundamentally, I don't think this suggestion offers anything at all, because it's overridden in each specific question type. Take this TS for example:
```ts
interface Base {
	type: string
}

interface Foo extends Base {
	type: 'foo'
}

const subject: Foo = { type: 'bar' }

// Type '"bar"' is not assignable to type '"foo"'.
// (property) Foo.type: "foo"
```
Because the `Base` interface is only referenced by the specific question types and not directly.

manuth: I totally agree - let's wait for @chigix's feedback.

jedmao: @manuth, I would refer back to [this comment](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35489#discussion_r286014246) concerning what to do about that particular use case. People can pass-in anything, so long as it extends `inquirer.Answers`. Gotta' love generics!

manuth: I hope I get you right - my English's pretty messy sometimes.
On one hand declaring `type: string` gives people writing their own prompts a plan on what to do about the `type`-property (which is: declaring an own `name` to access it using the `Question`s `type`-property).

On the other hand people could set a `type` that doesn't even exist as long as they're using the `QuestionCommon<A>`-interface rather than some more specific interface.

Though... to me the `QuestionCommon<A>` interface _must_ have a `type`-member as you _must_ set the `type` no matter what kind of prompt you're trying to show, right...?

jedmao: No, actually. The `type` [defaults to "input"](https://github.com/SBoudrias/Inquirer.js#question), which why I have it written this way: 

```ts
    interface InputQuestion<A> extends QuestionCommon<A>,
        Pick<QuestionOptions<A>, 'filter' | 'transformer'> {
        type?: 'input'
```
This is the only question type that has an optional `type` prop, because if the user doesn't supply a `type` an `InputQuestion` is assumed. It's not a required prop, but it does have a default value of "input".

Defining these specific types prevents the developer from making mistakes like:
1. Using a type that doesn't exist.
1. Adding props that are not supported for the specified type.

The current type definitions for Inquirer are too loose and do not prevent these things from happening, which is why I thought it was important to fix in this PR.

manuth: Aaah - so now I get your point.
Do you think there is any other change required in order to make the usage of `inquirer` more smoothly?

jedmao: Yes, I do, but this PR is already a bit much and I'd like to seal it off at this point. I might be adding another (much smaller) PR in the near future.

But yes, I've run into some hiccups during my Yeoman generator development that could be much smoother.

jedmao:
@chigix I think we're all just waiting on you at this point. Can you please approve the PR if you have nothing else to add? Thanks!

chigix:
The changes to the inquirer js are fine

jedmao:
@RyanCavanaugh we're good to go!

PR #35490, opened 2019-05-15
IChartistSeriesData.data is not only an array of number

pierre-fromager:
I changed an existing definition for Chartist as IChartistSeriesData interface, prop data:

according to this gist http://jsbin.com/lokozu/edit?html,css,js,output
following this issue https://github.com/gionkunz/chartist-js/issues/1076

We should be able to use an object with x/y props either typed as number/number or Date/number not only an array of number.


types/chartist/index.d.ts
```diff
@@ -99,7 +99,7 @@ declare namespace Chartist {
     interface IChartistSeriesData {
         name?: string;
         value?: number;
-        data?: Array<number>;
+        data?: Array<number> | Array<{ x: number, y: number }> | Array<{ x: Date, y: number }>;
```
mtgibbs: I'd say that we could probably simplify this down to 

```ts
data?: Array<number> | Array<{ x: number | Date, y: number}>;
```

It might make it easier to figure out when the compiler sees you made a mistake and does that thing where it says "Can not cast {x} to {x} | {y} | {z} | null | undefined | {}"

pierre-fromager: Agreed, i've just refacto as you suggested.

mtgibbs: Cool. 👍

pierre-fromager: Thanks Matt for this quick and pleasant review, 
I would propose a test case later, but for now i work on urgent project using timestamp on x-axis and must hurry up to deliver it in time.
Do you think the build will be done quickly on npm side "@types/chartist": "^0.9.45" ?

pierre-fromager: Unit test added

huchim:
I have this error with the package update.

**Error:**
```
Cannot invoke an expression whose type lacks a call signature. Type '((...items: number[]) =>
number) | ((...items: { x: number | Date; y: number; }[]) => number)' has no compatible call 
signatures.ts(2349)
```
**Code:**
``` typescript
spentValues.data.push(null);
spentValues.data.push(null);
```

pierre-fromager:
What about force casting <number> or <Date> as below :
spentValues.data.push(<number>0);
spentValues.data.push(<number>null);

PR #35497, opened 2019-05-15
[@types/lodash] _.chain().get() shorthand for empty array default

schmod:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://lodash.com/docs/4.17.11#get

---

## Original issue:

If I use `_.chain(obj).get('prop', []).value()` to grab an array off of an object (while providing an empty array as a fallback), the TypeScript compiler will throw an error when using the current types (v4.14.129).

This issue does not occur if I do _not_ provide a default value, or if I explicitly cast the empty array to the correct type (ie. `.get('prop', [] as number[])`). 

This is a regression from `@types/lodash` v4.14.123

**Note**: My fix is fairly limited in scope (and, TBH, feels kind of messy) -- however, I wasn't able to come up with any similar scenarios.  If I'm missing something, and this is indicative of a larger issue, please feel free to flag it!

---

### This works
```ts
interface Employee {
  phone?: number
}
interface Employees {
  employees?: Employee[]
}
const all:Employees = { employees: [{ phone: 1 }, {}, {}] };
const phoneUsers = _.chain(all)
    .get('employees')
    .filter((e) => !!e.phone)
    .value();
```

### This does not work
```ts
// same interfaces as above
const phoneUsers = _.chain(all)
    .get('employees', [])
    .filter((e) => !!e.phone) // tsc error
    .value();
```

`.get('employees', [] as Employee[])` _does_ work (but shouldn't be necessary).

This second example worked in `@types/lodash v4.14.123`.

### DIFF
--- a/types/lodash/ts3.1/common/object.d.ts
+++ b/types/lodash/ts3.1/common/object.d.ts
@@ -1125,7 +1125,7 @@ declare module "../index" {
         /**
          * @see _.get
          */
-        get<TKey extends keyof T, TDefault>(path: TKey | [TKey], defaultValue: []): T[TKey] extends Array<any> ? ExpChain<Exclude<T[TKey], undefined>> : never;
+        get<TKey extends keyof T>(path: TKey | [TKey], defaultValue: []): T[TKey] extends any[] ? ExpChain<Exclude<T[TKey], undefined>> : ExpChain<Exclude<T[TKey], undefined> | []>;
         /**
          * @see _.get
          */
--- a/types/lodash/ts3.1/common/object.d.ts
+++ b/types/lodash/ts3.1/common/object.d.ts
@@ -1125,7 +1125,7 @@ declare module "../index" {
         /**
          * @see _.get
          */
-        get<TKey extends keyof T>(path: TKey | [TKey], defaultValue: []): T[TKey] extends any[] ? ExpChain<Exclude<T[TKey], undefined>> : ExpChain<Exclude<T[TKey], undefined> | []>;
+        get<TKey extends keyof T>(path: TKey | [TKey], defaultValue: never[]): T[TKey] extends any[] ? ExpChain<Exclude<T[TKey], undefined>> : ExpChain<Exclude<T[TKey], undefined> | []>;
         /**
          * @see _.get
          */
--- a/types/lodash/ts3.1/common/object.d.ts
+++ b/types/lodash/ts3.1/common/object.d.ts
@@ -1125,7 +1125,7 @@ declare module "../index" {
         /**
          * @see _.get
          */
-        get<TKey extends keyof T>(path: TKey | [TKey], defaultValue: never[]): T[TKey] extends any[] ? ExpChain<Exclude<T[TKey], undefined>> : ExpChain<Exclude<T[TKey], undefined> | []>;
+        get<TKey extends keyof T>(path: TKey | [TKey], defaultValue: never[]): T[TKey] extends any[] ? ExpChain<Exclude<T[TKey], undefined>> : ExpChain<Exclude<T[TKey], undefined> | never[]>;
         /**
          * @see _.get
          */
--- a/types/lodash/ts3.1/common/object.d.ts
+++ b/types/lodash/ts3.1/common/object.d.ts
@@ -1125,6 +1125,10 @@ declare module "../index" {
         /**
          * @see _.get
          */
+        get<TKey extends keyof T, TDefault>(path: TKey | [TKey], defaultValue: []): T[TKey] extends Array<any> ? ExpChain<Exclude<T[TKey], undefined>> : never;
+        /**
+         * @see _.get
+         */
         get<TKey extends keyof T, TDefault>(path: TKey | [TKey], defaultValue: TDefault): ExpChain<Exclude<T[TKey], undefined> | TDefault>;
         /**
          * @see _.get
--- a/types/lodash/ts3.1/lodash-tests.ts
+++ b/types/lodash/ts3.1/lodash-tests.ts
@@ -5201,7 +5201,7 @@ fp.now(); // $ExpectType number
     _.chain({ a: undefined }).get("a"); // $ExpectType never
     _.chain({ a: value }).get("a", defaultValue); // $ExpectType StringChain | PrimitiveChain<false> | PrimitiveChain<true>
     _.chain({ a: undefined }).get("a", defaultValue); // $ExpectType PrimitiveChain<false> | PrimitiveChain<true>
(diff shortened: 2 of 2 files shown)


schmod:
@sandersn may also want to look at this, as he recently completed a huge refactor of these typings in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35241

types/lodash/ts3.1/common/object.d.ts
```diff
@@ -1122,6 +1122,10 @@ declare module "../index" {
          * @see _.get
          */
         get<TKey extends keyof T>(path: TKey | [TKey]): ExpChain<T[TKey]>;
+        /**
+         * @see _.get
+         */
+        get<TKey extends keyof T>(path: TKey | [TKey], defaultValue: []): T[TKey] extends any[] ? ExpChain<Exclude<T[TKey], undefined>> : ExpChain<Exclude<T[TKey], undefined> | []>;
```
sandersn: 1. You should use `never[]` instead of `[]` since that's the type of `[]`.
2. I don't understand how the conditional type helps. Why do you need to exclude `[]`?

sandersn (changes requested):

types/lodash/ts3.1/lodash-tests.ts
```diff
@@ -5201,6 +5201,7 @@ fp.now(); // $ExpectType number
     _.chain({ a: undefined }).get("a"); // $ExpectType never
     _.chain({ a: value }).get("a", defaultValue); // $ExpectType StringChain | PrimitiveChain<false> | PrimitiveChain<true>
     _.chain({ a: undefined }).get("a", defaultValue); // $ExpectType PrimitiveChain<false> | PrimitiveChain<true>
+    _.chain({ a: [1] }).get("a", []); // $ExpectType CollectionChain<number>
```
aj-r: If we support this use case for chaining, then we should also support it for the other variations of `get`, so add tests for those as well, e.g.

```ts
_.get({ a: [1] }, "a", []);
_({ a: [1] }).get("a", []);
```

aj-r (changes requested):

types/lodash/ts3.1/lodash-tests.ts
schmod: I'll add these tests, with the caveat that I have _not_ added this "workaround" to the types for `_.get()` and `_().get`.  Both of those signatures will continue to resolve to `number[] | never[]`.  

This PR doesn't actually change that behavior -- in my experience thus far, this appears to be far less of a problem in practice, as TypeScript appears to handle unions with `never[]` in a _mostly_-pragmatic fashion (ie. `.filter` predicates work like you'd expect them to, but `.map`'s argument gets widened to `any` for some reason)

If we to revisit this, I can add that to this PR (but I'd like to get some input from one of the maintainers first!). 

---

On the other hand, our original/underlying issue seems to be that `ExpChain<any[] | never[]>` expands to `CollectionChain<any> | CollectionChain<never>`, which is effectively an "impossible" type, as the predicates for many of the chained methods expand into something like `(val: any) => boolean | (val: never) => boolean` which can never be satisfied.

The tests don't really cover this scenario (ie. that one can actually chain multiple operations off of a `_.chain`), and I'm not really sure where to draw the line for providing test-coverage for this edge-case, as we're not going to reasonably be able to cover all possible combinations.  My test ensures that we don't end up with `CollectionChain<never>` in a union in this one narrow scenario, which seemed like a good-enough start.

---

Now that I've dug in a little further while writing this comment, I'm wondering if it makes more sense to remove my `_.chain().get()` workaround, and provide a more-explicit guards against ExpChain ever creating a `CollectionChain<never>`.

We'd change this:

```ts
type ExpChain<T> =
  // omitted a few lines...
  T extends List<infer U> | null | undefined ? CollectionChain<U> :
  // omitted a few lines...
  PrimitiveChain<T>;
```

to this:

```ts
type ExpChain<T> =
  // omitted a few lines...
  T extends List<infer U> | null | undefined ? U extends never ? never : CollectionChain<U> : 
  // omitted a few lines...
  PrimitiveChain<T>;
```

In some ways, this seems a bit better, but I'm not sure if it's only going to create _other_ undesirable scenarios.

types/lodash/ts3.1/common/object.d.ts
schmod: I left another comment explaining some of my rationale (and one potential alternative to this fix).  

I'm using a conditional type, because I only want the "empty-array" workaround to apply when `T[TKey]` is actually an array.  If we were to do something like:

```
_.chain({ a: 1 }).get('a', []).value();
```

I'd want to ensure that the return type is still `number | never[]`, rather than `number[]` (which is what it would be without the conditional).

sandersn (approved):

types/lodash/ts3.1/lodash-tests.ts
```diff
@@ -5181,6 +5181,7 @@ fp.now(); // $ExpectType number
     _.get({ a: undefined }, "a"); // $ExpectType undefined
     _.get({ a: value }, "a", defaultValue); // $ExpectType string | boolean
     _.get({ a: undefined }, "a", defaultValue); // $ExpectType boolean
+    _.get({ a: [1] }, "a", []); // $ExpectType number[] | never[]
```
aj-r: After reading your comment, I don't think these tests are necessary after all. I don't think they add any value so please remove them (sorry!)

aj-r (changes requested):

types/lodash/ts3.1/lodash-tests.ts
aj-r: Makes sense. I don't think we want to always prevent CollectionChain<never>. Your current solution is probably best.

types/lodash/ts3.1/lodash-tests.ts
schmod: Fixed!

types/lodash/ts3.1/common/object.d.ts
```diff
@@ -1122,6 +1122,10 @@ declare module "../index" {
          * @see _.get
          */
         get<TKey extends keyof T>(path: TKey | [TKey]): ExpChain<T[TKey]>;
+        /**
+         * @see _.get
+         */
+        get<TKey extends keyof T>(path: TKey | [TKey], defaultValue: never[]): T[TKey] extends any[] ? ExpChain<Exclude<T[TKey], undefined>> : ExpChain<Exclude<T[TKey], undefined> | []>;
```
aj-r: never[] instead of [] at the end?

aj-r (approved):

PR #35503, opened 2019-05-15
Add passport AuthenticationError class declaration

AlbertoFdzM:
Please fill in this template.

Select one of these and delete the others

  - https://github.com/jaredhanson/passport/blob/master/lib/errors/authenticationerror.js
  - https://github.com/passport/www.passportjs.org/pull/51/files

Previous PR https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35003


types/passport/lib/errors/authenticationerror.d.ts
```diff
@@ -0,0 +1,5 @@
+export default class AuthenticationError extends Error {
```
RyanCavanaugh: Why does this need its own file? Is it intended for users to import from their code? Otherwise it can just go in index.d.ts

RyanCavanaugh (changes requested):

types/passport/lib/errors/authenticationerror.d.ts
RyanCavanaugh: The linked file says
```
 * @api private
```
which really makes me believe this is not intended to be exposed

AlbertoFdzM: This is a type that can be thrown by passport if you set the option `failWithError: true`.

If you try to do `err instanceof AuthenticationError` in an error handler you will need to do the next in vanilla node js:
```
const AuthenticationError = require('passport/lib/errors/authenticationerror');

if (err instanceof AuthenticationError) {
  // Handle passport errors
} else {
  // Handle other errors
}
```

With these changes, you'll be able to do that same comparison in TS without compiler errors. If there is a cleaner way to do it please let me know.
```
import AuthenticationError from 'passport/lib/errors/authenticationerror';

if (err instanceof AuthenticationError) {
  // Handle passport errors
} else {
  // Handle other errors
}
```

PR #35510, opened 2019-05-16
[prompts] Fix `type` and `validate` function definitions

colinking:
1. Allows `type` to return `Falsy` values:

As defined in the documentation here: https://github.com/terkelg/prompts#type

The `type` field can be a function which returns a `Falsy` value to disable the prompt, which is currently not allowed by these type definitions:

> If `type` is a falsy value the prompter will skip that question.

2. Fixes the type of `values` in `PrevCaller`:

The definition for `PrevCaller` for `type` is also fixed in this PR, in that the current behavior sets `values: Answers<T>` with `T = PromptType`, instead of using `T` from `PromptObject` (so `values` contained keys like `"select"` instead of keys that corresponded to the `name`'s).

3. Updates `validate` to allow returning `boolean | string` instead of `void`

See: https://github.com/terkelg/prompts#textmessage-initial-style

> Receive user input. Should return true if the value is valid, and an error message String otherwise. If false is returned, a default error message is shown

----

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: (see above)
- [N/A] Increase the version number in the header if appropriate.
- [N/A] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

### DIFF
--- a/types/prompts/index.d.ts
+++ b/types/prompts/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for prompts 2.04
+// Type definitions for prompts 2.0
 // Project: https://github.com/terkelg/prompts
 // Definitions by: Berkay GURSOY <https://github.com/Berkays>
 //                 Daniel Perez Alvarez <https://github.com/danielpa9708>
@@ -60,7 +60,7 @@ declare namespace prompts {
     }
 
     interface PromptObject<T extends string = string> {
-        type: ValueOrFunc<PromptType> | Falsy;
+        type: PromptType | Falsy | PrevCaller<T, PromptType | Falsy>;
         name: ValueOrFunc<T>;
         message?: ValueOrFunc<string>;
         initial?: string | number | boolean | Date;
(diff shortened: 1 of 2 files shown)


types/prompts/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for prompts 2.04
+// Type definitions for prompts 2.0.4
```
colinking: `prompts` is currently available as 2.0.4: https://www.npmjs.com/package/prompts

colinking: Not sure if changing the version here like this is a "breaking change" in `DefinitelyTyped`, though.

colinking:
Feature request: `typescript-bot` ignoring Draft PRs 😅

types/prompts/index.d.ts
colinking: (Updated to `2.0`)

colinking:
Looks like `@types/prompts` is currently released under `2.4.0` because of the version typo.

In that case, not sure if it is best to leave as-is?

colinking:
Added a third fix, for the return type of `validate`.

kamontat (approved):

PR #35518, opened 2019-05-16
[Tern] fix inheritance of infer objects and added missing stuff.

nkappler:
The inheritance of abstract values and types was wrong.
Primitives were not yet implemented and some methods were on the wrong
level of inheritance, leading to unfounded complaints from the compiler

### DIFF
--- a/types/tern/index.d.ts
+++ b/types/tern/index.d.ts
@@ -2,7 +2,7 @@
 // Project: https://github.com/ternjs/tern
 // Definitions by: Nikolaj Kappler <https://github.com/nkappler>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 3.0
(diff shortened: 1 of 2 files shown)


sandersn:
Why does tern need typescript types? Are typescript users consuming the tern API to provide support in VS Code or something like that?

nkappler:
@sandersn Yes exactly.
We are building a web application where you can write scripts with basically a JavaScript subset, together with APIs we provide and strict types, quite similar to typescript in fact (but with many restrictions). We are using codemirror as editor and tern for validation, since those were already in the codebase we live in and we didn't want to add the typescript compiler.

First we had a partial declaration file for it but since it was not quite complete nor correct and it was a huge pain in the butt to make it work with the build pipeline of our department and I think it might be useful for others, I decided to make proper declarations and merge it to DefinitelyTyped.

PR #35535, opened 2019-05-17
@types/react-router: Adds support for typing the state on <Redirect />

guicamillo:
The current implementation of `<Redirect />` doesn't support any generic to type its state, so there's no way to validate it.

Imagine the following route:

```
interface MyAwesomeRouteState {
 foo: boolean;
}

<Redirect to={{
   pathname: "/my-awesome-route",
   state: {
      foo: true,
      bar: "this should fail, but it won't :'("
   }
}} />
````

With these changes, validating the state becomes possible:

```
interface MyAwesomeRouteState {
 foo: boolean;
}

<Redirect<MyAwesomeRouteState> to={{
   pathname: "/my-awesome-route",
   state: {
      foo: true,
      bar: "this will now fail :)"
   }
}} />
````
🎉

### DIFF
--- a/types/react-router/index.d.ts
+++ b/types/react-router/index.d.ts
@@ -23,7 +23,7 @@
 //                 Sebastian Silbermann <https://github.com/eps1lon>
 //                 Nicholas Hehr <https://github.com/HipsterBrown>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 2.9
 
 import * as React from 'react';
 import * as H from 'history';
--- a/types/react-router/index.d.ts
+++ b/types/react-router/index.d.ts
@@ -54,15 +54,15 @@ export interface PromptProps {
 }
 export class Prompt extends React.Component<PromptProps, any> { }
 
-export interface RedirectProps {
-  to: H.LocationDescriptor;
+export interface RedirectProps<LocationState = any> {
+  to: H.LocationDescriptor<LocationState>;
   push?: boolean;
   from?: string;
   path?: string;
   exact?: boolean;
   strict?: boolean;
 }
-export class Redirect extends React.Component<RedirectProps, any> { }
+export class Redirect<LocationState = any> extends React.Component<RedirectProps<LocationState>, any> { }
 
 export interface StaticContext {
   statusCode?: number;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by eps1lon — CHANGES_REQUESTED** [review]
Looks like a false positive in `use-default-type-parameter`. You can probably disable it.

Could you add your examples from the description as a test?

File: types/react-router/index.d.ts [hunk 1]
```diff
@@ -53,8 +53,7 @@ export interface PromptProps {
   when?: boolean;
 }
 export class Prompt extends React.Component<PromptProps, any> { }
-
-export interface RedirectProps<LocationState = any> {
+export interface RedirectProps<LocationState> {
```
eps1lon: We should default to any here. Otherwise this breaks peoples code needlessly. You probably changed this to fix the lint error? Could you try to `tslint:disable-next-line` it instead?

eps1lon (changes requested):
You can annotate failing lines with `$ExpectError`. This should allow you to illustrate the issue like you did in your description.

types/react-router/index.d.ts
guicamillo: Yeah, that was exactly what I was trying to solve - but somehow `npm run test` is not running properly locally, so I just get to see the error when it reaches the CI. 

I'll move it back.

eps1lon: @guicamillo

`npm run lint react-router` is probably what you're looking for.

guicamillo:
The only missing error to fix here is 

> Error: react-router-config depends on react-router but has a lower required TypeScript version.

Would you see any problem in bumping TS version up on `react-router-config` as well?

eps1lon:
2.9 is for default generic parameters?

guicamillo:
@eps1lon  That was the solution that the CI suggested, as you can see here.

https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/533764618?utm_source=github_status&utm_medium=notification

eps1lon:
> @eps1lon That was the solution that the CI suggested, as you can see here:
> 
> [travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/533764618?utm_source=github_status&utm_medium=notification](https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/533764618?utm_source=github_status&utm_medium=notification)

This looks like a false positive. Does tslint disable work?

PR #35539, opened 2019-05-17
Enable ban-types in express-serve-static-core

sugar700:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/expressjs/express/blob/master/History.md#400--2014-04-09


apvarun:
This PR breaks any code that tries to access the `options.req` from the express app engine callback since the type of options is changed to `object` which denotes an empty one (untyped).

I'm not sure if this also applies to template engines since I checked `ejs` and found that it tries to access the `options.filename` in it.

**Cause:** Change in type of `options` parameter of app.engine callback function from any to object, instead of an interface that defines the properties.

PR #35545, opened 2019-05-17
[@types/koa-bodyparser] Make body parameter on koa module definition optional

amaschas:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/koa-joi-router/index.d.ts

I'm using `koa-bodyparser` in my current project, and ran into an issue when I started using `koa-joi-router` alongside it. The issue is that the type definitions for `koa-joi-router` define the body parameter of the koa object as optional, while `koa-bodyparser` defines it as required. I've tested this change in my code, and everything seems to work fine if the body parameter is optional, hence this change. This would be an issue with any Koa middleware that needs to be aware of the body param on the Koa module, so this should enhance future compatibility with other middlewares as well.

### DIFF
--- a/types/koa-bodyparser/index.d.ts
+++ b/types/koa-bodyparser/index.d.ts
@@ -3,6 +3,7 @@
 // Definitions by: Jerry Chin <https://github.com/hellopao>
 //                 Anup Kishore <https://github.com/anup-2s>
 //                 Hiroshi Ioka <https://github.com/hirochachacha>
+//                 Alexi Maschas <https://github.com/amaschas>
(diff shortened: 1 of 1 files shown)


hirochachacha:
koa-joi-router have its own body parsing. https://github.com/koajs/joi-router/blob/509a0e6c1699ea7cb5bd9c8f340c8bcbae225247/joi-router.js#L368-L382
In certain situations, combination of koa-joi-router and koa-bodyparser do body parsing twice, may cause undesirable effects.
Perhaps, koa-router + koa-bodyparser is preferable since these features are orthogonal.

PR #35265, opened 2019-05-07
[styled-system] Fix definitions for mapProps and LowLevelStylefunctionArguments

Dhalton:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/styled-system/styled-system/blob/master/src/index.js
- [x] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #35556, opened 2019-05-17
[styled-system] Make N a generic on style and add typedef for scale

Dhalton:
Please fill in this template.

https://github.com/styled-system/styled-system/blob/master/src/index.js
https://styled-system.com/guides/theming

This is a follow-up PR after some suggestions were made on #35265.

@StevenLangbroek this *should* fit your needs; I used your example in a new test.
One thing that I'm still not totally sure on is the typedef for `scale`, so if anyone has any insight on that, I'd appreciate it.

Other than that, I apologize for the wait but I hope this works for y'all! 😄

### DIFF
--- a/types/styled-system/index.d.ts
+++ b/types/styled-system/index.d.ts
@@ -36,17 +36,20 @@ export interface styleFn {
     propTypes?: string[];
 }
 
-export interface LowLevelStylefunctionArguments {
-    prop: string;
-    cssProperty?: string;
-    alias?: string;
-    key?: string;
-    transformValue?: (n: string | number, scale: Array<string | number>) => any;
-    scale?: Array<string | number>;
+export type Scale = object | Array<string | number>;
+
+export interface LowLevelStylefunctionArguments<N> {
+  prop: string;
+  cssProperty?: string;
+  alias?: string;
+  key?: string;
+  transformValue?: (n: N, scale?: Scale) => any;
+  scale?: Scale;
 }
 
-export function style(
-    args: LowLevelStylefunctionArguments
+export function style<N = string | number>(
+  // tslint:disable-next-line no-unnecessary-generics
+  args: LowLevelStylefunctionArguments<N>
 ): { [cssProp: string]: string };
 
 export function compose(
--- a/types/styled-system/index.d.ts
+++ b/types/styled-system/index.d.ts
@@ -38,18 +38,18 @@ export interface styleFn {
 
 export type Scale = object | Array<string | number>;
 
-export interface LowLevelStylefunctionArguments<N> {
+export interface LowLevelStylefunctionArguments<N, S> {
   prop: string;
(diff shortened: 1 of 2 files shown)


types/styled-system/index.d.ts
```diff
@@ -36,17 +36,20 @@ export interface styleFn {
...
 }
 
-export interface LowLevelStylefunctionArguments {
-    prop: string;
-    cssProperty?: string;
-    alias?: string;
-    key?: string;
-    transformValue?: (n: string | number, scale: Array<string | number>) => any;
-    scale?: Array<string | number>;
+export type Scale = object | Array<string | number>;
```
alloy: Do those “nested objects” not follow a particular shape?

StevenLangbroek: Scales aren't known upfront, so should be generic below too actually. (we use deeply nested objects for example).

types/styled-system/index.d.ts
```diff
@@ -36,17 +36,20 @@ export interface styleFn {
...
+
+export interface LowLevelStylefunctionArguments<N> {
+  prop: string;
+  cssProperty?: string;
+  alias?: string;
+  key?: string;
+  transformValue?: (n: N, scale?: Scale) => any;
```
StevenLangbroek: Can you make this generic on `LowLevelStylefunctionArguments<N, S = Scale>`?

StevenLangbroek: Test could be my previous example:

```ts
const buttonSizes = {
  sm: '8px',
  md: '12px',
  lg: '24px',
};

const buttonSizeWithGeneric = style<keyof buttonSizes, typeof buttonSizes>({
  prop: 'size',
  cssProperty: 'font-size',
  transformValue: (size, sizes) => sizes[size]
});
```

StevenLangbroek (changes requested):
Thanks for taking the time to revisit this 🙇, requested 1 change.

peduarte (approved):

alloy:
@StevenLangbroek If you feel strongly about your request then you should leave a GH review with the “request changes” status, otherwise a maintainer with commit access is going to fly by and accidentally merge it.

types/styled-system/index.d.ts
Dhalton: Will do!

StevenLangbroek:
@alloy I did.

types/styled-system/index.d.ts
Dhalton: @StevenLangbroek I updated this PR to make scale a generic. I had to do it on `style`, the same way I did it for N, but it should work. Let me know if you're good with this change!

StevenLangbroek (approved):
Thanks so much! 💯🥂

alloy:
Great, thanks for all the discussion 👌

PR #35559, opened 2019-05-18
[mapbox__sphericalMercator] Update to properly have optional parameters

venture-vin:

 - Default size already gets set initially, so size should be optional in the constructor https://github.com/mapbox/sphericalmercator/blob/e33a8c8bf852ea7a7c57d1f9001b581fce4b6590/sphericalmercator.js#L20
 - Additionally the readme for https://github.com/mapbox/sphericalmercator specifies which parameters are optional for some of the function

### DIFF
--- a/types/mapbox__sphericalmercator/index.d.ts
+++ b/types/mapbox__sphericalmercator/index.d.ts
@@ -20,12 +20,12 @@ type BoundingBox = [number, number, number, number];
 type Projection = 'WGS84' | '900913';
 
 declare class SphericalMercator {
-    constructor(options: { size: number })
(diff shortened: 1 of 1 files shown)


venture-vin:
@nhusher Updated the type definitions for mapbox__sphericalmercator to reflect some of the optional parameters.

nhusher (approved):

nhusher:
As a minor follow-up: the doc comments source code (which was what I based the original typings off of) doesn't make the defaults clear. Might be worth opening a PR against https://github.com/mapbox/sphericalmercator to resolve that. It's a minor thing, though, I think.

PR #35561, opened 2019-05-18
[puppeteer] New mapping containing most of existing type. updated to v 1.16.0

UrielCh:

thx for your first comment.

## changes:
- Remove enumeration that had been replaced by <K extends keyof xxxxxxxxEventObj> ex: `BrowserEvents`, `PageEvents`, 
- add strict type for available keyboard key, 

```javascript
element.press('enter'); // fail
element.press('Enter'); // pass
element.press('f10'); // fail
element.press('F10'); // pass
```

- Merge all mergable interface as XXXXable interface. so that remove some duplicate interface declarations.
- split `index.d.ts` as the original files from puppeteer.
- add new call's from puppeteer 1.15.0 and 1.16.0
- add most of the class from puppeteer (Included non public ones) but removed them for this PR.
- each XXXXXable interface contains a JSdoc listing all the Implemented function and all the class using this interface in puppeteer.


types/puppeteer/index.d.ts
```diff
@@ -1528,11 +1589,20 @@ export interface SnapshopOptions {
  * exposing only the "interesting" nodes of the tree.
  */
 export interface Accessibility {
+  // constructor(client: CDPSession);
   snapshot(options?: SnapshopOptions): Promise<AXNode>;
 }
 
+
+interface TaskQueue {
```
SimonSchick: Don't think this is public API.

types/puppeteer/index.d.ts
```diff
@@ -1279,40 +1361,19 @@ export interface FrameBase extends Evalable {
...
-    xpath: string,
-    options?: WaitForSelectorOptions,
-  ): Promise<ElementHandle>;
 }
 
-export interface Frame extends FrameBase {
+
+export interface Frame extends Contentable, Evalable, Evaluateable, ExecutionContextable, FrameBase {
+  // constructor(frameManager: FrameManager, client: CDPSession, parentFrame: Frame | null | undefined, frameId: string);
+  _mainWorld: DOMWorld;
```
SimonSchick: We shouldn't expose this.

types/puppeteer/index.d.ts
```diff
@@ -2249,3 +2308,50 @@ export function executablePath(): string;
...
+ * see DeviceDescriptors.js for full list
+ */
+export const devices: {
+  [key: string]: Device;
+};
+export const error: Error | undefined;
+
+
+
+export type KeysF = 'F1' |  'F2' | 'F3' | 'F4' | 'F5' | 'F6' | 'F7' | 'F8' | 'F9' |'F10' | 'F11' | 'F12' | 'F13' | 'F14' | 'F15' | 'F16' | 'F17' | 'F18' | 'F19' |'F20' | 'F21' | 'F22' | 'F23' | 'F24';
```
SimonSchick: Do we really want to hardcode all of these constants, what about non-US layouts?

types/puppeteer/index.d.ts
```diff
@@ -2249,3 +2308,50 @@ export function executablePath(): string;
...
 export function createBrowserFetcher(options?: FetcherOptions): BrowserFetcher;
+/**
+ * Predefigned puppeter devices
+ * 
+ * see DeviceDescriptors.js for full list
+ */
+export const devices: {
+  [key: string]: Device;
+};
+export const error: Error | undefined;
```
SimonSchick: This doesn't appear to exist?

types/puppeteer/index.d.ts
```diff
@@ -1528,11 +1589,20 @@ export interface SnapshopOptions {
  * exposing only the "interesting" nodes of the tree.
  */
 export interface Accessibility {
+  // constructor(client: CDPSession);
```
SimonSchick: ?

types/puppeteer/index.d.ts
```diff
@@ -1141,42 +1276,60 @@ export interface WaitForSelectorOptionsHidden extends WaitForSelectorOptions {
...
+  setDefaultTimeout(timeout: number): void;
+  setDefaultNavigationTimeout(timeout: number): void;
+  navigationTimeout(): number;
+  timeout(): number;
+}
 
+/**
+ * internaly stored in `Frame._mainWorld` and `Frame._secondaryWorld`
+ */
+export interface DOMWorld extends Contentable, Evalable, Evaluateable, ExecutionContextable, FocusHoverable, WaitForable {
```
SimonSchick: As mentioned before, we shouldn't be exposing these as they are internals.

types/puppeteer/index.d.ts
```diff
@@ -1141,42 +1276,60 @@ export interface WaitForSelectorOptionsHidden extends WaitForSelectorOptions {
...
+  extraHTTPHeaders(): { [key: string]: string };
+  setOfflineMode(enabled: boolean): Promise<void>;
+  setUserAgent(userAgent: string): Promise<void>;
+  setCacheEnabled(enabled?: boolean): Promise<void>;
+  setRequestInterception(enabled: boolean): Promise<void>;
+}
 
-  /** Adds a `<script>` tag into the page with the desired url or content. */
-  addScriptTag(options: ScriptTagOptions): Promise<void>;
+export interface FrameManager extends EventEmitter {
```
SimonSchick: This is also not public API.

types/puppeteer/index.d.ts
```diff
@@ -1141,42 +1276,60 @@ export interface WaitForSelectorOptionsHidden extends WaitForSelectorOptions {
...
-   * If no elements match the selector, the return value resolve to [].
-   */
-  $$(selector: string): Promise<ElementHandle[]>;
-
-  /**
-   * The method evaluates the XPath expression.
-   * @param expression XPath expression to evaluate.
-   */
-  $x(expression: string): Promise<ElementHandle[]>;
+export interface NetworkManager extends EventEmitter {
```
SimonSchick: Not public API.

types/puppeteer/index.d.ts
```diff
@@ -1,17 +1,20 @@
-// Type definitions for puppeteer 1.12
+// Type definitions for puppeteer 1.15.0
```
SimonSchick: 1.16

SimonSchick (changes requested):
Really unsure about this, you are typing many non public API classes, exposing private properties and hardcoding key/modifier lists...

types/puppeteer/index.d.ts
UrielCh: Autocompletion work fine :)

types/puppeteer/index.d.ts
UrielCh: added in v1.15.0
see https://github.com/GoogleChrome/puppeteer/releases

types/puppeteer/index.d.ts
UrielCh: 4 constructor comment removed.

types/puppeteer/index.d.ts
UrielCh: Juste upgraded to 1.16.0
with 2 new methods.
@since XXXXX added in code.

types/puppeteer/index.d.ts
UrielCh: I only hardcode the Keys hardcoded in `USKeyboardLayout.js`,
About the non-US layout, since it is not available inside puppeteer code, with or without this typing that will not change anything.
If you wan to use a non US key, you will trigger a assert from `input.js`: 
    assert(definition, `Unknown key: "${keyString}"`);
the current puppeteer public interface do not allow non US typing.

types/puppeteer/index.d.ts
UrielCh: oups I forgot to remove this one, (this was the only one)

UrielCh:
Is that mandatory to have a single index.d.ts ?

SimonSchick (changes requested):
I'm sorry you are making too many changes to be able to see what's going on anymore.
Please keep your changes minimal.

UrielCh:
`DOMWorld`, `NetworkManager`, `DOMWorld`, `TaskQueue`

types/puppeteer/index.d.ts
UrielCh: Maybe I can add an enum of the existing devices  😁

types/puppeteer/index.d.ts
UrielCh: removed, but NetworkManager.d.ts stay because it contains `Response`, `ConsoleMessageLocation`, `Request`...

types/puppeteer/index.d.ts
UrielCh: removed

UrielCh:
If your point of view is too much change, I can not fix it.
If you want to keep a single large ts.d file with a lot duplicate (and not up-to-date duplicate) signature.
Reject this PR, It took me hours to make this typing to match the original source tree, and merged all duplicated signatures.
If you need some more test, that should be Ok, but no rollback.

buko106:
any update?

I want to use new API interface, browser.isConnected() from 1.16

UrielCh:
No update sorry. But you can test It. for me it works fine.

UrielCh:
next time

PR #26739, opened 2018-06-22
ClientContext field must be spelt 'custom' so that it is correctly se…

rmfbarker:
The ClientContext is passed between lambdas when they directly invoke each other. The 'custom' field is spelt with a lowercase 'c' in the Java API, and therefore is not correctly serialized and is missing if it is spelt 'Custom'.

PR #35562, opened 2019-05-18
aws-lambda: uppercase custom in client context

scamden:
per the aws docs:
https://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-context.html
```clientContext – (mobile apps) Client context provided to the Lambda invoker by the client application.

client.installation_id

client.app_title

client.app_version_name

client.app_version_code

client.app_package_name

env.platform_version

env.platform

env.make

env.model

env.locale

Custom – Custom values set by the mobile application.
```

We have been using the upper case value with no problems so i assume the docs are correct here.

### DIFF
--- a/types/aws-lambda/aws-lambda-tests.ts
+++ b/types/aws-lambda/aws-lambda-tests.ts
@@ -544,7 +544,7 @@ str = cloudwatchLogsDecodedData.logEvents[0].extractedFields!["example"];
 
 /* ClientContext */
 clientContextClient = clientCtx.client;
-anyObj = clientCtx.custom;
+anyObj = clientCtx.Custom;
 clientContextEnv = clientCtx.env;
 
 /* ClientContextEnv */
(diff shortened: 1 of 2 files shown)


buggy (approved):
LGTM

skyzenr (approved):

simonbuchan:
Hmm, I'm getting deja-vu on this one..., the same issue in lambda@edge or something?

scamden:
@simonbuchan unsure if you were asking me, but if you're waiting on my answer i'm not aware of the edge issue

simonbuchan:
More just if any of the other maintainers remember it. This package gets a lot of PRs!

khnicholas:
It was changed to lower case `custom` last year.  https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26739

scamden:
seems like according to https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26739#pullrequestreview-131053987

it can really be anything? but being inline with the official docs was a good default?

PR #35564, opened 2019-05-19
[cytoscape] Add StylesheetHelper type definitions and fix typo

freeqaz:
The comments that I added should describe this in more detail. I noticed this missing while developing and wanted to add it.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/cytoscape/cytoscape.js/blob/unstable/src/stylesheet.js


freeqaz:
Requesting @einsweniger & @sandersn 

I typed these out by hand. Didn't realize there was a helper I could have used. Also, I used only the Github UI for this change... So feel free to take ownership of these changes. Thanks for your help!

einsweniger:
@freeqaz not sure why you tagged me, but I'm pretty sure, you're supposed to add test cases.

PR #35574, opened 2019-05-19
allow arbitrary attrs on DecorationAttrs

### LINKED ISSUE #35369: [@types/prosemirror-view] DecorationAttrs interface usage
The DecorationAttrs interface only exposes the members `class`, `style`, and `nodeName` when it supposedly should allow arbitrary attributes which ProseMirror will add directly to the DOM node as HTML attributes.

Currently, I need to forcefully cast the object so it doesn't give a type error:
```js
{ class: 'some-class', 'data-placeholder': 'Some text...' } as DecorationAttrs
```

Is this intended behavior?
If not, should the DecorationAttrs be a simple type like `{ [key: string]: string }` instead?

@bradleyayers @davidka @timjb @patsimm

indfnzo:
This pull request modifies the DecorationAttrs interface to allow arbitrary strings which ProseMirror maps as-is to the target node's attributes.

This should resolve #35369.

@bradleyayers @davidka @timjb @patsimm

### DIFF
--- a/types/prosemirror-view/index.d.ts
+++ b/types/prosemirror-view/index.d.ts
@@ -128,6 +128,11 @@ export interface DecorationAttrs {
    * this type (and the other attributes are applied to this element).
    */
   nodeName?: string | null;
+  /**
+   * Specify additional attrs that will be mapped directly to the
+   * target node's DOM attributes.
+   */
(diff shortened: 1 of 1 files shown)


timjb (approved):
LGTM! Thanks!

PR #35589, opened 2019-05-20
[Algolia]: Fix AlgoliaPlaces (param language) 

aDEVE:
Please fill in this template.

Note: 

as ` hits: HitInterface[] | LocalizedHitInterface[];` didn't work well, I updated Query to handle it. 
If there is a param `language`, the result type will be different.

### DIFF
--- a/types/algoliasearch/index.d.ts
+++ b/types/algoliasearch/index.d.ts
@@ -1910,32 +1910,51 @@ declare namespace algoliasearch {
   }
 
   namespace Places {
+    interface LanguageInterface {
+      /**
+       * If specified, restrict the search results to a single language. You can pass two letters country codes (ISO 639-1).
+       * Warning: language parameter is case sensitive and should be lowercase otherwise it will fallback to default language.
+       * https://community.algolia.com/places/api-clients.html#api-options-language
+       */
+      language: string
+    }
+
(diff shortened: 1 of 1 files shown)


Haroenv:
@jonathanmontane

types/algoliasearch/index.d.ts
```diff
@@ -2013,24 +2026,18 @@ declare namespace algoliasearch {
...
-         */
-        language?: string;
     }
 
     /**
      * Result of search.
      * https://community.algolia.com/places/api-clients.html#json-answer
      */
-    interface ResultSearchInterface {
+    interface ResultSearchInterface<T> {
```
keichinger: I feel like we should add some kind of constraint to the generic parameter. Right now we could end up with something like `ResultSearchInterface<bool>`, which doesn't sound right. All hits are extending `HitInterface` or `LocalizedHitInterface`, right?

keichinger (changes requested):

types/algoliasearch/index.d.ts
aDEVE: You're right, all `hits` should extend `HitInterface` or `LocalizedHitInterface`. I add an extend to handle it. 😁

keichinger (approved):

PR #35594, opened 2019-05-20
[@loadable/server] Missing publicPath in ChunkExtractorOptions

moust:
Add missing `publicPath` parameter in `ChunkExtractorOptions` as documented [here](https://github.com/smooth-code/loadable-components/blob/9731e9c5f450f0c732c46d4dfdbbc6057a0f4d6a/website/src/pages/docs/server-side-rendering.mdx#override-statspublicpath-at-runtime) and from [here in the code](https://github.com/smooth-code/loadable-components/blob/f2a6bbd6d7e96780eed8422011942f2f2785c96f/packages/server/src/ChunkExtractor.js#L164).


martynaskadisa (approved):
Thanks!

PR #35602, opened 2019-05-20
clone and cloneMessage should be declared as generic methods

jcollins-axway:
The clone and cloneMessage methods create deep clones of the message and its fields.  It is accessible by any object which extends Message and should be callable without type casting.

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://developers.google.com/protocol-buffers/docs/reference/javascript-generated#message

### DIFF
--- a/types/google-protobuf/index.d.ts
+++ b/types/google-protobuf/index.d.ts
@@ -109,8 +109,8 @@ export abstract class Message {
   static equals(m1: Message, m2: Message): boolean;
   static compareExtensions(extension1: {}, extension2: {}): boolean;
   static compareFields(field1: any, field2: any): boolean;
-  cloneMessage(): Message;
-  clone(): Message;
+  cloneMessage<T extends Message>(): T;
+  clone<T extends Message>(): T;
   static clone<T extends Message>(msg: T): T;
   static cloneMessage<T extends Message>(msg: T): T;
   static copyInto(fromMessage: Message, toMessage: Message): void;


types/google-protobuf/index.d.ts
```diff
@@ -109,8 +109,8 @@ export abstract class Message {
   static equals(m1: Message, m2: Message): boolean;
   static compareExtensions(extension1: {}, extension2: {}): boolean;
   static compareFields(field1: any, field2: any): boolean;
-  cloneMessage(): Message;
-  clone(): Message;
+  cloneMessage<T extends Message>(): T;
+  clone<T extends Message>(): T;
```
sheetalkamat: I don't think this change is needed at all since TypeParameters do not affect anything. Person who is expecting to run `clone<SomeMessage>()` can always instead write it as `clone() as SomeMessage`

sheetalkamat (changes requested):

jcollins-axway (commented):
I don't disagree @sheetalkamat, I just think it is cleaner to be able to call clone or cloneMessage directly on a SomeMessage and get that same SomeMessage type back without having to cast it back to SomeMessage.

types/google-protobuf/index.d.ts
jcollins-axway: @sheetalkamat?

domino14:
someone please fix and merge

PR #35607, opened 2019-05-20
[react-tracking] Add types for useTracking hook

damassi:
Updates types for `react-tracking@7.0.0`, which includes [a new `useTracking` hook](https://github.com/nytimes/react-tracking/pull/124). In support of this,  `TrackingProp` now accepts a a generic, which allows us to restrict what props are passed into `tracking.trackEvent()`. 

#### Example: 

```tsx
interface Trackables {
    page: string;
    app: string;
}

 const App = track()((props: { foo: string }) => {
    const tracking = useTracking<Trackables>();

    return <div onClick={() => {
        tracking.trackEvent({
            page: 'Home'
        });
    }}/>;
});
```

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/nytimes/react-tracking/pull/124>
- [x] Increase the version number in the header if appropriate

### DIFF
--- a/types/react-tracking/index.d.ts
+++ b/types/react-tracking/index.d.ts
@@ -1,13 +1,14 @@
-// Type definitions for react-tracking 6.0
+// Type definitions for react-tracking 7.0
 // Project: https://github.com/NYTimes/react-tracking
 // Definitions by: Eloy Durán <https://github.com/alloy>
+//                 Christopher Pappas <https://github.com/damassi>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
 import * as React from "react";
 
-export interface TrackingProp {
-    trackEvent({}): any;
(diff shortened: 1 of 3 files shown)


types/react-tracking/index.d.ts
```diff
@@ -55,15 +56,26 @@ export type TrackingInfo<T, P, S> = T | ((props: P, state: S, args: any[any]) =>
...
-    tracking: Options<T> & { data?: {} }
+    tracking: Options<T> & { data?: {} };
 }>;
-
 export const ReactTrackingContext: TrackingContext;
 
+/**
+ * A React hook used to tap into the tracking context.
+ */
+export function useTracking<P = {}>(): TrackingProp<P>;
```
alloy: This doesn’t allow typing the function without using it, which means we’d need a wrapper function just to restrict it by default to our schema, but we can follow up with that in a later PR.

alloy:
👌

PR #35630, opened 2019-05-21
chart.js added types to generateLabels function

wertzui:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.chartjs.org/docs/latest/configuration/legend.html#legend-label-configuration

### DIFF
--- a/types/chart.js/index.d.ts
+++ b/types/chart.js/index.d.ts
@@ -15,6 +15,7 @@
 //                 Manuel Heidrich <https://github.com/mahnuh>
 //                 Conrad Holtzhausen <https://github.com/Conrad777>
 //                 Adrián Caballero <https://github.com/adripanico>
+//                 wertzui <https://github.com/wertzui>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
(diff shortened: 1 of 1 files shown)


types/chart.js/index.d.ts
```diff
@@ -270,7 +271,7 @@ declare namespace Chart {
         fontColor?: ChartColor;
         fontFamily?: string;
         padding?: number;
-        generateLabels?(chart: any): any;
+        generateLabels?(chart: ChartData): ChartLegendLabelItem[];
```
adripanico: According to the documentation, it should be generateLabels?(chart: Chart): ChartLegendItem[];

adripanico (changes requested):

types/chart.js/index.d.ts
wertzui: Thanks, I changed it to `Chart`

adripanico: Also ChartLegendLabelItem[] it's supposed to be ChartLegendItem[].

wertzui: No, it has the `datasetIndex` property and therefore is a `ChartLegendLabelItem`

adripanico: In that snippet they are attaching such property but the element is defined without it: https://www.chartjs.org/docs/latest/configuration/legend.html#legend-item-interface.

Actually, the link you shared is clearly identifying datasetIndex as an extra parameter.

wertzui: Then the definition of `onClick` must be updated too to reflect that change.

wertzui: @adripanico would you rather see the `onClick` method definition changed, or keep the `ChartLegendLabelItem`?

adripanico: It's ok. I'll approve it.

adripanico (approved):

PR #35633, opened 2019-05-21
[@types/stripe] Added Sources types

jekh:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://stripe.com/docs/api/sources
https://stripe.com/docs/sources

This PR adds supports for Sources. Stripe's naming convention is a little ambiguous, since it consistently refers to both cards and bank accounts as "source", but also has a separate Source object (https://stripe.com/docs/api/sources/object) that is a "peer" to card/bank account.

The Source object is used for more exotic payment types, like wire transfers, Alipay, SOFORT, etc. (see https://stripe.com/docs/sources).

**Notes**
- There was a type defined in the `sources` namespace called `sources.ISourceCreationOptions`, but it was an alias for `string | cards.ICardSourceCreationOptions`. This did not follow the convention of other resources, so I have replaced the card-specific usage of that alias with `string | cards.ICardSourceCreationOptions` so that `sources.ISourceCreationOptions` can represent the options passed to [Create Source](https://stripe.com/docs/api/sources/create)
- I added an `IStripeSource` alias that is a union of all source types, to make it easier to add (or remove) sources in the future.
- Stripe [removed support for bitcoin in April 2018](https://stripe.com/blog/ending-bitcoin-support), even though it is still included in the stripe-node package (even newly-released v7). `IStripeSource` includes `IBitcoinReceiver`, but I am happy to remove bitcoin in a separate PR.

### DIFF
--- a/types/stripe/index.d.ts
+++ b/types/stripe/index.d.ts
@@ -86,6 +86,7 @@ declare class Stripe {
     ephemeralKeys: Stripe.resources.EphemeralKeys;
     usageRecords: Stripe.resources.UsageRecords;
     usageRecordSummaries: Stripe.resources.UsageRecordSummaries;
+    sources: Stripe.resources.Sources;
 
     setHost(host: string): void;
     setHost(host: string, port: string|number): void;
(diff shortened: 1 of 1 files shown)


types/stripe/index.d.ts
```diff
@@ -7242,7 +7358,7 @@ declare namespace Stripe {
              * @param customerId The ID of the customer whose card needs to be retrieved.
              * @param sourceId The ID of the bank account to be updated.
              */
-            updateSource(customerId: string, sourceId: string, data: bankAccounts.IBankAccountUpdateOptions, options: HeaderOptions, response?: IResponseFn<cards.ICard>): Promise<cards.ICard>;
+            updateSource(customerId: string, sourceId: string, data: bankAccounts.IBankAccountUpdateOptions, options: HeaderOptions, response?: IResponseFn<bankAccounts.IBankAccount>): Promise<bankAccounts.IBankAccount>;
```
jekh: This appears to have been a copy/paste error.

PR #35636, opened 2019-05-21
Add RequestInit.signal to node-fetch

mike-marcacci:
Please fill in this template.

  - [code](https://github.com/bitinn/node-fetch/blob/master/src/index.js#L50)
  - [docs](https://github.com/bitinn/node-fetch#request-cancellation-with-abortsignal)


types/node-fetch/index.d.ts
```diff
@@ -14,6 +14,28 @@
 import { Agent } from "http";
 import { URLSearchParams } from "url";
 
+// This is defined here to prevent a dependency on a particular implementation
```
nikcorg:
```suggestion
// `Event` and `AbortSignal` are defined here to prevent a dependency on a particular implementation
```

nikcorg (changes requested):

nikcorg:
@hjr3 would you like to have a look at this, seeing as your PR [was auto-closed](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35333#issuecomment-494011680) due to inactivity. This one takes a slightly different approach than your contribution and I'd be interested to hear whether this would be amenable to your requirements.

types/node-fetch/index.d.ts
nikcorg: Just a thought, and this is definitely nitpicking, but how would you feel about moving these into an out-of-line file, e.g. `externals.d.ts` or similar, as they're not as such part of the node-fetch package, and including them in `tsconfig.json`?

mike-marcacci:
I made the suggested changes – I agree that adding an `externals.d.ts` looks prettier. I don't really have a strong preference on @hjr3's approach vs my own. There are really 3 options here:

1. Require the `dom` lib, which gives us access to the native types. This isn't great because most of the consuming packages won't be targeting an environment with these types/
2. Require and use a particular implementation. This also isn't great because for many use cases, we don't want to require the dependency, especially since we only really care that the signal object adheres to an _interface_.
3. Define the interface here. I didn't want to go too deep, which is why `Event` is really just `any`. This is fine because [the code doesn't even use the event](https://github.com/bitinn/node-fetch/blob/0c2294ec48fa5b84519f8bdd60f4e2672ebd9b06/src/index.js#L68) and really that's all we care about.

types/node-fetch/index.d.ts
```diff
@@ -14,6 +14,12 @@
 import { Agent } from "http";
 import { URLSearchParams } from "url";
 
+// `Event` and `AbortSignal` are defined here to prevent a dependency on a particular implementation
+// like the `abort-controller` package, and to avoid requiring the `dom` library
+// in `tsconfig.json`.
+import { AbortSignal } from "./externals";
```
nikcorg: We don't need to import the definitions, as we're including them in `tsconfig.ts`.

types/node-fetch/index.d.ts
```diff
@@ -14,6 +14,12 @@
...
+export * from "./externals";
```
nikcorg: I don't think we should be exporting the assistive types here, as the consuming application will use a concrete implementation.

types/node-fetch/tsconfig.json
```diff
@@ -19,6 +19,7 @@
     },
     "files": [
         "index.d.ts",
+        "externals.d.ts",
```
nikcorg: This should be in an `include` block, as it's only an assistive thing for this library, not a thing this library is exporting.

```
"include": ["externals.d.ts"],
```

types/node-fetch/externals.d.ts
```diff
@@ -0,0 +1,18 @@
+export type Event = any;
+export interface AbortSignal {
```
nikcorg: This is great, thanks for moving these out. 

However, I think they should be a declaration instead of an exported type, since we only want to describe the shape that the concrete implementation defined somewhere else should look like.

I think it's enough to replace `export` with `declare`.

nikcorg (changes requested):
Hi again, some more changes requested. Sorry, I wasn't being very clear in my initial comments.

~It looks like we're also going to lose TS 2.0 and you'll need to add `// TypeScript Version: 2.1` to `index.d.ts`. The test runner will inform you of this.~ (Nope, turns out that was just the prettier formatting of union types messing things up.)

Please also remember to add yourself to the list of contributors at the top of `index.d.ts`.

hjr3:
@mike-marcacci @nikcorg this approach works for me. Thank you both! Let me know if I can help in any way.

mike-marcacci:
Hi @nikcorg, I made several changes in the vein of your requests; hopefully this works, as I'm reluctant to fight the linter that DT is trying to standardize unless actually necessary.

types/node-fetch/tsconfig.json
mike-marcacci: Using files over include appears to be required: https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/536403252?utm_source=github_status&utm_medium=notification

types/node-fetch/index.d.ts
mike-marcacci: My rationale here is that by exporting an interface, someone can force that their implementation `implements` it. Either way, not a big deal to me.

types/node-fetch/index.d.ts
mike-marcacci: Changed.

types/node-fetch/index.d.ts
```diff
@@ -14,6 +14,12 @@
 import { Agent } from "http";
 import { URLSearchParams } from "url";
 
+// `Event` and `AbortSignal` are defined here to prevent a dependency on a particular implementation
```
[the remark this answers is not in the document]
mike-marcacci: Good catch.

mike-marcacci:
I added back explicit import/export, because while they work locally, once packaged in CI, relying on the implicit declarations breaks. I'm not sure why, but I also don't have the time to find out. Feel free to change the PR if you'd like these imported/exported differently – I definitely don't have any ego issues with "my pr" being accepted.

nikcorg (approved):
Hi @mike-marcacci. Sorry for the long reaction-time, but I was on an AFK holiday :-)

Anyway, approving as this now looks good to me 👍

PR #35658, opened 2019-05-22
Fix react-native-platform-touchable

iwikal:
The type of the `foreground` prop is a BackgroundPropType, not a
boolean. In fact, if you pass a boolean like the previous types
suggested, you get a crash.

https://github.com/react-native-community/react-native-platform-touchable#additional-props-used-by-touchablenativefeedback--default-android
> `foreground` - same as `background`


tngranados (approved):
Not sure how I missed that 😅

iwikal:
To be fair, it's easy to confuse the `foreground` prop with `useForeground` on `TouchableNativeFeedback`. Thanks for the quick response!

PR #35662, opened 2019-05-22
lodash: Add String.includes and StringChain.includes

### LINKED ISSUE #35627: [@types/lodash] Property 'includes' does not exist on type 'StringChain'.
  - Authors: @aj-r @sandersn
 
One more similar to https://github.com/DefinitelyTyped/DefinitelyTyped/issues/35400 and https://github.com/DefinitelyTyped/DefinitelyTyped/issues/35625 issue (ts is in strict mode):

```jsx
    chain('test').split('_').toLower().includes('test').value()
```

with `@types/lodash` **4.14.121** it works fine, but with **4.14.130** I get a type error:
```
 Property 'includes' does not exist on type 'StringChain'.
```
`includes` function exists here.

sandersn:
strings work with the static version even though the types look like they should only work with non-infinitely-typed collections.

Fixes #35627


aj-r (approved):
In general I think `StringChain` should be treated kind of like `CollectionChain<string>` (since a `string` is allso `ArrayLike<string>`). There may be other functions we need to add but this is probably good for now.

PR #35664, opened 2019-05-22
[luaparse] Added more complete typings for the luaparse library

thomasfn:
Added more complete typings for the luaparse library, including definitions for the lua ast. Previously this definition seemed to be a stub or placeholder.


types/luaparse/lib/ast.d.ts
```diff
@@ -0,0 +1,297 @@
+export const enum Type {
```
sheetalkamat: Do not use const enum here, There doesn't exist const of name 'Type' in your library.

types/luaparse/tslint.json
```diff
@@ -1,80 +1,9 @@
...
-        "trim-file": false,
-        "triple-equals": false,
-        "typedef-whitespace": false,
-        "unified-signatures": false,
-        "void-return": false,
-        "whitespace": false
+        // const string enum is deemed best way to represent the underlying js
+        "no-const-enum": false,
+        // no trivial way to split up ast.d.ts further to keep line limit without creating spaghetti
+        "max-line-length": false
```
sheetalkamat: Don't think that's true ., You can split the comments on multiple lines.

sheetalkamat (changes requested):

types/luaparse/lib/ast.d.ts
thomasfn: Resolved - const enum is removed.

types/luaparse/tslint.json
thomasfn: I don't fully understand. What do you mean by "split the comments on multiple lines"?

sheetalkamat: I meant please fix the error after removing this rule.

thomasfn: Ahh I was being silly and somehow mis-interpreted that linting rule as a line _count_ limit rather than a line _length_ limit. Now it makes sense about splitting the comments up. Sorry about that! Nethertheless this is now resolved.

PR #35668, opened 2019-05-22
[@types/react-table] Generics for accessor and cell

ghost:
- `Cell` and `accessor` are now generic types that will automatically be inferred from `data`. This includes string based accessors (will have an error if the property does not exist).
- Also, fixed a test.
- Lastly, fixed some linting issues regarding `[]` vs `Array<>`.

Didn't want to mess with `selectTable` and `treeTable` too much since I am not using those.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: The API is not changing at all. Just making the typings more strict by inferring.

### DIFF
--- a/types/react-table/index.d.ts
+++ b/types/react-table/index.d.ts
@@ -4,28 +4,29 @@
 //                 Pavel Sakalo <https://github.com/psakalo>,
 //                 Krzysztof Porębski <https://github.com/Havret>,
 //                 Andy S <https://github.com/andys8>,
-//                 Grzegorz Rozdzialik <https://github.com/Gelio>
-//                 Cam Pepin <https://github.com/cpepin>
+//                 Grzegorz Rozdzialik <https://github.com/Gelio>,
+//                 Cam Pepin <https://github.com/cpepin>,
+//                 Sarid Shinwari <https://github.com/ngrok-sarid>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 import * as React from 'react';
 
 export type ReactTableFunction = (value?: any) => void;
-export type AccessorFunction<D = any> = (row: D) => any;
-export type Accessor<D = any> = string | string[] | AccessorFunction<D>;
+export type AccessorFunction<D> = (row: D) => any;
+export type Accessor<D> = keyof D | Array<(keyof D)> | AccessorFunction<D>;
 export type Aggregator = (values: any, rows: any) => any;
-export type TableCellRenderer = ((cellInfo: CellInfo, column: any) => React.ReactNode) | React.ReactNode;
-export type FilterRender = (params: { column: Column, filter: any, onChange: ReactTableFunction, key?: string }) => React.ReactElement;
-export type PivotRenderer = ((cellInfo: CellInfo) => React.ReactNode) | (() => any) | string | React.ReactNode;
+export type TableCellRenderer<D> = ((cellInfo: CellInfo<D>, column: any) => React.ReactNode) | React.ReactNode;
+export type FilterRender<D> = (params: { column: Column<D>, filter: any, onChange: ReactTableFunction, key?: string }) => React.ReactElement;
+export type PivotRenderer<D> = ((cellInfo: CellInfo<D>) => React.ReactNode) | (() => any) | string | React.ReactNode;
 
 export type ComponentPropsGetter0 = (finalState: any, rowInfo: undefined, column: undefined, instance?: any) => object | undefined;
-export type ComponentPropsGetterR = (finalState: any, rowInfo?: RowInfo, column?: undefined, instance?: any) => object | undefined;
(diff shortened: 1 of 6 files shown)


types/react-table/react-table-tests.tsx
```diff
@@ -28,7 +28,7 @@ const columns: Array<Column<Data>> = [
         accessor: (data: Data) => data.age,
         Cell: props => <span className='number'>{props.value}</span>
       },
-      { Header: "Status", accessor: "status" }
+      { Header: "Status", accessor: "progress" }
```
ngrok-sarid: The updated typedefs found this error. `status` does not exist in interface `Data`.

ngrok-sarid:
@royxue @psakalo @Havret @andys8 @Gelio @cpepin Can you please review these [@types/react-table changes](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35668)? I think github webhooks is having problems today, so I am manually notifying.

types/react-table/index.d.ts
```diff
@@ -4,28 +4,29 @@
...
-//                 Cam Pepin <https://github.com/cpepin>
+//                 Grzegorz Rozdzialik <https://github.com/Gelio>,
+//                 Cam Pepin <https://github.com/cpepin>,
+//                 Sarid Shinwari <https://github.com/ngrok-sarid>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 import * as React from 'react';
 
 export type ReactTableFunction = (value?: any) => void;
-export type AccessorFunction<D = any> = (row: D) => any;
```
Gelio: Removing the default generic argument is a breaking change. I am not sure what the policy on that is. I am fine with this change, but I am not sure if the updated typings' version will reflect the breaking change

types/react-table/index.d.ts
```diff
@@ -4,28 +4,29 @@
...
-export type Accessor<D = any> = string | string[] | AccessorFunction<D>;
+export type AccessorFunction<D> = (row: D) => any;
+export type Accessor<D> = keyof D | Array<(keyof D)> | AccessorFunction<D>;
 export type Aggregator = (values: any, rows: any) => any;
-export type TableCellRenderer = ((cellInfo: CellInfo, column: any) => React.ReactNode) | React.ReactNode;
-export type FilterRender = (params: { column: Column, filter: any, onChange: ReactTableFunction, key?: string }) => React.ReactElement;
-export type PivotRenderer = ((cellInfo: CellInfo) => React.ReactNode) | (() => any) | string | React.ReactNode;
+export type TableCellRenderer<D> = ((cellInfo: CellInfo<D>, column: any) => React.ReactNode) | React.ReactNode;
```
Gelio: Adding a new generic parameter without a default value is also a breaking change :slightly_frowning_face:

Gelio (commented):
I will try to give these changes a more thorough look, but for now I have some concerns regarding introducing breaking changes. If we make those, will the package be released with the major version increased?

types/react-table/react-table-tests.tsx
Gelio: How about changing the header name to `Progress`, so that it matches the `accessor`? Having the header differ from the accessor may cause confusion for developers modifying the tests in the future 😄

```suggestion
      { Header: "Progress", accessor: "progress" }
```

Gelio (approved):
Overall, the changes look good to me 👍 The only matter that worries me is that those are breaking changes when it comes to types, as generic parameters are now required.

I am all for making them mandatory, as this is more explicit and it would be easier to notice room for improvement (improving the types in one's codebase), yet only if this package is released with an updated major version, due to those breaking changes.

robertrbairdii:
@Gelio @ngrok-sarid I noticed some breaking changes when updated dependencies this morning and it appears this was published as a patch release instead of a major version change. This seems like it might cause some people an issue today if they're not pinning their dependencies.

Gelio:
This is exactly what I wrote in my previous comments :)

robertrbairdii:
Yeah, but the bot actually published it to npm last night with the patch release so it’s actually breaking builds now

Gelio:
I have noticed :) @sheetalkamat do you have the authority to publish a new major version with this change, and a patch update that reverts this PR, so that 6.8.x does not break :slightly_smiling_face: ?

jeffy-g:
This change breaks typescript type validation 👎

sheetalkamat:
I have created PR to revert this change. Once that's done, this PR needs to redone where the major version is changed.

sheetalkamat:
@ngrok-sarid The change has been reverted, Please create new PR with update to major version to avoid breaking others.

iamandrewluca:
I think that types can be added without breaking changes.  
Also typescript tests need to be added for both version, typed generics and and untyped generics.  
I can give it a try and add  these back without breaking changes.

ps: I added this kind of generics in the previous days to `redux-form` and `final-form`

PR #35678, opened 2019-05-23
Create types for liquid-pid

kadupenido:

### DIFF
--- a/types/liquid-pid/index.d.ts
+++ b/types/liquid-pid/index.d.ts
@@ -1,0 +1,21 @@
+// Type definitions for liquid-pid 1.0
+// Project: https://github.com/hekike/liquid-pid
+// Definitions by: Kadu Penido <https://github.com/kadupenido>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export interface PIDControllerOption {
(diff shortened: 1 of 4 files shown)


types/liquid-pid/index.d.ts
```diff
@@ -0,0 +1,21 @@
...
+
+export interface PIDControllerOption {
+    temp?: { ref?: number };
+    Pmax?: number;
+    Kp?: number;
+    Ki?: number;
+    Kd?: number;
+}
+
+export class PIDController {
```
RyanCavanaugh: This isn't how the object is exported.

types/liquid-pid/liquid-pid-tests.ts
```diff
@@ -0,0 +1,21 @@
+import * as PIDController from "liquid-pid";
```
RyanCavanaugh: Your test should use (at a minimum) the code from the project's README. It says
```ts
var
  liquidPID = require('liquid-pid'),
  actualP = 0,
  pidController;

pidController = new liquidPID({
  temp: {
```

which will fail given the way you've written this .d.ts file

RyanCavanaugh (changes requested):

PR #35679, opened 2019-05-23
Add types for purdy@3.5

YellowKirby:
Please fill in this template.

If adding a new definition:

----

I was running into some issues with the linter, specifically with the `strict-export-declare-modifiers` rule. [purdy exports a function that also has methods on it](https://github.com/danielb2/purdy.js/blob/master/lib/index.js#L56-L90), so I'm not entirely sure if the function + namespace declaration is the right way to define the types (or if there's a better way to handle it without needing to ignore the `strict-export-declare-modifiers` error).

### DIFF
--- a/types/purdy/index.d.ts
+++ b/types/purdy/index.d.ts
@@ -7,15 +7,13 @@
 declare function Purdy(obj: object, options: Purdy.Options): void;
 
 declare namespace Purdy {
-    // tslint:disable-next-line strict-export-declare-modifiers
-    export interface Instance {
+    interface Instance {
         (obj: object, options: Options): void;
         print: (...args: any[]) => void;
         stringify: (...args: any[]) => string;
     }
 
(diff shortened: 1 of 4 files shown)


types/purdy/index.d.ts
```diff
@@ -0,0 +1,37 @@
+// Type definitions for purdy 3.5
+// Project: https://github.com/danielb2/purdy.js
+// Definitions by: YellowKirby <https://github.com/YellowKirby>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
+declare function Purdy(obj: object, options: Purdy.Options): void;
+
+declare namespace Purdy {
+    // tslint:disable-next-line strict-export-declare-modifiers
```
RyanCavanaugh: There's no reason to disable this rule

RyanCavanaugh (changes requested):

PR #35681, opened 2019-05-23
react-native: Fix the return type of queryCache

Gerhut:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/facebook/react-native/pull/23602>

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -3992,7 +3992,7 @@ export class Image extends ImageBase {
     static getSize(uri: string, success: (width: number, height: number) => void, failure: (error: any) => void): any;
     static prefetch(url: string): any;
     static abortPrefetch?(requestId: number): void;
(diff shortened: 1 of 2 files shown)


alloy (changes requested):
Looks good 👍 However, can you add some test coverage? And it’s also best to leave formatting (of the string quotes in this case) as it was.

Gerhut:
Hi @alloy,

I just found that there is no static methods in `ImageBackground` class, neither in docs nor in codes, including code history.

Could I remove it? If I remove it will it be a breaking change?

alloy:
Thanks for making those changes!!

Re the ImageBackground class, if there’s types in here for things that don’t exist at runtime, then yeah that should be cleaned up 👍 However, let’s do that in a follow-up PR? You may also want to check the original PR that added those to research co text for why they were added.

alloy (approved):

PR #35685, opened 2019-05-23
@types/algoliasearch-helper: Add missing method getQueryParams to SearchParameters

ValentinFunk:
Adds typings for method getQueryParams() from https://github.com/algolia/algoliasearch-helper-js/blob/develop/src/SearchParameters/index.js#L1534

This is useful for example for using searchOnce and using the SearchParameters API:
```ts
        const originalState = helperWithSearchRequest.getState();
        // SearchParameters is immutable - this does not modify the original search
        const sameSearchButForTrips: SearchParameters = originalState.addFacetRefinement('type', 'trip');
        return helper.searchOnce(sameSearchButForTrips.getQueryParams());
``` 
(if any of the original maintainers read this, it would also be nice to have a 'searchOnce' method that accepts a SearchParameters object)
#### Template Info

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/algolia/algoliasearch-helper-js/blob/develop/src/SearchParameters/index.js#L1534
- [x] Increase the version number in the header if appropriate: Not needed, present in current version

### DIFF
--- a/types/algoliasearch-helper/index.d.ts
+++ b/types/algoliasearch-helper/index.d.ts
@@ -460,7 +460,7 @@ declare namespace algoliasearchHelper {
       'index',
       'facets', 'disjunctiveFacets', 'facetsRefinements',
       'facetsExcludes', 'disjunctiveFacetsRefinements',
-      'numericRefinements', 'tagRefinements', 'hierarchicalFacets', 'hierarchicalFacetsRefinements'  
+      'numericRefinements', 'tagRefinements', 'hierarchicalFacets', 'hierarchicalFacetsRefinements'
     ];
 
     constructor(newParameters?: QueryParameters)
--- a/types/algoliasearch-helper/index.d.ts
+++ b/types/algoliasearch-helper/index.d.ts
@@ -385,6 +385,11 @@ declare namespace algoliasearchHelper {
     enableExactOnSingleWordQuery?: boolean;
   }
 
+  type ValueTypes<T> = T extends Array<infer U> ? U : never;
+  type ManagedParameters = {
+    [K in ValueTypes<SearchParameters["managedParameters"]>]: SearchParameters[K]
+  };
+
   export class SearchParameters implements QueryParameters {
     index?: string ;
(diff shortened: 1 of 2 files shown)


PikachuEXE:
Why not also update the test though?
Also it seems `getQueryParams` always return `object` instead of `any`

types/algoliasearch-helper/index.d.ts
```diff
@@ -478,6 +478,7 @@ declare namespace algoliasearchHelper {
     getHierarchicalRefinement(facetName: string): string[];
     getNumericRefinements(facetName: string): SearchParameters.OperatorList[];
     getNumericRefinement(attribute: string, operator: SearchParameters.Operator): Array<number | number[]>;
+    getQueryParams(): any;
```
Haroenv:
```suggestion
    getQueryParams(): SearchParameters;
```

Kamshak: @Haroenv Would it make sense to also update helper.searchOnce to accept type QueryParameters?

Haroenv: @Kamshak, yes, that's correct!

Haroenv (approved):

Kamshak:
@Haroenv 
I checked the code again, it looks like the query params that are returned by getQueryParams are restricted to "managedParameters" which is a subset of the full QueryParameters [1].

I've updated the types to reflect this. I've introduced a new (non-exported) type `ManagedParameters` which contains these types and is mapped from the `managedTypes` property (for which I also added typings). The field types map to the corresponding type in `SearchParameters` (and therefore also `QueryParameters` since `class SearchParameters implements QueryParameters`, but I felt this closer expresses what's actually happening).

I've added a test for the function and to make sure that the return can be used for querying

[1] https://github.com/algolia/algoliasearch-helper-js/blob/develop/src/SearchParameters/index.js#L1528

Haroenv:
that's a very good catch, thanks!

PR #35692, opened 2019-05-23
classnames: export types

JoshMcCullough:

This PR shouldn't change how people are currently using the types, but it exports types which are important in some circumstances.

### DIFF
--- a/types/classnames/classnames-tests.ts
+++ b/types/classnames/classnames-tests.ts
@@ -1,5 +1,5 @@
 import classNames = require('classnames');
-import * as classNames2 from 'classnames';
+import classNames2 from 'classnames';
 import * as cn from 'classnames/bind';
 
 classNames2('foo', 'bar'); // => 'foo bar'
--- a/types/classnames/classnames-tests.ts
+++ b/types/classnames/classnames-tests.ts
@@ -1,5 +1,5 @@
 import classNames = require('classnames');
-import classNames2 from 'classnames';
+import * as classNames2 from 'classnames';
 import * as cn from 'classnames/bind';
 
 classNames2('foo', 'bar'); // => 'foo bar'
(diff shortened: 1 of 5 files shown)


types/classnames/classnames-tests.ts
```diff
@@ -1,5 +1,5 @@
 import classNames = require('classnames');
-import * as classNames2 from 'classnames';
+import classNames2 from 'classnames';
```
seansfkelley: Why did this need to change? Did the supported import style change?

JoshMcCullough: Was not intentional -- fixed.

types/classnames/index.d.ts
```diff
@@ -6,18 +6,11 @@
 //                 Sean Kelley <https://github.com/seansfkelley>
 //                 Michal Adamczyk <https://github.com/mradamczyk>
 //                 Marvin Hagemeister <https://github.com/marvinhagemeister>
+//                 Josh McCullough <https://github.com/joshmccullough>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
-type ClassValue = string | number | ClassDictionary | ClassArray | undefined | null | boolean;
```
seansfkelley: Could you just export the names are defined here instead of putting them in a new file?

JoshMcCullough:
I tried but it broke the tests. I can try again maybe I missed something.

seansfkelley:
That's legitimately surprising, unless it introduces some kind of name clash?

JoshMcCullough:
Whoops, meant to comment inline. Since the current type def does `export = classNames;` specifically, you cannot export other things. I didn't want to remove that line because it is necessary AFAIK and if it's removed everything breaks.

seansfkelley:
Arrggg, right, that makes sense. I guess if you need the types then there's no other way around it.

¯\\\_(ツ)\_/¯

seansfkelley (approved):

PR #35693, opened 2019-05-23
feat(sql.js): Add binary WebAssembly instantiation

bennycode:
Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: - https://github.com/kripken/sql.js/blob/v1.0.0/dist/sql-wasm-debug.js#L2360

### DIFF
--- a/types/sql.js/index.d.ts
+++ b/types/sql.js/index.d.ts
@@ -12,7 +12,8 @@ export type ParamsObject = Record<string, ValueType>;
 export type ParamsCallback = (obj: ParamsObject) => void;
 
 export interface Config {
-    locateFile(fileName?: string): string;
+    locateFile?(fileName?: string): string;
+    wasmBinary?: Uint8Array;
 }
 
 export interface QueryResults {


hozuki (commented):
Thank you for the URL provided. It helped a lot.

Seeing the changes I got a little suspicious and did some more research. Considering that sql.js is compiled via Emscripten, it seems the `config` argument is of type `Module` ([here](https://emscripten.org/docs/api_reference/module.html) is its API documentation). Unfortunately this is not explicitly documented in the repo of sql.js.

After knowing this, I think we should reuse existing definition. DefinitelyTyped has a [definition file](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/emscripten/index.d.ts) for Emscripten related stuff, but it is quite tricky. Here, for the `Module` object, we need an interface, not a namespace (and apparently not `WebAssembly.Module`); also, all of its members should be optional. Neither directly referencing Emscripten definitions nor copying those definitions here (and make corresponding changes) is a "good" solution.

I cannot make a call for now. I suggest to discuss with maintainers of Emscripten's definitions and see how we can coordinate. @zakki @periklis 

---

Oh there is another thing. I noticed that sql.js supports both ways of import, import all and import default, not only the latter one. This should be treated in another PR though.

ffflorian:
I created a PR to include all `Module` functions and vars into the `config` object: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35713

@hozuki

> I noticed that sql.js supports both ways of import, import all and import default, not only the latter one.

What do you mean by "import all"?

hozuki:
@ffflorian 

OK I'll look into that PR.

> What do you mean by "import all"?

I mean this:

https://github.com/kripken/sql.js/blob/v1.0.0/dist/sql-wasm-debug.js#L6904

The `declare module "xxx" { export = xxx }` stuff.

ffflorian:
@hozuki I see. But AFAIC it's not possible to export as default and as `=` at the same time.

hozuki:
@ffflorian I think it still matters if someone wants to transform existing JS to TS, or insists using one of those import methods (well, they are both allowed). It's a rare case though.

Anyway, for the solution, I think I just got it, but it is a little complicated.

```typescript
// Some file name, for example 'module.d.ts'

export module SqlJS {
    // Other exported members; no 'declare'
    // ...

    export interface SqlJsStatic {
        Database: typeof Database;
        Statement: typeof Statement;
    }

    interface InitSqlJsStatic extends Function {
        (config?: SqlJS.Config): Promise<SqlJS.SqlJsStatic>;
        readonly default: InitSqlJsStatic;
    }

}

declare global {
    const initSqlJs: SqlJS.InitSqlJsStatic;
}
```

```typescript
// index.d.ts

/// <reference path="./module.d.ts" />

declare module "sql.js" {
    export = initSqlJs;
}
```

It may not be the most elegant solution: there is global name contamination, as those old declare module definitions.

But now it can be imported via `import *` and import default:

```typescript
import * as init1 from "sql.js";
import init2 from "sql.js";
import init3 = require("sql.js");

// TypeScript compiler can now deduce the types of init1, init2 and init3
console.log(init1.toString());
console.log(init1 === init2 && init2 === init3);
```

Plus, `initSqlJs` can still be accessed in the global context, should the module be loaded via a `<script>` tag. (I forgot this case. Thanks for your code.)

If the augmentation (`declare module "sql.js"`) is not separated into another file, there will be an annoying TS2665 error. It was really a headache. I got the solution idea from ajaxorg/ace-builds#133.

ffflorian:
Thanks @hozuki, I will implement your changes into https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35713.

bennyn:
Thanks @ffflorian & @hozuki for moving this issue further. I will close my PR in favor of #35713.

PR #35694, opened 2019-05-23
Add missing types for aliases querystring.encode and querystring.decode to v12 & v11

methodbox:
Please fill in this template.

v11.x:  https://nodejs.org/docs/latest-v11.x/api/querystring.html#querystring_querystring_encode
v12.x:  https://nodejs.org/api/querystring.html

### DIFF
--- a/types/node/querystring.d.ts
+++ b/types/node/querystring.d.ts
@@ -19,6 +19,14 @@ declare module "querystring" {
 
     function stringify(obj?: ParsedUrlQueryInput, sep?: string, eq?: string, options?: StringifyOptions): string;
     function parse(str: string, sep?: string, eq?: string, options?: ParseOptions): ParsedUrlQuery;
+    /**
+    * The querystring.encode() function is an alias for querystring.stringify().
+    */
+    function encode(obj?: ParsedUrlQueryInput, sep?: string, eq?: string, options?: StringifyOptions): string;
+    /**
+    * The querystring.encode() function is an alias for querystring.stringify().
+    */
+    function decode(str: string, sep?: string, eq?: string, options?: ParseOptions): ParsedUrlQuery;
     function escape(str: string): string;
     function unescape(str: string): string;
 }
--- a/types/node/querystring.d.ts
+++ b/types/node/querystring.d.ts
@@ -20,12 +20,12 @@ declare module "querystring" {
     function stringify(obj?: ParsedUrlQueryInput, sep?: string, eq?: string, options?: StringifyOptions): string;
     function parse(str: string, sep?: string, eq?: string, options?: ParseOptions): ParsedUrlQuery;
     /**
-    * The querystring.encode() function is an alias for querystring.stringify().
-    */
+     * The querystring.encode() function is an alias for querystring.stringify().
+     */
     function encode(obj?: ParsedUrlQueryInput, sep?: string, eq?: string, options?: StringifyOptions): string;
     /**
-    * The querystring.encode() function is an alias for querystring.stringify().
-    */
+     * The querystring.encode() function is an alias for querystring.stringify().
+     */
     function decode(str: string, sep?: string, eq?: string, options?: ParseOptions): ParsedUrlQuery;
     function escape(str: string): string;
     function unescape(str: string): string;
--- a/types/node/querystring.d.ts
+++ b/types/node/querystring.d.ts
@@ -24,7 +24,7 @@ declare module "querystring" {
      */
     function encode(obj?: ParsedUrlQueryInput, sep?: string, eq?: string, options?: StringifyOptions): string;
     /**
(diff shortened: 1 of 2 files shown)


types/node/querystring.d.ts
```diff
@@ -19,6 +19,14 @@ declare module "querystring" {
 
     function stringify(obj?: ParsedUrlQueryInput, sep?: string, eq?: string, options?: StringifyOptions): string;
     function parse(str: string, sep?: string, eq?: string, options?: ParseOptions): ParsedUrlQuery;
+    /**
+     * The querystring.encode() function is an alias for querystring.stringify().
+     */
+    function encode(obj?: ParsedUrlQueryInput, sep?: string, eq?: string, options?: StringifyOptions): string;
+    /**
+     * The querystring.encode() function is an alias for querystring.stringify().
```
Flarna: The comment is wrong here.

methodbox: Good call. Bad copy paste on my part :(. I'll approve your correction. Thanks for catching that!

methodbox (commented):
All good.

types/node/querystring.d.ts
methodbox: Sorry I got confused here. I'll push a correction then.

methodbox: Actually, no I'm confused where you're seeing this. I'm looking at my most recent commit and I don't see this:

https://github.com/methodbox/DefinitelyTyped/commit/e4f53b61d6ffffb8a2aadd86fd3e97ad30075cd1

Flarna: I look into the diff directly on GitHub. the file in v11 folder looks good but that one in node folder has the copy paste error.

methodbox: Okay, I’ve resolved it - seems like my browser was caching it or something :).

types/node/querystring.d.ts
```diff
@@ -19,6 +19,14 @@ declare module "querystring" {
 
     function stringify(obj?: ParsedUrlQueryInput, sep?: string, eq?: string, options?: StringifyOptions): string;
     function parse(str: string, sep?: string, eq?: string, options?: ParseOptions): ParsedUrlQuery;
+    /**
+     * The querystring.encode() function is an alias for querystring.stringify().
+     */
+    function encode(obj?: ParsedUrlQueryInput, sep?: string, eq?: string, options?: StringifyOptions): string;
```
islishude: I think that the `obj` should not a optional,If ignore it,`encode` always returns an empty string.https://github.com/nodejs/node/blob/master/lib/querystring.js#L160-L197.

methodbox (commented):
@islishude Can you look at this please? I understand your comment, but I think in this case it's incorrect.

islishude (approved):

Flarna:
I think it would be simpler to use
```
const encode: typeof stringify;
const decode: typeof parse;
```
to express that these are aliases. This way it's ensured that they stay in sync.

methodbox:
@Flarna But `stringify` and `parse` aren’t defined as types, they’re defined as functions which take specific types as parameters and return specific types.

I can rewrite that, but then it would be necessary to do so in all types files.

I can do that, if you think that’s best?

methodbox:
I guess I also consider the usefulness of the type definition. Using it as I have it makes it much more verbose; referencing `typeof` means the definition seen is just `typeof X` then you have to go look up what X is ... kind of missing the point of TS.

SimonSchick:
Duplication is not the point of TS.

methodbox:
> Duplication is not the point of TS.

That's a bit of a snarky answer that ignores _my point:_ documentation. And I agree, duplication isn't the point, but usefulness is certainly to be considered.

If the owners want it rewritten such that you have to take two steps instead of one to find the definition of a type just so we can prevent "duplication" of code (note that this is already effectively done within this same file - not my doing - just look at escape and unescape) so be it; I'll change it.

But the definition IMO is more useful when you can see it explicitly rather than saying one thing is a `typeof` another thing, then having to look that thing up.

Flarna:
@methodbox I tried this locally and there seems to be no need to rewrite `stringify` and `parse`. Regarding docs/code completion: At least VS Code is able to resolve this indirection and shows the full info as if you use `stringify` or `parse`.

The `typeof` solution is in use at several other places, for example in global.d.ts.

I will not block this as it is but would be nice if you could give this a try in your environment. If it doesn't work nice there keep it as it is.

methodbox:
In my VS Code I didn’t see that but I’ll give it another look as I’m not 100% sure I tested that.

I can put in a new commit with the fix. It seems most people agree this is the way to go, and at least it also gets fixed. I’m okay with that compromise.

Flarna (approved):

BrunoScheufler (approved):

Issue #32858, opened 2019-02-07
[redux-form] FieldArray fails with a custom component in TS 3.2+


sargant:
- [x] I tried using the `@types/redux-form` package and had problems.
- [x] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [ ] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @Kovensky @CarsonF @aikoven @LKay @bancek @Aleksion @weswigham

Using a `<FieldArray>` in TypeScript 3.2+ fails, where in TS 3.1 is was previously succeeding. Related to #31215.

Installed dependencies:
* `redux-form@8.1.0`
* `@types/redux-form@7.4.16`

Minimal example code:
```
import * as React from 'react'
import { FieldArray, WrappedFieldArrayProps } from 'redux-form'

export const ArrayComponent: React.FC<WrappedFieldArrayProps<any>> = () => null
export const Form: React.FC = () => <FieldArray name='test' component={ArrayComponent} />
```

Compiles without error using TypeScript 3.1. Using TS 3.2/3.3, provided the following error:
```
Type 'FunctionComponent<WrappedFieldArrayProps<any>>' is not assignable to type '(ComponentClass<{ name: "test"; component: FunctionComponent<WrappedFieldArrayProps<any>>; }, any> & FunctionComponent<WrappedFieldArrayProps<any>>) | (FunctionComponent<{ name: "test"; component: FunctionComponent<...>; }> & FunctionComponent<...>)'.
  Type 'FunctionComponent<WrappedFieldArrayProps<any>>' is not assignable to type 'ComponentClass<{ name: "test"; component: FunctionComponent<WrappedFieldArrayProps<any>>; }, any> & FunctionComponent<WrappedFieldArrayProps<any>>'.
    Type 'FunctionComponent<WrappedFieldArrayProps<any>>' is not assignable to type 'ComponentClass<{ name: "test"; component: FunctionComponent<WrappedFieldArrayProps<any>>; }, any>'.
      Types of property 'propTypes' are incompatible.
[...]

PR #35723, opened 2019-05-24
[@types/Redux-form] - Update FieldArray to type custom field arrays correctly in TS > 3.2

m-b-davis:

This PR fixes the issue specified at https://github.com/DefinitelyTyped/DefinitelyTyped/issues/32858.

Currently, typing your own custom field arrays creates errors with later versions of typescript (>3.2 and throws the following error: 

```Type 'FunctionComponent<WrappedFieldArrayProps<any>>' is not assignable to type '(ComponentClass<{ name: "test"; component: FunctionComponent<WrappedFieldArrayProps<any>>; }, any> & FunctionComponent<WrappedFieldArrayProps<any>>) | (FunctionComponent<{ name: "test"; component: FunctionComponent<...>; }> & FunctionComponent<...>)'.
  Type 'FunctionComponent<WrappedFieldArrayProps<any>>' is not assignable to type 'ComponentClass<{ name: "test"; component: FunctionComponent<WrappedFieldArrayProps<any>>; }, any> & FunctionComponent<WrappedFieldArrayProps<any>>'.
    Type 'FunctionComponent<WrappedFieldArrayProps<any>>' is not assignable to type 'ComponentClass<{ name: "test"; component: FunctionComponent<WrappedFieldArrayProps<any>>; }, any>'.
      Types of property 'propTypes' are incompatible.
        Type 'WeakValidationMap<WrappedFieldArrayProps<any>> | undefined' is not assignable to type 'WeakValidationMap<{ name: "test"; component: FunctionComponent<WrappedFieldArrayProps<any>>; }> | undefined'.
          Type 'WeakValidationMap<WrappedFieldArrayProps<any>>' has no properties in common with type 'WeakValidationMap<{ name: "test"; component: FunctionComponent<WrappedFieldArrayProps<any>>; }>'.
```

This PR makes it possible to create custom field arrays like so:

```
type MyFieldValue = { foo: string };
type MyProps = { extraProp: number; }

class MyCustomFieldArray extends FieldArray<MyProps, MyFieldValue> {}

const MyFieldComponent = (props: MyFieldProps & WrappedFieldArrayProps<MyFieldValue>) => null;
...
    <MyCustomFieldArray
          component={MyFieldComponent}
          props={{ myProps }}
     />

```

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://redux-form.com/7.2.3/docs/api/fieldarray.md/>>

### DIFF
--- a/types/redux-form/v7/lib/FieldArray.d.ts
+++ b/types/redux-form/v7/lib/FieldArray.d.ts
@@ -1,29 +1,30 @@
 import { Component, ComponentType } from "react";
 import { Validator } from "../index";
 
-export interface BaseFieldArrayProps<P = {}> {
+interface _BaseFieldArrayProps<P = {}, FieldValue = any> {
     name: string;
-    component: ComponentType<P>;
+    component: ComponentType<WrappedFieldArrayProps<FieldValue> & P>;
     validate?: Validator | Validator[];
     warn?: Validator | Validator[];
     withRef?: boolean;
-    props?: P;
     rerenderOnEveryChange?: boolean;
 }
 
-export interface GenericFieldArray<Field, P = {}> extends Component<BaseFieldArrayProps<P> & Partial<P>> {
+export type BaseFieldArrayProps<P = {}, FieldValue = any> = (P | { props: P}) & _BaseFieldArrayProps<P, FieldValue>;
+
+export interface GenericFieldArray<FieldValue = any, P = {}> extends Component<BaseFieldArrayProps<P, FieldValue>> {
     name: string;
     valid: boolean;
-    getRenderedComponent(): Component<WrappedFieldArrayProps<Field> & P>;
+    getRenderedComponent(): Component<WrappedFieldArrayProps<FieldValue> & P>;
 }
 
-export class FieldArray<P = {}> extends Component<BaseFieldArrayProps<P> & Partial<P>> implements GenericFieldArray<any, P> {
+export class FieldArray<P = {}, FieldValue = any> extends Component<BaseFieldArrayProps<P, FieldValue>> implements GenericFieldArray<FieldValue, P> {
     name: string;
     valid: boolean;
-    getRenderedComponent(): Component<WrappedFieldArrayProps<any> & P>;
+    getRenderedComponent(): Component<WrappedFieldArrayProps<FieldValue> & P>;
 }
 
-export interface WrappedFieldArrayProps<FieldValue> {
+export interface WrappedFieldArrayProps<FieldValue = any> {
     fields: FieldArrayFieldsProps<FieldValue>;
     meta: FieldArrayMetaProps;
 }
@@ -35,6 +36,7 @@ export interface FieldArrayFieldsProps<FieldValue> {
     get(index: number): FieldValue;
     getAll(): FieldValue[];
     removeAll(): void;
+
     insert(index: number, value: FieldValue): void;
     name: string;
     length: number;
--- a/types/redux-form/index.d.ts
+++ b/types/redux-form/index.d.ts
@@ -7,6 +7,7 @@
 //                 Alex Young <https://github.com/alsiola>
 //                 Anton Novik <https://github.com/tehbi4>
 //                 Huw Martin <https://github.com/huwmartin>
+//                 Matt Davis <https://github.com/m-b-davis>
 //                 Ethan Resnick <https://github.com/ethanresnick>
 //                 Tim de Koning <https://github.com/reggino>
 //                 Maddi Joyce <https://github.com/maddijoyce>
--- a/types/redux-form/lib/FieldArray.d.ts
+++ b/types/redux-form/lib/FieldArray.d.ts
@@ -1,29 +1,30 @@
 import { Component, ComponentType } from "react";
 import { Validator } from "../index";
 
-export interface BaseFieldArrayProps<P = {}> {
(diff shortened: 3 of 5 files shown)


types/redux-form/v7/lib/FieldArray.d.ts
```diff
@@ -1,29 +1,29 @@
...
-    component: ComponentType<P>;
+    component: ComponentType<WrappedFieldArrayProps<FieldValue> & P>;
     validate?: Validator | Validator[];
     warn?: Validator | Validator[];
     withRef?: boolean;
     props?: P;
     rerenderOnEveryChange?: boolean;
 }
 
-export interface GenericFieldArray<Field, P = {}> extends Component<BaseFieldArrayProps<P> & Partial<P>> {
```
huwmartin: @VincentBel, I see @m-b-davis is removing `& Partial<P>` here in favour of using `FieldValue` - would you be able to take a look and see what you think as I don't have the context for why `Partial<P>` was used originally and what potential use-cases might be out there.

huwmartin (commented):
Looks really good to me and addresses a problem I've seen with FieldArrays and adds some extra safety.

Would be keen for some input from @VincentBel on what the context/intent of his original work was to make sure we're not breaking any potential use-cases with this change!

types/redux-form/v7/lib/FieldArray.d.ts
VincentBel: @huwmartin I think it is necessary here because we can use `FieldArray` like this:

```js
const renderMembers = ({ fields, meta, foo }) => (
  <div>{foo}</div>
)

const FieldArraysForm = ({ foo }) => {
  return <FieldArray name="members" foo={foo} component={renderMembers} />
}
```

m-b-davis: @VincentBel - I don't believe this is a valid way of using FieldArray. Could you provide an example from the docs where this is supported behaviour? 

From the docs:

image
https://redux-form.com/8.1.0/docs/api/fieldarray.md/

As the props prop is the object which contains extra props to pass to the child, (I think) your example above should be written as :

```
const renderMembers = ({ fields, meta, foo }) => (
  <div>{foo}</div>
)

const FieldArraysForm = ({ foo }) => {
  return <FieldArray name="members" props={{ foo }} component={renderMembers} />
}
```

What are your thoughts on this?

VincentBel: Hi, @m-b-davis.

It is just like how we can pass props when using `Field` component ([Field doc](
https://redux-form.com/8.1.0/docs/api/field.md/#-code-props-object-code-optional-)):

image

I think it is just not documented at the `FieldArray` API page. Here is a demo how we can use it: https://codesandbox.io/s/reduxform-fieldarray-props-bqrgf

m-b-davis: @VincentBel I see - thanks for the clarification! Didn't realise that both methods were supported here.

I've updated the typings to support both of the use-cases. I've still removed partial - using partial here means that all properties of P are possibly undefined when being consumed by the child.

To keep type-safety the props have to be either of the variety where P is merged with the overall props, or provided as the prop named props. I can't see why you'd ever need to use both at the same time - and providing, for example, foo & props: { foo } would lead to undefined behaviour.

New type def is as follows:

```
export type BaseFieldArrayProps<P = {}, FieldValue = any> = (P | { props: P}) & _BaseFieldArrayProps<P, FieldValue>;

interface _BaseFieldArrayProps<P = {}, FieldValue = any> {
    name: string;
    component: ComponentType<WrappedFieldArrayProps<FieldValue> & P>;
    validate?: Validator | Validator[];
    warn?: Validator | Validator[];
    withRef?: boolean;
    rerenderOnEveryChange?: boolean;
}
```

So after defining a custom field array like so:

```
interface MyFieldValue {
    num: number;
}
interface MyFieldArrayCustomProps {
    foo: string;
    bar: number;
}

const MyCustomArrayField: React.StatelessComponent<MyFieldArrayCustomProps & WrappedFieldArrayProps<MyFieldValue>> = ({
    children,
    fields,
    foo,
    bar
}) => null;

const FieldArrayCustom = FieldArray as new () => GenericFieldArray<MyFieldValue, MyFieldArrayCustomProps>;

```

We can then use it 

```
                            {/* Good */}
                            <FieldArrayCustom
                                name="field10"
                                component={ MyCustomArrayField }
                                props={{
                                    foo: 'bar',
                                    bar: 123
                                }}
                            />

                            {/* Good */}
                            <FieldArrayCustom
                                name="field10"
                                component={ MyCustomArrayField }
                                foo="bar"
                                bar={23}
                            />

                            {/* Property 'bar' is missing in type '{ foo: string; }' but required in type 'MyFieldArrayCustomProps'. */}
                            <FieldArrayCustom
                                name="field10"
                                component={ MyCustomArrayField }
                                props={{
                                    foo: 'bar'
                                }}
                            />
```

Thoughts?

VincentBel: Yeah, totally agree.

VincentBel:
@m-b-davis I see you have only updated the `v7` typings. Would you mind updating the `v8` typings as well?

m-b-davis:
> @m-b-davis I see you have only updated the `v7` typings. Would you mind updating the `v8` typings as well?

Done!

vincentbel (approved):
LGTM!

huwmartin (approved):

PR #35748, opened 2019-05-26
[types/google-apps-script] Editing Types manually

oshliaer:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://developers.google.com/apps-script/reference/

I'm getting issues like 
> Argument of type 'String' is not assignable to parameter of type 'string'.
  'string' is a primitive, but 'String' is a wrapper object. Prefer using 'string' when possible.

For a quick changes find/replace is suitable I think.

Also It's hard to get serviceable app from https://github.com/grant/google-apps-script-dts

So commit it as it is. Please.


grant (approved):
Some of these changes have probably already been integrated. Thanks for the pr.

PR #35752, opened 2019-05-26
Added amazon-connect-streams from v1.3

andyhopp:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/amazon-connect-streams/index.d.ts
+++ b/types/amazon-connect-streams/index.d.ts
@@ -1,5 +1,5 @@
 // Type definitions for non-npm package Amazon Connect Streams API 1.3
-// Project: amazon-connect-streams
+// Project: https://github.com/aws/amazon-connect-streams
 // Definitions by: Andy Hopper <https://github.com/andyhopp>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.4
--- a/types/amazon-connect-streams/index.d.ts
+++ b/types/amazon-connect-streams/index.d.ts
@@ -1,0 +1,688 @@
+// Type definitions for non-npm package Amazon Connect Streams API 1.3
+// Project: amazon-connect-streams
+// Definitions by: Andy Hopper <https://github.com/andyhopp>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.4
+declare namespace connect {
+    /**
+     *
+     * A callback to receive agent details
+     *
+     * @param agent An Agent object containing information about the currently
+     * signed-in agent.
+     */
+    type AgentCallback = (agent: Agent) => void;
+
+    /**
+     * Register a callback to receive agent details
+     *
+     * @param callback A callback that will receive an {Agent} instance
+     *                 when agent information becomes available.
+     */
+    function agent(callback: AgentCallback): void;
+
+    /**
+     *
+     * A callback to receive contact details
+     *
+     * @param agent A Contact object containing information about the current contact.
+     */
+    type ContactCallback = (contact: Contact) => void;
+
+    /**
+     * Register a callback to receive contact details
+     *
+     * @param callback A callback that will receive a Contact instance
+     *                 when contact information is updated.
+     */
+    function contact(callback: ContactCallback): void;
+
+    /**
+     * Binds the given instance object as the context for
+     * the method provided.
+     *
+     * @param scope The instance object to be set as the scope
+     *              of the function.
+     * @param method The method to be encapsulated.
+     *
+     * All other arguments, if any, are bound to the method
+     * invocation inside the closure.
+     *
+     * @return A closure encapsulating the invocation of the
+     *    method provided in context of the given instance.
+     */
+    function hitch(scope: object, method: () => any): void;
+
+    interface Core {
+        initCCP(containerDiv: HTMLElement, options: InitCCPOptions): void;
+    }
+    let core: Core;
+
+    interface SoftPhoneOptions {
+        /*
+        * Whether to disable the ringtone.
+        */
+        disableRingtone?: boolean;
+        /*
+        * Whether to display the softphone in a frame.
+        */
+        allowFramedSoftphone?: boolean;
+        /*
+        * A URL for a custom ringtone.
+        */
+        ringtoneUrl?: string;
+     }
+
+    interface InitCCPOptions {
+        /*
+        * The URL for the Connect CCP.
+        */
+        ccpUrl: string;
+        /*
+        * Whether to display the login view.
+        */
+        loginPopup?: boolean;
+        /*
+        * Options specifying softphone configuration.
+        */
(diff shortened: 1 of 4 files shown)


andyhopp:
Forgot to check the boxes; fixed.

types/amazon-connect-streams/index.d.ts
```diff
@@ -0,0 +1,688 @@
+// Type definitions for non-npm package Amazon Connect Streams API 1.3
+// Project: amazon-connect-streams
```
sheetalkamat: This needs to be link to the project without that we cannot verify if definitions actually match the structure you are defining here.

types/amazon-connect-streams/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6",
+            "dom"
+        ],
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "strictNullChecks": false,
```
sheetalkamat: All new package definitions need to enable `strictNullChecks` so please enable and fix any errors.

sheetalkamat (changes requested):

types/amazon-connect-streams/index.d.ts
```diff
@@ -0,0 +1,688 @@
...
+// Definitions by: Andy Hopper <https://github.com/andyhopp>
```
sheetalkamat: There is already definition at https://github.com/aws/amazon-connect-streams/blob/master/src/index.d.ts
Since this is shipping with definition, I don't think you need them here.

andyhopp: There isn't an NPM package for this library yet; I'm the author of that definition (used my work account) and have added it to DT in order to make it easier for developers

types/amazon-connect-streams/tsconfig.json
andyhopp: Sorry! I had disabled it for the test script. Re-enabled and tweaked the test file.

types/amazon-connect-streams/index.d.ts
andyhopp: I have amended the .d.ts Project header to include a link to the amazon-connect-streams project.

andyhopp:
Hello! I believe I've addressed the reviewer comments. Since this package is not yet available on NPM, I wanted to give site developers a way to integrate the typings through the @types scope. I've updated the header to point to the GitHub repo for angular-connect-streams and resolved the strict null checks issue by refactoring the test script.

andyhopp:
@sheetalkamat Sorry, didn't "@" you... Please see above (and thanks for reviewing!).

types/amazon-connect-streams/index.d.ts
sheetalkamat: Since the definitions are available where the source is, I don't think we need definitions on DT. https://github.com/aws/amazon-connect-streams#downloading-streams shows the steps to get the js files, .d.ts file should be available along with that?

andyhopp: My chief concern is that, since we do not yet publish amazon-connect-streams as an NPM package, there are a couple of manual steps the developer would need to perform to modify their tsconfig to import the .d.ts file. Using the already-idiomatic @types scope simplifies the developer experience and is already a familiar process for developers using external JS libraries.
 
Once we start publishing ACS to NPM, we would, of course, remove the DT package.
 
Thanks again!

sheetalkamat: Adding @RyanCavanaugh and @sandersn for their opinion on hosting the types while they are still available elsewhere.

andyhopp: I should add; if having the types in the ACS GH repo is the issue, I could certainly remove them (I'd prefer *not* to, of course, but I'm willing to be persuaded otherwise!). I'm just trying to provide developers with a simple/easy way to code against ACS with Typescript while I drum up support for an official NPM package.

sheetalkamat: @andyhop making typings available easily seems to make sense. We see that your steps for install are to clone the repro, build and then copy js. Why not copy `.d.ts` as well next to the .js file? When the package is ready to ship, you can add `typings` field to package json pointing to the `.d.ts`

andyhopp: My concern there is that this becomes an additional step the developer needs to perform (modifying the tsconfig file to add the types). Most TS developers are familiar with the default behavior (to wit: including node_modules/@types), but once you start hacking the types property the developer must explicitly include that as well as the other types.

I certainly could work around this by instructing the developer to manually create a @types/amazon-connect-streams sub-folder, but I feel like that's an abuse.

It sounds like we're over-indexing on the fact that I included the types in the JS repo (which is officially only the JS library). I'll work with that team to address the fact that it's not published to NPM, but in the meantime, I really would like to keep the number of manual steps they need to perform to a minimum. I'm currently unhappy with the fact that users have to run a makefile to create the .js file for the library (again, working with the team to resolve that), and I'd be even unhappier with them having to know how to tweak up their tsconfig file (speaking from experience: the less you touch that file, the better).

Would deleting the index.d.ts file from the amazon-connect-streams repo resolve this to everyone's satisfaction?

sheetalkamat: I think the main concern is that we don't understand the extra step you are talking about. Why is copying the .d.ts file next to `.js` file while you are copying it difficult compared to installing types package?

andyhopp: Because if the developer is working with something like React or Angular, the location of the external javascript libraries is not typically under the project folder created by "ng new" or "create-react-app" and will therefore require modifying the tsconfig file to pull that file in. Performing an "npm i @types/xxx" is something I can expect most TS developers to be familiar with.

types/amazon-connect-streams/index.d.ts
```diff
@@ -0,0 +1,688 @@
...
+        * Whether to display the login view.
+        */
+        loginPopup?: boolean;
+        /*
+        * Options specifying softphone configuration.
+        */
+        softphone?: SoftPhoneOptions;
+     }
+
+     enum AgentStateType {
```
sheetalkamat: Is this really enum? Do the values like `AgentStateType` or `AgentAvailStates` etc exist at runtime. If not you want to declare this as union of string literal instead as :
```ts
type AgentStateType  = "init" | "routable" | "not_routable" | "offline";
```

sheetalkamat (changes requested):

types/amazon-connect-streams/index.d.ts
andyhopp: Hello! Yep, it is an enum; the expected usage per the ACS API docs is AgentStateType.INIT (ACS creates an enum object using the connect.makeEnum utility function).

andyhopp:
Thanks to everyone!!!

Issue #16474, opened 2017-05-11
[feature request] braintree-web-drop-in


buu700:
The Braintree v3 SDK now supports `braintree.dropin` as the v2 SDK did:

* https://github.com/braintree/braintree-web-drop-in

* https://developers.braintreepayments.com/guides/drop-in/javascript/v3

Edit: changed title for accuracy. My assumption had been that braintree-web-drop-in would need to be loaded via a side effect import and then a `dropin` property would be added to the exported value of braintree-web (as in their documented `<script>`-based examples), but actually this is used as an entirely separate module that just depends on braintree-web, so it would need to have its own independent type definition.

buu700:
Relevant reference documentation (just added for https://github.com/braintree/braintree-web-drop-in/issues/158): https://braintree.github.io/braintree-web-drop-in/docs/current

orta:
Hi thread, we're moving DefinitelyTyped to use [GitHub Discussions](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/53377) for conversations the `@types` modules in DefinitelyTyped.

To help with the transition, we're closing all issues which haven't had activity in the last 6 months, which includes this issue. If you think closing this issue is a mistake, please pop into the [TypeScript Community Discord](https://discord.gg/typescript) and mention the issue in the `definitely-typed` channel.

PR #35759, opened 2019-05-27
Add type definitions for braintree-web-drop-in

saoudrizwan:
---
---
These are the type definitions for Braintree's [`braintree-web-drop-in`](https://github.com/braintree/braintree-web-drop-in) library.

* https://github.com/braintree/braintree-web-drop-in/issues/157
* https://github.com/braintree/braintree-web/issues/429
* https://github.com/DefinitelyTyped/DefinitelyTyped/issues/16474


types/braintree-web-drop-in/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
```
sheetalkamat: new definitions cannot have anything except extends rule in tslint. Please fix any errors reported by linter after removing rules section,

types/braintree-web-drop-in/braintree-web-drop-in-tests.ts
```diff
@@ -0,0 +1,105 @@
+import * as dropin from "braintree-web-drop-in";
```
sheetalkamat: There is no package with this name. Also looking at `https://github.com/braintree/braintree-web-drop-in#full-example` doesn't seem like a module. You want to fix the tests and definitions to have global file declarations instead. 
Please refer to https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html for more information.

sheetalkamat (dismissed):

types/braintree-web-drop-in/braintree-web-drop-in-tests.ts
saoudrizwan: The package is here: https://www.npmjs.com/package/braintree-web-drop-in

Looking through TypeScript's handbook, a global package has "one or more assignments to `window.someName`", which `braintree-web-drop-in` does not. 

If it's any consolation, I've been using my modular type definitions with `braintree-web-drop-in` for almost all of its documented methods/properties, and I haven't run into any issues.

sheetalkamat: You still need to expose all this into Global since that's what the usage shows. Also add test to verify global usage as well. 

saoudrizwan: Thank you for the help 👌

saoudrizwan:
Hello @sheetalkamat, why did the typescript-bot fail [here](https://dev.azure.com/definitelytyped/DefinitelyTyped/_build/results?buildId=4331)? All I can see is a 
```
AssertionError [ERR_ASSERTION]: Required environment variable 'TYPESCRIPT_BOT_GITHUB_TOKEN' was not set
```
It looks like the 'Run comparisons' test failed, which doesn't make sense because there's nothing to compare to.

saoudrizwan:
Hello @sheetalkamat, I believe my PR is ready and passed all the checks

types/braintree-web-drop-in/index.d.ts
```diff
@@ -0,0 +1,130 @@
...
+}
+
+// Methods
+
+export function create(options: Options, callback: (error: object | null, dropin: Dropin | undefined) => void): void;
+export function create(options: Options): Promise<Dropin>;
+
+// Global
+
+declare global {
```
DanielRosenwasser: I think a more appropriate version of this would be

```ts
export as namespace braintree;
```

since it appears to be conditionally exposed as a global. Once that's fixed I think we can pull this in.

saoudrizwan: In a module, the user would have to import ;braintree-web-drop-in', and then access it like `dropin.create()`, but globally the user would access it as an extension of Braintree's main module like so `braintree.dropin.create()`
Using `declare global` I can make the `braintree` type declaration related to the global context, not to the current module.
`export as namespace braintree` wouldn't work here since `braintree` is not the module being exported, but `dropin` is.

DanielRosenwasser:
Thanks!

saoudrizwan:
Thanks for your help @DanielRosenwasser!

PR #35777, opened 2019-05-28
Update types for tedious authentication

csharpsi:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/tediousjs/tedious/blob/master/src/connection.js

### DIFF
--- a/types/tedious/index.d.ts
+++ b/types/tedious/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for tedious 3.0.0
+// Type definitions for tedious 4.0.0
 // Project: http://tediousjs.github.io/tedious/
 // Definitions by: Rogier Schouten <https://github.com/rogierschouten>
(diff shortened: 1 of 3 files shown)


types/tedious/index.d.ts
```diff
@@ -1,9 +1,10 @@
-// Type definitions for tedious 3.0.0
+// Type definitions for tedious 4.0.0
```
GuiAmPm: These are not breaking changes so there's no need to increment the major version, everything else is great.

csharpsi: I did intend to remove the deprecated properties so did this in anticipation, then decided I wouldn't do that and forgot to decrement the version number.

I will submit another PR to remove the deprecated properties.

csharpsi: Actually, looks like that's already been done

PR #35784, opened 2019-05-28
Add types for Google's closure-library.

ribrdb:
Please fill in this template.

If adding a new definition:


sheetalkamat (changes requested):
New definitions cannot  disable lint rules, Please fix them in tests as well as index.d.ts

ribrdb:
I don't really think that's feasible.
I ran Google's tool to convert the jsdoc into typescript types. The right place to fix these warnings would be in the jsdoc for the library itself, or in the conversion tool.
But I don't have the ability to change either of those. If I manually made the 8000+ fixes, which I obviously won't, the result would be so different from the autogenerated output that it wouldn't be practical to keep updated with new releases

sheetalkamat:
@ribrdb There seem to be lot of bad patterns allowed in your definitions and hence the reluctance to accept the definitions as is. (eg. usage of Object, Function). The library was last published 6 years ago. So it doesnt seem like its updated frequently.

Our team has few suggestions for you:
1. Include the types as part of library, then you don't have to rely on lint rules on DT.
2. Include the script you are using to convert the types along with the definitions(eg. `Lodash` does that)
3. If you really need to merge this as is, please include those rules in dtslint.json instead of per file, so they are more visible and clear.

cc: @RyanCavanaugh @sandersn

ribrdb:
A few things:
I'm not sure what to call this. I just noticed another lint message that says I should name this closure-library-browser if I'm going for the non-npm version. It is available on npm as google-closure-library, but the docs look like thats for usage with node and I've only tested the types for the browser and/or the closure compiler. The last release was two weeks ago, this is based off of the release from April.

Adding the types to the library itself seems possible, but it was requested in 2017 and hasn't gone anywhere: https://github.com/google/closure-library/issues/850

Including the script to generate the types is a good idea. I thought about doing that, but wasn't sure if it was allowed. I'll take a look at how lodash does it.

The usage of Object and Function in particular seem like issues with the library itself, especially older parts of the code. That really feels like something they need to figure out, and they've already got an issue for that: https://github.com/angular/clutz/issues/469

sandersn:
I would still name this google-closure-library and then make this work for node and the browser. It should be fairly simple. You'll need a single object to export, and then use `export=`.

```ts
export = goog;
export as namespace goog;
```

I haven't looked at the code yet since it doesn't display on github, but I expect you have some `goog` object suitable for export already. If not, you'll need to emit a wrapper around the generated code.

ribrdb:
I've found some more issues with the autogenerated file that our build system was hiding. I want to get those fixed before submitting it here. So I'll close this for now.

PR #35789, opened 2019-05-28
Support Yup nullable derived types

mauricedb:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: See below
- [x] ~Increase the version number in the header if appropriate.~
- [x] ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.~

It would be really nice if we can derive a TypeScript type alias from the Yup definition instead of having to redefine it. Using the `ReturnType` of the `cast()` function this is possible. The problem with this is that nullable defintions are lost. With this PR these are preserved.

Given the following Yup schema:
``` TypeScript
    firstName: yup.string(), // $ExpectType StringSchema<string>
    email: yup
        .string()
        .nullable()
        .email(),
    birthDate: yup
        .date()
        .nullable()
        .min(new Date(1900, 0, 1)),
    canBeNull: yup.string().nullable(true), // $ExpectType StringSchema<string | null>
    mustBeAString: yup
        .string()
        .nullable(true)
        .nullable(false),
    children: yup.array(yup.string()).nullable()
});
```

We can derive the `Person` type as follows:
``` TypeScript
type Person = ReturnType<typeof personSchema.cast>;
```

This is with this PR equivalent to:
``` TypeScript
type Person = {
    firstName: string;
    email: string | null;
    birthDate: Date | null;
    canBeNull: string | null;
    mustBeAString: string;
    children: string[] | null;
}
```

Without this PR it would lose the null types and be equivalent to:
``` TypeScript
type Person = {
    firstName: string;
    email: string;
    birthDate: Date;
    canBeNull: string;
    mustBeAString: string;
    children: string[];
}
```

Now we can create a new `Person` object as follows:
``` TypeScript
const person: Person = {
    firstName: "",
    email: null,
    birthDate: null,
    canBeNull: null,
    mustBeAString: "",
    children: null
};
```

### DIFF
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -71,7 +71,8 @@ export interface Schema<T> {
     withMutation(fn: (current: this) => void): void;
     default(value: any): this;
     default(): T;
-    nullable(isNullable?: boolean): this;
+    nullable(isNullable?: true): Schema<T | null>;
+    nullable(isNullable: false): Schema<Exclude<T, null>>;
     required(message?: TestOptionsMessage): this;
     notRequired(): this;
     typeError(message?: TestOptionsMessage): this;
@@ -169,7 +170,9 @@ export interface ArraySchema<T> extends Schema<T[]> {
     min(limit: number | Ref, message?: TestOptionsMessage): ArraySchema<T>;
     max(limit: number | Ref, message?: TestOptionsMessage): ArraySchema<T>;
     ensure(): ArraySchema<T>;
-    compact(rejector?: (value: T, index: number, array: T[]) => boolean): ArraySchema<T>;
+    compact(
+        rejector?: (value: T, index: number, array: T[]) => boolean
+    ): ArraySchema<T>;
 }
 
 export type ObjectSchemaDefinition<T extends object> = {
@@ -206,11 +209,11 @@ export interface ObjectSchema<T extends object> extends Schema<T> {
     constantCase(): ObjectSchema<T>;
 }
 
-export type TransformFunction<T> = ((
+export type TransformFunction<T> = (
     this: T,
     value: any,
     originalValue: any
-) => any);
+) => any;
 
 export interface WhenOptionsBuilder<T> {
     (value: any, schema: T): T;
@@ -230,7 +233,10 @@ export interface TestContext {
     parent: any;
     schema: Schema<any>;
     resolve: (value: any) => any;
-    createError: (params?: { path?: string; message?: string }) => ValidationError;
+    createError: (params?: {
+        path?: string;
+        message?: string;
+    }) => ValidationError;
 }
 
 export interface ValidateOptions {
@@ -297,7 +303,7 @@ export interface SchemaDescription {
     type: string;
     label: string;
     meta: object;
-    tests: Array<{ name: string, params: object }>;
+    tests: Array<{ name: string; params: object }>;
     fields: object;
 }
 
@@ -371,18 +377,16 @@ export class Ref {
 export interface Lazy extends Schema<any> {}
 
 export interface FormatErrorParams {
-  path: string;
-  type: string;
-  value?: any;
-  originalValue?: any;
+    path: string;
+    type: string;
+    value?: any;
+    originalValue?: any;
 }
 
-export type LocaleValue =
-    | string
-    | ((params: FormatErrorParams) => string);
+export type LocaleValue = string | ((params: FormatErrorParams) => string);
 
 export interface LocaleObject {
-    mixed?: { [key in keyof MixedSchema]?: string; } & { notType?: LocaleValue };
+    mixed?: { [key in keyof MixedSchema]?: string } & { notType?: LocaleValue };
     string?: { [key in keyof StringSchema]?: string };
     number?: { [key in keyof NumberSchema]?: string };
     boolean?: { [key in keyof BooleanSchema]?: string };
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -108,7 +108,7 @@ export interface StringSchemaConstructor {
     new (): StringSchema;
 }
 
-export interface StringSchema<T = string> extends Schema<T> {
+export interface StringSchema<T extends string | null = string> extends Schema<T> {
     length(limit: number | Ref, message?: TestOptionsMessage): StringSchema<T>;
     min(limit: number | Ref, message?: TestOptionsMessage): StringSchema<T>;
     max(limit: number | Ref, message?: TestOptionsMessage): StringSchema<T>;
@@ -133,7 +133,8 @@ export interface NumberSchemaConstructor {
     new (): NumberSchema;
 }
 
-export interface NumberSchema<T = number> extends Schema<T> {
+export interface NumberSchema<T extends number | null = number>
+    extends Schema<T> {
     min(limit: number | Ref, message?: TestOptionsMessage): NumberSchema<T>;
     max(limit: number | Ref, message?: TestOptionsMessage): NumberSchema<T>;
     lessThan(
@@ -169,7 +170,7 @@ export interface DateSchemaConstructor {
     new (): DateSchema;
 }
 
-export interface DateSchema<T = Date> extends Schema<T> {
+export interface DateSchema<T extends Date | null= Date> extends Schema<T> {
     min(
         limit: Date | string | Ref,
         message?: TestOptionsMessage
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -107,22 +107,24 @@ export interface StringSchemaConstructor {
     new (): StringSchema;
 }
 
-export interface StringSchema extends Schema<string> {
-    length(limit: number | Ref, message?: TestOptionsMessage): StringSchema;
-    min(limit: number | Ref, message?: TestOptionsMessage): StringSchema;
-    max(limit: number | Ref, message?: TestOptionsMessage): StringSchema;
+export interface StringSchema<T = string> extends Schema<T> {
+    length(limit: number | Ref, message?: TestOptionsMessage): StringSchema<T>;
+    min(limit: number | Ref, message?: TestOptionsMessage): StringSchema<T>;
+    max(limit: number | Ref, message?: TestOptionsMessage): StringSchema<T>;
     matches(
         regex: RegExp,
         messageOrOptions?:
             | TestOptionsMessage
             | { message?: TestOptionsMessage; excludeEmptyString?: boolean }
-    ): StringSchema;
-    email(message?: TestOptionsMessage): StringSchema;
-    url(message?: TestOptionsMessage): StringSchema;
-    ensure(): StringSchema;
-    trim(message?: TestOptionsMessage): StringSchema;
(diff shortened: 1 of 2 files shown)


types/yup/yup-tests.ts
```diff
@@ -594,3 +599,41 @@ yup.object<MyInterface>({
...
+// }
+
+const person: Person = {
+    firstName: "",
+    email: null,
+    birthDate: null,
+    canBeNull: null,
+    mustBeAString: "",
+    children: null
+};
```
sseppola: Would be nice to also ensure the properties can still be non-null, for example with some assignments like:
```
person.email = 'some@email.com';
person.birthDate = new Date();
person.children = ['1', '2', '3'];
```

sseppola (commented):
This looks like 🔥, love that we can derive the type from the schema with nullables!

As far as I can tell this PR is 👌, I'd just like a second opinion from another maintainer (@YashdalfTheGray perhaps?), and for the tests to confirm non-null values.

types/yup/index.d.ts
```diff
@@ -105,39 +107,49 @@ export interface StringSchemaConstructor {
     new (): StringSchema;
 }
 
-export interface StringSchema extends Schema<string> {
-    length(limit: number | Ref, message?: TestOptionsMessage): StringSchema;
-    min(limit: number | Ref, message?: TestOptionsMessage): StringSchema;
-    max(limit: number | Ref, message?: TestOptionsMessage): StringSchema;
+export interface StringSchema<T = string> extends Schema<T> {
```
YashdalfTheGray: I think, here, instead of defaulting it to string, we should say that `T extends string` because we're always going to want the user of `StringSchema` to pass us something that is derived from string specifically.

types/yup/index.d.ts
```diff
@@ -105,39 +107,49 @@ export interface StringSchemaConstructor {
...
-    min(limit: number | Ref, message?: TestOptionsMessage): NumberSchema;
-    max(limit: number | Ref, message?: TestOptionsMessage): NumberSchema;
-    lessThan(limit: number | Ref, message?: TestOptionsMessage): NumberSchema;
-    moreThan(limit: number | Ref, message?: TestOptionsMessage): NumberSchema;
-    positive(message?: TestOptionsMessage): NumberSchema;
-    negative(message?: TestOptionsMessage): NumberSchema;
-    integer(message?: TestOptionsMessage): NumberSchema;
-    truncate(): NumberSchema;
-    round(type: "floor" | "ceil" | "trunc" | "round"): NumberSchema;
+export interface NumberSchema<T = number> extends Schema<T> {
```
YashdalfTheGray: Same thing here, I think we should change `T = number` to `T extends number` because the second way locks the client from passing in some other type into this that might not be a number altogether.

types/yup/index.d.ts
```diff
@@ -146,16 +158,27 @@ export interface BooleanSchemaConstructor {
...
 
 export interface DateSchemaConstructor {
     (): DateSchema;
     new (): DateSchema;
 }
 
-export interface DateSchema extends Schema<Date> {
-    min(limit: Date | string | Ref, message?: TestOptionsMessage): DateSchema;
-    max(limit: Date | string | Ref, message?: TestOptionsMessage): DateSchema;
+export interface DateSchema<T = Date> extends Schema<T> {
```
YashdalfTheGray: Same here, `T extends Date`. This one matters less because users are _unlikely_ to have their own class derived from `Date` but it happens sometimes.

types/yup/index.d.ts
```diff
@@ -296,7 +324,7 @@ export interface SchemaDescription {
     type: string;
     label: string;
     meta: object;
-    tests: Array<{ name: string, params: object }>;
+    tests: Array<{ name: string; params: object }>;
```
YashdalfTheGray: I'm guessing some of these changes are like prettier changes?

types/yup/yup-tests.ts
```diff
@@ -594,3 +599,41 @@ yup.object<MyInterface>({
...
+    firstName: yup.string(), // $ExpectType StringSchema<string>
+    email: yup
+        .string()
+        .nullable()
+        .email(),
+    birthDate: yup
+        .date()
+        .nullable()
+        .min(new Date(1900, 0, 1)),
+    canBeNull: yup.string().nullable(true), // $ExpectType StringSchema<string | null>
```
YashdalfTheGray: Maybe have a boolean test case as well since we're adding types for `BooleanSchema` too?

types/yup/yup-tests.ts
```diff
@@ -594,3 +599,41 @@ yup.object<MyInterface>({
...
+const personSchema = yup.object({
+    firstName: yup.string(), // $ExpectType StringSchema<string>
+    email: yup
+        .string()
+        .nullable()
+        .email(),
+    birthDate: yup
+        .date()
+        .nullable()
+        .min(new Date(1900, 0, 1)),
```
YashdalfTheGray: It would also be cool to add type assertions to this as well so that we're checking one of each for the nullable case!

YashdalfTheGray (changes requested):

YashdalfTheGray:
> This looks like 🔥, love that we can derive the type from the schema with nullables!
> 
> As far as I can tell this PR is 👌, I'd just like a second opinion from another maintainer (@YashdalfTheGray perhaps?), and for the tests to confirm non-null values.

This is definitely clutch, have a couple of suggestions to add more type safety. Most 🔥part is `Schema<Exclude<T, null>>`! TIL!

types/yup/index.d.ts
mauricedb: Yes sorry, I have prettier with format on save enabled. You want me to revert these?

types/yup/index.d.ts
mauricedb: @YashdalfTheGray Looks like I need to change the definition to `interface StringSchema<T extends string | null = string> extends Schema<T>` because it can be null as well as a string type. And defaults to string is none specified.

types/yup/yup-tests.ts
mauricedb: 👍

types/yup/index.d.ts
mauricedb: 👍

types/yup/yup-tests.ts
mauricedb: 👍

mauricedb:
@YashdalfTheGray I am also looking at doing the same around `notRequired()`

types/yup/index.d.ts
YashdalfTheGray: I'm okay with this. What do y'all think? I like that we are forcing the client of this library to pass in something that has known string methods so that the validator will work properly but it might be too restrictive possibly? Thoughts?

types/yup/index.d.ts
YashdalfTheGray: Oh no. I like prettier being run on the code!

YashdalfTheGray:
> @YashdalfTheGray I am also looking at doing the same around `notRequired()`

[Aww yiss](https://media.giphy.com/media/VtcjgulX9Z6dW/giphy.gif)

types/yup/index.d.ts
sseppola: I think its fair to enforce string type here.
I didn't think of this before, but does 
`interface StringSchema<T extends string | null = string> extends Schema<T>`
prevent me from passing undefined into the schema? if so, should it be 
`interface StringSchema<T extends string | null | undefined = string> extends Schema<T>` instead?

mauricedb: @sseppola Yes it should prevent that. When using `--strictNullChecks` both `null` and `undefined` are distinctly different types. But that said when would you pass undefined into the `StringSchema` interface? As far as I can tell that should never happen at the moment. It will when I do the second PR with `notRequired()`/`required()` but I will do that as a separate PR.

sseppola: I tried to make a codesandbox to explore the default behavior of Yup:
https://codesandbox.io/s/formik-yup-form-dskdh

In summary
````
import * as Yup from "yup";

async function validate(value) {
  try {
    const res = await Yup.string().isValid(value);
    console.log(`"${value}" is valid`);
  } catch (error) {
    console.log(`"${value}" is not valid`);
  }
}
validate(undefined);
validate(null);
validate("");

```
the output is:
```
"undefined" is valid
"null" is not valid
"" is valid
```

So to be in line with Yup the type should be even more complicated:
`interface StringSchema<T extends string | null | undefined = string | undefined> extends Schema<T>`

However, as we go from `interface StringSchema extends Schema<string>`, which enforces string, then what you have currently is fine, and we can deal with `undefined` in another PR :)

sseppola (approved):

YashdalfTheGray (approved):

sandersn:
I guess that it would a more efficient to manually clone the Schema types, something like `StringSchema` and `StringNullableSchema`. Type parameters are a very general mechanism to turn a type into a factory of types, but you only ever need to construct two variants. 

I think it would also be easier to read for users of the library (especially JS users) to see `s.nullable() : StringNullableSchema` instead of `s.nullable() : StringSchema<string | null>`.

mauricedb:
@sandersn I am sure that copying the schema's to have a `StringSchema` and `StringNullableSchema` would be more efficient right now. However I would prefer not to do so. For one it would make removing `null` quite a bit harder with the `.nullable(false)` operation. And secondly it would result in just two copies of the schema now but in the next PR I am adding support for `undefined` through the `.required()` and `.notRequired()` operations exploding the possible schema's to: `StringSchema`, `StringNullableSchema`, `StringUndefinedSchema` and `StringNullableAndUndefinedSchema`. And that is just for `string`, the same happens for other types. Just imagine what happens if another option turns up, not that I am expecting one 😄

sandersn:
Yeah, four subtypes for every original type is a bit much. It might be easier to read if Schema took an extra `Nullable extends null | undefined = never` parameter and `StringSchema<Nullable = null | undefined extends never> extends Schema<string, Nullable>`, so that people would see `StringSchema<null>` or `StringSchema<never>`. But I'm not sure that's less confusing, actually.

YashdalfTheGray:
What if we used parameterized interfaces? Like `Nullable<string>`, that way, it could let us compose them like `Nullable<Undefinable<string>>` or something like that. I do like the idea of making these aliases though so we're not putting a huge thing in the parameter argument.

amcasey:
@YashdalfTheGray @mauricedb Do you want to revise this or merge it as-is?

YashdalfTheGray:
I'll defer to @mauricedb for it. I know he's got another PR in the works so I'm confident that these changes will make it into the repo in the near future. I think if he doesn't reply, a fallback is to just merge this as it is.

mauricedb:
Sorry for the delay, been a busy week with two conferences.

I would prefer to merge this as is. I am doing a second PR after this to add required/notRequired with undefined and will see if about updating the generic types and how that affects `Schema<Exclude<T, null>>`.

YashdalfTheGray:
🎉 Nice! @amcasey, we'll just merge this as is then!

PR #35808, opened 2019-05-29
react: add PropsWithChildren to memo return type

MeLlamoPablo:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://reactjs.org/docs/react-api.html#reactmemo
- [ ] Increase the version number in the header if appropriate. _- Not appropiate_
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. _- Not making substantial changes_

### The issue

When used on a `FunctionComponent`, which allow children by default, `memo` prevents the component from receiving children:

```typescript
import React, {FC, memo} from "react"

interface MyComponentProps {
	title: string
}

const MyComponentBase: FC<MyComponentProps> = ({children, title}) =>
	<div style={{backgroundColor: "cyan"}}>
		<h3>{title}</h3>
		{children}
	</div>

const MyComponent = memo(MyComponentBase)

const App: FC = () => <MyComponent title="Hello!">
	Hello world!
</MyComponent>

export default App
```

yields:

```
TypeScript error in /home/pablo/Projects/react-memo-children-test/src/App.tsx(15,24):
Type '{ children: string; title: string; }' is not assignable to type 'IntrinsicAttributes & MyComponentProps'.
  Property 'children' does not exist on type 'IntrinsicAttributes & MyComponentProps'.  TS2322
```

Even though this is perfectly valid React code. Nowhere on the react docs it says that memo-ized component can't accept children and in fact memo allows you to decide whether or not to update based on changes on the children.

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -784,7 +784,7 @@ declare namespace React {
     function memo<P extends object>(
         Component: SFC<P>,
         propsAreEqual?: (prevProps: Readonly<PropsWithChildren<P>>, nextProps: Readonly<PropsWithChildren<P>>) => boolean
-    ): NamedExoticComponent<P>;
+    ): NamedExoticComponent<PropsWithChildren<P>>;
     function memo<T extends ComponentType<any>>(
         Component: T,
         propsAreEqual?: (prevProps: Readonly<ComponentProps<T>>, nextProps: Readonly<ComponentProps<T>>) => boolean
--- a/types/react/test/tsx.tsx
+++ b/types/react/test/tsx.tsx
@@ -208,12 +208,14 @@ componentWithBadLifecycle.componentDidUpdate = (prevProps: {}, prevState: {}, sn
 
 const Memoized1 = React.memo(function Foo(props: { foo: string }) { return null; });
 <Memoized1 foo='string'/>;
+<Memoized1 foo='string'>Children</Memoized1>;
 
 const Memoized2 = React.memo(
     function Bar(props: { bar: string }) { return null; },
     (prevProps, nextProps) => prevProps.bar === nextProps.bar
 );
 <Memoized2 bar='string'/>;
+<Memoized2 bar='string'>Children</Memoized2>;
 
 const Memoized3 = React.memo(class Test extends React.Component<{ x?: string }> {});
 <Memoized3 ref={ref => { if (ref) { ref.props.x; } }}/>;


eps1lon:
Would this add `children` to

```tsx
function NoChildren({ label }: { label: string }) {
  return <nav aria-label={label}><NavItems /></nav>;
}

export default React.memo(NoChildren);
```
? If so then I'm not sure we want to pursue this. I guess most of the core maintainers want implicit children gone anyway. Extending this before we actually introduce the breaking change seems like a bad idea.

MeLlamoPablo:
Surprisingly, it doesn't:

```typescript
import React, {FC, memo} from "react"

interface MyComponentProps {
	title: string
}

const MyComponentBase = ({title}: MyComponentProps) =>
	<div style={{backgroundColor: "cyan"}}>
		<h3>{title}</h3>
	</div>

const MyComponent = memo(MyComponentBase)

const App: FC = () => <MyComponent title="Hello!">
	Hello world!
</MyComponent>

export default App
```

yields:

```
TypeScript error in /home/pablo/Projects/react-memo-children-test/src/App.tsx(23,10):
Type '{ children: (string | number | Element)[]; title: string; }' is not assignable to type 'IntrinsicAttributes & MyComponentProps'.
  Property 'children' does not exist on type 'IntrinsicAttributes & MyComponentProps'.  TS2322
```

However, I must ask for advice, because I'm trying to make a test case for it, but I can't get it to work.
If I add an error expectation:

```typescript
const Memoized8 = React.memo(
    ({ foo }: { foo: string }) => <div>{foo}</div>
);
<Memoized8 foo='string'>Children</Memoized8>; // $ExpectError
```

The tests fail with:

```
1> Error: /home/pablo/Projects/DefinitelyTyped/types/react/test/tsx.tsx:256:1
1> ERROR: 256:1   expect                  TypeScript@3.2: Expected an error on this line, but found none.
```

If I don't add the assertion:

```typescript
const Memoized8 = React.memo(
    ({ foo }: { foo: string }) => <div>{foo}</div>
);
<Memoized8 foo='string'>Children</Memoized8>;
```

The tests fail with:

```
1> Error: /home/pablo/Projects/DefinitelyTyped/types/react/test/tsx.tsx:256:2
1> ERROR: 256:2  expect  TypeScript@next compile error: 
1> Type '{ children: string; foo: string; }' is not assignable to type 'IntrinsicAttributes & { foo: string; }'.
1>   Property 'children' does not exist on type 'IntrinsicAttributes & { foo: string; }'.
1> 
1>     at /home/pablo/Projects/DefinitelyTyped/node_modules/dtslint/bin/index.js:189:19
1>     at Generator.next (<anonymous>)
1>     at fulfilled (/home/pablo/Projects/DefinitelyTyped/node_modules/dtslint/bin/index.js:5:58)
```

To reproduce, apply the following diff:

```diff
diff --git a/types/react/test/tsx.tsx b/types/react/test/tsx.tsx
index de9067cede..28379389bb 100644
--- a/types/react/test/tsx.tsx
+++ b/types/react/test/tsx.tsx
@@ -248,6 +248,13 @@ const Memoized6: React.NamedExoticComponent<object> = React.memo(props => null);
 // // $ExpectType boolean
 // Memoized7.type.defaultProps.test;
 
+// Make sure that memo doesn't add implicit children to a component that doesn't
+// have them:
+const Memoized8 = React.memo(
+    ({ foo }: { foo: string }) => <div>{foo}</div>
+);
+<Memoized8 foo='string'>Children</Memoized8>; // $ExpectError
+
 const LazyClassComponent = React.lazy(async () => ({ default: ComponentWithPropsAndState }));
 const LazyMemoized3 = React.lazy(async () => ({ default: Memoized3 }));
 const LazyRefForwarding = React.lazy(async () => ({ default: Memoized4 }));
```

to commit 0325042.

eps1lon:
It looks like `typescript@3.2` would not throw an error there but `typescript@next` would. Does it pass if you expect an error there and bump the minimal ts version to 3.4 (or 3.3)? Not saying we should bump the minimal version. Only want to see if this is only caught by typescript alphas.

MeLlamoPablo:
How do I test a single package? If I bump the version, I get:

```
Error: airbnb-prop-types depends on react but has a lower required TypeScript version.
    at checkTypeScriptVersions (/home/pablo/Projects/DefinitelyTyped/node_modules/types-publisher/src/check-parse-results.ts:51:23)
    at Object.checkParseResults [as default] (/home/pablo/Projects/DefinitelyTyped/node_modules/types-publisher/src/check-parse-results.ts:19:5)
    at <anonymous>
```

I tried bumping all packages with:

```bash
find ./types -type f -name index.d.ts -exec sed -i -E "s/\/\/ TypeScript Version: [0-9]+\.[0-9]+/\/\/ TypeScript Version: 3.3/g" {} +
```

but many have lint errors, and running the tests seems like would take hours.

ferdaber:
Even though it's not working with certain versions of TS, it does seem like the intention of this PR is to add children to memoized components. That really shouldn't be added for the reasons Sebastian outlined above. If the props of the inner component already contained explicit children prop, then the memorized component will also allow children.

MeLlamoPablo:
@ferdaber, I agree with @eps1lon in that adding children to a component that doesn't explicitly accept them is not a good idea, but the source of the problem is that `FunctionComponent` adds the children property. Ideally, we would have `FunctionComponent` and `FunctionComponentWithChildren` (the latter being opt-in if maintainers want implicit children gone).

In our codebase, we use `FC` and don't explicitly add the children prop because we expect that a component typed as `FC` will have it. My intention with this PR was to honor that expectation.

In my opinion, having `FC<{ children?: ReactNode }>` is redundant. If the intention was that developers would explicitly allow children on components that accept them, then `FunctionComponent` shouldn't have had the children property in the first place.

ferdaber:
That's correct and I agree with you that FC is not in its best place. However we weighed out the potential massive breaking change it would be to remove `children` from `FunctionComponent` and found that the best course of action for now is to not add to the cruft. Adding `children` to `memo` will only increase the technical debt once we do start to remove the `children` property.

eps1lon:
> Adding children to memo will only increase the technical debt once we do start to remove the children property.

I don't think this is about adding `children`. It's about not removing `children` in `memo` which is happening from the user perspective. At least that's how I understood it:

```tsx
const MyComponentBase: React.FC<{ foo: string }> = () => null;
<MyComponentBase>accepted at compile time, not handled at runtime</MyComponentBase>;

const MyComponent = memo(MyComponentBase);
<MyComponentBase>now it's rejected, why?</MyComponentBase>;
```

Maybe it's not possible to implement not removing it without also adding it. It's not a very "nice" way of teaching to use explicit children though.

ferdaber:
I suppose we could add an overload signature that catches `FC`s specifically, which is technically what this change does-- in that case I'd like a signature that actually accepts any JSX constructor function that doesn't use `PropsWithChildren` with this PR.

ferdaber (changes requested):
Requesting changes based on my last comment if we want to move forward with this.

PR #35816, opened 2019-05-29
Luxon@1.15: Fixes type of ZoneOffsetOptions

mdanka:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: The option here is not `localeCode` but `locale`: http://moment.github.io/luxon/docs/class/src/zone.js~Zone.html#instance-method-offsetName

### DIFF
--- a/types/luxon/luxon-tests.ts
+++ b/types/luxon/luxon-tests.ts
@@ -35,11 +35,13 @@ testIanaZone.formatOffset(dt.toMillis()); // $ExpectError
 testIanaZone.formatOffset(dt.toMillis(), 'narrow'); // $ExpectType string
 testIanaZone.formatOffset(dt.toMillis(), 'short'); // $ExpectType string
 testIanaZone.formatOffset(dt.toMillis(), 'techie'); // $ExpectType string
+testIanaZone.formatOffset(dt.toMillis(), 'other_string'); // $ExpectError
 testIanaZone.offsetName(dt.toMillis()); // $ExpectError
 testIanaZone.offsetName(dt.toMillis(), { format: 'short'}); // $ExpectType string
 testIanaZone.offsetName(dt.toMillis(), { format: 'long'}); // $ExpectType string
+testIanaZone.offsetName(dt.toMillis(), { format: 'other_string'}); // $ExpectError
 testIanaZone.offsetName(dt.toMillis(), { format: 'short', locale: 'en-us'}); // $ExpectType string
-testIanaZone.offsetName(dt.toMillis(), { format: 'short', locale: 'en-gb'}); // $ExpectType string
+testIanaZone.offsetName(dt.toMillis(), { locale: 'en-gb'}); // $ExpectType string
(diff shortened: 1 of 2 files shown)


mdanka:
@colbydehart @FourwingsY @jsiebern @mastermatt @pietrovismara @dawnmist @ycmjason - any chance you could review [this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35816) in the next few days? Thank you!

types/luxon/luxon-tests.ts
```diff
@@ -27,21 +27,26 @@ const fromObject = DateTime.fromObject({
...
+IANAZone.isValidSpecifier('Europe/London');
+IANAZone.isValidZone('Europe/London');
 IANAZone.resetCache();
-testIanaZone.formatOffset(dt.toMillis(), 'narrow'); // => +1
-testIanaZone.formatOffset(dt.toMillis(), 'short'); // => +01:00
-testIanaZone.formatOffset(dt.toMillis(), 'techie'); // => +0100
+testIanaZone.formatOffset(dt.toMillis(), 'narrow'); // $ExpectType string
+testIanaZone.formatOffset(dt.toMillis(), 'short'); // $ExpectType string
+testIanaZone.formatOffset(dt.toMillis(), 'techie'); // $ExpectType string
+testIanaZone.offsetName(dt.toMillis()); // $ExpectType string
```
pietrovismara: Why no options parameter here? Either you pass the param or use `$ExpectError`, since calling` offsetName` without options throws `Uncaught TypeError: Cannot read property 'format' of undefined`

types/luxon/luxon-tests.ts
```diff
@@ -27,21 +27,26 @@ const fromObject = DateTime.fromObject({
...
+testIanaZone.offsetName(dt.toMillis(), { format: 'short'}); // $ExpectType string
+testIanaZone.offsetName(dt.toMillis(), { format: 'long'}); // $ExpectType string
+testIanaZone.offsetName(dt.toMillis(), { format: 'short', locale: 'en-us'}); // $ExpectType string
+testIanaZone.offsetName(dt.toMillis(), { format: 'short', locale: 'en-gb'}); // $ExpectType string
```
mastermatt: 👍

mastermatt (approved):
Wow, this bug has been around for two years. Thanks for the update.

types/luxon/luxon-tests.ts
mastermatt: This is a weird case. Technically only Local and IANA zones need/use the options. I think that's the reason why it's marked optional in the types. That being said all the zones are meant to share a common interface. 
I think it's sane case to make that `options` on [this line](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/8a6de885da2ebbb895985eea03f385a84cad7eaa/types/luxon/index.d.ts#L503) should not be optional.

mastermatt (approved):

PR #35830, opened 2019-05-30
Add type definitions for expression-evaluator package

arturarushanyan:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/expression-evaluator/index.d.ts
+++ b/types/expression-evaluator/index.d.ts
@@ -1,0 +1,23 @@
+// Type definitions for expression-evaluator 0.2
+// Project: https://github.com/GreetzNL/expression-evaluator#readme
+// Definitions by: GreetzNL <https://github.com/GreetzNL>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export as namespace expressionEvaluator;
+
+export function evaluate(q: IComparisonOperator | ILogicalOperator, props?: any): boolean;
+export function and(query: ILogicalOperator, props: any): boolean;
+export function or(query: ILogicalOperator, props: any): boolean;
+export function eq(field: IComparisonOperator, props: any): boolean;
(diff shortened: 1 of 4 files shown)


types/expression-evaluator/index.d.ts
```diff
@@ -0,0 +1,23 @@
+// Type definitions for expression-evaluator 0.2
+// Project: https://github.com/GreetzNL/expression-evaluator#readme
```
sheetalkamat: Please provide the link to repo that works. This link is not accessible.

types/expression-evaluator/tslint.json
```diff
@@ -0,0 +1,6 @@
+{ 
+    "extends": "dtslint/dt.json",
```
sheetalkamat: New definition packages cannot have anything in .json file other than extends clause. Please remove the rules from there and fix the errors.

sheetalkamat (changes requested):

types/expression-evaluator/index.d.ts
arturarushanyan: Thanks for feedback. The link to repo is working within our organisation. It doenen't work for you because of you are not a contributor of GreetzNL. Is it ok to keep this like that?

sheetalkamat: If that's the case why not host the definitions inside the package itself.
Without working project link, we cannot verify the definitions represent correct structure or not.

arturarushanyan: We're doing similar thing now. We are using deprecated Typings package and keep all .d.ts fils there. But now we decided to upgrade to @types.
We will discuss with our team and organization what we can do about the link and I will update you.
Thanks

arturarushanyan:
We ended up sticking with what we have right now, and our organization ended up not making the sources publicly available. 
So closing this PR now.
Anyway, thanks and keep rocking guys :)

PR #35838, opened 2019-05-30
[recharts] fix React.ReactElement

morbalint:


morbalint:
recharts has a peer dependency on react ( `"react": "^15.0.0 || ^16.0.0"`,) and the typings on react 15.5.X has a required type argument for `ReactElement<T>`, which is missing in this typing of recharts. This PR was aiming to fix this issue, but then I realized that the versioning of typings in this repository has way too much problems, and abandoned this repository altogether.

PR #35840, opened 2019-05-30
babel__traverse: allow undefined scope for both traverse functions

coffee-cup:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/babel/babel/blob/master/packages/babel-traverse/src/index.js#L17
- [ ] ~Increase the version number in the header if appropriate.~

The scope can still be undefined when calling `traverse` with a non-default type parameter.

### DIFF
--- a/types/babel__traverse/index.d.ts
+++ b/types/babel__traverse/index.d.ts
@@ -11,8 +11,20 @@ import * as t from "@babel/types";
 
 export type Node = t.Node;
 
-export default function traverse<S>(parent: Node | Node[], opts: TraverseOptions<S>, scope: Scope, state: S, parentPath?: NodePath): void;
-export default function traverse(parent: Node | Node[], opts: TraverseOptions, scope?: Scope, state?: any, parentPath?: NodePath): void;
+export default function traverse<S>(
+    parent: Node | Node[],
+    opts: TraverseOptions<S>,
+    scope: Scope | undefined,
+    state: S,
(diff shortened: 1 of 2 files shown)


coffee-cup:
I am not sure if this change warrants an update to the contributors comment.

types/babel__traverse/index.d.ts
```diff
@@ -11,8 +11,20 @@ import * as t from "@babel/types";
 
 export type Node = t.Node;
 
-export default function traverse<S>(parent: Node | Node[], opts: TraverseOptions<S>, scope: Scope, state: S, parentPath?: NodePath): void;
-export default function traverse(parent: Node | Node[], opts: TraverseOptions, scope?: Scope, state?: any, parentPath?: NodePath): void;
+export default function traverse<S>(
```
amcasey: I'm probably missing something obvious, but why are there two overloads?  Why not just provide a default value for the type parameter of the first overload?

coffee-cup: I think because then you could do

```ts
traverse<StateType>(node, visitor)
```

In the visitor functions, the type of `state` would appear to be `StateType` even though no state was provided. Ideally, it would error if you specified a type parameter and did not pass in any state.

amcasey: Ah, I missed the extra question mark in the second overload.  Thanks!

amcasey (approved):

PR #35842, opened 2019-05-30
jasmine-given: make done arg non-optional

DmitryEfimenko:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/searls/jasmine-given/blob/master/app/js/jasmine-given.coffee#L71>
---
Having `done` argument as optional requires to have an `if` statement to check if it's defined:
```ts
Given((done) => {
  if (done) {
    done();
  }
});
```
However, this `if` statement is not actually required by *jasmine-given*. With new typings you can simply write:
```ts
Given((done) => {
  done();
});
```

### DIFF
--- a/types/jasmine-given/index.d.ts
+++ b/types/jasmine-given/index.d.ts
@@ -1,6 +1,7 @@
 // Type definitions for jasmine-given 2.6
 // Project: https://github.com/searls/jasmine-given
-// Definitions by: Shai Reznik <https://github.com/shairez>, Dmitry Efimenko <https://github.com/DmitryEfimenko>
+// Definitions by: Shai Reznik <https://github.com/shairez>
+//                 Dmitry Efimenko <https://github.com/DmitryEfimenko>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.1
 
--- a/types/jasmine-given/index.d.ts
+++ b/types/jasmine-given/index.d.ts
@@ -12,9 +12,9 @@ interface DoneFn {
   fail: (message?: Error | string) => void;
 }
 
-declare function Given(func: ((done: DoneFn) => void) | (() => void)): void;
-declare function When(func: ((done: DoneFn) => void) | (() => void)): void;
-declare function Then(func: ((done: DoneFn) => void) | (() => void)): void;
-declare function Then(label: string, func: ((done: DoneFn) => void) | (() => void)): void;
-declare function And(func: ((done: DoneFn) => void) | (() => void)): void;
-declare function Invariant(func: ((done: DoneFn) => void) | (() => void)): void;
+declare function Given(func: ((done: DoneFn) => void)): void;
+declare function When(func: ((done: DoneFn) => void)): void;
+declare function Then(func: ((done: DoneFn) => void)): void;
+declare function Then(label: string, func: ((done: DoneFn) => void)): void;
+declare function And(func: ((done: DoneFn) => void)): void;
+declare function Invariant(func: ((done: DoneFn) => void)): void;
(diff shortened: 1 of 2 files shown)


types/jasmine-given/index.d.ts
```diff
@@ -11,9 +12,9 @@ interface DoneFn {
   fail: (message?: Error | string) => void;
 }
 
-declare function Given(func: (done?: DoneFn) => void): void;
-declare function When(func: (done?: DoneFn) => void): void;
-declare function Then(func: (done?: DoneFn) => void): void;
-declare function Then(label: string, func: (done?: DoneFn) => void): void;
-declare function And(func: (done?: DoneFn) => void): void;
-declare function Invariant(func: (done?: DoneFn) => void): void;
+declare function Given(func: ((done: DoneFn) => void) | (() => void)): void;
```
demensky: Just use:

```ts
declare function Given(func: (done: DoneFn) => void): void;
```

This will work:

```ts
Given(() => { 
})
```

demensky (changes requested):

types/jasmine-given/index.d.ts
DmitryEfimenko: Indeed this would work and I was considering this.
However, this will result in intellisense suggesting (showing a tooltip with the signature of the method when you hover on `Given`) that the argument is required.
I wanted to make it explicit for documentation purposes.

@demensky What do you think?

shairez (approved):
LGTM

types/jasmine-given/index.d.ts
demensky: In typescript, in all cases, you can pass a function with fewer parameters.

Example:
You probably don’t use all `map` arguments every time you use it.

DmitryEfimenko: good point. Updated the code

types/jasmine-given/index.d.ts
```diff
@@ -11,9 +12,9 @@ interface DoneFn {
   fail: (message?: Error | string) => void;
 }
 
-declare function Given(func: (done?: DoneFn) => void): void;
-declare function When(func: (done?: DoneFn) => void): void;
-declare function Then(func: (done?: DoneFn) => void): void;
-declare function Then(label: string, func: (done?: DoneFn) => void): void;
-declare function And(func: (done?: DoneFn) => void): void;
-declare function Invariant(func: (done?: DoneFn) => void): void;
+declare function Given(func: ((done: DoneFn) => void)): void;
```
demensky: Let's remove the unnecessary `(` `)`.

types/jasmine-given/index.d.ts
```diff
@@ -1,7 +1,8 @@
 // Type definitions for jasmine-given 2.6
 // Project: https://github.com/searls/jasmine-given
-// Definitions by: Shai Reznik <https://github.com/shairez>
+// Definitions by: Shai Reznik <https://github.com/shairez>, Dmitry Efimenko <https://github.com/DmitryEfimenko>
```
demensky: Let's divide into separate lines for more convenient reading?

```ts
// Definitions by: Shai Reznik <https://github.com/shairez>
//                 Dmitry Efimenko <https://github.com/DmitryEfimenko>
```

demensky (changes requested):

types/jasmine-given/index.d.ts
DmitryEfimenko: Updated

types/jasmine-given/index.d.ts
DmitryEfimenko: oh, right. Updated as well

demensky (approved):

PR #35844, opened 2019-05-30
react-color - Implemented the types for v2.17.3 (latest in npm)

ErickTamayo:
Please fill in this template.

`npm run lint react-color` is currently giving an error, link to the issue: Cannot find module 'csstype'

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/casesandberg/react-color/tree/v2.17.3

### DIFF
--- a/types/react-color/index.d.ts
+++ b/types/react-color/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for react-color 2.17
+// Type definitions for react-color 3.0
 // Project: https://github.com/casesandberg/react-color/, http://casesandberg.github.io/react-color
 // Definitions by:  Karol Janyst <https://github.com/LKay>,
 //                  Marks Polakovs <https://github.com/markspolakovs>,
@@ -12,17 +12,11 @@
 
 import { ComponentClass, ClassAttributes, StatelessComponent, ReactNode } from "react";
 
-export interface HEXColor {
-    hex: string;
-    source?: string;
-}
-
 export interface HSLColor {
     a?: number;
     h: number;
     l: number;
     s: number;
-    source?: string;
 }
 
 export interface RGBColor {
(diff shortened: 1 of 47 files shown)


amcasey:
I hate to be the bearer of bad news, but I think the correct way to address this problem (it is definitely weird that the types have a newer version than the release on npm) is to introduce a "v2" folder and move these definitions into the subfolder.

amcasey:
There's some guidance here: https://github.com/DefinitelyTyped/DefinitelyTyped#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package

ErickTamayo:
Ok, I'll make the changes as soon as I can.

amcasey (approved):
Thanks for making those changes!

d4rekanguok:
Hi @ErickTamayo, thank you for updating the typing of `react-color`. This PR confuses me a bit however, would you mind helping me understand?

- Why was `ColorResult` renamed to `ColorState`?
- Definition for `ColorChangeHandler` was changed from `(color: ColorResult) => void` to `(color: Color) => void;`, [which is wrong](https://casesandberg.github.io/react-color/#api-onChange) & breaks my build.

@amcasey I have a question: the last PR that bumps this package versioning from `2.17` to `3.0` was not to match `react-color` v3.0 (which is still in development), but because of a [breaking change in typing](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33984). Thus, version 3.0 of this package is still meant for `react-color` 2.17. Now both `index.d.ts` and `/v2/index.d.ts` are meant for the same package... Going onward, how do we handle a breaking change for a `@types` package, without mismatching the version of the package it's meant for?

sandersn:
@d4rekanguok Regarding the version number, I think we should keep it the way it is now. It is extremely confusing for the DT version (3.0) to mismatch the package version (2.17), especially when there's a 3.0-beta as well. The current setup allows people to change the typings for 2.17 in v2, 3.0 in the root directory, and only has the downside that breaking changes can ship as patch versions. But all of DT has this problem.

The other break you found is best handled with a PR to Definitely Typed, which @ErickTamayo can then review.

dnlfrst:
Since it seems like people are very unhappy with my decision to increase the major version in the [previous PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33984), I'd like to know what the supposed version number would've been? Do you just ignore the presence of any breaking change in this package (which was due to a previous programming error and not induced by any changes in `react-color`) and update the patch version regardless? If so, how is a user of this package supposed to take notice of any breaking change when updating to the next patch version where he would not expect it due to semantic versioning?

sandersn:
Breaking changes to the `@types` package do have to be published as a patch version; major and minor version changes are reserved for matching the source package's version. That means if you care about exact type compatibility, you have to install an exact version of the types. It's a tradeoff between following semver and making it easy to guess which version of the types you should install. 
We decided it was more important that `npm install @types/node@10` install types for node 10, for example.

About your other question: The PR was merged because if nobody reviews changes to a package, then we assume (1) it works for the PR author (2) the community will be able to fix it if it does not.

hdra:
I'm also facing the same issue mentioned by @d4rekanguok . I also found that the `Color` is missing `string` in the type definition. Though this change also fixes some issue with `ColorWrap` that I was having. 

I was about to prepare a PR to fix these, but there are some things I'm not too clear about: 

- Am I supposed to fork and clone the entire DefinitelyTyped repository? Because thats a pretty massive repo
- How do I configure my project to pull the type definition from my fork to test it?

sandersn:
* Yes, that's the only way we have right now. If there's some git commands that would give you only types/react-color, types/react, and everything in the the root directory, you'd probably be OK too.
* A one-off way to copy the contents of types/react-color into node_modules/types/react-color. Or you can edit whatever's in node_modules/types/react-color and only copy it into the DT repo once you are confident it does what you want.

PR #35848, opened 2019-05-30
Remove PartialDeep from lodash, replace with Partial

sandersn:
Fixes the most common occurrence of microsoft/TypeScript#21592. 

I don't think anybody uses the full capability of `PartialDeep` as a iteratee parameter, and it doesn't really make sense as the return type of `pick`.

### DIFF
--- a/types/lodash/ts3.1/common/collection.d.ts
+++ b/types/lodash/ts3.1/common/collection.d.ts
@@ -1342,7 +1342,7 @@ declare module "../index" {
         /**
          * @see _.orderBy
          */
-        orderBy(iteratees?: Many<ListIterator<T, NotVoid> | PropertyName | PartialDeep<T>>, orders?: Many<boolean|"asc"|"desc">): Collection<T>;
+        orderBy(iteratees?: Many<ListIterator<T, NotVoid> | PropertyName | PartialObject<T>>, orders?: Many<boolean|"asc"|"desc">): Collection<T>;
     }
     interface Object<T> {
         /**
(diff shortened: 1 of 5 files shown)


aj-r:
Sorry for the late response - I'm worried this will break things for some people. It's probably fine to use `Partial<T>` as a return type (e.g. for `pick`), but I think it will be overly restrictive as a parameter type, especially for something like `IterateeShorthand` which is used in many places.

I think it would be safer to replace `PartialDeep<T>` with `object` when it's used as an input parameter. We lose some type safety, but I think that's better than breaking things.

Alternatively, we could keep most type safety without breaking things with something like this:

```ts
    type PartialDeep<T> = {
        [P in keyof T]?: any;
    };
```

or better yet:

```ts
    type PartialDeep<T> = {
        [P in keyof T]?: T[P] extends object ? object : T[P];
    };
```

We may want to make this change in the TS 2.X version, too, since microsoft/TypeScript#21592 seems to start in TS 2.7.

sandersn:
Yeah, we merged this too early because the team assumes that I know what I'm doing for `@types/lodash` which is not really true. :)

Anyway, I like your workaround ideas. The key here is whether we want to err on the side of

1. Too strict -- `Partial<T>`.
2. Too lenient -- (a) `object` or (b) `{ [P in keyof T]?: any }` or (c) `{ [P in keyof T]?: T[P] extends object ? object : T[P] }`.
3. Reliant on compiler quirks -- `{ [P in keyof T]?: T[P] extends never ? DeepPartial<T[P]> : DeepPartial<T[P]> }`.

I like (2.c). I'll try that first. (3) *works* but is likely to break on future versions of Typescript.

PR #35850, opened 2019-05-31
fix(hapi): make the auth settings property a separate type on routes

SimonSchick:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/hapijs/hapi/blob/74e523d68282838fbceca21d584df48ce6a3f181/lib/auth.js#L494

The `settings` property in general is hydrated or  'cleaned up' version the initial route options, there is more that needs fixing but this is a start :)

### DIFF
--- a/types/hapi__hapi/index.d.ts
+++ b/types/hapi__hapi/index.d.ts
@@ -201,6 +201,8 @@ export interface AuthCredentials {
     app?: AppCredentials;
 }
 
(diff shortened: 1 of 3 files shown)


SimonSchick:
@sheetalkamat ironically I messed something up (non breaking) but failed to fix the actual issue, I will create a follow up PR.

PR #35864, opened 2019-05-31
Improve the intellisense of lazypipe

neurolag:
Please fill in this template.

I've just noticed that there is no IntelliSense when using `lazypipe`.
This PR adds auto-completion to the `lazypipe`-module.

### DIFF
--- a/types/gulp-rev-replace/index.d.ts
+++ b/types/gulp-rev-replace/index.d.ts
@@ -2,7 +2,7 @@
 // Project: https://github.com/jamesknelson/gulp-rev-replace
 // Definitions by: Tanguy Krotoff <https://github.com/tkrotoff>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.2
+// TypeScript Version: 3.0
 
 /// <reference types="node" />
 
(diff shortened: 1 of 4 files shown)


manuth:
Uhm... Guys I need some help
This solution doesn't seem to work for methods with multiple overloads.

***test.d.ts***
```ts
import "node";
declare function example(name: string): NodeJS.ReadWriteStream;
declare function example(options: { name: string, author: string }): NodeJS.ReadWriteStream;
```
***index.ts***
```ts
import lazyPipe = require("lazypipe");
import test = require("./test");

// This here works
lazyPipe(test, { name: "", author: "" });
// While this here does not
lazyPipe(test, "");
```

manuth:
I've created an issue in the TypeScript repository: https://github.com/microsoft/TypeScript/issues/31716
I'll notify you guys as soon as my PR is ready.

amcasey:
@manuth, I believe that's a known (and intentional) limitation of TS's type inference - a more general system would be too expensive.  I think you might be able to work around the issue by using a union type `nameOrOptions: string | { name: string, author: string }`.

manuth:
@amcasey Well, ya see - actually `lazypipe.pipe(fn, args)` should take the arguments of `fn`, no matter what nature they have and what overloads `fn` has.

This means I sadly cannot work with union types.

Do you have any other idea?

amcasey:
Sorry, I blanked and read your example backwards - I thought it was `example` you were trying to get working, rather than `pipe`.

I think you're definition is correct and you're just hitting a limitation of the type system.

For now, I think you'll either need to use looser types (e.g. allowing `any`) or accept that some callers will see errors. 😞

manuth:
Alright - thanks for your support 🎉

PR #35871, opened 2019-05-31
mailgun-js: Allow sending templates

CodeLenny:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://documentation.mailgun.com/en/latest/user_manual.html#templating

---

The current definitions for `mailgun-js` don't allow sending templates or template data, so I've added the `template` field, and allowed `v:[variable name]` properties.


types/mailgun-js/index.d.ts
```diff
@@ -74,6 +75,11 @@ declare namespace Mailgun {
             "recipient-variables"?: BatchSendRecipientVars;
         }
 
+        interface SendTemplateData {
+            "template": string;
```
andipaetzold: We don't need these quotes here.

```suggestion
            template: string;
```

types/mailgun-js/index.d.ts
```diff
@@ -74,6 +75,11 @@ declare namespace Mailgun {
             "recipient-variables"?: BatchSendRecipientVars;
         }
 
+        interface SendTemplateData {
```
andipaetzold: The interface should also have the following properties from `SendData`: `from`, `to`, `cc`, `bcc`, `subject`, `attachment` and `inline`.
Especially `to` would be important since it is required to send the mail.

Could we do something like the following?
```typescript
interface SendDataBase {
    from?: string;
    to: string | string[];
    cc?: string;
    bcc?: string;
    subject?: string;
    attachment?: AttachmentData | ReadonlyArray<AttachmentData>;
    inline?: AttachmentData | ReadonlyArray<AttachmentData>;
}

interface SendData extends SendDataBase {
    text?: string;
    html?: string;
}

interface SendTemplateData extends SendDataBase {
    template: string;
    [templateVariable: string]: string;
}
```

andipaetzold (changes requested):
Thanks for your contribution!

We should use the existing types of `SendData` within `SendTemplateData` to have the same auto complete quality there.

CodeLenny:
@andipaetzold :man_facepalming: Yup, I intended to add `extends` but completely forgot.  I've implemented both suggested changes.

types/mailgun-js/index.d.ts
```diff
@@ -74,6 +75,11 @@ declare namespace Mailgun {
             "recipient-variables"?: BatchSendRecipientVars;
         }
 
+        interface SendTemplateData extends SendData {
```
andipaetzold: Let's just avoid using `any` by defining `SendTemplateData` as
```typescript
type SendTemplateData = SendData & {
    template: string;
    [templateVariable: string]: string;
}
```

Then I am happy to approve :)

CodeLenny: @andipaetzold Sorry for the delay, I've been really busy with other things.  I've implemented this fix.

andipaetzold (approved):

PR #35880, opened 2019-06-01
[next-nprogress] Add definitions

Ovyerus:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/next-nprogress/index.d.ts
+++ b/types/next-nprogress/index.d.ts
@@ -2,9 +2,10 @@
 // Project: https://github.com/sergiodxa/next-nprogress#readme
 // Definitions by: Ovyerus <https://github.com/Ovyerus>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
 
-import {NProgressOptions} from 'nprogress';
-import {Component, ComponentType} from 'react';
+import { NProgressOptions } from 'nprogress';
+import { Component, ComponentType } from 'react';
 
 export default function withNProgress<P extends object>(
     delayMs?: number,
--- a/types/next-nprogress/index.d.ts
+++ b/types/next-nprogress/index.d.ts
@@ -7,7 +7,7 @@
 import { NProgressOptions } from 'nprogress';
 import { Component, ComponentType } from 'react';
 
-export default function withNProgress<P extends object>(
+export default function withNProgress(
     delayMs?: number,
     options?: Partial<NProgressOptions>
-): (Page: ComponentType<P>) => ComponentType<P>;
+): <P extends object>(Page: ComponentType<P>) => ComponentType<P>;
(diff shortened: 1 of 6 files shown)


types/next-nprogress/index.d.ts
```diff
@@ -0,0 +1,13 @@
+// Type definitions for next-nprogress 1.4
+// Project: https://github.com/sergiodxa/next-nprogress#readme
+// Definitions by: Ovyerus <https://github.com/Ovyerus>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+import { NProgressOptions } from 'nprogress';
+import { Component, ComponentType } from 'react';
+
+export default function withNProgress<P extends object>(
```
amcasey: Do you have an example of a case where `P` is specified?  I could be mistaken, but I don't think it will be inferred.

Ovyerus: Thought it'd get inferred from line 13, but just did a test and I guess not. Any ideas on how to get it to?

amcasey: I think using type parameters exclusively in return types is discouraged for the very reason that they'll never be inferred (though there may be sophisticated ways of doing it that I just don't know).  I'm actually surprised you didn't get a lint error about it.

In any case, I think your options are to keep the type parameter and specify the type you want as a type argument at the call site or drop the type parameter and use a type assertion at the call site.  Generally speaking, the latter is preferred (as more idiomatic), but I can see why you'd want to specify just the variable part of the complicated type.

andrewbranch: Sorry to eavesdrop, but I think what you're looking for is simply to move the type parameter to the function that's returned: [Playground](https://www.typescriptlang.org/play/#src=type%20NProgressOptions%20%3D%20%7B%7D%0D%0A%0D%0Atype%20ComponentType%3CP%3E%20%3D%20%7B%0D%0A%20%20%20%20pretendImAComponent%3A%20boolean%2C%0D%0A%20%20%20%20props%3A%20P%0D%0A%7D%0D%0A%0D%0Adeclare%20function%20withNProgress(%0D%0A%20%20%20%20delayMs%3F%3A%20number%2C%0D%0A%20%20%20%20options%3F%3A%20Partial%3CNProgressOptions%3E%0D%0A)%3A%20%3CP%20extends%20object%3E(Page%3A%20ComponentType%3CP%3E)%20%3D%3E%20ComponentType%3CP%3E%3B%0D%0A%0D%0A%0D%0A%2F%2F%20Yes%20I%20know%20this%20is%20not%20how%20React%20works%0D%0Aconst%20App%20%3D%20%7B%0D%0A%20%20%20%20pretendImAComponent%3A%20true%2C%0D%0A%20%20%20%20props%3A%20%7B%20foo%3A%20'bar'%20%7D%0D%0A%7D%0D%0A%0D%0Aconst%20decoratedComponent%20%3D%20withNProgress()(App)%0D%0AdecoratedComponent.props.foo%3B)

Ovyerus: True

amcasey (approved):

PR #35889, opened 2019-06-01
[core-js] change Dict interface for strictNullChecks mode

mh-alahdadian:
Dict<T>[string] can have type of undefined and in strictNullChecks mode it should be checked


types/core-js/index.d.ts
```diff
@@ -337,9 +337,9 @@ declare var log: Log;
 // #############################################################################################
 
 interface Dict<T> {
-    [key: string]: T;
-    [key: number]: T;
-    // [key: symbol]: T;
+    [key: string]: T | undefined;
```
amcasey: Why just these?  Why not (e.g.) `get`, below?

MHA15: some more changes added

amcasey: I don't see an update?

MHA15: my bad, sorry

amcasey (approved):

PR #35890, opened 2019-06-01
[dompurify] Add typing for factory function

hgenru:
@See https://github.com/cure53/DOMPurify/issues/323

### DIFF
--- a/types/dompurify/index.d.ts
+++ b/types/dompurify/index.d.ts
@@ -2,27 +2,38 @@
 // Project: https://github.com/cure53/DOMPurify
 // Definitions by: Dave Taylor <http://davetayls.me>, Samira Bazuzi <https://github.com/bazuzi>, FlowCrypt <https://github.com/FlowCrypt>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-
 export as namespace DOMPurify;
+export = DOMPurify;
 
-export declare let version: string;
-export declare let removed: any[];
-export declare let isSupported: boolean;
+declare const DOMPurify: createDOMPurifyI
 
-export declare function sanitize(source: string | Node): string;
-export declare function sanitize(source: string | Node, config: Config & { RETURN_DOM_FRAGMENT?: false; RETURN_DOM?: false; }): string;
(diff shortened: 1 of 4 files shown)


types/dompurify/index.d.ts
```diff
@@ -3,26 +3,37 @@
 // Definitions by: Dave Taylor <http://davetayls.me>, Samira Bazuzi <https://github.com/bazuzi>, FlowCrypt <https://github.com/FlowCrypt>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export as namespace DOMPurify;
+export = createDOMPurify
```
amcasey: It looks like the new form isn't consumable as a global.  Was that intentional?

hgenru: @amcasey can you explain the difference? I am not familiar with typings, I just write as I know and test it.

amcasey: I think of the `as namespace` version as being a way to make the declarations available in non-modules (e.g. webpages).  There's a more thorough explanation here: https://www.typescriptlang.org/docs/handbook/modules.html#umd-modules

hgenru: Should I rename DOMPurify interface to DOMPurifyI and export createDOMPurify as DOMPurify namespace?

hgenru: I figured out. What you think, maybe we should add this into the FAQ?

quentez:
@hgenru Hi. Since updating this typings, we're not able to import the other interfaces that it previously contained anymore.

This worked before:

```
import DOMPurify, {Config, HookEvent, HookName} from 'dompurify';
```

Could this be because of this change?

hgenru:
@quentez, very likely. Config, HookEvent, HookName wasn't exported clearly and I have no idea why this worked before.

Sorry, I had no plans to break your code. Can you add a fix and new tests? It's will be very nice.

crimx:
Also break my code. Cannot import the namespace anymore. These types are useful when writing wrappers for dompurify.

crimx:
Changing to  `export declare interface DOMPurifyI {` seems to work. Can you verify that? @hgenru

PR #35893, opened 2019-06-02
Create types for logfmt

ebroder:
I've written up a type declaration for the [`logfmt` package](https://www.npmjs.com/package/logfmt). It has a pretty thorough README which seems to document all of the functionality that's exported from the main module of the package, so I used that to derive my tests.

If adding a new definition:

cc @csquared as the maintainer of the upstream package

### DIFF
--- a/types/logfmt/index.d.ts
+++ b/types/logfmt/index.d.ts
@@ -1,0 +1,70 @@
+// Type definitions for logfmt 1.2
+// Project: https://github.com/csquared/node-logfmt#readme
+// Definitions by: Evan Broder <https://github.com/ebroder>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
+/// <reference types="node" />
+
+import { IncomingMessage, ServerResponse } from 'http';
+
+type HTTPHandler =
+  (req: IncomingMessage, res: ServerResponse, next: (err?: any) => void) => any;
+
+interface WritableStream {
+    write(data: string): void;
+}
+
+interface RequestLoggerOptions {
+    immediate?: boolean;
+    elapsed?: string;
+}
+type RequestLoggerFormatter =
+  (req: IncomingMessage, res: ServerResponse) => object;
+
+interface RequestLogger {
+    (options?: RequestLoggerOptions, formatter?: RequestLoggerFormatter): HTTPHandler;
+    (formatter: RequestLoggerFormatter): HTTPHandler;
+    commonFormatter: (req: IncomingMessage, res: ServerResponse) => {
+        ip: string;
+        time: string;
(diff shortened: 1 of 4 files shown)


types/logfmt/logfmt-tests.ts
```diff
@@ -0,0 +1,117 @@
...
+
+logfmt.log({foo: "bar", a: 14, baz: 'hello kitty'});
+logfmt.log({foo: "bar", a: 14, baz: 'hello kitty'});
+logfmt.log({foo: "bar", a: 14, baz: 'hello kitty'}, process.stderr);
+
+// logfmt's trick of copying the prototype onto itself seems to trip up
+// TypeScript on module-level member assignment, so this should work but doesn't
+//
+// $ExpectError
+logfmt.stream = process.stderr;
```
ebroder: FTR, this is the one potential issue I'm aware of with these definitions. I couldn't figure out how to make the module-level `stream` (or `maxErrorLines`) variable writable without duplicating the class definition. Open to suggestions.

types/logfmt/index.d.ts
```diff
@@ -0,0 +1,66 @@
...
+    requestLogger: RequestLogger;
+
+    stream: WritableStream;
+    maxErrorLines: number;
+}
+
+// logfmt copies its prototype onto the exported class, so you can both use it
+// directly as an instance of the class and as a constructor to instantiate new
+// instances.
+declare const logfmt: Logfmt & typeof Logfmt;
```
DanielRosenwasser: Sorry about the delay, but thanks for your patience.

Instead of doing this, how about

```ts
interface Logfmt {
    stringify(data: object): string;
    parse(line: string): object;

    log(data?: object, stream?: WritableStream): void;
    time(label?: string): Logfmt;
    namespace(data: object): Logfmt;
    error(err: Error, id?: string): void;

    streamParser(options?: { end?: boolean; }): NodeJS.ReadWriteStream;
    streamStringify(options?: { delimiter?: string }): NodeJS.ReadWriteStream;

    bodyParser(options?: { contentType?: string }): HTTPHandler;
    bodyParserStream(options?: { contentType?: string }): HTTPHandler;

    requestLogger: RequestLogger;

    stream: WritableStream;
    maxErrorLines: number;
}

interface LogfmtStatic extends Logfmt {
    new (): Logfmt;
}

declare const logfmt: LogfmtStatic;
```

DanielRosenwasser (changes requested):

types/logfmt/index.d.ts
ebroder: Sure, that seems much cleaner :slightly_smiling_face:

Pushed a fix

DanielRosenwasser:
Thank you!

PR #35895, opened 2019-06-02
UUID:  reverse overloaded intersection so that ReturnType<T> will come up with string

scamden:

https://github.com/Microsoft/TypeScript/issues/24275#issuecomment-390701982

ReturnType defaults to the last overload's return type since it is assumed to be the most general. Unfortunately, this isn't true with intersections which are treated as overloads like these ones. I assume the string implementation is the much more common case and should probably be the default return type since we sadly have to choose. Use case: this is currently breaking our use of jest types at 23 and above as it uses ReturnType in the mock types.

### DIFF
--- a/types/uuid/interfaces.d.ts
+++ b/types/uuid/interfaces.d.ts
@@ -14,12 +14,12 @@ export type V4Options = {random: number[]} | {rng(): number[]};
 
 export type v1String = (options?: V1Options) => string;
 export type v1Buffer = <T extends OutputBuffer>(options: V1Options | null | undefined, buffer: T, offset?: number) => T;
-export type v1 = v1String & v1Buffer;
+export type v1 = v1Buffer & v1String;
 
 export type v4String = (options?: V4Options) => string;
 export type v4Buffer = <T extends OutputBuffer>(options: V4Options | null | undefined, buffer: T, offset?: number) => T;
-export type v4 = v4String & v4Buffer;
+export type v4 = v4Buffer & v4String;
 
 export type v5String = (name: string | number[], namespace: string | number[]) => string;
 export type v5Buffer = <T extends OutputBuffer>(name: string | number[], namespace: string | number[], buffer: T, offset?: number) => T;
-export type v5 = v5String & v5Buffer;
+export type v5 = v5Buffer & v5String;


cjbarth:
While I'm not thrilled about having to do this, it does seem that this would be the correct pattern given the current state of TypeScript. I wouldn't mind a second opinion on this though.

cjbarth (commented):
While I'm not thrilled about having to do this, it does seem that this would be the correct pattern given the current state of TypeScript. I wouldn't mind a second opinion on this though.

scamden:
@cjbarth Ya i agree it's not great. I've been playing locally with using standard overrides, which would give us the fully inclusive type for the final one but i'm having some trouble getting it right. i'll send a pr if i get it since that would probably be the more complete solution..

scamden:
@cjbarth ya as far as i can tell, it's impossible to get the more permissive return type without also getting a more permissive call signature.

```
export class UuidStatic {
  v4(options?: V4Options): string;
  v4<T extends OutputBuffer>(options: V4Options | null | undefined, buffer: T, offset?: number): T;
  v4<T extends OutputBuffer>(options: V4Options | null | undefined, buffer?: T, offset?: number): T | string {
    if (buffer) {
      return buffer;
    }
    return '' as any
  }
}

const uuid = new UuidStatic().v4(null, undefined, 2);  // fails as expected cause no matching call signature
```

in the above it seems we get the more permissive return type in the implementation signature but actually ReturnType still uses the last overload definition and not the final line. if i take those three signatures and put them on an interface instead, the ReturnType becomes more permissive but suddenly you can make incorrect calls as well like:

```
export interface UuidStatic {
  v4(options?: V4Options): string;
  v4<T extends OutputBuffer>(options: V4Options | null | undefined, buffer: T, offset?: number): T;
  v4<T extends OutputBuffer>(options: V4Options | null | undefined, buffer?: T, offset?: number): T | string;
}

const V4: UuidStatic['v4'] = () => '' as any;

const uuid = V4(null, undefined, 2);
```

it doesn't seem like with this particular function signature it's possible to get the permissive return type without the permissive call type.. pretty annoying. let me know if you see an alternative.

scamden:
@cjbarth can we merge this? it's a pretty small change

PR #35907, opened 2019-06-02
[@wordpress/data] add new definitions

dsifford:

**NOTE TO REVIEWER:** This is currently being flagged with `no-unnecessary-generics` warnings, but I'd argue that the generics being used are indeed necessary in these circumstances. How should we proceed here?

If adding a new definition:

### DIFF
--- a/types/wordpress__data/index.d.ts
+++ b/types/wordpress__data/index.d.ts
@@ -51,7 +51,7 @@ export interface StoreConfig<S> {
         [k: string]: (...args: readonly any[]) => any;
     };
     controls?: {
-        [k: string]: (action: Action) => Promise<any> | any;
+        [k: string]: (action: Action) => any;
     };
     initialState?: S;
 }
--- a/types/wordpress__data/index.d.ts
+++ b/types/wordpress__data/index.d.ts
@@ -1,0 +1,148 @@
+// Type definitions for @wordpress/data 4.5
+// Project: https://github.com/WordPress/gutenberg/tree/master/packages/data/README.md
+// Definitions by: Derek Sifford <https://github.com/dsifford>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.4
+
+import { ComponentType, Consumer, Provider, useContext } from "react";
+import { AnyAction as Action, combineReducers, Reducer } from "redux";
+
+/**
+ * Re-exports
+ */
+export { Action, combineReducers };
+
+//
+// Core functionality
+//
+export type SelectorMap = Record<
+    string,
+    <T = unknown>(...args: readonly any[]) => T
+>;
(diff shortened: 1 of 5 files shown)


types/wordpress__data/index.d.ts
```diff
@@ -0,0 +1,148 @@
...
+//
+// Core functionality
+//
+export type SelectorMap = Record<
+    string,
+    <T = unknown>(...args: readonly any[]) => T
+>;
+export type DispatcherMap = Record<
+    string,
+    <T = unknown>(...args: readonly any[]) => T
```
amcasey: The reason the type parameter is being flagged as unnecessary is that there are no circumstances in which it will be inferred - it will always need to be specified (or allowed to default to `unknown`).  Since the type system has no role in determining the type, it is regarded as more "honest" to put an explicit type assertion at the call site.

types/wordpress__data/wordpress__data-tests.ts
```diff
@@ -0,0 +1,25 @@
+import * as data from "@wordpress/data";
+
+data.select("core/block-editor").isTyping<boolean>();
```
amcasey: The `<boolean>` here amounts to assertion that the result will be a `boolean`.  Since that type isn't either inferred or checked, it might be clearer to explicitly assert that it's `boolean`.

types/wordpress__data/index.d.ts
dsifford: Yep, that's preferred... Because the return value of `dispatch()` will return a map of dispatcher functions that can literally return anything, depending on the namespace provided to the `dispatch` function, and it can be extended indefinitely by users. So there's no way to know ahead of time of the return value of these and I think it would be safer to default it to `unknown` if not explicitly marked.

I'd have it return `unknown` always, but then you can't do this...

```ts
var something: boolean = dispatch('foo').isSomething();
```

because type `boolean` is not compatible with type `boolean`. 

The only way around that would be to do an `as` cast, but that's less ideal and more sloppy (IMHO).

types/wordpress__data/wordpress__data-tests.ts
dsifford: Can't do that because there's no way to know ahead of time of all the possibilities that `select` may return.

See [here](https://github.com/dsifford/academic-bloggers-toolkit/blob/master/types/augments.d.ts) for an example of how a user might augment the definitions to be more type-safe.

types/wordpress__data/index.d.ts
amcasey: You're aware of the reasoning for the rule and you've made an informed decision not to apply it - that's good enough for me.  Get the build passing and I'll merge it.

dsifford: Cool! All set. :+1:

amcasey (approved):

PR #35911, opened 2019-06-03
Add type definitions for frida-gum

oleavr:

### DIFF
--- a/types/frida-gum/frida-gum-tests.ts
+++ b/types/frida-gum/frida-gum-tests.ts
@@ -1,0 +1,54 @@
+Frida.version; // $ExpectType string
+
+const otherPuts = new NativeCallback(() => {
+    return 0;
+}, "int", ["pointer"]);
+
+const puts = new NativeFunction(Module.getExportByName(null, "puts"), "int", ["pointer"]);
+
+// $ExpectType NativeFunction
+puts;
+
+const message = Memory.allocUtf8String("Hello!");
+
+// $ExpectType NativePointer
+message;
+
+// $ExpectType NativeReturnValue
+puts.call(otherPuts, message);
+
+// $ExpectType NativeReturnValue
+puts.apply(otherPuts, [message]);
+
+// $ExpectType NativeReturnValue
+puts(message);
+
+const open = new SystemFunction(Module.getExportByName(null, "open"), "int", ["pointer", "int"]);
+
+// $ExpectType SystemFunction
+open;
+
+const path = Memory.allocUtf8String("/etc/hosts");
+const result = open(path, 0) as UnixSystemFunctionResult;
+
+// $ExpectType NativeReturnValue
+result.value;
+
+// $ExpectType number
+result.errno;
+
+Interceptor.attach(puts, {
+    onEnter(args) {
+        // $ExpectType NativePointer[]
+        args;
+    },
+    onLeave(retval) {
+        // $ExpectType InvocationReturnValue
+        retval;
+    }
+});
+
+const obj = new ObjC.Object(ptr("0x42"));
+
+// $ExpectType Object
+obj;
--- a/types/frida-gum/index.d.ts
+++ b/types/frida-gum/index.d.ts
@@ -12,7 +12,7 @@
  */
 declare function hexdump(target: ArrayBuffer | NativePointerValue, options?: HexdumpOptions): string;
 
-declare interface HexdumpOptions {
+interface HexdumpOptions {
     /**
      * Specifies byte offset of where to start dumping. Defaults to 0.
      */
@@ -69,9 +69,9 @@ declare function recv(callback: MessageCallback): MessageRecvOperation;
  */
 declare function recv(type: string, callback: MessageCallback): MessageRecvOperation;
 
-declare interface MessageCallback { (message: any, data: ArrayBuffer | null): void; }
+interface MessageCallback { (message: any, data: ArrayBuffer | null): void; }
 
-declare interface MessageRecvOperation {
+interface MessageRecvOperation {
     /**
      * Blocks until the message has been received and callback has returned.
      */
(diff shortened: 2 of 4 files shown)


amcasey:
Hey, @oleavr!  It looks like you already have this file in your project: https://github.com/frida/frida-gum/blob/master/bindings/gumjs/types/frida-gum/frida-gum.d.ts.  I'd be interested to know your motivation for publishing it to DefinitelyTyped (as well? instead?).

oleavr:
Hey @amcasey!

Thanks for reviewing; and sorry for the lack of context in my PR.  My intention is the latter – to move them here to aid discovery and make it look more familiar to TypeScript users.

Long story (in case you're curious :-) )
===
The Frida project has an instrumentation core, Gum (written in C), that currently provides two different language bindings: C++ and JavaScript. The latter is what these type definitions are for, and as most of the APIs are implemented in native code there isn't any implementation module published to npm where these type definitions can live (or be generated by).

Users of these type definitions will typically use [frida-compile](https://github.com/frida/frida-compile) to build their TypeScript agent and turn it into a self-contained blob of JS that they can ask GumJS to load to apply the desired instrumentation. Their project might look something like:
https://github.com/oleavr/frida-agent-example

So my thinking is that `@types/frida-gum` will look beautiful next to `@types/foo`, which would be types for some non-TS module that they consume from npm. (frida-compile uses Browserify under the hood and plugs in some shims, so Frida agents are able to consume thousands of existing modules from npm.)

amcasey:
Cool, thanks for the detailed response!  I take it that means these types are already fairly mature, since you've already been consuming them?

oleavr:
Yep, they should be fairly mature at this point. I've used the `frida-gum-types` module to iterate on the types until all APIs were covered and I felt happy about the ergonomics. So the other day I finally felt it was time to open a PR here. :-)

amcasey:
They look pretty solid to me.  My only real question is whether you're actually depending on features of TS 3.4.  We usually specify the lowest possible version so more users can consume it.

oleavr:
Glad to hear!  Ah yes, good catch. The current definitions do not depend on any TS features above 2.4, so I just lowered the requirement. Thanks!

amcasey:
Oops, I missed the disabled lint rules.  I'll admit that some are a bit tedious, but others (e.g. no const enums) are important for consumers.

types/frida-gum/index.d.ts
```diff
@@ -0,0 +1,6668 @@
...
+     */
+    or(v: Int64 | number | string): Int64;
+
+    /**
+     * Makes a new Int64 whose value is `this` ^ `v`.
+     */
+    xor(v: Int64 | number | string): Int64;
+
+    /**
+     * Makes a new Int64 whose value is `this` << `v`.
```
amcasey: I admit I don't use it very often, but isn't `<<` left-shift?

types/frida-gum/frida-gum-tests.ts
```diff
@@ -0,0 +1 @@
+Frida.version; // $ExpectType string
```
amcasey: Consider fleshing this out a bit. :wink:

types/frida-gum/index.d.ts
```diff
@@ -0,0 +1,6668 @@
...
+/**
+ * Intercepts execution through inline hooking.
+ */
+declare namespace Interceptor {
+    /**
+     * Intercepts calls to function/instruction at `target`. It is important
+     * to specify a `InstructionProbeCallback` if `target` is not the first
+     * instruction of a function.
+     */
+    function attach(target
```
amcasey: This could also be written as a union type: `type Arm64ConditionCode = "eq" | "ne" | ...`

types/frida-gum/index.d.ts
oleavr: Oops, indeed, the docstrings somehow ended up reversed. 😅

types/frida-gum/index.d.ts
oleavr: Yes I was tempted to do that, but leaned towards using an enum to make the code clearer.

types/frida-gum/frida-gum-tests.ts
oleavr: Done 😊

oleavr:
Regarding the tslint suppressions:

> interface-name

Triggered by this false positive:

```ts
declare interface IA32CpuContext extends PortableCpuContext {
```

I renamed it to `Ia32CpuContext`, as it's actually what's consistent with the other enums – e.g. `Arm64CpuContext` is not named `ARM64CpuContext`

> no-const-enum

The GumJS runtime does not currently have any runtime-visible metadata for enums, and adding this would be quite a bit of work, so it's something I've got on my backlog for quite far down the road. OTOH I do like the clarity of using enums vs magic strings, so that's why I've opted for depending on const enums, at least for now.

> no-unnecessary-class

I couldn't figure out how to avoid this one for this particular class:

```ts
declare class Backtracer {
    static ACCURATE: Backtracer;
    static FUZZY: Backtracer;
}
```

But I've now reduced the scope of the suppression to only that particular class.

> no-unnecessary-qualifier

This one was needed for `ObjC.Object`, which would otherwise violate the `ban-types` rule. I've now limited the suppression to the `ObjC` namespace.

> strict-export-declare-modifiers

I take it we're stuck with this one as these are all globals and nothing can be exported?

amcasey:
@oleavr thanks for the update and sorry for the delay.

> interface-name

That's a legitimate exception.  Just use a line suppression and it will be clear to readers why it's there.

> no-const-enum

Don't forget that "magic strings" can be strongly typed in TS.  That's why I suggested a union of string literals.  You should get good completions too. 😄 

> no-unnecessary-class 

I have to admit that I get this one mixed up from time to time, but it seems like you might just want a namespace with some constants in it.

> no-unnecessary-qualifier

Best effort. :wink:

> strict-export-declare-modifiers

I don't have your code downloaded at the moment, but I think I skimmed these errors and they were just pointing out that some of your `declare` keywords were redundant.

oleavr:
Thanks a lot @amcasey!

And sorry for the delay here, I got knocked offline by a flu. I have addressed the remaining feedback in #36293, which finally gets rid of the const enums. (Well, all except two, which would require breaking the underlying API.)

PR #35929, opened 2019-06-03
Improve MetadataType definition in @types/chromecast-caf-receiver

Serabe:
According to [documentation,
MetadataType](https://developers.google.com/cast/docs/reference/caf_receiver/cast.framework.messages#.MetadataType)
is a number, not a string.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://developers.google.com/cast/docs/reference/caf_receiver/cast.framework.messages#.MetadataType

### DIFF
--- a/types/chromecast-caf-receiver/cast.framework.messages.d.ts
+++ b/types/chromecast-caf-receiver/cast.framework.messages.d.ts
@@ -196,12 +196,14 @@ export type QueueType =
     | "LIVE_TV"
     | "MOVIE";
 
-export type MetadataType =
-    | "GENERIC"
-    | "MOVIE"
-    | "TV_SHOW"
-    | "MUSIC_TRACK"
-    | "PHOTO";
+export enum MetadataType {
+    GENERIC = 0,
+    MOVIE = 1,
+    TV_SHOW = 2,
+    MUSIC_TRACK = 3,
+    PHOTO = 4,
+    AUDIOBOOK_CHAPTER = 5
+}
 
 /**
  * RefreshCredentials request data.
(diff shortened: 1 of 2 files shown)


Serabe:
@andrewbranch any hint on what I did wrong (if any) to make "getQuickInfoAtPosition" go up that way? Thank you!

andrewbranch:
I need to investigate—it seems very suspicious that the _median_ went up so much without the mean or worst-case going up at all... it's quite possible there's something wonky with the math.

craigrbruce:
@Serabe .. could you please double check this in your code? The documentation has been known to be wrong in several areas.

It has been 6 months since I worked on our Chromecast solution, but it is in production with many thousands of users (I work for Australia's biggest broadcaster) and we use metadata like this with success:

```         const metadata = {
          metadataType: 'GENERIC',  ```

craigrbruce (changes requested):
Please see my comment: Check metadata in your debugger -- documentation may not be correct.

If it is correct then we need to reach out to Google CC developers to make their documentation clear. For example, if `metadataType` is indeed a number then please specify it as an `enum` with keys and values e.g. `GENERIC = 1` etc

Serabe:
Their documentation says that it is a number. Check here: https://developers.google.com/cast/docs/reference/caf_receiver/cast.framework.messages#.MetadataType

For the values, I just did `copy(JSON.stringify(cast.framework.messages.MetadataType))`. Don't know if they accept both `'GENERIC'` and `cast.framework.messages.GENERIC`, but the documentation is explicit about what MetadataType contains (numbers). I can check tomorrow and make adjustments to allow both versions.

craigrbruce (approved):

craigrbruce:
I checked the docs, and in the past, they have been incorrect (number instead of string, lowercase instead of uppercase, object instead of array etc.. etc..) which is why I raised the red flag.

in this case, you appear to be correct, I checked with our production site a short time ago

thanks for raising it - it could be this enum has changed over time, we will need to adjust our code accordingly :(

Cheers.

amcasey:
@andrewbranch any perf concerns before I merge this?

andrewbranch:
@amcasey go ahead 👍

PR #35930, opened 2019-06-03
Add request option 'accessType' to passport-google-oauth20

thebrengun:
There are many other request options. This URL mainly goes over PHP definitions because Google has its own client library for Node. Still, this and other request options would be useful to define.

https://developers.google.com/identity/protocols/OAuth2WebServer


ezintz:
Your changes are looking good and seem to be reasonable. Even if the bot tells me to approve or request a change, I have no user interface for this.

@yasupeke if you are bothered by this I can remove you as an author. I just duplicated one of the strategies...

yasupeke (approved):
Sorry for being late...

PR #35931, opened 2019-06-03
Add webpack-bugsnag-plugins types

skovy:
This was originally proposed here: https://github.com/bugsnag/webpack-bugsnag-plugins/pull/33.

Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/webpack-bugsnag-plugins/index.d.ts
+++ b/types/webpack-bugsnag-plugins/index.d.ts
@@ -2,6 +2,7 @@
 // Project: https://github.com/bugsnag/webpack-bugsnag-plugins#readme
 // Definitions by: Spencer Miskoviak <https://github.com/skovy>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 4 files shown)


types/webpack-bugsnag-plugins/index.d.ts
```diff
@@ -0,0 +1,151 @@
+// Type definitions for webpack-bugsnag-plugins 1.4
+// Project: https://github.com/bugsnag/webpack-bugsnag-plugins#readme
+// Definitions by: Spencer Miskoviak <https://github.com/skovy>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.4
```
amcasey: It doesn't look like you're using any TS features that require version 3.4.  You might want to specify a lower version to make the types more broadly consumable.

skovy: Ah, good point 👍 lowered to match webpack's TS version

amcasey (approved):

PR #35940, opened 2019-06-04
update react-instantsearch-dom & core

foray1010:
Please fill in this template.

1. `appId` and `appKey` no longer recommended way to init the `InstantSearch` (see https://www.algolia.com/doc/api-reference/widgets/instantsearch/react/), it suggests we use `searchClient` now

2. `InfiniteHits` does not have `refine` but `refineNext` and `refinePrevious` (see https://www.algolia.com/doc/api-reference/widgets/infinite-hits/react/?language=react#full-example)

### DIFF
--- a/types/react-instantsearch-dom/index.d.ts
+++ b/types/react-instantsearch-dom/index.d.ts
@@ -55,21 +55,19 @@ interface CommonWidgetProps {
 }
 
 export interface InstantSearchProps {
-  apiKey: string;
-  appId: string;
   indexName: string;
+  searchClient: any;
 
-  algoliaClient?: any;
-  searchClient?: any;
   createURL?: (...args: any[]) => any;
   searchState?: any;
   refresh?: boolean;
   onSearchStateChange?: (...args: any[]) => any;
   onSearchParameters?: (...args: any[]) => any;
   resultsState?: any;
+  stalledSearchDelay?: number;
   root?: {
     Root: string | ((...args: any[]) => any);
-    props: any;
+    props?: object;
   };
 }
 /**
--- a/types/react-instantsearch-dom/index.d.ts
+++ b/types/react-instantsearch-dom/index.d.ts
@@ -58,7 +58,7 @@ export interface InstantSearchProps {
   apiKey?: string;
   appId?: string;
   indexName: string;
-  searchClient: any;
+  searchClient?: any;
 
   createURL?: (...args: any[]) => any;
   searchState?: any;
--- a/types/react-instantsearch-dom/index.d.ts
+++ b/types/react-instantsearch-dom/index.d.ts
@@ -55,6 +55,8 @@ interface CommonWidgetProps {
 }
 
 export interface InstantSearchProps {
+  apiKey?: string;
+  appId?: string;
   indexName: string;
   searchClient: any;
 
--- a/types/react-instantsearch-dom/react-instantsearch-dom-tests.tsx
+++ b/types/react-instantsearch-dom/react-instantsearch-dom-tests.tsx
@@ -7,8 +7,9 @@ import { Hit, connectRefinementList, connectMenu } from 'react-instantsearch-cor
 () => {
   const App = () => (
     <InstantSearch
+      appId="latency"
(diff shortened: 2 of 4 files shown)


gburgett:
Thanks @foray1010 !

Regarding `<InstantSearch />`, is it still possible to use `apiKey` and `appId`?  If so I'd prefer we not remove them from the type definitions.  Better just to add the new parameter.

foray1010:
hi @gburgett ! I have read the source code and seems it is no longer possible to use `apiKey` and `appId`
https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-core/src/components/InstantSearch.js#L176-L204

gburgett:
Aha, I figured out the issue!  That component which you've linked is actually not the one exported by `react-instantsearch-dom`.  If you do the following:

```ts
import { InstantSearch } from 'react-instantsearch-dom'
```
then you actually get this object: https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-dom/src/widgets/InstantSearch.js  
which is created by this connector: https://github.com/algolia/react-instantsearch/blob/master/packages/react-instantsearch-core/src/core/createInstantSearch.js  
which wraps that InstantSearch component you have linked above.  The actual prop types therefore ought to be:

```js
static propTypes = {
      algoliaClient: PropTypes.object,
      searchClient: PropTypes.object,
      appId: PropTypes.string,
      apiKey: PropTypes.string,
      children: PropTypes.oneOfType([
        PropTypes.arrayOf(PropTypes.node),
        PropTypes.node,
      ]),
      indexName: PropTypes.string.isRequired,
      createURL: PropTypes.func,
      searchState: PropTypes.object,
      refresh: PropTypes.bool.isRequired,
      onSearchStateChange: PropTypes.func,
      onSearchParameters: PropTypes.func,
      resultsState: PropTypes.oneOfType([PropTypes.object, PropTypes.array]),
      root: PropTypes.shape({
        Root: PropTypes.oneOfType([
          PropTypes.string,
          PropTypes.func,
          PropTypes.object,
        ]).isRequired,
        props: PropTypes.object,
      }),
};
```

gburgett:
@Haroenv can you confirm?

gburgett (changes requested):
I have to explicitly leave a comment with "request changes" - please revert your changes to the InstantSearchProps, or improve them according to the propTypes linked above.

types/react-instantsearch-dom/index.d.ts
```diff
@@ -55,21 +55,19 @@ interface CommonWidgetProps {
 }
 
 export interface InstantSearchProps {
-  apiKey: string;
-  appId: string;
```
samouss: Those values aren't documented anymore but it's still possible to provide them. Note that they will be removed with the next major version of React InstantSearch. Until the release it's probably better to keep them.

types/react-instantsearch-dom/index.d.ts
```diff
@@ -55,21 +55,19 @@ interface CommonWidgetProps {
...
   indexName: string;
+  searchClient: any;
 
-  algoliaClient?: any;
```
samouss: This one can be removed though. The `searchClient` is the prefer alternative.

samouss (changes requested):

types/react-instantsearch-dom/react-instantsearch-dom-tests.tsx
```diff
@@ -7,9 +7,8 @@ import { Hit, connectRefinementList, connectMenu } from 'react-instantsearch-cor
 () => {
   const App = () => (
     <InstantSearch
-      appId="latency"
-      apiKey="3d9875e51fbd20c7754e65422f7ce5e1"
```
samouss: We should keep at least one test that use this notation to check that it's actually working.

samouss (approved):

foray1010:
thank you very much @gburgett and @samouss ! I have marked `searchClient`, `apiKey` and `appId` as optional field

gburgett (approved):

PR #35945, opened 2019-06-04
aws-lambda: Add CloudFrontRequestBody

Marwes:
CloudFront events can contain the body of the request by enabling a setting. This adds `CloudFrontRequestBody` which exposes that extra (optional field).

https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-event-structure.html

### DIFF
--- a/types/aws-lambda/index.d.ts
+++ b/types/aws-lambda/index.d.ts
@@ -768,8 +768,16 @@ export interface CloudFrontResponse {
     headers: CloudFrontHeaders;
 }
 
+export interface CloudFrontRequestBody {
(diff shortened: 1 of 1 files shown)


simonbuchan (changes requested):
Sorry for the late response, this looks pretty good.
Could you add something to the tests to exercise this? They're a bit of a mess, sorry, but it doesn't need to be exhaustive, just enough to show that it's usable.
Ideally it would show usage for each event or result it shows up in, but I'm not sure how CloudFront handlers should work here, given they seem to expect you to update the event then return bits of it?

Issue #31628, opened 2018-12-24
 Variable 'require' must be of type 'NodeRequire', but here has type 'Require'


kumaresan-subramani:
- [x] I tried using the `@types/requirejs` package and had problems.

**package.json**
`
"devDependencies": {
    "@syncfusion/ej2-build": "*",
    "@types/chai": "^3.4.28",
   "@types/jasmine": "2.8.9",
    "@types/jasmine-ajax": "^3.1.27",
    "@types/requirejs": "^2.1.31"
  }

`
**When i try  compile my application it throws below error:**

> [11:47:10] Starting 'scripts-gen'...
> D:/workspace/components/ej2-button-components/node_modules/@types/requirejs/index.d.ts(396,13): error TS2403: **Subsequent variable declarations must have the same type.  Variable 'require' must be of type 'NodeRequire', but here has type 'Require'.**
> [11:47:23] 'scripts-gen' errored after 12 s
> [11:47:23] TypeScript error: D:/workspace/components/ej2-button-components/node_modules/@types/requirejs/index.d.ts(396,13): error TS2403: Subsequent variable declarations must have the same type.  Variable 'require' must be of type 'NodeRequire', but here has type 'Require'.
>     at Object.getError (D:\workspace\components\ej2-button-components\node_modules\gulp-typescript\release\utils.js:53:15)

**versions:**
node - v8.11.3
npm  - 5.6.0
typescript -3.0.3

Please check this and provide any solution for this ASAP

kumaresan-subramani:
In here, i need updation https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/requirejs/index.d.ts#L421

kumaresan-subramani:
Any update on this?

jonesmac:
I think I am facing this issue as well.  Any update?

thiagomajesk:
Same here, I get the error:

```
C:/Users/{userName}/Source/Repos/project/Project.Web/node_modules/@types/node/module.d.ts(2,5): error TS2300: Duplicate identifier 'mod'.
C:/Users/{userName}/Source/Repos/project/Project.Web/node_modules/@types/requirejs/index.d.ts(38,11): error TS2300: Duplicate identifier 'mod'.
[...]

Issue #15851, opened 2017-04-13
@types/requirejs  Variable 'require' must be of type 'NodeRequire'


cyandream:
- [X] I tried using the `@types/requirejs` package and had problems.
    Error from console:
```
node_modules/@types/requirejs/index.d.ts
(416,13): error TS2403: Subsequent variable declarations must have the same type.  Variable 'require' must be of type 'NodeRequire', but here has type 'Require'.

```
typescript error - 

in @types/requirejs/index.d.ts 
ln416    **declare var require: Require;**

And in @types/node/index.d.ts
ln73    **declare var require: NodeRequire;**

Thanks

Michelle

- [ ] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [ ] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [ ] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @....

ETE-Jens-Ihrig:
Same Problem here with error message with different order:
node_modules/@types/webpack-env/index.d.ts
`Subsequent variable declarations must have the same type.  Variable 'require' must be of type 'Require', but here has type 'NodeRequire'.`

And node_modules/@types/node/index.d.ts
` error TS2403: Subsequent variable declarations must have the same type.  Variable 'require' must be of type 'Require', but here has type 'NodeRequire'.
`

e-cloud:
see a may-be solution here https://medium.com/@mindginative/typescript-its-magic-number-ts2403-ef76202185c9

kumaresan-subramani:
i am also facing same problem. this breaks my entire application

> [11:20:44] D:/workspace/components/ej2-button-components/node_modules/@types/requirejs/index.d.ts(421)
>  Subsequent variable declarations must have the same type.  Variable 'require' must be of type 'NodeRequire', but here has type 'Require'.
> Rendering [========================================] 100%
> [11:20:46] Finished 'typedoc' after 12 s
> [11:20:46] Finished 'execute-generator' after 12 s
> [11:20:46] Starting 'scripts'...
> [11:20:46] Starting 'scripts-gen'...
[...]

PR #35946, opened 2019-06-04
Prevent `@types/node` and `@types/requirejs`-confusion

neurolag:
Please fill in this template.

  I created this PR as many people are having trouble with `requirejs` overriding the variable `require`:
    - #31628
    - #15851
    - <https://github.com/rollup/rollup/issues/2142>

## About this PR
According to the documentation, `@types/requirejs` can be invoked through `require()` and `requirejs()`.  
As `require()` is declared by `@types/node`, too, which is part of most projects.

For that reason I decided to make `@types/requirejs` exporting `requirejs()` rather than `require()`.

### DIFF
--- a/types/requirejs/index.d.ts
+++ b/types/requirejs/index.d.ts
@@ -29,15 +29,6 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 OTHER DEALINGS IN THE SOFTWARE.
 */
 
-declare module 'requirejs/module' {
-	var mod: {
-		config: () => any;
-		id: string;
-		uri: string;
-	}
-	export = mod;
-}
-
 interface RequireError extends Error {
 
 	/**
--- a/types/requirejs/index.d.ts
+++ b/types/requirejs/index.d.ts
@@ -29,7 +29,7 @@ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
 OTHER DEALINGS IN THE SOFTWARE.
 */
 
-declare module 'module' {
+declare module 'requirejs/module' {
 	var mod: {
 		config: () => any;
 		id: string;
--- a/types/requirejs/index.d.ts
+++ b/types/requirejs/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for RequireJS 2.1.20
+// Type definitions for RequireJS 3.0.0
 // Project: http://requirejs.org/
 // Definitions by: Josh Baldwin <https://github.com/jbaldwin>
 //                 Manuel Thalmann <https://github.com/manuth>
--- a/types/requirejs/index.d.ts
+++ b/types/requirejs/index.d.ts
@@ -1,6 +1,7 @@
 // Type definitions for RequireJS 2.1.20
 // Project: http://requirejs.org/
 // Definitions by: Josh Baldwin <https://github.com/jbaldwin>
+//                 Manuel Thalmann <https://github.com/manuth>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 /*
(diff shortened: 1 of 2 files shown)


RyanCavanaugh:
This seems extremely dangerous in terms of breaking existing code and JS users. I would need to see data in terms of what % of extant code uses `require` vs `requirejs` to be OK publishing this.

manuth:
Oh - I thought this should not damage any code as I increased the major version-number.

manuth:
@RyanCavanaugh is this a wrong assumption?
Is there anything else I need to do in order to increase the major version-number but changing the file-header of `index.d.ts`?

manuth:
So I just checked the guide of `DefinitelyTyped` and it says that there is nothing to do but increasing the version-number in the first line of the `index.d.ts`-file:

I updated the major version which means that no package will be affected.
Is there anything more I have to do?

Eagle94T:
@RyanCavanaugh what's the status for this PR?
With updating the major version, this shouldn't affect any existing projects.

Eagle94T:
I'm sorry to ask again, but why is this PR still not merged?
I need this change, since I'm using aurelia-bundler (uses @types/node) and @types/select2 (uses @types/requirejs)

andrewbranch:
> Oh - I thought this should not damage any code as I increased the major version-number.
>
> ...is this a wrong assumption?

Yes—the version number is supposed to represent the version of the _JavaScript library on npm_. The latest release of RequireJS is 2.3.6. In fact, @sandersn _just_ added a CI rule to prevent type definitions from referencing non-existent versions of their JS counterparts.

Echoing Ryan, I think we would want to see some data here. I think my _hunch_ is that lots of RequireJS users probably use `require()` (and want to keep it that way), in which case this change is probably a non-starter, although well-intentioned. If you can prove me wrong, then we can proceed. 👍

andrewbranch (changes requested):
It appears that @typescript-bot might have interpreted my trailing “👍” in that last comment as an approval? I’m requesting changes just so it’s clear the status is not mergeable without further research/discussion.

manuth:
Honestly I don't think "people may want to keep using `require()`" is a pretty bad argument.
By now you cannot write typescript-modules which have both `@types/requirejs` and `@types/node` in their dependency-tree which often might be the case as some types in DefinitelyTyped use either parts of `@types/requirejs` or `@types/node`.

To me its a bad idea to prevent people from writing typescript-code just because  "people want to use `require()`.

sandersn:
Here's a workaround: provide a node-specific version of the package which basically implements this change, but copy it into a subdirectory named `node`. Then people who have the conflict can add the following path mapping to their tsconfig:

```
        "baseUrl": ".",
        "paths": {
            "requirejs": ["node_modules/@types/requirejs/node/index.d.ts"]
        }
```

It should be possible to share the types between the main file and the node file, so that they only differ in the last few lines -- the main file would have 3 globals and the node file would have a single `export =`.

andrewbranch:
@sandersn I tried this locally, but `node_modules/@types/requirejs/index.d.ts` (with its ambient declarations) still get pulled into the program. I think you’d have to combine this approach with specifying `typeRoots` or `types` in your tsconfig, which could be pretty frustrating if you have tons of @types packages and you just need to _prevent_ one file in them from being included. (Related: https://github.com/microsoft/TypeScript/issues/18588)

manuth:
@RyanCavanaugh. @andrewbranch I'd be glad if you guys could discuss this further so I can open a new PR for resolving the `node`/`requirejs`-intersection.

sandersn:
@andrewbranch I was wrong about my fix, the errors you saw got lost in a bunch of others. I'm out of ideas on this one.

manuth:
Should I wait for you guys to work out a solution or should I rather open a new PR?

andrewbranch:
My only idea is to publish a separate package, e.g. @types/requirejs-node.

PR #35950, opened 2019-06-04
Add types for didyoumean.js

jamesgeorge007:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/didyoumean/index.d.ts
+++ b/types/didyoumean/index.d.ts
@@ -1,0 +1,25 @@
+// Type definitions for didyoumean 1.2.1
+// Project: https://github.com/dcporter/didyoumean
+// Definitions by: James George <https://github.com/jamesgeorge007>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare function didyoumean(str: any, list: any, key: any): any;
+
+export interface didyoumean {
+    const caseSensitive: boolean;
+
+    const nullResultValue: any;
+
+    const prototype: {
+    };
+
+    const returnFirstMatch: boolean;
+
+    const returnWinningObject: any;
+
+    const threshold: number;
+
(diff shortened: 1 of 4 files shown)


types/didyoumean/index.d.ts
```diff
@@ -0,0 +1,18 @@
+// Type definitions for didyoumean 1.2.1
```
andrewbranch: FYI, this is what’s causing the builds to fail—the version number should only be major/minor, no patch number.

andrewbranch:
/azp run DefinitelyTyped.BenchmarkPR

(I'm testing making the perf results a little quieter on updates, ignore me 😄)

jamesgeorge007:
cc @amcasey

types/didyoumean/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+export = didYouMean;
+
+declare function didYouMean(str: string, list: string[], key?: string): string | string[];
+
+declare namespace didYouMean {
+    let caseSensitive: boolean;
+
+    let nullResultValue: any;
+
+    let prototype: {
```
amcasey: I don't believe you need to explicitly declare `prototype` - you should get it from the merged function declaration.

jamesgeorge007:
@amcasey Hope it's fine now :thinking:

amcasey (approved):
LGTM.  Thanks!

iabdo9:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/fc9a888ab36a8122d73ab42a3b9f4fc473728454/types/didyoumean/index.d.ts#L3

I think we should add `object[]` to list parameter 
```js
declare function didYouMean(str: string, list: string[] | object[], key?: string): string | string[];
```

PR #35958, opened 2019-06-05
[Prettier] Run Prettier over repo

damassi:
Related: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35672

Per [this comment](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35672#pullrequestreview-245718700), performing a trial run that applies Prettier over entire DefinitelyTyped repo. 

cc @sandersn


types/activex-excel/index.d.ts
```diff
@@ -11,38 +11,53 @@
...
-        T extends XlSheetType.xlChart ? Chart :
-        T extends XlSheetType ? Worksheet :
-        T extends undefined ? Sheet :
-        never;
-
-    type InternationalResult<T> =
-        T extends XlApplicationInternationalStrings ? string :
-        T extends XlApplicationInternationalNumbers ? number :
-        T extends XlApplicationInternationalBooleans ? boolean :
-        any;
```
zspitz: The intention of a conditional type might be clearer if formatted as above.

types/activex-excel/index.d.ts
```diff
@@ -11,38 +11,53 @@
...
-        TSourceType extends undefined ? undefined :
-        never;
-
-    type PivotTableWizardAutoPage<TSourceType> =
-        TSourceType extends XlPivotTableSourceType.xlConsolidation ? boolean :
-        undefined;
-
-    type PrintToFileName<T> =
-        T extends true ? string :
-        undefined;
```
zspitz: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35958/files#r290613127

types/activex-excel/index.d.ts
```diff
@@ -3156,10 +3174,84 @@ declare namespace Excel {
...
+            Macro?: any,
+            Description?: any,
+            HasMenu?: any,
+            MenuText?: any,
+            HasShortcutKey?: any,
+            ShortcutKey?: any,
+            Category?: any,
+            StatusBar?: any,
+            HelpContextID?: any,
+            HelpFile?: any
```
zspitz: Perhaps if all the function arguments are `any` the arguments should be on a single line, or however many lines it takes; instead of one argument per line?

types/activex-excel/index.d.ts
```diff
@@ -3330,18 +3507,70 @@ declare namespace Excel {
...
+            Arg22?: Range,
+            Arg23?: Range,
+            Arg24?: Range,
+            Arg25?: Range,
+            Arg26?: Range,
+            Arg27?: Range,
+            Arg28?: Range,
+            Arg29?: Range,
+            Arg30?: Range
+        ): Range;
```
zspitz: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35958/files#r290613779 Or even if the arguments are all the same type?

types/activex-excel/index.d.ts
```diff
@@ -3849,12 +4196,77 @@ declare namespace Excel {
...
+            | 39
+            | 40
+            | 41
+            | 42
+            | 43
+            | 44
+            | 45
+            | 46
+            | 47
+            | 48;
```
zspitz: Perhaps types defined as a set of numeric constants should also be on a single line?

Jessidhia: Not much that can be done here, unless a `// prettier-ignore` is added to the definition. This is prettier's chosen format because the line got over its max `printWidth`.

types/activex-excel/index.d.ts
Jessidhia: This is also prettier's formatting system; if the function declaration, all the way to the `{`, does not fit within `printWidth`, all the arguments are printed one per line.

types/activex-excel/index.d.ts
Jessidhia: There is some discussion about how to format the conditional operator in the prettier repository itself, but nothing we can change from here as it's not a tunable option.

types/activex-excel/index.d.ts
zspitz: Is that an acceptable strategy -- to add `// prettier-ignore`

Jessidhia: I'm not sure of what the team thinks; but personally, and at work, `// prettier-ignore` is _never_ used unless it's to work around a prettier bug where it would create invalid syntax or unstable formatting (i.e. when the formatting would change every time prettier runs).

So, in this case, I think we'd just have to deal with the newlines.

alloy: Same for us. In the long run, we find that letting go of some of these formatting preferences makes it easier/more efficient for people to contribute as all thinking about formatting is taken out of the equation and leads to a consistent code-base that reduces cognitive load on maintainers.

damassi:
(Many apologies to all of the maintainers automatically pinged by this PR.)

damassi:
(Going to close this so that automated interval pings don't go out.)

PR #35970, opened 2019-06-05
Namespace JSX under React.JSX

rymndhng:
When using both `preact` and `react` in typescript projects, the type checker gets conflicting type definitions because both projects declare global JSX definitions. Typescript since 2.8 has recommended libraries to internally namespace `JSX` to avoid conflicting definitions. See factory functions: https://www.typescriptlang.org/docs/handbook/jsx.html#factory-functions.

Preact 10.x already switched to internally namespaced JSX. See https://github.com/preactjs/preact/pull/1448

---

Please fill in this template.

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.typescriptlang.org/docs/handbook/jsx.html#factory-functions

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -2803,7 +2803,7 @@ type ReactManagedAttributes<C, P> = C extends { propTypes: infer T; defaultProps
             ? Defaultize<P, D>
             : P;
 
-declare global {
+namespace React {
     namespace JSX {
         // tslint:disable-next-line:no-empty-interface
         interface Element extends React.ReactElement<any, any> { }


ferdaber:
This is an enormously large breaking change. I'm all for it obviously since it's a positive change, but do understand that this PR will probably need some better callouts (and maybe even a FAQ section?) for the inevitable horde of folks wondering why their typings suddenly stop working.

ferdaber:
You will also likely need to change the numerous other dependent packages so that they reference the correct type (I think a lot of them rely on global `JSX` right now).

Jessidhia:
I actually didn't know the compiler would check `React.JSX` before checking for `JSX` the global; sounds useful.

The hard part is going to fix the many things that assume `JSX` is global...

ferdaber:
I can probably work on a codemod that will convert previously-global JSX references. That + a FAQ on this PR can help alleviate the breaking change.

eps1lon:
It might be a good idea to start a changelog for this package. Would give us the opportunity to describe releases in SemVer terms even though the version doesn't follow SemVer.

We probably need to coordinate with the types publisher to arrange that they whitelist `CHANGELOG.md` for published files.

ferdaber:
We could release it as non-breaking but I'm not sure if it will have any positive impact anymore. It's not like we're the React team and can go tell people to not use the global JSX namespace anymore. Maybe if we mark it as deprecated? Is it possible to annotate a namespace as deprecated?

PR #35973, opened 2019-06-05
Lodash add partialshallow

sandersn:
Based on @aj-r's idea, switch away from `Partial` to the less restrictive `PartialShallow`, which only requires types to match for primitives.

This didn't require any test changes on my machine, which is odd.


aj-r (approved):
I think not changing any tests is the correct thing here. You've only made an input type more permissive, so I'd be a little surprised if something broke. If anything, it might be worth adding a test with a deep partial object literal (maybe 2 levels deep?), to make sure we don't accidentally break that in the future.

Another option is below. I didn't suggest it before because I assumed it would have the same recursion issues as the old `PartialDeep`. But based on your suggestion #3 in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35848#issuecomment-498716825, maybe it will work? (I'll admit I don't fully understand how #3 fixes the problem though)

```ts
type PartialDeep<T> = {
    [P in keyof T]?: T[P] extends object ? PartialDeep<T[P]> : T[P]
}
```

sandersn:
Option (3) only works because Typescript can't relate conditional types, so it gives up when it encounters them before instantiation happens. That might change someday, so it's not a good idea to rely on it.

I'll add a double-partial test.

sandersn:
Note: I tried switching to PartialShallow in pre-3.1 but it messed up type parameter inference for the 1-parameter version of fp.every. I think it's because it only has one argument to infer from, and we're stricter about inferences to non-recursive mapped types.

It might still be worth fixing, but in a separate PR.

PR #35977, opened 2019-06-05

regevbr:
Please fill in this template.

### DIFF
--- a/types/underscore/index.d.ts
+++ b/types/underscore/index.d.ts
@@ -6,6 +6,7 @@
 //                 Ard Timmerman <https://github.com/confususs>,
 //                 Julian Gonggrijp <https://github.com/jgonggrijp>,
 //                 Florian Keller <https://github.com/ffflorian>
+//                 Regev Brody <https://github.com/regevbr>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
@@ -3713,6 +3714,11 @@ declare module _ {
         pick(
             object: any,
             ...keys: any[]): any;
+        pick(
+            object: any,
+            keys: any[]): any;
+        pick<T, K extends keyof T>(obj: T, ...keys: K[]): Pick<T, K>;
+        pick<T, K extends keyof T>(obj: T, keys: K[]): Pick<T, K>;
 
         /**
         * @see _.pick
--- a/types/underscore/index.d.ts
+++ b/types/underscore/index.d.ts
@@ -3711,14 +3711,13 @@ declare module _ {
         * @keys The key/value pairs to keep on `object`.
         * @return Copy of `object` with only the `keys` properties.
         **/
+        pick<T, K extends keyof T>(obj: T, ...keys: K[]): Pick<T, K>;
         pick(
             object: any,
             ...keys: any[]): any;
         pick(
             object: any,
             keys: any[]): any;
-        pick<T, K extends keyof T>(obj: T, ...keys: K[]): Pick<T, K>;
-        pick<T, K extends keyof T>(obj: T, keys: K[]): Pick<T, K>;
 
         /**
         * @see _.pick
--- a/types/underscore/index.d.ts
+++ b/types/underscore/index.d.ts
@@ -88,6 +88,10 @@ declare module _ {
         (element: T, key: string, list: Dictionary<T>): TResult;
     }
 
+    interface FullObjectIterator<T, TResult, K extends keyof T = keyof T> {
+        (value: T[K], key: K, object: T): TResult;
+    }
+
     type IterateePropertyShorthand = string | number;
 
     type IterateeMatcherShorthand<T> = Dictionary<T>;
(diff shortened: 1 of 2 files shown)


types/underscore/index.d.ts
```diff
@@ -3710,9 +3711,13 @@ declare module _ {
         * @keys The key/value pairs to keep on `object`.
         * @return Copy of `object` with only the `keys` properties.
         **/
+        pick<T, K extends keyof T>(obj: T, ...keys: K[]): Pick<T, K>;
```
jgonggrijp: I'm missing two signatures from [the documentation](https://underscorejs.org/#pick):

```ts
pick<T, K extends keyof T>(obj: T, keys: K[]): Pick<T, K>;
pick<T, K extends keyof T, V extends typeof T[K]>(obj: T, predicate: ObjectIterator<V, boolean>): Pick<T, K>;
```

In the second line, I'm not 100% sure about the notation of the third type parameter (`V extends typeof T[K]`), but I'm sure such a signature needs to be there because `pick` accepts a predicate.

types/underscore/index.d.ts
```diff
@@ -3710,9 +3711,13 @@ declare module _ {
...
         pick(
             object: any,
             ...keys: any[]): any;
```
jgonggrijp: In my opinion, if you make the strongly typed signatures completely cover all valid use cases of `pick`, which I think will be the case if you add both signatures from my previous comment, then this old, overly permissive signature should be removed.

types/underscore/index.d.ts
```diff
@@ -3710,9 +3711,13 @@ declare module _ {
...
+        pick(
+            object: any,
+            keys: any[]): any;
```
jgonggrijp: This new overload certainly shouldn't be added; apart from the fact that it is overly permissive, it doesn't add any information because all call signatures that match this will also match the previous one.

jgonggrijp (changes requested):
These changes don't actually seem to fix a problem. One strongly typed signature is added, but it doesn't cover all valid calls of the `pick` function. Moreover, the existing, overly permissive signature is preserved, so the set of permitted call signatures doesn't change at all. Please see my comments below for details and suggestions for improvement.

Apart from the comments below, please add tests to confirm that all three valid call signatures pass the type checker.

types/underscore/index.d.ts
regevbr: Hi,

The reason I didn't add the first one is that when I tested it, TypeScript failed to infer K automatically so I decided to drop it - but now I realize I was wrong... 
Same goes for the 2nd signature - TypeScript can't infer K by itslef but you are right we should at least enforce the predicate and return types.

So, I added the 2 missing signatures. but regarding the predicate signature - I changed it a bit and created a new `ObjectIterator` called `FullObjectIterator` as the object values doesn't have to be the same for all keys (like in the tests). I'm not sure about the name I gave it so please let me know your feedback on the matter.

types/underscore/index.d.ts
regevbr: done

types/underscore/index.d.ts
regevbr: Yes it was a typo. Removed it

types/underscore/index.d.ts
jgonggrijp: In my opinion, this is an improvement, but I think `T[K]` and the `ObjectIterator` interface work a bit differently than you seem to assume. If I'm right, I want to request a few more changes.

Let's say we have the following `T`.

```ts
interface T {
    apple: string;
    banana: number;
    cherry: boolean;
}
```

Then I think we agree that `keyof T` is `'apple' | 'banana' | 'cherry'`. I think we also agree that `T['apple']` is `string`. However, you seem to have different expectations about `T[keyof T]`. It expands to `T['apple' | 'banana' | 'cherry']`, so I think it will be `string | number | boolean`. The typescript handbook is not 100% clear on this situation, however.

Furthermore, I think `T` is a subset of `Dictionary<string | number | boolean>`. Note that the element values don't need to be the same for all keys in such a `Dictionary`, either. This would mean that any `ObjectIterator` that accepts `string | number | boolean` as its first argument will be able to iterate over `T`. In this case, you don't need `FullObjectIterator`, although `FullObjectIterator` is arguably slightly more correct than `ObjectIterator`.

Replacing `ObjectIterator`, or giving it a slightly different sibling, doesn't seem like a good idea to me. So I'd prefer that you stick closer to the predicate signature I suggested and just use `ObjectIterator`.

In `FullObjectIterator`, you wrote `K extends keyof T = keyof T` to provide a default when TS can't infer `K`. You can apply this trick to the `pick` function signatures too, if necessary.

jgonggrijp (changes requested):
I think you're on the right track, but please see my comment below.

types/underscore/index.d.ts
regevbr: Thanks for the detailed response!
I agree with you completely! I removed my new predicate interface and reverted to use the already existing one.
As to your last suggestion - I don't it applies to the pick method as it doesn't make sense to perform a pick on all the keys of an object as it will act like a `copy` method, so I think it is better to let the user choose the subset of keys to pick.

jgonggrijp (approved):
I think it's ready to merge! 👍

PR #35992, opened 2019-06-06
selenium-webdriver: add "defineCommand" method to executor class

beta-vulgaris:
method "defineCommand" was missing that would block executing custom commands, that are needed e.g. for enabling file download in headless testing.

### DIFF
--- a/types/selenium-webdriver/index.d.ts
+++ b/types/selenium-webdriver/index.d.ts
@@ -1461,6 +1461,21 @@ export class Capabilities {
  */
 export class Executor {
   /**
+   * Defines a new command for use with this executor. When a command is sent,
+   * the {@code path} will be preprocessed using the command's parameters; any
+   * path segments prefixed with ':' will be replaced by the parameter of the
+   * same name. For example, given '/person/:name' and the parameters
+   * '{name: 'Bob'}', the final command path will be '/person/Bob'.
+   *
+   * @param {string} name The command name.
+   * @param {string} method The HTTP method to use when sending this command.
+   * @param {string} path The path to send the command to, relative to
+   *     the WebDriver server's command root and of the form
+   *     '/path/:variable/segment'.
+   */
+  defineCommand(name: string, method: string, path: string): void;
+  
+  /**
    * Executes the given {@code command}. If there is an error executing the
    * command, the provided callback will be invoked with the offending error.
    * Otherwise, the callback will be invoked with a null Error and non-null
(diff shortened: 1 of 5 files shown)


types/selenium-webdriver/index.d.ts
```diff
@@ -1460,6 +1461,21 @@ export class Capabilities {
  * @interface
  */
 export class Executor {
```
cnishina: Executor is not exported from the index.js file and does not belong here. https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/index.js

It is exported in http.d.ts. It could also be added to a file command.d.ts but it still needs to be exported from http.d.ts. See https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/http.js#L619 and https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/command.js#L198

I think this could be moved to a command.d.ts file then http.d.ts would also have to export the Executor class.

cnishina (changes requested):

types/selenium-webdriver/index.d.ts
beta-vulgaris: Sounds reasonable!
I also added a test for the command executor.

types/selenium-webdriver/index.d.ts
```diff
@@ -14,7 +15,8 @@ import * as edge from './edge';
 import * as firefox from './firefox';
 import * as ie from './ie';
 import { By, ByHash } from './lib/by';
-import { Command, ICommandName, Name } from './lib/command';
+import * as command from './lib/command';
+import * as http from './http';
```
cnishina: I don't think you will need this import. See comment below about `command.Executor`.

cnishina (changes requested):
You still need to export the types for Executor in http.d.ts

See https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/http.js#L619

types/selenium-webdriver/index.d.ts
```diff
@@ -2072,7 +2056,7 @@ export class WebDriver {
    * @param {!command.Executor} executor The executor to use when sending
    *     commands to the browser.
    */
-  constructor(session: Session|Promise<Session>, executor: Executor);
+  constructor(session: Session|Promise<Session>, executor: http.Executor);
```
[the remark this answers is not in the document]
beta-vulgaris: @cnishina http executor should be right, see:

https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/index.js#L641

PR #35994, opened 2019-06-06
D3 V3 (Added Index as optional parameter in interface Axis tickformat)

vbinithyanandamv:
Index can be use in tickFormat function. Hence added it as optional Parameter.

Please fill in this template.

If adding a new definition:

- [x ] Increase the version number in the header if appropriate.

If removing a declaration:


types/d3/v3/index.d.ts
```diff
@@ -2576,7 +2577,7 @@ declare namespace d3 {
             tickPadding(): number;
             tickPadding(padding: number): Axis;
 
-            tickFormat(): (t: any) => string;
+            tickFormat(): (t: any , index ?: number) => string;
```
gustavderdrache: 1. The signature for this function should be `(t: any, index: number) => string` - we can't guarantee that the function doesn't use the second parameter. (If you know it doesn't, you can use a type assertion to ignore it.)
2. Make sure you're reflecting this extra parameter in the setter below this one. The tests are failing because the two-argument function you passed in don't match the declaration.

gustavderdrache (changes requested):
Close! I think I found why the test are failing, so I made a note. I also left some feedback on the optional second parameter - I'm not sure it should always be optional (see [Optional Parameters in Callbacks](http://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#optional-parameters-in-callbacks) in the handbook for why).

types/d3/v3/index.d.ts
vbinithyanandamv: Thanks for the feedback. Changes made.

gustavderdrache (approved):
:+1:

types/d3/v3/index.d.ts
```diff
@@ -2576,8 +2577,8 @@ declare namespace d3 {
             tickPadding(): number;
             tickPadding(padding: number): Axis;
 
-            tickFormat(): (t: any) => string;
-            tickFormat(format: (t: any) => string): Axis;
+            tickFormat(): (t: any , index : number) => string;
```
amcasey: I feel like I'm missing something obvious, but is `index` optional or not?

vbinithyanandamv: Hi casey, can you please refer previous comments. You may get the idea of why i did it.

denisname:
Where is this documented ?

types/d3/v3/d3-tests.ts
```diff
@@ -786,7 +786,7 @@ function populationPyramid() {
         .scale(y)
         .orient("right")
         .tickSize(-width)
-        .tickFormat(function (d) { return Math.round(d / 1e6) + "M"; } );
+        .tickFormat(function (d,i) { console.log(i); return Math.round(d / 1e6) + "M"; } );
```
denisname: Try to make the test more "precise": `console.log` works whatever the type of `i` is.
Also a tests for the getter and with one parameter, to be sure it still works, would be nice.

Please respect the file style indentation:
```suggestion
        .tickFormat(function (d, i) { console.log(i); return Math.round(d / 1e6) + "M"; } );
```
Same below: `(t: any, index: number)`

vbinithyanandamv:
> Where is this documented ?

Hi, Not able to find the exact document link. However in d3 V4, type is already been updated with index  which you can reference here. 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/8d217f368733b640b90ffc0c8d8224746733f6cf/types/d3-axis/index.d.ts#L219

types/d3/v3/index.d.ts
amcasey: I'm assuming you're referring to this comment: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35994#discussion_r291305510.  I saw it, but didn't follow - could you possibly paraphrase?

vbinithyanandamv: Don’t use optional parameters in callbacks unless you really mean it:

/* WRONG */
interface Fetcher {
    getObject(done: (data: any, elapsedTime?: number) => void): void;
}

This has a very specific meaning: the done callback might be invoked with 1 argument or might be invoked with 2 arguments. The author probably intended to say that the callback might not care about the elapsedTime parameter, but there’s no need to make the parameter optional to accomplish this – it’s always legal to provide a callback that accepts fewer argument.

What i understood from here is. It's good to keep it as not optional since it provide reasonable value to the function. It's totally up to us to decide whether it should be optional or not. After considering the comment and mapping it with newer type. I decided to change it to not optional. Hope this make sense.

types/d3/v3/d3-tests.ts
```diff
@@ -786,7 +786,7 @@ function populationPyramid() {
         .scale(y)
         .orient("right")
         .tickSize(-width)
-        .tickFormat(function (d) { return Math.round(d / 1e6) + "M"; } );
+        .tickFormat(function (d, i) { console.log(i); return Math.round(d / 1e6) + "M"; } );
```
rbuckton: Since the tests aren't actually run as code, you just need to verify the correct type is received. You can do that with an `// $ExpectType` comment:

```suggestion
        .tickFormat(function (d, i) {
            // $ExpectType number
            i;
            return Math.round(d / 1e6) + "M";
        });
```

rbuckton (changes requested):

types/d3/v3/d3-tests.ts
vbinithyanandamv: Thanks for the suggestion. I have updated it.

vbinithyanandamv (commented):
Done.

PR #36003, opened 2019-06-06
validator: split out all library packages into their own .d.ts files

osdiab:
Because all the declarations are made in the root `validator/index.d.ts` file, if you don't automatically include the types from this library (i.e. don't define `types` in `tsconfig.json` `compilerOptions` field, or explicitly set `"types": ["validator"]`), then all the submodule types aren't accessible.

this puts all of them in their own files.

- [ ] **N/A** Add or edit tests to reflect the change. (Run with `npm test`.)

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [Similar issue from Material UI](https://github.com/mui-org/material-ui/issues/7853#issuecomment-324155653)
- [ ] **N/A** Increase the version number in the header if appropriate.
- [ ] **N/A** If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

Lib files generated with this script:

```javascript
const path = require("path");
const fs = require("fs");

function findAll(regexPattern, sourceString) {
    let output = [];
    let match;
    // make sure the pattern has the global flag
    let regexPatternWithGlobal = RegExp(regexPattern, "g");
    while ((match = regexPatternWithGlobal.exec(sourceString))) {
        // get rid of the string copy
        delete match.input;
        // store the match data
        output.push(match);
    }
    return output;
}

const existing = fs.readFileSync("./index.d.ts").toString();
const regex = /declare module "validator\/(.+)" {[^}]+}/s;
const result = findAll(regex, existing);
result.forEach(r => {
    const [code, filename] = r;
    const parts = filename.split("/");
    const target = `${path.resolve(".", ...parts)}.d.ts`;
    const [declarationLine, constDeclaration, ...restOfCode] = code.split("\n");
    const importValidator = 'import * as validator from "validator";';
    const declareConstProperly = `declare ${constDeclaration.trim()}`;
    const toWrite = [
        importValidator,
        declareConstProperly,
        ...restOfCode.slice(0, -1).map(s => s.trim())
    ].join("\n");
    fs.writeFileSync(target, toWrite);
    console.log(`"${filename}.d.ts",`);
});
```


osdiab:
BTW one thing I noticed is that not all the functions have definitions defined

kvcpr (approved):
LGTM

types/validator/lib/blacklist.d.ts
```diff
@@ -0,0 +1,5 @@
+declare module "validator/lib/blacklist" {
```
amcasey: Is the module declaration redundant now that the export is in a file with that path?

osdiab: Let me give it a shot!

osdiab: nope that wasn't happy lol

osdiab: says you can’t declare variables and exports outside of a declare block

osdiab:
Updated so that each lib file doesn't declare its filename explicitly.

amcasey:
Thanks!

PR #36010, opened 2019-06-06
[coinbase] Update Account.getTransactions

jrnt30:
This is to update the signature of `Account.getTransactions` of the `coinbase` module to match that which is present in the underlying NPM package.

https://github.com/coinbase/coinbase-node/blob/master/lib/model/Account.js#L98 for relevant JS code.

### DIFF
--- a/types/coinbase/index.d.ts
+++ b/types/coinbase/index.d.ts
@@ -966,29 +966,29 @@ export class Buy implements Resource {
 }
 
 export interface Fee {
-  /**
-   * Amount associated to this fee
-   */
-  amount: MoneyHash;
-  /**
-   * Fee beneficiary ("bank", "coinbase", ...)
-   */
-  type: string;
+    /**
+     * Amount associated to this fee
+     */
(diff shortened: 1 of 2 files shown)


rogierschouten (approved):

types/coinbase/index.d.ts
```diff
@@ -600,7 +600,7 @@ export class Account implements Resource {
      * Lists account’s transactions.
      * Scope: wallet:transactions:read
      */
-    getTransactions(cb: (error: Error | null, result: Transaction[]) => void): void;
+    getTransactions(opts: {}, cb: (error: Error | null, result: Transaction[]) => void): void;
```
amcasey: Was this just wrong or has the underlying library changed?  If the latter, should the version number in the header be updated?

jrnt30:
It was just wrong. The library extracts things from the data block only in
select cases. These more primitive types came back as is and this just
wouldn’t have worked.

PR #36017, opened 2019-06-07
update(miniprogram-wxs): modified es5

NewFuture:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://developers.weixin.qq.com/miniprogram/en/dev/framework/view/wxs/07basiclibrary.html>

### DIFF
--- a/types/miniprogram-wxs/es5.d.ts
+++ b/types/miniprogram-wxs/es5.d.ts
@@ -1,0 +1,1634 @@
+/*! *****************************************************************************
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the Apache License, Version 2.0 (the "License"); you may not use
+this file except in compliance with the License. You may obtain a copy of the
+License at http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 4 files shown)


rbuckton:
Why is this including an entire copy of the lib.es5.d.ts file?

NewFuture:
> Why is this including an entire copy of the lib.es5.d.ts file?

It is almost the same with lib.est.d.ts, but some interfaces are modified.
* All Constructors are deleted
* All buffer related interfaces are deleted
* some vars are deleted

All of those are cannot use in `wxs`.
And I could not find a way to disable them in the default es5 lib.

Is there a better way to to do that?

rbuckton:
There is no other way to do it, but you will constantly need to keep this up to date with changes in lib.es5.d.ts since we also put common TypeScript interfaces and types in that declaration file as well.

NewFuture:
Yes. And in fact, the interface of  `wxs`  is not keeping up to date immediately with `lib.es5`.
Currently, it makes sense to maintain a copy of this lib.

Thanks, @rbuckton.

PR #32210, opened 2019-01-16
Define KnockoutObservableType to match returned type of fromJS

turdwaster:
Please fill in this template.

- [X] Use a meaningful title for the pull request. Include the name of the package modified.
- [X] Test the change in your own code. (Compile and run.)
- [X] Add or edit tests to reflect the change. (Run with `npm test`.)
- [X] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [X] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [X] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [X] Provide a URL to documentation or source code which provides context for the suggested changes: Uhm, https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html ?
- [X] Increase the version number in the header if appropriate. - there was none - changed TS version
- [X] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #36025, opened 2019-06-07
Knockout mapping return properly typed with TS 2.8

ltlombardi:
- Reduction of duplication due to Conditional Typing present in TS 2.8
- To allow the KnockoutObservableType<T> to work properly, it was necessary to update to TS 2.8.
Now, the methods that return a mapped object, if the object source is typed, the return type will also be typed properly.
- Changed some tslint rules to improve code.
- Small fixes

- Removed knockout.mapping plugin tests present in father project (knockout). It should not have dependency in it's children. It's the other way around.

- [x] Add or edit tests to reflect the change. (Run with `npm test`.) // failing due to other projects


ltlombardi:
So, I posted this on this project in Gitter.

> Hi,
> I've made improvements in the typing knockout.mapping, a plugin to knockout framework, and it required upping the TS Version. I'm getting this error in Travis testing:
> Error: knockout depends on knockout.mapping but has a lower required TypeScript version. at checkTypeScriptVersions (/home/travis/build/DefinitelyTyped/DefinitelyTyped/node_modules/types-publisher/src/check-parse-results.ts:51:23) at Object.checkParseResults [as default] (/home/travis/build/DefinitelyTyped/DefinitelyTyped/node_modules/types-publisher/src/check-parse-results.ts:19:5)
> 
> Knockout doesn't really depend on knockout.mapping, but someone added tests for knockout.mapping in the knockout tests.
> So:
> 1) Is this normal / default behavior to add tests of the plugins in the base library?? Should I keep them there?
> 2) Should I just up the TS version of the base library (knockout) ?


After 4 days I got no response. So I've decided to remove the dependency. There shouldn't be references to child projects in the main project. It's the other way around. 

This way, the child projects can have different typescript version requirements.

types/knockout.mapping/index.d.ts
```diff
@@ -15,13 +15,23 @@ export = self;
...
+        any;
+
     type KnockoutObservableType<T> = {
-        [P in keyof T]: KnockoutObservable<KnockoutObservableType<T[P]>> | T[P];
+        [P in keyof T]: T[P] extends Primitives ? KnockoutObservable<T[P]> :
+                        T[P] extends any[] ? KnockoutObservableArrayType<T[P][number]> :
+                        MappedType<T[P]>;
     };
 
+    // Could not get his to return any when T is any. It returns a Union type of the possible values.
```
rbuckton:
```suggestion
    // Could not get this to return any when T is any. It returns a Union type of the possible values.
```

rbuckton (changes requested):

ltlombardi:
@rbuckton The fix has been made. Can you approve it?

types/knockout.mapping/index.d.ts
```diff
@@ -4,7 +4,7 @@
 //                 Mathias Lorenzen <https://github.com/ffMathy>
 //                 Leonardo Lombardi <https://github.com/ltlombardi>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+// TypeScript Version: 2.8
```
Igorbek: I would suggest to have redirection of types and leave old version for compatibility.
As I know most users of ko is considering it legacy code with older ts.
So we should be careful with uplifting ts requirements.

Igorbek (changes requested):

types/knockout.mapping/index.d.ts
ltlombardi: I don't know what you mean by: "redirection of types".

I thought long and hard about upping the TS version, but the only way for this typing to work is with version 2.8. 

It's my understanding that when someone needs and older version of a typing, he needs to check the git history and get the version that satisfy he's needs.

There is also a way to command to npm to download a specific version of the typings correlated with the TS version using `npm install @types/knockout.mapping@ts2.2`
Although I know this command is not popular.

[Examples of versions from the  "versions" in npm site:](https://www.npmjs.com/package/@types/knockout.mapping). 
image



I don't think it's a good idea to limit development and usage of new features, because the userbase is using older versions. :disappointed:

Igorbek: I'm sorry, I should've been clearer. I mean this types redirection: https://github.com/microsoft/TypeScript/wiki/What%27s-new-in-TypeScript#version-selection-with-typesversions

So it allows to have types for lower versions of TS as well as newer versions, in the same package.

ltlombardi: The problem I see is that it only works in TS 3.1 or above. So people with TS 2.8 thru 3.0 would be getting the older knockout.mapping file. Basically the version present in the PR will only be available to TS 3.1 users or above. 

I guess this is acceptable, but I really believe this is not the way to do it.

My reasoning:
1) If the project doesn't use a package delivery system, the dev already included the `.d.ts` file in it's project repository, so this PR has no harm for the legacy project and it's developer.
2) If the project uses Nuget, again, this changes won't affect him. He is stuck to the typings version in his projects.
3) If the project uses NPM, with a specific version set, again, this changes won't affect him. He is stuck to the typings version in his projects.
4) If the project uses NPM, without any version established, then this is the devs fault / decision to use the latest version and deal with the consequences. 

So my conclusion is that this won't really harm anyone already using knockout.mapping. 
Someone that will start to use now, will be actively developing on a legacy code... so It won't really be legacy, will it? That person would be more than better off upping the TSC version. But again he is not required, he can just set a specific version in NPM packages.json and be happy.

ltlombardi: Hey @Igorbek , the bot wrongly things there has been no activity on this PR and will close it tomorrow. Can you put a hold on it?

ltlombardi:
:-(
Can it be reopen? It would be a shame to loose the discussion creating a new PR.

ltlombardi:
@Igorbek , it's been 2 weeks. I put a lot of time and effort on this code. Could you get back to this?

rbuckton:
@ltlombardi: I've reopened the PR per your request. The build is failing, so can you ensure your branch is up-to-date with the latest master?

I think the questions to be addressed are these:

1. Will increasing the TypeScript Version to 2.8 negatively impact anyone in the community?
2. Will requiring TypeScript 3.1 or later for the newer definitions be a barrier to anyone wanting to use knockout.mapping?

Unfortunately, there's no way of knowing how many people are using TypeScript 2.8 with knockout.mapping. TypeScript 2.8 was published over a year ago, so it is reasonably safe to upgrade the version number without requiring `"typesVersions"` support, as the older versions will still remain on NPM. 

Unless @Igorbek has any additional concerns, I will review this again over the next few days.

rbuckton:
I think typescript-bot will continue to close this until a new commit is made. If you can update the PR with the latest changes in master then I'll reopen it.

ltlombardi:
Thank you @rbuckton . Branch is updated now.
I believe the branch is ok to merge and won't negatively impact anyone in the community, because NPM already supports versioning, as you stated.

maskmaster:
First, let me say that an update to the mapping types is sorely needed.

knockout.mapping might be seen as legacy, but the people that have invoked Typescript for it is definitely not using it as legacy, instead want to improve the code quality and catch more mistakes. I am all for using newer ts-versions for this package.

There was an old PR #32210 previously that unfortunately seems to have been abandoned. I think that PR really nailed the `fromJS` function. Please take a look at that PR for inspiration on how to catch some more edge cases. Also, something similar could be implemented for the `toJS` for flawless conversions.

rbuckton:
Ok, reopening (hopefully typescript-bot will leave it alone)… I'll review this now.

types/knockout/test/index.ts
```diff
@@ -173,7 +173,7 @@ function test_bindings() {
 
     ko.bindingHandlers.yourBindingName = {
         init: function (element, valueAccessor, allBindingsAccessor, viewModel, bindingContext) {
-        	return { "controlsDescendantBindings": true };
+            return { "controlsDescendantBindings": true };
```
rbuckton: Are these indentation changes needed?

types/knockout.mapping/index.d.ts
```diff
@@ -15,13 +15,23 @@ export = self;
 
 
 declare global {
-    // the idea behind this is good, but doesn't work as intended. This will make object properties T | Observable<T>, 
-    //and you will have to put the correct Type in each property to use them, what defeats the purpose. Besides, this gives  
-    //RangeError: Maximum call stack size exceeded in TSC when used in all the mapping signatures. Maybe this can be used with TS 2.8 and conditional typing
+
+    type Primitives = string | number | boolean;
```
rbuckton: Is this the full list of primitives supported by knockout, or should this also include `bigint | symbol`?

types/knockout.mapping/index.d.ts
```diff
@@ -15,13 +15,23 @@ export = self;
...
+
+    type MappedType<T> =
+        T extends Primitives ? KnockoutObservable<T> :
+        T extends object ? KnockoutObservableType<T> :
+        any;
+
     type KnockoutObservableType<T> = {
-        [P in keyof T]: KnockoutObservable<KnockoutObservableType<T[P]>> | T[P];
+        [P in keyof T]: T[P] extends Primitives ? KnockoutObservable<T[P]> :
+                        T[P] extends any[] ? KnockoutObservableArrayType<T[P][number]> :
```
rbuckton: What happens here if `T[P]` is a `ReadonlyArray<any>` ?

types/knockout.mapping/index.d.ts
rbuckton: Also, I would rather see the types for `Primitives` inlined below rather than have its own declaration, or defined *outside* of the `declare global`. Adding something like `Primitives` to the global scope is bound to end up with a "duplicate identifier" conflict at some point.

types/knockout.mapping/knockout.mapping-tests.ts
```diff
@@ -1,105 +1,160 @@
...
-};
+    create: (options: KnockoutMappingCreateOptions) => { }, // $ExpectError
+}
 
 let userMappingOptions: KnockoutMappingOptions<User> = {
-    ignore: ['age'],
-    include: ['name'],
-    copy: ['height'],
-    mappedProperties: ['age', 'name'],
+    ignore: ["age"],
```
rbuckton: Is there a reason to change the quotemark? The Prettier config for DT uses single quotes, so you should stay consistent with that:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3ad3eefe64b17cbb77f2a1a3932f72f1be73c028/.prettierrc.json#L7

types/knockout.mapping/tslint.json
```diff
@@ -50,21 +46,17 @@
...
         "one-line": false,
         "one-variable-per-declaration": false,
-        "only-arrow-functions": false,
-        "prefer-conditional-expression": false,
-        "prefer-const": false,
         "prefer-declare-function": false,
         "prefer-for-of": false,
         "prefer-method-signature": false,
         "prefer-template": false,
+        "quotemark": [true, "double", "avoid-escape" ],
```
rbuckton: The Prettier configuration for DT uses single quotes. Is this lint rule necessary?

rbuckton (changes requested):
A few minor comments.

types/knockout/test/index.ts
ltlombardi: Needed is a strong word. They were generated by vs code auto formatted and reviewed by me as **useful**. Original indentation was non-standard (didn't respect the rest of the file indentation).

types/knockout.mapping/index.d.ts
ltlombardi: Added support for this scenario.

types/knockout.mapping/index.d.ts
ltlombardi: Added support for this scenario, and made necessary changes.

types/knockout.mapping/knockout.mapping-tests.ts
ltlombardi: Repo didn't use prettier back when I did the changes. I don't use it my self. Maybe someday. 
I saw that each project used it's own formatting guidelines, so used my opinionated in this one. I prefer double quotes because it is consistent with other common languages like C# and Java. 

In the interest of consistency with this repo and saving time, against my preference, I reverted this changes.

types/knockout.mapping/tslint.json
ltlombardi: Explained in other comment.

ltlombardi:
@rbuckton , It's been two weeks 😅

types/knockout.mapping/index.d.ts
Igorbek: @ltlombardi I think it should be quite easy to follow suggesting described here: https://github.com/DefinitelyTyped/DefinitelyTyped#i-want-to-use-features-from-typescript-31-or-above

I still think that there're projects that use older versions of TS and wouldn't upgrade.
However, I don't think I can prevent from upgrading to new TS now, and if any update for older version is required in the future, we can make a new redirection folder.

Igorbek (approved):

ltlombardi (commented):
All done

PR #36030, opened 2019-06-07
Fix passReqToCallback type in passport-azure-ad

kaaloo:
Fixes the following compilation error:

error TS2345: Argument of type '{ allowHttpForRedirectUrl: true; clientID: string; clientSecret: string | undefined; identityMetadata: string; passReqToCallback: boolean; redirectUrl: string; responseMode: "form_post"; responseType: "code id_token"; scope: string[]; validateIssuer: false; }' is not assignable to parameter of type 'IOIDCStrategyOption'.
  Object literal may only specify known properties, and 'passReqToCallback' does not exist in type 'IOIDCStrategyOption'.

69     passReqToCallback: false,

Please fill in this template.

### DIFF
--- a/types/passport-azure-ad/oidc-strategy.d.ts
+++ b/types/passport-azure-ad/oidc-strategy.d.ts
@@ -21,6 +21,10 @@ export interface IOIDCStrategyOptionWithRequest extends IOIDCStrategyOption {
     passReqToCallback: boolean;
 }
 
+export interface IOIDCStrategyOptionWithoutRequest extends IOIDCStrategyOption {
+    passReqToCallback: false;
+}
+
 export interface IProfile {
     sub?: string;
     oid?: string;
@@ -37,7 +41,7 @@ export interface IProfile {
 }
 
 export type VerifyOIDCFunction =
-    ((profile: IProfile, done: VerifyCallback) => void) | 
+    ((profile: IProfile, done: VerifyCallback) => void) |
     ((iss: string, sub: string, done: VerifyCallback) => void) |
     ((iss: string, sub: string, profile: IProfile, done: VerifyCallback) => void) |
     ((iss: string, sub: string, profile: IProfile, access_token: string, refresh_token: string, done: VerifyCallback) => void) |
@@ -45,7 +49,7 @@ export type VerifyOIDCFunction =
     ((iss: string, sub: string, profile: IProfile, jwtClaims: any, access_token: string, refresh_token: string, params: any, done: VerifyCallback) => void);
 
 export type VerifyOIDCFunctionWithReq =
-    ((req: Request, profile: IProfile, done: VerifyCallback) => void) | 
+    ((req: Request, profile: IProfile, done: VerifyCallback) => void) |
     ((req: Request, iss: string, sub: string, done: VerifyCallback) => void) |
     ((req: Request, iss: string, sub: string, profile: IProfile, done: VerifyCallback) => void) |
     ((req: Request, iss: string, sub: string, profile: IProfile, access_token: string, refresh_token: string, done: VerifyCallback) => void) |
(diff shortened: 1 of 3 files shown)


kaaloo:
Hi, I need to do a bit of work to perhaps add some tests to this patch.  It's a very small change but please let me know if you require some more input from me.

types/passport-azure-ad/oidc-strategy.d.ts
```diff
@@ -18,7 +18,7 @@ export interface IOIDCStrategyOption extends IBaseStrategyOption {
 }
 
 export interface IOIDCStrategyOptionWithRequest extends IOIDCStrategyOption {
-    passReqToCallback: true;
+    passReqToCallback: boolean;
```
rbuckton: You should not change this. The `true` type for the property is required to pick a different overload in the constructor:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6148c753d84131f8c8581fa8974443e39d4d7a38/types/passport-azure-ad/oidc-strategy.d.ts#L56-L60

If you pass an `IOIDCStrategyOptionWithRequest` (which specifies `passReqToCallback: true`), then it picks the constructor overload that takes a `VerifyOIDCFunctionWithReq`. If you pass a regular `IOIDCStrategyOption`, it picks the overload that takes a `VerifyOIDCFunction`.

If the problem is that you need to pass `false` for `passReqToCallback` then you may need to add a separate `IOIDCStrategyOptionWithoutRequest` interface like this:

```ts
export interface IOIDCStrategyOptionWithoutRequest extends IOIDCStrategyOption {
  passReqToCallback?: false;
}
```

And then replace the constructor overload on Line 60:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6148c753d84131f8c8581fa8974443e39d4d7a38/types/passport-azure-ad/oidc-strategy.d.ts#L60

With this:

```ts
    constructor(options: IOIDCStrategyOptionWithoutRequest, verify: VerifyOIDCFunction);
```

rbuckton (changes requested):

types/passport-azure-ad/oidc-strategy.d.ts
kaaloo: Thank you for your detailed comment!  I updated the `pasReqToCallback` type locally to `boolean` and I'm indeed passing `false`.  Things seem to be working ok but I have not tried passing `true` since that is not my use case.  Presumably the first constructor is still being called. It is somewhat confusing to me to see that `pasReqToCallback` is being used as a tag though since I was expecting it to be a simple property.  I will try your approach on my project and post my results.

kaaloo:
Hi @rbuckton , I updated my branch according to your instructions and added a quick test.

types/passport-azure-ad/oidc-strategy.d.ts
rbuckton:
```suggestion
    passReqToCallback: true;
```

This needs to stay `true` for the correct overload selection to work.

rbuckton (changes requested):

rbuckton:
I would suggest adding tests for both cases to ensure overload resolution is working correctly.

kaaloo:
Oops, sorry about the leftover "true".  Thanks for catching that.  There are actually tests for both cases.  I added a test with `passReqToCallback: false` but there were already existing tests for the `passReqToCallback: true` case.  In fact that is why this issue never came up.  Existing tests did not test for the `false` case.

Rick-Kirkham:
@kaaloo @rbuckton Treating true and false as "types" seems strange to put it mildly. It also causes a compiler error. If I have a config object with Boolean property passReqToCallback set to `false`, then the following errors: 

```
passport.use(new OIDCStrategy({
    ...
    
    passReqToCallback: myConfigObject.passReqToCallback, // Error: "Type Boolean is not assignable to type false."
    ...
  }, ...
```

To fix it, I have to cast the value to "type false": 

```
passport.use(new OIDCStrategy({
    ...
    
    passReqToCallback: <false>myConfigObject.passReqToCallback, 
    ...
  }, ...
```

This seems hacky. What is it that you are trying to accomplish by treating true and false as types?

PR #30663, opened 2018-11-19
[country-list] Updated country-list types to work with version 2.1

Kimahriman:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/fannarsh/country-list/commit/ef4633adc993834fea4ba11df7bc75f627df1b0f https://github.com/fannarsh/country-list/commit/a0102f2839011665abcd02224eac1e68a9def5f4
- [x] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #36036, opened 2019-06-07
[styled-components] Omit overlapping props from the base type of wrapped components.

ajmadsen:
Context:

We, internally, are building a UI Component Library based on Styled Components & Styled System.
Styled System leverages helper functions that accept arrays or maps of potential values and translate those to the appropriate string. This exposed the issue demonstrated in the test I've added which I've added functionality to cover: _that props that overlap with provided html attributes intersect incorrectly resulting in TS errors  but working functionality_.

I've tried to think of reasons why overlapping props should have their types merged with the base component, and couldn't think of a valid example. Nonetheless, I'm concerned that this touches the core type.

### DIFF
--- a/types/styled-components/index.d.ts
+++ b/types/styled-components/index.d.ts
@@ -74,13 +74,10 @@ export type StyledComponentProps<
     A extends keyof any
 > = WithOptionalTheme<
     Omit<
-        OmitOverlapping<
-            ReactDefaultizedProps<
-                C,
-                React.ComponentPropsWithRef<C>
-            >,
-            O
-        > & O,
+        ReactDefaultizedProps<
+            C,
+            OmitOverlapping<React.ComponentPropsWithRef<C>, O> & O
+        >,
         A
     > & Partial<Pick<React.ComponentPropsWithRef<C> & O, A>>,
     T
--- a/types/styled-components/index.d.ts
+++ b/types/styled-components/index.d.ts
@@ -56,6 +56,13 @@ type ReactDefaultizedProps<C, P> = C extends { defaultProps: infer D; }
     ? Defaultize<P, D>
     : P;
 
+// Omit props from P that are also contained in O. Follows similar logic to Defaultize
+type OmitOverlapping<P, O> = P extends any
+    ? string extends keyof P
+        ? P
+        : Pick<P, Exclude<keyof P, keyof O>>
+    : never;
+
 export type StyledComponentProps<
     // The Component from whose props are derived
     C extends keyof JSX.IntrinsicElements | React.ComponentType<any>,
@@ -67,9 +74,12 @@ export type StyledComponentProps<
     A extends keyof any
 > = WithOptionalTheme<
     Omit<
-        ReactDefaultizedProps<
-            C,
-            React.ComponentPropsWithRef<C>
+        OmitOverlapping<
+            ReactDefaultizedProps<
+                C,
+                React.ComponentPropsWithRef<C>
+            >,
+            O
         > & O,
         A
     > & Partial<Pick<React.ComponentPropsWithRef<C> & O, A>>,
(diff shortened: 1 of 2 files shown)


types/styled-components/index.d.ts
```diff
@@ -56,6 +56,13 @@ type ReactDefaultizedProps<C, P> = C extends { defaultProps: infer D; }
     ? Defaultize<P, D>
     : P;
 
+// Omit props from P that are also contained in O. Follows similar logic to Defaultize
+type OmitOverlapping<P, O> = P extends any
+    ? string extends keyof P
```
eps1lon: This seems like C&P from `Defaultize`? I think the original intent was to distribute this type over unions. This is already the case by using the conditional above (`P extends any`) and can therefore be omitted.

eps1lon (changes requested):
Be aware that a similar change had to be rolled back previously because it caused a sever performance regression in typescript 3.4. This should be fine in 3.5 as per https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36036#issuecomment-500023685. See #30663 for more context

types/styled-components/index.d.ts
ajmadsen: It was C&P from `Defaultize` because without it, I get complaints from its usage down below with the `A` type parameter.

```
Type 'A' is not assignable to type 'Exclude<symbol, keyof O>'.
    Type 'string | number | symbol' is not assignable to type 'Exclude<symbol, keyof O>'.
        Type 'string' is not assignable to type 'Exclude<symbol, keyof O>'.
```

in the context

```
export type StyledComponentProps<
    // The Component from whose props are derived
    C extends keyof JSX.IntrinsicElements | React.ComponentType<any>,
    // The Theme from the current context
    T extends object,
    // The other props added by the template
    O extends object,
    // The props that are made optional by .attrs
    A extends keyof any
> = WithOptionalTheme<
    Omit<
        ReactDefaultizedProps<
            C,
            OmitOverlapping<React.ComponentPropsWithRef<C>, O> & O
        >,
        A
    > & Partial<Pick<React.ComponentPropsWithRef<C> & O, A>>,
    T
> & WithChildrenIfReactComponentClass<C>;
```

types/styled-components/index.d.ts
```diff
@@ -74,13 +74,10 @@ export type StyledComponentProps<
...
 > = WithOptionalTheme<
     Omit<
-        OmitOverlapping<
-            ReactDefaultizedProps<
-                C,
-                React.ComponentPropsWithRef<C>
-            >,
-            O
-        > & O,
+        ReactDefaultizedProps<
```
ajmadsen: I realized looking back here that I likely initially mis-ordered the application here, we'd still want `defaultProps` to apply `Partial` after re-picking the correct types.

types/styled-components/index.d.ts
ajmadsen: I wasn't sure how else to work around it, but the reasoning around `Defaultize` seems like it would apply here as well.

eps1lon: Yep got it. I guess if we omit `P extends any` we don't distribute over union props anymore and it squashes properties. 

Just to be clear: If you remove the `P extends any` it also fails the introduced tests?

ajmadsen: Yeah, it fails similarly.

```
Type 'A' is not assignable to type 'Exclude<keyof StyledComponentPropsWithRef<C>, keyof O>'.
  Type 'string | number | symbol' is not assignable to type 'Exclude<keyof StyledComponentPropsWithRef<C>, keyof O>'.
    Type 'string' is not assignable to type 'Exclude<keyof StyledComponentPropsWithRef<C>, keyof O>'.
```

ajmadsen:
I realized that I never properly tested `.attrs()` behavior with my changes, and ended up discovering that it wasn't behaving properly. I added some tests to uncover that behavior, and made an attempt at fixing the mangled types in a similar way.

Issue #33809, opened 2019-03-12
[@types/hapi] ApplicationState problem


Ackos95:
- [x] I tried using the `@types/hapi` package and had problems.
- [x] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @rafaelsouzaf, @jhsimms, @SimonSchick, @saboya 

I have a problem with definition of `ApplicationState`, and `Request.payload` (same problem).

I tried to add some custom parameters to `server.app`, same as I saw in [documentation](https://hapijs.com/api#-serveroptionsapp), and I got compiler issues. First problem is on line where value is set.

```ts
server.app.key = 'value';  // Property 'key' does not exist on type 'ApplicationState'.
```

And second one is when reading that value (same issue).

Both can be hacked with casting, but from my perspective this is unnecessary (and ugly), as you can see in example bellow.

```ts
(<any> server.app).key = 'value';  // to fix first issue
// and when reading
console.log((<{ key: string }> server.app).key);  // pretty ugly
```

I checked source, and saw that `ApplicationState` is defined as an empty interface, I'm not sure why it is done like that, could you maybe update it to at least support adding new keys, or in better case to be parametrized (so that one could add specific type for it, through generators). Same thing goes for `Request.payload`, from which is also impossible to read values (which are validated through `Joi` validators) without casting it. My suggestion would be to maybe add generators to `Server` and `Request` (maybe some others as well).

Example bellows solves only problems I've come up on, it may need some more stuff, it's just to demonstrate the logic.

```ts
class Server<TApplicationState = {}> {
  /* rest of the code */
  app: TApplicationState,
}

export interface Request<TPayload = {}, TApplicationState = {}> extends Podium {
  app: TApplicationState,
  payload: TPayload,
  // rest of the code
}
```

Any opinions on subject?

saboya:
`ApplicationState` is an empty interface to allow extension through module augmentation.

This is how you would do it:

```typescript
declare module 'hapi' {
  interface ApplicationState {
    key: string,
  }
}
```

[...]

PR #36037, opened 2019-06-07
[hapi] Add generic types for ServerApplicationState and RequestApplicationState

sprice237:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33809

This is my first PR to an open source project, so I apologize if conventions were not followed. If there's anything I need to change, please let me know.


ptorian:
Thoughts? Am I on the right track with this, do I need to go a different direction?

jhsimms:
I'll review this soon. This is a pretty significant change. Are there breaking changes in here?

We periodically get PRs trying to generically type application and request state, but they usually aren't really an improvement over just casting. I'm not saying that's what this is, I just want to be sure in advance that's not the case.

ptorian:
As far as I could tell there are no breaking changes. All of the tests still pass and all of the new generic inputs default to `any` so completely omitting them shouldn't break anything. Only thing that would perform differently is if someone was taking the module augmentation approach for modifying `ApplicationState` interface, as that interface doesn't exist anymore. It won't break anything for those users, they would just lose typechecking. I wouldn't be opposed to defaulting the type generics to `ApplicationState` instead of `any` in principle to preserve that workflow, but that needs work, as currently `ApplicationState` is used as the type for both `request.app` and `server.app` when those are actually different types.

SimonSchick (commented):
This has the same issues as making requests generic for payloads etc. massive amounts of generics that will for most users default to `any` since it will be extremely tedious to pass those around all the time and make the lib harder to use.

Especially since there are many more other rather unspecified object that could be made generic as well.

ptorian:
If this isn't an appropriate way to do it, then how do we get type checking on `request.app` and `server.app`? Casting every time it's accessed isn't any better. I disagree that it makes the library harder to use; the way I use it, I create my own type alias to the hapi type with the generic type arguments provided and then I use my type alias everywhere instead of the hapi type, it works flawlessly for me.

jhsimms:
I'm in agreement with Simon on this. I'm saying this based on memory when I did a large amount of work for Hapi 18 and had the same idea. 

As I review this, I'm not sure we're getting much benefit. The problem is that the way Hapi is structured, some things are not accommodating to type checking, particularly with generics. The flip side to this is that it's a huge change and we're introducing a large maintenance burden on the Hapi types, for a suspect benefit.

In most cases, with request payloads, etc, server state, app state, the type system can't really connect the dots, and so you have to specify the generics all over the place. This is what I meant by it not being much benefit over casting. I know you mentioned you're aliasing the generic server type, but I see problems with this. For those that don't choose this method, if they drop the type, you're defaulting to any, so they may expect they have a more specific type but they have any, if they forget to generically type a function. Furthermore, in all cases where you've generically typed, can you confirm that if you specify a different type to the generic function, would it break, given what you've put on the server type? If we can't say this is 100% true, I'm not sure what we've gained here. It's basically glorified casting via generic type parameters.

jhsimms:
One more thing: isn't request.app potentially different by request?

ptorian:
`request.app` being different per request is a good point, but I don't see needing to pass around generics as a major con, it's how a lot of other type definitions handle problems like this. If passing around generics isn't the answer, what is the recommended way for defining a type for `server.app`?

ptorian:
@jhsimms I'm not 100% sure what you're getting at with the second half of your third paragraph, but if for example you define your server with defined `IServerApplicationState` and `IRequestApplicationState` types and then you try to add a route with a different `IServerApplicationState` or `IRequestApplicationState` type, it will do what you expect typescript to do and throw an error if the interface signatures are not equivalent.

If someone using the hapi classes sometimes provides generic types and sometimes doesn't, then yes, `any` will be assumed for the times they don't provide generic types and they won't have type checking, but I don't know how that's any different than it is now. With the exception of `request.app` possibly being different for every request (which is a good point and needs discussing), these modifications allow the use of static type checking for `server.app` and `request.app` if you want it, while defaulting to `any` if you don't want it. I'm open to suggestions to make it better, but I don't think outright rejection is the right answer. Needing configurable, static typing for `server.app` and `request.app` is a very valid use case.

RyanCavanaugh:
This is a fairly large change. I'd really like concrete sign-off from an existing owner -  @rafaelsouzaf @jhsimms @SimonSchick @saboya  ?

saboya:
Sorry about taking so long to look at this.

I'm not sure this is an objective quality of life upgrade compared to the module augmentation route. I think this is only a clear positive if you're dealing with multiple `Server` instances with different `ApplicationState` or `RequestApplicationState`, but I'd argue that this is rather uncommon. So I'd rather augment the hapi module once on my project  than having to pass generic parameters in all my route declarations.

Maybe with some practical examples you can shed a light into the benefits of change.

sandersn:
This PR has been open for too long with no discussion. I'm going to close it for now. Let me know if anything changes.

PR #36040, opened 2019-06-07
yup - Fix WhenOptions "is" prop constrained to boolean

shiznit013:
Please fill in this template.

- [ x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).

- [ x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/jquense/yup#mixedwhenkeys-string--arraystring-builder-object--value-schema-schema-schema
https://github.com/jquense/yup/blob/master/src/Condition.js#L26

"is" does not need to be a boolean. It will perform a === check for whatever value "is" is.

- [x ] Increase the version number in the header if appropriate.

Not necessary.

- [x ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

Changes are not substantial.


types/yup/index.d.ts
```diff
@@ -212,18 +212,29 @@ export type TransformFunction<T> = ((
     originalValue: any
 ) => any);
 
-export interface WhenOptionsBuilder<T> {
+export interface WhenOptionsBuilderFunction<T> {
```
YashdalfTheGray: I think we should call this type `WhenOptionsFunction<T>` because the docs call the second argument `builder` when it's an object.

types/yup/index.d.ts
```diff
@@ -212,18 +212,29 @@ export type TransformFunction<T> = ((
...
     (value: any, schema: T): T;
     (v1: any, v2: any, schema: T): T;
     (v1: any, v2: any, v3: any, schema: T): T;
     (v1: any, v2: any, v3: any, v4: any, schema: T): T;
 }
 
-export type WhenOptions<T> =
-    | WhenOptionsBuilder<T>
-    | { is: boolean | ((...values: any[]) => boolean); then: any; otherwise: any }
+export type WhenOptionsBuilderObjectIs =
```
YashdalfTheGray: Since `is` can actually be anything, are we losing of the type information if we just change this type to `((...values: any[]) => boolean) | any` rather than a union of all these types?

Furthermore, can we name this type something like `BuilderIsProperty` or `WhenIsProperty` or simply `IsProperty` so that it doesn't sound like a ruby method name?

types/yup/index.d.ts
```diff
@@ -212,18 +212,29 @@ export type TransformFunction<T> = ((
...
+    | ((...values: any[]) => boolean)
+    | boolean
+    | number
+    | null
+    | object
+    | string;
+
+export type WhenOptionsBuilderObject =
```
YashdalfTheGray: We should just call this `WhenOptionsBuilder` to match the documentation for yup.

YashdalfTheGray (changes requested):

types/yup/index.d.ts
shiznit013: `builder` is the name of the argument when it is an object or function. I was trying to stick with the existing naming scheme, but add more specificity.

types/yup/index.d.ts
shiznit013: Yes. Unioning anything with `any` resolves to just `any`. I believe this is even a linter warning. The way I have it allows for all the types I tested with, but still gives some intellisense/direction about what the `when` method expects.

I think `IsProperty` is too generic. What if there were another `is` property that took different types. Likewise with `BuilderIsProperty`. There could be other methods that take a builder of a different sort. Again, I was trying to stick with and expand the existing naming convention.

types/yup/index.d.ts
shiznit013: Referring back to the 1st comment. `builder` can be an object or function, so suffixing them makes sense to me to distinguish them.

types/yup/index.d.ts
YashdalfTheGray: Yeah, okay. I'll concede. 😄

types/yup/index.d.ts
YashdalfTheGray: I'm still not a fan of the name but I don't see that as a reason to block this PR from being merged.

YashdalfTheGray (approved):

PR #36041, opened 2019-06-07
[styled-system] Update typings to v5

elliotbonneville:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://styled-system.com/guides/migrating

## Update notes

I'm relatively new to TypeScript, so the `system` and `compose` typings may not be totally correct. I tried to make sure they were accurate, but they would certainly benefit from a more experienced eye. In particular, both functions accept and return `styleFn`, which just accepts and returns `any` values. This seems necessary to allow `styled-system` to be slurped up by the flavor-of-the-day CSS-in-JS library, but I would be much happier with stricter typings.

### Testing

With the introduction of groupings by category for built in style functions, testing becomes much simpler. We can simply import the typings for each category (which in turn contain the individual style function typings) and test on those instead. I think this should be fine, but let me know if it isn't and we can revert to the more verbose style.

### New typings

I added `paddingX`, `paddingY`, `marginX`, and `marginY` typings to the `SpaceProps` interface. I'm not sure if these were introduced in v5 or if there was a reason for not including them in previous version, but happy to remove these if needed.

I also added `borderX` and `borderY` typings to the `BorderProps` interface.

Finally, I introduced `FlexGrowProps` and `FlexShrinkProps` interfaces, which did not previously exist but are used in the new `flexbox` style functions category.

### DIFF
--- a/types/styled-system/index.d.ts
+++ b/types/styled-system/index.d.ts
@@ -783,7 +783,7 @@ export interface GridTemplateRowsProps<TLength = TLengthStyledSystem> {
 
 export const gridTemplateRows: styleFn;
 
-export interface GridTemplatesAreasProps {
+export interface GridTemplateAreasProps {
     /**
      * The grid-template-areas CSS property specifies named grid areas.
      *
--- a/types/styled-system/index.d.ts
+++ b/types/styled-system/index.d.ts
@@ -984,9 +984,8 @@ export const border: styleFn;
 
 export interface BoxShadowProps {
     /**
-     * The box-shadow CSS property adds shadow effects around an element's frame. You can set multiple effects
-     * separated by commas. A box shadow is described by X and Y offsets relative to the element, blur and spread
-     * radii, and color.
+     * The box-shadow CSS property adds shadow effects around an element's frame. You can set multiple effects separated
+     * by commas. A box shadow is described by X and Y offsets relative to the element, blur and spread radii and color.
      *
      * [MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow)
      */
@@ -994,6 +993,24 @@ export interface BoxShadowProps {
 }
 
 export const boxShadow: styleFn;
+
+export interface TextShadowProps {
+    /**
+     * The `text-shadow` CSS property adds shadows to text. It accepts a comma-separated list of shadows to be applied
+     * to the text and any of its `decorations`. Each shadow is described by some combination of X and Y offsets from
+     * the element, blur radius, and color.
+     *
+     * [MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/text-shadow)
+     */
+    textShadow?: ResponsiveValue<CSS.TextShadowProperty | number>;
+}
+
+export const textShadow: styleFn;
+
+export interface ShadowsProps
+    extends BoxShadowProps, TextShadowProps {}
+
+export const shadows: styleFn;
 
 export interface OpacityProps {
     /**
--- a/types/styled-system/index.d.ts
+++ b/types/styled-system/index.d.ts
@@ -51,7 +51,8 @@ export interface CustomStyleDefinition {
     /** The CSS property to use in the returned style object (overridden by `properties` if present). */
     property?: string;
     /** an array of multiple properties (e.g. `['marginLeft', 'marginRight']`) to which this style's value will be
-      * assigned (overrides `property` when present). */
+      * assigned (overrides `property` when present).
+      */
     properties?: string[];
     /** A string referencing a key in the `theme` object. */
     scale?: string;
--- a/types/styled-system/index.d.ts
+++ b/types/styled-system/index.d.ts
@@ -50,9 +50,10 @@ export interface styleFn {
 export interface CustomStyleDefinition {
     /** The CSS property to use in the returned style object (overridden by `properties` if present). */
     property?: string;
-    /** an array of multiple properties (e.g. `['marginLeft', 'marginRight']`) to which this style's value will be
-      * assigned (overrides `property` when present).
-      */
+    /**
+     * An array of multiple properties (e.g. `['marginLeft', 'marginRight']`) to which this style's value will be
+     * assigned (overrides `property` when present).
+     */
     properties?: string[];
     /** A string referencing a key in the `theme` object. */
     scale?: string;
--- a/types/styled-system/index.d.ts
+++ b/types/styled-system/index.d.ts
@@ -21,7 +21,13 @@ import * as CSS from "csstype";
 
 export function get(obj: any, ...paths: Array<string | number>): any;
 
-export type Scale = object | Array<string | number>;
+export type ObjectOrArray<T> = T[] | { [K: string]: T | ObjectOrArray<T> };
+
+export type Scale = ObjectOrArray<number | string>;
+
+export type TLengthStyledSystem = string | 0 | number;
+
+export type ResponsiveValue<T> = T | Array<T | null> | { [key: string]: T };
 
 // Preserved to support v4 shim:
 // https://github.com/styled-system/styled-system/blob/master/packages/styled-system/src/index.js#L108
@@ -45,47 +51,44 @@ export function style<N = string | number, S = Scale>(
 
 export interface styleFn {
     (...args: any[]): any;
+    config?: object;
+    propNames?: string[];
+    cache?: object;
 }
 
-export interface CustomStyleDefinition {
+export interface ConfigStyle {
     /** The CSS property to use in the returned style object (overridden by `properties` if present). */
-    property?: string;
+    property?: keyof CSS.Properties;
     /**
      * An array of multiple properties (e.g. `['marginLeft', 'marginRight']`) to which this style's value will be
      * assigned (overrides `property` when present).
      */
-    properties?: string[];
+    properties?: Array<keyof CSS.Properties>;
     /** A string referencing a key in the `theme` object. */
     scale?: string;
     /** A fallback scale object for when there isn't one defined in the `theme` object. */
-    defaultScale?: any[];
+    defaultScale?: Scale;
     /** A function to transform the raw value based on the scale. */
     transform?: (value: any, scale?: Scale) => any;
 }
 
-export interface CustomStyleDefinitions {
+export interface Config {
     /** Property name exposed for use in components */
-    [customStyleName: string]: CustomStyleDefinition | boolean;
+    [customStyleName: string]: ConfigStyle | boolean;
 }
 
 export function compose(...parsers: styleFn[]): styleFn;
-export function system(styleDefinitions: CustomStyleDefinitions): styleFn;
+export function system(styleDefinitions: Config): styleFn;
 
 export interface VariantArgs {
     key?: string;
-    // Defaults to "variant"
+    /** Component prop, defaults to "variant" */
     prop?: string;
+    /** theme key for variant definitions */
+    scale?: string;
 }
 
 export function variant(props: VariantArgs): (...args: any[]) => any;
-
-export type TLengthStyledSystem = string | 0 | number;
-export type ResponsiveValue<T> =
-    | T
-    | Array<T | null>
-    | {
-          [key: string]: T;
-      };
 
 /**
  * Converts shorthand or longhand margin and padding props to margin and padding CSS declarations
@@ -174,12 +177,9 @@ export interface MarginProps
         | "mx"
     > {}
 export interface MarginTopProps extends Pick<SpaceProps, "mt" | "marginTop"> {}
-export interface MarginBottomProps
-    extends Pick<SpaceProps, "mb" | "marginBottom"> {}
-export interface MarginLeftProps
-    extends Pick<SpaceProps, "ml" | "marginLeft"> {}
-export interface MarginRightProps
-    extends Pick<SpaceProps, "mr" | "marginRight"> {}
+export interface MarginBottomProps extends Pick<SpaceProps, "mb" | "marginBottom"> {}
+export interface MarginLeftProps extends Pick<SpaceProps, "ml" | "marginLeft"> {}
+export interface MarginRightProps extends Pick<SpaceProps, "mr" | "marginRight"> {}
 
 export const margin: styleFn;
 export const marginTop: styleFn;
@@ -203,14 +203,10 @@ export interface PaddingProps
         | "py"
         | "px"
     > {}
-export interface PaddingTopProps
-    extends Pick<SpaceProps, "pt" | "paddingTop"> {}
-export interface PaddingBottomProps
-    extends Pick<SpaceProps, "pb" | "paddingBottom"> {}
-export interface PaddingLeftProps
-    extends Pick<SpaceProps, "pl" | "paddingLeft"> {}
-export interface PaddingRightProps
-    extends Pick<SpaceProps, "pr" | "paddingRight"> {}
+export interface PaddingTopProps extends Pick<SpaceProps, "pt" | "paddingTop"> {}
+export interface PaddingBottomProps extends Pick<SpaceProps, "pb" | "paddingBottom"> {}
+export interface PaddingLeftProps extends Pick<SpaceProps, "pl" | "paddingLeft"> {}
+export interface PaddingRightProps extends Pick<SpaceProps, "pr" | "paddingRight"> {}
 
 export const padding: styleFn;
 export const paddingTop: styleFn;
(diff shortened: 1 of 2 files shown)


peduarte:
Nice, gonna do a proper review on Monday.

peduarte:
I mean Tuesday 😂

peduarte:
I mean Wednesday 🙄

types/styled-system/index.d.ts
```diff
@@ -13,50 +13,63 @@
...
-export interface LowLevelStylefunctionArguments<N, S> {
-  prop: string;
-  cssProperty?: string;
-  alias?: string;
-  key?: string;
-  transformValue?: (n: N, scale?: S) => any;
-  scale?: S;
+export interface CustomStyleDefinition {
+    /** The CSS property to use in the returned style object (overridden by `properties` if present). */
+    property?: string;
```
focux: You can force here to use a valid CSS property, e.g.
```ts
property?: (keyof CSS.Properties)
```

types/styled-system/index.d.ts
```diff
@@ -13,50 +13,63 @@
...
+    /**
+     * An array of multiple properties (e.g. `['marginLeft', 'marginRight']`) to which this style's value will be
+     * assigned (overrides `property` when present).
+     */
+    properties?: string[];
```
focux: You can do the same here, e.g.:
```ts
properties?: (keyof CSS.Properties)[]
```

types/styled-system/index.d.ts
```diff
@@ -13,50 +13,63 @@
...
+    /** A string referencing a key in the `theme` object. */
+    scale?: string;
+    /** A fallback scale object for when there isn't one defined in the `theme` object. */
+    defaultScale?: any[];
```
focux: Shouldn't this be a number's array?

focux (changes requested):
I'm glad this PR is already in place because I'm needing those types right now haha. I made a few reviews of types that can be improved.

types/styled-system/index.d.ts
peduarte: I think we should have a scale type

focux: How that type should look? Actually, I think it should be something like this:
```ts
defaultScale: number[] | string[];
```
Because if I'm not wrong, a scale can be a number array e.g. `[1, 4, 5]` or a string array: `['5px', '2rem']`

types/styled-system/index.d.ts
elliotbonneville: Awesome, I was looking for a way to do this!

types/styled-system/index.d.ts
```diff
@@ -13,50 +13,61 @@
...
+    /** Property name exposed for use in components */
+    [customStyleName: string]: CustomStyleDefinition | boolean;
+}
 
-export function mapProps(mapper: (...args: any[]) => any): (func: any) => (...props: any[]) => any;
+export function compose(...parsers: styleFn[]): styleFn;
+export function system(styleDefinitions: CustomStyleDefinitions): styleFn;
 
 export interface VariantArgs {
     key?: string;
```
focux: Variant function is missing `scale` property, should be of type string. ref: https://styled-system.com/variants/

focux (changes requested):

pleunv (changes requested):
Quickly browsing through the types: `zIndeces` on line 254 probably needs to be `zIndices`. Not sure whether to mention here or in a separate issue as this part of the code is untouched by the PR. (Can't add a line comment)

types/styled-system/index.d.ts
peduarte: A scale can also be nested object, for example:
```
theme = {
  colors: {
    gray: '#ccc',
    grays: [
      '#eee',
      '#ddd',
      '#333',
    ]
  }
}
```

peduarte:
@elliotbonneville It's working well for me, tried it on [Radix](https://github.com/modulz/radix/pull/128) and didn't have any issues. 

Looking forward to getting this release 🚀

peduarte:
@elliotbonneville One little thing I noticed when creating my custom Style function:

```js
import { system, CustomStyleDefinitions } from 'styled-system';

const config: CustomStyleDefinitions = {
  textColor: {
    property: 'color',
    scale: 'colors',
  },
};

export const textColor = system(config);
```

When we call `system` with a config, it adds a few extra props, such as `propNames`. 
In our current typings, `system` returns a `styleFn`, which doesn't include `propNames`

So...

```
textColor.propNames
```

image

peduarte:
Also while we're at it, I've noticed that `GridTemplatesAreasProps` is named wrong, it should be `GridTemplateAreasProps`

types/styled-system/index.d.ts
jxnblk: This is something I've been curious about as well, since I'm less familiar with TS. Since scales are defined in a user's own `theme` object, wouldn't it be ideal to define the scale types to match the schema that you're using in your theme? That might be easier said that done, but I'm curious of how something like that *could* be supported from Styled System side

peduarte:
@pleunv 
>Quickly browsing through the types: zIndeces on line 254 probably needs to be zIndices.

Done

types/styled-system/index.d.ts
peduarte: Done

types/styled-system/index.d.ts
peduarte: @jxnblk yeah this is the default theme, but can be extended on the user's side. Since the theme is dynamic is impossible for us to type check before hand.

On that note, maybe the Theme shouldn't be so flexible :) haha let's not go there.

types/styled-system/index.d.ts
peduarte: Done

types/styled-system/index.d.ts
peduarte: Done

peduarte:
@elliotbonneville 

Pushed a few changes. Addressed all PR Comments.

Renamed a few things, take a look and let me know.

Also updated the `Theme` typing to match v5.

peduarte (approved):

types/styled-system/index.d.ts
pleunv: Yeah, I doubt this is something that can be tackled on the side of styled-system. Coincidentally I dug into this myself this week as we're currently adopting typescript and styled-system at the same time. If you go the custom type route you can do cool things like type the length & values of your scales with a tuple, i.e.

```
import { Theme } from 'styled-system';
import { ColorProperty } from 'csstype';

type CustomTheme = Theme & {
  colors: {
    red: [ColorProperty, ColorProperty, ColorProperty, ColorProperty, ColorProperty]
  }
};
```

But you'll have to implement your own types and you don't really have a way to easily feed these into the default styled-system setup. Plus with the amount of flexibility styled-system offers out of the box I doubt you'll ever be able to strictly type this. Also utils like `themeGet` and color props that take deeply nested string values are pretty much untypeable.

peduarte:
Added createParser and createStyleFunction functions

Lavoaster:
On my own perusal through this I can see that it's missing the new `shadows` function.

I intend to go through this and compare any other functions /w styled-system v5.

peduarte:
I've been thinking about something. Styled System v5 is now publishing modules under different packages.

For example, you can now only import `@styled-system/core` if you just need the core functions, such as `system`, `get`, etc...

Alternatively, you can install `@styled-system/space` to get `margin` and `padding`.

Here's a [list of all packages](https://github.com/styled-system/styled-system/blob/master/packages/space/src/index.js).

Right now I'm building a wrapper on top of `@styled-system/core` and this package won't work for me, I need `@types/styled-system__core`.

What do you guys think?

focux:
@peduarte I think we could keep the same repository but separating all the types by the modules they belong. e.g. 
```typescript
declare module '@styled-system/core' {
  ...
}

declare module '@styled-system/space' {
  ...
}

```

peduarte:
@focux interesting, I thought the convention was to create a new package.

I may be wrong, but when you declare the modules like that, they're not included in the bundle, are they?

focux:
@peduarte Actually, I don't know what the convention is, I just thought it will be easier to maintain this way, but you are right, we should research what the convention is.

And ya, modules are included in the bundle.

peduarte:
Yeah the thing is when you install `@styled-system/core` you'll also install `@types/styled-system__core` 

And that'll live under `node_modules/types` so everything works out of the box.

Looking at other packages, that's how they do it.

focux:
@peduarte  Great, then that's how we should do it. Also, besides publishing modules under different packages, they also publish the full bundle with all the packages, so we should look for a way to keep this repo up to date, reflecting the updates made on the submodules.

For example, if we update `@styled-system/core`, automatically reflect the updates in this repo. Maybe we could install all the submodules as dependencies and export them all.

Also, how the other modules handle shared dependencies? e.g. `styledFn`.

peduarte:
@focux great points. To be fair, I don't know. This is all new to me 😂 

But I would like to be able to install `@types/styled-system__core` as a dependency of other packages

peduarte:
UPDATE:

For now, I've published Style System's Core package under `@peduarte/styled-system__core` so I could continue with what I was working on.

If anyone's interested in using it, you can alias it to `@types/styled-system__core` like this:

```sh
yarn add @types/styled-system__core@npm:@peduarte/styled-system__core
```

types/styled-system/index.d.ts
```diff
@@ -709,20 +808,51 @@ export interface GridAreaProps {
...
+        GridColumnGapProps,
+        GridRowGapProps,
+        GridColumnProps,
+        GridRowProps,
+        GridAutoFlowProps,
+        GridAutoColumnsProps,
+        GridAutoRowsProps,
+        GridTemplateColumnsProps,
+        GridTemplateRowsProps,
+        GridTemplatesAreasProps,
```
rathpc: I think this has been an ongoing typo that just keeps popping up. Shouldn't this be:

```suggestion
        GridTemplateAreasProps,
```

rathpc: Also on line 786

rathpc (changes requested):

types/styled-system/index.d.ts
rathpc: Don't forget to refactor the other misspelled references as well 👍 thanks!

rathpc:
I'd love to help get this out the door if you all would like another pair of hands to address feedback. Let me know 🎉

types/styled-system/index.d.ts
elliotbonneville: > This is all pretty cool & interesting stuff but maybe it should be part of a separate discussion somewhere :)

Yeah, very cool discussion, but I agree that it doesn't fall under the scope of this PR. Unless there are any objections, let's leave in the `any` scale typing for now.

types/styled-system/index.d.ts
elliotbonneville: Thanks for the catch, @rathpc. Refactored other references and added a test.

elliotbonneville:
> On my own perusal through this I can see that it's missing the new `shadows` function.
> 
> I intend to go through this and compare any other functions /w styled-system v5.

Good spot on the `shadows` function -- I added this and also `textShadow` which was also not typed. I also ran through and checked quickly to see if there were any other functions missing, but didn't see any. Feel free to verify that.

> For example, if we update @styled-system/core, automatically reflect the updates in this repo. Maybe we could install all the submodules as dependencies and export them all.

@focux and @peduarte Super interesting discussion, I'd love to see this done. However, it's outside the scope of this PR, so let's get this merged and then figure out the best way to do that. Sound good?

> I'd love to help get this out the door if you all would like another pair of hands to address feedback. Let me know 🎉

Thanks, @rathpc! I think we have everything covered by now.

## Fin

Are we ready to merge this? As far as I can see, all outstanding comments have been addressed, build is passing, and this should be good to go.

rathpc (approved):
Nice work!

alloy:
As I have no experience with v5 yet, I’m waiting to hear from @focux and @pleunv before I'll merge.

focux:
@alloy Looks good to me!

pleunv (approved):

types/styled-system/index.d.ts
```diff
@@ -843,6 +994,24 @@ export interface BoxShadowProps {
...
+     */
+    textShadow?: ResponsiveValue<CSS.TextShadowProperty | number>;
+}
+
+export const textShadow: styleFn;
+
+export interface ShadowsProps
+    extends BoxShadowProps, TextShadowProps {}
+
+export const shadows: styleFn;
```
erickeno: Isn't this supposed to be `shadow` instead of shadows?
https://github.com/styled-system/styled-system/blob/master/packages/shadow/src/index.js#L14

types/styled-system/index.d.ts
```diff
@@ -843,6 +994,24 @@ export interface BoxShadowProps {
...
+     * the element, blur radius, and color.
+     *
+     * [MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/text-shadow)
+     */
+    textShadow?: ResponsiveValue<CSS.TextShadowProperty | number>;
+}
+
+export const textShadow: styleFn;
+
+export interface ShadowsProps
```
rathpc:
```suggestion
export interface ShadowProps
```

types/styled-system/styled-system-tests.tsx
```diff
@@ -1,316 +1,108 @@
...
     Theme,
+    layout,
+    LayoutProps,
+    typography,
+    TypographyProps,
+    flexbox,
+    FlexboxProps,
+    grid,
+    GridProps,
+    shadows,
```
rathpc:
```suggestion
    shadow,
```

types/styled-system/styled-system-tests.tsx
```diff
@@ -1,316 +1,108 @@
...
+    ShadowsProps,
```
rathpc:
```suggestion
    ShadowProps,
```

types/styled-system/styled-system-tests.tsx
```diff
@@ -1,316 +1,108 @@
...
         BackgroundProps,
-        BackgroundRepeatProps,
-        BackgroundSizeProps,
-        BorderColorProps,
         BorderProps,
         BordersProps,
-        BottomProps,
-        BoxShadowProps,
-        ColorProps,
+        ShadowsProps,
```
rathpc:
```suggestion
        ShadowProps,
```

types/styled-system/index.d.ts
rathpc: good catch! it should be

```suggestion
export const shadow: styleFn;
```

types/styled-system/styled-system-tests.tsx
```diff
@@ -1,316 +1,108 @@
...
+    typography,
+    flexbox,
+    border,
+    background,
+    color,
+    display,
+    background,
+    borders,
+    position,
+    shadows,
```
rathpc:
```suggestion
    shadow,
```

types/styled-system/index.d.ts
rathpc: the only place `shadows` plural should be is in themes

rathpc (changes requested):
per @erickeno 's catch I added a bunch of code suggestion changes to fix the misspelling of shadows -> shadow and ShadowsProps -> ShadowProps

types/styled-system/index.d.ts
```diff
@@ -13,62 +13,85 @@
...
+    /** A function to transform the raw value based on the scale. */
+    transform?: (value: any, scale?: Scale) => any;
+}
+
+export interface Config {
+    /** Property name exposed for use in components */
+    [customStyleName: string]: ConfigStyle | boolean;
+}
+
+export function compose(...parsers: styleFn[]): styleFn;
```
grrowl: just wondering if theres any way to type that _only_ `styleFn`s should be passed to `compose`? currently regular functions without the extra properties are allowed, since [all the extra properties are optional](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36041/files#diff-a483ed9502305df77fe429527806d00dR52). If any aren't truly optional for `styled-system` functions it'd be great to assert at least once, since `compose` will silently fail when a "regular" userland style function is passed in.

(p.s. thanks for doing this work! y'all are very fantastic)

alloy:
@elliotbonneville There’s a few suggestions by @rathpc, once those are in let’s get this merged and other stuff can be follow-ups?

types/styled-system/index.d.ts
alloy: Let’s make this a follow-up and not let perfect become the enemy of good enough.

peduarte:
Uh oh 😅

liamcmitchell-sc:
I opened a new PR with fixes suggested by @rathpc.

alloy:
Hah, can't blame the bot. Thanks @liamcmitchell-sc 👍

rathpc:
Thanks @liamcmitchell-sc !!!

PR #36047, opened 2019-06-08
[@ember/test-helpers] Add `Window` and `Document` for `triggerEvent()` and `triggerKeyEvent()`

gossi:
I added `Window` and `Document` as valid types for `triggerEvent()` and `triggerKeyEvent()`

---

Please fill in this template.

  - https://github.com/emberjs/ember-test-helpers/blob/master/addon-test-support/%40ember/test-helpers/dom/-get-element.ts
  - https://developer.mozilla.org/de/docs/Web/API/EventTarget/addEventListener

### DIFF
--- a/types/ember__test-helpers/ember__test-helpers-tests.ts
+++ b/types/ember__test-helpers/ember__test-helpers-tests.ts
@@ -56,6 +56,8 @@ test('DOM interactions', async () => {
     await triggerEvent(window, 'custom-event');
     await triggerEvent(document, 'custom-event');
     await triggerKeyEvent(messageElement, 'keydown', 'Enter', { ctrlKey: true });
+    await triggerKeyEvent(window, 'keydown', 'Enter', { ctrlKey: true });
+    await triggerKeyEvent(document, 'keydown', 'Enter', { ctrlKey: true });
     await fillIn(messageElement, 'content');
     await typeIn(messageElement, 'content');
 
--- a/types/ember__test-helpers/ember__test-helpers-tests.ts
+++ b/types/ember__test-helpers/ember__test-helpers-tests.ts
@@ -53,13 +53,15 @@ test('DOM interactions', async () => {
     await focus(messageElement);
     await blur(messageElement);
     await triggerEvent(messageElement, 'custom-event');
-    await triggerEvent(window, 'custom-event');
-    await triggerEvent(document, 'custom-event');
     await triggerKeyEvent(messageElement, 'keydown', 'Enter', { ctrlKey: true });
-    await triggerKeyEvent(window, 'keydown', 'Enter', { ctrlKey: true });
-    await triggerKeyEvent(document, 'keydown', 'Enter', { ctrlKey: true });
     await fillIn(messageElement, 'content');
     await typeIn(messageElement, 'content');
+
+    await triggerEvent(window, 'custom-event');
+    await triggerKeyEvent(window, 'keydown', 'Enter', { ctrlKey: true });
+
+    await triggerEvent(document, 'custom-event');
+    await triggerKeyEvent(document, 'keydown', 'Enter', { ctrlKey: true });
 
     const allMessages = findAll('.message');
     for (const element of allMessages) {
--- a/types/ember__test-helpers/ember__test-helpers-tests.ts
+++ b/types/ember__test-helpers/ember__test-helpers-tests.ts
@@ -53,6 +53,8 @@ test('DOM interactions', async () => {
     await focus(messageElement);
     await blur(messageElement);
     await triggerEvent(messageElement, 'custom-event');
+    await triggerEvent(window, 'custom-event');
+    await triggerEvent(document, 'custom-event');
     await triggerKeyEvent(messageElement, 'keydown', 'Enter', { ctrlKey: true });
     await fillIn(messageElement, 'content');
     await typeIn(messageElement, 'content');
--- a/types/ember__test-helpers/index.d.ts
+++ b/types/ember__test-helpers/index.d.ts
@@ -94,7 +94,7 @@ declare module '@ember/test-helpers/dom/blur' {
 declare module '@ember/test-helpers/dom/trigger-event' {
     import { Target } from '@ember/test-helpers';
 
-    export default function(target: Target | Window, eventType: string, options?: object): Promise<void>;
+    export default function(target: Target | Window | Document, eventType: string, options?: object): Promise<void>;
 }
 
 declare module '@ember/test-helpers/dom/trigger-key-event' {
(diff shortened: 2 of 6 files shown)


types/ember__test-helpers/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for @ember/test-helpers 0.7
+// Type definitions for @ember/test-helpers 0.8
```
jamescdavis: This is actually referring to the `major.minor` version of `@ember/test-helpers` that these definitions are appropriate for (and there is no `@ember/test-helpers@0.8`).

See: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library

This should probably be updated to the latest stable version of `@ember/test-helpers` (`1.5` as of right now).

types/ember__test-helpers/ember__test-helpers-tests.ts
```diff
@@ -53,7 +53,11 @@ test('DOM interactions', async () => {
     await focus(messageElement);
     await blur(messageElement);
     await triggerEvent(messageElement, 'custom-event');
+    await triggerEvent(window, 'custom-event');
+    await triggerEvent(document, 'custom-event');
```
jamescdavis: IMHO, since this group is all testing passing an element, testing `window` and `document` should each be moved into a separate sections below:

```js
.
.
.
await fillIn(messageElement, 'content');
await typeIn(messageElement, 'content');

await triggerEvent(window, 'custom-event');
await triggerKeyEvent(window, 'keydown', 'Enter', { ctrlKey: true });

await triggerEvent(document, 'custom-event');
await triggerKeyEvent(document, 'keydown', 'Enter', { ctrlKey: true });
```

jamescdavis (changes requested):
Thanks for adding `Window` and `Document` tot he event trigger helpers, @gossi! See my notes about versioning and test structure. Also, are there any of the other helpers that should have `Window` and `Document` added as possible targets (since they all use `getElement()`)? Maybe it only makes sense for the event triggering helpers?

dfreeman:
> Also, are there any of the other helpers that should have `Window` and `Document` added as possible targets (since they all use `getElement()`)? Maybe it only makes sense for the event triggering helpers?

Maybe we should update the definition of `Target` here to match [the one upstream](https://github.com/emberjs/ember-test-helpers/blob/9331615660437783a69828640306ef9bdeb319aa/addon-test-support/%40ember/test-helpers/dom/-target.ts#L1) instead of updating each function signature individually?

types/ember__test-helpers/ember__test-helpers-tests.ts
gossi: Haha - I've seen this, there were two ways of doing it. The other I  had in mind was the one you suggested now =)

types/ember__test-helpers/index.d.ts
gossi: Ah, I see. I thought the types run their own versioning system as mostly all `@types/ember*` do not match their package versions.

gossi:
Thanks for your comments. Basically I've only seen this to happen for `triggerEvent()` (that's the one I was using) and while updating them, was doing it for `triggerKeyEvent()`, too. Any more to update?

types/ember__test-helpers/index.d.ts
mike-north: > This should probably be updated to the latest stable version of @ember/test-helpers (1.5 as of right now).

I disagree with this. It should be the _oldest version of the package that the types accurately describe_.

mike-north:
Do we have a clear signal that passing a `Window` or `Document` is a deliberate part of the `@ember/test-helpers` public API?

mike-north (changes requested):
I need to see the following before approving this change
- [ ] Clear indication that this change aligns with the intended public API of `@ember/test-helpers`
- [ ] A proper version bump. If these types still accurately describe `@ember/test-helpers` v0.7, no version bump in the `index.d.ts` header should be made.

types/ember__test-helpers/index.d.ts
jamescdavis: Ah, you're right @mike-north! The upstream type changes were introduced in https://github.com/emberjs/ember-test-helpers/commit/40d263a2c06ee515ffe8d1c6a93045a5c1422a5d as part of https://github.com/emberjs/ember-test-helpers/pull/463 and first included in `1.1.0`. It's possible that this type change is functionally accurate for earlier versions, though.

mike-north: aha, so `1.1` would be the correct version.

gossi:
> Do we have a clear signal that passing a `Window` or `Document` is a deliberate part of the `@ember/test-helpers` public API?

Yes and no. I think, the types in javadoc part are outdated in the [source](https://github.com/emberjs/ember-test-helpers/blob/9331615660437783a69828640306ef9bdeb319aa/addon-test-support/%40ember/test-helpers/dom/trigger-event.ts#L11) in comparison to what the implementation is doing. However, you can `triggerEvent()` on a `Target` (given just the DOM).

I do use it to test event triggers against `window` and it works. To me, it seems the types are not accurate here and most probably the javadoc, too.

I changed the version to `1.1` as suggested and also reorganized the test code.

jamescdavis:
@typescript-bot what do you mean doesn't exist on master?: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/ember__test-helpers

mike-north:
> @typescript-bot what do you mean doesn't exist on master?: /types/ember__test-helpers@master 

@andrewbranch

andrewbranch:
@mike-north DefinitelyTyped allows supporting multiple major versions of a package at a time via different folders ([example](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/d3)), so we use a compound key of name + major version to identify packages. Since the last commit bumped the major version from 0 to 1, the assumption was that this is a new set of typings. It doesn’t currently try to do anything smart with the git diff. I did write some code to allow a comparison against the previous major version with a disclaimer that we might be comparing apples and oranges, so I’m not sure why that code path wasn’t chosen.

gossi:
@andrewbranch like this? I copied over the existing ones to `v0/` folder

andrewbranch:
@gossi I thiiiink you only need to do that if you need to continue making new changes to v0 while separately maintaining a v1. /cc @sandersn?

sandersn:
@andrewbranch You are right. I would only create a v0/ directory if you want to keep maintaining v0 types while work happens on the v1 types at the same time. Otherwise you are just making work for yourself.

jamescdavis:
The only [breaking change from v0 to v1](https://github.com/emberjs/ember-test-helpers/blob/master/CHANGELOG.md#v100-2018-10-29) is dropping support for node 4 and old versions of ember-cli so v1 types should continue to work for v0 (they just *might* include some features not implemented in v0). IMHO, we don't need to maintain separate types for v0.

PR #36048, opened 2019-06-08
[@ember/object] update `addListener()` and `removeListener()`

gossi:
Copy over existing type definitions from ember itself.

PS. dtslint forced me not to use `Function` but extract it into a `Callback` type. Dunno if this is the right way to do it.

---

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/emberjs/ember.js/blob/v3.10.1/packages/%40ember/-internals/metal/lib/events.ts

### DIFF
--- a/types/ember__object/events.d.ts
+++ b/types/ember__object/events.d.ts
@@ -3,32 +3,26 @@ import { ObserverMethod } from '@ember/object/-private/types';
 /**
  * Add an event listener
  */
-export function addListener<Context, Target>(
-    obj: Context,
-    key: keyof Context,
-    target: Target,
-    method: ObserverMethod<Target, Context>,
+export function addListener(
+    obj: object,
+    eventName: string,
+    target: object | Function | null,
+    method?: Function | string,
     once?: boolean
-): void;
-export function addListener<Context>(
-    obj: Context,
-    key: keyof Context,
-    method: ObserverMethod<Context, Context>
-): void;
+): void
+
 /**
  * Remove an event listener
+ *
+ * Arguments should match those passed to `addListener`.
  */
-export function removeListener<Context, Target>(
-    obj: Context,
-    key: keyof Context,
-    target: Target,
-    method: ObserverMethod<Target, Context>
-): any;
-export function removeListener<Context>(
-    obj: Context,
-    key: keyof Context,
-    method: ObserverMethod<Context, Context>
-): any;
+export function removeListener(
+    obj: object,
+    eventName: string,
+    target: object | null,
+    method?: Function | string
+): void
+
 /**
  * Send an event. The execution of suspended listeners
  * is skipped, and once listeners are removed. A listener without
--- a/types/ember__object/events.d.ts
+++ b/types/ember__object/events.d.ts
@@ -1,15 +1,17 @@
 import { ObserverMethod } from '@ember/object/-private/types';
 
+type Callback = () => void;
+
 /**
  * Add an event listener
  */
 export function addListener(
     obj: object,
     eventName: string,
-    target: object | Function | null,
-    method?: Function | string,
+    target: object | Callback | null,
+    method?: Callback | string,
     once?: boolean
-): void
+): void;
 
 /**
  * Remove an event listener
@@ -20,8 +22,8 @@ export function removeListener(
     obj: object,
     eventName: string,
     target: object | null,
-    method?: Function | string
-): void
+    method?: Callback | string
+): void;
 
 /**
  * Send an event. The execution of suspended listeners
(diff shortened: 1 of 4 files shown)


jednano (approved):

mike-north (changes requested):
Leaving additional comments in a moment. Blocking now so @jedmao's approval doesn't result in an auto-publish.

types/ember__object/events.d.ts
```diff
@@ -1,34 +1,32 @@
 import { ObserverMethod } from '@ember/object/-private/types';
 
+export {};
```
mike-north: This package doesn't export an empty object in reality, so we can't type it as such. Please revert this.

types/ember__object/events.d.ts
```diff
@@ -1,34 +1,32 @@
...
+
+type Callback = () => void;
```
mike-north: 1) 

Some versions of typescript have problems when you use a non-exported type in an exported type/signature. You'll want this alias to live in a package that's unlikely to collide with anything the `@ember/object` package will contain. 

Look at the other `@types/ember__*` modules for examples. We usually put things like this in folders like `./-private/types.d.ts` or something.


2) 

If you want to create a type alias, the proper one that's equivalent to `Function` is 

```ts
(...args: any[]) => any
```

types/ember__object/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for non-npm package @ember/object 3.1
+// Type definitions for non-npm package @ember/object 3.2
```
mike-north: No need to bump this. Please revert this change.

mike-north (changes requested):

types/ember__object/events.d.ts
gossi: I did this, because dtslint wanted me to do that :/ Based on your next point 👇

types/ember__object/events.d.ts
gossi: I was using `Function` but dtslint was against me, so I created an alias type. And because I had an alias type it forced me to have this `export {}`. Very weird. The code I pushed now is what I had first, before dtslint.

gossi:
Thanks for your comments. Pushed new code, please have a look.

gossi:
I kinda expected CI to fail now. Can you tell how to make it happy?

types/ember__object/events.d.ts
```diff
@@ -1,34 +1,28 @@
...
 
 /**
  * Add an event listener
  */
-export function addListener<Context, Target>(
-    obj: Context,
-    key: keyof Context,
-    target: Target,
-    method: ObserverMethod<Target, Context>,
+export function addListener(
```
mike-north: the removal of these type params is a backwards-incompatible change that will break anyone who has things like

```ts
addListener<C,T>(...
```
in their app. We'd want to think very carefully before doing something like this

types/ember__object/events.d.ts
```diff
@@ -1,34 +1,28 @@
...
-export function addListener<Context>(
-    obj: Context,
-    key: keyof Context,
-    method: ObserverMethod<Context, Context>
-): void;
+
 /**
  * Remove an event listener
+ *
+ * Arguments should match those passed to `addListener`.
```
mike-north: this JSDoc stuff shows up in tooltips as users hover over various symbols and navigate within intellisense autocompletes. I tend to find that it's a better experience to find complete information (as opposed to a reference to another function's information) in these tooltips. 

What are your thoughts around the way to provide the best developer experience?

mike-north (changes requested):
You have provided a description at the top of this PR:
> Copy over existing type definitions from ember itself.

The type definitions from the [emberjs/ember.js](https://github.com/emberjs/ember.js/) codebase are not an accurate representation of the framework's public API. We recognize the ember API documentation, and public statements from the framework core team as the "source of truth" for what these types should align with.

All of this being said, I don't really understand what your code change is intended to add or fix. 

Could you explain a little bit more about your intended outcome in making this change? What problem are you trying to solve?

gossi:
@mike-north I was actually adding/correcting the types to what they are _now_ (given this PR is already a little older).

in essence:

```ts
const handler = () => {
  //...
};

addListener(obj, 'foo', handler);
removeListener(obj, 'foo', handler);

// somewhere else
sendEvent(obj, 'foo');
```

which is what the official API docs say: https://api.emberjs.com/ember/3.15/functions/@ember%2Fobject%2Fevents/addListener

what apparently the current types will not be happy with. Perhaps needs some versioning then.

PR #36049, opened 2019-06-08
@types/react-test-renderer fix: children type of JSON output

michelle-laurenti:
ReactTestRendererJSON interface did not match the flow type. 

https://github.com/facebook/react/blob/v16.0.0/src/renderers/testing/ReactTestRendererFiberEntry.js#L39

The `children` attribute is now an array of `ReactTestRendererJSON | string` or null.


jgoz (approved):

rbuckton:
Thanks for the contribution!

PR #36054, opened 2019-06-08
async: fix: return promise if no callback for map

bertyhell:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: http://caolan.github.io/async/v3/docs.html#mapLimit

```
Wherever we accept a Node-style async function, we also directly accept an ES2017 async function. 
In this case, the async function will not be passed a final callback argument, 
and any thrown error will be used as the err argument of the implicit callback, 
and the return value will be used as the result value. 
(i.e. a rejected of the returned Promise becomes the err callback argument, 
and a resolved value becomes the result.)
```

### DIFF
--- a/types/async/index.d.ts
+++ b/types/async/index.d.ts
@@ -123,8 +123,11 @@ export const eachOf: typeof forEachOf;
 export const eachOfSeries: typeof forEachOf;
 export const eachOfLimit: typeof forEachOfLimit;
 export function map<T, R, E = Error>(arr: T[] | IterableIterator<T> | Dictionary<T>, iterator: AsyncResultIterator<T, R, E>, callback?: AsyncResultArrayCallback<R, E>): void;
-export const mapSeries: typeof map;
-export function mapLimit<T, R, E = Error>(arr: IterableCollection<T>, limit: number, iterator: AsyncResultIterator<T, R, E>, callback?: AsyncResultArrayCallback<R, E>): void;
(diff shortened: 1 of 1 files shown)


erossignon:
may duplicate #35825

types/async/index.d.ts
```diff
@@ -123,18 +123,28 @@ export const eachOf: typeof forEachOf;
 export const eachOfSeries: typeof forEachOf;
 export const eachOfLimit: typeof forEachOfLimit;
 export function map<T, R, E = Error>(arr: T[] | IterableIterator<T> | Dictionary<T>, iterator: AsyncResultIterator<T, R, E>, callback?: AsyncResultArrayCallback<R, E>): void;
-export const mapSeries: typeof map;
```
DanielRosenwasser: These actually had the same signature, so why couldn't you just add an overload to `map` and keep using `typeof map`?

bertyhell: I didn't know if would work, since map has 2 overloads now
So i thought the typeof wouldn't know which one to take, so i played it safe and duplicated the the type with both overloads.

DanielRosenwasser: Yeah, it will work. Just give it a shot, since I think the two were meant to stay in sync.

uniqueiniquity (changes requested):
Marking as changes requested until @DanielRosenwasser's comment is addressed.

PR #36063, opened 2019-06-09
Update types for kafkajs from official repository

mahulst:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/tulios/kafkajs/blob/master/index.d.ts

### DIFF
--- a/types/kafkajs/index.d.ts
+++ b/types/kafkajs/index.d.ts
@@ -7,111 +7,134 @@
 /// <reference types="node" />
 
 import * as tls from "tls";
+import * as net from "net";
 
 export class Kafka {
-    constructor(options: KafkaOptions);
-
-    producer(options?: ProducerOptions): Producer;
-    consumer(options?: ConsumerOptions): Consumer;
-    admin(options?: AdminOptions): Admin;
+    constructor(config: KafkaConfig);
+    producer(config?: ProducerConfig): Producer;
+    consumer(config?: ConsumerConfig): Consumer;
+    admin(config?: AdminConfig): Admin;
+    logger(): Logger;
 }
 
-export const PartitionAssigners: {
-    roundRobin: PartitionAssigner;
-};
-
-export namespace AssignerProtocol {
-    interface MemberMetadataOptions {
-        version: number;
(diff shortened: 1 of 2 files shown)


michal-b-kaminski:
I think you should delete the DefinitelyTyped package, not update it - at least according to [readme](https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package).

mahulst:
It is not yet released on npm with types, they will be included in 1.9.0. These are the typings for 1.8.0, and they are more precise, as they are from the official package maintainers.

michal-b-kaminski (approved):

rbuckton:
Thanks for the contribution!

dolegi:
Hello, this seems to be incompatible with the latest version of kafkajs (1.8.0).
Specifically we have issues with the change of type of `KafkaOptions.ssl` which was `tls.ConnectionOptions`, but is now `KafkaConfig.ssl` and `tls.SecureContextOptions`, this stops us being able to pass the `rejectUnauthorized` option in. It also looks like a few other things have changed too.

Downgrading to `@types/kafkajs@1.8.0` works fine with `kafkajs@1.8.0`

Maybe I'm missing something or doing something wrong, but the default behaviour of
`npm install kafkajs && npm install --save-dev @types/kafkjs`
Installs kafkajs@1.8.0 and @types/kafkajs@1.8.1 which are incompatible

michal-b-kaminski:
@mahulst Besides `KafkaConfig.ssl` mentioned above, I also see following problems:
- in `KafkaAdmin` - your typings have `fetchTopicMetadata` but in `kafkajs` there is `getTopicMetadata`
- in `ProducerBatch` - `acks`, `timeout` and `compression` are required event tough they have default values

Where there any changes in new typings that you need and are compatible with the library? Otherwise I think we have to restore previous version.

rbuckton:
It sounds like these changes caused some significant breaks, so I will revert this PR.

rbuckton:
This PR has been reverted. @mahulst: If you'd like to review your typings and file a new PR, please do so.

Issue #31914, opened 2019-01-06
[@types/react] - Feature Request - Use DeepReadonly so Components can't mutate their parameters


carpben:
One of the principles of React is that components are supposed to be pure - they shouldn't mutate the parameters they receive. If they do mutate parameters they receive they can mutate the state of a parent, or the state of the app.

We can now statically check mutation of objects by defining their type as DeepReadonly ([source1](https://stackoverflow.com/questions/41879327/deepreadonly-object-typescript), [2](https://www.npmjs.com/package/utility-types), [3](https://t.co/LnQjZHW0A9)). This could further be used for functions, by defining their parameters as DeepReadonly. I've started using it in my code and it works quite well. I get an error in the editor once I try to change a property of a parameter defined this way. It would be great if we can add this feature to the type of components (React.Compo ent and React.SFC). This can help keep state immutable, and components pure.

I'm not sure whom of the contributors I should mention, so I'll just mention every thrid contributor.   @johnnyreilly @digiguru @DovydasNavickas @guilhermehubner @pascaloliv @Kovensky

carpben:
@bbenezech @pzavolinsky @ericanderson @tkrotoff @onigoetz @theruther4d @ferdaber @jrakotoharisoa @hotell @franklixuefei @Kovensky @pshrmn 
I haven't received any reply yet. Could you give it a thought? Many use 3rd libraries such as Immutable and Immer and special syntax to avoid mutation. I think DeepReadonly provides a better approach to achieve immutability in React. We don't need to replace the type of Component, but maybe add another type, which wraps state and props in DeepReadonly. If you think this is necessary I'd be happy to prepare a PR.

Jessidhia:
Oh, right, I need to update my username in the comments too.

*ahem* I should take a look tomorrow. I am totally in favor of deep
immutability of props and state but it won't actually do what we want until
TS3.4 (making Readonly<Array> actually become ReadonlyArray)

ferdaber:
We already type the `Component` interface with `Readonly<P>` and `Readonly<S>` as their `props` and `state` instance properties, respectively, so it should be very easy to just change them to `DeepReadonly` instead.

carpben:
> but it won't actually do what we want until TS3.4 (making Readonly<Array> actually become ReadonlyArray)
@Jessidhia , My experience is that if you use a properly defined DeepReadonly generic, mutating an array is not possible. 

[...]

PR #36066, opened 2019-06-09
[@types/react] New component types - typing props and state as DeepReadonly

carpben:
Please fill in this template.

- [x ] Test the change in your own code. (Compile and run.)
- [x ] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts>>

If removing a declaration:

Following[ feature suggestion 31914](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/31914).
Currently this commit adds:
1. a new functional component type DRFC which types props as DeepReadonly.
2. A new DRPureComponent which extends the class component. I think that wrapping state and props in DeepReadonly goes along with using a pure component. Using immutable data guarantees that if a reference hasn't changed, none of its properties and deep properties have changed.
```ts
class DRPureComponent<P = {}, S = {}, SS = any> extends React.PureComponent<DR<P>, DR<S>, SS> { }
```
3. An adjustment to `JSXElementConstructor` to handle the new component types. 

It's important to note that if we would type state or props as `DeepReadonly` on an existing component type, this would be a breaking change, since you can't pass a `ReadonlyArray` to a function that expects an oridnary `Array` (which lacks methods such as `.push`). Therefor it is necessary to offer a new type, and not to replace a current one.


johnnyreilly:
Yo @andrewbranch!

What do all the numbers mean? They're definitely numbers... But it's hard to know what they represent ... And what response they're designed to provoke 😁

andrewbranch:
Hiya @johnnyreilly, fancy running into you here!

These particular numbers shouldn’t provoke much response at all, because they didn’t change much. The short explanation is that the “batch compilation” section contains some diagnostics about what happens memory-wise when running `tsc` on the tests + typings. We want to look out for huge increases here, particularly in memory usage / type count (memory usage isn’t displayed here because it wasn’t being reliably measured when the benchmark against `master` was recorded, but it’s fairly highly correlated with type count). The “language service” section summarizes the results of pretending the test files are in VS Code and someone is typing around and hovering identifiers to get information about them. We obviously want response times for those kinds of interactions to be as fast as possible.

We’re not currently running on dedicated hardware, so what I’ve seen so far is (speaking anecdotally) for response times, a fluctuation of around 10–15% could easily be environmental/measurement error. If any of the metrics move _a lot_, @typescript-bot will bold them, annotate them with an emoji, and gently suggest that the PR might have a performance impact that you need to consider. (Or, if the response times go uniformly and significantly down, you get a 🌟 and a message of congratulations.)

Since that didn’t happen here, you can safely ignore the metrics, or better yet, be reasonably confident that this PR will not destroy performance for folks consuming React typings.

johnnyreilly:
> Hiya @johnnyreilly, fancy running into you here!

Ha! I know, I get around 😄 (I'm one of the original DTers. My particular claim to fame was JSDOCing jQuery back in the day - so you get that lovely intellisense)

> Also note that on my to-do list is including a detailed explanation of each metric and some overall guidance on how to interpret them

I think this is a great idea. I think including some kind of examples of good / bad practice and performance gotchas here might help. I'd say this is generally not a well known / understood area. (Certainly by me) Performance of TypeScript itself when it comes to type definitions I'd suggest is not generally considered. Looks like you're hoping to fix that!

carpben:
@johnnyreilly, Any thoughts about this PR? 
I think it's an important PR, since using the new components types prevents the mutation of state or props (so it prevents bugs) without using another library (such as ImmutableJs) or relaying on the programmers caution.

types/react/index.d.ts
```diff
@@ -3007,4 +3013,33 @@ declare global {
             view: React.SVGProps<SVGViewElement>;
         }
     }
+
+    type DeepReadonly<T> =
```
ferdaber: For anything that's not public API, can you move these outside the `React` namespace? So that it's inaccessible to consumers.

types/react/index.d.ts
```diff
@@ -509,6 +513,8 @@ declare namespace React {
         displayName?: string;
     }
 
+    type DRFC<T= {}> = FC<DeepReadonly<T>>;
```
ferdaber:
```suggestion
    type DRFC<P= {}> = FC<DeepReadonly<P>>;
```

types/react/index.d.ts
```diff
@@ -74,7 +74,9 @@ declare namespace React {
 
     type JSXElementConstructor<P> =
         | ((props: P) => ReactElement | null)
-        | (new (props: P) => Component<P, any>);
+        | (new (props: P) => Component<P, any>)
+        | DRFC<P>
+        | (new (props: P) => DRPureComponent<P, any>);
```
ferdaber:
```suggestion
        | (new (props: DeepReadonly<P>) => DRPureComponent<P, any>);
```

Why does this only produce pure components? What about regular components?

types/react/index.d.ts
carpben: Sure. I haven't noticed it's inside `declare Global`. I'll move it to be above it.

types/react/index.d.ts
carpben: DRPureComponent wraps P in DeepReadonly. I guess the benefit of your suggestions is that it prevents the constructor function itself from mutating P - right? 
As for only adding DRPureComponent, I'm not sure about it any more. As I wrote yesterday (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36066) immutable data goes together with `PureComponent`. If the data is immutable we can trust that if the references in props hasn't changed, then the object hasn't changed, there is no reason to rerender the component, and it makes sense to use a PureComponent instead. 
However, the fact that our component won't mutate Props doesn't mean that no one will mutate Props, as it could be mutated before it reaches our component. The new component types make it more compelling to use a DRPureComponent, but you are right there probably is a need for a new DRComponent that just extends the "regular" Component.

ferdaber: Yeah I was mostly just wondering what the scope of this change would be. Theoretically no props should be mutated so we could change nearly everything to be deep-readonly, so I was just trying to get a sense of that scope.

carpben: Following this discussion I realized that Props isn't protected as DeepReadonly in the constructor, and could be mutated there. Can we remedy this?

ferdaber: It *should* be okay if you change the constructor of the `Component` class to have a deep-readonly props, but I'm not 100% sure as deep-readonly types are subtypes of non-readonly, and some weird assignability checks might fail unrelatedly.

carpben: Following this discussion I added a new type DRComponent. 
As for your suggestion, since every type should extend it's equivalent DeepReadonly type, your suggestion seems valid.

ferdaber (approved):

carpben:
This new component types still have two week spots. 
1. With the currently suggested DRComponent it is possible to mutate props in the construcotr. This is not a common pattern any more, as React move more towards functional components, and since there is no need to define a constructor for using state, or bind methods. There are two alternative solutions to this scenario: 

- Wrap props type in DR where Props is defined. for example: 
```ts
// Instead of 
interface IProps {
   people: IPeople
}
// Use
type Props = DeepReadonly<{
   people: IPeople
}>
```
- Use a guard property in DRComponent's definition. 
```
/* DeepReadonly Class Component Props */
type DRCP<P> = DR<P> & { __DRCP_GUARD: ["💀 You should define the constructor parameter as DeepReadonly<", P, "> 💀"]}
class DRComponent<P = {}, S = {}, SS = any> extends Component<DRCP<P>  , DR<S>, SS> {}
```
This will enforce that when a constructor is defined, props in the constructor will be typed as DeepReadonly. It will however require to slightly update `LibraryManagedAttributes` (just ommiting the guard property):        
```ts
type LibraryManagedAttributes<C, P> = C extends React.MemoExoticComponent<infer T> | React.LazyExoticComponent<infer T>
            ? T extends React.MemoExoticComponent<infer U> | React.LazyExoticComponent<infer U>
                ? ReactManagedAttributes<U, Omit<P, "__gaurd">>
                : ReactManagedAttributes<T, Omit<P, "__gaurd">>
            : ReactManagedAttributes<C, Omit<P, "__gaurd">>;
```
This approach was suggested to me by Titian Cernicova-Dragomir. Before submitting a PR to solve this issue, I'd like to get some feedback/opinion about this approach. 
[Here is a CodeSnadbox demo](https://codesandbox.io/s/friendly-moore-upsq6?fontsize=14), without changing LibraryManagedAttributes. 

2. the current useState hook doesn't protect state from mutation (see demo). If we would change the definition of useState and wrap the returned State in DeepReadonly it would  be a breaking change. A possible solution is to create a new variable drUseState to set it to useState, and to cast it so the state it returns is protected as DeepReadonly. 
```ts
const drUseState = useState as <S>(initialState: DR<S> | (() => DR<S>)) => [DR<S>, Dispatch<SetStateAction<S>>]
```
Check it out in the demo. 
Is it possible in DefenitelyTyped modules, or is it out of scope?

PR #36076, opened 2019-06-10
Add rsocket types

adrianhopebailie:
> Currently failing, see error below...


adrianhopebailie:
Need to still add some tests but getting an irritating compile error using `Partial<T>` and hoping for some wisdom of the crowd to resolve.

Error:
```
error TS2416: Property 'map' in type 'FlowableProcessor<T, R>' is not assignable to the same property in base type 'IPublisher<R>'.
  Type '<S>(fn: (a: R) => S) => this' is not assignable to type '<R>(fn: (data: R) => R) => IPublisher<R>'.
    Type 'this' is not assignable to type 'IPublisher<R>'.
      Type 'FlowableProcessor<T, R>' is not assignable to type 'IPublisher<R>'.
        Types of property 'subscribe' are incompatible.
          Type '(subscriber?: Partial<ISubscriber<R>> | undefined) => void' is not assignable to type '(subscriber?: Partial<ISubscriber<R>> | undefined) => void'. Two different types with this name exist, but they are unrelated.
            Types of parameters 'subscriber' and 'subscriber' are incompatible.
              Type 'Partial<ISubscriber<R>> | undefined' is not assignable to type 'Partial<ISubscriber<R>> | undefined'. Two different types with this name exist, but they are unrelated.
                Type 'Partial<ISubscriber<R>>' is not assignable to type 'Partial<ISubscriber<R>>'. Two different types with this name exist, but they are unrelated.
                  Types of property 'onNext' are incompatible.
                    Type '((value: R) => void) | undefined' is not assignable to type '((value: R) => void) | undefined'. Two different types with this name exist, but they are unrelated.
                      Type '(value: R) => void' is not assignable to type '(value: R) => void'. Two different types with this name exist, but they are unrelated.
                        Types of parameters 'value' and 'value' are incompatible.
                          Type 'R' is not assignable to type 'R'. Two different types with this name exist, but they are unrelated.
                            'R' is assignable to the constraint of type 'R', but 'R' could be instantiated with a different subtype of constraint '{}'.
```

Also getting lots of these but linting rules don't allow anything in `types` in `tsconfig.json`.

```
../rsocket-types/ReactiveSocketTypes.d.ts(139,42): error TS2591: Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i @types/node` and then add `node` to the types field in your tsconfig.
```

types/rsocket-core/RSocketMachine.d.ts
```diff
@@ -0,0 +1,165 @@
...
+  FRAME_TYPES,
+  MAX_REQUEST_N,
+  MAX_STREAM_ID,
+} from './RSocketFrame';
+import { PayloadSerializers, IdentitySerializers } from './RSocketSerialization';
+
+export type Role = 'CLIENT' | 'SERVER';
+
+export class ResponderWrapper<D, M> implements Responder<D, M> {
+  _responder: PartialResponder<D, M>;
```
rbuckton: A `_` prefix often indicates a "private" or "protected" member. Is this intended to be public?

types/rsocket-core/RSocketResumableTransport.d.ts
```diff
@@ -0,0 +1,121 @@
...
+ * that:
+ * - any frames the server has received are removed from _sentFrames
+ * - the remaining frames are merged (in correct order) into _pendingFrames
+ *
+ * Then the connection proceeds as above, where all pending frames are flushed.
+ * If anything other than RESUME_OK is received, resumption is considered to
+ * have failed and the connection is set to the ERROR status.
+ */
+export default class RSocketResumableTransport implements DuplexConnection {
+  _bufferSize: number;
```
rbuckton: Same with these, they seem like they should be `private`.

types/rsocket-core/RSocketMachine.d.ts
adrianhopebailie: No, I wasn't aware that the typings should only be for the public members. Will review all the definitions and revert.

adrianhopebailie:
@rbuckton - think this is ready for another review although I see that the bot didn't remove the CI fail label

PR #36080, opened 2019-06-10
[react] Only allow thunks in useState when setting state that is a function

lemonmade:
This PR updates the typing for `useState` from React. Currently, the following is considered valid during type check:

```ts
const [state, setState] = useState<() => boolean | null>(null);
setState(() => false);
```

However, this is actually incorrect at runtime, because React sees the function form of the state setter and calls it. So, after this call, `state` will be `boolean`, which it should never be given the type definition. Yes, it's an unusual case (and probably an anti pattern in most cases), but saving a function in state does have some valid use cases.

Please fill in this template.

- ~~[ ] Provide a URL to documentation or source code which provides context for the suggested changes:~~ this change is just reflective of the reality of `useState` when a function is provided.
- ~~[ ] Increase the version number in the header if appropriate.~~
- ~~[ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.~~

### DIFF
--- a/types/react/test/hooks.tsx
+++ b/types/react/test/hooks.tsx
@@ -201,6 +201,14 @@ function useEveryHook(ref: React.Ref<{ id: number }>|undefined): () => boolean {
     // make sure the generic argument does reject actual potentially undefined inputs
     // $ExpectError
     React.useState<number>(undefined)[0];
+    // When state is a function type, it must be returned from a function,
+    // not provided directly.
+    // $ExpectError
+    React.useState<(() => boolean)>(() => true)[1](() => false);
+    // Returning a function for state is fine
+    React.useState<(() => boolean)>(() => true)[1](() => () => false);
+    // As is returning non-function members of a union
+    React.useState<(() => boolean) | number>(() => true)[1](() => 42);
 
     // useReducer convenience overload
 
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -805,7 +805,7 @@ declare namespace React {
     // based on the code in https://github.com/facebook/react/pull/13968
 
     // Unlike the class component setState, the updates are not allowed to be partial
-    type SetStateAction<S> = S | ((prevState: S) => S);
+    type SetStateAction<S> = S extends ((...args: any[]) => any) ? ((prevState: S) => S) : S | ((prevState: S) => S);
     // this technically does accept a second argument, but it's already under a deprecation warning
     // and it's not even released so probably better to not define it.
     type Dispatch<A> = (value: A) => void;
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -806,7 +806,9 @@ declare namespace React {
     // based on the code in https://github.com/facebook/react/pull/13968
 
     // Unlike the class component setState, the updates are not allowed to be partial
-    type SetStateAction<S> = S extends ((...args: any[]) => any) ? ((prevState: S) => S) : S | ((prevState: S) => S);
+    type SetStateAction<S> = S extends ((...args: never[]) => unknown) ?
+        (prevState: S) => S :
+        (S | ((prevState: S) => S));
     // this technically does accept a second argument, but it's already under a deprecation warning
     // and it's not even released so probably better to not define it.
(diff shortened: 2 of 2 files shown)


ferdaber:
Functions aren't allowed to be state because React can't tell the difference between the dispatch form and the non-dispatch form when setting, so ideally shouldn't we just change the type param to extend non function types?

lemonmade:
Is there some document that explicitly notes this to be disallowed? It works fine in practice once you compensate for the "must be a thunk" bit, and you can certainly think up a number of cases where you legitimately have a function that is stateful (my case: async loading a component, and needing to force update the component doing the loading once it has been fetched).

ferdaber:
Oh, I see what you mean; AFAIK there is no documentation but that's just what I commonly hear. So the workaround would be to always use the dispatch form when setting the state. The way people usually work around this issue is to encapsulate it inside an object:

```ts
const [{ myMethod }, setMethods] = useState<{ myMethod(): void }>({ myMethod(): {} })
/*...*/
setMethods({ myMethod() { /* some other impl */ }})
```

Either way, I'm fine with this change, I just got confused by the wording of the PR, sorry!

ferdaber (approved):

types/react/test/hooks.tsx
```diff
@@ -201,6 +201,14 @@ function useEveryHook(ref: React.Ref<{ id: number }>|undefined): () => boolean {
     // make sure the generic argument does reject actual potentially undefined inputs
     // $ExpectError
     React.useState<number>(undefined)[0];
+    // When state is a function type, it must be returned from a function,
+    // not provided directly.
+    // $ExpectError
+    React.useState<(() => boolean)>(() => true)[1](() => false);
+    // Returning a function for state is fine
+    React.useState<(() => boolean)>(() => true)[1](() => () => false);
```
Jessidhia: The `() => true` initializer will create a `boolean` state (`useState` will treat function arguments as lazy initializers instead of values)

Jessidhia (commented):
I didn't see this in time 😢

I am good with this change, the only note is that it missed dealing with the initializers.

Fonger:
This used to work. Now it breaks
```ts
const [showPwd, setShowPwd] = useState<boolean>(false);
<button onClick={() => setShowPwd(ps => !ps)}>show/hide</button>
/* works well @types/react@16.8.20 */
/* Parameter 'ps' implicitly has an 'any' type.ts(7006) @types/react@16.8.21 */
```

MayhemYDG:
Same here, on TS 3.5.2:
```ts
setBool(bool => !bool);
```
```
src/hooks.ts:31:11 - error TS7006: Parameter 'bool' implicitly has an 'any' type.

31   setBool(bool => !bool);
             ~~~~

src/hooks.ts:31:11 - error TS2345: Argument of type '(bool: any) => boolean' is not assignable to parameter of type 'boolean | ((prevState: false) => false) | ((prevState: true) => true)'.
  Type '(bool: any) => boolean' is not assignable to type '(prevState: false) => false'.
    Type 'boolean' is not assignable to type 'false'.

31   setBool(bool => !bool);
             ~~~~~~~~~~~~~
```
---
```ts
setBool((bool: boolean) => !bool);
```
```
src/hooks.ts:31:11 - error TS2345: Argument of type '(bool: boolean) => boolean' is not assignable to parameter of type 'boolean | ((prevState: false) => false) | ((prevState: true) => true)'.
  Type '(bool: boolean) => boolean' is not assignable to type '(prevState: false) => false'.
    Type 'boolean' is not assignable to type 'false'.

31   setBool((bool: boolean) => !bool);
             ~~~~~~~~~~~~~~~~~~~~~~~~
```

joealden:
@Jessidhia do you know if a fix for this in the works?

This regression is blocking me from updating to `16.8.21`, and I'm sure it is for many others.

Tagging @lemonmade  @ferdaber and @DanielRosenwasser for visibility.

ferdaber:
I can create a revert PR later today while we look into the issue.

ferdaber:
For now please go back to 16.8.20.

PR #36087, opened 2019-06-10
Add new type to support arearange graph

dolanmiu:
Area Range needs 3 inputs for it to `addPoint`:

http://jsfiddle.net/k8EKj/

Please fill in this template.

### DIFF
--- a/types/highcharts/index.d.ts
+++ b/types/highcharts/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for Highcharts 7.1
+// Type definitions for Highcharts 5.0
 // Project: http://www.highcharts.com/
 // Definitions by: Damiano Gambarotto <https://github.com/damianog>
 //                 Dan Lewi Harkestad <https://github.com/baltie>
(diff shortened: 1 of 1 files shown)


jgresham (approved):

types/highcharts/index.d.ts
```diff
@@ -1,11 +1,12 @@
-// Type definitions for Highcharts 5.0
+// Type definitions for Highcharts 7.1
```
rbuckton: When upgrading the major version, please consider the steps outlined here: https://github.com/DefinitelyTyped/DefinitelyTyped#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package

rbuckton (changes requested):

types/highcharts/index.d.ts
baltie:
```suggestion
// Type definitions for Highcharts 5.0
```

Highcharts are supplying their own TypeScript definitions for Highcharts 7 and onwards. Also, this pull request does not include anything to lift the type definitions to version 7 either, so it is misleading. I suggest you leave this comment as it was.

dolanmiu:
Done, I reverted the version back to 5.0

rbuckton (approved):

rbuckton:
Thanks for the contribution!

PR #36091, opened 2019-06-10
Adds type definitions for 'promise-nodeify'

jgeth:
Adds type definitions and tests
Adds tsconfig and tslint config

### DIFF
--- a/types/promise-nodeify/index.d.ts
+++ b/types/promise-nodeify/index.d.ts
@@ -17,7 +17,6 @@ declare const promiseNodeify: {
     ): Promise<T> | undefined;
 
     nodeifyThis<T>(
-        promise: Promise<T>,
         callback: (error: Error | undefined, value: T | undefined) => any
     ): Promise<T> | undefined;
 };
--- a/types/promise-nodeify/index.d.ts
+++ b/types/promise-nodeify/index.d.ts
@@ -1,0 +1,25 @@
+// Type definitions for promise-nodeify 3.0
+// Project: https://github.com/kevinoid/promise-nodeify
+// Definitions by: jgeth <https://github.com/jgeth>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export as namespace promiseNodeify;
+
+declare const promiseNodeify: {
+    <T>(
+        promise: Promise<T>,
+        callback: (error: Error | undefined, value: T | undefined) => any
+    ): Promise<T> | undefined;
+
+    delegated<T>(
+        promise: Promise<T>,
+        callback: (error: Error | undefined, value: T | undefined) => any
+    ): Promise<T> | undefined;
+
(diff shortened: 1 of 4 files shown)


types/promise-nodeify/index.d.ts
```diff
@@ -0,0 +1,25 @@
...
+        promise: Promise<T>,
+        callback: (error: Error | undefined, value: T | undefined) => any
+    ): Promise<T> | undefined;
+
+    delegated<T>(
+        promise: Promise<T>,
+        callback: (error: Error | undefined, value: T | undefined) => any
+    ): Promise<T> | undefined;
+
+    nodeifyThis<T>(
```
rbuckton: This doesn't seem correct. The usage in the documentation is this:
```ts
Promise.prototype.nodeify = require('promise-nodify').nodeifyThis;
...
promise.nodeify(callback);
```

I think the definition you want is:
```ts
nodeifyThis<T>(
  this: Promise<T>,
  callback: (error: Error | undefined, value: T | undefined) => any
): Promise<T> | undefined;
```

types/promise-nodeify/promise-nodeify-tests.ts
```diff
@@ -0,0 +1,8 @@
+import * as promiseNodeify from "promise-nodeify";
+
+const promise = new Promise<any>((resolve, reject) => {});
+const callback = () => {};
+
+promiseNodeify(promise, callback);
+promiseNodeify.delegated(promise, callback);
+promiseNodeify.nodeifyThis(promise, callback);
```
rbuckton: This does not match the documentation.

rbuckton (changes requested):

types/promise-nodeify/index.d.ts
jgeth: Good catch. It does use `this` but it's not in the method signature. Will update with correct signature.

types/promise-nodeify/promise-nodeify-tests.ts
jgeth: As above

jgeth:
@rbuckton - Can we get a review on this? 🙏🏻

types/promise-nodeify/index.d.ts
rbuckton: Still not correct. As I mentioned, `nodifyThis<T>` requires a `this`, so it would be better to change it to the following:
```suggestion
    nodeifyThis<T>(
      this: Promise<T>,
```

TypeScript uses a `this` parameter to denote the required `this` type for a function. Otherwise, it seems like `nodifyThis` can be called on its own with just a callback and produce a `Promise`, which is incorrect.

orta:
Cool, this looks good to me - sorry it took so long to get back to the PR, there's an awful lot of them and being on rotation effectively kills your inbox. Thanks for adding this module!

PR #36094, opened 2019-06-10
Emscripten: Add EmscriptenModule interface type, make Module an instance of it.

### LINKED ISSUE #24791: Support for MODULARIZE in emcripten definitions
Thanks for the convenient type definitions.  Emscripten now has a MODULARIZE (see https://github.com/kripken/emscripten/blob/incoming/src/settings.js#L564) setting that makes it produce better-behaved code.  With that setting, `Module` is an interface rather than a namespace.  Would it be easy to update the typescript definitions to support that?

- Authors: @zakki @periklis

wffurr:
Fixes #24791.  I plan to use this when sending code to Emscripten for [generating TypeScript typings from embind bindings](https://github.com/emscripten-core/emscripten/issues/7083).

Note that this changed some type names:

 * `Module.WebAssemblyImports` → `Emscripten.WebAssemblyImports`
 * `Module.WebAssemblyExports` → `Emscripten.WebAssemblyExports`

This was required because `Module` is now an instance of the `EmscriptenModule` interface, and can't host type aliases.

@cpitclaudel @periklis please review and comment if this meets your needs.

Please fill in this template.

### DIFF
--- a/types/emscripten/index.d.ts
+++ b/types/emscripten/index.d.ts
@@ -13,9 +13,6 @@ declare namespace WebAssembly {
 declare namespace Emscripten {
     interface FileSystemType {
     }
-}
-
-declare namespace Module {
     type EnvironmentType = "WEB" | "NODE" | "SHELL" | "WORKER";
 
     type WebAssemblyImports =  Array<{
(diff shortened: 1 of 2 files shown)


types/emscripten/index.d.ts
```diff
@@ -28,96 +25,104 @@ declare namespace Module {
...
-    function _malloc(size: number): number;
-    function _free(ptr: number): void;
+    _malloc(size: number): number;
+    _free(ptr: number): void;
 }
 
+// By default Emscripten emits a single global Module.  Users setting -s
+// MODULARIZE=1 -s EXPORT_NAME=MyMod should declare their own types, e.g.
+// declare var MyMod: EmscriptenModule;
+declare var Module: EmscriptenModule;
```
boyanio: Maybe we could omit this one, so that every person defines the correct module name. If one uses the default settings, Emscripten will indeed emit `Module`, but otherwise this will "pollute" the global types. What do you think?

wffurr: The default Emscripten configuration uses the global Module name.  It's unfortunate that it's so generic but I don't know of any specific conflicts with it.  @periklis in the linked issue (#24791) claims that the MODULARIZE option with other names is not much used.

Of the two use cases I have encountered at work where we have just imported this typings definition, one uses Module and the other sets a name with MODULARIZE.

wffurr: I'd prefer to keep the annotation in order to make using the typings definition simple for Emscripten users with the default set up.

uniqueiniquity:

wffurr:
Thank you Benjamin,

PR #36095, opened 2019-06-10
Update Ember tests to treat 'string & number' and 'never' the same

ahejlsberg:
In preparation for https://github.com/microsoft/TypeScript/pull/31838 this modifies Ember's tests to treat empty intersection types like `string & number` the same as `never`.


mike-north (approved):
Thanks for these updates! We'll replace these with some more meaningful tests as a follow-on PR

PR #36102, opened 2019-06-11
Add type definitions for d3-require

kindy:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/d3-require/index.d.ts
+++ b/types/d3-require/index.d.ts
@@ -1,0 +1,45 @@
+// Type definitions for d3-require 1.2
+// Project: https://github.com/d3/d3-require
+// Definitions by: Kindy Lin <https://github.com/kindy>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+// Last module version validated against: 1.2.2
+
+export interface require {
+    (name: string, ...names: string[]): Promise<any>;
+
+    /**
+     * Returns a `require` with the specified *aliases*.
+     * `require.alias({"react": React, "react-dom": ReactDOM})`
+     */
+    alias(aliases: {[k: string]: any}): require;
+
+    /**
+     * The resolver passed to `requireFrom`
+     */
+    resolve: resolver;
+}
+
+/**
+ * Returns a promise to the URL to load the module with the specified *name*.
+ */
+export type resolver = (name: string, base?: string) => Promise<string>;
+export type requireFrom = (resolver: resolver) => require;
+
+declare class RequireError extends Error {
+}
+
+/**
+ * Loads modules from jsDelivr.
+ * Like `require('d3@5')`
+ *
+ * @param name Any package name optionally followed by the 'at sign' and a semver range
+ * @returns A promise to the loaded module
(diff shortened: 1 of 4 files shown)


types/d3-require/index.d.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @param name Any package name optionally followed by the 'at sign' and a semver range
+ * @returns A promise to the loaded module
+ */
+export const require: require;
+
+/**
+ * Returns a new `require` which loads modules from the specified `resolver`.
+ * Like `requireFrom(name => ('https://unpkg.com/'+name))`
+ */
+export const requireFrom: requireFrom;
```
rbuckton: It looks like `d3-require` is a UMD module and can be used in scripts via the `d3` global. I would recommend adding the following at the bottom:

```ts
export as namespace d3
```

rbuckton (changes requested):

types/d3-require/index.d.ts
kindy: will add that

kindy: Should I add some tests for this namespace?
is this correct? Thanks. @rbuckton 

```diff
diff --git a/types/d3-require/d3-require-tests.ts b/types/d3-require/d3-require-tests.ts
index f77d0e2250..78e79158a9 100644
--- a/types/d3-require/d3-require-tests.ts
+++ b/types/d3-require/d3-require-tests.ts
@@ -1,4 +1,5 @@
 import { require as d3Require, requireFrom } from 'd3-require';
+import d3 = require('d3-require');
 
 d3Require('abc').then(x => null);
 
@@ -6,3 +7,5 @@ d3Require.alias({a: 'a'})('a', 'b').then(x => null);
 d3Require.resolve('b').then(x => x.length);
 
 requireFrom(async a => ('/~/' + a))('a', 'b').then(x => null);
+
+d3.require('abc@x');
```

rbuckton: My apologies. After further review we probably should *not* use `export as namespace d3;`. This package is intended to be used alongside the `d3` package which *also* has an `export as namespace d3`, and using both together would result in a conflict.

I'm not sure how best to resolve that conflict at the moment, so for the time being I would remove the `export as namespace d3` line and I'll go ahead and merge.

kindy: namespace removed.

types/d3-require/tslint.json
```diff
@@ -0,0 +1,7 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "unified-signatures": false,
+        "no-unnecessary-generics": false
```
DanielRosenwasser: Why do you need these?

kindy: @DanielRosenwasser  this **tslint.json** file is copy from other module, should I fix it?

DanielRosenwasser: Yes please

orta:
I've given this a look over, @kindy - does this look good to go now?

PR #36106, opened 2019-06-11
chore(node): add Readable.from utility

imcotton:
Please fill in this template.

If adding a new definition:

    - nodejs/node#27660
    - [Readable.from(iterable, [options])](https://nodejs.org/dist/latest-v12.x/docs/api/stream.html#stream_readable_from_iterable_options)

If removing a declaration:

### DIFF
--- a/types/node/test/stream.ts
+++ b/types/node/test/stream.ts
@@ -192,3 +192,17 @@ function stream_readable_pipe_test() {
     z.close();
     rs.close();
 }
+
+async function readable_from() {
+    const list = [ 1, 2, 3 ];
+    const asyncList = (async function *() { yield *list; }());
+
+    const readableSync = Readable.from(list);
+    const readableAsync = Readable.from(asyncList);
+
+    let i = 0;
+    readableSync.on('data', n => assert(n === list[i++]));
+
+    let j = 0;
+    readableAsync.on('data', n => assert(n === list[j++]));
+}
--- a/types/node/stream.d.ts
+++ b/types/node/stream.d.ts
@@ -18,6 +18,7 @@ declare module "stream" {
         }
 
         class Readable extends Stream implements NodeJS.ReadableStream {
+            static from(iterable: Iterable<any> | AsyncIterable<any>, opts?: ReadableOptions): NodeJS.ReadableStream;
             readable: boolean;
             readonly readableHighWaterMark: number;
             readonly readableLength: number;


ZaneHannanAU (approved):

islishude (commented):
Please add tests for it.

imcotton:
Added test, but there are two issues remain:

1. The `iterable` argument accept either `Iterable<any>` or `AsyncIterable<any>` type, but `AsyncIterable` seems requires TS **v3.2**, how to annotate the difference?

2. This PR affects `@types/readable-stream` which for https://www.npmjs.com/package/readable-stream, however its implementations pinned to Node **v10** only, should pin the Node version in `@types/readable-stream` first? How?

types/node/test/stream.ts
```diff
@@ -192,3 +192,17 @@ function stream_readable_pipe_test() {
     z.close();
     rs.close();
 }
+
+async function readable_from() {
+    const list = [ 1, 2, 3 ];
+    const listPromise = list.map(n => Promise.resolve(n));
```
Touffy: a iterable of promises is not an asyncIterable. You could generate your mock asyncIterable with a generator, like this:
```javascript
const asyncList = (async function *() {yield *[1, 2, 3]})()
```

imcotton: Thank you for the correction, fix amended.

Touffy (approved):

Flarna:
Regarding the CI errors:
`Cannot find name 'AsyncIterable'.` ==> can be solved by adding a forward declaration in index.d.ts
`'async' modifier cannot be used here.` ==> I think this can be only solved by moving the test (at least the async part) into ts3.2\node-tests.ts
The error in readable-stream could be fixed by adding `static from` method to `class _Readable.Duplex` or maybe my by changing `Duplex` to `extends _Readable implements Writable` similar as it is in node. eventually @TeamworkGuy2 has also some hints in this.

TeamworkGuy2:
Least change to `readable-stream` would be to just add the `static from()` method (currently the Duplex definition manually implements _Readable to work around the extends vs implements issue for node classes like Duplex that 'extend' two classes).
Note: adding `from` to `readable-stream` should be part of a major version bump of the `@types/readable-stream` definition from 2.x -> 3.x since this from() method is a new API of node v12 and readable-stream@~3.0.0 is aligned with node v10+ (?) I don't fully follow how the version numbers match up between node and readable-stream...

imcotton:
About updating `readable-stream` alone, I've had issue nodejs/readable-stream#411, as the result I think:

1. wait till PR nodejs/node#27660 backported to Node **v10**, then update `readable-stream` correspondingly, then it unblocks this PR.
1. find way to pin `@types/node@10` in `@types/readable-stream`, then it will also unblocks this PR.

_Thoughts? @Flarna @TeamworkGuy2_

TeamworkGuy2:
I haven't had to specify a node definition version in a dts file before, `readable-stream` uses `/// <reference types="node" />` because some of the classes/interfaces the definition references from node are ambient.
I know the [package.json](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/readable-stream/package.json) can be used to force a specific major version of an imported dependency (like [v10](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/node)) to be used, but I've only see that used for `import ...` not triple slash `/// <reference`'s, so I don't have much help to offer with that question.

Flarna:
using `/// <reference types="node" />` actually specifies latest node type definition which is node.js 12 at this time. I'm not sure if it is even possible to select an older node type definition as reference here. And even if it is possible there is some risk that this results in conflicts for end users as pinning a node version in one module could conflict with others.

@sandersn Do you know if it is possible for type definitons to reference specific, older version as dependencies? and if yes, is it a good idea to do this?

I think the best solution is to modify `Duplex` in `readable-stream` to `extends _Readable implements Writable` similar as it is in node. As `readable-stream` is actually a mirror of node streams I see no reason for such diff.

sandersn:
Yes, this is an accepted technique: for example, angularfire depends on `@types/firebase@^2.4.1`. I think it should work the same for triple-slash references. 

However, I'm not sure it's a good idea for `@types/node` since it's the most common dependency.
Which packages need the versioned dependency on `@types/node`? 

You have to add an entry to types-publisher/dependenciesWhitelist.txt in order to pass CI.

Flarna:
It's `readable-stream` which is a standalone version of node.js 10 streams.

I'm also not sure if it is a good idea. In special I would expect issues if some user has two versions of `@types/node` in it's dependencies.

PR #36111, opened 2019-06-11
Add imurmurhash

jimexist:


types/imurmurhash/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+
+interface MurmurHash3 {
+    result(): number;
+
+    reset(seed?: number): MurmurHash3;
+
+    hash(value: string): MurmurHash3;
+}
+
+declare function MurmurHash(text?: string, seed?: number): MurmurHash3;
```
DanielRosenwasser: Since this can be both constructed and called, is it appropriate to declare this as the following?

```ts
declare var MurmurHash: {
    (text?: string, seed?: number): MurmurHash3;
    new (text?: string, seed?: number): MurmurHash3;
}
```

PR #36117, opened 2019-06-11
Update types for react-autosuggest containerProps

jseminck:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: (see below)

I tested in two places and the current definitions seem incorrect.

CodePen (https://codepen.io/anon/pen/YozMEe):

Screen Shot 2019-06-11 at 17 34 02

Our own code: 

Screen Shot 2019-06-11 at 17 29 27

### DIFF
--- a/types/react-autosuggest/index.d.ts
+++ b/types/react-autosuggest/index.d.ts
@@ -93,8 +93,9 @@ declare namespace Autosuggest {
         containerProps: {
             id: string;
             key: string;
+            className: string;
             ref: any;
-            style: any;
+            role: string;
         };
         children: React.ReactNode;
         query: string;
--- a/types/react-autosuggest/index.d.ts
+++ b/types/react-autosuggest/index.d.ts
@@ -95,7 +95,7 @@ declare namespace Autosuggest {
             key: string;
             className: string;
             ref: any;
-            role: string;
+            role?: string;
         };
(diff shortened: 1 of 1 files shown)


types/react-autosuggest/index.d.ts
```diff
@@ -93,8 +93,9 @@ declare namespace Autosuggest {
         containerProps: {
             id: string;
             key: string;
+            className: string;
             ref: any;
-            style: any;
+            role?: string;
```
cdeutsch: I don't think `role` should be optional based on this (which is what is ultimately calling `renderSuggestionsContainer`
https://github.com/moroshko/react-autowhatever/blob/a74c882d4f63fabe7856e5d90b93c2af81c89a5b/src/Autowhatever.js#L342

cdeutsch (commented):
I see the same `containerProps` being set as your code (not the Pen)

image

So removing `style` and adding `className` looks correct.

I'm not sure why `role` is missing in your Pen.

I think whatever unpkg.com is doing is wrong.

When I change the settings to the following `role` is present.
image

I'd just make `role` non-optional, then should be good.

cdeutsch:
Looks good 👍

cdeutsch (approved):

Issue #36019, opened 2019-06-07
How to use sinon types?


nicojs:
Authors: @mrbigdog2u, @rationull, @lumaxis, @43081j, @joshuakgoldberg, @gjednaszewski, @johnjesse, @alecf, @SimonSchick, @bergundy

Hi fellow sinon type authors 👋 . Sorry for the inconvenient cry for attention (it always comes at a bad time, feel free to ignore).

I've updated to the latest version `@types/sinon` yesterday. I needed to add generic classes to my stubs like so:

![image](https://user-images.githubusercontent.com/1828233/59089103-99d38d00-8909-11e9-96d6-6b592e10a9f8.png)

My question is: am I using the sinon types correctly? I don't feel like these typings improved the maintainability of my code. As soon as a method changes its signature I will need to update the mocks in all classes. That doesn't feel helpful. Am I doing something wrong here? 

The problem is that with mocha we declare stubs before we initialize them (in `beforeEach`). Is there a way around it that I'm unaware of? Or is everyone typing all these generic types by hand?

Thanks in advance for any help.

43081j:
the aim isn't to let you write less, its to have stronger types. those stubs will now have the correct type information with all the assertion methods correctly typed (e.g. `returns` will only accept what you said the method returns). 

having to be a bit more verbose is the cost of declaring all your variables before they are assigned to, its not a problem and is making good use of the types still.

nicojs:
I've found a way around having to manually type the stubs using a small helper class:

```ts
    class TestHelper {
        existsSyncStub = sinon.stub(fs, 'existsSync');
        streamStub = {
            pipe: sinon.stub(),
            on: sinon.stub()
        };
        streamToPromiseStub = sinon.stub(Utils, 'streamToPromise');
        gotStreamStub = sinon.stub(got, 'stream');
        mkdirSyncStub = sinon.stub(fs, 'mkdirSync');
        statSyncStub = sinon.stub(fs, 'statSync');
        createWriteStream = sinon.stub(fs, 'createWriteStream');

        constructor() {
            this.gotStreamStub.returns(this.streamStub as any);
            this.streamStub.pipe.returns('expected piped value');
        }
    }
    let testHelper: TestHelper;

    beforeEach(() => {
        log = sinon.stub();
        testHelper = new TestHelper();
    });
```

Is this the way to go? Could we maybe document this somewhere so people use this way of writing code? Reaping the benefits from a short syntax and strong typing.

[...]

PR #36119, opened 2019-06-11
sinon: resolve any in worst case

43081j:

---

If we have an uninferrable return type by one of these cases (for example):

* Multiple overloads of a function or method
* `any`
* `unknown`

`resolves` will be typed as accepting `never`, making it completely unusable.

we should instead fallback to `any`, as "better typing when possible" is better than "better typing or nothing".

### DIFF
--- a/types/sinon/ts3.1/index.d.ts
+++ b/types/sinon/ts3.1/index.d.ts
@@ -404,7 +404,7 @@ declare namespace Sinon {
          * The Promise library can be overwritten using the usingPromise method.
          * Since sinon@2.0.0
          */
-        resolves(value?: TReturnValue extends PromiseLike<infer TResolveValue> ? TResolveValue : never): SinonStub<TArgs, TReturnValue>;
+        resolves(value?: TReturnValue extends PromiseLike<infer TResolveValue> ? TResolveValue : any): SinonStub<TArgs, TReturnValue>;
         /**
          * Causes the stub to return a Promise which resolves to the argument at the provided index.
          * stub.resolvesArg(0); causes the stub to return a Promise which resolves to the first argument.
--- a/types/sinon/ts3.1/sinon-tests.ts
+++ b/types/sinon/ts3.1/sinon-tests.ts
@@ -491,6 +491,7 @@ function testStub() {
         foo(arg: string): number { return 1; }
         promiseFunc() { return Promise.resolve('foo'); }
         promiseLikeFunc() { return Promise.resolve('foo') as PromiseLike<string>; }
+        unresolvableReturnFunc(): any { return Promise.resolve(); }
         fooDeep(arg: { s: string }): void { return undefined; }
     };
     const instance = new obj();
@@ -501,6 +502,10 @@ function testStub() {
 
     const promiseStub = sinon.stub(instance, 'promiseFunc');
     promiseStub.resolves('test');
+
+    const promiseUnresolvableReturn =
+        sinon.stub(instance, 'unresolvableReturnFunc');
+    promiseUnresolvableReturn.resolves(['anything', 123, true]);
 
     const promiseLikeStub = sinon.stub(instance, 'promiseLikeFunc');
     promiseLikeStub.resolves('test');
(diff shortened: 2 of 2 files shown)


types/sinon/ts3.1/index.d.ts
```diff
@@ -404,7 +404,7 @@ declare namespace Sinon {
          * The Promise library can be overwritten using the usingPromise method.
          * Since sinon@2.0.0
          */
-        resolves(value?: TReturnValue extends PromiseLike<infer TResolveValue> ? TResolveValue : never): SinonStub<TArgs, TReturnValue>;
+        resolves(value?: TReturnValue extends PromiseLike<infer TResolveValue> ? TResolveValue : any): SinonStub<TArgs, TReturnValue>;
```
alecf: but this now means that if you do pass in the wrong type, the whole thing falls back to any and you lose all strong typing.

i.e. if your stub returns `Promise<number>` and you do:

```typescript
myStub.resolves('hello');
```

Now the inference will fail, and this falls back to `'hello'` which means this is broken.

What happens if you resolve to `unknown` instead, then you could at least cast bad values:

```typescript
myStub.resolves('hello' as unknown);
```

alecf (commented):
I don't like this approach and I think the burden is on the consumer to do casting to get the correct type:

```typescript
myStub.resolves('hello' as unknown as number);
```

alecf:
see #31401 for the original change that made this `never`

Another approach: when you're declaring your stub, declare it with `Promise<any>`:

```typescript
const stub = sinon.stub<Promise<any>>();
```

43081j:
It really doesn't work like you are saying...

`resolves` has a single-use type on the way in, you're not losing any type information here.

```ts
declare fn: () => Promise<boolean>;
stub(fn).resolves(true); // `resolves(value?: boolean)` is inferred

declare fn: () => any;
stub(fn).resolves(true); // `resolves(value?: any)` is inferred

declare function foo(): Promise<boolean>;
declare function foo(num: number): void;
stub(foo).resolves(true); // `resolves(value?: any)` is inferred
```

If we fall back to `never` rather than `any`, we are effectively disallowing use of `resolves` where the type can't be inferred. this is a terrible idea, the fallback should be a looser type rather than a stricter one.

If your stub returns a `Promise<number>` as you said above, it will be typed as `resolves(value?: number)`. Why wouldn't it? its a conditional type.. the return type _does_ extend `PromiseLike<number>` in your example, so it _does_ infer `number`. Not `any`.

types/sinon/ts3.1/index.d.ts
43081j: no, it will not.

`TReturnValue` is from the type parameters of the stub. You can't pass a string in, `TReturnValue` is `Promise<number>` which extends `PromiseLike<number>` which means `resolves(value?: number)` here.

43081j: there you go i've even added a test now to show you this is the case. does it make more sense now?

nicojs (approved):
LGTM. Just to be clear, this solves the 3rd issue you've discussed in this comment right?
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/36019#issuecomment-500821065

43081j:
It does. It solves the 2nd too (in case you have a return type of `any` or a mixed return type).

alecf:
Here is a use case that no longer breaks, when it should have broken:

```typescript
const o = {
  fn() {
    return 1;
  },
};

const s = sinon.stub(o);
s.fn.resolves('hello');
```

Before this PR, you'd get:
```
Argument of type '"hello"' is not assignable to parameter of type 'never'.ts(2345)
```

With this PR you get no error. I believe this is incorrect behavior. Thoughts, @nicojs ?

43081j:
This really isn't incorrect behaviour.

`never` was preventing people from using the sinon types for any un-inferrable types (overloads, mixed returns, anys, unknowns, etc.).

if someone wants to non-promise-returning function return a promise, why should you prevent them from doing that? 

the types were broken before, hence all the comments in the related issues of people being unable to use them anymore. 

do you want to break the types for everyone who has a non-simple function, and make them completely unusable, or do you want to allow people to make functions which previously didn't return a promise, return a promise?

```
function foo() {
  if (x) {
    return Promise.resolve();
  }
  return true;
}
```

what is the return type? it will be inferred as a union type or won't be inferrable at all. i still want to stub it and make it resolve to something in my tests, why shouldn't i? the types allow this, because the developer knows best in these cases where the compiler cannot infer it for them.

PR #36127, opened 2019-06-11
little-loader types

chrisdrackett:
Please fill in this template.

### DIFF
--- a/types/little-loader/index.d.ts
+++ b/types/little-loader/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for little-loader
+// Type definitions for little-loader 0.2
 // Project: https://github.com/walmartlabs/little-loader
 // Definitions by: Chris Drackett <https://github.com/chrisdrackett>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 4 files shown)


types/little-loader/index.d.ts
```diff
@@ -0,0 +1,12 @@
+// Type definitions for little-loader 0.2
+// Project: https://github.com/walmartlabs/little-loader
+// Definitions by: Chris Drackett <https://github.com/chrisdrackett>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare function loader(
+    module: string,
+    callBack: (err: string) => void,
+    context?: any
+): void;
```
DanielRosenwasser: This doesn't model the options object case. Would a more correct version be the following?

```ts
interface Options<T> {
    context?: T;
    setup?: (this: T, script: any) => void;
    callback?: (this: T, err: string) => void;
}

declare function loader<T>(
    module: string,
    callback?: (this: T, err: string) => void,
    context?: T
): void;

declare function loader<T>(
    module: string,
    options?: Options<T>
): void;

export = loader;

```

chrisdrackett: good call, I had never used that form!

DanielRosenwasser:
Thanks!

PR #36133, opened 2019-06-12
types for dis-gui

mhermher:
Please fill in this template.

- [ ] Types for package dis-gui. Types do not exist anywhere else that I could find.

If adding a new definition:

### DIFF
--- a/types/dis-gui/dis-gui-tests.ts
+++ b/types/dis-gui/dis-gui-tests.ts
@@ -1,0 +1,78 @@
+import * as React from 'react';
+import * as dg from './index';
+
+React.createElement(
+    dg.GUI,
+    {
+        style : {width : '200px'},
+        expanded : false
+    },
+    React.createElement(
+        dg.Folder,
+        {
+            label : 'Some Folder',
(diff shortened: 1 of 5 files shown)


mhermher:
Honestly, not sure how to resolve this failure. It is due to my setting one of the types to `React.CCSProperties`. React is a dependency of this package. In the React type def is : 
```export interface CSSProperties extends CSS.Properties<string | number> {
        /**
         * The index signature was removed to enable closed typing for style
         * using CSSType. You're able to use type assertion or module augmentation
         * to add properties or an index signature of your own.
         *
         * For examples and more information, visit:
         * https://github.com/frenic/csstype#what-should-i-do-when-i-get-type-errors
         */
    }```

and `import * as CSS from 'csstype';` at the top of the file. This is caught in vscode's linter by the way. So I am not sure if this failure is in the react dir rather than the dir that I am adding. I added ccstype as a dependency (and added a package.json) to get the type import to actually work. I had to npm install from inside the dir though. Not sure if I am doing this right. I would like to allow for the full range of CSS properties, but don't want to write them all out myself for the sake of passing this test.

PR #36145, opened 2019-06-12
(got) fix: add init hook

jaulz:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/sindresorhus/got/blob/master/readme.md#hooksinit

### DIFF
--- a/types/got/got-tests.ts
+++ b/types/got/got-tests.ts
@@ -347,7 +347,11 @@ got('example.com', {
             options => {
                 options.baseUrl = 'https://google.com';
             }
-        ],
+        ]
+    }
+});
+got('example.com', {
+    hooks: {
         beforeRequest: [
             options => {
                 options.headers!['x-foo'] = 'bar';
--- a/types/got/got-tests.ts
+++ b/types/got/got-tests.ts
@@ -343,6 +343,11 @@ got('http://todomvc.com', { timeout: 1 }).catch((err) => err instanceof got.Time
 // Test hooks.
 got('example.com', {
     hooks: {
+        init: [
+            options => {
+                options.baseUrl = 'https://google.com';
+            }
+        ],
         beforeRequest: [
             options => {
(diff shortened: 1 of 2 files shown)


types/got/got-tests.ts
```diff
@@ -343,6 +343,11 @@ got('http://todomvc.com', { timeout: 1 }).catch((err) => err instanceof got.Time
 // Test hooks.
 got('example.com', {
     hooks: {
+        init: [
```
ikokostya: As you can see each test contains only one case, i.e. hook. So, move this hook test to separate case.

ikokostya (changes requested):
`init` hook was added in [9.6.0 release](https://github.com/sindresorhus/got/releases/tag/v9.6.0). Need to  increase the version number in the header:

```diff
- // Type definitions for got 9.4
+ // Type definitions for got 9.6
```

types/got/index.d.ts
```diff
@@ -140,12 +140,18 @@ declare namespace got {
...
         beforeRequest?: Array<BeforeRequestHook<Options>>;
         beforeRedirect?: Array<BeforeRedirectHook<Options>>;
         beforeRetry?: Array<BeforeRetryHook<Options>>;
         afterResponse?: Array<AfterResponseHook<Options, Body>>;
     }
 
+    /**
+     * @param options Unnormalized request options.
+     */
+    type InitHook<Options> = (options: Options) => any;
```
ikokostya: Documentation [says](https://github.com/sindresorhus/got/blob/master/readme.md#hooksinit) 

> Note: This hook must be synchronous!

Maybe  more appropriate return type here is `void`. `any` type is used for other hooks, because they can return `Promise`.

jaulz:
Thanks for the comments! Should be fixed now.

ikokostya (approved):

PR #36151, opened 2019-06-12
[dirty-chai] fix 'esModuleInterop' requirement

gstamac:
Please fill in this template.

Change was required since the definitions resulted in "error TS2497: This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.".

### DIFF
--- a/types/dirty-chai/index.d.ts
+++ b/types/dirty-chai/index.d.ts
@@ -1,28 +1,30 @@
 // Type definitions for dirty-chai 2.0
 // Project: https://github.com/prodatakey/dirty-chai
 // Definitions by: Piotr Roszatycki <https://github.com/dex4er>
+//                 Gregor Stamać <https://github.com/gstamac>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 /// <reference types="chai" />
 /// <reference types="chai-as-promised" />
 
-declare global {
-    namespace Chai {
-        interface LanguageChains {
-            always: Assertion;
-        }
+import Assertion = Chai.Assertion;
 
-        interface Assertion {
-            (message?: string): Assertion;
-            ensure: Assertion;
-        }
+declare module 'dirty-chai' {
+    function dirtyChai(chai: any, utils: any): void;
 
-        interface PromisedAssertion extends Eventually, PromiseLike<any> {
-            (message?: string): Assertion;
-            ensure: Assertion;
-        }
-    }
+    namespace dirtyChai { }
+
+    export = dirtyChai;
 }
 
-declare function dirtyChai(chai: any, utils: any): void;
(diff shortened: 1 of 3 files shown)


dex4er (approved):
Works for me

DanielRosenwasser:
None of this seems appropriate to me. We specifically have lint rules against this because the empty namespace thing is a hack based on an old TypeScript bug.

I also don't know why this package ever declared types globally to begin with, but that can be a separate PR.

gstamac:
@DanielRosenwasser Does that mean this PR will not be merged? Is there another way to do this because currently 'dirty-chai' typings don't work?

dex4er:
@gstamac Why do you need this patch? `dirty-chai-tests.ts` file uses this module without `esModuleInterop` setting so maybe your patch is not required at all?

gstamac:
@dex4er It definitelly doesn't work on our projects. Here is our tsconfig flie
```
{
  "compilerOptions": {
    "module": "commonjs",
    "target": "es2017",
    "strict": true,
    "declaration": true,
    "noImplicitAny": true,
    "moduleResolution": "node",
    "sourceMap": true,
    "forceConsistentCasingInFileNames": true,
    "outDir": "dist",
    "baseUrl": ".",
    "paths": {
      "*": [...]
    },
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true
  },
  "include": [...],
  "typeRoots": [
    "node_modules/@types",
    "src/types"
  ],
  "noUnusedParameters": true
}
```

dex4er:
@gstamac I'm pretty sure that you try to `import * as dirtyChai from 'dirty-chai'` or `import dirtyChai from 'dirty-chai'` rather than `import dirtyChai = require("dirty-chai")`.

Typings for `dirty-chai` works for me with or without `esModuleInterop`. With `false` I use 3rd option, with `true` 2nd option.

gstamac:
3rd option is not allowed because of tslint rules and it shouldn't be used with typescript in my opinion. If I use 1st I geterror `This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.` if I use 2nd option I get error `This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.`

dex4er:
@gstamac If you don't want to use `import = require()` syntax in Typescript then enable `esModuleInterop` option. It was designed just for this case. If you can't use it because of compatibility issues or whatever then you must use `require()` and then ignore this bullshit from `tslint`.

I'm afraid you want to eat cake it have it too.

gstamac:
@dex4er I defined the typings so it works correctly without `esModuleInterop` and without `import = require()`. Like all the other `chai` add-ons.

PR #36154, opened 2019-06-12
Add definitions for jscodeshift/src/testUtils

bjacobel:
Adds types for a helper module exported as `/src/testUtils` from jscodeshift. 

(No lint errors introduced to the ones already on master.)

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/jscodeshift/blob/master/src/testUtils.js

### DIFF
--- a/types/jscodeshift/index.d.ts
+++ b/types/jscodeshift/index.d.ts
@@ -5,7 +5,6 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.9
 
-/// <reference path="./src/testUtils.d.ts" />
 import core = require("./src/core");
 
 export = core;
--- a/types/jscodeshift/index.d.ts
+++ b/types/jscodeshift/index.d.ts
@@ -5,7 +5,7 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.9
 
+/// <reference path="./src/testUtils.d.ts" />
 import core = require("./src/core");
-import testUtils = require('./src/testUtils');
 
 export = core;
--- a/types/jscodeshift/index.d.ts
+++ b/types/jscodeshift/index.d.ts
@@ -1,9 +1,11 @@
 // Type definitions for jscodeshift 0.6
 // Project: https://github.com/facebook/jscodeshift#readme
 // Definitions by: Brie Bunge <https://github.com/brieb>
+//                 Brian Jacobel <https://github.com/bjacobel>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.9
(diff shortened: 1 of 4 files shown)


brieb:
Happy to review this once the build is passing.

types/jscodeshift/index.d.ts
```diff
@@ -1,9 +1,11 @@
 // Type definitions for jscodeshift 0.6
 // Project: https://github.com/facebook/jscodeshift#readme
 // Definitions by: Brie Bunge <https://github.com/brieb>
+//                 Brian Jacobel <https://github.com/bjacobel>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.9
 
 import core = require("./src/core");
+import testUtils = require('./src/testUtils');
```
brieb: nit: `"` instead of `'` for consistency

brieb (approved):

brieb (changes requested):
Had one minor nit. Otherwise looks good. Thank you for the contribution!

types/jscodeshift/index.d.ts
brieb: Does a `// <reference ... />` work here instead of a noop import?

brieb: Also, why is this import necessary?

bjacobel: It's needed to avoid an `Error: Unused file` from the parser in types-publisher. However, a reference works as well, and makes more sense, so I will change to that.

Thank you for the review!

brieb (approved):

types/jscodeshift/index.d.ts
```diff
@@ -1,9 +1,11 @@
 // Type definitions for jscodeshift 0.6
 // Project: https://github.com/facebook/jscodeshift#readme
 // Definitions by: Brie Bunge <https://github.com/brieb>
+//                 Brian Jacobel <https://github.com/bjacobel>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.9
 
+/// <reference path="./src/testUtils.d.ts" />
```
DanielRosenwasser: I think it'd probably be better to place this in the `tsconfig.json`.

DanielRosenwasser (changes requested):

types/jscodeshift/index.d.ts
bjacobel: I agree, that makes more sense - thank you for the review.

DanielRosenwasser:
Thank you!

types/jscodeshift/index.d.ts
brieb: Thanks for weighing in -- that's a much better idea! 😄

PR #36155, opened 2019-06-12
Support Yup required/notRequired derived types

mauricedb:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: See below
- [x] ~~Increase the version number in the header if appropriate.~~
- [x] ~~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.~~

Adding support for deriving a TypeScript type alias with `required()` and `notRequired()` properties.

Consider the following Yup schema:
```TypeScript
    firstName: yup.string(),
    email: yup
        .string()
        .nullable()
        .notRequired()
        .email(),
    birthDate: yup
        .date()
        .nullable()
        .notRequired()
        .min(new Date(1900, 0, 1)),
});
```

We can now derive the TypeScript type as follows:
```TypeScript
type Person = yup.InferType<typeof personSchema>;
```

Which is equivalent to the following type:
```TypeScript
type Person = {
    firstName: string;
    email?: string | null | undefined;
    birthDate?: Date | null | undefined;
}
```

Making the following code valid both for TypeScript and Yup validation:
```TypeScript
const minimalPerson: Person = {
    firstName: "Jack"
};

const fullPerson: Person = {
    firstName: "Jack",
    email: "jack@somewhere.com",
    birthDate: new Date(2000, 0, 1)
};
```

### DIFF
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -96,10 +96,10 @@ export interface MixedSchemaConstructor {
     new (options?: { type?: string; [key: string]: any }): MixedSchema;
 }
 
-// tslint:disable-next-line:no-empty-interface
 export interface MixedSchema<T = any> extends Schema<T> {
     nullable(isNullable?: true): MixedSchema<T | null>;
     nullable(isNullable: false): MixedSchema<Exclude<T, null>>;
+    nullable(isNullable?: boolean): MixedSchema<T>;
     required(message?: TestOptionsMessage): MixedSchema<Exclude<T, undefined>>;
     notRequired(): MixedSchema<T | undefined>;
 }
@@ -128,6 +128,7 @@ export interface StringSchema<T extends string | null | undefined = string>
     uppercase(message?: TestOptionsMessage): StringSchema<T>;
     nullable(isNullable?: true): StringSchema<T | null>;
     nullable(isNullable: false): StringSchema<Exclude<T, null>>;
+    nullable(isNullable?: boolean): StringSchema<T>;
     required(message?: TestOptionsMessage): StringSchema<Exclude<T, undefined>>;
     notRequired(): StringSchema<T | undefined>;
 }
@@ -156,6 +157,7 @@ export interface NumberSchema<T extends number | null | undefined = number>
     round(type: "floor" | "ceil" | "trunc" | "round"): NumberSchema<T>;
     nullable(isNullable?: true): NumberSchema<T | null>;
     nullable(isNullable: false): NumberSchema<Exclude<T, null>>;
+    nullable(isNullable?: boolean): NumberSchema<T>;
     required(message?: TestOptionsMessage): NumberSchema<Exclude<T, undefined>>;
     notRequired(): NumberSchema<T | undefined>;
 }
@@ -165,11 +167,11 @@ export interface BooleanSchemaConstructor {
     new (): BooleanSchema;
 }
 
-// tslint:disable-next-line:no-empty-interface
 export interface BooleanSchema<T extends boolean | null | undefined = boolean>
     extends Schema<T> {
     nullable(isNullable?: true): BooleanSchema<T | null>;
     nullable(isNullable: false): BooleanSchema<Exclude<T, null>>;
+    nullable(isNullable?: boolean): BooleanSchema<T>;
     required(
         message?: TestOptionsMessage
     ): BooleanSchema<Exclude<T, undefined>>;
@@ -193,6 +195,7 @@ export interface DateSchema<T extends Date | null | undefined = Date>
     ): DateSchema<T>;
     nullable(isNullable?: true): DateSchema<T | null>;
     nullable(isNullable: false): DateSchema<Exclude<T, null>>;
+    nullable(isNullable?: boolean): DateSchema<T>;
     required(message?: TestOptionsMessage): DateSchema<Exclude<T, undefined>>;
     notRequired(): DateSchema<T | undefined>;
 }
@@ -215,6 +218,7 @@ export interface NotRequiredNullableArraySchema<T>
     of<U>(type: Schema<U>): NotRequiredNullableArraySchema<U>;
     nullable(isNullable?: true): NotRequiredNullableArraySchema<T>;
     nullable(isNullable: false): NotRequiredArraySchema<T>;
+    nullable(isNullable?: boolean): ArraySchema<T>;
     required(message?: TestOptionsMessage): NullableArraySchema<T>;
     notRequired(): NotRequiredNullableArraySchema<T>;
 }
@@ -223,6 +227,7 @@ export interface NullableArraySchema<T> extends BasicArraySchema<T[] | null> {
     of<U>(type: Schema<U>): NullableArraySchema<U>;
     nullable(isNullable?: true): NullableArraySchema<T>;
     nullable(isNullable: false): ArraySchema<T>;
+    nullable(isNullable?: boolean): ArraySchema<T>;
     required(message?: TestOptionsMessage): NullableArraySchema<T>;
     notRequired(): NotRequiredNullableArraySchema<T>;
 }
@@ -232,6 +237,7 @@ export interface NotRequiredArraySchema<T>
     of<U>(type: Schema<U>): NotRequiredArraySchema<U>;
     nullable(isNullable?: true): NotRequiredNullableArraySchema<T>;
     nullable(isNullable: false): NotRequiredArraySchema<T>;
+    nullable(isNullable: boolean): ArraySchema<T>;
     required(message?: TestOptionsMessage): ArraySchema<T>;
     notRequired(): NotRequiredArraySchema<T>;
 }
@@ -239,7 +245,7 @@ export interface NotRequiredArraySchema<T>
 export interface ArraySchema<T> extends BasicArraySchema<T[]> {
     of<U>(type: Schema<U>): ArraySchema<U>;
     nullable(isNullable?: true): NullableArraySchema<T>;
-    nullable(isNullable: false): ArraySchema<T>;
+    nullable(isNullable: false | boolean): ArraySchema<T>;
     required(message?: TestOptionsMessage): ArraySchema<T>;
     notRequired(): NotRequiredArraySchema<T>;
 }
@@ -279,6 +285,7 @@ export interface ObjectSchema<T extends object | null | undefined = object>
     constantCase(): ObjectSchema<T>;
     nullable(isNullable?: true): ObjectSchema<T | null>;
     nullable(isNullable: false): ObjectSchema<Exclude<T, null>>;
+    nullable(isNullable?: boolean): ObjectSchema<T>;
     required(message?: TestOptionsMessage): ObjectSchema<Exclude<T, undefined>>;
     notRequired(): ObjectSchema<T | undefined>;
 }
@@ -312,7 +319,7 @@ export type WhenOptionsBuilderObject =
       }
     | object;
 
-export type WhenOptions<T> =
+    export type WhenOptions<T> =
     | WhenOptionsBuilderFunction<T>
     | WhenOptionsBuilderObject;
 
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -204,6 +204,67 @@ export interface ArraySchemaConstructor {
     new (): ArraySchema<{}>;
 }
 
+interface NotRequiredNullableArraySchema<T>
+    extends Schema<T[] | null | undefined> {
+    of<U>(type: Schema<U>): NotRequiredNullableArraySchema<U>;
+    min(
+        limit: number | Ref,
+        message?: TestOptionsMessage
+    ): NotRequiredNullableArraySchema<T>;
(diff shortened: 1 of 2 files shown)


mauricedb:
Note, there is quite an expansion of the type count reported. I am not quite sure how this is determined or how I can run this locally but assume it is the result of the generic types that can also be null or undefined. As I am not sure I fully understand this metric I am not sure how much of a problem this is of what can be done to decrease the type count.

types/yup/index.d.ts
```diff
@@ -183,13 +195,76 @@ export interface DateSchema<T extends Date | null = Date> extends Schema<T> {
...
+    required(message?: TestOptionsMessage): DateSchema<Exclude<T, undefined>>;
+    notRequired(): DateSchema<T | undefined>;
 }
 
 export interface ArraySchemaConstructor {
     <T>(schema?: Schema<T>): ArraySchema<T>;
     new (): ArraySchema<{}>;
 }
 
+interface NotRequiredNullableArraySchema<T>
```
YashdalfTheGray: I wonder if there is a way to write this so there isn't so much repetition but that might make this typings file more convoluted probably. I definitely don't think it's necessary but all of these types might have something to do with the automated analysis marking the changes as big impact or whatever. 

I'm mostly just thinking out loud here.

types/yup/index.d.ts
```diff
@@ -435,3 +514,18 @@ export interface LocaleObject {
...
 }
+
+export type InferType<T> = T extends Schema<infer P>
+    ? InnerInferType<P>
+    : never;
+
+// Shut off automatic exporting after this statement
+export {};
+
+type KeyOfUndefined<T> = {
```
YashdalfTheGray: I think having these types are clutch!

types/yup/index.d.ts
mauricedb: @YashdalfTheGray I would prefer to remove the duplication as well. I tried but my attempts didn't work out. Any good suggestions on how to do so are appreciated and will be used in my PR.

types/yup/index.d.ts
mauricedb: Thanks :-)

When this PR is merged I will do a PR on the yup docs to show how to use this.

types/yup/index.d.ts
YashdalfTheGray: Man, I was trying to think of any but really, I can't think of any that wouldn't make the code harder to decipher. In my mind, if you have lots of `Type<<<>>>` things in the code, it makes it harder to see what's actually going on. So I think I'm okay to let it ride like this. 

I wonder if @ssepola has any ideas!

YashdalfTheGray:
> Note, there is quite an expansion of the type count reported. I am not quite sure how this is determined or how I can run this locally but assume it is the result of the generic types that can also be null or undefined. As I am not sure I fully understand this metric I am not sure how much of a problem this is of what can be done to decrease the type count.

@andrewbranch can you help us interpret these typescript-bot metrics?

andrewbranch:
Well, the good news is that language service response times didn’t go up much. It does seem like these types incur a significantly larger memory penalty. You can get memory and type count numbers by compiling your project with `tsc --diagnostics -p .` At a quick glance I’m not 100% sure if there’s a single culprit or an easy optimization here... It’s not my place (this week, anyway) to make a value judgement about these metrics, just to make sure you have the info you need to ask yourself whether the improvements in this PR justify 378% higher memory usage. I know that’s fairly abstract to think about, but in the future, we’ll be able to show where that puts yup compared to other packages on DT. As the bot said, this is still a work-in-progress; hopefully we’ll be able to provide more guidance on this stuff soon.

mauricedb:
@andrewbranch Thanks for the explanation. Let me see if I can do anything about this and het the number of types and memory consumption down a bit. It's worth a shot I would say.

mauricedb:
Please hold on merging this PR for a bit while I investigate the memory and type count later today or tomorrow.

types/yup/index.d.ts
sseppola: I'm not sure, tell me if you've tried this before, but if we can write NullableArraySchema as `ArraySchema<T | null>`, then we can reverse this operation with `ArraySchema<Exclude<T, null>>`, and same goes for notRequired/required.

Since T in ArraySchema is the type of the content of the array, I think this should be possible.

mauricedb: @sseppola I tried that but unless I made a mistake that doesn't work. With `interface ArraySchema<T> extends Schema<T[]>` that means `ArraySchema<T | null>`  is a `Schema<(T | null)[]>` instead of `Schema<(T[]) | null>`. So the array can contain `null` values as well as their designated type instead of the array reference being `null`.

sseppola: True..
I'm not sure and a bit short on time to explore this in depth, so excuse me for throwing out ideas. 

1. Conditional typing:
`type ArraySchema<T> = null | undefined extends T ? [something] : [other]`
However that will probably get really really messy because we have to juggle both undefined and null. I'm not even sure if `null | undefined extends T` works as I expect, so that may trip people up.

2. Overloading
Is there a way to overload the ArraySchema, where if null/undefined is present in T we modify the types?  I don't think it's possible to overload interfaces, I think that's just for functions.

This is really tough..

mauricedb:
The type count issue has been solved. Turned out the be because of the `nullable()`/`required()` functions in `Schema<T>`.

I also reduced the duplication with the `ArraySchema<T>` so I am happy with the current state of the PR. If everyone else agrees this PR can be merged.

types/yup/index.d.ts
mauricedb: When I try overloading `BasicArraySchema` I get an error: `All declarations of 'BasicArraySchema' must have identical type parameters.ts(2428)`. As I was expecting these declarations get merged into one based on their name so require identical type parameters.

types/yup/index.d.ts
```diff
@@ -102,6 +98,8 @@ export interface MixedSchemaConstructor {
 
 // tslint:disable-next-line:no-empty-interface
 export interface MixedSchema<T = any> extends Schema<T> {
+    nullable(isNullable?: true): MixedSchema<T | null>;
+    nullable(isNullable: false): MixedSchema<Exclude<T, null>>;
```
sandersn: can `nullable` be called with anything but a literal? It doesn't work with an unknown boolean:

```ts
function wrapper(b: boolean, msx: MixedSchema) {
  return msx.nullable(b)
}
```

Gives an error "Argument of type 'boolean' is not assignable to 'false'."

types/yup/index.d.ts
```diff
@@ -153,6 +156,8 @@ export interface NumberSchema<T extends number | null = number>
     round(type: "floor" | "ceil" | "trunc" | "round"): NumberSchema<T>;
     nullable(isNullable?: true): NumberSchema<T | null>;
     nullable(isNullable: false): NumberSchema<Exclude<T, null>>;
```
sandersn: (reposting comment on PR instead of commit)
can nullable be called with anything but a literal? It doesn't work with an unknown boolean:

```ts
function wrapper(b: boolean, msx: MixedSchema) {
  return msx.nullable(b)
}
```

Gives an error "Argument of type 'boolean' is not assignable to 'false'."

YashdalfTheGray: We might need to flip the order of these function declarations for the types to flow correctly?

YashdalfTheGray:
@mauricedb solid work on getting the numbers lower my dude! There is one outstanding comment, any ideas on how to solve that piece?

types/yup/index.d.ts
sandersn: I think you actually need another overload for `boolean`, but I'm not sure what to return. Just `NumberSchema<T>`?

sandersn:
@YashdalfTheGray That might be an issue for a separate PR since this PR didn't come up with the pattern of incomplete overloads.

types/yup/index.d.ts
mauricedb: @sandersn 	

Adding an extra overload to the `MixedSchema`:
```TypeScript
export interface MixedSchema<T = any> extends Schema<T> {
    nullable(isNullable?: boolean): MixedSchema<T>;
    nullable(isNullable?: true): MixedSchema<T | null>;
    nullable(isNullable: false): MixedSchema<Exclude<T, null>>;
    required(message?: TestOptionsMessage): MixedSchema<Exclude<T, undefined>>;
    notRequired(): MixedSchema<T | undefined>;
}
```

And overloading the wrapper like:
```TypeScript
function wrapper<T>(b: false, msx: MixedSchema<T>): MixedSchema<T>;
function wrapper<T>(b: true, msx: MixedSchema<T>): MixedSchema<T | null>;

function wrapper<T>(b: boolean, msx: MixedSchema<T>) {
    return msx.nullable(b);
}
``` 

Should IMHO result in the correct typing:
```TypeScript
const resultingSchema1 = wrapper<string | number>(false, yup.mixed().oneOf(['1', 2]));
const resultingSchema2 = wrapper<string | number>(true, yup.mixed().oneOf(['1', 2]));
```

Resulting type signatures:
```TypeScript
const resultingSchema1: yup.MixedSchema<string | number>
const resultingSchema2: yup.MixedSchema<string | number | null>
```

sandersn: Shouldn't the boolean overload go last since it will otherwise match literal `true` and `false`?

YashdalfTheGray: Yeah, I'm thinking this too.

mauricedb: Agreed, added the various overloads and a test for them.

types/yup/index.d.ts
```diff
@@ -100,15 +96,20 @@ export interface MixedSchemaConstructor {
     new (options?: { type?: string; [key: string]: any }): MixedSchema;
 }
 
-// tslint:disable-next-line:no-empty-interface
-export interface MixedSchema extends Schema<any> {}
+export interface MixedSchema<T = any> extends Schema<T> {
+    nullable(isNullable?: true): MixedSchema<T | null>;
+    nullable(isNullable: false): MixedSchema<Exclude<T, null>>;
+    nullable(isNullable?: boolean): MixedSchema<T>;
```
DanielRosenwasser: Instead of overloads, what about

```ts
nullable(isNullable?: boolean): MixedSchema<T extends true ? T | null : Exclude<T, null>>;

Comment (mauricedb) [reply types/yup/index.d.ts]: @DanielRosenwasser 	

Sweet, I like removing the overloads, makes the definition more readable.

However when I try this works fine but using a wrapper function becomes harder. The [wrapper test code here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36155/files#diff-3ed4607a6b19a001ccf1d37f42e3f387R663) needs to be updated to the following. 

```TypeScrript
function wrapper<T>(b: false, msx: MixedSchema<T>): MixedSchema<T>;
function wrapper<T>(b: true, msx: MixedSchema<T>): MixedSchema<T | null>;

function wrapper<T>(b: boolean, msx: MixedSchema<T>): MixedSchema<T extends true ? T | null : Exclude<T, null>> {
    return msx.nullable(b);
}

const resultingSchema1 = wrapper<string | number>(false, yup.mixed().oneOf(['1', 2])); // $ExpectType MixedSchema<string | number>
const resultingSchema2 = wrapper<string | number>(true, yup.mixed().oneOf(['1', 2])); // $ExpectType MixedSchema<string | number | null>
```

So the overloads are still required else the resulting type changes and the return of the wrapper function is a lot more verbose.

Personally I would favor more complexity in the type definition with a simple API over a simple type definition and a more complex API for developers to use. 

Lets put this to the vote and see what the others think and go with that.

YashdalfTheGray: I agree with 👇 statement pretty wholeheartedly

> Personally I would favor more complexity in the type definition with a simple API over a simple type definition and a more complex API for developers to use.

I do wonder though, how many of our clients are using wrapper code like that because if that's somewhat of an "advanced" usecase, then I'm okay having the types be conditional so that we can get some of the benefits of simpler types on the type author side as well.

chmanie:
Could it be that this functionality requires `strictNullChecks` to be enabled (it is disabled by default)?

See http://www.typescriptlang.org/play/?strictNullChecks=false#code/JYWwDg9gTgLgBAKjgQwM5wJ4FcxwGZQQhwDk2YJA3ALABQdAxhAHaryoMAWApiMnAF5MOAHQQARgCtuDGAAoA3nTgq4AEyLJgzAJJqAXMLAjmWEOO5Q5AShFRuARyzB7amwBplq5shDdD5CJsUNoA5jZ2js6uHnQAvtY09LRMrPASkgGiOsx4lgAqGGDcADwwRdwQeHAcPHwAfIJwCuqa2nqGAIzucD5+hiQ8ADZDECRwcUl02jCWeMgM3HAAogBu3MzwLRp87Qa9ZhZQlL2+-jUwIcyhE1MpLGxwGQBMhmsb8EIZlEA

mauricedb:
@chmanie Looks like it does. I never realized because I always run with strict enabled.

PR #36157, opened 2019-06-12
[@types/enzyme] adding `invoke` definition to support enzyme@3.10

rgibson518:
Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<
https://github.com/airbnb/enzyme/blob/b8289d02355882c57bf334ce001307807a674661/packages/enzyme/src/ShallowWrapper.js#L1305,
https://github.com/airbnb/enzyme/blob/master/packages/enzyme/src/ReactWrapper.js#L837
>>

If removing a declaration:

#### Misc
Just not sure how the process moves to support each version of Enzyme. I asked on their gitter, but they weren't really sure, either.  Do I need to bump the version in the `index.d.ts`?

#### Module Augmentation
I made this PR since I had already done a module augmentation to support `invoke`.  For those interested, here is the augmentation for the ShallowWrapper. I did not test this in my project using the (Common|React)Wrapper, however it should be trivial to make it work for them.
```typescript
import { Component } from 'react';

declare module 'enzyme' {
  interface ShallowWrapper<P = {}, S = {}, C = Component> {
    /**
     * Invokes a function prop.
     * @param invokePropName The function prop to call.
     * @param ...args The argments to the invokePropName function
     * @returns The value of the function.
     */
    invoke<K extends keyof P>(invokePropName: K): P[K];
  }
}
```


types/enzyme/index.d.ts
```diff
@@ -92,6 +92,14 @@ export interface CommonWrapper<P = {}, S = {}, C = Component<P, S>> {
      */
     hasClass(className: string | RegExp): boolean;
 
+    /**
+     * Invokes a function prop.
+     * @param invokePropName The function prop to call.
+     * @param ...args The argments to the invokePropName function
+     * @returns The value of the function.
+     */
+    invoke<K extends keyof P>(invokePropName: K): P[K];
```
ljharb: `invoke` returns a function, which when again invoked, calls the prop function.

Thus, the type here should perhaps be (assuming P is the props):
```suggestion
    invoke<K extends keyof P>(invokePropName: K): P[K] extends Function
```
but ideally it'd also give you a type error when you passed a `K` such that `P[K]` isn't a function.

types/enzyme/enzyme-tests.tsx
```diff
@@ -16,6 +16,7 @@ import { Component, ReactElement, ReactNode, HTMLAttributes, ComponentClass, Sta
 interface MyComponentProps {
     stringProp: string;
     numberProp: number;
+    functionProp(...args: string[]): boolean;
```
jwbay: Would like to this prop made optional so it's less invasive to the rest of the tests that don't care about invoke. If strictNullChecks is on, using ! is fine imo, or declaring a separate component/props for the invoke tests.

types/enzyme/enzyme-tests.tsx
```diff
@@ -640,6 +663,21 @@ function ReactWrapperTest() {
         boolVal = reactWrapper.find('.my-button').hasClass('disabled');
     }
 
+    function test_invoke() {
+        let is_bool: boolean;
+        is_bool = reactWrapper.invoke('functionProp')(); // no args
```
jwbay: Assignability can be a weak type-level assertion. Invoke could start returning any, for example, and all these tests would pass. Can you try `// $ExpectType number` instead for these? I'm not sure where the magic comment can be placed (above, same line, etc)

types/enzyme/index.d.ts
jwbay: The return type isn't currently a valid position for a constraint in TS. This seems to do what we want

```ts
type FunctionPropertyNames<T> = { [K in keyof T]: T[K] extends Function ? K : never }[keyof T];

interface Wrapper<P> {
    invoke<K extends keyof Pick<P, FunctionPropertyNames<P>>>(invokePropName: K): P[K]
}
```

jwbay (changes requested):

types/enzyme/index.d.ts
rgibson518: Haha I was just working to update to use `FunctionPropertyNames`! Found that in the TS Handbook after a colleague at work mentioned it.  

This is what I had:
```typescript
/**
   * Invokes a function prop.
   * @param invokePropName The function prop to call.
   * @param ...args The argments to the invokePropName function
   * @returns The value of the function.
   */
  invoke<K extends NonNullable<{
      [K in keyof P]: P[K] extends () => void ? K : never
  }[keyof P]>>(invokePropName: K): P[K];
```

I inlined the `FunctionPropertyNames` because it seems that DT auto exports all declarations, so I wasn't sure if this was an export that should be owned by `@types/enzyme`. 

I was thinking of using `NonNullable` since the conditional allows for `undefined` and the inlined `FunctionPropertyNames` already "picks" those keys. I hadn't considered `Pick`. However, I do think `Pick` might make the intent more clear.  Thoughts on which to use?

types/enzyme/enzyme-tests.tsx
rgibson518: After working with the aboves changes, I've started pulling out the "invoked" assignment tests. After thinking about it, these tests just prove that `functionProp` works as expect, not `invoke`.   I wrote these originally after looking at the enzyme docs for `invoke`. Being inspired by the `(...args)` in the docs, it felt necessary, but I'm feeling less so now.

types/enzyme/enzyme-tests.tsx
rgibson518: I'll look into this, it felt bad updating so many tests that didn't rely on `functionProp`.

types/enzyme/enzyme-tests.tsx
ljharb: is "$ExpectType" a TS thing, or a DT tests thing?

rgibson518:
As a kind of tertiary improvement, thoughts on DRYing up the `invoke` tests into a `CommonWrapperHelper`? 

Example: 
```typescript
// outside of tests
const CommonWrapperHelper = {
    test_invoke: (wrapper: CommonWrapper<MyComponentProps, any, any>) => {
        // should only accept the names of props that are functions
        let is_matching_function: MyComponentProps['functionProp'] = wrapper.invoke('functionProp');
        // $ExpectError
        is_matching_function = wrapper.invoke('stringProp'); // should not accept non-function prop
        // $ExpectError
        const is_not_matching_function: () => string = wrapper.invoke('functionProp'); // should return matching function type
    },
};
// ShallowWrapper
function ShallowWrapperTest() {
    // ...other tests
    function test_invoke() {
        CommonWrapperHelper.test_invoke(shallowWrapper);
    }
};
// ReactWrapper
function ReactWrapperTest() {
    // ...other tests
    function test_invoke() {
        CommonWrapperHelper.test_invoke(reactWrapper);
    }
}
```

types/enzyme/enzyme-tests.tsx
rgibson518: Seems to be part of [dtslint](https://github.com/microsoft/dtslint).

types/enzyme/index.d.ts
jwbay: 👍For inlining

jwbay:
Do any other maintainers have any opinion here?

I'm a bit concerned about introducing abstractions in the tests that might make them harder to follow. There's already a lot going on in them. However, the current duplication is definitely and consistently annoying. I think I lean slightly toward your approach personally 👍

rgibson518:
I have some more changes coming (as outlined in a few comments), please do not merge.

rgibson518:
Still more coming... I think I need to make the `functionProps` less invasive.  Leaning toward extending the `MyComponentProps` with the `functionProps`, however I'm not 100% sure on this approach.

I'm also planning to prove that an undefined function prop should not be invokable.  At runtime, `invoke` will throw a TypeError if the prop is not a function, so I think it's reasonable to just "*prevent*" that with types in the first place.
```typescript
//example
interface Props {
  functionProp?(): void;
}
const MyComp: React.FC<Props> = ({ functionProp }) => <div></div>;
const wrapper = shallow(<MyComp/>);
// I believe invoke will accept `never` as argument since there are no props that satisfy it
// but I'm not very fluent in `never` nuances.
wrapper.invoke('functionProp') // $ExpectError
```

Or something like the above.

types/enzyme/enzyme-tests.tsx
```diff
@@ -497,9 +521,9 @@ function ShallowWrapperTest() {
 // ReactWrapper
 function ReactWrapperTest() {
     let reactWrapper: ReactWrapper<MyComponentProps, MyComponentState> =
-        mount<MyComponentProps, MyComponentState>(<MyComponent stringProp="value" numberProp={1} />);
+        mount<MyComponentProps, MyComponentState>(<MyComponent stringProp="value" numberProp={1} functionProp={() => true} />);
 
-    let reactElement: ReactElement;
+            let reactElement: ReactElement;
```
jwbay: Please revert indentation bumps like this as well

jwbay (changes requested):
(Not sure why it marked this as approved)

rgibson518:
> (Not sure why it marked this as approved)

Me either, lol..

types/enzyme/enzyme-tests.tsx
rgibson518: Is there a formatter for DefinitelyTyped? Does the dtslinter have a --fix option or is there a prettier config? I was looking but didnt' see anything.

rgibson518:
After some playing around with the function key selector portion of the invoke signature, I was able to get optional and required functions, with varying signatures (w/o using Function), while not getting optional non-function props.  I think this make the most sense. That said I think this is GTG unless of course there is more feedback!

types/enzyme/enzyme-tests.tsx
jwbay: Not that I know of :(

jwbay (approved):

cherealnice:
Any chance this gets merged soon? Currently blocking a project I'm working on, thanks!

rgibson518:
@cherealnice Beyond merging, this will have to get published to @types/eznyme I think under the 3.10 tag.  However, I'm pretty sure if this PR is doesn't encompass everything needed to move to 3.10.  In the meantime you can use [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html) if you're blocked. 

Add this to your project: 
```typescript
import { Component } from 'react';

declare module 'enzyme' {
  interface ShallowWrapper<P = {}, S = {}, C = Component> {
    /**
     * Invokes a function prop.
     * @param invokePropName The function prop to call.
     * @param ...args The argments to the invokePropName function
     * @returns The value of the function.
     */
    invoke<
      K extends NonNullable<
        { [K in keyof P]: P[K] extends ((...arg: any[]) => void) | undefined ? K : never }[keyof P]
      >
    >(
      invokePropName: K
    ): P[K];
  }
}
```

uniqueiniquity (changes requested):
@rgibson518 If you'd like this to be published as `3.10.0`, please bump the version at the top of `index.d.ts` to `3.10`. I'm not sure I follow your point about needing to add something to your code if you consume the package.

rgibson518:
> @rgibson518 If you'd like this to be published as `3.10.0`, please bump the version at the top of `index.d.ts` to `3.10`. I'm not sure I follow your point about needing to add something to your code if you consume the package.

@uniqueiniquity I bumped the version since this covers a function from enzyme@3.10, this PR is definitely not comprehensive to all the changes from enzyme@3.10, though.  Perhaps it's better to leave it as 3.9 until 3.10 is more comprehensively covered?

uniqueiniquity:
> @uniqueiniquity I bumped the version since this covers a function from enzyme@3.10, this PR is definitely not comprehensive to all the changes from enzyme@3.10, though. Perhaps it's better to leave it as 3.9 until 3.10 is more comprehensively covered?

That's fine; I think 3.10 is a better choice here.

uniqueiniquity (approved):

PR #36160, opened 2019-06-13
set-cookie-parser: add CookieMap support

hjr3:
Allow the parser result to return a CookieMap type.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/2634fd5ae8a654f7f8d40c4ab1883abb71b88bb1/types/set-cookie-parser/index.d.ts#L32

### DIFF
--- a/types/set-cookie-parser/index.d.ts
+++ b/types/set-cookie-parser/index.d.ts
@@ -9,10 +9,14 @@
 declare module "set-cookie-parser" {
     import http = require("http");
 
-    function SetCookieParser(input: string | ReadonlyArray<string> | http.IncomingMessage, options?: SetCookieParser.Options): SetCookieParser.Cookie[];
+    function SetCookieParser(input: string | ReadonlyArray<string> | http.IncomingMessage, options: SetCookieParser.Options & { map: true }): SetCookieParser.CookieMap;
+    function SetCookieParser(input: string | ReadonlyArray<string> | http.IncomingMessage, options?: SetCookieParser.Options & { map?: false }): SetCookieParser.Cookie[];
+    function SetCookieParser(input: string | ReadonlyArray<string> | http.IncomingMessage, options?: SetCookieParser.Options): SetCookieParser.Cookie[] | SetCookieParser.CookieMap;
 
     namespace SetCookieParser {
-        function parse(input: string | ReadonlyArray<string> | http.IncomingMessage, options?: Options): Cookie[];
+        function parse(input: string | ReadonlyArray<string> | http.IncomingMessage, options: Options & { map: true }): CookieMap;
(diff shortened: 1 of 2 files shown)


hjr3:
The `set-cookie-parser` type definitions only expose `Cookie[]` as return type for the parser. Here is what currently has to be done to use `set-cookie-parser` with the `{map: true}` option:

```ts
// make my own version
interface CookieMap {
  [s: string]: Cookie;
}

// do some type casting
const cookies = (SetCookieParser(response.header['set-cookie'], {
  decodeValues: true,
  map: true,
}) as unknown) as CookieMap;

// now i can use cookie.foo
```

Instead, I want to upstream `CookieMap` so the code can be simplified to:

```ts
// no custom type needed
// more straight-forward type cast
const cookies = SetCookieParser(response.header['set-cookie'], {
  decodeValues: true,
  map: true,
}) as CookieMap;

// now i can use cookie.foo
```

I did not bother changing the parser return type to `Cookie[] | CookieMap` as that looks to break backwards compatibility with the current type definition.

nickp10 (approved):

types/set-cookie-parser/index.d.ts
```diff
@@ -27,6 +27,10 @@ declare module "set-cookie-parser" {
             httpOnly?: boolean;
         }
 
+        interface CookieMap {
+          [s: string]: Cookie;
```
ilyaztsv: probably, using `name` instead of `s` as key name will be better for readability, because in case of `map: true` key is name of a cookie.

iztsv (approved):

iztsv (approved):
LGTM

rbuckton (changes requested):
I would suggest the following change:

```ts
function parse(input: string | ReadonlyArray<string> | http.IncomingMessage, 
  options: Options & { map: true }): CookieMap;
function parse(input: string | ReadonlyArray<string> | http.IncomingMessage, 
  options?: Options & { map?: false }): Cookie[];
function parse(input: string | ReadonlyArray<string> | http.IncomingMessage, 
  options?: Options): Cookie[] | CookieMap;
```

This way the correct type will be picked up automatically.

hjr3:
> I would suggest the following change:
> 
> ```ts
> function parse(input: string | ReadonlyArray<string> | http.IncomingMessage, 
>   options: Options & { map: true }): CookieMap;
> function parse(input: string | ReadonlyArray<string> | http.IncomingMessage, 
>   options?: Options & { map?: false }): Cookie[];
> function parse(input: string | ReadonlyArray<string> | http.IncomingMessage, 
>   options?: Options): Cookie[] | CookieMap;
> ```
> 
> This way the correct type will be picked up automatically.

Thanks! I was thinking of how to do this and it is obvious in retrospect.

rbuckton (approved):

PR #36162, opened 2019-06-13
[pikaday] Remove `moment` from dependency of @types/pikaday

### LINKED ISSUE #27942: @types/pikaday: Moment in pikaday
- [x] I tried using the `@types/pikaday` package and had problems.

Authors: @MidnightDesign, @wake42

`@types/pikaday` has `moment` in dependencies although it is optional in original `pikaday` as well as removed from `package.json` (dbushell/Pikaday#718)

PikachuEXE:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/Pikaday/Pikaday/issues/718 [X] Increase the version number in the header if appropriate.

Pikaday 1.8 does **not** require `moment`
So using this typing package right now includes huge `moment` package in produced bundle as mentioned in #27942
Fixes #27942

Best fix is to have separate typing package for `moment` and reference it (without including the actual code)

### DIFF
--- a/types/pikaday/index.d.ts
+++ b/types/pikaday/index.d.ts
@@ -1,8 +1,9 @@
-// Type definitions for pikaday 1.7
+// Type definitions for pikaday 1.8
 // Project: https://github.com/dbushell/Pikaday, https://pikaday.com
 // Definitions by: Rudolph Gottesheim <https://github.com/MidnightDesign>
 //                 Åke Wivänge <https://github.com/wake42>
 //                 Istvan Mezo <https://github.com/mezoistvan>
+//                 PikachuEXE <https://github.com/PikachuEXE>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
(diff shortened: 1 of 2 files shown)


MidnightDesign (commented):
I don't really know what to do here. Changing a return type hint to `any` seems like a bad idea. How does the library behave if Moment isn't loaded?

MidnightDesign:
Can't you just change the dependency to `@types/moment`?

PikachuEXE:
@MidnightDesign 
I agree that's a better solution
So I started #36252 
But there is a package `@types/moment` already which is *deprecated*
image
Unless the deprecation is reverted I am not sure if I should use it

PikachuEXE:
If I just point to @types/moment in `package.json`, this error shows up:
```
For example, if package *P* used to have types on Definitely Typed at @types/P,
but now has its own types, a dependent package *D* will need to use package.json
to refer to @types/P if it relies on old versions of P's types.
In this case, please make a pull request to types-publisher adding @types/P to `dependenciesWhitelist.txt`.
```

PikachuEXE:
Closing for #36252

PR #36164, opened 2019-06-13
yet other fixes for swagger types

XVincentX:
It turns out there was a lot more missing for Swagger — this PR should hopefully close the gap once forever.


philsturgeon (approved):
😅

XVincentX:
@mohsen1 Mind checking this out?

types/swagger-schema-official/index.d.ts
```diff
@@ -129,11 +150,12 @@ export interface Response {
...
 
 // ------------------------------ Schema -------------------------------------
-export interface BaseSchema {
+export type BaseSchema = {
+  type?: ParameterType;
   format?: string;
   title?: string;
   description?: string;
-  default?: string | boolean | number | {};
+  default?: any[] | boolean | number | null | object | string;
```
mohsen1: is it `any` at this point?

mohsen1 (approved):

types/swagger-schema-official/index.d.ts
XVincentX: Hmm I'm against any in general, I would say I was too generous and I should check it out again and narrow it down, although it can really be anything here. I'll check it out again.

XVincentX:
@mohsen1 Although I think the correct type for that things is probably `unknown` I decided to stick with `any` as you suggested. Maybe we can iterate once more when I'll use this stuff more seriously.

PR #36170, opened 2019-06-13
add notifme-template typings

ziv:
Please fill in this template.

If adding a new definition:
- [ ] Create it with `dts-gen --dt`, not by basing it on an existing project. **created from module written in Babel using it's Flow output**

### DIFF
--- a/types/notifme-template/index.d.ts
+++ b/types/notifme-template/index.d.ts
@@ -1,0 +1,128 @@
+// Type definitions for notifme-template
+// Project: https://github.com/notifme/notifme-template
+// Definitions by: Ziv <https://github.com/ziv>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped/
+// TypeScript Version: 3.2
+
+/// <reference types="node" />
+
+declare module 'notifme-template' {
+
+    export default function RenderFactory(renderer: RendererType, folder: string, options?: OptionsType): RendererFunction;
+
+    export type  RendererFunction = (templateName: string, data: object, lang?: string) => Promise<TemplateType>;
+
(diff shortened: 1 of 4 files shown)


DanielRosenwasser:

types/notifme-template/index.d.ts
```diff
@@ -0,0 +1,125 @@
+// Type definitions for notifme-template 1.0
+// Project: https://github.com/notifme/notifme-template
+// Definitions by: Ziv <https://github.com/ziv>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped/
+// TypeScript Version: 3.2
+
+/// <reference types="node" />
+
+export default function RenderFactory(renderer: RendererType, folder: string, options?: OptionsType): RendererFunction;
```
uniqueiniquity: It seems like the [underlying library](https://github.com/notifme/notifme-template/blob/master/index.js) is not using `export default`, so this .d.ts file should not either. Instead, you should follow the appropriate template from [the TypeScript handbook](http://www.typescriptlang.org/docs/handbook/declaration-files/templates.html). Note that if you want to use `import name from 'package';` in your consuming code, you can use the `esModuleInterop` compiler setting.

uniqueiniquity (changes requested):

ziv (commented):
seems right

types/notifme-template/index.d.ts
uniqueiniquity: I'm not sure why you marked this as resolved; is there something I'm missing about why this .d.ts should be using `export default`? Please let me know if there's any part I should explain more.

PR #36172, opened 2019-06-13
[parse] Add generic support for (database) models to Parse.Object

dirkgroenen:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: _There's no such documentation since this is Typescript specific_

**Additional info:**
When working with Parse.Object's it's currently not possible to define the actual database model. This means Intellisense can't provide us with meaningful keys when using methods like `Query.equalTo` or `Object.get`.  

This change allows a Parse.Object to be provided with a generic interface representing the actual (database) model:
```typescript
interface GameScoreModel {
    arrayKey: number[];
    score: number;
    difficult: number;
    skills: string[];
}

class GameScore extends Parse.Object<GameScoreModel> {
    constructor(options?: any) {
        super("GameScore", options);
    }
}
```

Intellisense will now provide you with meaningful column names when calling Object methods like `get()` or query methods like `equalTo`.

image
image


types/parse/index.d.ts
```diff
@@ -272,31 +278,32 @@ declare namespace Parse {
...
-        static createWithoutData<T extends Object>(id: string): T;
-        static destroyAll<T extends Object>(list: T[], options?: Object.DestroyAllOptions): Promise<T[]>;
+        static createWithoutData<T extends Object>(this: { new(): T }, id: string): T;
+        static destroyAll<T extends Object>(this: { new(): T }, list: T[], options?: Object.DestroyAllOptions): Promise<T[]>;
         static extend(className: string | { className: string }, protoProps?: any, classProps?: any): any;
         static fetchAll<T extends Object>(list: T[], options: Object.FetchAllOptions): Promise<T[]>;
         static fetchAllIfNeeded<T extends Object>(list: T[], options: Object.FetchAllOptions): Promise<T[]>;
         static fetchAllWithInclude<T extends Object>(list: T[], keys: string | Array<string | Array<string>>, options: RequestOptions): Promise<T[]>;
-        static fromJSON(json: any, override?: boolean): Object;
+        static fromJSON<T>(this: { new(): T }, json: any, override?: boolean): T;
```
yagotome: What about `json`'s type as the `K` for the returned `Object<K>`?

```suggestion
        static fromJSON<T extends Object<K>, K>(this: { new(): T }, json: K, override?: boolean): T;
```

yagotome (commented):
@dirkgroenen Why not generics for `User` and `Session` too?

types/parse/index.d.ts
dirkgroenen: It would indeed make sense to add those as well. 
Unfortunately I'm struggling with the User type:

Adding the generic to User and passing it along to Object causes `Query` to infer the type incorrectly. 
```typescript
class User<K = any> extends Object<K> { }
class Query<T extends Object, K = Unobject<T>> extends BaseObject { }
```
causes the following to fail:
```typescript
const queryB = new Parse.Query(Parse.User);
queryB.near("a", userGeoPoint); // Argument of type '"a"' is not assignable to parameter of type 'never'.
```
The type for `queryB` was set to `const queryB: Parse.Query<Parse.User<{}>, {}>`, where `Parse.Query<Parse.User<any>, any>` was expected.

RyanCavanaugh:
@yagotome :+1: or :-1: ?

types/parse/index.d.ts
yagotome: @dirkgroenen, sorry for the long time without replying.

The issue you pointed out is also a problem for `Parse.Object`. The problem is in the inference of `T` by the type of the constructor's parameter `objectClass`. If you do `const queryB = new Parse.Query(Parse.Object);`, you will see that the `typeof queryB` is also wrong (`Parse.Query<Parse.Object<unknown>, unknown>`).

I am not sure why, but if you do `const queryB = new Parse.Query(typeof ({} as Parse.Object));` it works. Another option that works is `const queryB = new Parse.Query<Parse.Object>(Parse.Object);`. However, a possible solution for this would be changing the constructor of `Parse.Query`. I am going to send you a suggestion for that in a new thread.

But what about the suggested change I proposed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36172#discussion_r294058705?

types/parse/index.d.ts
```diff
@@ -458,9 +465,9 @@ declare namespace Parse {
      *   }
      * });</pre></p>
      */
-    class Query<T extends Object = Object> extends BaseObject {
+    class Query<T = Object, K extends Unobject<T> = Unobject<T>> extends BaseObject {
```
yagotome: Why removing `extends Object`?

```suggestion
    class Query<T extends Object = Object, K extends Unobject<T> = Unobject<T>> extends BaseObject {
```

types/parse/index.d.ts
```diff
@@ -458,9 +465,9 @@ declare namespace Parse {
...
 
-        objectClass: any;
+        objectClass: T;
         className: string;
 
         constructor(objectClass: string);
```
yagotome: @dirkgroenen  The other constructor `constructor(objectClass: new (...args: any[]) => T);` could be `constructor(objectClass: new (...args: any[]) => Object);`. In this way, it is possible to use `new Parse.Query(Parse.Object)`, although it is no more able to infer the generic type from constructor. It would be necessary to do `new Parse.Query<MyObj>(MyObj)` if I have a `class MyObj extends Object<{ foo: string }> {}` for example.

What do you think?

types/parse/index.d.ts
dirkgroenen: Good question, not sure what happened over there. I've reverted it.

types/parse/index.d.ts
dirkgroenen: The `fromJson` suggestion doesn't seem to work correctly. Using `Unobject` does give us the desired result:
```
static fromJSON<T extends Object<K>, K>(this: { new(): T }, json: K, override?: boolean): T;
```
image

types/parse/index.d.ts
dirkgroenen: Giving up on our inferred type feels a bit like giving up one of the better improvements in this PR. I'm eager on finding the reason, or a better solution, which will solve the problems we're having. 

I have a few ideas which I would like to try before falling back to manually providing our type. I'll try to come back on this thread this week.

weswigham:
@yagotome 👍 or 👎 ?

dirkgroenen:
@weswigham @yagotome Due holidays and all kinds of other stuff I haven't yet found the time to take a look out the current outstanding issues. 

Given the current state of this PR isn't really 'publishable' I suggest we close it for now and come back on it after I've made the appropriate changes?

weswigham:
> Given the current state of this PR isn't really 'publishable' I suggest we close it for now and re-open it after I've made the appropriate changes?

You may do so, if that is your desire.

yagotome:
@dirkgroenen no worries! Unfortunately I also haven't found the time to dive into this problem. I wish I will be able to help with those changes then.

dirkgroenen:
Well let's keep it open then, for ourselves and others to see. 
I'm not going to be able to work on this in the next few weeks, but will definitely try to take a look at it asap.

dirkgroenen:
@yagotome  it has been a while now, but I finally pushed some changes which, if I remember correctly, address your concerns:
- Added generics to Session and User, including default attributes
- Fixed `Unobject` to return `any` when no types were defined, preventing `never` from being the `K` in `set()`-like methods. This will also allow `new Parse.Query(Parse.Object);`, see parse-tests.ts.

types/parse/index.d.ts
```diff
@@ -79,6 +80,11 @@ declare enum ErrorCode {
 
 declare namespace Parse {
 
+    // Helper types for extracting an extended generic type from an
+    // (Parse.)Object or Array.
+    type Unarray<T> = T extends Array<infer U> ? U : T;
+    type Unobject<X> = X extends Object<infer U> ? U extends any ? any : U : X;
```
yagotome: `Unobject` of something that isn't an `Object` should `never` happen.

```suggestion
    type Unobject<X> = X extends Object<infer U> ? U : never;
```

types/parse/index.d.ts
```diff
@@ -79,6 +80,11 @@ declare enum ErrorCode {
 
 declare namespace Parse {
 
+    // Helper types for extracting an extended generic type from an
+    // (Parse.)Object or Array.
+    type Unarray<T> = T extends Array<infer U> ? U : T;
```
yagotome: Similarly.
```suggestion
    type Unarray<T> = T extends Array<infer U> ? U : never;
```

types/parse/index.d.ts
yagotome: Anything `extends` `any`, doesn't it? So, if `X extends Object` this would return `any`. That's why that (following) test passed. The problem isn't solved yet :confused: 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f1a13f991a2d4abab3f8fd3e0a3082160af85743/types/parse/parse-tests.ts#L665-L666

yagotome (changes requested):

types/parse/index.d.ts
dirkgroenen: I agree, except that I'm (mis)using the side-effect of a non-unarrayable type to fallback to it's given type by default. 

I've modified Unarray and Unobject to fallback to `never` by default, unless provided otherwise.

types/parse/index.d.ts
dirkgroenen: Ah you're right, obviously. 

The `new Parse.Query(Parse.Object)` is giving me some real headaches... I've pushed a new draft which adds an `UnpackKeys` type to  hopefully solves these 'never' and 'unknown' cases.

rommyarb:
> Please fill in this template.
> 
> * [x]  Use a meaningful title for the pull request. Include the name of the package modified.
> * [x]  Test the change in your own code. (Compile and run.)
> * [x]  Add or edit tests to reflect the change. (Run with `npm test`.)
> * [x]  Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
> * [x]  Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
> * [x]  Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).
> 
> Select one of these and delete the others:
> 
> If changing an existing definition:
> 
> * [x]  Provide a URL to documentation or source code which provides context for the suggested changes: _There's no such documentation since this is Typescript specific_
> * [x]  Increase the version number in the header if appropriate.
> * [x]  If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.
> 
> **Additional info:**
> When working with Parse.Object's it's currently not possible to define the actual database model. This means Intellisense can't provide us with meaningful keys when using methods like `Query.equalTo` or `Object.get`.
> 
> This change allows a Parse.Object to be provided with a generic interface representing the actual (database) model:
> 
> ```ts
> interface GameScoreModel {
>     arrayKey: number[];
>     score: number;
>     difficult: number;
>     skills: string[];
> }
> 
> class GameScore extends Parse.Object<GameScoreModel> {
>     constructor(options?: any) {
>         super("GameScore", options);
>     }
> }
> ```
> 
> Intellisense will now provide you with meaningful column names when calling Object methods like `get()` or query methods like `equalTo`.
> 
> image
> image

I've been searching my whole life how to do this. Thank you! Thank you! Thank you! ❤️❤️❤️

PR #36186, opened 2019-06-13

regevbr:
Please fill in this template.

To continue #35977

@jgonggrijp I just noticed that I didn't take care of the typings for the chain and wrapped object use cases.
I fixed the code, but my tests failed, then I realized that the Generic type of Underscore<T> and Chain<T> when using a dictionary, is the union of the value types of the dictionary - which can be completely wrong for some methods, since, for example - pick is operating on the type of the object and not its value types.

I'll give an example in order to clarify the issue:

When using the current static wrapped "constructor"
```typescript
        <T>(value: _.Dictionary<T>): Underscore<T>;
```
You have no way to describe Pick on it, as you are loosing the type of the object itself which is essential here...

I added non breaking changes to support the use case, but in time more methods will need to be revised as there are a lot of wrongly types methods in the chain process, at they can probably be better handled now with the new typing.

I did however had to change the typescript version to 2.8 which caused a chain reaction of changing the typescript version for 18 more dependent libraries. I'm not sure what I did is correct, but I increased the major version and updated the typescript version in all related libraries.

as to the following complicated line:

```typescript
        <T extends TypeOfDictionary<V>, V extends _.Dictionary<any> = _.Dictionary<T>>(value: V): Underscore<T, V>;
```

This was the only way I could get Typescript to properly infer T. If using the obvious typing

```typescript
        <T, V extends _.Dictionary<T> = _.Dictionary<T>>(value: V): Underscore<T, V>;
```

The last test fails as it infers T as `{}`


regevbr:
closing in favor of #36230

PR #36193, opened 2019-06-14
@types/react Event add event.target.xxx

kszitt:
### Reproduction problem
``` javascript
 <p data-type="p" onClick={e => {this.click(e)}}>click</p>
...
  click(e: MouseEvent<HTMLDivElement>): void {
    let target = e.target;

    console.log(target);
    console.log(target.innerHTML);  // Error:(46, 24) TS2339: Property 'innerHTML' does not exist on type 'EventTarget'.
    console.log(target.getAttribute("data-type"));  // Error:(48, 24) TS2339: Property 'getAttribute' does not exist on type 'EventTarget'.
  }
```

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -1059,7 +1059,7 @@ declare namespace React {
      * This might be a child element to the element on which the event listener is registered.
      * If you thought this should be `EventTarget & T`, see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/12239
      */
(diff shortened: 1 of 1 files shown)


Jessidhia:
It's intentionally made like this for events that bubble; non-bubbling ones like `ChangeEvent` do get `& T` on their `target`.

This is because the `target` is not necessarily the element the handler is attached to, it could be any of its children elements.

eps1lon:
Two lines above the change is a link that gives you more context why this is typed this way.

PR #36204, opened 2019-06-14
[mailgun-js] Add missing SendData typings

ZhangYiJiang:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: Additional sending options from: https://documentation.mailgun.com/en/latest/api-sending.html#sending. I tried adding `[key: string]: string` for `v:` and `h:` keys, but TS didn't like that, so I just encoded the most common headers. Using `boolean` is valid for some of the boolean params because mailgun-js casts booleans to string first. 

### DIFF
--- a/types/mailgun-js/index.d.ts
+++ b/types/mailgun-js/index.d.ts
@@ -61,13 +61,30 @@ declare namespace Mailgun {
         interface SendData {
             from?: string;
             to: string | string[];
-            cc?: string;
-            bcc?: string;
+            cc?: string | string[];
+            bcc?: string | string[];
             subject?: string;
             text?: string;
             html?: string;
+            'amp-html'?: string;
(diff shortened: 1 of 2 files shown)


types/mailgun-js/index.d.ts
```diff
@@ -61,13 +61,30 @@ declare namespace Mailgun {
...
+            bcc?: string | string[];
             subject?: string;
             text?: string;
             html?: string;
+            'amp-html'?: string;
             attachment?: AttachmentData | ReadonlyArray<AttachmentData>;
             inline?: AttachmentData | ReadonlyArray<AttachmentData>;
+
+            // Mailgun options
+            'o:tag'?: string[];
```
epiphone: To be exact this should be `'o:tag'?: string | string[];`, no?

PR #36207, opened 2019-06-14
Add typings for npm module uid2

levibostian:
Please fill in this template.

If adding a new definition:

---

Note: The npm module these typings are created for only uses patch as version [releases](https://github.com/coreh/uid2/releases). Therefore, I had to disable the dt-header rule for tslint as it was not liking that I included the patch version in my header of the index.d.ts file.

### DIFF
--- a/types/uid2/index.d.ts
+++ b/types/uid2/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for uid2 0.3
+// Type definitions for uid2 0.0
 // Project: https://github.com/coreh/uid2
 // Definitions by: Levi Bostian <https://github.com/levibostian>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
@@ -6,4 +6,4 @@
 declare function uid2(length: number): string;
 declare function uid2(length: number, callback: (err: Error | null, result?: string) => void): void;
 
-export default uid2;
+export = uid2;
--- a/types/uid2/index.d.ts
+++ b/types/uid2/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for uid2 0.3.3
+// Type definitions for uid2 0.3
 // Project: https://github.com/coreh/uid2
 // Definitions by: Levi Bostian <https://github.com/levibostian>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
--- a/types/uid2/index.d.ts
+++ b/types/uid2/index.d.ts
@@ -3,6 +3,7 @@
 // Definitions by: Levi Bostian <https://github.com/levibostian>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export default function(length: number, callback?: Callback): string;
+declare function uid2(length: number): string;
+declare function uid2(length: number, callback: (err: Error | null, result?: string) => void): void;
 
-export type Callback = (err: Error | null, result?: string) => void;
+export default uid2;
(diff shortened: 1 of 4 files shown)


types/uid2/index.d.ts
```diff
@@ -0,0 +1,6 @@
+// Type definitions for uid2 0.3.3
+// Project: https://github.com/coreh/uid2
+// Definitions by: Levi Bostian <https://github.com/levibostian>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export default function(length: number): string;
```
rbuckton: It looks like uid2 also supports an overload with a callback:

https://github.com/coreh/uid2/blob/de1b40992074e01983dd1d145af369eb8bc41137/index.js#L31-L37

rbuckton (changes requested):

types/uid2/index.d.ts
```diff
@@ -0,0 +1,8 @@
+// Type definitions for uid2 0.3.3
+// Project: https://github.com/coreh/uid2
+// Definitions by: Levi Bostian <https://github.com/levibostian>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export default function(length: number, callback?: Callback): string;
```
rbuckton: I would suggest this change:

```suggestion
declare function uid2(length: number): string;
declare function uid2(length: number, callback: (err: Error | null, result?: string) => void): void;
export default uid2;
```

Along with removing the extranous `Callback` type.

levibostian: Thank you for your suggest. I can see a few ways in which your suggestion improves the typings. 

However, I do not understand the reasoning behind using 2 function declarations instead of 1 with the callback parameter as optional? I decided to use 1 function after reading [this part](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#use-optional-parameters) of the do's and don'ts guide in the handbook.

rbuckton: The signatures I proposed differ in *more* than just trailing parameters, they also differ in return types. The underlying function at runtime will return a `string` if you do *not* pass a callback, and returns nothing if you *do* pass a callback.

levibostian: Ah, yes. Thank you very much for pointing that out for me.

levibostian:
Not sure why this PR is currently labeled, "Needs Author Attention" in the project board as my most recent activity on this PR is a commit. 

Either way, this PR is ready for another review.

levibostian:
The requested changes have been resolved. 

This PR is ready for another review, @rbuckton. Thanks!

types/uid2/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {        
+        "dt-header": false
```
rbuckton: Is there a reason not to enforce `dt-header`?

levibostian: This is similar to [another PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36211) I made for another npm module. 

For reference, here is the conversation from that other PR I made. This conversation goes over why I disabled this rule in this PR. 

2019-07-18 at 9 32 AM

I have made this same change in this PR now. I changed the version number for the uid2 package to 0.0 from 0.0.3 and then re-enabled the lint rule.

rbuckton (approved):

PR #36211, opened 2019-06-14
Add typings for connect-trim-body npm module

levibostian:
Please fill in this template.

If adding a new definition:

---

Note: The npm module these typings are created for only uses patch as version releases. Therefore, I had to disable the dt-header rule for tslint as it was not liking that I included the patch version in my header of the index.d.ts file.

### DIFF
--- a/types/connect-trim-body/tslint.json
+++ b/types/connect-trim-body/tslint.json
@@ -1,6 +1,3 @@
 {
-    "extends": "dtslint/dt.json",
-    "rules": {
-        "dt-header": false
-    }
+    "extends": "dtslint/dt.json"
 }
--- a/types/connect-trim-body/tslint.json
+++ b/types/connect-trim-body/tslint.json
@@ -1,0 +1,6 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "dt-header": false
+    }
+}
--- a/types/connect-trim-body/connect-trim-body-tests.ts
+++ b/types/connect-trim-body/connect-trim-body-tests.ts
@@ -1,0 +1,6 @@
+import connectTrimBody from 'connect-trim-body';
+import express from 'express';
+
+const app = express();
+
+app.use(connectTrimBody());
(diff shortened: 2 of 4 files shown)


types/connect-trim-body/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "dt-header": false
```
uniqueiniquity: Please remove the patch version in your header and re-enable this rule. Thanks!

uniqueiniquity (changes requested):

types/connect-trim-body/tslint.json
uniqueiniquity: Follow-up: I saw your comment in the original post. Unfortunately, a new patch version of this package needs to be released every time a change is made, so you cannot specify the patch version in the header.

levibostian: When you say, "the patch version in the header", are you referring to this code? 

```
// Type definitions for connect-trim-body 0.0.3
```

The `0.0.3` part? That version is referring to the [version of the npm package](https://github.com/samora/connect-trim-body/blob/master/package.json#L3), not these typings? 

Are you saying that we cannot merge this PR because the npm package `connect-trim-body` decided to use only patch versions for releases instead of a major/minor version? The project has not been updated in years. I doubt that I could get the author to release a new version. 

What do you recommend we do here?

levibostian:
I have replied to your questions, @uniqueiniquity. Thank you.

types/connect-trim-body/tslint.json
uniqueiniquity: I'd suggest just leaving it as `// Type definitions for connect-trim-body 0.0`.
This package will then be initially published as `@types/connect-trim-body@0.0.0` and will bump the patch version with each change.

levibostian: Got it. Done.

uniqueiniquity (approved):

PR #36216, opened 2019-06-15
Add types for places.js (Algolia Places)

samjetski:
Please fill in this template.

If adding a new definition:

---

My first TS type definition, any feedback welcome.

### DIFF
--- a/types/places.js/index.d.ts
+++ b/types/places.js/index.d.ts
@@ -14,7 +14,14 @@ export = places;
 declare function places(options: places.PlacesStaticOptions): places.PlacesAutocomplete;
 
 declare namespace places {
-    type PlaceType = "country" | "city" | "address" | "busStop" | "trainStation" | "townhall" | "airport";
+    type PlaceType
+        = "country"
+        | "city"
+        | "address"
+        | "busStop"
+        | "trainStation"
+        | "townhall"
+        | "airport";
 
     interface PlaceLatLng {
         lat: number;
--- a/types/places.js/index.d.ts
+++ b/types/places.js/index.d.ts
@@ -14,16 +14,7 @@ export = places;
 declare function places(options: places.PlacesStaticOptions): places.PlacesAutocomplete;
 
 declare namespace places {
-    // tslint:disable-next-line:no-const-enum
-    const enum PlaceType {
-        Country = "country",
-        City = "city",
-        Address = "address",
-        BusStop = "busStop",
-        TrainStation = "trainStation",
-        Townhall = "townhall",
-        Airport = "airport"
-    }
+    type PlaceType = "country" | "city" | "address" | "busStop" | "trainStation" | "townhall" | "airport";
 
     interface PlaceLatLng {
         lat: number;
--- a/types/places.js/index.d.ts
+++ b/types/places.js/index.d.ts
@@ -14,7 +14,8 @@ export = places;
 declare function places(options: places.PlacesStaticOptions): places.PlacesAutocomplete;
 
 declare namespace places {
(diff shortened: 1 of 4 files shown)


types/places.js/index.d.ts
```diff
@@ -0,0 +1,340 @@
...
+import { EventEmitter } from 'events';
+
+export as namespace places;
+
+export = places;
+
+declare function places(options: places.PlacesStaticOptions): places.PlacesAutocomplete;
+
+declare namespace places {
+    enum PlaceType {
```
rbuckton:
```suggestion
    const enum PlaceType {
```

`places` doesn't actually have an exported *value* named `PlaceType`, so callers would be broken if they wrote `PlaceType.City` since `PlaceType` is `undefined`.

types/places.js/index.d.ts
```diff
@@ -0,0 +1,340 @@
...
+         */
+        useDeviceLocation?: boolean;
+
+        /**
+         * Allows you to override computed query parameters sent to the {@link https://community.algolia.com/places/rest.html | Algolia Places REST API}.
+         */
+        computeQueryParams?: algoliasearch.QueryParameters;
+    }
+
+    class PlacesAutocomplete extends EventEmitter {
```
rbuckton:
```suggestion
    interface PlacesAutocomplete extends EventEmitter {
```

`places` doesn't export the class, so there's no actual *value* name `PlacesAutocomplete`

rbuckton (changes requested):

types/places.js/index.d.ts
samjetski: Ah cheers, that makes a lot of sense! The default dtslint rules appear to disallow `const enum`. I'll put an exception, but is that supposed to be so?

samjetski:
@rbuckton Thanks for the feedback, let me know if any other changes are needed.

types/places.js/index.d.ts
```diff
@@ -14,7 +14,8 @@ export = places;
 declare function places(options: places.PlacesStaticOptions): places.PlacesAutocomplete;
 
 declare namespace places {
-    enum PlaceType {
+    // tslint:disable-next-line:no-const-enum
+    const enum PlaceType {
```
sheetalkamat: This shouldnt be enum.. This should be union of type literals instead

sheetalkamat (changes requested):

types/places.js/index.d.ts
samjetski: Thanks @sheetalkamat, should be fixed now.

sheetalkamat (changes requested):

types/places.js/places.js-tests.ts
```diff
@@ -0,0 +1,110 @@
+import * as places from 'places.js';
+// tslint:disable-next-line:no-duplicate-imports
```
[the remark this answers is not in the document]
samjetski: Got it!

samjetski:
Hi @sheetalkamat your requested changes have been made, is this PR ready to progress?

aub:
It would be great to get this in!

It seems worth noting that there are already definitions under [types/places](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/places) that seem clearly intended to support this same library. They're much less complete and also can't be used because the name of the package doesn't match the name of the library (places.js vs places) so it seems like they should be removed.

Haroenv:
cc @JonathanMontane

orta:
OK great, looks like these are shipping with the library now. Closing.

PR #36230, opened 2019-06-16

regevbr:
Please fill in this template.

To continue #35977

@jgonggrijp I just noticed that I didn't take care of the typings for the chain and wrapped object use cases.
I fixed the code, but my tests failed, then I realized that the Generic type of Underscore<T> and Chain<T> when using a dictionary, is the union of the value types of the dictionary - which can be completely wrong for some methods, since, for example - pick is operating on the type of the object and not its value types.

I'll give an example in order to clarify the issue:

When using the current static wrapped "constructor"
```typescript
        <T>(value: _.Dictionary<T>): Underscore<T>;
```
You have no way to describe Pick on it, as you are loosing the type of the object itself which is essential here...

I added non breaking changes to support the use case, but in time more methods will need to be revised as there are a lot of wrongly types methods in the chain process, at they can probably be better handled now with the new typing.

I did however had to change the typescript version to 2.8 which caused a chain reaction of changing the typescript version for 18 more dependent libraries. I'm not sure what I did is correct, but I increased the major version and updated the typescript version in all related libraries.

as to the following complicated line:

```typescript
        <T extends TypeOfDictionary<V>, V extends _.Dictionary<any> = _.Dictionary<T>>(value: V): Underscore<T, V>;
```

This was the only way I could get Typescript to properly infer T. If using the obvious typing

```typescript
        <T, V extends _.Dictionary<T> = _.Dictionary<T>>(value: V): Underscore<T, V>;
```

The last test fails as it infers T as `{}`

### DIFF
--- a/types/backbone-associations/index.d.ts
+++ b/types/backbone-associations/index.d.ts
@@ -1,8 +1,8 @@
-// Type definitions for Backbone-associations 0.6.4
+// Type definitions for Backbone-associations 0.7.0
 // Project: http://dhruvaray.github.io/backbone-associations
 // Definitions by: Craig Brett <https://github.com/craigbrett17>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+// TypeScript Version: 2.8
 
 import * as Backbone from 'backbone';
 
--- a/types/backbone-associations/index.d.ts
+++ b/types/backbone-associations/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for Backbone-associations 0.7.0
+// Type definitions for Backbone-associations 0.6
 // Project: http://dhruvaray.github.io/backbone-associations
 // Definitions by: Craig Brett <https://github.com/craigbrett17>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
--- a/types/underscore/index.d.ts
+++ b/types/underscore/index.d.ts
@@ -4115,7 +4115,7 @@ declare module _ {
         chain<T extends {}>(obj: T): _Chain<T>;
     }
 
-    interface Underscore<T, V> {
+    interface Underscore<T, V = T> {
 
         /* *************
         * Collections *
--- a/types/underscore/index.d.ts
+++ b/types/underscore/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for Underscore 2.0
+// Type definitions for Underscore 1.9
 // Project: http://underscorejs.org/
 // Definitions by: Boris Yankov <https://github.com/borisyankov>,
 //                 Josh Baldwin <https://github.com/jbaldwin>,
--- a/types/underscore/index.d.ts
+++ b/types/underscore/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for Underscore 1.8
+// Type definitions for Underscore 2.0
 // Project: http://underscorejs.org/
 // Definitions by: Boris Yankov <https://github.com/borisyankov>,
 //                 Josh Baldwin <https://github.com/jbaldwin>,
@@ -8,7 +8,7 @@
 //                 Florian Keller <https://github.com/ffflorian>
 //                 Regev Brody <https://github.com/regevbr>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+// TypeScript Version: 2.8
 
 declare var _: _.UnderscoreStatic;
 export = _;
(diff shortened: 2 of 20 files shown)


jgonggrijp (changes requested):
You bumped lots of package versions because of the TypeScript version bump. While I agree that's technically a breaking change, the DefinitelyTyped readme seems to indicate it is not a valid reason to bump the version. In fact, the version comment of an `index.d.ts` should always only include `major.minor` (so no `.patch`) and the `major.minor` should reflect the actual version of the corresponding library that it provides typings for:

https://github.com/DefinitelyTyped/DefinitelyTyped#i-want-to-use-features-from-typescript-21-or-above

https://github.com/DefinitelyTyped/DefinitelyTyped#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library (see especially the red exclamation mark at the bottom of this section)

I have added change suggestions in all places where you did a version bump, showing what I believe should be the actual version. Please hold off following up on them until @sandersn has replied to my question below, because I'm only 95% sure. Also, keep in mind that if you accept my change suggestions through the GitHub web interface, each one will show up as an individual commit with both of us as authors and a rather undescriptive commit message. I'm fine with it if you follow up on them with a single commit that has only you as the author.

@sandersn Could you please check that what I'm writing about package versions above is correct?

Other than that, there is one breaking change that I commented on, but I trust the rest of your changes.

types/underscore/index.d.ts
```diff
@@ -104,16 +104,18 @@ declare module _ {
...
         * as the first parameter can be invoked through this function.
         * @param key First argument to Underscore object functions.
         **/
-        <T>(value: _.Dictionary<T>): Underscore<T>;
-        <T>(value: _.List<T>): Underscore<T>;
-        <T>(value: Array<T>): Underscore<T>;
-        <T>(value: T): Underscore<T>;
+        <T>(value: _.List<T>): Underscore<T, _.List<T>>;
+        <T>(value: Array<T>): Underscore<T, Array<T>>;
+        <T extends TypeOfDictionary<V>, V extends _.Dictionary<any> = _.Dictionary<T>>(value: V): Underscore<T, V>;
```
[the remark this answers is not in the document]
regevbr: haha yes I was surprised as well :-)

types/underscore/index.d.ts
```diff
@@ -5068,7 +5063,7 @@ declare module _ {
         value<TResult>(): TResult;
     }
 
-    interface _Chain<T> {
+    interface _Chain<T, V = T> {
```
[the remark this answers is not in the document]
regevbr: sure sounds good

types/underscore/index.d.ts
```diff
@@ -4115,12 +4110,12 @@ declare module _ {
...
         **/
-        chain<T>(obj: T[]): _Chain<T>;
-        chain<T>(obj: _.Dictionary<T>): _Chain<T>;
+        chain<T>(obj: T[]): _Chain<T, T[]>;
+        chain<T extends TypeOfDictionary<V>, V extends _.Dictionary<any> = _.Dictionary<T>>(obj: V): _Chain<T, V>;
         chain<T extends {}>(obj: T): _Chain<T>;
     }
 
-    interface Underscore<T> {
+    interface Underscore<T, V> {
```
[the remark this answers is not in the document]
regevbr: I agree, will update the code accordingly

regevbr:
@jgonggrijp thanks for reviewing.

I added a fix for the breaking change.

Indeed it seems that I should revert the versions changes, but let's indeed wait for @sandersn to give his answer on the matter.

sandersn:
@jgonggrijp You are correct. People who are using old versions of typescript should be using `@types/underscore@ts2.6` to acquire types, in which case they will not even see this new version.

regevbr:
@sandersn thanks for you answer!
@jgonggrijp I have made the relevant changes to comply with the semver of this project. Please review the changes.

jgonggrijp (approved):
Looks good to me!

jgonggrijp:
@sandersn thanks for helping out.

types/backbone-associations/index.d.ts
```diff
@@ -1,8 +1,8 @@
-// Type definitions for Backbone-associations 0.6.4
+// Type definitions for Backbone-associations 0.6
 // Project: http://dhruvaray.github.io/backbone-associations
 // Definitions by: Craig Brett <https://github.com/craigbrett17>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+// TypeScript Version: 2.8
```
DanielRosenwasser: Why did this need to change? This package wasn't otherwise changed.

DanielRosenwasser (changes requested):

types/backbone-associations/index.d.ts
regevbr: Hi @DanielRosenwasser 

The `backbone-associations` library depends on `backbone` which in turn depends on `underscore` which we just modified and it now requires TypeScript 2.8 to operate.

That change forces TypeScript version change to `backbone-associations` (and many other libraries as you can see in the changed files)

If you look at @jgonggrijp comment below, you will see that this change won't affect current users as they can choose to import a version that supports the TypeScript version they actually use.
 
> @jgonggrijp You are correct. People who are using old versions of typescript should be using `@types/underscore@ts2.6` (to use TS 2.6 as an example) to acquire types, in which case they will not even see this new version.

If you believe that this is not an issue, please mark the requested changes as resolved, otherwise please let us know.

Thanks!

jgonggrijp: @regevbr I agree with you, but that quote belongs to `@sandersn` (back-quoting in order to not needlessly attract his attention).

@DanielRosenwasser you can find the quote in question over here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36230#issuecomment-502837876 which in turn was a reply to my question over here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36230#pullrequestreview-250386425. Please observe that my question mentions *package versions*, which indeed should not be bumped in the packages that didn't change. The *TypeScript versions* did need to change because of the "viral" cascade that @regevbr described.

regevbr: @jgonggrijp my bad for the quote attribution :-)

PR #36237, opened 2019-06-17
Node: ref and unref for timers returns this + new method hasRef

dex4er:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/api/timers.html#timers_immediate_unref

### DIFF
--- a/types/node/v11/globals.d.ts
+++ b/types/node/v11/globals.d.ts
@@ -1087,15 +1087,18 @@ declare namespace NodeJS {
     }
 
     class Immediate {
-        ref(): void;
-        unref(): void;
+        hasRef(): boolean;
+        ref(): this;
+        refresh(): this;
+        unref(): this;
         _onImmediate: Function; // to distinguish it from the Timeout class
     }
 
     class Timeout implements Timer {
-        ref(): void;
-        refresh(): void;
-        unref(): void;
+        hasRef(): boolean;
+        ref(): this;
+        refresh(): this;
+        unref(): this;
     }
 
     class Module {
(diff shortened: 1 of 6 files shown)


types/node/v11/globals.d.ts
```diff
@@ -1081,21 +1081,25 @@ declare namespace NodeJS {
     }
 
     interface Timer {
+        hasRef(): boolean;
         ref(): void;
         refresh(): void;
```
Flarna: Shouldn't these APIs here also return `this`?

dex4er: Yes, of course. I'm surprised that test didn't catch that.

Flarna (approved):

types/node/v11/globals.d.ts
Flarna: Most likely because your test uses `class Timeout` which was adapted already. I find it interesting that typescript allows a class to implement an interface where return type of methods differs. Maybe because interface uses `void` so the nothing get's broken if interface api is used.

DanielRosenwasser:
@andrewbranch @sandersn it looks like tests for Node are timing out and I don't have access to reboot them. See #36263 as well.

andrewbranch:
Yeaaah, so, what’s happening is we limit to one hour per benchmark, but separate versions are separate benchmarks, and if we don’t have a baseline against master, we have to run one for master too. Node is so big that we’re hitting the one hour limit, but it looks like this PR changes v10, v11, and v12, which means we’ll run as many as 6 benchmarks, and Azure Pipelines cuts us off at 6 hours. This is kinda edge casey; I would just ignore the benchmark status check.

PR #36243, opened 2019-06-17
[ldapjs-client] Basic Typings

colthreepv:
Please fill in this template.

If adding a new definition:

# ldapjs-client
Those typings are not the best they can get, as they are inferred by my use of the library, and my ability to read the original source code.

All of the `Promise<any>` that you see are quite correct, the response from LDAP servers is implementation dependent.

I have met a weird tslint error, it kept telling me the following line had unnecessary generics
```typescript
search<T extends any>(base: string, options: PartialSearchOptions): Promise<T[]>;
```

But without (an optional) generic the user cannot specify what kind of response it's going to get. I find that preferable instead of using a type cast after having received the data.

Also I had to disable "npm-naming" in tslint because it kept arguing the original library has no "default" in it, but it just uses `module.exports`. I can't use `module.exports` in Typescript, of course.

Any Insight would be appreciated, that's my first added typing

### DIFF
--- a/types/ldapjs-client/index.d.ts
+++ b/types/ldapjs-client/index.d.ts
@@ -4,36 +4,32 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
-declare class Filter {
-    matches(obj: any): boolean;
-    type: string;
-}
+declare namespace LdapClient {
+    interface Change {
+        operation: 'add' | 'delete' | 'replace';
+        modification: {
+            [key: string]: any;
+        };
+    }
 
-interface Change {
-    operation: string;
-    modification: {
-        [key: string]: any;
-    };
-}
+    interface SearchOptions {
+        scope?: string;
+        filter?: string;
+        attributes?: string[];
+        sizeLimit?: number;
+        timeLimit?: number;
+        typesOnly?: boolean;
+    }
 
-interface SearchOptions {
-    scope: string;
-    filter: string | Filter;
-    attributes: string[];
-    sizeLimit: number;
-    timeLimit: number;
-    typesOnly: boolean;
-}
-type PartialSearchOptions = Partial<SearchOptions>;
-
-interface ClientOptions {
-    url: string;
-    tlsOptions?: object;
-    timeout?: number;
+    interface ClientOptions {
+        url: string;
+        tlsOptions?: object;
+        timeout?: number;
+    }
 }
 
 declare class LdapClient {
-    constructor(options: ClientOptions);
+    constructor(options: LdapClient.ClientOptions);
 
     /**
      * Adds an entry to the LDAP server.
@@ -67,7 +63,7 @@ declare class LdapClient {
      * @param dn the DN of the entry to modify.
      * @param change update to perform (can be [Change]).
      */
-    modify(dn: string, change: Change): Promise<any>;
+    modify(dn: string, change: LdapClient.Change): Promise<any>;
 
     /**
      * Performs an LDAP modifyDN against the server.
@@ -80,7 +76,7 @@ declare class LdapClient {
      * Note that the defaults for options are a 'base' search.
      */
     // tslint:disable-next-line: no-unnecessary-generics
-    search(base: string, options: PartialSearchOptions): Promise<unknown[]>;
+    search(base: string, options: LdapClient.SearchOptions): Promise<unknown[]>;
 
     /**
      * Unbinds this client from the LDAP server.
--- a/types/ldapjs-client/index.d.ts
+++ b/types/ldapjs-client/index.d.ts
@@ -4,19 +4,19 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
-export class Filter {
+declare class Filter {
     matches(obj: any): boolean;
     type: string;
 }
 
-export interface Change {
+interface Change {
     operation: string;
     modification: {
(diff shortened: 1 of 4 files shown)


types/ldapjs-client/index.d.ts
```diff
@@ -0,0 +1,92 @@
...
+     */
+    modifyDN(dn: string, newName: string): Promise<any>;
+
+    /**
+     * Performs an LDAP search against the server.
+     *
+     * Note that the defaults for options are a 'base' search.
+     */
+    // tslint:disable-next-line: no-unnecessary-generics
+    search<T extends any>(base: string, options: PartialSearchOptions): Promise<T[]>;
```
DanielRosenwasser: I think this is legitimate - the generic is really acting as a type assertion. You should instead pick `unknown[]` and let users cast.

DanielRosenwasser (changes requested):

types/ldapjs-client/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "npm-naming": false
```
DanielRosenwasser: Why is this needed?

DanielRosenwasser (changes requested):

types/ldapjs-client/tslint.json
colthreepv: I wrote it in the PR description, to be more verbose, it gives the following error:
```
npm-naming  The types for ldapjs-client specify 'export default' but the source does not mention 'default' anywhere.
```

The sources of ldapjs-client have `module.exports = Client;`, but I have to use `export default LdapClient` on the typings

types/ldapjs-client/index.d.ts
```diff
@@ -0,0 +1,92 @@
...
+    search(base: string, options: PartialSearchOptions): Promise<unknown[]>;
+
+    /**
+     * Unbinds this client from the LDAP server.
+     *
+     */
+    unbind(): Promise<any>;
+}
+
+export default LdapClient;
```
DanielRosenwasser: This package was declared using `module.exports = ...` or some variant. It should be declared with an `export = ...` like one of the following:

### Variables

```ts
declare var xyz: SomeType;

export = xyz;
```

### Classes

```ts
class Foo {
  // ...
}

export = Foo
```

### Functions

```ts
function foo(): any;

export = foo;
```

```ts
import foo = require("foo");
```

colthreepv: Ok perfect. Sorry for have missed that part.

There is a problem though, I don't seem to be able to export interfaces and types alongside the default export being LdapClient

Is there a way in Typescript to do this? I have googled but not many people use commonjs exports  in the wild (outside definitelytyped)

colthreepv:
Ok, I have re-enabled the `npm-naming` rule, still my question stands, if anyone more expert than me on TS could answer

> There is a problem though, I don't seem to be able to export interfaces and types alongside the default export being LdapClient
> 
> Is there a way in Typescript to do this? I have googled but not many people use commonjs exports in the wild (outside definitelytyped)

types/ldapjs-client/index.d.ts
```diff
@@ -0,0 +1,92 @@
+// Type definitions for ldapjs-client 0.1
+// Project: https://github.com/zont/ldapjs-client#readme
+// Definitions by: Valerio Coltrè <https://github.com/colthreepv>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+declare class Filter {
```
DanielRosenwasser: Is `Filter` actually a class, or is it an interface?

types/ldapjs-client/index.d.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+interface Change {
+    operation: string;
+    modification: {
+        [key: string]: any;
+    };
+}
+
+interface SearchOptions {
```
DanielRosenwasser: If this is only used in one place, I'd really just avoid writing two types.

DanielRosenwasser (changes requested):
If you really want to be able to reference the types too, you'll need to do something called namespace merging in which you write something like

```
class LdapClient {
    // ...
}

namespace LdapClient {
    export interface Change {
      // ...
    }
}

types/ldapjs-client/index.d.ts
colthreepv: I have never used that.
Blatantly copied from [ldapjs](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ldapjs/index.d.ts#L354)

Honestly [reading carefully the source code](https://github.com/zont/ldapjs-client/blob/master/src/index.js#L96), I'm going to remove this altogether, and just use `string`

Even if the ldapjs-client library has an [entire directory devoted to support Class-Filters](https://github.com/zont/ldapjs-client/tree/master/src/filters)

types/ldapjs-client/index.d.ts
colthreepv: sure

colthreepv:
Thanks a lot for this insight, clever use of merging indeed!
I have refined the definitions again

colthreepv (commented):
Thanks a lot for this insight, clever use of merging indeed!
I have refined the definitions again

colthreepv:
@DanielRosenwasser sorry I think my forced-update broke the PR change request flow
can you review again?

DanielRosenwasser:
Thanks!

PR #36246, opened 2019-06-17
react-stripe-elements: add definition for createPaymentMethod

cnasc:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/stripe/react-stripe-elements#setting-up-your-payment-form-injectstripe>> (used in the first example under `handleSubmit` in that section)

### DIFF
--- a/types/react-stripe-elements/index.d.ts
+++ b/types/react-stripe-elements/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for react-stripe-elements 1.1
+// Type definitions for react-stripe-elements 1.2
 // Project: https://github.com/stripe/react-stripe-elements#readme
 // Definitions by: dan-j <https://github.com/dan-j>
 //                 Santiago Doldan <https://github.com/santiagodoldan>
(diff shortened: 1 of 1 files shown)


yhnavein (approved):

types/react-stripe-elements/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for react-stripe-elements 1.1
+// Type definitions for react-stripe-elements 1.2
```
DanielRosenwasser: This version isn't supposed to be incremented unless it's to reflect version 1.2 of the library.

Also, this library is at version 3.0 so can you clarify if the major version should be changed?

DanielRosenwasser (changes requested):

types/react-stripe-elements/index.d.ts
cnasc: Ah I see. It looks like it should be 3.0, as the signature of `createPaymentMethod` I added to the types is consistent with the 3.0 version. Most of the rest remains compatible, but there is one breaking change: the `PostalCodeElement` class was removed in 3.0. Should I also remove that one when I bump the version?

PR #36247, opened 2019-06-17
Add type support for ioredis to connect-redis

vaskevich:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.npmjs.com/package/connect-redis#custom-redis-clients

### DIFF
--- a/types/connect-redis/index.d.ts
+++ b/types/connect-redis/index.d.ts
@@ -3,14 +3,16 @@
 // Definitions by: Xavier Stouder <https://github.com/xstoudi>
 //				   Seth Butler <https://github.com/sbutler2901>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.2
+// TypeScript Version: 2.8
 
 /// <reference types="express" />
 /// <reference types="express-session" />
+/// <reference types="ioredis" />
 /// <reference types="redis" />
 
 declare module "connect-redis" {
     import * as express from "express";
+    import * as ioredis from "ioredis";
     import * as session from "express-session";
     import * as redis from "redis";
 
@@ -18,12 +20,14 @@ declare module "connect-redis" {
     function s(options: (options?: session.SessionOptions) => express.RequestHandler): s.RedisStore;
 
     namespace s {
-        interface RedisStore extends session.Store {
-            new (options: RedisStoreOptions): RedisStore;
-            client: redis.RedisClient;
+        type RedisClient = redis.RedisClient | ioredis.Redis;
+
+        interface RedisStore<T extends RedisClient = redis.RedisClient> extends session.Store {
+            new <T extends RedisClient = redis.RedisClient>(options: RedisStoreOptions<T>): RedisStore<T>;
+            client: T;
         }
-        interface RedisStoreOptions {
-            client?: redis.RedisClient;
+        interface RedisStoreOptions<T extends RedisClient> {
+            client?: T;
             host?: string;
             port?: number;
             socket?: string;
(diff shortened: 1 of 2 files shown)


sbutler2901 (approved):

types/connect-redis/index.d.ts
```diff
@@ -3,27 +3,31 @@
...
-            client: redis.RedisClient;
+        type RedisClient = redis.RedisClient | ioredis.Redis;
+
+        interface RedisStore<T extends RedisClient = redis.RedisClient> extends session.Store {
+            new <T extends RedisClient = redis.RedisClient>(options: RedisStoreOptions<T>): RedisStore<T>;
+            client: T;
         }
-        interface RedisStoreOptions {
-            client?: redis.RedisClient;
+        interface RedisStoreOptions<T extends RedisClient> {
```
sbutler2901: It would be nice if this extended `redis.ClientOpts` and/or `ioredis.RedisOptions`. For the project I am working on the `tls` property is not recognized and requires `@ts-ignore`. I believe the same would apply to you for `tls` or other options not defined in the interface. Can you see if there is a technique to extend either `redis.ClientOpts` or `ioredis.RedisOptions`?

I can change the typing to `interface RedisStoreOptions<T extends RedisClient> extends redis.ClientOpts {` and tls will be considered a valid property in my project, but I am fairly certain this will not work for ioredis when it comes to property naming discrepancies, such as, `enable_offline_queue` in `redis.ClientOpts` versus `enableOfflineQueue` in `ioredis.RedisOptions`.

sbutler2901 (approved):

types/connect-redis/index.d.ts
vaskevich: I guess one way to do this would be to add another parameter `U`, and change the constructor to something like:

```ts
new <T extends RedisClient = redis.RedisClient, U extends RedisStoreOptions<T> = extends RedisStoreOptions<T>>(options: U): RedisStore<T, U>;
```

does that work?

uniqueiniquity (changes requested):
Marking as changes requested until @sbutler2901's comment is resolved.

types/connect-redis/index.d.ts
sbutler2901: I think there was a typo and reduced it down to 
```typescript
new <T extends RedisClient = redis.RedisClient, U extends RedisStoreOptions<T>>(options: U): RedisStore<T, U>;
```
but unfortunately it didn't work. I don't think its possible with the current design of the typings.

 Considering this and rather than polluting the `RedisStoreOptions` with properties that only concern the client, such as `tls`, I believe it would be better (replacing redis with ioredis where necessary) to intersect the types:
```typescript 
const storeOptions: connectRedis.RedisStoreOptions & redis.ClientOpts = {...};
```
That being said, the generic in `RedisStoreOptions` should be defaulted like it is in the constructor:
```typescript
interface RedisStoreOptions<T extends RedisClient = redis.RedisClient> {
```
otherwise the error, `TS2314: Generic type 'RedisStoreOptions<T>' requires 1 type argument(s).` will be reported to other users when they upgrade the typings if they are typing the options such as:
```typescript
const storeOptions: connectRedis.RedisStoreOptions = {
    host: process.env.REDIS_HOST,
};
```
or
```typescript
const store = new RedisStore({...} as connectRedis.RedisStoreOptions);
```
and are using `redis.RedisClient`

vaskevich: I'm not going to have time to get around to making this change in the next week before the bot closes the PR. Feel free to adapt my PR and sending something out you think should work.

PR #36249, opened 2019-06-17
rename `findDescendants` to `findDescendant`

onlyann:
Slate doesn't have a `findDescendants` method on `BaseNode` but `findDescendant`.

https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/interfaces/element.js#L395


hanstarj (approved):
Looks right to me. Thanks!

PR #36252, opened 2019-06-18
[moment] Create types: @types/moment

PikachuEXE:
Please fill in this template.

If adding a new definition:

## Details
This is for fixing #27942 
Since including `moment` in dependency makes it included in final JS bundle

To make tslint pass I disabled many rules
Made 1 fix to `isBetween`
https://momentjs.com/docs/
```ts
// Before
isBetween(a: MomentInput, b: MomentInput, granularity?: unitOfTime.StartOf, inclusivity?: "()" | "[)" | "(]" | "[]"): boolean;
// After (granularity accepts null)
isBetween(a: MomentInput, b: MomentInput, granularity?: unitOfTime.StartOf | null, inclusivity?: "()" | "[)" | "(]" | "[]"): boolean;
```
Other changes are made on test code (missing var, incorrect function type, etc)

### DIFF
--- a/types/moment/index.d.ts
+++ b/types/moment/index.d.ts
@@ -1,0 +1,751 @@
+// Type definitions for moment 2.24
+// Project: http://momentjs.com
+// Definitions by: PikachuEXE <https://github.com/PikachuEXE>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+// tslint:disable:typedef-whitespace
+// tslint:disable:whitespace
+// tslint:disable:array-type
+// tslint:disable:unified-signatures
+// tslint:disable:no-redundant-undefined
+// tslint:disable:no-padding
+// tslint:disable:strict-export-declare-modifiers
+// tslint:disable:void-return
+// tslint:disable:ban-types
+// tslint:disable:no-var-keyword
+declare function moment(inp?: moment.MomentInput, format?: moment.MomentFormatSpecification, strict?: boolean): moment.Moment;
+declare function moment(inp?: moment.MomentInput, format?: moment.MomentFormatSpecification, language?: string, strict?: boolean): moment.Moment;
+
+declare namespace moment {
+    type RelativeTimeKey = 's' | 'ss' | 'm' | 'mm' | 'h' | 'hh' | 'd' | 'dd' | 'M' | 'MM' | 'y' | 'yy';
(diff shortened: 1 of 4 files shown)


DanielRosenwasser:
Can you explain a bit more? We generally don't add packages that already have their own types.

PikachuEXE:
The main problem is
Having package `moment` in dependency makes it impossible to exclude it from built JS bundle as it's required by `pikaday` when it's available
Best case is to have a separate package, maintained by `moment` author, which contains type definition **only**
So we can include type packages depending on `moment` without installing `moment` runtime

PikachuEXE:
I started an issue to request a separate typing package for `moment`: https://github.com/moment/moment/issues/5154
Please keep this PR opened until they have taken action that fixes #27942

andrewbranch:
> Having package moment in dependency makes it impossible to exclude it from built JS bundle as it's required by pikaday when it's available

@PikachuEXE what are you using as your bundler? The [original issue against pikaday](https://github.com/Pikaday/Pikaday/issues/718) suggests a workaround for Webpack, which is also what I would suggest.

I sympathize with the issue; this kind of stuff is super annoying. But it’s not justification to copy moment’s typings back to here—DefinitelyTyped generally can’t be held responsible for the bad bundling behavior of other packages (pikaday, in this case). I won’t close this PR just yet, but I’m not sure how keeping it open in the meantime will help 😕 

I _do_ want to help you find a solution that doesn’t involve merging this PR, though, and I’m a bit skeptical that moment is going to publish their own typings separately. So I think your best bet would be to configure a workaround in your bundler. But if that’s not possible, you could publish these typings somewhere _other_ than DefinitelyTyped. You could simply push them to a GitHub repo of your own, and install them from there: `npm install --save PikachuEXE/moment-typings`. It’s not very beautiful as a long-term solution, but it should get you unstuck. I’ve had to do this myself a few times in the past—it makes for a perfectly reasonable last resort.

PikachuEXE:
Thanks for your suggestion
Right now I am using my own `pikaday` typing (so `moment` won't be installed)
I am about to switch to the [webpack workaround](https://github.com/Pikaday/Pikaday/issues/718#issuecomment-334738391) you mentioned 
since I confirmed that workaround is working (inspect JS bundle with [`webpack-bundle-analyzer`](https://www.npmjs.com/package/webpack-bundle-analyzer))

PR #20695, opened 2017-10-18
[node] Update older node types (7,6,4) for more consistency with 8

kevin-greene-ck:
* Specifically add interfaces for reqeust headers
  * Make the naming of client request options consistent (ClientRequestArgs)

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: #20579 
- [x] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #36265, opened 2019-06-18
fix(nodejs): remove undefined from OutgoingHttpHeaders

boopathi:
In Node, if a request header is `undefined`, then making a http.request `throws` (synchronously) with a TypeError - [ERR_HTTP_INVALID_HEADER_VALUE](https://nodejs.org/api/errors.html#errors_err_http_invalid_header_value). This is something that should be caught at compile time.

For example,

```js
const http = require("http");

try {
  http.request({
    host: "localhost",
    headers: {
      "Something": undefined
    }
  }, (res) => {
    console.log("Success");
  });
} catch (e) {
  console.log("Sending request with undefined header throws this error - ");
  throw e;
}
```

I've attempted a fix where the OutgoingHttpHeaders used as the input path of a request does not allow `undefined` as one of the possible values and the OutgoingHttpHeaders at the output path - such as `getHeaders()` - undefined is one of the possible values.

I will change the types for older node versions after receiving an initial feedback

---

Please fill in this template.

### DIFF
--- a/types/node/http.d.ts
+++ b/types/node/http.d.ts
@@ -65,7 +65,7 @@ declare module "http" {
 
     // outgoing headers allows numbers (as they are converted internally to strings)
     interface OutgoingHttpHeaders {
-        [header: string]: number | string | string[] | undefined;
+        [header: string]: number | string | string[];
     }
 
     interface ClientRequestArgs {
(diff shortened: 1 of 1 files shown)


Flarna:
see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/20695#discussion_r147345373 why `| undefined` was added.

There are discussions in typescript issue tracker (sorry, have no issue number) regarding should `| undefined` part of an index signature or not as it's inconsistent currently. Manually adding `| undefined` is a workaround with some drawbacks. Most likely this change will break some users applications as they augment `OutgoingHeaders` as described in above comment.

abdulfarid-dev:
Hi, just noting that this topic has come up again in Node.js issue nodejs/node#60418.
The runtime still throws when setting a header to undefined, but the type definitions still allow it (| undefined).
Might be worth revisiting this discussion?

PR #36274, opened 2019-06-19
[react-big-calendar] Change TEvent restriction to object

onlyann:
`TEvent` can have any shape when using property accessors.
The only restriction is to be `object`.
This change fixes Typescript errors when using a custom event structure that doesn't match the default `Event` interface.

There are only a few changes but prettier/linting adds quite a lot of formatting.

### DIFF
--- a/types/react-big-calendar/react-big-calendar-tests.tsx
+++ b/types/react-big-calendar/react-big-calendar-tests.tsx
@@ -16,16 +16,16 @@ class CalendarEvent {
     title: string;
     allDay: boolean;
     start: Date;
-    end: Date;
+    endDate: Date;
     desc: string;
     resourceId?: string;
     tooltip?: string;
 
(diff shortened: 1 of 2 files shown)


types/react-big-calendar/react-big-calendar-tests.tsx
```diff
@@ -16,16 +16,16 @@ class CalendarEvent {
     title: string;
     allDay: boolean;
     start: Date;
-    end: Date;
```
TomasHubelbauer: Why rename this? Did `end` clash with something as an identifier?

onlyann: I arbitrarily chose one of the fields of the custom `CalendarEvent` to be different from the default `Event` interface definition.
Before this PR, it would turn into a Typescript error, complaining that `CalendarEvent` doesn't match `Event`. However, this is a valid example and shouldn't fail.
I modified the example to prevent a regression error on future updates to the declaration file.

TomasHubelbauer (approved):

PR #36279, opened 2019-06-19
[types/google-apps-script] Blob extends BlobSource

oshliaer:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://developers.google.com/apps-script/reference/base/blob-source

Based on the question https://stackoverflow.com/questions/53965253

As I can see the `Class Blob` implements `Interface BlobSource`. We use the namespace with the interfaces (not in the types).  A reference to the extension of the Interface is possible instead of a direct reference to the implementation of a Class. It does not violate the heritage.

Just reject it if I'm wrong.

Thanks.


grant (approved):
Thanks for the test. It may get reverted since that type is generated, but should be detected since there is a test.

oshliaer:
> but should be detected since there is a test.

I think how to do it.

PR #36283, opened 2019-06-19
fixing typing for CurrentRefinementsProvided in react-instantsearch-core

e-krebs:
This PR is fixing a small error in **react-instantsearch-core**.

The `CurrentRefinementsProvided` interface has a `refine` function, which argument should be of type `Refinement | Refinement[]` (instead of `RefinementValue | RefinementValue[]`).

If this `refine` function is called with a `RefinementValue[]` it throws an error.
Whereas it works as expected when calling it with a `Refinement[]`.

_relevant part of the doc:_
https://www.algolia.com/doc/api-reference/widgets/current-refinements/react/?language=react#full-example

### DIFF
--- a/types/react-instantsearch-core/index.d.ts
+++ b/types/react-instantsearch-core/index.d.ts
@@ -227,7 +227,7 @@ export interface CurrentRefinementsExposed {
 
 export interface CurrentRefinementsProvided {
   /** a function to remove a single filter */
-  refine: (refinement: RefinementValue | RefinementValue[]) => void;
+  refine: (refinement: Refinement | Refinement[]) => void;
   /**
    * all the filters, the value is to pass to the refine function for removing all currentrefinements,
    * label is for the display. When existing several refinements for the same atribute name, then you
--- a/types/react-instantsearch-core/react-instantsearch-core-tests.tsx
+++ b/types/react-instantsearch-core/react-instantsearch-core-tests.tsx
@@ -299,14 +299,15 @@ import {
 };
 
 () => {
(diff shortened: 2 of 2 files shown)


gburgett:
Hi @e-krebs ,
The documentation seems to show that the proper thing to do is to pass `refinement.value` to the `refine` function:

image
https://www.algolia.com/doc/api-reference/widgets/current-refinements/react/?language=javascript#connector-param-provided-refine

Which version of react-instantsearch-core are you using?

gburgett:
FYI, my team is using v5.2.3 and we pass `item.value` to the `refine` call which works fine.
```tsx
    return <button key={item.label} className="btn btn-secondary" onClick={() => refine(item.value)}>
        {item.label}
        <i className="material-icons">clear</i>
      </button>
```
```
"@types/react-instantsearch-core@*":
  version "5.2.3"
  resolved "https://registry.yarnpkg.com/@types/react-instantsearch-core/-/react-instantsearch-core-5.2.3.tgz#9642d15cae7749bd91a0952e59a25ecd357846c7"
  integrity sha512-ujEGrMCrLuB4MOG8GJ/Gp0TxdJi2YaNiE8UC65+Cae0oGo+4B0sa42SAJnDO4VpRnh84OxfvrsZkgbUH9wXYlg==
  dependencies:
    "@types/react" "*"
```

gburgett (changes requested):
I suspect this change is incorrect according to the docs and have not heard back from @e-krebs

e-krebs:
sorry, I was on holidays.
no problem, I'll look into it and open a new PR if necessary

PR #36284, opened 2019-06-19
Add type field to simple-peer/SignalData

sky0014:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [https://developer.mozilla.org/en-US/docs/Web/API/RTCSessionDescription](https://developer.mozilla.org/en-US/docs/Web/API/RTCSessionDescription)

The SignalData should have `type` field according to [https://developer.mozilla.org/en-US/docs/Web/API/RTCSessionDescription](https://developer.mozilla.org/en-US/docs/Web/API/RTCSessionDescription), or it will cause error when calling `setRemoteDescription` function.


uniqueiniquity:

PR #36288, opened 2019-06-19
lodash: Make _.invert more precise

danvk:
Context: [this question/answer][1] on Stack Overflow.

The idea is that if you invert an `as const` object, its type can be more precise:

```ts
const kv = {
  x: 'a',
  y: 'b'
} as const;
_.invert(kv);  // type should be { readonly a: 'x', readonly b: 'y' }, not Dictionary<string>.
```

Please fill in this template.

[1]: https://stackoverflow.com/questions/56415826/is-it-possible-to-precisely-type-invert-in-typescript/56416192

### DIFF
--- a/types/lodash/common/object.d.ts
+++ b/types/lodash/common/object.d.ts
@@ -1990,14 +1990,8 @@ declare module "../index" {
     // helpers
     type Keyify<T> = T extends PropertyKey ? T : string;
     type Stringify<T> = T extends string ? T : string;
-    type KeyValues<T> = {
-        [P in keyof T]: {
-            key: Stringify<P>,
-            value: Keyify<T[P]>,
-        }
-    }[keyof T];
-    type InvertResult<T> = {} & {
-        [P in Keyify<T[keyof T]>]: Extract<KeyValues<T>, { value: P }>['key']
+    type InvertResult<T> = {
+        [P in Keyify<T[keyof T]>]: Stringify<keyof T>;
     };
 
     interface LoDashStatic {
--- a/types/lodash/common/object.d.ts
+++ b/types/lodash/common/object.d.ts
@@ -1988,12 +1988,17 @@ declare module "../index" {
     // invert
 
     // helpers
-    type InvertAllValues<T extends Record<PropertyKey, PropertyKey>> = {
-        [P in keyof T]: { key: P, value: T[P] }
-    }[keyof T]
-    type InvertResult<T extends Record<PropertyKey, PropertyKey>> = {
-        [P in InvertAllValues<T>['value']]: Extract<InvertAllValues<T>, { value: P }>['key']
-    }
+    type Keyify<T> = T extends PropertyKey ? T : string;
+    type Stringify<T> = T extends string ? T : string;
+    type KeyValues<T> = {
+        [P in keyof T]: {
+            key: Stringify<P>,
+            value: Keyify<T[P]>,
+        }
+    }[keyof T];
+    type InvertResult<T> = {} & {
+        [P in Keyify<T[keyof T]>]: Extract<KeyValues<T>, { value: P }>['key']
+    };
 
     interface LoDashStatic {
         /**
--- a/types/lodash/common/object.d.ts
+++ b/types/lodash/common/object.d.ts
@@ -1991,7 +1991,7 @@ declare module "../index" {
     type InvertAllValues<T extends Record<PropertyKey, PropertyKey>> = {
         [P in keyof T]: { key: P, value: T[P] }
     }[keyof T]
-    type InvertResult<T extends Record<PropertyKey, PropertyKey>> = {} & {
+    type InvertResult<T extends Record<PropertyKey, PropertyKey>> = {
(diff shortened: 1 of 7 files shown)


danvk:
Maintainers: I'm consistently getting out of memory errors with this change. Is this just a bad idea?

types/lodash/common/object.d.ts
```diff
@@ -1987,32 +1987,47 @@ declare module "../index" {
 
     // invert
 
+    // helpers
+    type InvertAllValues<T extends Record<PropertyKey, PropertyKey>> = {
+        [P in keyof T]: { key: P, value: T[P] }
+    }[keyof T]
+    type InvertResult<T extends Record<PropertyKey, PropertyKey>> = {} & {
+        [P in InvertAllValues<T>['value']]: Extract<InvertAllValues<T>, { value: P }>['key']
+    }
```
aj-r: Tests are running out of memory so we need to find a simpler way of doing this. Would something like this work?

```ts
type Invert<T> = T extends Record<infer U, infer V> ? Record<V, U> : Dictionary<string>;
```

I'm not sure if TS will like that though. Also, even if it compiles, it will lose some information - your test ocject would probably get mapped to something like `{ x: "a" | "b", y: "a" | "b" }`, which is an improvement over the old version, but not perfect.

aj-r (changes requested):
Once you find something that works:
- do `npm run generate` (see readme) to update the FP types.
- update the ts3.1 types (including  `npm run generate` there, too)

types/lodash/common/object.d.ts
```diff
@@ -1987,32 +1987,47 @@ declare module "../index" {
 
     // invert
 
+    // helpers
+    type InvertAllValues<T extends Record<PropertyKey, PropertyKey>> = {
+        [P in keyof T]: { key: P, value: T[P] }
+    }[keyof T]
+    type InvertResult<T extends Record<PropertyKey, PropertyKey>> = {} & {
```
aj-r: What does the `{} & ` do? Maybe that is causing the performance issues?

danvk:
I've removed the `{} &`. See [this comment](https://stackoverflow.com/questions/56415826/is-it-possible-to-precisely-type-invert-in-typescript/56416192#comment99430271_56416192) for context on why it was there to begin with. I'm still getting OOM errors locally, so I'm not optimistic.

aj-r:
Apparently that made it worse somehow?

Here's a slight modification that (might) simplify things a bit.

```ts
type Keyify<T> = T extends PropertyKey ? T : string;
type Stringify<T> = T extends string ? T : string;
type KeyValues<T> = {
    [P in keyof T]: {
        key: Stringify<P>,
        value: Keyify<T[P]>,
    }
}[keyof T];
type Invert<T> = {
  [P in Keyify<T[keyof T]>]: Extract<AllValues<T>, { value: P }>['key']
}
```

It might not help since it still has the problem of being O(n^2). However it is a bit more accurate, since the resulting values will *always* be strings. For example:

```ts
_.invert({ 1: 'a' }); // -> { 'a': '1' }
```

If that doesn't work, here's the simplest version I can think of. Again, not perfect, but hopefully better than `Dictionary<string>`. I'd be surprised if this had performance issues.

```ts
type Keyify<T> = T extends PropertyKey ? T : string;
type Stringify<T> = T extends string ? T : string;
type Invert<T> = {
  [P in Keyify<T[keyof T]>]: Stringify<keyof T>;
};
```

andrewbranch:
I'm wondering if this is something wonky on the compiler side. It feels suspicious that batch compilation (`tsc`) looks fine (1% more memory) but language service stuff is running out of memory.

danvk:
@aj-r I pushed a change with the "slight modification" you suggested above. I had to modify `fp.d.ts` manually to make it parse after running `npm run generate`.

`npm run lint lodash` OOMs on my machine so I'm not optimistic. I'll try your "simplest version" once this fails on CI.

types/lodash/fp.d.ts
```diff
@@ -1789,7 +1789,7 @@ declare namespace _ {
     }
     type LodashIntersectionWith1x5<T1> = (array: lodash.List<T1> | null | undefined) => T1[];
     type LodashIntersectionWith1x6<T1, T2> = (comparator: lodash.Comparator2<T1, T2>) => T1[];
-    type LodashInvert = (object: object) => lodash.Dictionary<string>;
+    type LodashInvert = <T extends Record<PropertyKey, PropertyKey>>(object: T) => lodash.InvertResult<T>;
```
danvk: FYI there is a bug in `generate-fp.ts`; this comes out as:

```ts
type LodashInvert = <T extends Record<PropertyKey>(object: T) => lodash.InvertResult<T>;
```

which is a syntax error. I had to manually edit.

danvk:
@aj-r @andrewbranch Even the "simplest version I can think of" version OOMs and has regressions on the performance benchmark. I'm out of ideas here!

andrewbranch:
I think something weird and bad is happening in the language service. I filed a TypeScript bug: https://github.com/microsoft/TypeScript/issues/32111

PR #36289, opened 2019-06-19
Add definitions for cypress-cucumber-preprocessor

aleccool213:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/cypress-cucumber-preprocessor/index.d.ts
+++ b/types/cypress-cucumber-preprocessor/index.d.ts
@@ -3,4 +3,9 @@
 // Definitions by: Alec Brunelle <https://github.com/aleccool213>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export {};
+export default function preprocessor(options?: any): (file: any) => Promise<string>;
+
+/**
+ * @returns NodeJS.ReadableStream
+ */
+export function transform(file: any): any;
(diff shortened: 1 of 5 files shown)


types/cypress-cucumber-preprocessor/index.d.ts
```diff
@@ -0,0 +1,6 @@
+// Type definitions for cypress-cucumber-preprocessor 1.12
+// Project: https://github.com/TheBrainFamily/cypress-cucumber-preprocessor
+// Definitions by: Alec Brunelle <https://github.com/aleccool213>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export {};
```
DanielRosenwasser: This seems exceedingly incomplete, so this will be a regression for users in `noImplicitAny`.

At the very least, this module should provide `default` and `transform. `steps/index.d.ts` should define

- `given`
- `when`
- `then`
- `and`
- `but`
- `Given`
- `When`
- `Then`
- `And`
- `But`
- `defineParameterType`
- `defineStep`

DanielRosenwasser (changes requested):

types/cypress-cucumber-preprocessor/index.d.ts
aleccool213: Good point :) I will add these.

aleccool213: @DanielRosenwasser Checkout new commits.

DanielRosenwasser:
Thanks!

PR #36290, opened 2019-06-19
Using distributed pick over unions props

radiosterne:
PRevents losing type information when wrapping components with union props

### DIFF
--- a/types/styled-components/index.d.ts
+++ b/types/styled-components/index.d.ts
@@ -7,6 +7,7 @@
 //                 Jason Killian <https://github.com/jkillian>
 //                 Sebastian Silbermann <https://github.com/eps1lon>
 //                 David Ruisinger <https://github.com/flavordaaave>
+//                 Andrey Kurnoskin <https://github.com/radiosterne>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.9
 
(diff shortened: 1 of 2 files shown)


eps1lon (approved):

DanielRosenwasser:
Looks okay, but can I get @weswigham and @sandersn to take a quick look?

weswigham:
This is just a revert of #34499. It's okayish so long as we're OK with TS 3.4 users going back to having horrendous performance with the newest `styled-components`...

DanielRosenwasser:
No, I don't think we are unless we're willing to backport non-trivial changes to 3.4. Instead of this, we could create a 3.5 folder and let `typesVersions` redirect 3.5 users appropriately.

DanielRosenwasser:
Alternatively, create a 3.4 folder that redirects 3.4 users to the "worse" types.

eps1lon:
> No, I don't think we are unless we're willing to backport non-trivial changes to 3.4.

Which means this change can never be accepted because of 3.4? Maintenance is already pretty slow for types packages. Having to maintain a fork would only worsen the current situation.

I'd say we just mark this as wontfix until 3.4 has reached its end of life.

radiosterne:
3.4 folder and redirection to not-so-good types for these poor souls looks good to me, tbh: if we're good to go, I may be able to do this in a few days.

Lavoaster:
I feel like it's a good compromise to have a version specifically for 3.4 - and not have it be maintained. 

The "fix" for consumers would simply be upgrading TS to a new version, which _should_ be a fairly easy thing for them to do. From experience, I've never had a problem updating TS versions between "minor" versions, but that's only one perspective.

uniqueiniquity (changes requested):
Marking as changes requested until 3.4 folder is created as discussed.

PR #36291, opened 2019-06-19
Fix react-addons-update export

### LINKED ISSUE #12509: `react-addons-update` has no default export
- [x] I tried using the latest `react-addons-update/index.d.ts` file in this repo and had problems.

When using the `react-addons-update` typings, I get the following error:
```js
Module '"[...]/node_modules/@types/react-addons-update/index"' has no default export.
```

If I change in the definition file:
```js
export = React.__Addons.update;

to

export default React.__Addons.update;
```

It works fine.

The syntax I am using to import is `import update from "react-addons-update"`.

skovy:
This resolves https://github.com/DefinitelyTyped/DefinitelyTyped/issues/12509

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.npmjs.com/package/react-addons-update

### DIFF
--- a/types/react-addons-update/index.d.ts
+++ b/types/react-addons-update/index.d.ts
@@ -6,7 +6,7 @@
 
 import * as React from 'react';
 
-export = React.__Addons.update;
+export default React.__Addons.update;
 
 declare module 'react' {
     interface UpdateSpecCommand {
@@ -32,4 +32,3 @@ declare module 'react' {
         export function update(value: {}, spec: UpdateSpec): any;
     }
 }
-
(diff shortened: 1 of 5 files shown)


lukyth (approved):

types/react-addons-update/index.d.ts
```diff
@@ -6,7 +6,7 @@
 
 import * as React from 'react';
 
-export = React.__Addons.update;
+export default React.__Addons.update;
```
uniqueiniquity: It seems like the [underlying library](https://unpkg.com/react-addons-update@15.6.2/index.js) is not using `export default`, so this .d.ts file should not either. Instead, you should continue to use `export =`. Note that if you want to use `import name from 'package';` in your consuming code, you can use the `esModuleInterop` compiler setting.

uniqueiniquity (changes requested):

PR #36294, opened 2019-06-19
Fixes for early June TS3.6 DOM update

sandersn:
1. Apollo-link used GlobalFetch, which has been removed. Until my PR goes in (apollographql/apollo-link#1095), I added a shim in apollo-upload-client, which uses apollo-link-http-common.
2. dom-inputevent is updated to match the spec-standard InputEvent types that will ship with Typescript 3.6.
3. Same for webappsec-credential-management
4. Fixed to-markdown's tests to work with the new, non-nullable definition of node.style.fontStyle.

### DIFF
--- a/types/apollo-upload-client/index.d.ts
+++ b/types/apollo-upload-client/index.d.ts
@@ -2,7 +2,7 @@
 // Project: https://github.com/jaydenseric/apollo-upload-client#readme
 // Definitions by: Edward Sammut Alessi <https://github.com/Slessi>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.7
+// TypeScript Version: 3.1
 
 import { ApolloLink } from "apollo-link";
 import { HttpOptions } from "apollo-link-http-common";
@@ -11,7 +11,7 @@ export { ReactNativeFile } from "extract-files";
 
 declare global {
     interface GlobalFetch {
-        fetch: (input: string | Request, init?: RequestInit) => Promise<Response>; // WindowOrWorkerGlobalScope['fetch'];
+        fetch: WindowOrWorkerGlobalScope["fetch"];
     }
 }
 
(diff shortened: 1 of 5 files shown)


SimenB:
@sandersn this explodes when I do not have the DOM typings, which I don't have when using React Native. Any workaround (beyond downgrading, which I just did)?

image

sandersn:
@SimenB seems like the correct fix is actually to put in a type for `fetch` that doesn't depend on the DOM, like some of my earlier commits on this PR did. I didn't think enough of non-DOM users here. 

I changed apollo-link itself too, but with the same code, so I'll send another PR there before they ship a new version.

sandersn:
Actually, that still depends on types `Request`, `RequestInit` and `Response`. I'm not sure how `GlobalFetch` was working without the DOM typings before.

sandersn:
Yep, `GlobalFetch` is in lib.dom.d.ts in TS3.5 and below. @SimenB what happens in your react native project when you write `let x: GlobalFetch["fetch"]`? Does it error? If not, where does it go when you goto-def? Is it lib.dom.d.ts? If not, then where? What about Request, RequestInit and Response?

SimenB:
It worked since the react native typings defines `GlobalFetch`: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/efa6b9e11eb0167547d6611490be48b94bc8d7ea/types/react-native/globals.d.ts#L24-L136

sandersn:
Ah, then the right fix is to change react-native types to define that piece of WindowOrWorkerGlobalScope as well.

SimenB:
Might be nice to have a shared `fetch` type that core `WindowOrWorkerGlobalScope `, `node-fetch` and `react-native` (and potentially `node` if https://github.com/nodejs/node/issues/19393 ever lands) could share? Might break down, and just adding the second declaration to RN is probably quicker for now 🙂

sandersn:
Yeah, I'm doing the latter right now. I'm not sure where the shared definition would live, to make it accessible to all 3 of those types.

PR #36302, opened 2019-06-19
node-hid: add new definitions

peitschie:
Updating based on https://github.com/node-hid/node-hid/commit/0ff75bf17212243d7f1d05e445a3dbead0f5ab08
Additional event handles added as HID class inherits directly from Node's EventEmitter: https://github.com/node-hid/node-hid/blob/0ff75bf17212243d7f1d05e445a3dbead0f5ab08/nodehid.js#L33


thegecko (approved):
LGTM

PR #36322, opened 2019-06-20
[got] Fix got.extend types

avaly:
Please fill in this template.

This PR changes the types so that the following code is valid:

```js
const extended = got.extend({ timeout: 100 });

extended('/example', {
	body: { foo: 'bar' },
	form: true
});
```

### DIFF
--- a/types/got/got-tests.ts
+++ b/types/got/got-tests.ts
@@ -264,6 +264,10 @@ got('http://todomvc.com', { cookieJar: new tough.CookieJar() });
 // Test extension
 got.extend({ method: 'POST' })('/example');
 got.extend({ method: 'POST' }).extend({ headers: {} }).stream('/example');
+got.extend({ timeout: 100 })('/example', { body: {}, json: true });
+got.extend({ timeout: 100 })('/example', { body: { foo: 'bar' }, json: true });
+got.extend({ timeout: 100 })('/example', { body: {}, form: true });
+got.extend({ timeout: 100 })('/example', { body: { foo: 'bar' }, form: true });
 
 // JSON options:
 // $ExpectType Promise<any>
--- a/types/got/index.d.ts
+++ b/types/got/index.d.ts
@@ -120,6 +120,7 @@ declare namespace got {
     };
 
     interface GotExtend {
+        (options: GotOptions<string | null>): GotInstance;
         (options: GotJSONOptions): GotInstance<GotJSONFn>;
         (options: GotFormOptions<string>): GotInstance<GotFormFn<string>>;
         (options: GotFormOptions<null>): GotInstance<GotFormFn<null>>;


types/got/got-tests.ts
```diff
@@ -266,20 +266,24 @@ got.extend({ method: 'POST' })('/example');
 got.extend({ method: 'POST' }).extend({ headers: {} }).stream('/example');
 
 // JSON options:
-// $ExpectType Promise<any>
+got.extend({ timeout: 100 })('/example', { body: {}, json: true });
+got.extend({ timeout: 100 })('/example', { body: { foo: 'bar' }, json: true });
+// $ExpectType Promise<string>
 got.extend({ json: true })('/example').then(({ body }) => body);
-// $ExpectType Promise<any>
+// $ExpectType Promise<string>
```
ikokostya: This is incorrect. If you pass `json: true` option, then result must `Promise<any>`, because `JSON.parse` will be called.

types/got/index.d.ts
```diff
@@ -120,11 +106,11 @@ declare namespace got {
     };
 
     interface GotExtend {
-        (options: GotJSONOptions): GotInstance<GotJSONFn>;
```
ikokostya: Why this was removed?

ikokostya (changes requested):

ikokostya (commented):
Can you explain your problem with the provided code example:

```ts
const extended = got.extend({ timeout: 100 });

extended('/example', {
	body: { foo: 'bar' },
	form: true
});
```

?

avaly:
> Can you explain your problem with the provided code example:

The issue with the example code without the changes in this PR is the following:

1. because the options sent in the `got.extend()` call do no contain any `json` or `form` field, that call gets through this `GotExtend` override:

`        (options: GotBodyOptions<string>): GotInstance<GotBodyFn<string>>;`

2. So the extended calls are now expecting to be called with `GotBodyFn<string>` - it will disallow any calls with `json` or `form` - which are actually valid call signatures in `got`.

wingsbob:
@avaly Looks like you need to add some more overloads to `GotBodyFn`. There may also be some needed for `GotFormFn`, but I'm not sure

The existing tests should all be left as-is, and a new one added:
```typescript
// $ExpectType Promise<any>
got.extend({})('/example', { json: true }).then(({ body }) => body);
```

ikokostya:
`got.extend()` call without any `json` or `form` field returns `got.GotInstance<got.GotBodyFn<string>>`, i.e. the same result as `got()` call without any `json` or `form` field:

```ts
const res: got.GotPromise<string> = got('/exmaple')
```

Removing overrides resolves provided case, but breaks others.

I think that right approach is to add another override for `got.extend()` for most common case. Also, there is question how handle the following cases:

```ts
const client = got.extend({json: true});
const client2 = client.extend({form: true});
```
?

ikokostya:
Adding new override should fix all cases:

```diff
interface GotExtend {
+       (options: GotOptions<string | null>): GotInstance;
        (options: GotJSONOptions): GotInstance<GotJSONFn>;
        (options: GotFormOptions<string>): GotInstance<GotFormFn<string>>;
        (options: GotFormOptions<null>): GotInstance<GotFormFn<null>>;
        (options: GotBodyOptions<string>): GotInstance<GotBodyFn<string>>;
        (options: GotBodyOptions<null>): GotInstance<GotBodyFn<null>>;
    }
}
```
Override must be first.

ikokostya:
> Adding new override should fix all cases:
> 
> ```diff
> interface GotExtend {
> +       (options: GotOptions<string | null>): GotInstance;
>         (options: GotJSONOptions): GotInstance<GotJSONFn>;
>         (options: GotFormOptions<string>): GotInstance<GotFormFn<string>>;
>         (options: GotFormOptions<null>): GotInstance<GotFormFn<null>>;
>         (options: GotBodyOptions<string>): GotInstance<GotBodyFn<string>>;
>         (options: GotBodyOptions<null>): GotInstance<GotBodyFn<null>>;
>     }
> }
> ```
> 
> Override must be first, because it should match call without `body`, `json`, or `form` options.

No, It doesn't handle Buffer case

```ts
const extended = got.extend({ timeout: 100, encoding: null });
```

PR #36325, opened 2019-06-20
Updates for Microsoft Azure AD authentication

kaaloo:
Please fill in this template.

Please see the table describing the parameters for the signin url here.

https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc#send-the-sign-in-request

### DIFF
--- a/types/openid-client/index.d.ts
+++ b/types/openid-client/index.d.ts
@@ -73,9 +73,7 @@ export class Client {
     oauthCallback(redirectUri: string, parameters: {}, checks?: {
         readonly response_type?: string;
         readonly state?: string;
-        readonly nonce?: string;
         readonly code_verifier?: string;
-        readonly max_age?: number;
     }): Promise<TokenSet>;
 
     userinfo(accessToken: string | TokenSet): Promise<{ readonly [name: string]: {} | null | undefined }>;
--- a/types/openid-client/index.d.ts
+++ b/types/openid-client/index.d.ts
@@ -39,6 +39,7 @@ export interface AuthorizationUrlParameters {
 
     readonly response_mode?: string;
     readonly nonce?: string;
+    readonly prompt?: string;
     readonly resource?: string;
     readonly code_challenge?: string;
     readonly code_challenge_method?: string;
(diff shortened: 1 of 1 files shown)


types/openid-client/index.d.ts
```diff
@@ -69,6 +70,14 @@ export class Client {
         readonly max_age?: number;
     }): Promise<TokenSet>;
 
+    oauthCallback(redirectUri: string, parameters: {}, checks?: {
+        readonly response_type?: string;
+        readonly state?: string;
+        readonly nonce?: string;
+        readonly code_verifier?: string;
+        readonly max_age?: number;
+    }): Promise<TokenSet>;
```
ulrichb: Here we should drop max_age and nonce. 

See 
> If you're using pure OAuth 2.0 then client.oauthCallback(redirectUri, parameters[, checks[, extras]]) is the OAuth 2.0 variant of this method, it has the same signature with the exception of checks only supporting code_verifier, state and response_type.

https://github.com/panva/node-openid-client/blob/master/docs/README.md#clientcallbackredirecturi-parameters-checks-extras

... and see also the code: https://github.com/panva/node-openid-client/blob/8d15d302f7fa53d953f032d07fc03b7f0a008285/lib/client.js#L488

types/openid-client/index.d.ts
```diff
@@ -78,13 +87,17 @@ export class Client {
...
+    refresh(refreshToken: string | TokenSet): Promise<TokenSet>;
 }
 
 export class TokenSet {
     readonly access_token?: string;
     readonly token_type?: string;
     readonly id_token?: string;
     readonly refresh_token?: string;
+    readonly expires_at?: number;
+    readonly scope?: string;
```
ulrichb: `expires_at`: Okay.

But `scope` ?

See https://github.com/panva/node-openid-client/blob/master/docs/README.md#new-tokensetinput

ulrichb (changes requested):
... could you also please add test cases  ...

types/openid-client/index.d.ts
kaaloo: Thank you for the catch!

types/openid-client/index.d.ts
kaaloo: `scope` is a Microsoft specific required parameter and is the rationale for this PR.  Without it I was unable to use the types as is.

https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-protocols-oidc#send-the-sign-in-request

> This request is similar to the first leg of the OAuth 2.0 authorization code flow, with these important distinctions:
> 
> The request must include the openid scope in the scope parameter.
> The response_type parameter must include id_token.
> The request must include the nonce parameter.

ulrichb: You linked a `.../oauth2/v2.0/authorize?..scope=..` URL => so the `AuthorizationUrlParameters` input applies (where `scope` already exists). I don't quite get why you are extending the `TokenSet` (which is primarily an output type).

... also please add test cases for your changes.

Thanks!

kaaloo:
Closing this PR since node-openid client now carries its own types.

PR #36334, opened 2019-06-20
[openid-client]: fix Issuer constructor; improve object typing and return types

YangusKhan:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: The current types are not accurate to the documentation and source code of the package. Specifically, you should be able to instantiate an Issuer with a regular constructor, even if it isn't the preferred method. [https://github.com/panva/node-openid-client/blob/master/docs/README.md](https://github.com/panva/node-openid-client/blob/master/docs/README.md)

### DIFF
--- a/types/openid-client/index.d.ts
+++ b/types/openid-client/index.d.ts
@@ -16,7 +16,7 @@ export interface IssuerMetadata {
     readonly authorization_endpoint?: string;
     readonly userinfo_endpoint?: string;
     readonly token_endpoint?: string;
-    readonly jwks_uri: string;
+    readonly jwks_uri?: string;
     readonly revocation_endpoint?: string;
     readonly introspection_endpoint?: string;
     readonly end_session_endpoint?: string;
--- a/types/openid-client/index.d.ts
+++ b/types/openid-client/index.d.ts
@@ -81,11 +81,11 @@ export class Client {
 
     endSessionUrl(parameters?: EndSessionUrlParameters): string;
 
-    callbackParams(input: string | IncomingMessage): object;
+    callbackParams(input: string | IncomingMessage): {};
 
     callback(
         redirectUri: string,
-        parameters: object,
+        parameters: {},
         checks?: {
             readonly response_type?: string;
             readonly state?: string;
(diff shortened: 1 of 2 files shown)


ulrichb (changes requested):
Please don't change the `{} | null | undefined }` output contracts to `any` (reasoning: it's about type safety; the unknown JSON results should be explicitly casted to whatever is expected).

Further: All the reformatting makes it heard to review the changes. Maybe propose an own PR for the reformatting.

ulrichb (changes requested):
See comment above.

YangusKhan:
> Please don't change the `{} | null | undefined }` output contracts to `any` (reasoning: it's about type safety; the unknown JSON results should be explicitly casted to whatever is expected).

Can you give an example of how you would type cast from this interface, then? Because my change was motivated by using the `Client.userinfo` function, which returns strings instead of objects. Here is my example usage:
```
const userinfo = await client.userinfo(token);
const username = snakeCase(userinfo.name || "anonymous") + "-" + randomBytes(2).toString("hex");
```
Explicitly casting my `userinfo` result would be something like this:
```
type UserInfo = {
  [k in "sub" | "name"]: string | null | undefined;
};
const userinfo = (await client.userinfo(token)) as UserInfo;
```
and it seems a little strange to type cast from your `{}` type to a `string`.

> Further: All the reformatting makes it heard to review the changes. Maybe propose an own PR for the reformatting.

I'm only using the `prettier` config that's set up by default in this repository.

ulrichb:
Using you sample:

```ts
interface UserInfo {
  sub?: string;
  name?: string;
};
const userinfo = (await client.userinfo(token)) as UserInfo;
console.log(userinfo.name.toUpperCase(), userinfo.invalid_prop);
```

With the explicit cast to `UserInfo` (which can be defined much easier as simple interface), TSC would correctly warn about a) the missing undefined-check for name and b) warn about using `invalid_prop`.

In contrast, when using `any`, we don't force the cast; and without the cast, TSC would not emit any errors for the erroneous code above (because of the inferred `any` type).

> I'm only using the prettier config that's set up by default in this repository.

Hmm, still it doesn't make reviewing the PR easier :)

ulrichb (approved):
Nice. 

Many thanks for your PR!

PR #36338, opened 2019-06-21
[react] add tests for useState

joehillen:
Add a test case to prevent [this](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/598ffd0b99dcfa511ab6056d9652c99665a7bbe4#r34006320) from happening again.

Please fill in this template.

### DIFF
--- a/types/react/test/hooks.tsx
+++ b/types/react/test/hooks.tsx
@@ -201,6 +201,12 @@ function useEveryHook(ref: React.Ref<{ id: number }>|undefined): () => boolean {
     // make sure the generic argument does reject actual potentially undefined inputs
     // $ExpectError
     React.useState<number>(undefined)[0];
+    // make sure useState does not widen
+    const [toggle, setToggle] = React.useState(false);
+    // $ExpectType boolean
+    toggle;
+    // make sure setState accepts a function
+    setToggle(r => !r);
 
     // useReducer convenience overload
 


ferdaber:
So how does this pass here but not in the reported cases?

joehillen:
@ferdaber Correct. Here is the test output if I revert your revert:

```
Error: /home/joe/src/DefinitelyTyped/types/react/test/hooks.tsx:217:15
ERROR: 217:15  expect  TypeScript@next compile error: 
Parameter 'r' implicitly has an 'any' type.
ERROR: 217:15  expect  TypeScript@next compile error: 
Argument of type '(r: any) => boolean' is not assignable to parameter of type 'boolean | ((prevState: false) => false) | ((prevState: true) => true)'.
  Type '(r: any) => boolean' is not assignable to type '(prevState: false) => false'.
    Type 'boolean' is not assignable to type 'false'.

    at /home/joe/src/DefinitelyTyped/node_modules/dtslint/bin/index.js:190:19
    at fulfilled (/home/joe/src/DefinitelyTyped/node_modules/dtslint/bin/index.js:5:58)
    at <anonymous>
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

DanielRosenwasser:
Is it effectively just a difference between versions? (CC @weswigham)

joehillen:
I don't understand your question. Different versions of what?

joehillen:
To clarify, this test case is a common pattern in my React codebase and is what started returning an error when I updated @types/react and TS 3.5 yesterday. I suspect the error has to do with both the bad [commit](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/598ffd0b99dcfa511ab6056d9652c99665a7bbe4) and the [smarter type unions change](https://github.com/microsoft/TypeScript/pull/30779) introduced in 3.5.

What I think happened is `(prevState: boolean) => boolean` was widened to `((prevState: false) => false) | ((prevState: true) => true)`, which doesn't make sense because why would you ever call `setState` but not change the value.

Because it is such a common pattern in React, I thought it made an obvious test case.

PR #36340, opened 2019-06-21
Type definitions for susyweb 1.0

SUSYLABS:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/susyweb/utils.d.ts
+++ b/types/susyweb/utils.d.ts
@@ -1,0 +1,94 @@
+import BigNumber = require("bn.js");
+import * as us from "underscore";
+
+type Unit =
+    | "nosophy"
+    | "wei"
+    | "kwei"
+    | "Kwei"
+    | "babbage"
+    | "femtosophy"
+    | "mwei"
+    | "Mwei"
+    | "lovelace"
+    | "picosophy"
+    | "gwei"
+    | "Gwei"
+    | "shannon"
+    | "nanosophy"
+    | "nano"
+    | "szabo"
+    | "microsophy"
+    | "micro"
+    | "finney"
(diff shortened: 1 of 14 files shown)


types/susyweb/utils.d.ts
```diff
@@ -0,0 +1,94 @@
...
+        value: string;
+    }
+    | {
+        t: string;
+        v: string;
+    };
+
+type Hex = string | number;
+
+export default interface Utils {
```
uniqueiniquity: I don't think `export default interface` is really what you want here. From [the source](https://unpkg.com/susyweb@0.20.7/lib/utils/utils.js), it looks like the `utils` module is an object with each of these as properties. To achieve this, you should export each individually (e.g. `export function fromUtf8(val: string): string;`). I don't see a good reason to ever have `export default interface` in any of these files, but this seemed to be the most concrete instance to point out.

uniqueiniquity (changes requested):

types/susyweb/utils.d.ts
SUSYLABS: Thanks, we wish to stay consistent with:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/web3/utils.d.ts

uniqueiniquity: Since `web3` ships [its own types](https://github.com/ethereum/web3.js/tree/2.x/packages/web3-utils), you should stay consistent with them, rather than the out-of-date types on Definitely Typed. The format of those types more correctly reflects what you should do. In fact, I will be putting up a PR to remove the web3 type declarations from DefinitelyTyped.

PR #36342, opened 2019-06-21
selenium-webdriver: add "defineCommand" method to executor class

beta-vulgaris:
method "defineCommand" was missing that would block executing custom commands, that are needed e.g. for enabling file download in headless testing.

Previous pull request: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35992

### DIFF
--- a/types/selenium-webdriver/index.d.ts
+++ b/types/selenium-webdriver/index.d.ts
@@ -1461,6 +1461,21 @@ export class Capabilities {
  */
 export class Executor {
   /**
+   * Defines a new command for use with this executor. When a command is sent,
+   * the {@code path} will be preprocessed using the command's parameters; any
+   * path segments prefixed with ':' will be replaced by the parameter of the
+   * same name. For example, given '/person/:name' and the parameters
+   * '{name: 'Bob'}', the final command path will be '/person/Bob'.
+   *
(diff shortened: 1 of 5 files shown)


uniqueiniquity:

types/selenium-webdriver/index.d.ts
```diff
@@ -2072,7 +2056,7 @@ export class WebDriver {
    * @param {!command.Executor} executor The executor to use when sending
    *     commands to the browser.
    */
-  constructor(session: Session|Promise<Session>, executor: Executor);
+  constructor(session: Session|Promise<Session>, executor: http.Executor);
```
cnishina: http.Executor could have been command.Executor... maybe for a future PR? I didn't have time to review this but I had this comment in the previous PR and it was not addressed.

beta-vulgaris: Hi @cnishina ,
Thanks for your comment!
As far as I can see http.Executor is correct, see:

https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/index.js#L641

The code is not super clear in the original selenium-webdriver package. I agree that in a future PR we could clear things up a bit more.

PR #36349, opened 2019-06-21
[@types/slate] Support nested plugin stacks on the "plugins" c…

lukasbuenger:
Both [the docs (see example code)][1] and [the source][2] show that the `plugins` constructor argument can be a nested list of plugins. I'd even argue that this is quite pivotal when it comes to composing atomic behaviors to something more specific. 

As for the names of the intermediate types, I'm not super-opinionated here. Just let me know if you'd like to have them changed.   
   
[1]: https://docs.slatejs.org/guides/plugins#feature-plugins
[2]: https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/controllers/editor.js#L645-L649

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [docs][1], [source][2]

### DIFF
--- a/types/slate/index.d.ts
+++ b/types/slate/index.d.ts
@@ -1131,8 +1131,7 @@ export interface Plugin {
     schema?: SchemaProperties;
 }
 
-export type PluginOrPlugins = Plugin | Plugins;
-export interface Plugins extends Array<PluginOrPlugins> {}
+export interface Plugins extends Array<Plugin | Plugins> {}
 
 export interface EditorProperties {
     onChange?: (change: { operations: Immutable.List<Operation>; value: Value }) => void;
--- a/types/slate/index.d.ts
+++ b/types/slate/index.d.ts
@@ -1131,9 +1131,12 @@ export interface Plugin {
     schema?: SchemaProperties;
 }
 
+export type Plugins = Plugin | Plugin[] | PluginStack;
+export interface PluginStack extends Array<Plugins> {}
+
 export interface EditorProperties {
     onChange?: (change: { operations: Immutable.List<Operation>, value: Value }) => void;
-    plugins?: Plugin[];
+    plugins?: PluginStack;
     readOnly?: boolean;
     value?: Value;
 }
--- a/types/slate/index.d.ts
+++ b/types/slate/index.d.ts
@@ -1131,12 +1131,12 @@ export interface Plugin {
     schema?: SchemaProperties;
 }
 
-export type Plugins = Plugin | Plugin[] | PluginStack;
-export interface PluginStack extends Array<Plugins> {}
+export type Plugins = Plugin | NestedPlugins;
+export interface NestedPlugins extends Array<Plugins> {}
 
 export interface EditorProperties {
     onChange?: (change: { operations: Immutable.List<Operation>; value: Value }) => void;
-    plugins?: PluginStack;
+    plugins?: Plugins;
     readOnly?: boolean;
     value?: Value;
 }
--- a/types/slate/index.d.ts
+++ b/types/slate/index.d.ts
@@ -1131,8 +1131,8 @@ export interface Plugin {
     schema?: SchemaProperties;
 }
 
-export type Plugins = Plugin | NestedPlugins;
-export interface NestedPlugins extends Array<Plugins> {}
+export type PluginOrPlugins = Plugin | Plugins;
+export interface Plugins extends Array<PluginOrPlugins> {}
 
 export interface EditorProperties {
     onChange?: (change: { operations: Immutable.List<Operation>; value: Value }) => void;
--- a/types/slate/index.d.ts
+++ b/types/slate/index.d.ts
@@ -1135,7 +1135,7 @@ export type Plugins = Plugin | Plugin[] | PluginStack;
 export interface PluginStack extends Array<Plugins> {}
 
 export interface EditorProperties {
-    onChange?: (change: { operations: Immutable.List<Operation>, value: Value }) => void;
+    onChange?: (change: { operations: Immutable.List<Operation>; value: Value }) => void;
     plugins?: PluginStack;
     readOnly?: boolean;
     value?: Value;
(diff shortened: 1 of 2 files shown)


types/slate/index.d.ts
```diff
@@ -1131,9 +1131,12 @@ export interface Plugin {
     schema?: SchemaProperties;
 }
 
+export type Plugins = Plugin | Plugin[] | PluginStack;
```
hanstar17: I don't think Plugin[] is needed here because PluginStack covers it.
Is there any reason giving it a name 'stack'? How about CompositePlugin? Or, from the documentation, FeaturePlugin or FrameworkPlugin?

types/slate/index.d.ts
```diff
@@ -1131,9 +1131,12 @@ export interface Plugin {
...
+export interface PluginStack extends Array<Plugins> {}
+
 export interface EditorProperties {
     onChange?: (change: { operations: Immutable.List<Operation>, value: Value }) => void;
-    plugins?: Plugin[];
+    plugins?: PluginStack;
```
hanstar17: PluginStack is good but Plugins seems to be more intuitive here?

types/slate/index.d.ts
lukasbuenger: > I don't think Plugin[] is needed here because PluginStack covers it.
Good catch. Thanks!

> Is there any reason giving it the name 'stack'?
Maybe there's a better name than `PluginStack`, but I surely wouldn't use any of the names you suggested as they all imply some kind of behavior, which at this point can be misleading bit of info. It's a stack as opposed to the readonly property on the controller, which is always a flat list, so more about the possible data structure than its content. Maybe `NestedPlugins`?

types/slate/index.d.ts
lukasbuenger: Sure, why not.

lukasbuenger: Ugh, I got this wrong, sorry, I'm switching contexts a lot right now ;). Will fix asap.

types/slate/index.d.ts
```diff
@@ -1131,12 +1131,12 @@ export interface Plugin {
...
 
-export type Plugins = Plugin | Plugin[] | PluginStack;
-export interface PluginStack extends Array<Plugins> {}
+export type Plugins = Plugin | NestedPlugins;
+export interface NestedPlugins extends Array<Plugins> {}
 
 export interface EditorProperties {
     onChange?: (change: { operations: Immutable.List<Operation>; value: Value }) => void;
-    plugins?: PluginStack;
+    plugins?: Plugins;
```
hanstar17:
```ts
plugins?: Plugin | Plugins;
```

hanstar17: I think you can pass a plain Plugin, so
```ts
plugins?: Plugin | Plugins;
```
is correct?

hanstarj (changes requested):

types/slate/slate-tests.ts
```diff
@@ -114,7 +114,7 @@ const plugin: Plugin = {
     schema: {...schema},
 };
 
-const plugins = [plugin];
+const plugins = [plugin, [plugin, [plugin]]];
```
hanstar17: I think we are missing a test case ```plugins = plugin;```, which is why the new commit passed the test.

hanstar17:
I like the change, thanks for the work and also applying some of my feedback, I appreciate it. As soon as you fix the bug I will approve, as all look good to me!

types/slate/index.d.ts
lukasbuenger: > I think you can pass a plain Plugin

I really think you can't. 

Slate's `CorePlugin` most definitely expects a plugin array at `plugins`. Here's some pointers.
https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/plugins/core.js#L20  
https://github.com/ianstormtaylor/slate/blob/master/packages/slate/src/plugins/core.js#L200 

The «act as one plugin»-case is covered by the fact that the editor controller/component in `slate-react` accepts all properties  of a regular plugin. A case that @types/slate-react already tests. 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/slate-react/slate-react-tests.tsx#L49-L56

hanstarj (approved):
Hey, you are right! Thanks for letting me know!

types/slate/index.d.ts
hanstar17: Hey, you are right! Thanks for letting me know!

hanstarj (approved):

types/slate/slate-tests.ts
lukasbuenger: The new commit passed the test, because the fixtures were not invalid per se. I'm by no means a Typescript expert, not to mention DefinitelyTyped, so do you know of a way of testing a supposedly faulty test? 
All I get is the notion to provide code that does not fail neither `tslint` nor `tsc`.

hanstar17:
How do I approve this issue? I am still getting used to open source contribution :P

lukasbuenger:
> How do I approve this issue? I am still getting used to open source contribution :P

You're doing perfectly fine, mate! Thanks a lot for the feedback and for challenging Captain Obvious. You approve PR's by starting a review and, when you're finished, choose whether you approve or not. Just like you did.

types/slate/slate-tests.ts
hanstar17: Your changes are perfectly fine :) I wrongly thought that originally `plugins` were the type of `Plugin | Plugins[]`. Sorry for the confusion!!

lukasbuenger: At one point they were. 
Thas was part of my ...

> Ugh, I got this wrong, sorry

...initiative. *I* am sorry. And also, sorry for the confusion.

hanstar17:
> > How do I approve this issue? I am still getting used to open source contribution :P
> 
> You're doing perfectly fine, mate! Thanks a lot for the feedback and for challenging Captain Obvious. You approve PR's by starting a review and, when you're finished, choose whether you approve or not. Just like you did.

Thanks for warm words, dude! Have a nice day.

DanielRosenwasser:
I sent a quick commit to inline `PluginOrPlugins` since it was only used once.

PR #36380, opened 2019-06-23
Add types for `react-beforeunload`

nem035:
:tada: add types for [`react-beforeunload`](https://github.com/jacobbuck/react-beforeunload)

Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/react-beforeunload/index.d.ts
+++ b/types/react-beforeunload/index.d.ts
@@ -6,9 +6,9 @@
 
 import * as React from 'react';
 
-export function useBeforeunload(handler?: (arg?: Event) => string | void): void;
+export function useBeforeunload(handler?: (arg: Event) => string | void): void;
 
 export const Beforeunload: React.FC<{
     children?: React.ReactNode;
-    onBeforeunload: (arg?: Event) => string | void;
+    onBeforeunload: (arg: Event) => string | void;
 }>;
--- a/types/react-beforeunload/index.d.ts
+++ b/types/react-beforeunload/index.d.ts
@@ -6,9 +6,9 @@
 
 import * as React from 'react';
 
-export function useBeforeunload(handler: (arg: Event) => string | void): void;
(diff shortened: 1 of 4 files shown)


types/react-beforeunload/index.d.ts
```diff
@@ -0,0 +1,14 @@
...
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+import * as React from 'react';
+
+export function useBeforeunload(handler?: (arg: Event) => string | void): void;
+
+export const Beforeunload: React.FC<{
+    children?: React.ReactNode;
+    onBeforeunload: (arg: Event) => string | void;
```
uniqueiniquity: Please change this return type to `string | undefined`; `string | void` isn't really a meaningful type even though it seems like it might work. Thanks!

uniqueiniquity (changes requested):

types/react-beforeunload/index.d.ts
nem035: Good point, completely missed this 👍

nem035:
@uniqueiniquity just remembered that I had the `void` in `() => string | void` because [`event.preventDefault` is declared as returning `void`](https://github.com/microsoft/TypeScript/blob/master/src/lib/dom.generated.d.ts#L5266) (and one of the allowed usages in this lib is `e => e.preventDefault`.

Would it make sense to use `string | undefined | void` in this case?

uniqueiniquity:
@nem035 Ah, I see the use case. In this situation, I'd suggest using `type UseBeforeUnloadHandler = ((arg: Event) => string) | ((arg: Event) => void);` 
I realize it's not clear why that should be different from what you have, but `void` can behave pretty strangely in that way.

uniqueiniquity (approved):

uniqueiniquity:

PR #36398, opened 2019-06-24
ramda: add support of applying `ap` to function type argument

jituanlin:
Please fill in this template.

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://gist.github.com/jituanlin/168c2eb147f69b60cc063c08964cbf58>>

### DIFF
--- a/types/ramda/ramda-tests.ts
+++ b/types/ramda/ramda-tests.ts
@@ -2197,6 +2197,7 @@ class Rectangle {
 () => {
     const x: number[] = R.ap([R.multiply(2), R.add(3)], [1, 2, 3]); // => [2, 4, 6, 4, 5, 6]
     const y: number[] = R.ap([R.multiply(2), R.add(3)])([1, 2, 3]); // => [2, 4, 6, 4, 5, 6]
+    const z: string = R.ap<string, string, string>(R.concat, R.toUpper)('Ramda'); // => 'RamdaRAMDA'
 };
 
 () => {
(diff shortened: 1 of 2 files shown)


googol:
Looks good. Adding a test, for example the one in the documentation would be a good addition :+1:

DanielRosenwasser:
Can you add a test please?

jituanlin:
> Can you add a test please?

OK, I will finish the test lately.

types/ramda/ramda-tests.ts
```diff
@@ -2197,6 +2197,7 @@ class Rectangle {
 () => {
     const x: number[] = R.ap([R.multiply(2), R.add(3)], [1, 2, 3]); // => [2, 4, 6, 4, 5, 6]
     const y: number[] = R.ap([R.multiply(2), R.add(3)])([1, 2, 3]); // => [2, 4, 6, 4, 5, 6]
+    const z: string = R.ap<string, string, string>(R.concat, R.toUpper)('Ramda'); // => 'RamdaRAMDA'
```
DanielRosenwasser: Will this actually require providing type arguments? If not, please remove them. If so, then maybe we should consider the declaration here.

jituanlin: Provide the type arguments explicitly is required here.
This is cause by Typescript `strange` type infer behavior of overload function.
In this case, the `concat`  function not be infer `correctly`(you can see it be infer a very not make sense if you check the code in your editor), so we need provider type arguments
explicitly.

See [this issues](https://github.com/microsoft/TypeScript/issues/32081)

PR #36409, opened 2019-06-24
Added type definition for `codegen.macro`

charlee:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/codegen.macro/index.d.ts
+++ b/types/codegen.macro/index.d.ts
@@ -1,0 +1,7 @@
+// Type definitions for codegen.macro 3.0
+// Project: https://github.com/kentcdodds/codegen.macro
+// Definitions by: Charlee Li <https://github.com/charlee>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare function codegen(strings: TemplateStringsArray, ...values: any[]): any;
+export = codegen;
--- a/types/codegen.macro/codegen.macro-tests.ts
+++ b/types/codegen.macro/codegen.macro-tests.ts
@@ -1,0 +1,5 @@
+import codegen = require('codegen.macro');
+
+const eg1: string = codegen`module.exports = "'hello world'"`;
+const eg2: number = codegen`module.exports = "42"`;
+const eg3: { [key: string]: string } = codegen`module.exports = "({ hello: 'world' })"`;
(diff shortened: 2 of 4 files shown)


types/codegen.macro/index.d.ts
```diff
@@ -0,0 +1,7 @@
+// Type definitions for codegen.macro 3.0
+// Project: https://github.com/kentcdodds/codegen.macro
+// Definitions by: Charlee Li <https://github.com/charlee>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare function codegen(strings: TemplateStringsArray, ...values: any[]): any;
```
uniqueiniquity: I would suggest the return type be `unknown`. This way, consumers of the library will have to cast the result of `codegen` in order to use it. As you have it now, you can "cast" it using type annotations as you do in your tests, but if they don't do this, then there's no error and they might introduce `any` into their program.

uniqueiniquity (changes requested):

types/codegen.macro/index.d.ts
charlee: Yes totally agree. But do we still need to support TypeScript versions older than 3.0?

uniqueiniquity: Ah, good point, I sometimes forget how new `unknown` is. In that case, you could use `{} | null | undefined`.

charlee: I think the return value of `codegen` can be literally anything, not limited to `{}` or `null` or `undefined`.
I feel `any` should be a good representation of its return type, or `unknown` to be more accurate, but any specific types could limit its potential.

uniqueiniquity: `unknown` and `{} | null | undefined` are actually the same thing. But I understand your concern, so I'm fine merging either that or `any`; whichever you prefer.

charlee: Thanks for pointing that out. I tried `{} | null | undefined` and even `{} | null | undefined | number | string`, not sure if I was doing something wrong, but looks like it was not compatible with the types given in the test case. For example, `{}` is not assignable to type `{ [key: string]: string; }`.
Given that, I would like it to be `any` which makes it much simpler.

charlee:
Hi @uniqueiniquity , as commented above I would like to choose using `any`. Do I need to do anything else?

uniqueiniquity (approved):

uniqueiniquity:

uniqueiniquity:
Sorry for the delay!

PR #36412, opened 2019-06-24
Fix incorrect AddressType value for street_number

thewunder:
Please fill in this template.

- [ x ] Provide a URL to documentation or source code which provides context for the suggested changes: https://developers.google.com/places/web-service/details#PlaceDetailsResults

### DIFF
--- a/types/google__maps/index.d.ts
+++ b/types/google__maps/index.d.ts
@@ -808,8 +808,6 @@ export type GeocodedWaypointStatus = (
 );
 
 export type AddressType = (
-    /** indicates a precise street address. */
-    'street_address' |
     /** indicates a named route (such as "US 101"). */
     'route' |
     /** indicates a major intersection, usually of two major roads. */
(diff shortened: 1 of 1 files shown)


indrimuska (changes requested):
From the [official documentation](https://developers.google.com/places/web-service/supported_types), street_number and street_address are both available, so you should append the new type rather than replace the old one.

indrimuska:
Sorry, I had a mistake.

You are linking the documentation of the **PlaceDetailsResults** API, and as part of the `address_components` array, there is the `type` property. This latter one should be of type `PlaceType[]` (as per [doc](https://developers.google.com/places/web-service/details#PlaceDetailsResults)), instead of `AddressType[]` (that is used for [geocoding results](https://developers.google.com/maps/documentation/geocoding/intro#Types)).

Anyway `AddressComponent` has been defined here as a fixed type, and used in both geocoding and placetype api. Perhaps it should be a generic type, providing the given `type` property type (e.g. defined as`interface AddressComponent<T> { types: T[]; ... }` and used as `AddressComponent<PlaceType1 | PlaceType2>`).

thewunder:
@indrimuska Yes, this would work plus street_address would need to be moved into GeocodingAddressComponentType, and street_number must be moved into a new PlaceDetailsAddressComponentType

thewunder:
> From the [official documentation](https://developers.google.com/places/web-service/supported_types), street_number and street_address are both available, so you should append the new type rather than replace the old one.

In practice, it seems that place details returns street_number and never street_address

orta:
@thewunder sorry it took us a while to get to this, are you happy with this PR?

thewunder:
@orta Yeah I think this is a good fix

PR #36415, opened 2019-06-24
papaparse: add transformHeader option

dobesv:
Version 5 introduces a new option `transformHeader`.

See https://www.papaparse.com/docs#config

This PR also includes a fix for TS@next from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36470


dobesv:
If anyone knows the fix for that Travis failure I'd love to know.  It's not caused by my change, but rather a change in behavior of typescript.  Maybe remove that overload that allows 1 as a parameter?  Not sure what that would break.

PR #36416, opened 2019-06-24
Adds type definitions for 'node-zendesk'

jgeth:
Adds type definitions and tests
Adds tsconfig and tslint config

### DIFF
--- a/types/node-zendesk/index.d.ts
+++ b/types/node-zendesk/index.d.ts
@@ -7,6 +7,7 @@
 /// <reference types="node"/>
 
 import { PathLike } from 'fs';
+import { type } from 'os';
 
 export type ZendeskCallback<TResponse, TResult> = (
     error: Error | undefined,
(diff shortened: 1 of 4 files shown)


types/node-zendesk/index.d.ts
```diff
@@ -0,0 +1,659 @@
...
+            url?: string;
+            request_id?: number;
+            body?: string;
+            html_body?: string;
+            public?: boolean;
+            author_id?: ZendeskID;
+            uploads?: ReadonlyArray<string>;
+        }
+
+        // TODO: support VoiceComment model @jgeth 2019-01-22
```
ckknight: It'd probably be wise to notate all the TODOs in one place (probably near the top of the file).

For any types that are missing, it might be wise to expose them as either `any` or `unknown`.

types/node-zendesk/index.d.ts
```diff
@@ -0,0 +1,659 @@
...
+            readonly html_body: string;
+            readonly plain_body: string;
+            readonly public: boolean;
+            readonly author_id: ZendeskID;
+            readonly attachments: ReadonlyArray<Attachments.Model>;
+            readonly via?: Tickets.Via;
+            readonly metadata?: Tickets.Comments.Metadata;
+        }
+
+        enum Type {
```
ckknight: Is this an enum that exists at runtime? I can't find it within https://github.com/blakmatrix/node-zendesk

If it isn't, it needs to either be a `const enum` or a type alias.

ckknight: This comment probably applies to all uses of `enum`

ckknight (changes requested):

types/node-zendesk/index.d.ts
jgeth: enums have been replaced with type aliases consisting of unions of literals.

types/node-zendesk/index.d.ts
jgeth: Resolved this sole TODO

types/node-zendesk/index.d.ts
```diff
@@ -7,6 +7,7 @@
 /// <reference types="node"/>
 
 import { PathLike } from 'fs';
+import { type } from 'os';
```
ckknight: This doesn't seem right.

ckknight (changes requested):

ckknight (approved):
Looks good. Thanks, @jgeth.

PR #36420, opened 2019-06-25
[@wordpress/blocks] add new definitions

dsifford:
Please fill in this template.

If adding a new definition:


types/wordpress__blocks/index.d.ts
```diff
@@ -0,0 +1,403 @@
...
+     */
+    readonly parent?: readonly string[];
+    /**
+     * This is set internally when registering the type.
+     */
+    readonly name: string;
+    /**
+     * Component to render on the frontend.
+     */
+    readonly save: ComponentType<BlockSaveProps<T>>;
```
zanona: hey @dsifford, sorry to reach you here regarding this. But I was wondering how can I correctly type an external function to match the `save` method available under block?

I am not too advanced in Typescript and wondered if this declaration (`readonly save: ComponentType<BlockSaveProps<T>>;`) is the same of:

```ts
import {ComponentType} from '@wordpress/element';
interface BlockData { foo:string }
function save(props:<BlockSaveProps<BlockData>): ComponentType { ... }
```

although I couldn't get it working unless I would do:

```ts
import {ReactElement} from '@wordpress/element';
interface BlockData { foo:string }
function save(props:<BlockSaveProps<BlockData>): ReactElement { ... }
```
but then I fear I may be losing some validations with this?

It would be great in case you could help.

All the best

zanona: Just a small update here as I did look for the `ComponentType` definition on React's declaration and did find this:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7d4caa8dd6f01035092c0b28b6b160ac93f35ca1/types/react/index.d.ts#L164

which solved my problem definition the function as:

```ts
import {FunctionComponentElement} from '@wordpress/element';
interface BlockData { foo:string }
function save(props:<BlockSaveProps<BlockData>): FunctionComponentElement<BlockData> { ... }
```
I'm not sure this is correct, but hopefully already a bit better than the general `ReactElement` solution?

Cheers

dsifford: @zanona `ComponentType` is just a union of `ComponentClass<Props>` and `FunctionComponent<Props>` so with that in mind, if we look at your definition of...

```ts
function save(props:<BlockSaveProps<BlockData>): ComponentType;
```
What this really amounts to is one of the following...

```tsx
function save(props) {
  return function(props)  {
    return <h1>Hello world</h1>;
  }
}
```

or

```tsx
function save(props) {
  return class extends Component {
    render() {
      return <h1>Hello world</h1>;
    }
  }
}
```

Which is not correct. 

Instead, your type should look like this...

```tsx
import {ComponentType} from '@wordpress/element';
interface BlockData { foo:string }
function save(props:<BlockSaveProps<BlockData>): JSX.Element;
//       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Which would amount to being...

```tsx
function save(props) {
  return <h1>Hello world</h1>;
}
```

Looking at the modified definition, everything above the `^`s is part of the `ComponentType` definition.

Hope that makes sense.

zanona: Wow, that's great! Thanks so much for the explanation. I didn't know about `JSX.Element`  ;)

types/wordpress__blocks/index.d.ts
```diff
@@ -0,0 +1,403 @@
...
+
+declare module '@wordpress/data' {
+    function dispatch(key: 'core/blocks'): typeof import('./store/actions');
+    function select(key: 'core/blocks'): typeof import('./store/selectors');
+}
+
+export type BlockAlignment = 'left' | 'center' | 'right' | 'wide' | 'full';
+
+export interface BlockEditProps<T extends Record<string, any>> extends BlockSaveProps<T> {
+    readonly className: string;
```
zanona: Apparently, we now also have access to the following properties sent to the edit function:
`clientId`,
`insertBlocksAfter`,
`isSelectionEnabled`,
`mergeBlocks`,
`name`,
`onReplace` and
`toggleSelection`.

types/wordpress__blocks/index.d.ts
```diff
@@ -0,0 +1,403 @@
...
+    background?: string;
+    foreground?: string;
+    shadowColor?: string;
+    src: Dashicon.Icon | ComponentType;
+}
+
+export type BlockIcon = BlockIconNormalized['src'] | BlockIconNormalized;
+
+export interface BlockSaveProps<T extends Record<string, any>> {
+    readonly attributes: Readonly<T>;
```
zanona: and whenever using `InnerBlocks` an `innerBlocks` property (array) is also sent to the save function.

types/wordpress__blocks/index.d.ts
dsifford: No time at the moment to get to this, but feel free to handle and open a PR.

PR #36434, opened 2019-06-25
fix(slonik): adjust types according to slonik unit tests and documentation

n1ru4l:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/gajus/slonik#slonik-query-methods

If removing a declaration:

### DIFF
--- a/types/slonik/index.d.ts
+++ b/types/slonik/index.d.ts
@@ -221,15 +221,15 @@ export type QueryResultRowType<ColumnName extends string = string> = {
     [name in ColumnName]: QueryResultRowColumnType;
 };
 
-export type QueryAnyFirstFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<Array<T[keyof T]>>;
-export type QueryAnyFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<T[]>;
-export type QueryFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<QueryResultType<T>>;
-export type QueryManyFirstFunctionType = QueryAnyFirstFunctionType;
+export type QueryAnyFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<ReadonlyArray<T>>;
+export type QueryAnyFirstFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<ReadonlyArray<T>>;
 export type QueryManyFunctionType = QueryAnyFunctionType;
-export type QueryMaybeOneFirstFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<T[keyof T] | null>;
+export type QueryManyFirstFunctionType = QueryAnyFirstFunctionType;
 export type QueryMaybeOneFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<T | null>;
-export type QueryOneFirstFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<T[keyof T]>;
+export type QueryMaybeOneFirstFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<T | null>;
 export type QueryOneFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<T>;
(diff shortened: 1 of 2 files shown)


types/slonik/index.d.ts
```diff
@@ -221,15 +221,15 @@ export type QueryResultRowType<ColumnName extends string = string> = {
...
-export type QueryFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<QueryResultType<T>>;
-export type QueryManyFirstFunctionType = QueryAnyFirstFunctionType;
+export type QueryAnyFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<ReadonlyArray<T>>;
+export type QueryAnyFirstFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<ReadonlyArray<T>>;
 export type QueryManyFunctionType = QueryAnyFunctionType;
-export type QueryMaybeOneFirstFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<T[keyof T] | null>;
+export type QueryManyFirstFunctionType = QueryAnyFirstFunctionType;
 export type QueryMaybeOneFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<T | null>;
-export type QueryOneFirstFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<T[keyof T]>;
+export type QueryMaybeOneFirstFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<T | null>;
```
mmkal: This isn't right. `T[keyof T]` was correct. The `...First` functions return the first column value of the row objects. Which means if `T = {foo: string}`, `T[keyof T] = string`.

types/slonik/index.d.ts
```diff
@@ -221,15 +221,15 @@ export type QueryResultRowType<ColumnName extends string = string> = {
     [name in ColumnName]: QueryResultRowColumnType;
 };
 
-export type QueryAnyFirstFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<Array<T[keyof T]>>;
-export type QueryAnyFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<T[]>;
-export type QueryFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<QueryResultType<T>>;
-export type QueryManyFirstFunctionType = QueryAnyFirstFunctionType;
+export type QueryAnyFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<ReadonlyArray<T>>;
+export type QueryAnyFirstFunctionType = <T>(...args: QueryMethodParams<T>) => Promise<ReadonlyArray<T>>;
```
mmkal: These aren't readonly arrays, and there's no reason to type them that way. You lose the ability to use utility functions for regular arrays by doing this.

types/slonik/slonik-tests.ts
```diff
@@ -110,18 +110,18 @@ const typedQuery = async () => {
   await pool.query(getFooBarQuery(10));
 
   // $ExpectType string
-  await pool.oneFirst(getFooQuery(10));
+  await pool.oneFirst<string>(getFooQuery(10));
```
mmkal: The generic in all these query methods is supposed to represent the row type, which could never be string.

mmkal (changes requested):
These changes don't reflect how the library works - if there's something you're finding is going wrong for you, let me know and I can help with adding a test case, or some documentation to the types.

n1ru4l:
@mmkal

How would you type this correctly?

```ts
const amountOfRecords: number = ((await this.slonik.oneFirst(sql`
  SELECT COUNT(*) FROM "records"
  WHERE "ownerId" = 1;
`)) as unknown) as number;
```

mmkal:
@n1ru4l 
```typescript
const amountOfRecords = await slonik.oneFirst<{records: number}>(sql`
  SELECT COUNT(*) FROM "records"
  WHERE "ownerId" = 1;
`)
```

I created a package under the `@slonik` org which will do that automatically for you. It's very new, but feel free to try it out and see if it works for you: https://npmjs.com/package/@slonik/typegen

PR #36435, opened 2019-06-25
[electron-packager] Add all electron-osx-sign options

ffflorian:
Please fill in this template.

  - https://github.com/electron-userland/electron-osx-sign/blob/master/index.d.ts#L26
  - https://github.com/electron-userland/electron-packager/blob/master/src/mac.js#L10
  - https://github.com/electron-userland/electron-packager/blob/master/src/mac.js#L293
  - https://github.com/electron-userland/electron-packager/blob/master/src/mac.js#L346

### DIFF
--- a/types/electron-packager/index.d.ts
+++ b/types/electron-packager/index.d.ts
@@ -5,6 +5,7 @@
 //                 John Kleinschmidt <https://github.com/jkleinsc>
 //                 Brendan Forster <https://github.com/shiftkey>
 //                 Mark Lee <https://github.com/malept>
+//                 Florian Keller <https://github.com/ffflorian>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 /// <reference types="node" />
@@ -57,9 +58,19 @@ declare namespace electronPackager {
     }
 
     interface ElectronOsXSignOptions {
-        identity?: string;
         entitlements?: string;
         "entitlements-inherit"?: string;
+        "gatekeeper-assess"?: boolean;
+        // see https://github.com/electron-userland/electron-packager/blob/92d09bba34599283a794fd6f24b88470f0cb1074/src/mac.js#L359
+        identity?: string | true;
+        "identity-validation"?: boolean;
+        ignore?: string;
+        keychain?: string;
+        "pre-auto-entitlements"?: boolean;
+        "pre-embed-provisioning-profile"?: boolean;
+        "provisioning-profile"?: string;
+        requirements?: string;
+        type?: string;
     }
 
     /**


types/electron-packager/index.d.ts
```diff
@@ -57,9 +57,18 @@ declare namespace electronPackager {
     }
 
     interface ElectronOsXSignOptions {
-        identity?: string;
         entitlements?: string;
         "entitlements-inherit"?: string;
+        identity?: string | boolean;
```
cortopy: [electron-osx-sign](https://www.npmjs.com/package/electron-osx-sign#opts) docs say that identity is till a string. Could you clarify?

cortopy (changes requested):

types/electron-packager/index.d.ts
ffflorian: Please see `mac.js` lines 356-361:

https://github.com/electron-userland/electron-packager/blob/92d09bba34599283a794fd6f24b88470f0cb1074/src/mac.js#L356-L361

cortopy: awesome, thanks!

cortopy (approved):

types/electron-packager/index.d.ts
ffflorian: The more I think about it, it should actually be `string | true`, what do you think?

cortopy: Definitely, well spotted. There doesn't seem to be logic for handling `false`. Could you also add an inline comment with the link you shared here for future reference?

ffflorian: Done :slightly_smiling_face:

cortopy: :+1:

types/electron-packager/index.d.ts
```diff
@@ -57,9 +58,19 @@ declare namespace electronPackager {
...
+        // see https://github.com/electron-userland/electron-packager/blob/92d09bba34599283a794fd6f24b88470f0cb1074/src/mac.js#L359
+        identity?: string | true;
+        "identity-validation"?: boolean;
+        ignore?: string;
+        keychain?: string;
+        "pre-auto-entitlements"?: boolean;
+        "pre-embed-provisioning-profile"?: boolean;
+        "provisioning-profile"?: string;
+        requirements?: string;
+        type?: string;
```
malept: Is there a way to say something like

```typescript
import { SignOptions } from 'electron-osx-sign';

// ...
interface ElectronOsxSignOptions extends SignOptions {
  identity?: string | true;
}
```

and have `identity` get overridden properly by TypeScript? Personally I'd rather have as little copypasta as possible.

malept (changes requested):

types/electron-packager/index.d.ts
malept: Oh, the change I described is in #36441? I'm a little confused.

ffflorian: @malept #36441 will be merged after this PR.

I'm also against copypasta, but to import `electron-osx-sign`, https://github.com/microsoft/types-publisher/pull/627 needs to be merged. And this sometimes takes a few days... :slightly_smiling_face:

cortopy: @malept is that valid typescript? See this question in [stackoverflow](https://stackoverflow.com/questions/41285211/overriding-interface-property-type-defined-in-typescript-d-ts-file)

What about once the types get published:

```typescript
import { SignOptions } from 'electron-osx-sign';

interface ElectronOsxSignOptions extends Omit<SignOptions, 'identity'>  {
  identity?: string | true;
}
```

malept: > is that valid typescript?

I did say "something like" as my TypeScript skills are not very good. @ffflorian answered my question.

malept (approved):
This is fine given the existence of #36441.

PR #36444, opened 2019-06-25
Updated type defs for @testing-library/user-event

weyert:
Please fill in this template.

If adding a new definition:

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/testing-library/user-event/releases/tag/v4.1.0

If removing a declaration:

### DIFF
--- a/types/testing-library__user-event/index.d.ts
+++ b/types/testing-library__user-event/index.d.ts
@@ -3,17 +3,21 @@
 // Definitions by: Wu Haotian <https://github.com/whtsky>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export interface UserOpts {
+export interface IUserOptions {
     allAtOnce?: boolean;
     delay?: number;
 }
+
+type TargetElement = Element | Window;
+
 declare const userEvent: {
-    click: (element: Element | Window) => void;
-    dblClick: (element: Element | Window) => void;
+    click: (element: TargetElement) => void;
+    dblClick: (element: TargetElement) => void;
+    selectOptions: (element: TargetElement, values: string | string[]) => void;
     type: (
-        element: Element | Window,
+        element: TargetElement,
         text: string,
-        userOpts?: UserOpts
+        userOpts?: IUserOptions
     ) => Promise<void>;
 };
 
--- a/types/testing-library__user-event/index.d.ts
+++ b/types/testing-library__user-event/index.d.ts
@@ -3,11 +3,6 @@
 // Definitions by: Wu Haotian <https://github.com/whtsky>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-
-
-
-
-
 export interface UserOpts {
     allAtOnce?: boolean;
     delay?: number;
--- a/types/testing-library__user-event/index.d.ts
+++ b/types/testing-library__user-event/index.d.ts
@@ -1,5 +1,4 @@
 // Type definitions for @testing-library/user-event 4.1
-
 // Project: https://github.com/testing-library/user-event
 // Definitions by: Wu Haotian <https://github.com/whtsky>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
@@ -13,7 +12,6 @@ declare const userEvent: {
     click: (element: Element | Window) => void;
     dblClick: (element: Element | Window) => void;
     selectOptions: (element: Element | Window, values: string | string[]) => void;
-
     type: (element: Element | Window, text: string, userOpts?: IUserOptions) => Promise<void>;
 };
 
--- a/types/testing-library__user-event/index.d.ts
+++ b/types/testing-library__user-event/index.d.ts
@@ -6,6 +6,8 @@
 
 
 
+
+
 export interface UserOpts {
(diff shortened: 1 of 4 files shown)


johnnyreilly:
Heya!

Looks like there's some errors:

```
Error in testing-library__user-event
Error: /home/travis/build/DefinitelyTyped/DefinitelyTyped/types/testing-library__user-event/index.d.ts:6:18
ERROR: 6:18   interface-name  interface name must not have an "I" prefix
ERROR: 18:26  eofline         file should end with a newline
/home/travis/build/DefinitelyTyped/DefinitelyTyped/types/testing-library__user-event/testing-library__user-event-tests.ts:5:1
ERROR: 5:1    expect          TypeScript@next expected type to be:
  void
got:
  any
ERROR: 5:11   expect          TypeScript@next compile error: 
Property 'selectOptions' does not exist on type '{ click: (element: Element | Window) => void; dblClick: (element: Element | Window) => void; type: (element: Element | Window, text: string, userOpts?: UserOpts | undefined) => Promise<...>; }'.
ERROR: 6:1    expect          TypeScript@next expected type to be:
  void
got:
  any
ERROR: 6:11   expect          TypeScript@next compile error: 
Property 'selectOptions' does not exist on type '{ click: (element: Element | Window) => void; dblClick: (element: Element | Window) => void; type: (element: Element | Window, text: string, userOpts?: UserOpts | undefined) => Promise<...>; }'.
```

https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/550307788?utm_source=github_status&utm_medium=notification

types/testing-library__user-event/index.d.ts
```diff
@@ -1,20 +1,18 @@
-// Type definitions for @testing-library/user-event 4.0
+// Type definitions for @testing-library/user-event 4.1
 // Project: https://github.com/testing-library/user-event
 // Definitions by: Wu Haotian <https://github.com/whtsky>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export interface UserOpts {
+export interface IUserOptions {
```
whtsky: any particular reason for changing the interface's name?

IMO it's a breaking change and the "I" prefix is making tslint unhappy

weyert: Yes, sorry, on my computer it was complaining about the I prefix within my project. I have resolved it now :)

Regarding the name just felt better to me but reverted the name change

johnnyreilly:
Okay it looks like I was wrong about the relative import.... Want to substitute for the namespaced one instead?

johnnyreilly:
Maybe it doesn't need the prefix? Not sure...

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/babel-types/index.d.ts

johnnyreilly:
Huh failed. Weird. Maybe we import fron: `testing-library__user-event`?

johnnyreilly:
Or do a `require` like in babel-template: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/babel-template/babel-template-tests.ts

I think that's an example of a namespaced / scoped definition

johnnyreilly:
@andrewbranch @drosenwasser - sorry to bother you but we're bumping up against something I haven't had to do before. Namely, write tests for a scoped / namespaced package. 

We don't seem to be able to import the type definition in our tests. We've tried:

- `import ... from 'user-event'` - this is what is committed now and seems incorrect; it refers to the legacy (pre-scoped) location of the package
- `import ... from './` - this is not allowed
- `import ... from '@testing-library/user-event' - this seems correct but does not work.

I feel like we're missing a trick / technique here and I was wondering if you knew what it was.  I've tried inspecting other scoped packages on DT but the answer has not presented itself...

I suspect we could hack this by just updating the legacy definition of `user-event` but that doesn't seem like the right thing to do 🤗

sandersn:
@johnnyreilly I think you have to put explicit entries in tsconfig. Take a look at `@babel/core` for an example.

andrewbranch:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/efa6b9e11eb0167547d6611490be48b94bc8d7ea/types/babel__core/tsconfig.json#L18-L31

johnnyreilly:
Aha! The ol' `paths` magic! Thanks @sandersn and @andrewbranch!

@weyert shall we give this a go?

whtsky (approved):

johnnyreilly:
@sandersn @andrewbranch - unfortunately the paths magic didn't work 😢 

We tried approach suggested:

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36444/commits/3c9cb6d021e69fb56c91c0cdd1d82d788ae662ea

But we still ended up seeing the `Cannot find module '@testing-library/user-event'` error message:

https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/550712352#L704

@weyert has (for now) switched to use the exisiting approach where we point at the existing `user-event` module.  I'm going to merge so I can unblock him.  

But I'd like to understand why we couldn't get this to work.  As far as I can tell, we did exactly the same approach as the `@babel/core` and yet it didn't work.  I've picked through this line by line but kind see any reason for failure. Do you have any insights?

PR #36453, opened 2019-06-25
Add WindowOrWorkerGlobalScope.fetch

sandersn:
This is compatible with TS3.6 and above, which remove GlobalFetch.

I didn't remove GlobalFetch from react-native, though, since I want it to be compatible with lower versions of typescript.

### DIFF
--- a/types/react-native/globals.d.ts
+++ b/types/react-native/globals.d.ts
@@ -31,6 +31,10 @@ declare interface GlobalFetch {
 
 declare function fetch(input: RequestInfo, init?: RequestInit): Promise<Response>;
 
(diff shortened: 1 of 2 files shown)


sandersn:
@SimenB this should fix your problem when it ships.

SimenB:
Thanks!

alloy:
I’m unsure of the details of the change without any further context, but if the react-native tests still pass then I’m fine with this change.

alloy (approved):
I’m unsure of the details of the change without any further context, but if the react-native tests still pass then I’m fine with this change.

PR #36469, opened 2019-06-26
[react] add createFactory(type: ComponentType) definition

tpucci:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://reactjs.org/docs/react-api.html#createfactory


types/react/index.d.ts
```diff
@@ -209,7 +210,7 @@ declare namespace React {
         type: ClassType<P, ClassicComponent<P, ComponentState>, ClassicComponentClass<P>>): CFactory<P, ClassicComponent<P, ComponentState>>;
     function createFactory<P, T extends Component<P, ComponentState>, C extends ComponentClass<P>>(
         type: ClassType<P, T, C>): CFactory<P, T>;
-    function createFactory<P>(type: ComponentClass<P>): Factory<P>;
+    function createFactory<P>(type: ComponentType<P>): Factory<P>;
```
ferdaber: I would actually use `JSXElementConstructor` instead, actually. It's more broad and doesn't require passing in function/class components that have the correct static properties).

tpucci: Hi! I tried to replace with `ComponentClass` with `JSXElementConstructor`.
Then I encounter the following error:

```
ERROR: 254:7  expect  TypeScript@next compile error: 
Type 'Factory<SCProps>' is not assignable to type 'FunctionComponentFactory<SCProps>'.
  Type 'ReactElement<SCProps, string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<any, any, any>)>' is not assignable to type 'FunctionComponentElement<SCProps>'.
    Types of property 'type' are incompatible.
      Type 'string | ((props: any) => ReactElement<any, string | ... | (new (props: any) => Component<any, any, any>)> | null) | (new (props: any) => Component<any, any, any>)' is not assignable to type 'FunctionComponent<SCProps>'.
        Type 'string' is not assignable to type 'FunctionComponent<SCProps>'.
ERROR: 259:7  expect  TypeScript@next compile error: 
Type 'Factory<SCProps>' is not assignable to type 'FunctionComponentFactory<SCProps>'.
```

I am not sure how to solve this problem. Any guess ?

RyanCavanaugh:
It'd be great to add a test so it's obvious to reviewers why this change is necessary

RyanCavanaugh (changes requested):
We really need tests for any changes for a definition file this popular

tpucci:
> We really need tests for any changes for a definition file this popular

I added a test. Hope everything is OK 😄 

Perhaps there is something I am missing but is there a way to efficiently test such type definition ?

I keep running `yarn test` at the root of the repo but it takes some times 🕐

PR #36470, opened 2019-06-26
papaparse: Declare parse/unparse overloads all at once

dobesv:
There was a problem in TS@next with the TS checker choosing the wrong overload.  Combining all the variants into one declaration seems to fix it.


johnnyreilly:
Thanks!

PR #32659, opened 2019-01-31
[react] Breaking! Update hooks API to (what's expected to be) the final 16.8.0 API

Jessidhia:
**NOTE**: This is a **breaking change** to current users of the hooks API if they pass explicit generics to `useReducer`, or if they use `useReducer`'s third argument. Two-argument `useReducer` is unaffected. If you pass explicit generic arguments to `useReducer`, uh, don't do that.

**NOTE 2**: The updated `useReducer` is _not available in any alpha_, it is the `useReducer` that will be in the stable release of hooks _currently_ due Monday.

---

I also did a small rename pass changing the `inputs` wording to `deps` as was also changed internally in React.

---

I'll preferably merge this when a build is actually available with the new `useReducer`. In the interim, the type for `useDebugValue`, which is available in `-alpha.1`, can be added with an ambient declaration:

```ts
import {} from 'react' // this is important

declare module 'react' {
  // this is written as a 'const' to break when this PR is merged
  export const useDebugValue: <T>(value: T, format?: (value: T) => any) => void
}
```

PR #36476, opened 2019-06-26
[react] simplify `useReducer` generics.

saltycrane:
Change generics for `useReducer` from

```
useReducer<Reducer<State, Action>>
```

to

```
useReducer<State, Action>
```

This is closer to [Flow's definitions for `useReducer`](https://github.com/facebook/flow/blob/88762e497b36796b24bc307793d97497168c05a6/lib/react.js#L329-L344)

I see the [previous method was done to support autocompletion](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32659). Is the autocompletion still broken with this method?

<hr/>

- [x] (Run with `npm test`.)

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -810,10 +810,6 @@ declare namespace React {
     type Dispatch<A> = (value: A) => void;
     // Unlike redux, the actions _can_ be anything
     type Reducer<S, A> = (prevState: S, action: A) => S;
-    // types used to try and prevent the compiler from reducing S
-    // to a supertype common with the second argument to useReducer()
-    type ReducerState<R extends Reducer<any, any>> = R extends Reducer<infer S, any> ? S : never;
-    type ReducerAction<R extends Reducer<any, any>> = R extends Reducer<any, infer A> ? A : never;
     // The identity check is done with the SameValue algorithm (Object.is), which is stricter than ===
     // TODO (TypeScript 3.0): ReadonlyArray<unknown>
     type DependencyList = ReadonlyArray<any>;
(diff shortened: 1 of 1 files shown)


types/react/index.d.ts
```diff
@@ -897,18 +893,11 @@ declare namespace React {
 
     // I'm not sure if I keep this 2-ary or if I make it (2,3)-ary; it's currently (2,3)-ary.
     // The Flow types do have an overload for 3-ary invocation with undefined initializer.
-
-    // NOTE: without the ReducerState indirection, TypeScript would reduce S to be the most common
```
eps1lon: This note makes it sound like that the previous behavior was intentional. Could you do a git blame and ping the person that added this? I think this is not about type checking but autocompletion only which has to be tested manuall.y

eps1lon (changes requested):

saltycrane:
@Jessidhia does this PR cause autocompletion to break for you?

types/react/index.d.ts
Jessidhia: It seems that, while autocomplete _does_ still work, the behavior is... strange.

Here is autocomplete working; note the inferred S generic:
image

But note how it is inferred after you enter the initializer:

image

---

Like, it's not **wrong**, but I wonder why does the object prevent the type alias from being used? The generic is invariant here (you can't either omit or add keys), but typescript seems to still construct a singleton common subtype instead of reusing the alias.

saltycrane: I see the difference:

**current behavior in master**
Screen Shot 2019-07-03 at 11 46 22 AM

**new behavior in my branch**
Screen Shot 2019-07-03 at 11 48 20 AM

PR #36479, opened 2019-06-27
Update index.d.ts

mattirizarry:
Math.eval doesnt work for some reason, this fixes it

Please fill in this template.

- [ x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x ] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If adding a new definition:
- [ ] The package does not already provide its own types, or cannot have its `.d.ts` files generated via

If removing a declaration:

### DIFF
--- a/types/mathjs/index.d.ts
+++ b/types/mathjs/index.d.ts
@@ -319,7 +319,7 @@ declare namespace math {
          * @param scope Scope to read/write variables
          * @returns The result of the expression
          */
-        eval(
+        evaluate(
             expr: MathExpression | MathExpression[] | Matrix,
             scope?: object
         ): any;


pawkrol:
Hi @mattirizarry, thanks for the PR!

Note that there are multiple mentions and declarations (like docs and methods) of `eval` function across the edited file, please adjust them as well if it's required.

andnp:
Why are we getting rid of `math.eval`?

andnp (changes requested):
Why are we getting rid of `math.eval`? As far as I am aware this is still a valid function. Should we add a new definition for `math.evaluate`?

pawkrol:
@andnp I believe that the required change is dictated by [this](https://github.com/josdejong/mathjs/blob/41efed79d3fcdcd5043952378f089df662d94ab1/src/expression/node/Node.js#L35)
```javascript
// TODO: Deprecated since v6.0.0. Clean up some day
Node.prototype.eval = function (scope) {
    warnOnce('Method Node.eval is renamed to Node.evaluate. Please use the new method name.')
...
```
and [this](https://github.com/josdejong/mathjs/blob/41efed79d3fcdcd5043952378f089df662d94ab1/src/expression/Parser.js#L114)
```javascript
// TODO: Deprecated since v6.0.0. Clean up some day
Parser.prototype.eval = function (expr) {
    warnOnce('Method Parser.eval is renamed to Parser.evaluate. Please use the new method name.')

...
```
Math.js lib changes, backed by [this](https://github.com/josdejong/mathjs/blob/ca933489d68225db2dc3bca4997a5797250ac0c1/HISTORY.md#breaking-changes) changelog

andnp:
Got it, thanks! Definitely helpful to have links to the docs in the PR. I think I still vote to keep support for `eval` given that it is still used on the front-page of the [math.js docs](https://mathjs.org/) in the "example" box. However, I don't feel too strongly either way.

@mattirizarry Quick question. Was it specifically `math.eval` that wasn't working as listed in your description? What was the specific issue that you were seeing? How did changing `math.eval` to `math.evaluate` fix the issue, or did you additionally change your code to use `math.evalutate` as well?

Admittedly, I'm a bit confused as to the original impetus of the change. I'm a bit wary that the original issues that led to the change are now being covered up by implicit `any`s, and that there was another issue at play.

---

Also. We need to bump the version of the `mathjs` types to `6.0` in this PR.

If we are to drop `math.eval` in favor of `math.evaluate`, that would be a breaking change in the type definitions that is unnecessary from the perspective of the original library. I'm okay with that because the original library is logging deprecation warnings; however, we need to make sure we match the correct major version number of the types and the library before making that change.

pawkrol:
@andnp Definitely agree with you, and I also believe that we should keep the `eval` function present. It is also present in the `math.js` lib, so as you said it will give warnings and will be easier to move from `eval` to `evaluate` for some projects.

PR #36493, opened 2019-06-27
Add missed path mapping to wordpress__plugins

sandersn:


dsifford (approved):
Thanks! :+1:

dsifford:
@sandersn While we're on this topic, would there be any way in the future for `types-publisher` to support the glob-based path mapping?

So, instead of what we have here, this...

```json
{
  "paths": {
    "@wordpress/*": [ "wordpress__*"]
  }
}
```

I tried in an earlier PR and it didn't work. Not sure how difficult that would be technically on the end of `types-publisher`, but it would definitely prevent these sorts of issues as more of these packages get merged.

sandersn:
That's a good idea. I filed https://github.com/microsoft/types-publisher/issues/628 to track that. I suspect that fixing it is not too hard.

PR #36499, opened 2019-06-27
[mailgun-js] Add types for generic requests 

ZhangYiJiang:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://bojand.github.io/mailgun-js/#/generic. Also allow `string` to be used for `o:tag` which is not explicitly documented but is allowed simply because `string` and `[string]` are treated the same by the library 

### DIFF
--- a/types/mailgun-js/index.d.ts
+++ b/types/mailgun-js/index.d.ts
@@ -3,7 +3,7 @@
 // Definitions by: Sampson Oliver <https://github.com/sampsonjoliver>
 //                 Andi Pätzold <https://github.com/andipaetzold>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.2
+// TypeScript Version: 2.3
 
 /// <reference types="node" />
 
@@ -55,6 +55,11 @@ declare namespace Mailgun {
     interface MailgunExport {
         new (options: ConstructorParams): Mailgun;
         (options: ConstructorParams): Mailgun;
+    }
+
+    interface MailgunRequest {
+        (resource: string, data: any, callback: (error: Error, response: any) => void): void;
+        (resource: string, data: any): Promise<any>;
     }
 
     namespace messages {
@@ -194,6 +199,12 @@ declare namespace Mailgun {
         validate(address: string, opts?: validation.ValidationOptionsPublic): Promise<validation.ValidateResponse>;
         validate(address: string, isPrivate: false, opts?: validation.ValidationOptionsPublic): Promise<validation.ValidateResponse>;
         validate(address: string, isPrivate: true, opts?: validation.ValidationOptionsPrivate): Promise<validation.ValidateResponse>;
+
+        // Generic requests
+        get: MailgunRequest;
+        post: MailgunRequest;
+        put: MailgunRequest;
+        delete: MailgunRequest;
     }
 
     interface Lists {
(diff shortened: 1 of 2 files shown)


types/mailgun-js/mailgun-js-tests.ts
```diff
@@ -76,3 +76,20 @@ const validationResult6: mailgunFactory.validation.ValidateResponse = {
...
+// Generic requests
+mailgun.get('/samples.mailgun.org/stats', { event: ['sent', 'delivered'] }, (error, body) => {
+    if (error) {
+        console.error(error);
+    } else {
+        console.log(body);
+    }
+});
+
+mailgun.get('/samples.mailgun.org/stats', { event: ['sent', 'delivered'] })
```
andipaetzold: The code is never executed. You could simply write `const getResponse: Promise<any> = ...`. This would save some lines that don't add any value.

types/mailgun-js/mailgun-js-tests.ts
```diff
@@ -76,3 +76,20 @@ const validationResult6: mailgunFactory.validation.ValidateResponse = {
         local_part: "foo"
     }
 };
+
+// Generic requests
+mailgun.get('/samples.mailgun.org/stats', { event: ['sent', 'delivered'] }, (error, body) => {
```
andipaetzold: You don't need to add any logic to the callback. `(err, body) => {}` is enough.
Or `(err: Error, body: any) => {}`. That way we have type checking for the callback function.

types/mailgun-js/index.d.ts
```diff
@@ -3,7 +3,7 @@
 // Definitions by: Sampson Oliver <https://github.com/sampsonjoliver>
 //                 Andi Pätzold <https://github.com/andipaetzold>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.2
```
andipaetzold: What's the reason for the version change? Which type definition requires TypeScript 2.3?

andipaetzold (changes requested):
Thanks for the contribution. The actual types look good to me. Just minor comments/questions.

types/mailgun-js/index.d.ts
ZhangYiJiang: When running the linter it says that this changeset fails for 2.2 but passes on 2.3. I'm not actually sure which specific feature requires 2.3, but it is reasonably old that I don't think it makes a big difference?

andipaetzold: Strange - I just tested it and linting worked for both versions 🤔 
But as you said, it's a quite old version (Feb 2017 vs April 2017).

types/mailgun-js/mailgun-js-tests.ts
```diff
@@ -76,3 +76,7 @@ const validationResult6: mailgunFactory.validation.ValidateResponse = {
         local_part: "foo"
     }
 };
+
+// Generic requests
+mailgun.get('/samples.mailgun.org/stats', { event: ['sent', 'delivered'] }, (error: any, body: any) => { });
```
andipaetzold:
```suggestion
mailgun.get('/samples.mailgun.org/stats', { event: ['sent', 'delivered'] }, (error: Error, body: any) => { });
```

andipaetzold (approved):

Issue #36308, opened 2019-06-20
@types/underscore error TS2322 for `_.chain` after upgrade to v1.9


smileLilith:
After upgrade of `@types/underscore` from v1.8.20 to v1.9.0 all usages of `_.chain` in my angular project were broken. For example
```
const arr1: string[] = ['z', 'x', 'y'], query = 'z';
let arr2: string[] = ['a', 'b', 'c'];
...
arr2 = _.chain(arr1)
      .union(arr2)
      .without(query)
      .value()
```
has returned
```
error TS2322: Type 'string' is not assignable to type 'string[]'
```

It seems it was caused by changes done in [this commit](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/9cf5c98aa8c9edc65e70dd078e0f94d72d4aa6d1#diff-bc995fed3367d28165753a2581d1af54).

@regevbr, @DanielRosenwasser was this expected? If yes, what is upgrade path?

regevbr:
I'm sorry for that. I will get right on it.
Thanks for reporting!

smileLilith:
No worries, thank you

regevbr:
The changes made at #36230 caused it
@jgonggrijp the changes we just made caused a breaking chnage and I verified it in the test of the library.
As I mentioned in the PR - I will need to revise all the methods under the _.Chain interface so they will not break existing code.
Any suggestions for now?

jgonggrijp:
@regevbr I have nothing to add to what you already said. Add tests that replicate the problem, then revise the code until all tests pass again.

And then wait until the next issue comes up... it's a fact of life.

regevbr:
@jgonggrijp I already did it and the fix takes a second but I found myself revising the entire library as almost 90% of the _Chain interface is completely wrong (even from before the change). I also noticed that there are some missing methods declarations and dome deprecated ones.
I want to finish the complete revision of the library but it is going to take a few days for us to ping pong on the review as it will be a big one

jgonggrijp:
@regevbr Would it be doable for you to submit a PR that only fixes the current issue first? I don't mind reviewing twice, and while I agree it's important for the Chain interface to be correct, I'd say that newly introduced bugs are a bit more urgent to fix than long-existing ones.

regevbr:
@jgonggrijp sure coming right up

PR #36319, opened 2019-06-20
Fix #36308 - underscore has bad chain typings

regevbr:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

PR #36510, opened 2019-06-28
[@types/underscore] Fix #36308 - underscore has bad chain typings fix only for function 'map' (#36319)

Radullus:
Fix typing in 'underscore' for function 'map'

Please fill in this template.

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: https://underscorejs.org/#map
https://underscorejs.org/#chain

### DIFF
--- a/types/underscore/index.d.ts
+++ b/types/underscore/index.d.ts
@@ -5094,7 +5094,7 @@ declare module _ {
         * Wrapped type `any[]`.
         * @see _.map
         **/
-        map<TArray>(iterator: _.ListIterator<T, TArray[]>, context?: any): _ChainOfArrays<TArray>;
+        map<TArray>(iterator: _.ListIterator<T, TArray>, context?: any): _ChainOfArrays<TArray>;
 
         /**
         * Wrapped type `any[]`.
@@ -5106,7 +5106,7 @@ declare module _ {
         * Wrapped type `any[]`.
         * @see _.map
         **/
-        map<TArray>(iterator: _.ObjectIterator<T, TArray[]>, context?: any): _ChainOfArrays<TArray>;
+        map<TArray>(iterator: _.ObjectIterator<T, TArray>, context?: any): _ChainOfArrays<TArray>;
 
         /**
         * Wrapped type `any[]`.
--- a/types/underscore/index.d.ts
+++ b/types/underscore/index.d.ts
@@ -5100,7 +5100,7 @@ declare module _ {
         * Wrapped type `any[]`.
         * @see _.map
         **/
-        map<TResult>(iterator: _.ListIterator<T, TResult>, context?: any): _Chain<TResult>;
+        map<TResult>(iterator: _.ListIterator<T, TResult>, context?: any):  _Chain<TResult, TResult[]>;
 
         /**
         * Wrapped type `any[]`.
@@ -5112,7 +5112,7 @@ declare module _ {
         * Wrapped type `any[]`.
         * @see _.map
         **/
-        map<TResult>(iterator: _.ObjectIterator<T, TResult>, context?: any): _Chain<TResult>;
+        map<TResult>(iterator: _.ObjectIterator<T, TResult>, context?: any): _Chain<TResult, TResult[]>;
 
         /**
         * @see _.map
--- a/types/underscore/index.d.ts
+++ b/types/underscore/index.d.ts
@@ -5094,7 +5094,7 @@ declare module _ {
         * Wrapped type `any[]`.
         * @see _.map
         **/
-        map<TArray>(iterator: _.ListIterator<T, TArray>, context?: any): _ChainOfArrays<TArray>;
+        map<TArray>(iterator: _.ListIterator<T, TArray[]>, context?: any): _ChainOfArrays<TArray>;
 
         /**
         * Wrapped type `any[]`.
@@ -5106,7 +5106,7 @@ declare module _ {
         * Wrapped type `any[]`.
         * @see _.map
         **/
-        map<TArray>(iterator: _.ObjectIterator<T, TArray>, context?: any): _ChainOfArrays<TArray>;
+        map<TArray>(iterator: _.ObjectIterator<T, TArray[]>, context?: any): _ChainOfArrays<TArray>;
 
         /**
         * Wrapped type `any[]`.
--- a/types/underscore/underscore-tests.ts
+++ b/types/underscore/underscore-tests.ts
@@ -221,13 +221,31 @@ var usersData: _.Dictionary<{ age: number; name: string }> = {
 };
 
 let youngPeopleId: string[] = _.chain(usersData)
-    .filter(p => p.age < 60)
     .map((p, k: string) => k)
     .value();
 
 let usersTable: { age: number; name: string; id: string }[] = _.chain(usersData)
     .map<{ age: number; name: string; id: string }>((p, k: string) => {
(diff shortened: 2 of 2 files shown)


Radullus:
@jgonggrijp @regevbr @renjfk  Can you please review?

jgonggrijp:
I defer this one to @regevbr because he is currently working on a complete fix for the chain typings. I'm not sure whether the fix in the current PR will combine well with his work, but I think the unittest would be a welcome addition in any case.

regevbr (changes requested):
@Radullus Thanks for the PR!

The `T` in `_Chain<T,V>` represents the type of the object and `V` represents the object.  
So for example, `string[]` should be `_Chain<string,string[]>` and `_.Dictionary<number>` should be `_Chain<number, _.Dictionary<number>` 

The lines you changed made your test pass, but they would cause other tests, that should be there but are not :), to fail.

Please see the following suggestion to fix the issue:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d200340ecb521b3856f8bbf6e5b61a33182f9363/types/underscore/index.d.ts#L5103

should be changed to 
```ts
 map<TResult>(iterator: _.ListIterator<T, TResult>, context?: any): _Chain<TResult, TResult[]>; 
```

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d200340ecb521b3856f8bbf6e5b61a33182f9363/types/underscore/index.d.ts#L5115

should be changed to 
```ts
 map<TResult>(iterator: _.ObjectIterator<T, TResult>, context?: any): _Chain<TResult, _.Dictionary<TResult>>;
```

Can you please try it (after reverting your change) and see if the test passes? Also I agree with @jgonggrijp that more tests are always welcome, so can you please also add a test for map operation on an object?

types/underscore/index.d.ts
```diff
@@ -5094,25 +5094,25 @@ declare module _ {
         * Wrapped type `any[]`.
         * @see _.map
         **/
-        map<TArray>(iterator: _.ListIterator<T, TArray[]>, context?: any): _ChainOfArrays<TArray>;
+        map<TArray>(iterator: _.ListIterator<T, TArray>, context?: any): _ChainOfArrays<TArray>;
```
regevbr: this should be reverted as current behavior is ok

types/underscore/index.d.ts
```diff
@@ -5094,25 +5094,25 @@ declare module _ {
...
         **/
-        map<TResult>(iterator: _.ListIterator<T, TResult>, context?: any): _Chain<TResult>;
+        map<TResult>(iterator: _.ListIterator<T, TResult>, context?: any):  _Chain<TResult, TResult[]>;
 
         /**
         * Wrapped type `any[]`.
         * @see _.map
         **/
-        map<TArray>(iterator: _.ObjectIterator<T, TArray[]>, context?: any): _ChainOfArrays<TArray>;
+        map<TArray>(iterator: _.ObjectIterator<T, TArray>, context?: any): _ChainOfArrays<TArray>;
```
regevbr: this should be reverted as current behavior is ok

Radullus:
- I added more test, for mapping 'Dictionary'.
 - I apply charges suggested for Line 5103 in line DefinitelyTyped/types/underscore/index.d.ts d200340
 - I modified Line 5115 DefinitelyTyped/types/underscore/index.d.ts d200340. 
Map function for 'chain' object don't return 'Dictionary' but return 'Array'.  You can see some of my tested https://jsfiddle.net/Radullus/fsmj2rxa/37/

types/underscore/underscore-tests.ts
```diff
@@ -209,6 +209,28 @@ _(stooges)
...
+    .map(p => p.name)
+    .value();
+
+var usersData: _.Dictionary<{ age: number; name: string }> = {
+    'user id': { name: 'moe', age: 40 },
+    'other user Id': { name: 'larry', age: 50 },
+    'fake id': { name: 'curly', age: 60 },
+};
+
+let youngPeopleId: string[] = _.chain(usersData)
```
regevbr: This test should actually fail, as performing filter on a a Dictionary<T> returns T[], thus performing map on it will need the following iterator:
```
(value: T, index: number, list: T[]): TResult
```
The reason this test passes now is because of missing signature of filter for _Chain which should be:
```ts
filter(iterator: _.ObjectIterator<T, boolean>, context?: any): _Chain<T, T[]>;
```
But fixing it will not even help as the entire _Chain interface is malformed :-( and I'm working on a wide fix for it.
I would rather you only add tests that use the fixed map method

regevbr:
Thanks @Radullus!
The _Chain interface is completely messed up currently :(
I added some comments for you to review

Radullus:
Hi @regevbr . I have applied the changes you had suggested.

regevbr:
I added new comment for you to review

types/underscore/underscore-tests.ts
Radullus: You have right. I will stick in this test to 'map' function. I interface definition I was trying to test.
And, I will add some test for 'filter' separate.

types/underscore/index.d.ts
Radullus: done.
I have add test for this definition of 'map', but I have to comment part of the 'testing' witch check output because it will cause the test to fail.

types/underscore/index.d.ts
Radullus: done.
I have add test for this definition of 'map', but I have to comment part of the 'testing' witch check output because it will cause the test to fail.

types/underscore/underscore-tests.ts
```diff
@@ -209,6 +209,46 @@ _(stooges)
...
+    .value();
+
+let usersTable: { age: number; name: string; id: string }[] = _.chain(usersData)
+    .map<{ age: number; name: string; id: string }>((p, k: string) => {
+        return { id: k, ...p };
+    })
+    .value();
+
+// Test map function with _ChainOfArrays<>
+let usersTable_2 /*: { age: number; name: string; id: string }[][]*/ = _.chain(usersData)
```
regevbr: why did you remove the type here?
It is best to use the flatten() method here and then get the value

types/underscore/underscore-tests.ts
```diff
@@ -209,6 +209,46 @@ _(stooges)
...
+    .map<{ age: number; name: string; id: string }>((p, k: string) => {
+        return [{ id: k, ...p }];
+    })
+    .value();
+
+let usersTable_3 /*: { score: number; fullName: string; login: string }[][]*/ = _.chain(usersTable)
```
regevbr: same here - why did you remove the type?

types/underscore/underscore-tests.ts
Radullus: because the type by the underscore should be like this one in the comment, But with current typescript definition, There is missing one level of array nest. 
`{ age: number; name: string; id: string }[][]  // type return by underscore
{ age: number; name: string; id: string }[]    // type with is calculate base of current definitions
`
It would require more changes in the underscore typescript definition. And I'm not certain what should be changed?
`interface _ChainOfArrays<T> extends _Chain<T[]> `
to 
`interface _ChainOfArrays<T> extends _Chain<T[], T[][]> `
it will not match with data return from function 'groupBy'

regevbr: You are correct. Lets leave it be for now

regevbr (approved):

PR #36525, opened 2019-06-28
[react] Support <a ping> attribute

tony-scio:
See spec:
https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -1805,6 +1805,7 @@ declare namespace React {
         href?: string;
         hrefLang?: string;
         media?: string;
+        ping?: string;
         rel?: string;
         target?: string;
         type?: string;


saranshkataria (commented):
Since this is an experimental attribute as of now, I don't think it makes sense to add it right now?

tony-scio:
@saranshkataria what makes you say <a ping> is experimental? It's in the HTML5 spec which I linked to in the description and it's also in MDN which also doesn't mark it experimental:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a

saranshkataria:
Scroll down below in the table on MDN, it has an icon next to it which tells you it's experimental

johnnyreilly:
I think it's probably fine to merge this - it's been around for a long time (3 years === infinity in the web)

https://davidwalsh.name/ping-attribute

It can be used and I don't see a downside to it being added. Have I missed anything?

johnnyreilly:
Actually reading some of the answers here: https://stackoverflow.com/questions/22428764/how-cross-browser-is-the-ping-attribute-of-anchor-elements

Maybe it shouldn't be merged. It sounds like it's not here for good.

tony-scio:
I'll just make one more appeal before going away. That SO answer is 5 years old and the big argument is that WHATWG defines it and W3C does not. Since then, the W3C has basically just decided to follow what the WHATWG defines (see https://www.w3.org/2019/04/WHATWG-W3C-MOU.html) and &lt;a ping> is still here.

Also, at the time IE and Opera didn't implement, but since then, they've both changed to chromium implementations.

Regardless, I wanted this in the context of a chrome extension where it is supported. As an aside, if this is not going to be merged, does anyone happen to know if there's anything cleaner I can do on my end to prevent the error besides this hack:
```jsx
          {/*
           // @ts-ignore */}
```

saranshkataria:
https://caniuse.com/#search=ping
No firefox and IE support yet

PR #36550, opened 2019-06-29
[is-number] Added `num is number` user defined type guard as return type

jedster1111:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards>>

Hopefully this looks okay, give me a shout if I missed anything or anything needs changing.

### DIFF
--- a/types/is-number/index.d.ts
+++ b/types/is-number/index.d.ts
@@ -1,6 +1,7 @@
-// Type definitions for is-number 3.0
+// Type definitions for is-number 7.0
 // Project: https://github.com/jonschlinkert/is-number
 // Definitions by: Harry Shipton <https://github.com/harryshipton>
+//                 Jed Thompson <https://github.com/jedster1111>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 export = is_number;
@@ -10,4 +11,4 @@ export = is_number;
  * @param num Any value that should be tested for being a number
  * @returns true if the parameter is a valid number, otherwise false
  */
-declare function is_number(num: any): boolean;
+declare function is_number(num: any): num is number;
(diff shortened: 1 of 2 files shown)


svens-uk (commented):
The JSDoc return signature doesn't reflect the added type guard — I'm not an expert at it but I'm reasonably sure that JSDoc supports type guard definitions. Is this something we should add?

harryshipton:
Thanks for adding a type guard though, it seems like a much more sensible way of typing the module! :)

jedster1111:
No problem 😄
I added the type guard to the JSDoc like this:
```js
/**
 * Will test to see if the argument is a valid number, excluding Infinity and NaN.
 * @param num Any value that should be tested for being a number
 * @returns {num is number} true if the parameter is a valid number, otherwise false
 */
declare function is_number(num: any): num is number;
```
but I got an error when I ran the lint tests:
```
ERROR: 12:13  no-redundant-jsdoc-2  Type annotation in JSDoc is redundant in TypeScript code.
```
which I guess makes sense.

jedster1111:
I think I just realised this change isn't quite right as this library checks both string and number inputs. Maybe you could type guard to a string or number though, not sure if that would be helpful though.

harryshipton:
I'll take a closer look this evening 🙂

jedster1111:
Okay, I added string to the type guard, and think it works quite nicely from what I can see.
  - Narrows `any` and `unknown` to `string | number`.
  - Keeps `string` as `string`
  - Keeps `number` as `number`

Wasn't 100% sure how to write a meaningful test for it though 😄

harryshipton:
Okay, I agree that the JSDoc error does make sense and we don't need to add typings to it.

Testing the type when isNumber() returns true should be fine — the type when isNumber() returns false is undetermined so I don't think we need to test it. I'm happy with the changes unless someone else raises any issues. 👍

svens-uk (approved):

PR #36562, opened 2019-07-01
feat: node v12.5

SimonSchick:
Please fill in this template.

- [ ] Add or edit tests to reflect the change. (Run with `npm test`.) Unrelated test is failing...

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/node/releases

Async iterables are tricky, haven't added those.

### DIFF
--- a/types/node/index.d.ts
+++ b/types/node/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for non-npm package Node.js 12.5
+// Type definitions for non-npm package Node.js 12.6
 // Project: http://nodejs.org/
 // Definitions by: Microsoft TypeScript <https://github.com/Microsoft>
 //                 DefinitelyTyped <https://github.com/DefinitelyTyped>
--- a/types/node/index.d.ts
+++ b/types/node/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for non-npm package Node.js 12.0
+// Type definitions for non-npm package Node.js 12.5
 // Project: http://nodejs.org/
 // Definitions by: Microsoft TypeScript <https://github.com/Microsoft>
 //                 DefinitelyTyped <https://github.com/DefinitelyTyped>
@@ -68,6 +68,7 @@ interface Map<K, V> {}
 interface ReadonlySet<T> {}
 interface IteratorResult<T> { }
 interface Iterable<T> { }
+interface AsyncIterable<T> { }
 interface Iterator<T> {
     next(value?: any): IteratorResult<T>;
 }
--- a/types/node/stream.d.ts
+++ b/types/node/stream.d.ts
@@ -18,6 +18,11 @@ declare module "stream" {
         }
 
         class Readable extends Stream implements NodeJS.ReadableStream {
+            /**
+             * A utility method for creating Readable Streams out of iterators.
+             */
+            static from(iterable: Iterable<any> | AsyncIterable<any>, options?: ReadableOptions): Readable;
+
             readable: boolean;
             readonly readableHighWaterMark: number;
             readonly readableLength: number;
@@ -29,7 +34,7 @@ declare module "stream" {
             resume(): this;
             isPaused(): boolean;
             unpipe(destination?: NodeJS.WritableStream): this;
-            unshift(chunk: any): void;
+            unshift(chunk: any, encoding?: BufferEncoding): void;
             wrap(oldStream: NodeJS.ReadableStream): this;
             push(chunk: any, encoding?: string): boolean;
             _destroy(error: Error | null, callback: (error?: Error | null) => void): void;
@@ -110,7 +115,8 @@ declare module "stream" {
         }
 
         class Writable extends Stream implements NodeJS.WritableStream {
-            writable: boolean;
+            readonly writable: boolean;
+            readonly writableFinished: boolean;
             readonly writableHighWaterMark: number;
             readonly writableLength: number;
             constructor(opts?: WritableOptions);
@@ -208,7 +214,8 @@ declare module "stream" {
 
         // Note: Duplex extends both Readable and Writable.
         class Duplex extends Readable implements Writable {
-            writable: boolean;
+            readonly writable: boolean;
+            readonly writableFinished: boolean;
             readonly writableHighWaterMark: number;
             readonly writableLength: number;
             constructor(opts?: DuplexOptions);
--- a/types/node/tls.d.ts
+++ b/types/node/tls.d.ts
@@ -64,70 +64,72 @@ declare module "tls" {
         version: string;
     }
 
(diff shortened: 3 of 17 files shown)


SimonSchick:
Ugh, seems like there are some packages that depent on the statics of the `Readable` class...

Flarna:
see #36106

types/node/stream.d.ts
```diff
@@ -18,6 +18,11 @@ declare module "stream" {
         }
 
         class Readable extends Stream implements NodeJS.ReadableStream {
+            /**
+             * A utility method for creating Readable Streams out of iterators.
+             */
+            static from(iterable: Iterable<any>, options?: ReadableOptions): Readable;
```
Flarna: Should the type of iterable be `Iterable<any> | AsyncIterable<any>`?

types/node/tls.d.ts
```diff
@@ -64,70 +64,72 @@ declare module "tls" {
         version: string;
     }
 
+    export interface TLSSocketOptions {
```
Flarna: I think `enableTrace?: boolean` should be added here also.

types/node/tls.d.ts
```diff
@@ -292,41 +312,47 @@ declare module "tls" {
         addListener(event: "OCSPRequest", listener: (certificate: Buffer, issuer: Buffer, callback: (err: Error | null, resp: Buffer) => void) => void): this;
         addListener(event: "resumeSession", listener: (sessionId: Buffer, callback: (err: Error, sessionData: Buffer) => void) => void): this;
         addListener(event: "secureConnection", listener: (tlsSocket: TLSSocket) => void): this;
+        addListener(event: "keylog", listener: (line: Buffer, tlsSocket: TLSSocket) => void): this;
```
Flarna: A `keylog` event was also added to `TLSSocket` but with only `line: Buffer` as argument.

types/node/worker_threads.d.ts
```diff
@@ -73,7 +73,12 @@ declare module "worker_threads" {
         postMessage(value: any, transferList?: Array<ArrayBuffer | MessagePort>): void;
         ref(): void;
         unref(): void;
-        terminate(callback?: (err: Error, exitCode: number) => void): void;
+        /**
+         * Stop all JavaScript execution in the worker thread as soon as possible.
+         * Returns a Promise for the exit code that is fulfilled when the `exit` event is emitted.
+         * @param callback Deprecated, use promise instead.
+         */
+        terminate(callback?: (err: Error, exitCode: number) => void): Promise<number>;
```
Flarna: `terminate()` no longer accepts a callback now

Flarna (commented):
I think `tls.rootCertificates: ReadonlyArray<string>` was also added in 12.3.0.

types/node/stream.d.ts
SimonSchick: `AsyncIterable` isn't supported by the mimum required node version, I tried adding it as an overload in the ts3.2 folder but that didn't work.

types/node/worker_threads.d.ts
SimonSchick: It's deprecated, not entirely removed but useless regardless.
We can remove it to encourage people to migrate however.

types/node/tls.d.ts
SimonSchick: Not according to the changelog, this is just cleanup.

SimonSchick:
I will probably amend changes for 12.6 while I'm at it.

types/node/stream.d.ts
SimonSchick: Nvm, adding shallow definition.

types/node/tls.d.ts
Flarna: but according to the docs: https://nodejs.org/dist/latest-v12.x/docs/api/tls.html#tls_new_tls_tlssocket_socket_options and https://nodejs.org/dist/latest-v12.x/docs/api/tls.html#tls_tls_connect_options_callback

SimonSchick: *sigh* I wish the changelog was more detailed.

types/node/tls.d.ts
```diff
@@ -212,65 +176,119 @@ declare module "tls" {
...
         once(event: "secureConnect", listener: () => void): this;
         once(event: "session", listener: (session: Buffer) => void): this;
+        once(event: "keylog", listener: (line: Buffer) => void): this;
 
         prependListener(event: string, listener: (...args: any[]) => void): this;
         prependListener(event: "OCSPResponse", listener: (response: Buffer) => void): this;
         prependListener(event: "secureConnect", listener: () => void): this;
         prependListener(event: "session", listener: (session: Buffer) => void): this;
+        prependListener(event: "keylog", listener: (line: Buffer) => void): this;
+        prependListener(event: "keylog", listener: (line: Buffer) => void): this;
```
TeamworkGuy2: why are lines 217 and 218 the same?

TeamworkGuy2: why are lines 217 and 218 the same?

SimonSchick: Copy & Paste is magic.

TeamworkGuy2 (approved):

types/node/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for non-npm package Node.js 12.0
+// Type definitions for non-npm package Node.js 12.5
```
Flarna: should be 12.6 now

types/node/tls.d.ts
```diff
@@ -212,65 +176,118 @@ declare module "tls" {
...
+         * used to debug TLS connection problems.
+         * @default false
+         */
+        enableTrace?: boolean;
+        /**
+         * If true the server will request a certificate from clients that
+         * connect and attempt to verify that certificate. Defaults to
+         * false.
+         */
         requestCert?: boolean;
```
Flarna: I think `requestCert` is only applicable to `TLSSocket` and `createServer` but not to `connect` therefore it should be not part of `CommonConnectionOptions`.
There are also other options here where I'm not sure if they are correct like `NPNProtocols` which is not in docs but maybe it's best to update this via a separate PR.

SimonSchick: I went through the code and made sure `requestCert` is passed through.
Also NPNProtocols was kinda weird, there are some referenced to it in https://github.com/nodejs/node/search?q=NPNProtocols&unscoped_q=NPNProtocols but it appears to have been removed here: https://github.com/nodejs/node/pull/19403 so I will probably remove it.

The reason I extracted those options is due to wanting to avoid even more duplication.

Flarna (approved):

BrunoScheufler (approved):

PR #36564, opened 2019-07-01
Support for UMD module definitions

gtrifidis:
Adding support for UMD module definitions, so types can be used in a project with no module loading.
[UMD](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#support-for-umd-module-definitions)

In a TS project where module code generation is set to 'none', types cannot be resolved without using an "import".

If "import" is used compiler complains **"Cannot use imports, exports or module augmentations when '--module' is 'none"**

Credit to [Aluan Haddad](https://stackoverflow.com/users/1915893/aluan-haddad) and [Mu-Tsun Tsai](https://stackoverflow.com/users/9953396/mu-tsun-tsai)

[Original stackoverflow issue](https://stackoverflow.com/questions/56774764/reference-declaration-file-d-ts-in-a-ts-file-with-compiler-module-option/)

### DIFF
--- a/types/i18next/index.d.ts
+++ b/types/i18next/index.d.ts
@@ -685,3 +685,4 @@ declare namespace i18next {
 
 declare const i18next: i18next.i18n;
 export = i18next;
+export as namespace i18next;


mxl (approved):

lenovouser (changes requested):
You should re-open this at the original i18next repository, they are shipping their own definitions now.

I’ll open a PR to delete these types shortly.

gtrifidis:
So I should just close this one, then?

lenovouser:
@gtrifidis yes, open a new one with the same changes at [i18next/i18next](https://github.com/i18next/i18next)

PR #36581, opened 2019-07-01
R.zip compatible with a tuple array

rwx222:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [url here](https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple)

### The problem

Currently this is not posible:
```typescript
type TNameAge = [string, number]; // this is a tuple (not an array) to typescript

const namesAndAges: TNameAge[] = R.zip<string, number>(["John", "Juliet", "Melanie"], [21, 22, 23]);
const namesAndAges2: TNameAge[] = R.zip(["John", "Juliet", "Melanie"])([21, 22, 23]);
```

**TNameAge** is a [typescript tuple](https://www.typescriptlang.org/docs/handbook/basic-types.html#tuple), so you can't assign it the result of **R.zip** to a typed tuple, because they aren't compatible. But they should be, because once compiled to javascript there's no difference.

I added some tests to make sure is compatible with both of them, a tuple array (`Array<TNameAge>`), and a common array (`Array<TNames>`), to make sure I'm not breaking the current behavior:
```typescript
type TNameAge = [string, number]; // this is a tuple (not an array) to typescript
const namesAndAges: TNameAge[] = R.zip<string, number>(["John", "Juliet", "Melanie"], [21, 22, 23]);
const namesAndAges2: TNameAge[] = R.zip(["John", "Juliet", "Melanie"])([21, 22, 23]);

type TNames = string[]; // this is an array
const fullNames: TNames[] = R.zip<string, string>(["John", "Juliet", "Melanie"], ["Titor", "Burke", "Cross"]);
const fullNames2: TNames[] = R.zip(["John", "Juliet", "Melanie"])(["Titor", "Burke", "Cross"]);
```

### DIFF
--- a/types/ramda/index.d.ts
+++ b/types/ramda/index.d.ts
@@ -556,13 +556,10 @@ declare namespace R {
 
     type Path = ReadonlyArray<(number | string)>;
 
+    type KeyValueTuple<K, V> = [K, V];
+
     interface Functor<T> {
         map<U>(fn: (t: T) => U): Functor<U>;
-    }
-
-    interface KeyValuePair<K, V> extends Array<K | V> {
-        0: K;
-        1: V;
     }
 
     interface ArrayLike {
@@ -1428,8 +1425,8 @@ declare namespace R {
         /**
          * Creates a new object out of a list key-value pairs.
          */
-        fromPairs<V>(pairs: Array<KeyValuePair<string, V>>): { [index: string]: V };
-        fromPairs<V>(pairs: Array<KeyValuePair<number, V>>): { [index: number]: V };
+        fromPairs<V>(pairs: Array<KeyValueTuple<string, V>>): { [index: string]: V };
+        fromPairs<V>(pairs: Array<KeyValueTuple<number, V>>): { [index: number]: V };
 
         /**
          * Splits a list into sublists stored in an object, based on the result of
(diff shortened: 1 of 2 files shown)


googol:
Hey! This definitely is a problem. I think it would make more sense to fix the definition of KeyValuePair, which most likely has been defined before the new Tuple goodness was implemented. It is meant to specify the semantics of `[K, V]` in pre-tuple type syntax

drebits:
Hi @googol, please check the [changes](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36581/files) again.

You mean something like that, I think It's what you said, because **R.zip** *always will return a tuple array*.
So I reimplemented **KeyValuePair** as **KeyValueTuple** and the tests and everything looks ok.

Do you agree?

By the other hand, I'm not sure when tuples where implemented to typescript, but at least in [version 3.1](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#mapped-types-on-tuples-and-arrays) already existed, and as [this type definition says Typescript 3.3](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e4a0d4f532b177fc800e8ade7f1b39e9879d4b3c/types/ramda/index.d.ts#L34) I think is ok.

googol (commented):
I'd retain the name `KeyValuePair`, since that is used by user code. This way we won't create a breaking change. Otherwise this looks good

drebits:
You're right @googol. I already fixed it.

googol (approved):
Looks good to me

PR #36591, opened 2019-07-02
fix(types/next): Add req and res NextContext typedefs used in `next export`

jtomaszewski:
On `next export` command, `req` and `res` objects are hardcoded simple objects: https://github.com/zeit/next.js/blob/master/packages/next/export/worker.js#L42-L43 and thus they have different type definition (i.e. `res.writeHead` is missing which might cause [such code](https://github.com/zeit/next.js/wiki/Redirecting-in-%60getInitialProps%60) to blow up with a `TypeError: res.writeHead is not a function` ).

- [x] Provide a URL to documentation or source code which provides context for the suggested changes

### DIFF
--- a/types/next/index.d.ts
+++ b/types/next/index.d.ts
@@ -49,9 +49,9 @@ declare namespace next {
         /** String of the actual path (including the query) shows in the browser */
         asPath: string;
         /** HTTP request object (server only) */
-        req?: http.IncomingMessage & CustomReq;
+        req?: http.IncomingMessage & CustomReq | { url: string };
         /** HTTP response object (server only) */
-        res?: http.ServerResponse;
+        res?: http.ServerResponse | {};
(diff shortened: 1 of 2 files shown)


jtomaszewski:
Related: https://github.com/zeit/next.js/pull/7729

scottdj92 (approved):

scottdj92:
Looks good, but the CI build failed. Looks like we'll need to update the tests. Can you go ahead and do that?

jtomaszewski:
@scottdj92 I updated it a littlebit so the type is actually a `Partial<Request>` instead of `Request | {}`. This will avoid TS raising errors on `req && req.getHeaderNames ? req.getHeaderNames() : []` and avoid the need to do sth like `req && (req as ...).getHeaderNames ? (req as ...).getHeaderNames() : []` .

Tests should be fixed now as well.

PR #36604, opened 2019-07-02
[@types/enzyme] Add missing Wrapper debug options parameter

erickzhao:

* `ShallowWrapper`: https://airbnb.io/enzyme/docs/api/ShallowWrapper/debug.html
* `ReactWrapper`: https://airbnb.io/enzyme/docs/api/ReactWrapper/debug.html
**N.B. although the `ReactWrapper` docs are missing the `verbose` option, it's implemented in the code. I've opened a PR here to fix the docs: https://github.com/airbnb/enzyme/pull/2184. See PR for more context.**
- [x] ~Increase the version number in the header if appropriate.~ **(N/A)**
- [x] ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.~ **(N/A)**

### DIFF
--- a/types/enzyme/index.d.ts
+++ b/types/enzyme/index.d.ts
@@ -102,9 +102,15 @@ export interface CommonWrapper<P = {}, S = {}, C = Component<P, S>> {
      * @param ...args The argments to the invokePropName function
      * @returns The value of the function.
      */
-    invoke<K extends NonNullable<{
-        [K in keyof P]: P[K] extends ((...arg: any[]) => void) | undefined ? K : never
-    }[keyof P]>>(invokePropName: K): P[K];
+    invoke<
+        K extends NonNullable<
+            {
+                [K in keyof P]: P[K] extends ((...arg: any[]) => void) | undefined ? K : never;
+            }[keyof P]
+        >
+    >(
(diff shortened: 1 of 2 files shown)


types/enzyme/index.d.ts
```diff
@@ -300,7 +307,10 @@ export interface CommonWrapper<P = {}, S = {}, C = Component<P, S>> {
      * Returns an html-like string of the wrapper for debugging purposes. Useful to print out to the console when
      * tests are not passing when you expect them to.
      */
-    debug(): string;
+    debug(options?: {
+        ignoreProps?: boolean; // Whether props should be omitted in the resulting string. Props are included by default.
```
jwbay: These would be useful as JSDoc comments so they show up in intellisense. Something like this, I think:
```suggestion
        /** Whether props should be omitted in the resulting string. Props are included by default. */
        ignoreProps?: boolean; 
```

erickzhao: @jwbay great suggestion! didn't know that intellisense picked definitions up that way. thanks :)

jwbay (approved):

PR #32311, opened 2019-01-19
[@types/ts-nameof] Add package

dsherret:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [ ] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes). -- Need to use the unnecessary generics because it's a compile time transformation
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If adding a new definition:
- [ ] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration` -- It does have its own package, but it would be better if it were a types package so the global functions can be more easily resolved. Maybe there's a better way, but I'm not so sure.
- [x] If this is for an NPM package, match the name. If not, do not conflict with the name of an NPM package.
- [x] Create it with `dts-gen --dt`, not by basing it on an existing project.
- [x] `tslint.json` should be present, and `tsconfig.json` should have `noImplicitAny`, `noImplicitThis`, `strictNullChecks`, and `strictFunctionTypes` set to `true`. -- Note: dts-gen did not add `strictFunctionTypes`

PR #36608, opened 2019-07-03
[@types/ts-nameof] Update to 3.2

dsherret:
**MAINTAINERS READ ME:** This package does distribute its own types, but having it as a `@types` package allows these global functions to be resolved without any effort other than the install. Maybe there is a better way, but I've asked about this in the past (see this [previous PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32311)).

- [ ] Add or edit tests to reflect the change. (Run with `npm test`.) -- Fails saying `Error: In package.json: Don't use a package.json for @types dependencies.` I think unrelated since there is no package.json here? Edit: I probably just forgot to run `npm install`. It's been a while since I ran that on my local branch.
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes). -- Unnecessary generics are necessary because this is a compiler transform.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/dsherret/ts-nameof/commit/b4f8f1152170d65f9f128da028626ede731478af

### DIFF
--- a/types/ts-nameof/index.d.ts
+++ b/types/ts-nameof/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for ts-nameof 2.0
+// Type definitions for ts-nameof 3.2
 // Project: https://github.com/dsherret/ts-nameof#readme
 // Definitions by: David Sherret <https://github.com/dsherret>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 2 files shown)


dsherret:
Seems @typescript-bot is editing its comment with the same comment every hour?

image

PR #36612, opened 2019-07-03
extending Profile interface and changing VerifyCallback

ngtan:
Please fill in this template.

### DIFF
--- a/types/passport-google-oauth20/index.d.ts
+++ b/types/passport-google-oauth20/index.d.ts
@@ -42,7 +42,7 @@ export interface Profile extends passport.Profile {
     _json: any;
 }
 
-export type VerifyCallback = (err?: any, user?: any, info?: any) => void;
+export type VerifyCallback = (err?: string | Error, user?: any, info?: any) => void;
 
 export class Strategy extends oauth2.Strategy {
     constructor(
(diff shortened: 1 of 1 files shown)


types/passport-google-oauth20/index.d.ts
```diff
@@ -36,16 +37,21 @@ export interface StrategyOptionsWithRequest extends _StrategyOptionsBase {
 
 export interface Profile extends passport.Profile {
     profileUrl: string;
+
+    _raw: string;
+    _json: any;
 }
 
+export type VerifyCallback = (err?: any, user?: any, info?: any) => void;
```
armanio123: Why are you changing from using `passport-oauth2` type to implement it's own?

ngtan: Hello @armanio123 
Thanks for your response on my pull request. 
Actually, for the first parameter when using `VerifyCallback` from `passport-oauth2`, it will show the error (include stack trace) on user view instead of a friendly error (just an error message) from developer. 
So, that's a reason why I implemented another one. Please give your advices if you have other ideas. 
Thanks.

armanio123: You mean that `VerifyCallback` is different in a sense that the ` err` parameter contains different information but the type would be the same?

Also,  if `err` is always `string`, could you use that type instead?

ngtan: Thanks @armanio123 , that's make sense. Actually, I think I will add 1 more type for it. To keep `VerifyCallback` like from `passport-oauth2`.

armanio123 (approved):

PR #36626, opened 2019-07-03
[recharts] Added missing baseValue prop

veddermatic:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: http://recharts.org/en-US/api/AreaChart#baseValue

### DIFF
--- a/types/recharts/index.d.ts
+++ b/types/recharts/index.d.ts
@@ -12,6 +12,7 @@
 //                 Robert Stigsson <https://github.com/RobertStigsson>
 //                 Kosaku Kurino <https://github.com/kousaku-maron>
 //                 Leon Ng <https://github.com/iflp>
+//                 Dave Vedder <https://github.com/veddermatic>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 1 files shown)


types/recharts/index.d.ts
```diff
@@ -96,6 +98,7 @@ export interface CategoricalChartWrapper<L = LayoutType> {
     barCategoryGap?: number | string;
     barGap?: number | string;
     barSize?: number | string;
+    baseValue?: BaseValueType;
```
armanio123: I'm not very familiar the library but based on the documentation, seems like `baseValue` is only specific of `AreaChart`. Adding it to the interface will add it to another bunch of classes where the property is not documented. Is that the intended behavior?

armanio123 (changes requested):

types/recharts/index.d.ts
veddermatic: Because you can put  `Area` inside of composed charts, the prop needs to be there as well so it can be passed down to children of `ComposedChart`.   For a real-world example, if you go to their Composed Char example: http://recharts.org/en-US/examples/LineBarAreaComposedChart you can add `baseValue={450}` to the  `ComposedChart` component and it will have the desired effect. So I'm just documenting existing behavior with this change.

armanio123 (approved):

types/recharts/index.d.ts
armanio123: Thanks for clarifying.

PR #36633, opened 2019-07-03
[@types-baseui] Style override signature

alexzywiak:
Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://baseweb.design/theming/understanding-overrides/

If removing a declaration:

### DIFF
--- a/types/baseui/baseui-tests.tsx
+++ b/types/baseui/baseui-tests.tsx
@@ -111,6 +111,17 @@ const NewStyledProgressSteps = styled<ProgressStepsStyleProps, typeof StyledProg
 // Avatar
 <Avatar name='user' size='10'/>;
 <Avatar name='user' size={10}/>; // $ExpectError
+<Avatar
+    name="user"
+    overrides={{ Root: { style: ({ $theme }) => ({ height: $theme.sizing && $theme.sizing.scale1000 }) } }}
+/>;
+<Avatar
+    name="user"
+    overrides={{ Root: { style: ({ $didImageFailToLoad }) => ({ height: $didImageFailToLoad ? '0' : '10px' }) } }}
+/>;
+<Avatar name="user" overrides={{ Root: { style: ({ children }) => ({ height: children ? '10px' : '0'})} }} />;
+<Avatar name="user" overrides={{ Root: { style: ({ $theme }) => ({ height: $theme.sizing && $theme.sizing.notasize })} }} />; // $ExpectError
+<Avatar name="user" overrides={{ Root: { style: ({ $nonExistantProp }) => ({}) }}} />; // $ExpectError
 
 // Block
 <Block as='h1'/>;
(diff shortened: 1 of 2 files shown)


types/baseui/baseui-tests.tsx
```diff
@@ -111,6 +111,8 @@ const NewStyledProgressSteps = styled<ProgressStepsStyleProps, typeof StyledProg
 // Avatar
 <Avatar name='user' size='10'/>;
 <Avatar name='user' size={10}/>; // $ExpectError
+<Avatar name="user" overrides={{ Root: { style: ({$theme}) => ({height: $theme.sizing && $theme.sizing.notasize})} }} />; // $ExpectError
+<Avatar name="user" overrides={{ Root: { style: ({$nonExistantProp}) => ({}) }}} />; // $ExpectError
```
jh3y: Maybe an extra test here that verifies correct params are allowed too.
`<Avatar name="user" overrides={{ Root: { style: ({$size, role, children}) => ({}) }}} />;`

alexzywiak: @jh3y I added happy path tests for the the $theme and the passed in shared props. 

I have a question about `role` and `children` as these aren't included in the StylePropsT type and so not available to the style function in this implementation.  Should those props be added to the generic [here](https://github.com/uber-web/baseui/blob/cc272a2d999f6d39cccde26087113955101bd32e/src/avatar/types.js#L29) or should React intrinsic props, children for example, be defined in the style function type along with $theme?

jh3y: Hey @alexzywiak 👋 

Yeah, this is kinda what I was getting at. Because these are the props that would be available in that function and they might be used. For example, styling might change dependant on the number of children.

I think maybe the latter would make more sense.

alexzywiak: @jh3y, made a change to include React children in props.  Two issues I'm not sure how to approach:

1.  Additional props that are put directly onto the root component.  `role` and `aria-label` in Avatar, `onClick` in Button.  For Avatar, `role` would be available for Root overrides, but not Initials.
1. The `style` and the `component` override have access to the value of the `props` override key.

These are beyond my typescript prowess to figure out at the moment.  I think this style function interface is an improvement over `props: any`, but it's still not accurate.  Let me know what you want to do with it.

alexzywiak: Number 1. is easy once I thought about it.  Just change the Root override generic in Avatar specifically.

armanio123:
@jh3y Is it approved or there are more changes?

jh3y (approved):

types/baseui/baseui-tests.tsx
jh3y: I have approved the PR @alexzywiak 👍 
If you're happy with it and it won't break other definitions whilst making this one more accurate.

tajo:
We are [moving](https://github.com/uber-web/baseui/pull/1538) types into our repo since we plan to support and test TS definitions better and faster. I'm happy to port these changes over once merged.

armanio123:
@tajo thanks for letting us know. I appreciate if you could send a PR for removing the types from DefinitelyTyped once they are on the original library. Meanwhile I'll merge this pull request.

PR #36639, opened 2019-07-04
Add typings for cometd's BinaryExtension

dirk-olmes:
Add typings for the BinaryExtension that's part of the cometd distribution. Heavily inspired by the types definition of the AckExtension :-)


hagl (approved):
Looks good! 
Thanks for adding more typing to cometd!

PR #36668, opened 2019-07-05
[aws-lambda] Add missing fields on CognitoUserPoolTriggerEvent

dalen:
See:
https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-token-generation.html#cognito-user-pools-lambda-trigger-syntax-pre-token-generation

The linkParameter doesn't seem to be in the documentation, but it is there on the events.

Please fill in this template.

### DIFF
--- a/types/aws-lambda/index.d.ts
+++ b/types/aws-lambda/index.d.ts
@@ -293,6 +293,7 @@ export interface CognitoUserPoolTriggerEvent {
         userAttributes: { [key: string]: string };
         validationData?: { [key: string]: string };
         codeParameter?: string;
+        linkParameter?: string;
         usernameParameter?: string;
         newDeviceUsed?: boolean;
         session?: Array<{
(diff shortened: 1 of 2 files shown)


simonbuchan (commented):
Probably should also update the tests?
Also seems a bit lame these are all optional - can any of these be not present? The contents of `claimsOverrideDetails` in particular.

types/aws-lambda/index.d.ts
```diff
@@ -322,6 +323,15 @@ export interface CognitoUserPoolTriggerEvent {
         messageAction?: "SUPPRESS";
         desiredDeliveryMediums?: Array<"EMAIL" | "SMS">;
         forceAliasCreation?: boolean;
+        claimsOverrideDetails?: {
+            claimsToAddOrOverride?: { [key: string]: string };
+            claimsToSuppress?: string[];
+            groupOverrideDetails?: {
```
simonbuchan: Docs has "If you provide an empty or null object in the response, then the groups are suppressed.", so maybe allowing `null` for this?

dalen:
> Probably should also update the tests?

sure, will do.

> Also seems a bit lame these are all optional - can any of these be not present? The contents of `claimsOverrideDetails` in particular.

In their examples they only set some of them, so I think they are all optional.
But we should likely split this whole event by `event.triggerSource` sometime as both valid request and response fields vary by trigger source.

types/aws-lambda/index.d.ts
dalen: Good point, fixed.

PR #36671, opened 2019-07-05
[fluent] Add 'undefined' to return type of FluentBundle.getMessage()

Seally:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: http://projectfluent.org/fluent.js/fluent/class/src/bundle.js~FluentBundle.html

My tests indicate that when `id` is not found in the `FluentBundle` when calling `getMessage(id)`, it will return `undefined`. This makes it a little awkward since the type definition suggests otherwise. This will be a breaking change for existing code if it doesn't test for this already though (should I do something about the version number??).

I've also tweaked `FluentBundle.hasMessage()`'s signature since the [API docs](http://projectfluent.org/fluent.js/fluent/class/src/bundle.js~FluentBundle.html) call the parameter `id`, not `source`. My current understanding of TypeScript is that this shouldn't break anything, but not 100% sure on that.

### DIFF
--- a/types/fluent/fluent-tests.ts
+++ b/types/fluent/fluent-tests.ts
@@ -13,7 +13,7 @@ if (errors1.length) {
 
 const welcome = bundle1.getMessage('welcome');
 
-bundle1.format(welcome, { name: 'Anna' });
+bundle1.format(welcome || [], { name: 'Anna' });
 
 // FluentBundle constructor examples:
 const bundle2 = new FluentBundle(['en-US']);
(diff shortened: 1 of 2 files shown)


JamesNimlos (approved):
Just validated this can return `undefined` for us as well; we are using `fluent@^0.12.0`.

JamesNimlos:
Hey good fix @Seally! Thanks for contributing 👍

PR #36702, opened 2019-07-07
Yup: Support concat() method with better typings

### LINKED ISSUE #36555: [@types/yup] concat() method in Schema<T> has wrong parameters
- [x] I tried using the `@types/yup` package and had problems.
- [x] I tried using the latest stable version of tsc (3.5.2). https://www.npmjs.com/package/typescript
  - Authors: @dhardtke @vtserman @MoretonBayRC @sseppola @YashdalfTheGray @vincentjames501 @robertbullen @sat0yu @dancrumb @deskoh @mauricedb 

Right now, the concat method looks like this:
```
export interface Schema<T> {
    ...
    concat(schema: this): this;
    ...
}
```
That means, if you have an ObjectSchema with many different values, the T parameter will be all of those values with their types, making it impossible to concatenate two ObjectSchemas (which is possible according to yup documentation, mixed.concat()) with different keys.

Maybe the solution could be something like:
```
concat<E>(schema: this | ObjectSchema<E>): this | ObjectSchema<T & E>
```

Thank you for your contribution and help.

mauricedb:
Fixes #36555

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: See issue and below
- [x] ~~Increase the version number in the header if appropriate.~~
- [x] ~~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.~~

Fixes the following which is valid for Yup:
```TypeScript
const schemaA = yup.object({ a: yup.string() });
const schemaB = yup.object({ b: yup.date() });
const mixedSchema = schemaA.concat(schemaB);
type MixedType = yup.InferType<typeof mixedSchema>;
```
This resolves `MixedType` to `{ a: string; b: Date; }`

In the case of conflicting property types it resolves them to `never`
```
const schemaA = yup.object({ a: yup.string() });
const schemaB = yup.object({ a: yup.bool() });
const mixedSchema = schemaA.concat(schemaB);
type MixedType = yup.InferType<typeof mixedSchema>;
```
This resolves `MixedType` to `{ a: never; }` which makes sense as the `schemaA.concat(schemaB);` statement would result in a Yup _TypeError: You cannot `concat()` schema's of different types: string and boolean_ at run-time. Note: With some type combinations TypeScript resolves to the likes of `string & Date` with is less clear but also means you can't assign anything the the property unless it is if type `any`. Not perfect and it would still produce the same run-time error.

Also fixed the `concat()` on the `MixedSchema`
```TypeScript
const mixedStringOrDateSchema = yup.mixed<string>().concat(yup.date())
```
This will now resolve to `yup.MixedSchema<string | Date>`

Note: Doesn't fix the following case which is valid for Yup:
```TypeScript
const schemaA = yup.object({ a: yup.mixed<string>() });
const schemaB = yup.object({ a: yup.date() });
const mixedSchema = schemaA.concat(schemaB);
type MixedType = yup.InferType<typeof mixedSchema>;
```

Where `MixedType` is inferred as `{ a: string & Date; }` where it should be: `{ a: string | Date; }` as both a `string` and `Date` are valid according to Yup

### DIFF
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -12,7 +12,7 @@
 //                 Desmond Koh <https://github.com/deskoh>
 //                 Maurice de Beijer <https://github.com/mauricedb>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 2.9
 
 export function reach<T>(
     schema: Schema<T>,
@@ -296,6 +296,8 @@ export interface ObjectSchema<T extends object | null | undefined = object>
     nullable(isNullable?: boolean): ObjectSchema<T>;
     required(message?: TestOptionsMessage): ObjectSchema<Exclude<T, undefined>>;
     notRequired(): ObjectSchema<T | undefined>;
+    concat(schema: this): this;
+    concat<U extends object>(schema: ObjectSchema<U>): ObjectSchema<T & U>;
 }
 
 export type TransformFunction<T> = (
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -12,7 +12,7 @@
 //                 Desmond Koh <https://github.com/deskoh>
 //                 Maurice de Beijer <https://github.com/mauricedb>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.9
+// TypeScript Version: 2.8
 
 export function reach<T>(
     schema: Schema<T>,
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -104,6 +104,8 @@ export interface MixedSchema<T = any> extends Schema<T> {
     nullable(isNullable?: boolean): MixedSchema<T>;
     required(message?: TestOptionsMessage): MixedSchema<Exclude<T, undefined>>;
     notRequired(): MixedSchema<T | undefined>;
+    concat(schema: this): this;
+    concat<U >(schema: MixedSchema<U>): MixedSchema<T | U>;
 }
 
 export interface StringSchemaConstructor {
--- a/types/yup/yup-tests.ts
+++ b/types/yup/yup-tests.ts
@@ -199,6 +199,10 @@ mixed.test({
     test: testContext,
 });
 
+// mixed with concat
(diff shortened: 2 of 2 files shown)


mauricedb:
@RyanCavanaugh Not sure about Travis not running. Seems to have run just fine here: https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/555240614

emiliskiskis:
@mauricedb This looks a lot better than the previous solution. Even if it is possible to get the run-time TypeError in an edge case, at least this allows for correct usage, unlike the previous solution.

YashdalfTheGray:
Sorry y'all. I'm on vacation so I didn't reply to the conversation. I think I'm okay with this for now until we get a better solution. So I'm going to approve and let this merge. I haven't caught up with the conversation, do we need to merge @emiliskiskis's PR first and then this?

YashdalfTheGray (approved):

mauricedb:
@YashdalfTheGray No, the original PR by @emiliskiskis is closed, my PR replaces that.

Enjoy your holiday 🌞⛵️🏖

YashdalfTheGray:
Yeah, okay. Sounds good. I'm excited to see this merged in!

YashdalfTheGray:
🤔 travis just woke up?

mauricedb:
@YashdalfTheGray Nope, I did a `pull --rebase` and `push yolo` in the hope to get the Travis status and a merge but unfortunately there seems to be an error around `form-data` being included in a `package.json`. AFAICT not related to this PR.

types/yup/index.d.ts
```diff
@@ -12,7 +12,7 @@
 //                 Desmond Koh <https://github.com/deskoh>
 //                 Maurice de Beijer <https://github.com/mauricedb>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 2.9
```
andrewbranch: Just curious, what necessitates the bump to 2.9?

mauricedb: @andrewbranch When doing this update I had a error which was fixed by updating the TS version to 2.9. But trying to reproduce the same error now by reverting the version update gives me no error. So I will revert this change.

PR #36708, opened 2019-07-08
Refactor the type-declaration of inquirer

neurolag:

# About the Changes I've Made
So here's the thing...
I decided to completely refactor the type-declarations of `inquirer` which turned out to be pretty hard but finally I'm done.

This PR contains breaking changes which is why I increased the major version-number of this type-declaration.

I tried my best to make the type-declaration as extensible, intuitively and helpful as possible by allowing other modules to register their own question-types while preserving as much IntelliSense features as possible. Additionally I payed attention to get the whole `inquirer`-module covered.
In the previous type-declaration a lot components such as the ui- and prompt-modules have been missing.

I'd be glad if you could leave some feedback if there's anything I can improve or simplify.

## Registering Own Question-Types
As TypeScript does not allow to augment `type`s I decided to use an interface called `QuestionMap` to provide a collection of all available question-types.

People may now add their own prompt-types to the `inquirer`-module by writing a bunch of code like this:
```ts
import * as inquirer from "inquirer";
import InputPrompt = require("inquirer/lib/prompts/input");
/**
 * Provides options for the `ChalkPrompt`.
 */
export interface ChalkQuestionOptions<A extends inquirer.poll.Answers = inquirer.poll.Answers> extends inquirer.poll.Question<A> {
    previewColors: boolean;
}

/**
 * Provides options for a question for the `ChalkPrompt`.
 */
export interface ChalkQuestion<A extends inquirer.poll.Answers = inquirer.poll.Answers> extends ChalkQuestionOptions<A> {
    type: "chalk";
}

export class ChalkPrompt<TQuestion extends ChalkQuestionOptions = ChalkQuestionOptions> extends InputPrompt<TQuestion> {
    /* ... */
}

declare module "inquirer" {
    interface QuestionMap<T> {
        chalk: ChalkQuestion<T>;
    }
}

inquirer.registerPrompt("chalk", ChalkPrompt);
inquirer.prompt([ { type: "chalk", previewColors: true } ]); // Dah - `ChalkQuestion`s can be asked now!
```

## Intellisense-improvements
Things like property-names of choices and names of questions are now auto-completed in order to make writing questions even easier.

***Example:***
```ts
import Choices = require("inquirer/lib/objects/choices");

let choices = new Choices([], {});
choices.pluck("disabled"); // This name is auto-completed as it is a key of `CheckboxChoiceOptions`.
choices.pluck("helloWorld"); // This is working, too, even if it is not a key of any choice-type.
```

Also names of questions are auto-completed:
```ts
import inquirer = require("inquirer");

inquirer.prompt<{ width: number }>([
    {
        type: "number",
        name: "width", // This is auto-completed
    },
    {
        type: "number",
        name: "height" // This is working, too, even if it is not part of the answers-type.
    }]);
```

## Completeness
In this PR I've added type-declarations for every module of the `inquirer`-package which means that people can now extend from existing prompts in a TypeScript manner easily.

```ts
import InputPrompt = require("inquirer/lib/prompts/input");

class MyOwnPrompt extends InputPrompt {
    /* My Stuff */
}
```


types/inquirer/tslint.json
```diff
@@ -1,80 +1 @@
...
-        "strict-export-declare-modifiers": false,
-        "trim-file": false,
-        "triple-equals": false,
-        "typedef-whitespace": false,
-        "unified-signatures": false,
-        "void-return": false,
-        "whitespace": false
-    }
-}
+{ "extends": "dtslint/dt.json" }
```
andrewbranch: 🎉

types/inquirer/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for Inquirer.js 6.x
+// Type definitions for Inquirer.js 7.0
```
andrewbranch: This number is supposed to be purely just the npm version of the JavaScript library these typings are for. Since the type defs themselves have breaking changes, perhaps the best solution would be to wait until Inquirer actually ships a new major version 😬

andrewbranch (commented):
Thanks for all your effort here. We can’t merge something this significant without sign-off from the owners, though. The changes sound like valuable improvements, but they need to be a) validated by people familiar with Inquirer, and b) weighed against the cost of potentially breaking users. I’m not in the position to do the former (nor is anyone else on the TypeScript team as far as I know), and our job with the latter is to air on the side of caution. To that end, all I can do at the moment is ping the owners again to beg for help 😁 

@dfreeman @tkqubo @ppathan @jouderianjr @bang88 @bitjson @synarque @jrockwood @kwkelly @chigix @jedmao @bolasblack

types/inquirer/index.d.ts
manuth: It's a pity that it's not possible to change the version of the generated package somehow.
The major version has not been changed for about a year now so I really don't think there's going to be another major version in the near future.

types/inquirer/index.d.ts
```diff
@@ -10,430 +10,909 @@
...
+            new(question: any, readLine: ReadlineInterface, answers: Answers): PromptBase;
         }
 
         /**
-         * Sticky bottom bar user interface
+         * Provides a set of prompt-constructors.
          */
-        interface BottomBar extends BaseUI {
-            new(opt?: StreamOptions & { bottomBar?: string }): BottomBar;
+        interface PromptCollection {
```
jedmao:
```suggestion
        type PromptCollection = Record<string, PromptConstructor>;
```
You would also need to delete some lines below.

types/inquirer/inquirer-tests.ts
```diff
@@ -1,765 +1,49 @@
-import chalkPipe = require("chalk-pipe");
-import * as inquirer from "inquirer";
-import { from as rxjsFrom, Observable, Observer } from "rxjs";
-
-/**
- * Bottom Bar example
```
jedmao: I had copied all of these examples directly from https://github.com/SBoudrias/Inquirer.js/tree/master/packages/inquirer/examples and put each one inside their own `{}` scope, so as not to affect the other test cases. I'm not sure where you've gotten these new tests, but I think it makes more sense if they came from the provided examples in their GitHub repo than hand-typed. If you want to add more custom tests at the very bottom, that's fine, but why were all of these original tests removed?

I'm shocked, actually, that you've effectively just added a bunch more definitions, but deleted 99% of the tests in the process. Why was this done? I'm extremely concerned about this.

jednano (changes requested):
I appreciate the effort it took to flush out these definitions, complete with TSDoc comments, but I do have some concerns as well.

types/inquirer/index.d.ts
manuth: @jedmao
If the major version is not being increased, people using `@types/inquirer` might have to refactor parts of their code due to breaking changes.

types/inquirer/index.d.ts
```diff
@@ -10,430 +10,909 @@
...
+
+    /**
+     * Provides options for a question.
+     *
+     * @template T
+     * The type of the answers.
+     */
+    interface Question<T extends Answers = Answers> {
+        /**
+         * The type of the question.
```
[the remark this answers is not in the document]
manuth: I'm generally using `completed-docs` to ensure my js-docs covered every single part of my code.
This generally improves the IntelliSense in many code-editors.

types/inquirer/index.d.ts
```diff
@@ -10,430 +10,909 @@
...
-        when?: boolean | ((answers: A) => boolean | Promise<boolean>);
+        checked?: boolean;
     }
 
-    interface QuestionOptions<A> {
+    /**
+     * Provides options for a choice of the `ExpandPrompt`.
+     *
+     * @template T
+     * The type of the answers.
```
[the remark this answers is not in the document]
manuth: This might be a good idea.
I tried to do some coding similar to the ".NET"-library where all generic parameters are named `T` expect there are multiple generic types (like, for example, `List<T>` and `Dictionary<TKey, TValue>`).

If you want me to change the names of the generic parameters to `TAnswers` rather than `T` I'm, of course, willing to do so.

types/inquirer/index.d.ts
```diff
@@ -10,430 +10,909 @@
...
      */
-    interface Answer {
-        name: string,
-        answer: any,
+    interface QuestionMap<T extends Answers = Answers> {
+        /**
+         * The `InputQuestion` type.
```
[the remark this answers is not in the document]
manuth: As  I mentioned before I'm always paying attention to add js-docs to each part of my code.
I see this issue as resolved.

types/inquirer/index.d.ts
manuth: Thanks for the suggestion - I'm going to commit this instantly.

types/inquirer/inquirer-tests.ts
manuth: The code-examples provided here are very verbose.
The test-file is intended for testing each use-case of a module (in this case prompting questions, creating a prompt-module, creating a custom UI and stuff).

To me it's ways more clear to have these things covered with some brief test-snippets rather than using a bunch of code provided by a package's documentation.

If you want me to add the examples back I can do so, though.

types/inquirer/index.d.ts
jedmao: If all the pre-existing tests pass, I don't see why this would be the case. What's breaking? Sounds like an assumption.

types/inquirer/index.d.ts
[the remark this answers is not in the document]
jedmao: I'm not going to fight this, because it's a bit subjective, but I'm not loving the extra noise.

types/inquirer/index.d.ts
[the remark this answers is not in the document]
jedmao: If you're not going to remove the comment, then `T` is just fine.

types/inquirer/inquirer-tests.ts
jedmao: I'm not sure these DT tests can be compared to traditional app tests. The whole point of tests is to increase confidence. If the author of a library says it should work a certain way and provides example usage, it stands to reason that these examples should work in TypeScript w/o any kind of errors. The only way to increase my confidence to 100% is to include these examples in the DT tests; otherwise, I wonder if something was missed.

Are you 100% sure all of these examples work w/o errors? If so, how are you so sure? You already bumped the major version of these DT types because you were worried that existing implementations would break, so that tells me you're not 100% confident, defeating the whole purpose of the tests in the first place.

If you really want to delete all of these tests, then I suggest this PR be squashed down into 2 commits. The first commit would be all of your changes with the pre-existing tests preserved and this first commit would get a green checkmark on the PR before pushing the 2nd commit, where most of those tests are deleted. This would give me 100% confidence that this PR didn't break anything, but it would make future changes destroy that confidence (not my recommendation).

types/inquirer/index.d.ts
manuth: I renamed `Question<T>` to `DistinctQuestion<T>` in favor of adding a question-interface which represents the base of all question-interfaces.

That's the only breaking change so far.

jedmao: To @andrewbranch's point, I'm not sure you understand what this version represents. If I'm using an older version of Inquirer, say 6.3.1 and I `npm install @types/inquirer@6` then I would expect to get the latest types for Inquirer@6, which means less than 7. If you bump this "types" version to 7, this creates two problems:
1. `npm install @types/inquirer@6` isn't getting the latest types for the Inquirer's major version of 6, as they would expect.
1. When Inquirer v7 is actually published, the types for v7 are actually for v6, which is terribly confusing and wrong.

In short, this version is _never_ supposed to reference a non-existing or future version of any package. It's not a shame or a pity that we can't "change this somehow" because we're not supposed to be breaking current implementations and preserving existing tests is a great way to ensure that.

jednano (changes requested):
Still waiting for the tests to be restored.

types/inquirer/index.d.ts
manuth: Aah I get you.
I just updated the version in the header to "6.5" which is the actual current minor-version of inquirer.

types/inquirer/index.d.ts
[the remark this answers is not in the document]
manuth: Alright!
Let's keep it as it is then.

types/inquirer/inquirer-tests.ts
```diff
@@ -1,4 +1,3 @@
-import "./test";
```
jedmao: I'm not sure if this is going to work the way you think it is. If you look [here](https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package), you'll see that `foo-tests.ts` is the only file that will be tested; thus, the one that needs to do the importing.

Just thinking we might get some false positives with this new way you're doing it. Can you please just leave the tests as is? I understand your motivation to make these changes, but in fear that it goest against the DT guidelines, I think we should avoid this type of change.

jednano (changes requested):

types/inquirer/inquirer-tests.ts
manuth: @jedmao no worries
Some type-declarations such as [athenajs](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/athenajs) use the same approach.

Do you rather prefer the splitted version of the tests or should I put them back into the Tests-file?

jedmao: I mean, it looks promising, but I don't see it documented anywhere. I'd feel a lot better if this PR were 2 commits, where the first didn't move the files around and the 2nd was JUST the files being moved around. I'd also like to see the green checkmark on the 1st commit, most importantly.

jedmao: Unfortunately, the logging on Travis-CI isn't verbose enough to show me that these exact files were actually tested; only, they at least show up in the diff.

jednano (changes requested):
I test drove these types by pasting them into my `node_modules/@types/inquirer` folder and it broke all sorts of stuff I had relied on. I lost so much type information it's not even funny. I even went into `@types/yeoman-generator/index.d.ts` and changed `inquirer.Question` to `inquirer.DistinctQuestion`. And although my existing implementations didn't fail, if I tried to change the data types in the prompt questions it would also not fail. For example:

```ts
	public async prompting() {
		this.answers = await this.prompt([
			{
				type: 'confirm',
				message: 42,
				name: () => 'hi',
				when: 'iliketo',
			},
		])
	}
```

That code above should fail hard if the types are guarding properly, but these new types have no issues with that. When I restored the previous Inquirer types it complained about all of the issues real fast:

image

☝️ this is what I would expect to see when providing the right types.

I'm afraid in the current state of this PR, I cannot approve these changes. I also noticed you hadn't updated the `yeoman-generator/index.d.ts` to use the new `DistinctQuestion` in this PR?

types/inquirer/inquirer-tests.ts
manuth: Aah I got your point
I just tested whether errors in one of the `types/inquirer/test/*.ts`-files are reported when running either `npm run lint inquirer` or `npm test`.

These files are considered in the test-commands as they are added to the `files`-property in the `types/inquirer/tsconfig.json`-file.

Also files placed in a directory called `test/` or files which are called `{package-name}-tests.ts(x)?` are considered as test-files:
https://github.com/microsoft/types-publisher/blob/c4762a372a40fa3dd26f7b7308d8257ba551e093/src/lib/definition-parser.ts#L237-L251

manuth:
Oh geez you're right!
I'm sorry I totally forgot about changing the depdendent packages.

I'll merge this immediately

manuth:
@jedmao dependent packages should be fixed now.
Could you have a look, please?
Thanks for the effort! :smile:

jednano (changes requested):
I don't think you're understanding what I'm saying. I already changed `yeoman-generator` to use the `DistinctQuestion` in my local environment and still the type information I relied upon before was lost. Even now, I pulled down your branch and pasted all your changes into my `@types` folder and now it's still broken.

For example, if I change a question type to `foo` it should complain like this:

Screen Shot 2019-07-26 at 11 38 00 AM

But instead, your types are allowing me to do this without any kind of errors. I think, conceptually, what you're trying to do with the `DistinctQuestion` is noble, but it doesn't actually work the way you think. It's too forgiving.

In lieu of all this, I think we need an additional test to show that adding a question of `type: 'foo'` fails.

manuth:
Oooh I see
Well the types I wrote are supposed to allow any string value in the `type`-field while still providing proposed values using IntelliSense.

Is this a bad idea?
Because now that you say it:
If anyone adds a custom `type` or an alias or something they are supposed to add this change to the question-map, right?

jedmao:
I do think it's a bad idea, because what if someone accidentally typed `type: 'inputs'` with an `s` on the end of `input` and didn't realize it? We should be explicit about these things. And yes, if they have a custom type they just need to add it to the question map, as you say.

manuth:
So it looks just like my current solution is treating `type` restrictive already... but TypeScript doesn't seem to handle it correctly, hope you have some advices for me - here's the thing:
If I declare a variable either as a `DistinctQuestion` or a `ReadonlyArray<DistinctQuestion>` providing an incorrect `type` causes an error, as desired:
```ts
const x: DistinctQuestion = {
    // Causes an error
    type: "inputs"
};

const y: ReadonlyArray<DistinctQuestion> = [
    {
        // Causes an error
        type: "inputs"
    }
];
```

But if I declare a variable whose type is `DistinctQuestion | ReadonlyArray<DistinctQuestion>` it doesn't show an error...?
```ts
const z: DistinctQuestion | ReadonlyArray<DistinctQuestion> = [
    {
        // Causes no error
        type: "inputs"
    }
];
```

@jedmao is there anything I can do?

jedmao:
@manuth can you reproduce this in the [TypeScript Playground](http://www.typescriptlang.org/play/) in a minimal example to illustrate the point? We need to make sure it's not a TS bug first. If it is, it needs to be filed. If it's not, there's work to do on this PR.

manuth:
I finally found what's causing the error and it's TypeScript's expected behavior.
Since `Question` has a `filter(input: any): any)`-method in common with the `Array`-class any sort of array is assignable to `Question`.

All fields of the `InputQuestion` are optional which is the reason why any sort of `Array`-class is assignable to `InputQuestion`, too.

This means you could also to something like
```ts
let x: InputQuestion = [
    {
        fizz: "buzz"
    }
];
```
or
```ts
let x: InputQuestion = [
    {
        type: "doesnotexist"
    }
];
```
So when I provide an array with an invalid question to the `prompt`-method it will just be considered as an `InputQuestion`.

[This TypeScript-Playground](Playground) makes it a lot easier to understand - took me a while to find what's causing the issue...

@jedmao Any shall we try to find some way to patch this?
Honestly I don't think there is some way to do so... expect to make some field non-optional like, for example, "message" which is required by inquirer anyway.

[Playground]: https://www.typescriptlang.org/play/index.html#code/JYOwLgpgTgZghgYwgAgIoFcIGczAPYjIDeAsAFDLJgCeADhAPwBcyOUoA5gNzmUzAAbSFAYAKULXRgWcENQCUMuTzIBfcuVDD4SZAEkQksBmy4CyCAA9IIACZY0mHPkKkKlKnUYsARBKk+KupkmuDQOigAorbAYHhQJs7mVjb2jqYuxLweNPS+EDFxUIHkwaHaiCiJZiAAsnC0We6U-tL6hlLVLioeFoXxLNGx8V0EQRpkuSgAIsDOIAjGTjXIALzpSXUNANoA1hDUeDAbNfW0ALoq5FMnLgDCeAICEIuZ67Pzi6OEAD7IAEoQOC2AgCagAQSgUDg1AAPB9cAslhkCAA+K5kZ5gZCWFgGIzfNbIbbZShuXq9Ka+AAWECeeB8pOQwUuEyxyGog36CWWb2JTPJFMoVOQPlp9MZzWZ5HObIg2IAXixvg8ni8VusSe5BcKvDS6QIGdkWRjyEA

manuth:
@jedmao, @andrewbranch What do you guys think of forcing the user to provide an array of questions rather than allowing them to provide a single question only?

jedmao:
I'm also not sure why you're trying to set a variable as a union type of a single question or an array of questions when you know it's an array, based on the value you are signing to it.

manuth:
As you might remember prompt-methods allow either a single question or an array of questions - a union type of `ReadonlyArray<Question>` and `Question`.

Providing arguments to a method which accepts `ReadonlyArray<Question> | Question` and assigning a variable of type `ReadonlyArray<Question> | Question` both causes the same tslint/tsc-errors.

I hope I could make clear why [my playground][Playground] looks like it does.

[Playground]: https://www.typescriptlang.org/play/index.html#code/JYOwLgpgTgZghgYwgAgIoFcIGczAPYjIDeAsAFDLJgCeADhAPwBcyOUoA5gNzmUzAAbSFAYAKULXRgWcENQCUMuV2QB6VcgAqUalTzIEeALZGI4KgAtgWZANARyAX3LlQw+EmQBJEJLAZsXAJkCAAPSBAAExsAnHxCUgpKKjpGFgAiCSl0njJnMldwaA8UAFFI4DA8KFigwjCI6LRMOODE5JT6DIgKqqgcpxcyN2LEFFr4gFk4WmJeZKzpb18pCYJcjp7K6pZy7ZqWutz88hp6ZAARa1wQBH9D+OQAXmbAqZmAbQBrCGo8GFerRA01oAF1cqdUoC6gBhPACAQQO6PF5XOK3e5vYIAH2QACUIHBIgQBNQAIJQKBwagAHjRNzuaxAAD4IWREWBkKEWD4-EznsgPvNKO0OskzhAMhYIAi8Olhch8uChhzkNRdr1qvyXkKkiKFR0JVKZQI5QqlSqIJyAF4sJlwhFIuoC3X6vWdSXIdLS2XypJKtnkIA

manuth:
Yet if we allow the user to provide either an array of questions or a single question we will have the issue that the user can provide any kind of array to the `prompt`-method as an array is considered as an `InputQuestion`.
Please note that this error also existed for the current version `6.x` of `@types/inquirer` so we won't face a loss of quality.

There are two things coming to my mind we could change in order to fix this issue, which is:
1. Drop `Question` from the `QuestionCollection`-union to make sure only array or observables are provided to the `prompt`-method
2. forcing the user to specify the type of the question (make `type: "input"` mandatory)

manuth:
Personally I'd prefer to change the `QuestionCollection` from
```ts
   type QuestionCollection<T extends Answers = Answers> =
        | DistinctQuestion<T>
        | ReadonlyArray<DistinctQuestion<T>>
        | Observable<DistinctQuestion<T>>;
```
to
```ts
    type QuestionCollection<T extends Answers = Answers> =
        | ReadonlyArray<DistinctQuestion<T>>
        | Observable<DistinctQuestion<T>>;
```

This would improve the intellisense significantly as it prevents `Array`-`Question`-confusion.

jedmao:
> Yet if we allow the user to provide either an array of questions or a single question we will have the issue that the user can provide any kind of array

The current types prevent you from having any kind of array. Refer to my previous screenshot in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36708#pullrequestreview-267291378:

Screen Shot 2019-07-26 at 11 38 00 AM

manuth:
Haven't noticed this yet, my bad.
It's because the current `filter()` method of the questions is incorrect:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1895488dec906b3a1c022899df50e6699dc20695/types/inquirer/index.d.ts#L125

As you can see the filter-method accepts a `string` while the result of a prompt could be anything.
For the `NumberPrompt`, for example, a number is passed to the question's `filter`-method rather than a string:
https://github.com/SBoudrias/Inquirer.js/blob/e05ae81019a6b524917b2453c107aa93d661b705/packages/inquirer/lib/prompts/input.js#L23
https://github.com/SBoudrias/Inquirer.js/blob/e05ae81019a6b524917b2453c107aa93d661b705/packages/inquirer/lib/prompts/number.js#L13-L26

So - as you can see - the `filter()`-method **must** be able to accept any type of `input` as you can literally resolve your prompt with whatever answer-type you want.

The only reason why arrays can't be assigned to `Question`s currently is because the type of `Question["filter"]` is incorrect.

Hope I could make things clear.
Thanks for taking time

jednano (approved):
I 100% understand what you mean. Thanks for the clarification!

I'm a bit bothered that I'm going to lose this type safety, but I'm not sure I see a way around it. The new (correct) types would not break existing implementations, so I think I'm 👍 with these changes.

Thanks for your comprehensive work on these type definitions! 🎉

manuth:
Well I think I remember that an error in the inquirer code says that the name-property of a question is required.

We could just make `name` non-optional which would cause that arrays are not assignable to questions anymore.
What do you think about that?

jedmao:
I don't remember that. Pretty sure all properties are optional in a `Question` until I see documentation or tests that say otherwise.

manuth:
The only thing I could find was [this here](https://github.com/SBoudrias/Inquirer.js/#prompt-types)
According to this document `type`, `name` and `message` are always required.

This is very inaccurate as `type` might be skipped which causes the default prompt to be used.
`message` could also be skipped as it would be replaced by `name` according to [the docs](https://github.com/SBoudrias/Inquirer.js/#question).

jedmao:
Yes, I think I also came across the same inconsistency (at least with the `type` field) when I was reading the docs. Of course, I prefer type definitions over misleading documentation.

manuth:
I totally agree
I think we could just leave my PR as it is.

jedmao:
@manuth I thought you fixed this to be a number?

Screen Shot 2019-07-31 at 2 17 50 PM

manuth:
Ah no - I didn't do so - all Question-Type's filter-method takes an `any`-parameter

PR #36715, opened 2019-07-08
[phoenix] Fix SocketConnectOptions for phoenix >= 1.4.3

ybycode:
Adds/fixes types of SocketConnectOptions attributes:

- Adds the type of `binaryType`,
- Adds the type of `rejoinAfterMs`,
- Fixes the type of `reconnectAfterMs`.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://hexdocs.pm/phoenix/js/#socket
- [n/a] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

### DIFF
--- a/types/phoenix/index.d.ts
+++ b/types/phoenix/index.d.ts
@@ -36,18 +36,21 @@ export class Channel {
   push(event: string, payload: object, timeout?: number): Push;
 }
 
+export type BinaryType = 'arraybuffer' | 'blob';
 export type ConnectionState = 'connecting' | 'open' | 'closing' | 'closed';
 
 export interface SocketConnectOption {
+  binaryType: BinaryType;
   params: object | (() => object);
   transport: string;
   timeout: number;
   heartbeatIntervalMs: number;
-  reconnectAfterMs: number;
   longpollerTimeout: number;
   encode: (payload: object, callback: (encoded: any) => void) => void;
   decode: (payload: string, callback: (decoded: any) => void) => void;
   logger: (kind: string, message: string, data: any) => void;
+  reconnectAfterMs: (tries: number) => number;
+  rejoinAfterMs: (tries: number) => number;
 }
 
 export class Socket {
--- a/types/phoenix/phoenix-tests.ts
+++ b/types/phoenix/phoenix-tests.ts
@@ -1,7 +1,12 @@
 import { Socket, Channel, Presence } from 'phoenix';
 
 function test_socket() {
-  const socket = new Socket('/ws', {params: {userToken: '123'}});
+  const socket = new Socket('/ws', {
+      binaryType: 'arraybuffer',
+      params: { userToken: '123' },
+      reconnectAfterMs: tries => 1000,
+      rejoinAfterMs: tries => 1000,
+  });
   socket.connect();
 }
 


types/phoenix/index.d.ts
```diff
@@ -39,15 +39,17 @@ export class Channel {
 export type ConnectionState = 'connecting' | 'open' | 'closing' | 'closed';
 
 export interface SocketConnectOption {
+  binaryType: string;
```
princemaple: do we know all the possible binary types? If so, we may want to list them here.

princemaple (approved):
That one question

ybycode:
Hi all,
None of the classes attributes are provided in the type. Is it on purpose, or simply because so far no one needed them?
I'm asking because in my app (I'm transitioning an existing JS app to TS), we use `channel.joinPush.receive('ok', callback)` to know when the channel is up.
`channel.join()` returns this `Push` instance, but a bit too late to subscribe to its events.
So I'm tempted to describe this attribute in the types. Any opinions?

types/phoenix/index.d.ts
ybycode: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/binaryType lists "arraybuffer" and "blob", but I'm not sure if that's an exhaustive list. I couldn't find its use in the (elixir) phoenix source code either.

princemaple:
@ybycode I don't completely get what you were saying. Mind try a more detailed example? Or point out what you are going to add.

types/phoenix/index.d.ts
princemaple: MDN tends to have comprehensive docs. The two are probably all the possibilities. Would you like to update the PR? I'll take a look at the Phoenix source tomorrow, too. Past midnight in Australia.

ybycode:
@princemaple in my JS code I currently use the `joinRef`attribute of the class `Channel`. But the types don't describe any of the attributes of the class at the moment. So I was wondering if that was because they were considered private, or simply because no one had use of them so far.

As an example, here's how I use it in my JS code:
```js
    onStatusUpdate (channel, callback) {
      // Adds a callback to lifecycle events of the given channel.
      // The given callback will get called with the new channel status as argument.
      channel.onError(() => callback('down'))
      channel.onClose(() => callback('closed'))
      // because there's no 'channel.onJoined': 
      channel.joinPush.receive('ok', () => callback('up'))
    },
```

And here's the minimal addition I'd make to the Channel class declaration (for this specific need): 

```ts
export class Channel {
  constructor(topic: string, params?: object | (() => object), socket?: Socket);

  // (...)

  joinPush: Push; // <--
}
```

types/phoenix/index.d.ts
ybycode: I updated the commit to define the `BinaryType` type.

armanio123:
@ybycode seems like you're still working on this pull request, I'll wait from merging until you can confirm is ready. Let me know.

princemaple:
@ybycode I think it's considered private, as an implementation detail. Also, it's returned by `Channel#join`. The return type of `Channel#join` correctly states `Push`. 

The official way of doing it is in the phoenix js docs:

```js
 channel.join()
   .receive("ok", ({messages}) => console.log("catching up", messages) )
   .receive("error", ({reason}) => console.log("failed join", reason) )
   .receive("timeout", () => console.log("Networking issue. Still waiting..."))
```

ybycode:
@princemaple ok, then let's keep the PR as it currently is if that's ok
with you.

princemaple:
Hi @armanio123 please take another look. @ybycode and I have agreed that this one is good to go in.

armanio123:
> Hi @armanio123 please take another look. @ybycode and I have agreed that this one is good to go in.

Sounds good, merging.

PR #36739, opened 2019-07-09
[react-syntax-highlighter] update path declarations to respect 11.0.1

maxchehab:

The 11.0.1 version of react-syntax-highlighter builds both `cjs` and `esm` distributions. [The package, by default, uses `/dist/cjs` directory.](https://github.com/conorhastings/react-syntax-highlighter/blob/99fb9f3ec36e0d539b3110716e22d645fad860d3/package.json#L5-L6)

This PR updates the declaration paths to use the default `cjs` directory.


guoyunhe:
I see the [README of react-syntax-highlighter](https://github.com/conorhastings/react-syntax-highlighter#prism) always use `esm` module.

package.json
```
{
  "main": "dist/cjs/index.js",
  "module": "dist/esm/index.js",
}
```
means if it is imported with webpack/babel (most likely for a project that uses TypeScript), it will use `esm` module.

So I suggest to use `esm` module instead of `cjs`.

andrewbranch (changes requested):
From what I can tell, I think it wouldn’t hurt to declare types for both cjs _and_ esm directories. TypeScript can handle either import and it should have no effect on the types. (That said, it seems like a wildcard declaration might be more ergonomic since these are just exporting `any` over and over again.)

I have no real stake or opinion in the decision, but I’m requesting changes on behalf of @guoyunhe’s comment so that @typescript-bot will categorize the status correctly.

maxchehab:
> (That said, it seems like a wildcard declaration might be more ergonomic since these are just exporting `any` over and over again.)
@andrewbranch what does a wildcard declaration look like?

I'm sorry for the lag guys 🙍‍♂

andrewbranch:
```ts
declare const _: any;
declare module "react-syntax-highlighter/dist/*" {
  export default _;
}
```

The disadvantage is that you don’t get validation that the paths actually exist. I guess I probably wouldn’t recommend that, on second thought. Sorry for the delay, I accidentally missed this notification :/

Hypnosphi:
> I think it wouldn’t hurt to declare types for both cjs and esm directories

Sorry, but why the final decision was not to do that? Using `cjs` imports makes sense in a project that uses Jest, which doesn't transpile `node_modules` without extra configuration

guoyunhe:
es6 modules are best for webpack etc. keeping two type of types will increase the size of typing and it is not easy to maintain. if you have issue of importing node_modules, it is probably `react-syntax-highlighter`'s issue.

Hypnosphi:
Ok, to give more context:

1. I want to use `react-syntax-highlighter` in a library `A`
2. End users of library `A` test their components with Jest

If I use imports from `dist/esm`, all the end users will have to update their `jest.config.js` with a line like that:
```js
transformIgnorePatterns: ['node_modules/(?!react-syntax-highlighter)']
```

Technically, that means that adding a dependency on `react-syntax-highlighter` becomes a breaking change for `A`.

At the same time, if I use imports from `dist/cjs`, all the end users can keep using `A` without any changes to their configs, so there can be no breaking changes.

Does it make sense?

guoyunhe:
Jest should be able to consume ESM, even raw TypeScript. You just need to add some Babel/TypeScript presets. Like in our project, we have `jest.config.json`:

```json
{
  "bail": true,
  "coverageThreshold": {
    "global": {
      "branches": 20,
      "functions": 20,
      "lines": 30,
      "statements": 30
    }
  },
  "preset": "ts-jest/presets/js-with-babel",
  "setupFilesAfterEnv": ["<rootDir>/jest.setup.ts"],
  "snapshotSerializers": ["enzyme-to-json/serializer"],
  "testMatch": [
    "**/?(*.)+(spec|test).(js|ts)?(x)"
  ],
  "testPathIgnorePatterns": [
    "<rootDir>/dist/",
    "<rootDir>/node_modules/"
  ]
}
```

Hypnosphi:
@guoyunhe do you import any ESM from `node_modules` in your project?

guoyunhe:
Many packages provide two format: UMD and ES6.

```json
{
  "main": "dist/foobar.umd.js",
  "module": "lib/index.js"
}
```

We import without path:

```js
import something from 'foobar'
```

Jest or webpack will pick the format they like.

The problem of `react-syntax-highlighter` is that they expose multiple files. So this magic doesn't work anymore. Maybe using cjs format is the best solution.

However, this PR isn't the way to go. It only supports cjs format. It should support both cjs and esm formats. I can make a PR for this change.

guoyunhe:
Checkout #41345

PR #36752, opened 2019-07-09
Elasticsearch: Changed type to optional field in IndicesPutMappingParams interface to bring it in line with the optional nature of the 7.x api version and it's future deprecation

Dmajster:
https://www.elastic.co/guide/en/elasticsearch/reference/current/removal-of-types.html


Dmajster:
Type repeats over several other definitions as well. Will update them when i'll be able to.

PR #36758, opened 2019-07-09
TypeScript definition for react-transition-group's new SwitchTransition component

AdrianFleming:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/react-transition-group/SwitchTransition.d.ts
+++ b/types/react-transition-group/SwitchTransition.d.ts
@@ -1,5 +1,5 @@
 import { Component } from "react";
-import { TransitionProps } from "react-transition-group/Transition";
+import { TransitionProps } from "./Transition";
 
 declare namespace SwitchTransition {
     interface SwitchTransitionProps extends TransitionProps {
--- a/types/react-transition-group/SwitchTransition.d.ts
+++ b/types/react-transition-group/SwitchTransition.d.ts
@@ -1,0 +1,12 @@
+import { Component } from "react";
+import { TransitionProps } from "react-transition-group/Transition";
+
+declare namespace SwitchTransition {
+    interface SwitchTransitionProps extends TransitionProps {
+        mode?: 'out-in' | 'in-out';
+    }
+}
+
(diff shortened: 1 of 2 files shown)


types/react-transition-group/SwitchTransition.d.ts
```diff
@@ -0,0 +1,12 @@
+import { Component } from "react";
+import { TransitionProps } from "./Transition";
+
+declare namespace SwitchTransition {
+    interface SwitchTransitionProps extends TransitionProps {
+        mode?: 'out-in' | 'in-out';
```
ybiquitous: In the [original code](https://github.com/reactjs/react-transition-group/blob/f50fae0392e0beeb3989db9ddc59c04ecb20cf09/src/SwitchTransition.js#L23-L26) `modes` is also exported, so what about adding it like this? 👇 

```tsx
const modes = {
  out: 'out-in',
  in: 'in-out'
}

interface SwitchTransitionProps extends TransitionProps {
  mode?: modes.out | modes.in;
}
```

ybiquitous (changes requested):
Sorry for too late replay 🙇 

It seems that new test cases are missing. Could you add new test cases to the following file?

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f18c0148617a59a32e4f1a377a1b9a996bc5abfc/types/react-transition-group/react-transition-group-tests.tsx#L1

types/react-transition-group/SwitchTransition.d.ts
AdrianFleming: @ybiquitous The above fails to build with an error "A 'const' initializer in an ambient context must be a string or numeric literal or literal enum reference"

ybiquitous: Oh, sorry. It seems good to use `enum` instead of `const` here.

By the way, I notice PR #37061 has been opened. 👀

AdrianFleming: @ybiquitous thanks, PR #37061 does what I wanted it to :)

PR #36765, opened 2019-07-09
react-router Make Route autocomplete its props

TedDriggs:
The `T extends RouteProps = RouteProps` only provides property autocompletion on the first property entered, after which it forgets what type we're using. Adding `& RouteProps` to the `extends React.Component` declaration fixes this issue.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: N/A

### DIFF
--- a/types/react-router/index.d.ts
+++ b/types/react-router/index.d.ts
@@ -92,7 +92,7 @@ export interface RouteProps {
   sensitive?: boolean;
   strict?: boolean;
 }
-export class Route<T extends RouteProps = RouteProps> extends React.Component<T, any> { }
+export class Route<T extends RouteProps = RouteProps> extends React.Component<T & RouteProps, any> { }
 
 export interface RouterProps {
   history: H.History;


TedDriggs:
@johnnyreilly I'm struggling to figure out how this change caused the redux test to fail; any chance I could ask for your help on understanding how those two relate?

johnnyreilly:
@TedDriggs I'm really sorry but I'm kinda maxed out right now and so can't help I'm afraid. My guess is that there's a test in redux which is affected by your changes; though that seems surprising.

johnnyreilly:
Actually I had a quick look at the error in the logs:

```
ERROR: 23:50  expect  TypeScript@next compile error: 
Type 'State' does not satisfy the constraint 'ReducersMapObject<any, any>'.
  Index signature is missing in type 'State'.
```

My guess is you're being bitten by new strictness in the TypeScript compiler. You may want to ask one of the TypeScript team to advise... @andrewbranch any thoughts?

andrewbranch:
I haven’t looked at this super closely nor https://github.com/microsoft/TypeScript/pull/32100 super closely _yet_, but I wonder if they might be related? In general, I would probably recommend against type hacks that make the editing experience nicer in favor of opening TypeScript bugs when the editing experience isn’t good. (That’s not to say we can _never_ accept a clever workaround that makes the experience better, but hopefully if there’s an experience bug, we can just fix it for everyone.)

If you’re determined to keep working on this, I would recommend you give that PR a try in the process ☝️

andrewbranch:
I’m looking at the failed build logs now, though, and while my point still stands, I think the build failures are probably completely unrelated to this, as @johnnyreilly posited.

PR #36779, opened 2019-07-09
[braintree] Add paymentMethodNonce to ValidatedResponse

schmod:
This appears to have been accidentally omitted from the ValidatedResponse type

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://developers.braintreepayments.com/reference/response/payment-method-nonce/node

### DIFF
--- a/types/braintree/braintree-tests.ts
+++ b/types/braintree/braintree-tests.ts
@@ -1,7 +1,8 @@
 import braintree = require('braintree');
 import { BraintreeGateway, Address, AddressCreateRequest,
     CreditCard, Customer, PayPalAccount, ApplePayCard, AndroidPayCard,
-    VisaCheckoutCard, SamsungPayCard, MasterpassCard, PaymentMethod, Transaction
+    VisaCheckoutCard, SamsungPayCard, MasterpassCard, PaymentMethod,
+    PaymentMethodNonce, Transaction
 } from 'braintree';
 
 /**
@@ -55,6 +56,11 @@ const gateway: BraintreeGateway = new braintree.BraintreeGateway({
     const visaCheckoutCard = <VisaCheckoutCard> response.paymentMethod;
     const samsungPayCard = <SamsungPayCard> response.paymentMethod;
     const masterpassCard = <MasterpassCard> response.paymentMethod;
+})();
+
+(async () => {
+  const response = await gateway.paymentMethodNonce.create('token');
+  const nonce = response.paymentMethodNonce.nonce;
 })();
 
 (async () => {
--- a/types/braintree/index.d.ts
+++ b/types/braintree/index.d.ts
@@ -59,6 +59,7 @@ declare namespace braintree {
         dispute: T extends Dispute ? Dispute : never;
         merchantAccount: T extends MerchantAccount ? MerchantAccount : never;
         paymentMethod: T extends PaymentMethod ? PaymentMethod : never;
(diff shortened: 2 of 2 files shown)


schmod:
@andrewbranch -- Love the new performance tool!  Unfortunately, it looks like it's quite noisy in the case of this very simple change (which seems extremely unlikely to have caused any meaningful regression).

andrewbranch:
Thanks for the feedback! It’s useful to hear when things aren’t optimal.

I’m not terribly concerned about the noise level, and unfortunately very simple changes can sometimes be very expensive—there’s really not a surefire way to know that a change is innocuous without actually testing it.

I do agree that this change is obviously not a problem for performance, which is why we don’t block merges or even fail builds when the perf testing turns something up—it’s designed to be a tool that human reviewers can pair with their human sensibilities.

I’m not exactly sure why this thinks there might be a problem, but my guess is that the test file doesn’t instantiate very many conditional types, so calling one more `create()` actually causes the checker to do, proportionally, a significant amount more work than it was doing before. In this case, it means that your tests improved, not that your definitions regressed.

I do plan, when I get a chance, to set some thresholds below which we won’t give you any warnings. E.g., 125 ms is 54% worse than 80 ms, but they’re both really good, so it’s nothing to worry about. We’ve also talked about doing comparisons on a curve of _all_ definitions on DT, not just against the master branch version of the affected definitions. I’m pretty excited about that, but it will take some more time and investment.

I’m going to kick off another run, just to determine if the higher numbers were a fluke of environmental noise or if my theory was right. Cheers!

andrewbranch:
Yep, it’s more or less consistent. So, false positive? Depends on how you look at it. All we can possibly measure is a proxy of what it would be like to edit the test file, as a further proxy of the editing experience of users consuming these typings. Certain things in the tests _did_ get slightly slower, but the impact is probably due to the tests themselves rather than the definitions. (This is why we report the number of identifiers found in test files.) I tend to think the bot did its job, if a bit inelegantly—even though there are a couple warnings, a human can easily correlate them with the changes and see that this PR is safe.

smrubin (approved):
Great catch @schmod and thanks for adding the test too!

schmod:
@andrewbranch -- Is there any reason why the PR and `master` benchmark isn't on the same machine?  I'd imagine that you'd get a more consistent reference point for comparisons if you ran the benchmarks back-to-back.

andrewbranch:
_Some_ definitions are extremely expensive (time-wise) to measure because the test files are _massive_. In order to expedite the PR process, we try to get a head start by running baselines against master overnight. That way, when a PR is opened, we check to see if there’s a usable baseline against master already saved, and we use that. It cuts the feedback cycle time in half, which is significant for larger definitions. The machines have to be _similar_ in order to reuse an existing baseline. We’re primarily concerned with catching order-of-magnitude-scale regressions, so the small differences are an acceptable price to pay for making folks wait less to see results. And again, that’s why we _report_ the machine differences when they exist—so a human can take them into account.

In the future, we could consider re-running baselines on definitions we already know to be fast, but I don’t see a really strong reason to do that—we know that we’re looking at a fairly wide margin of error, and if we want to narrow it in the future, we’ll have to invest in dedicated hardware in order to have total control over background processes and to be immune to noisy neighbor issues. We’re running these baselines on standard cloud build agents at the moment as we try things out, because they’re cheap, convenient, and sufficient for detecting large regressions.

PR #36785, opened 2019-07-10
[enzyme] Add `wrappingComponent` and other related props support to `mount` function

jtomaszewski:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/airbnb/enzyme/blob/master/docs/api/mount.md#arguments

### DIFF
--- a/types/enzyme/index.d.ts
+++ b/types/enzyme/index.d.ts
@@ -12,6 +12,7 @@
 //                 Mateusz Sokoła <https://github.com/mateuszsokola>
 //                 Braiden Cutforth <https://github.com/braidencutforth>
 //                 Erick Zhao <https://github.com/erickzhao>
+//                 Jack Tomaszewski <https://github.com/jtomaszewski>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.1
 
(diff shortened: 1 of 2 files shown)


types/enzyme/index.d.ts
```diff
@@ -642,15 +650,15 @@ export interface ShallowRendererProps {
...
+     * See the [getWrappingComponent() docs](https://airbnb.io/enzyme/docs/api/ShallowWrapper/getWrappingComponent.html) for an example.
+     * **Note**: `wrappingComponent` must render its children.
      */
     wrappingComponent?: ComponentType<any>;
     /**
-     * Initial props to pass to the wrappingComponent if it is specified.
+     * Initial props to pass to the `wrappingComponent` if it is specified.
      */
-    wrappingComponentProps?: any;
+    wrappingComponentProps?: {};
```
andrewbranch: Nit: [`object` is probably a better match for your intention here](https://www.typescriptlang.org/play/#code/G4QwTgBAHgXBDeBfCBeCBydAoUkCecA9gEYBWApgMYAuqG6QA), although I guess there’s precedent elsewhere in this file for `{}` ¯\\_(ツ)_/¯

types/enzyme/index.d.ts
```diff
@@ -597,6 +598,13 @@ export class ReactWrapper<P = {}, S = {}, C = Component> {
      * Returns a wrapper with the direct parent of the node in the current wrapper.
      */
     parent(): ReactWrapper<any, any>;
+
+    /**
+     * If a wrappingComponent was passed in options,
+     * this methods returns a ReactWrapper around the rendered wrappingComponent.
+     * This ReactWrapper can be used to update the wrappingComponent's props and state
+     */
+    getWrappingComponent: () => ReactWrapper;
```
andrewbranch: I’m supposing this returns `undefined` in some circumstances, but the same can be said about almost literally every other method here—I think conveniently “optimistic” types like these make sense for a testing library. (Just thinking out loud)

andrewbranch (approved):

PR #36789, opened 2019-07-10
slate-react: Separated type for onChange function

yurist38:
A small change in order to be able to import and re-use types of `onChange` function and its parameter.

### DIFF
--- a/types/slate-react/index.d.ts
+++ b/types/slate-react/index.d.ts
@@ -114,12 +114,18 @@ export interface Plugin extends CorePlugin {
 export type PluginOrPlugins = Plugin | Plugins;
 export interface Plugins extends Array<PluginOrPlugins> {}
 
+export interface OnChangeParam {
+    operations: Immutable.List<Operation>;
+    value: Value
+}
+export type OnChangeFn = (change: OnChangeParam) => any;
+
 export interface BasicEditorProps {
     value: Value;
     autoCorrect?: boolean;
     autoFocus?: boolean;
     className?: string;
-    onChange?: (change: { operations: Immutable.List<Operation>; value: Value }) => any;
+    onChange?: OnChangeFn;
     placeholder?: any;
     plugins?: Plugins;
     readOnly?: boolean;
--- a/types/slate-react/index.d.ts
+++ b/types/slate-react/index.d.ts
@@ -116,7 +116,7 @@ export interface Plugins extends Array<PluginOrPlugins> {}
 
 export interface OnChangeParam {
     operations: Immutable.List<Operation>;
-    value: Value
+    value: Value;
 }
 export type OnChangeFn = (change: OnChangeParam) => any;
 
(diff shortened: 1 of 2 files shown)


andrewbranch:
I would suggest, rather than using this PR, doing

```ts
import { BasicEditorProps } from 'slate-react';
type OnChangeFunction = BasicEditorProps['onChange'];

// Or, if you need to remove `undefined` from the domain:
type OnChangeFunction = Exclude<BasicEditorProps['onChange'], undefined>;

// To get the parameter type:
type OnChangeParam = OnChangeFunction extends (p: infer P) => unknown ? P : never;
```

It seems like having a unique reference to this type is a bit of an edge case, and as such I think it’s better not to add new types to the exports when you can get them other ways.

yurist38:
@andrewbranch To be honest, I don't agree that it's an edge case. If you write the event handler function (what you most probably will do when using this component) you will need those types. And the way you suggested not a friendly one, too complicated, as for me. Consumers of the package should not do this, it's much easier to simply export it. I don't see any problem with exporting it.

andrewbranch:
Consumers of the package needn’t do either. I’m actually a Slate user myself, coincidentally, and I used contextual typing:  https://github.com/andrewbranch/blog/blob/50939b0baf65839a724543694578ef23952d8697/src/components/InteractiveCodeBlock/InteractiveCodeBlock.tsx#L235

If I had really wanted to write it as a separate function, I don’t think it’s unfriendly to do

```tsx
const onChange: EditorProps['onChange'] = ({ value, operations }) => {
  //
};

<Editor onChange={onChange} />
```

([Note this technique is already used in slate-react typings hundreds of times.](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36789/files#diff-5d1ba452e78bcfee7592a6cc1c9b7ccaL151))

There’s no real problem with exporting it, it’s just a bit unusual. And it’s unclear why `onChange` should get its own exported type and not, for example, [decorateNode, renderEditor, renderMark, renderBlock, renderDocument, renderInline, and shouldNodeComponentUpdate](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36789/files#diff-5d1ba452e78bcfee7592a6cc1c9b7ccaR85). I think things would start to get messy if all of them had named, exported types for their functions and their params.

But, my job isn’t to stop you, I just wanted to offer an alternative in the absence of @andykent, @majelbstoat, @JanLoebel, @PatrickSachs, @YangusKhan, @isubasti, @sgreav, @Kornil, @jackall3n, and @benjiro—I would really prefer for one of them to weigh in. I’ll wait a little longer after this second ping, and then come back around after a couple days to merge this if nobody else reviews. Thanks for hearing me out!

yurist38:
Thank you for the explanation @andrewbranch ! I've got your point and the alternative you've suggested is a good option too. Let's see maybe we will get more feedback on this. Thanks again for your time! 👍

andrewbranch:
Well, no one else showed up, so I guess you win by default 😄

PR #36791, opened 2019-07-10
Add type definitions for service-worker-mock

remcohaszing:
Please fill in this template.

If adding a new definition:

Alternative typings are available [here](https://github.com/udacity/cloudflare-typescript-workers/tree/master/packages/%40udacity/types-service-worker-mock), but the readme states they are incomplete.

### DIFF
--- a/types/service-worker-mock/index.d.ts
+++ b/types/service-worker-mock/index.d.ts
@@ -4,8 +4,6 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
-export as namespace makeServiceWorkerEnv;
-
 export = makeServiceWorkerEnv;
 declare function makeServiceWorkerEnv(): WorkerGlobalScope;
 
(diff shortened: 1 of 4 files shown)


types/service-worker-mock/index.d.ts
```diff
@@ -0,0 +1,58 @@
+// Type definitions for service-worker-mock 2.0
+// Project: https://github.com/pinterest/service-workers/tree/master/packages/service-worker-mock#readme
+// Definitions by: Remco Haszing <https://github.com/remcohaszing>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+export as namespace makeServiceWorkerEnv;
```
armanio123: I believe the package is not an UMD module. This export is incorrect.

types/service-worker-mock/service-worker-mock-tests.ts
```diff
@@ -0,0 +1,18 @@
+import makeServiceWorkerEnv from 'service-worker-mock';
```
armanio123: nit: For correctness, I prefer to use `import makeServiceWorkerEnv = require('service-worker-mock');` instead and remove the flag `esModuleInterop`. And to clarify, both of the methods are correct. I just prefer this method as I prefer not to enable flags.

armanio123 (changes requested):

armanio123 (approved):

PR #36798, opened 2019-07-10
@postlight/mercury-parser: `html` param can be a `Buffer` as well

moshest:
See here for reference:
https://github.com/postlight/mercury-parser/issues/362


malob (approved):
Looks right to me, thanks!

PR #36799, opened 2019-07-10
[react-select] Export component props.

thchia:
Please fill in this template.

### Motivation

I'm facing an issue best illustrated with a code sample:

```tsx
import Select from 'react-select'

type SelectProps = React.ComponentProps<typeof Select>

const MySelect = () => {
  const props: SelectProps = {
    onChange: (v, actionMeta) => { ... } <----- v and actionMeta are any
  }
  return <Select {...props} />
}
```
Somehow the arguments to onChange are not getting through. This can be fixed by exporting the prop declarations directly and using those instead of the `ComponentProps` util.

### DIFF
--- a/types/react-select/index.d.ts
+++ b/types/react-select/index.d.ts
@@ -3,6 +3,7 @@
 // Definitions by: Claas Ahlrichs <https://github.com/claasahl>
 //                 Jon Freedman <https://github.com/jonfreedman>
 //                 Nathan Bierema <https://github.com/Methuselah96>
+//                 Thomas Chia <https://github.com/thchia>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.9
 
--- a/types/react-select/src/AsyncCreatable.d.ts
+++ b/types/react-select/src/AsyncCreatable.d.ts
@@ -6,7 +6,7 @@ import { cleanValue } from './utils';
 
 export type Props<OptionType> = AsyncProps<OptionType> & CreatableProps<OptionType>;
 
-type State<OptionType> = AsyncState<OptionType> & CreatableState<OptionType>;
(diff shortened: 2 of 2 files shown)


types/react-select/index.d.ts
```diff
@@ -3,13 +3,18 @@
...
 // TypeScript Version: 2.9
 
 import { StateManager } from './src/stateManager';
 
 export default StateManager;
 
+export { Props as SelectProps } from './src/Select';
+export { Props as AsyncProps } from './src/Async';
+export { Props as CreatableProps } from './src/Creatable';
+export { Props as AsyncCreatableProps } from './src/AsyncCreatable';
```
Methuselah96: I'm not in favor of exporting these from `index.d.ts` since the Flow types in the original repo aren't exported like that. Can't you import them via `import { Props } from 'react-select/src/Select`?

Methuselah96 (changes requested):

types/react-select/index.d.ts
thchia: This makes total sense. I'll update this to just export the `Props` from `AsyncCreatable` then.

types/react-select/src/AsyncCreatable.d.ts
```diff
@@ -4,7 +4,7 @@ import { Props as CreatableProps, State as CreatableState } from './Creatable';
 import { OptionsType, ValueType, ActionMeta, InputActionMeta } from './types';
 import { cleanValue } from './utils';
 
-type Props<OptionType> = AsyncProps<OptionType> & CreatableProps<OptionType>;
+export type Props<OptionType> = AsyncProps<OptionType> & CreatableProps<OptionType>;
 
 type State<OptionType> = AsyncState<OptionType> & CreatableState<OptionType>;
```
Methuselah96: Can we also export the `State` type since we're exporting the `Props` type?

Methuselah96 (changes requested):

PR #36803, opened 2019-07-10
Add typings for duckduckgo-images-api

bartduisters:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/duckduckgo-images-api/index.d.ts
+++ b/types/duckduckgo-images-api/index.d.ts
@@ -1,0 +1,32 @@
+// Type definitions for duckduckgo-images-api
+// Project: https://github.com/KshitijMhatre/duckduckgo-images-api
+// Definitions by: Bart Duisters <https://github.com/bartduisters>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export function image_search(params: {
+    query: string, 
+    moderate: boolean, 
+    retries: number, 
+    iterations: number
+}): Promise<Array<{ 
+    image: string,
+    title: string,
+    height: number,
+    thumbnail: string
+    width: number,
+    url: string,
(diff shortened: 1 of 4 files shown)


types/duckduckgo-images-api/index.d.ts
```diff
@@ -0,0 +1,32 @@
+// Type definitions for duckduckgo-images-api 1.0
+// Project: https://github.com/KshitijMhatre/duckduckgo-images-api
+// Definitions by: Bart Duisters <https://github.com/bartduisters>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export function image_search(params: {
+    query: string,
+    moderate: boolean,
```
armanio123: Documentation states that `moderate`, `iterations` and `retries` are all "optional". You could add the `?` to the end of the parameter to make it optional. e.g. `moderate?: boolean`.

types/duckduckgo-images-api/index.d.ts
```diff
@@ -0,0 +1,32 @@
...
+    image: string,
+    title: string,
+    height: number,
+    thumbnail: string,
+    width: number,
+    url: string,
+    source: string}>>;
+
+export function image_search_generator(params: {
+    query: string,
```
armanio123: nit: Both `params` and the returned properties are the same for `image_search` and `image_search_generator`. You could extract that to its own type and reuse them on both functions. It will make it easier to read as well.

armanio123 (changes requested):

types/duckduckgo-images-api/index.d.ts
bartduisters: Extracted to an interface.

armanio123 (approved):

PR #36590, opened 2019-07-02
[@types/webappsec-credential-management] Change RP ID to optional.

lgarron:
See https://www.w3.org/TR/webauthn/#sctn-rp-credential-params

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [ ] Test the change in your own code. (Compile and run.)
- [ ] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.w3.org/TR/webauthn/#sctn-rp-credential-params
- [x] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #36812, opened 2019-07-11
[@types/webappsec-credential-management] Add optional icon to `PublicKeyCredentialEntity`.

lgarron:
See https://w3c.github.io/webauthn/#dictdef-publickeycredentialentity

### DIFF
--- a/types/webappsec-credential-management/index.d.ts
+++ b/types/webappsec-credential-management/index.d.ts
@@ -405,8 +405,8 @@ interface PublicKeyCredentialRequestOptions {
  * @see {@link https://w3c.github.io/webauthn/#dictdef-publickeycredentialentity}
  */
 interface PublicKeyCredentialEntity {
-    name: string;
     icon?: string;
+    name: string;
 }
 
 /**
--- a/types/webappsec-credential-management/index.d.ts
+++ b/types/webappsec-credential-management/index.d.ts
@@ -402,19 +402,25 @@ interface PublicKeyCredentialRequestOptions {
 }
 
 /**
+ * @see {@link https://w3c.github.io/webauthn/#dictdef-publickeycredentialentity}
+ */
+interface PublicKeyCredentialEntity {
+    name: string;
+    icon?: string;
+}
+
+/**
  * @see {@link https://w3c.github.io/webauthn/#dictdef-publickeycredentialrpentity}
  */
-interface PublicKeyCredentialRpEntity {
+interface PublicKeyCredentialRpEntity extends PublicKeyCredentialEntity {
     id?: string;
-    name: string;
 }
 
 /**
  * @see {@link https://w3c.github.io/webauthn/#dictdef-publickeycredentialuserentity}
  */
-interface PublicKeyCredentialUserEntity {
+interface PublicKeyCredentialUserEntity extends PublicKeyCredentialEntity {
     id: BufferSource;
-    name: string;
     displayName: string;
 }
 


Basaingeal (changes requested):
It looks to me as though you reverted the change you made in your last pull request (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36590) making the `id` property of `PublicKeyCredentialRpEntity` required again. The [spec](https://w3c.github.io/webauthn/#dictdef-publickeycredentialrpentity) says it's not required, so I think that should be fixed before merging.

Basaingeal:
I think you're hitting on a piece of major refactoring that needs doing.
There are (in my opinion; I only started maintaining this repo a few weeks ago) implementations defined here that don't really align with the spec. For instance `PublicKeyCredential` extending `CredentialData` instead of `Credential` or `CredentialBase` as this repo has it now. Or the `fetch()` monkey patching that doesn't seem to be part of the spec anymore.

I have a definitions file that aligns with the spec [here](https://github.com/Basaingeal/DefinitelyTyped/blob/feature/w3c-webauthn/types/w3c-webauthn/index.d.ts) but I haven't had the chance to merge it in yet because there are some major changes to the implementation.

Basaingeal:
Actually, it looks like they've updated lib.dom.d.ts in a recent version of `typescript@next` to have most of the types in the WebAuthn spec (not the CredMan spec though) so this definition as is wouldn't pass the tests anymore anyway, and never will unless we can exclude typescript 3.6 from the list of versions to check 'Duplicate identifiers' against.

lgarron:
> Actually, it looks like they've updated lib.dom.d.ts in a recent version of `typescript@next` to have most of the types in the WebAuthn spec (not the CredMan sp
ec though) so this definition as is wouldn't pass the tests anymore anyway, and never will unless we can exclude typescript 3.6 from the list of versions to check 'Duplicate identifiers' against.

Hmm, where are you seeing this?
I can't find it at https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts

Basaingeal:
It was done in this PR https://github.com/microsoft/TypeScript/pull/32335 which updated the [`dom.generated.d.ts`](https://github.com/microsoft/TypeScript/blob/master/src/lib/dom.generated.d.ts) file, which is what drives the `typescript@next` package I think.

The same TS maintainer merged a PR into this repo yesterday https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36837 which was supposed to solve these conflicts by inclining some type definitions.

From my testing, I think everything works now. You might just recommit after rebasing to get the CI tests to run again, and I'll think it'll be good.

lgarron:
> It was done in this PR [microsoft/TypeScript#32335](https://github.com/microsoft/TypeScript/pull/32335) which updated the [`dom.generated.d.ts`](https://github.com/microsoft/TypeScript/blob/master/src/lib/dom.generated.d.ts) file, which is what drives the `typescript@next` package I think.

Hmm, does that make this PR redundant?

lgarron:
Also, I just realized that there's definitely some more outdated stuff in this file, e.g. `SiteBoundCredential` (which seems to have been gone for 2 years?)

Basaingeal:
> Also, I just realized that there's definitely some more outdated stuff in this file, e.g. `SiteBoundCredential` (which seems to have been gone for 2 years?)

Yeah. With most of the WebAuthn stuff being added to TypeScript, I imaging that this repo will 1. be used even less than it is now, and 2. begin to only focus on stuff in the CredMan spec and not WebAuthn (as the name implies anyway), and with that should come some trimming of things removed from the spec.

Basaingeal (approved):

PR #36817, opened 2019-07-11
parse5: more strict typings for getParentNode

yavanosta:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [here](https://github.com/inikulin/parse5/blob/master/packages/parse5/lib/tree-adapters/default.js)

As `DefaultTreeParentNode` was extracted, it's great idea to have `DefaultTreeChildNode`. It makes possible add more strict typings for `getParentNode`.

### DIFF
--- a/types/parse5/index.d.ts
+++ b/types/parse5/index.d.ts
@@ -139,6 +139,16 @@ export interface DefaultTreeParentNode extends DefaultTreeNode {
 }
 
 /**
+ * Default tree adapter ChildNode interface.
+ */
+export interface DefaultTreeChildNode extends DefaultTreeNode {
+    /**
+     * Parent node.
+     */
+    parentNode: DefaultTreeParentNode;
+}
+
+/**
  * Default tree adapter DocumentType interface.
  */
 export interface DefaultTreeDocumentType extends DefaultTreeNode {
@@ -187,7 +197,7 @@ export interface DefaultTreeDocumentFragment extends DefaultTreeParentNode {
 /**
  * Default tree adapter Element interface.
  */
-export interface DefaultTreeElement extends DefaultTreeParentNode {
+export interface DefaultTreeElement extends DefaultTreeChildNode, DefaultTreeParentNode {
     /**
      * The name of the node. Equals to element {@link tagName}.
      */
@@ -205,10 +215,6 @@ export interface DefaultTreeElement extends DefaultTreeParentNode {
      */
     attrs: Attribute[];
     /**
-     * Parent node.
-     */
-    parentNode: DefaultTreeParentNode;
-    /**
      * Element source code location info. Available if location info is enabled via {@link ParserOptions}.
      */
     sourceCodeLocation?: ElementLocation;
@@ -217,7 +223,7 @@ export interface DefaultTreeElement extends DefaultTreeParentNode {
 /**
  * Default tree adapter CommentNode interface.
  */
-export interface DefaultTreeCommentNode extends DefaultTreeNode {
+export interface DefaultTreeCommentNode extends DefaultTreeChildNode {
     /**
      * The name of the node.
      */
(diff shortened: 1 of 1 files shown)


types/parse5/index.d.ts
```diff
@@ -264,6 +262,11 @@ export interface DefaultTreeTextNode extends DefaultTreeNode {
  * Cast to the actual AST interface (e.g. {@link parse5.DefaultTreeNode}) to get access to the properties.
  */
 export type Node = DefaultTreeNode | object;
+/**
+ * Generic ChildNode interface.
+ * Cast to the actual AST interface (e.g. {@link parse5.DefaultTreeChildNode}) to get access to the properties.
+ */
+export type ChildNode = DefaultTreeChildNode | object;
```
andrewbranch: This all looks reasonable given the types already defined here, but nothing is _actually_ stricter because of this, right? Because `DefaultTreeChildNode | object` reduces to just `object` (unless the union is being used to check a fresh object literal, in which case it operates a little differently).

andrewbranch (commented):
Looks fine, but I’m super curious if you could expound a little on what effect this actually has. I’m not familiar with parse5 but I’m familiar with parsers and ASTs in general, and would love to see a motivating example. (In fact, maybe the best place for an example would be the tests—add something that demonstrates the effect of this change and I can approve 👍)

types/parse5/index.d.ts
Yavanosta: Oh @andrewbranch i'm sorry, I had a lot of work to do on my main job, so completely forgot about this PR and now received the notification about it's merge.

The main idea is that:
1. now you can't pass `DefaultTreeDocumentType`  to `getParentNode`
2. now you have a generic type for all nodes that could have parent.

One more thing I have to fix here is type `childNodes`  property in `DefaultTreeParentNode` as `DefaultTreeChildNode` so this give an ability to write node traversal code like `node.childNodes[0].parentNode` (which return same node).

andrewbranch: > now you can't pass DefaultTreeDocumentType to getParentNode

But that’s not true. You can pass literally any object type, because that’s what `object` means.

image

All these `| object` basically make all the other typings meaningless. I merged this because that was already the case for _everything_ in these typings, but if you want to fix them in the future, you need to get rid of those union types.

andrewbranch: And by the way, if you intend to ensure that, e.g., “now you can't pass DefaultTreeDocumentType to getParentNode,” you should add that as a test case with `// $ExpectError`, so you (and reviewers here) don’t have to guess at what the effect of a PR is.

gribnoysup: @andrewbranch thank you so much for this comment, I'm not that familiar with `ts` and I was banging my head on this for a long time.

Do you see any reason why the `| object` is even there? Can it be removed? Not only it creates the case that you described in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36817/files#r305446160, but using these typedefs is completely meaningless as typescript always thinks that anything returned by parser is just `object` with no interface props available 😭 

image

andrewbranch: Sorry, I'm not familiar enough with parse5 to know if this would break anyone or not :/

It definitely seems wrong, but it's possible that it's wrong for a _reason_. This is kind of a weird package because the JS is used by millions of people every day (it’s a dependency of jsdom, notably), but relatively few people are depending on the types. Still, the types package has some popular-ish dependencies, so my advice would be to test whether your changes break some of the more popular ones there: https://www.npmjs.com/browse/depended/@types/parse5

A quick sampling shows these are fairly popular:

- dom5
- polymer-analyzer
- polyserve

gribnoysup: Thanks for some guidance! I'll try looking into it!

PR #36825, opened 2019-07-11
Add basic types def for lodash-deep

uzust01:


armanio123 (changes requested):
Could you please add a test file as well as tslint and tsconfig files? Check the [create a new package](https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package) documentation for details.

lodash-deep/lodash-deep.d.ts
```diff
@@ -0,0 +1,14 @@
+// Type definitions for lodash-deep v 2.0
+
```
DanielRosenwasser: Remove the extra newline

lodash-deep/lodash-deep.d.ts
```diff
@@ -0,0 +1,14 @@
...
+// Project: https://github.com/marklagendijk/lodash-deep
+// Definitions by: Remo uzust01 <https://github.com/uzust01/>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+
+declare module "lodash-deep" {
```
armanio123: This library is an UMD module. Instead you'll need to extend the `lodash` types. You'll find documentation about this on the [library structures page of the TypeScript handbook](http://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#umd) and on the [template module-plugin.d.ts](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-plugin-d-ts.html).

The types should end like this:

```ts
import * as _ from 'lodash';

declare module 'lodash' {
    interface LoDashStatic { // lodash uses this interface for augmentation
        deepMapValues(
            object: any,
            callback: any,
            propertyPath?: any
        ): any;
    }
}

export function deepMapValues(
    object: any,
    callback: any,
    propertyPath?: any
): any;
```

lodash-deep/lodash-deep-tests.ts
```diff
@@ -0,0 +1,6 @@
+import lDeep = require('lodash-deep');
```
armanio123: You'll need to change the test file to use `lodash` instead.
```ts
import _ = require("lodash");
import lDeep = require('lodash-deep');
_.mixin(lDeep);

const object = { null: null };
const callback = () => { };

const value = _.deepMapValues(object, callback, 'local-path');
```

lodash-deep/lodash-deep.d.ts
```diff
@@ -0,0 +1,14 @@
+// Type definitions for lodash-deep v 2.0
```
armanio123: The name of this file should be `index.d.ts`. Also, the `lodash-deep` folder should be inside the `types/` folder.

armanio123 (changes requested):

lodash-deep/lodash-deep.d.ts
uzust01: @armanio123, Thank you for all the help. I am new to this, can you send me some literature I can read so I can become better at it ?

armanio123: You can find most of the documentation on [How can I contribute](https://github.com/DefinitelyTyped/DefinitelyTyped#how-can-i-contribute) section of the DefinitelyTyped readme. 

Also the [TypeScript handbook](https://www.typescriptlang.org/docs/handbook/basic-types.html) is a great source for understanding details about the types.

uzust01: @armanio123, Thank you and could you please re-review ?

armanio123 (approved):

PR #36837, opened 2019-07-11
Dom fixes 2019 07 11

sandersn:
1. `$(window)` will have a different type in TS 3.6, `JQuery<Window & typeof globalThis>`, so `ExpectType JQuery<Window>` assertions won't work anymore. I weakened the test to test assignability:

```ts
const jqw: JQuery<Window> = $(window);
```

2. webappsec-credential-management needs to avoid conflicts with the DOM version that 3.6 will include in order to keep compiling. I copied interfaces and inlined type aliases where needed.

Inlining type aliases is not the ideal solution, but they don't merge, so I couldn't copy them like interfaces. And they are (1) used only once or twice (2) self-explanatory literal unions, so I think it's fine for a package that will see reduced use once 3.6 is released.

I also renamed CredentialBase to Credential and changed it to an interface, also to match the DOM code in 3.6.

3. webgl2 is also now in the DOM, so the types need to match exactly. I deleted an extra property. Again, I expect this package to stop being used as much once 3.6 is released.

### DIFF
--- a/types/jquery/jquery-tests.ts
+++ b/types/jquery/jquery-tests.ts
@@ -190,9 +190,6 @@ function JQueryStatic() {
                 target;
             });
             const myDocForced: JQuery<Document> = $(document);
-            const myWindow = $(window);
-            // $ExpectType JQuery<Window>
-            myWindow;
             const myWindowForced: JQuery<Window> = $(window);
             // $ExpectType JQuery<Window>
             myWindowForced;
--- a/types/jquery/test/jquery-slim-window-module-tests.ts
+++ b/types/jquery/test/jquery-slim-window-module-tests.ts
@@ -1,5 +1,4 @@
 import jq = require('jquery/dist/jquery.slim');
 
 const $window = jq(window);
-// $ExpectType JQuery<Window>
-$window;
+const forced: JQuery<Window> = $window;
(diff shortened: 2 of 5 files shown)


Basaingeal (changes requested):
We were just discussing this problem in a recent pull request https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36812 for the webappsec-credential-management type. I was trying to figure out a way to resolve the type collisions with 3.6 without in-lining all of them, but this might be the only way. Once this is in, we should be able to resolve the CI issues we were having there.

However, since some of the webauthn types that were previously declared are now missing, it could constitute a breaking change in some people's code, so could you bump the version in line 1 up to `0.5`?

Also, (I'm not entirely clear how all this works but) it may be better to submit this as multiple PRs, one for each type, so that individual type maintainers can approve/request changes for the projects that they maintain.

sandersn:
I bumped the version to 0.5. 

I suspect that effort for typing webappsec-credential-management would be better spent on making the built-in DOM types better at microsoft/TSJS-lib-generator, although that might mean some new specs might need to be added there. 

There are not many users of webappsec-credential-management now (as measured by npm downloads) and I expect that number to go down if most people can upgrade Typescript versions. So I think removing some type aliases is not a big problem, considering that they'll be available in TS3.6.

Basaingeal (approved):

PR #36840, opened 2019-07-11
[compression-webpack-plugin]:  Add 'brotliCompress' to the list of supported algorithms

Chris3773:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://webpack.js.org/plugins/compression-webpack-plugin/#using-brotli

### DIFF
--- a/types/compression-webpack-plugin/index.d.ts
+++ b/types/compression-webpack-plugin/index.d.ts
@@ -19,7 +19,7 @@ declare namespace CompressionPlugin {
     type Algorithm<O> = (source: string, options: O, callback: AlgorithmCallback) => void;
 
     // NOTE: These are the async compression algorithms on the zlib object.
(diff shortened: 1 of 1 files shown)


types/compression-webpack-plugin/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for compression-webpack-plugin 2.0
+// Type definitions for compression-webpack-plugin 2.1
```
andrewbranch: From what I can tell, I don’t think a version bump is needed. It sounds like this is passed through to Node, which means it would have worked in 2.0 as well as 2.1+.

andrewbranch (changes requested):

andrewbranch (approved):

PR #36844, opened 2019-07-11
[@types/node] Add missing parameters for process.emitWarning

murgatroid99:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/api/process.html#process_process_emitwarning_warning_options, https://nodejs.org/api/process.html#process_process_emitwarning_warning_type_code_ctor

### DIFF
--- a/types/node/globals.d.ts
+++ b/types/node/globals.d.ts
@@ -842,10 +842,10 @@ declare namespace NodeJS {
     }
 
     interface ProcessWarningOptions {
-        type?: String;
-        code?: String;
+        type?: string;
+        code?: string;
         ctor?: Function;
-        detail?: String;
+        detail?: string;
     }
 
     interface Process extends EventEmitter {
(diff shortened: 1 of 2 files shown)


SimonSchick (changes requested):
Test pls

types/node/globals.d.ts
```diff
@@ -841,6 +841,13 @@ declare namespace NodeJS {
         voluntaryContextSwitches: number;
     }
 
+    interface ProcessWarningOptions {
+        type?: String;
```
SimonSchick: This should be `string` not `String`.

SimonSchick (changes requested):

murgatroid99:
Type fixed and some tests added

KSXGitHub:
From the docs:

> If warning is passed as an `Error` object, it will be passed through to the `'warning'` event handler unmodified (and the optional `type`, `code` and `ctor` arguments will be ignored)

I think you should remove `Error` from `string | Error` and add an overload for it:

```typescript
emitWarning(warning: Error): void;
```

SimonSchick (approved):

murgatroid99:
I disagree with separating out the `Error` override like that. It is still valid in either case to pass the extra arguments if the first argument is an `Error`.

PR #36845, opened 2019-07-11
Add types for `rockset` node module

kwadhwa18:
Please fill in this template.

If adding a new definition:


types/rockset/tslint.json
```diff
@@ -0,0 +1,4 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": { "no-empty-interface": false  }
```
armanio123: Is there a reason why you need this off?

types/rockset/rockset-tests.ts
```diff
@@ -0,0 +1,5 @@
+import { Collection } from "rockset";
+
+function rockset_tests() {
+    // TODO
```
armanio123: Could you please add a couple of tests before merging the pull request?

armanio123 (changes requested):

types/rockset/tslint.json
kwadhwa18: I had few interfaces which are empty and are used in other interfaces. e.g. `AddDocumentsRequestData`. Is there any other way to resolve this?

types/rockset/rockset-tests.ts
kwadhwa18: sure, we had a user waiting for the definitions so created a pull request without tests. Will add some.

types/rockset/tslint.json
armanio123: The linter documentation states: `An empty interface is equivalent to its supertype (or {}).`

So basically it serves no purpose other than saying the type is an `object`. Instead you could just use:
```ts
export interface AddDocumentsResponse {
    data: {}[];
}
```

kwadhwa18: done

types/rockset/rockset-tests.ts
kwadhwa18: Added some tests

kwadhwa18:
The master branch seems to be failing causing Travis to fail for my pull request

armanio123 (approved):

PR #36856, opened 2019-07-12
[stripe-v3] Update Stripe checkout types to include server implementation

harrygr:
There are 2 ways to use Stripe Checkout: the client implementation and
the server implementation. The former requires more parameters, such as
redirect URLs and the products for the checkout session. In the server
implementation the session is created server-side so the only parameter
required for `redirectToCheckout` is the session ID.

Function overloading has been used in favour of a union type for the options parameter as this enforces exclusivity between the two implementations. See https://github.com/Microsoft/TypeScript/issues/14094 (hence the tslint rule being disabled).

Please fill in this template.

---

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://stripe.com/docs/payments/checkout/client

### DIFF
--- a/types/stripe-v3/index.d.ts
+++ b/types/stripe-v3/index.d.ts
@@ -28,7 +28,10 @@ declare namespace stripe {
         createSource(element: elements.Element, options?: { owner?: OwnerInfo }): Promise<SourceResponse>;
         createSource(options: SourceOptions): Promise<SourceResponse>;
         retrieveSource(options: RetrieveSourceOptions): Promise<SourceResponse>;
+        // We use function overloading instead of a union here to ensure that redirectToCheckout can only be
+        // called with either the server options or the client options - not a mix of both.
         redirectToCheckout(options: StripeClientCheckoutOptions): Promise<StripeRedirectResponse>;
+        // tslint:disable-next-line unified-signatures
         redirectToCheckout(options: StripeServerCheckoutOptions): Promise<StripeRedirectResponse>;
         paymentRequest(options: paymentRequest.StripePaymentRequestOptions): paymentRequest.StripePaymentRequest;
         createPaymentMethod(
--- a/types/stripe-v3/index.d.ts
+++ b/types/stripe-v3/index.d.ts
@@ -28,7 +28,8 @@ declare namespace stripe {
         createSource(element: elements.Element, options?: { owner?: OwnerInfo }): Promise<SourceResponse>;
         createSource(options: SourceOptions): Promise<SourceResponse>;
         retrieveSource(options: RetrieveSourceOptions): Promise<SourceResponse>;
-        redirectToCheckout(options: StripeCheckoutOptions): Promise<StripeRedirectResponse>;
+        redirectToCheckout(options: StripeClientCheckoutOptions): Promise<StripeRedirectResponse>;
+        redirectToCheckout(options: StripeServerCheckoutOptions): Promise<StripeRedirectResponse>;
         paymentRequest(options: paymentRequest.StripePaymentRequestOptions): paymentRequest.StripePaymentRequest;
         createPaymentMethod(
             type: paymentMethod.paymentMethodType,
@@ -75,15 +76,19 @@ declare namespace stripe {
     };
 
     type billingAddressCollectionType = 'required' | 'auto' | '';
-    interface StripeCheckoutOptions {
+
+    interface StripeClientCheckoutOptions {
         items: StripeCheckoutItem[];
         successUrl: string;
         cancelUrl: string;
         clientReferenceId?: string;
         customerEmail?: string;
         billingAddressCollection?: billingAddressCollectionType;
-        sessionId?: string;
         locale?: string;
+    }
+
+    interface StripeServerCheckoutOptions {
+        sessionId: string;
     }
 
     interface StripeCheckoutItem {
--- a/types/stripe-v3/stripe-v3-tests.ts
+++ b/types/stripe-v3/stripe-v3-tests.ts
@@ -196,13 +196,21 @@ describe("Stripe elements", () => {
         });
     });
 
-    it("should use checkout API", () => {
+    it("should use checkout API for client implementations", () => {
         stripe.redirectToCheckout({
             items: [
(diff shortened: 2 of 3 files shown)


amritk (approved):
Looks good to me

andrewbranch (changes requested):
> Function overloading has been used in favour of a union type for the options parameter as this enforces exclusivity between the two implementations.

That’s the case when overload signatures differ in more than just one parameter type. E.g., if each overload had a different return type, they wouldn’t be unifiable. Or, if each overload had a second parameter that also differed in type, the exclusivity would be important—a union would be too permissive. But in the code here, the overload signatures differ only in a single parameter type position, which means there’s effectively not a difference between doing that and using a union type, which is what the linter is complaining about. If each overload even had separate JSDoc comments, that would be a reasonable justification for using overloads, but neither of these signatures has documentation attached. Maybe I’m missing something, but it doesn’t seem like the overload is particularly helpful here. Can you elaborate on what the benefit is?

harrygr:
>  Can you elaborate on what the benefit is?

Sure. So as far as I understand, you'd never call `redirectToCheckout` with options containing both a `sessionId` and the stuff for using it purely client-side (like `items`, `successUrl` etc).

If a union type is used for `options` then TypeScript allows this, due to the structural typing approach of the compiler. 

One way to get around this is a Xor operator as discussed in the linked issue. One could also define a `Xor` type constructor instead.

If we use function overloading then we're saying that you either have to call the function with the server options or the client options, but you can't mix the two.

e.g.

```ts
declare function redirectToCheckout(options: StripeClientCheckoutOptions): Promise<any>;
declare function redirectToCheckout(options: StripeServerCheckoutOptions): Promise<any>;

redirectToCheckout({
    sessionId: 'sess_test',
    successUrl: 'http://example.com/success'
})
```

results in:
> Argument of type '{ sessionId: string; successUrl: string; }' is not assignable to parameter of type 'StripeServerCheckoutOptions'.
  Object literal may only specify known properties, and 'successUrl' does not exist in type 'StripeServerCheckoutOptions'.

which is what we want.

whilst

```ts
declare function redirectToCheckout2(options: StripeClientCheckoutOptions | StripeServerCheckoutOptions): Promise<any>;

redirectToCheckout2({
    sessionId: 'sess_test',
    successUrl: 'http://example.com/success'
})
```

has no error.

[link to playgound](http://www.typescriptlang.org/play/#code/JYOwLgpgTgZghgYwgAgMpisADhAwgG2AnFwAsIEBrAewFcwB5LMYakAZ2QG8AoZf5MEgBbdgC5kcEAE8A2gF0A3HwHtaCJO3YBVKPgnsMoAObKByBFKT5d+5IcwhTK-gkLEwAJQgxoxJACSACYA-AZGTmYCCLSG1MLQAKLCcMD4YfYRzub41Jb4EBkOJsoAvjw8oJCwiCjomDio0ABu0GQUNPRMLGycvObsEFqsIMHhjs7lPEEU+HBQKDC0IAg9IMgLQcALqwAq1O1UdGAAFNTMI+JoRjgERCTkR10XvQCUEgAKUPHAgwA8UmkAD5lDM3PNFstViMNhAtjswPtDp1Tuc1ld6tgIE0oK0oMjjt1Lu9kF8fv9ASCKptthREQdHiiTv1VEN2CMxsgAOSDLQAfUghi5ABoXPZ1JodHoJFzSGAwFgxAB6JUQAAecGEWAKADoEPElWoNGyuTxSq8KmC5gtkEsVmtYfC6UjGccAExnF4cCSY27uB4dQlezgAH2uDWxLTarue6JJZOEvwgAJkVJ4NIRLsD9A9LP4vPZbE5PLZAqGYBFYqEEFEEgUZteQA)

Maybe I've misinterpreted the API, in which case I'm happy to amend my PR, but as it stands the types prevent using `redirectToCheckout` in the server implementation without a type assertion to an `any` for the options.

andrewbranch:
@harrygr got it, thanks. FWIW, this only applies to fresh object literal checking, as object types with extra properties are still strictly subtypes of ones with a subset of those properties, so there’s no _true_ way to prohibit extra properties from making it in: [playground](http://www.typescriptlang.org/play/#code/JYOwLgpgTgZghgYwgAgMpisADhAwgG2AnFwAsIEBrAewFcwB5LMYakAZ2QG8AoZf5MEgBbdgC5kcEAE8A2gF0A3HwHtaCJO3YBVKPgnsMoAObKByBFKT5d+5IcwhTK-gkLEwAJQgxoxJACSACYA-AZGTmYCCLSG1MLQAKLCcMD4YfYRzub41Jb4EBkOJsoAvjw8oJCwiCjomDio0ABu0GQUNPRMLGycvObsEFqsIMHhjs7lPEEU+HBQKDC0IAg9IMgLQcALqwAq1O1UdGAAFNTMI+JoRjgERCTkR10XvQCUEgAKUPHAgwA8UmkAD5lDM3PNFstViMNhAtjswPtDp1Tuc1ld6tgIE0oK0oMjjt1Lu9kF8fv9ASCKghemBkGjLsgALzcFz2IbsEZjZAAckGWgA+pBDDyADRstQaDm2CQ80hgMBYMQAemVEAAHnBhFgCgA6GnCZWSzTsMVswjVOD4fDSCQABnF5kBYFIJgkAEZHQIIPhBhIAExe-jSOgSADMQeQAHcpGAJAAWHilZQ8TbbCiIg6PFFnF4cV48IA)

But, for the typical case, I buy that the overload gives a better UX. My only request is that you use `// tslint:disable-next-line` instead of disabling the rule outright, and maybe leave a comment with a brief justification so nobody hassles you about it again 🙂

andrewbranch:
> there’s no true way to prohibit extra properties from making it in

Well, as you pointed out, you could prohibit _specific_ extra properties from making it in by declaring them as `undefined` or `never` or something, a la that horrifying/beautiful/not-sure-how-to-feel-about-it XOR mapped type. 😄  But you can’t guard against _arbitrary_ extra properties.

harrygr:
Hey @andrewbranch thanks for your feedback and apologies for the delay.

I've amended my PR as requested.

andrewbranch (approved):
Thanks!

PR #36862, opened 2019-07-12
[highcharts] Allow null y values in datapoints

dyladan:
https://api.highcharts.com/highcharts/series.line.data

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://api.highcharts.com/highcharts/series.line.data

### DIFF
--- a/types/highcharts/index.d.ts
+++ b/types/highcharts/index.d.ts
@@ -5910,7 +5910,7 @@ declare namespace Highcharts {
          *            [3, 7, 8]
          *
          */
-        data?: Array<number | [number, number] | [string, number] | [string, number, number] | [number, number, number] | DataPoint>;
+        data?: Array<number | null | [number, number] | [number, null] | [string, number] | [string, number, number] | [number, number, number] | DataPoint | null>;
         /**
          * A description of the series to add to the screen reader information about the series.
          * @since 5.0.0
@@ -6134,7 +6134,7 @@ declare namespace Highcharts {
         /**
          * The y value of the point.
          */
-        y?: number;
+        y?: number | null;
         /**
          * The size value for each bubble. The bubbles' diameters are computed based on the z, and controlled by series
          * options like minSize, maxSize, sizeBy, zMin and zMax.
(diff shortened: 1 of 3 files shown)


types/highcharts/index.d.ts
```diff
@@ -5910,7 +5910,7 @@ declare namespace Highcharts {
          *            [3, 7, 8]
          *
          */
-        data?: Array<number | [number, number] | [string, number] | [string, number, number] | [number, number, number] | DataPoint>;
+        data?: Array<number | null | [number, number] | [number, null] | [string, number] | [string, number, number] | [number, number, number] | DataPoint | null>;
```
dolanmiu: It doesn't say anywhere in the documentation that data can be null. Please revert.

types/highcharts/index.d.ts
```diff
@@ -6134,7 +6134,7 @@ declare namespace Highcharts {
         /**
          * The y value of the point.
          */
-        y?: number;
+        y?: number | null;
```
dolanmiu: Correct, y can be null:

https://github.com/highcharts/highcharts/blob/8f802d8c7/js/parts/Series.js#L5730-#L5736

dolanmiu (changes requested):
Revert the `data?:` change

`y` change seems ok

types/highcharts/index.d.ts
dyladan: From the documentation:

> An array of numerical values. In this case, the numerical values will be interpreted as `y` options

> An array of arrays with 2 values. In this case, the values correspond to `x,y`.

In addition the type annotation at https://github.com/highcharts/highcharts/blob/8f802d8c7/js/parts/Series.js#L5613 seems to allow it

> `* @type      {Array<number|Array<(number|string),(number|null)>|null|*>}`

dyladan: Examples:

Data with number and null: https://jsfiddle.net/6vptua08/
Data with [number,number] and [number,null] pairs: https://jsfiddle.net/gzya3e7c/
Data with [string, number] and [string, null] (x value is inferred, in this case 1 per point): https://jsfiddle.net/gzya3e7c/2/

dolanmiu:
My bad, I think you are right

dolanmiu (approved):

dolanmiu:
Approved

PR #36868, opened 2019-07-13
Node-Fetch: Generic for "body".json() method

favna:
Please fill in this template.

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: **N.A. This is coming from a frustration I had in my own code where I would have to cast my data in some way or another when it's much cleaner when the function would just support generics.**

---

I've had to add a `// tslint:disable-next-line:no-unnecessary-generics` to add this to avoid [this linting rule](https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md). I opted for this because the suggested adjustments from the dtslint documentation are not applicable to this specific case and I believe that adding type information in the way this PR does is worth having a single line disable a single dtslint rule.

The key benefit for this can be seen in the added test cases where type information is *not* available in the code where `.json()` resolves to `Promise<any>`, but  *does* have type information when passing in the generic.

---

Example using actual code of before and after.

**Before**:
```ts
(async function() {
    try {
      // Fetch Reddit comments for a user
      const response = await fetch(`https://www.reddit.com/user/favna/comments.json?after=''&limit=100`);
      const json: RedditResponse<'comments'> = await response.json();
      // or
      const json2 = await response.json() as RedditResponse<'comments'>;
      //  Do something with the JSON
    } catch (err) {
        msg.reply(err);
    }
})();
```

**After**:
```ts
(async function() {
    try {
      // Fetch Reddit comments for a user
      const response = await fetch(`https://www.reddit.com/user/favna/comments.json?after=''&limit=100`);
      const json = await response.json<RedditResponse<'comments'>>();
      //  Do something with the JSON
    } catch (err) {
        msg.reply(err);
    }
})();
```

---

p.s. I also tried to pull the generic all the way up to the `fetch` function and while in terms of typing I managed, it made many tests crash and basically requiring a higher typescript version, which in turn caused a lot of dependency issues as other packages depend on node-fetch. Figured that was not the best way going forward, but maybe it'll be something to keep in mind if the node-fetch library itself does a major version upgrade.

### DIFF
--- a/types/node-fetch/index.d.ts
+++ b/types/node-fetch/index.d.ts
@@ -141,7 +141,8 @@ export class Body {
     body: NodeJS.ReadableStream;
     bodyUsed: boolean;
     buffer(): Promise<Buffer>;
-    json(): Promise<any>;
+    // tslint:disable-next-line:no-unnecessary-generics
+    json<T extends any>(): Promise<T>;
     size: number;
     text(): Promise<string>;
     textConverted(): Promise<string>;
--- a/types/node-fetch/index.d.ts
+++ b/types/node-fetch/index.d.ts
@@ -8,6 +8,7 @@
 //                 Jason Li <https://github.com/JasonLi914>
 //                 Brandon Wilson <https://github.com/wilsonianb>
 //                 Steve Faulkner <https://github.com/southpolesteve>
+//                 Jeroen Claassens <https://github.com/favna>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 /// <reference types="node" />
--- a/types/node-fetch/node-fetch-tests.ts
+++ b/types/node-fetch/node-fetch-tests.ts
@@ -144,3 +144,21 @@ function test_ResponseInit() {
         });
     });
 }
+
+interface test_JsonPlaceholderBody {
+    userId: number;
+    id: number;
+    title: string;
+    completed: boolean;
+}
+
+function test_ResponseJsonBody() {
+    fetch('https://jsonplaceholder.typicode.com/todos/1', {}).then(response => {
+        response.json().then(dataWithoutGeneric => {
+            // no type information available => Attempting to access 'id' throws "Property 'id' does not exist on type '{}'" in tests
+        });
+        response.json<test_JsonPlaceholderBody>().then(dataWithGeneric => {
+            const id = dataWithGeneric.id; // type information available
+        });
+    });
+}


kyranet:
I think disabling that TSLint rule already shows how a bad idea this can be, plus I don't really see the point on the changes, is there an issue with casting? `no-unnecessary-generics` forbids this code style:

```typescript
function parse<T>(): T;
const x = parse<number>();
```

As it could be written with less syntax doing a cast:

```typescript
function parse(): {};
const x = parse() as number;
```

Furthermore, this also applies to awaited promises, your code example includes this line:

```typescript
const json = await response.json<RedditResponse<'comments'>>();
```

Which can be rewritten to this, and wouldn't need generics nor extra syntax:

```typescript
const json = await response.json() as RedditResponse<'comments'>;
```

Or if you use ES6:

```typescript
response.json()
  .then((json: RedditResponse<'comments'>) => { /* code */ });
```

Favna:
@Kyranet yes you're not wrong in that you could do `json() as <yourType>` or `const json: <yourType> = await json()`, however personally I'm off the opinion that type casting and `any`/`unknown` should be avoided when possible and instead either use generics and type inferring wherever possible. This because using type casting is essentially sidestepping the protection the type system gives us which is very core reason why we're even writing TypeScript and not plain JavaScript. If I'm going to cast everything I might as well just make everything `any` and then realize I might as well just write JavaScript.

My personal view on this is strongly influenced by how typecasting is frowned upon by Java, C# and similar language communities. For example in [this stackoverflow](https://stackoverflow.com/questions/4167304/why-should-casting-be-avoided) user Jerry Coffin rightfully says:

> Looking at things more generally, the situation's pretty simple (at least IMO): a cast (obviously enough) means you're converting something from one type to another. When/if you do that, it raises the question "Why?" If you really want something to be a particular type, why didn't you define it to be that type to start with? That's not to say there's never a reason to do such a conversion, but anytime it happens, it should prompt the question of whether you could re-design the code so the correct type was used throughout

kyranet:
The rule is to use `any` for literally any JavaScript value (can be used without casting), and `unknown` (much newer, must be casted).

Generic parameters do not help you with inferred types in this context, as it's basically impossible to do, you need to pass the value, which casts the value.

> My personal view on this is strongly influenced by how typecasting is frowned upon by Java, C# and similar language communities.

`any` and `unknown` exist because TypeScript is a superset of an untyped language where you can pass any arbitrary value of any type into a function, you could even do `add({}, function() {});` in a function with the TS signature of `add(a: number, b: number);`, though TS will forbid this on compile time.

Casting in `any` and `unknown` are completely fine.

Favna:
Well I think we can agree to disagree then. We have different opinions on how these kind of type situations should be handled. That said however since I'm not swayed I'm going to keep this PR open and see where it goes from here by means of other reviewers.

southpolesteve:
I'm inclined not to accept. I see both sides of the argument, but this would diverge from the TS lib.dom definitions (https://github.com/microsoft/TSJS-lib-generator/blob/master/baselines/dom.generated.d.ts#L2516) and I would prefer not to do that. My rec would be to try to convince them to make the change.

Favna:
I can't really say anything against what @southpolesteve brought up as argument. Using TS' `lib.dom` as a "source of truth" for the typings is pretty solid imo. 

That said, I've given what @southpolesteve said a shot and created a PR: https://github.com/microsoft/TSJS-lib-generator/pull/739

For the record then , @southpolesteve and @kyranet would you guys approve this PR in the instance that the Microsoft devs approve my PR to TSJS-lib-generator repo?  Not sure how much time approval or denial will take, but I'll ensure this PR stays open one way or another at least until the other PR is handled.

southpolesteve:
Yes. If the change lands there, I would want it here too

Favna:
Going to tag @andrewbranch and @sandersn here because of the comment from @typescript-bot. To TL;DR over the conversation in bulletpoints:

- We agreed that we'd stick to what is defined in DOM definitions in TypeScript
- I created a PR to TSJS-lib-generator, linked above
- That PR is as of yet without response and I personally have no idea how long that will take
- Ideally this PR stays open until some closing action is taken on the definitions to TSJS-lib-generator.
  - If the PR there is accepted then this PR can be accepted
  - If the PR there is denied then this PR can be denied

andrewbranch:
> however personally I'm off the opinion that type casting and any/unknown should be avoided when possible and instead design the code in such a way that casting is not required

This _is_ a typecast—there’s no difference aside from syntax. I do think that the generic looks and feels better, but that actually makes it worse. but that’s entirely an illusion, and the illusion of type safety is worse than something that is transparently unsafe. Especially because those coming from a C# background who are unfamiliar with TypeScript would have good reason to believe that specifying a class as the explicit type parameter would ensure that the result is an instance of that class—C# would have no trouble implementing runtime reflection to that effect, which of course isn’t possible with this syntax in TypeScript. The generic truly is useless, and I think that can lead to confusion, which is exactly why the lint rule exists.

There _may_ come a time in the future where the DOM libs could use `unknown` instead of `any`, possibly via a compiler flag or something (don’t take this as a promise; we don’t have any specific plans for this right now), in which case you would be forced to _safely_ narrow the type before using it:

```ts
function isMyResource(json: unknown): json is MyResource {
  // check the shape... return true/false
}

// ...
const json = await response.json(); // unknown
if (isMyResource(json)) {
  json; // MyResource
}
```

This would be the canonical, safe way to type such a construct, at which point the generic would be egregiously wrong by comparison. (What you get from an API call is inherently unknown to TypeScript, and there’s no real way around that.) For now, `any` is often used where `unknown` should, and it’s not great, but I think a guiding principle in the meantime could be that we don’t want to write types that are still unsafe in a world where we simply replace `any` with `unknown`:

```ts
// Unsafe
interface Response {
  json(): Promise<any>;
}

// find-and-replace any -> unknown makes it safe:
interface Response {
  json(): Promise<unknown>;
}
```

```ts
interface Response {
  json<T extends any>(): Promise<T>;
}

// find-and-replace any -> unknown has no effect:
interface Response {
  json<T extends unknown>(): Promise<T>;
}
```

I do appreciate your time and effort here, so I’m sorry that it didn’t achieve the results you were hoping for. Thanks for starting a meaningful discussion; keep the contributions coming 🙂

southpolesteve:
> There may come a time in the future where the DOM libs could use unknown instead of any, possibly via a compiler flag or something

Yes, please!

andrewbranch:
@southpolesteve you can always sub in your own custom, safer lib files, but I know this is about as helpful as telling a lactose intolerant person at an ice cream party that they’re welcome to bake themselves a cake

Favna:
Thank you for the great explanation Andrew

PR #35920, opened 2019-06-03
1.5.0.Leaflet.fix type for html property in DivIconOptions

ivkusto:
According https://leafletjs.com/reference-1.5.0.html#divicon-html we can use HTMLElement for html property. I pick 'Element' type instead of 'HTMLElement', because also we can use SVGElement
<<https://leafletjs.com/reference-1.5.0.html#divicon-html >>

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<url here>>
- [x] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #36883, opened 2019-07-14
fix type for html property in DivIconOptions

ivkusto:
According https://leafletjs.com/reference-1.5.0.html#divicon-html we can use HTMLElement for html property. Update because of this comment https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35920#pullrequestreview-247360014
<<https://leafletjs.com/reference-1.5.0.html#divicon-html >>

Please fill in this template.

### DIFF
--- a/types/leaflet/index.d.ts
+++ b/types/leaflet/index.d.ts
@@ -1511,7 +1511,7 @@ export namespace Icon {
 export function icon(options: IconOptions): Icon;
 
 export interface DivIconOptions extends BaseIconOptions {
-    html?: string | false;
+    html?: string | false | HTMLElement | SVGElement;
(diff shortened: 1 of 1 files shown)


life777 (approved):
Everything is ok.

According to Leaflet documentation DivIconOptions property HTML may be only String|HTMLElement [see here](https://leafletjs.com/reference-1.5.0.html#divicon-html).

But according to code they accept all objects that are `instanceof Element` [see here](https://github.com/Leaflet/Leaflet/blob/master/src/layer/marker/DivIcon.js#L49)

Thanks a lot

ivkusto (commented):
ok

PR #36884, opened 2019-07-14
[@types/mongoose] Check static property types more strongly

lightnet328:
# Changes

## Base code

```typescript
import { Document, DocumentQuery, Model, Schema } from "mongoose";

interface AnimalDocument extends Document {
  name: string;
}

interface AnimalModel extends Model<AnimalDocument> {
  findByName(name: any, cb: Function): DocumentQuery<AnimalDocument[], AnimalDocument, {}>;
}

const animalSchema = new Schema<AnimalDocument, AnimalModel>({
  name: String
});
```

## Before

```typescript
animalSchema.statics.findByName = function(this: AnimalModel, name: any, cb: string) { // allowed :(
  return this.find({ name: new RegExp(name, "i") }, cb);
};
```

## After

```typescript
animalSchema.statics.findByName = function(this: AnimalModel, name: any, cb: string) { //  error :)
  return this.find({ name: new RegExp(name, "i") }, cb);
};
```


andrewbranch (commented):
I thiiiink this looks ok, but would really like feeback from an owner:  @horiuchi @lukasz-zak @Alorel @jendrikw @ethanresnick @vologab @jussikinnula @ondratra @alfirin @idandrd @various89 @Fazendaaa @NormanPerrin @danmana @stablio @emmanuelgautier @frontendmonster @mingchen @penumbra1 @orblazer @HughKu @niuware @Vladmel1234 @jloveridge @grimmer0125, pinging you again, thanks 🔔 

I would maybe recommend updating the version string with a change like this—mongoose itself is on 5.6, so you’re ok to bump the types versions up to 5.6, and that will make this change a little less disruptive (if anything about it becomes disruptive—I think it’s probably fine, since you default the type parameter to `any`, and I don’t think it can be inferred).

If you update the header version, please also go ahead and remove the lint rule exception for `dt-header` from tslint.json so the header can be validated.

Issue #36857, opened 2019-07-12
Chart.js - Property 'options' does not exist on type 'Chart'


ZaLiTHkA:
- [x] I tried using the `@types/chart.js@2.7.55` package and had problems.
- [ ] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @anuti @FabienLavocat @KentarouTakeda @larrybahr @mernen @josefpaij @danmana @guillaume-ro-fr @archy-bold @braincore @frabnt @alexdor @mahnuh @Conrad777 @adripanico @wertzui @lekoaf

If you do not mention the authors the issue will be ignored.

>note: I didn't tick the "latest tsc" checkbox because my environment is running TS `3.5.2`, but `3.5.3` won't make any difference here.

The `onRefresh` callback from `chartjs-plugin-streaming` is called with the chart instance as it's only argument, so I've assigned the `Chart` type to the argument in the function I'm assigning to that key.

Inside this function, when I try access `chart.options` as per the [official documentation](https://www.chartjs.org/docs/latest/developers/updates.html#updating-options), I get the error `Property 'options' does not exist on type 'Chart'`.

---

If I change the type assignment to `ChartConfiguration`, then I get no errors accessing either `chart.data` or `chart.options`, but the naming convention there seems strange..

Should I be using `ChartConfiguration` here, even though I'm working with a reference to the *rendered* chart?

I know the constructor for elements in the `chart.data.datasets` array does change between the configuration I give it and the instance after the chart has rendered, so do I run a risk of mismatching the type definitions with the actual object here by using `ChartConfiguration`?

adripanico:
I don't think this question belongs to here.

If you get the error accessing to chart.options with tsc, try to access it in runtime. If the property actually exists, then change the type definitions and make a PR to this repo. If the property does not exists, open an issue into the proper repo.

ZaLiTHkA:
@adripanico, my point here is that the property does exist on the root `Chart` instance attached to the DOM at runtime.

[...]

PR #36893, opened 2019-07-15
fix(chart.js): add missing Chart instance fields

makepanic:
Please fill in this template.

* https://github.com/chartjs/Chart.js/blob/master/src/core/core.controller.js#L180-L183

The `Chart` instance has some fields which weren't specified yet.
I've used `null` as the unset option because it's the same unset type that `canvas` uses which is used to decide whether to set them or not.


makepanic:
Seems like CI fails on an unrelated issue `Error: In package.json: Dependency form-data not in whitelist.`

danmana:
Closes #36857
Closes #28248

PR #36896, opened 2019-07-15
webpack: fix FmModule.reasons typing

yavanosta:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://webpack.js.org/api/stats/#module-objects

### DIFF
--- a/types/webpack/index.d.ts
+++ b/types/webpack/index.d.ts
@@ -1297,7 +1297,7 @@ declare namespace webpack {
             prefetched: boolean;
             profile: any; // TODO
             providedExports?: any; // TODO
-            reasons: {
+            reasons: [{
                 explanation: string | undefined;
                 loc?: string;
                 module: string | null;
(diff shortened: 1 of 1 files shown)


types/webpack/index.d.ts
```diff
@@ -1297,7 +1297,7 @@ declare namespace webpack {
             prefetched: boolean;
             profile: any; // TODO
             providedExports?: any; // TODO
-            reasons: {
+            reasons: [{
```
mohsen1: if it can be an object like this or an array of objects like this then you need to put this type into a type reference or interface and then use it here. Something like:

```ts
type Reason = {
    explanation: string | undefined;
    // ...
};

```
```suggestion
            reasons: Reason[] | Reason
```

mohsen1 (changes requested):
You're mixing the tuple syntax with array syntax.

types/webpack/index.d.ts
Yavanosta: Seems it can't be an object and always is an array. At least on [webpack](https://github.com/webpack/webpack/blob/master/lib/Stats.js#L535) current version. 

Extract `Reason` type is a great idea.

PR #36899, opened 2019-07-15
Lodash: `overEvery`: add support for user defined type guards

OliverJAsh:
---

Please fill in this template.

### DIFF
--- a/types/lodash/ts3.1/common/util.d.ts
+++ b/types/lodash/ts3.1/common/util.d.ts
@@ -744,6 +744,8 @@ declare module "../index" {
         over<TResult>(...iteratees: Array<Many<(...args: any[]) => TResult>>): FunctionChain<(...args: any[]) => Array<ReturnType<T> | TResult>>;
     }
 
+    type Refinement<A, B extends A> = (a: A) => a is B;
+
     interface LoDashStatic {
         /**
          * Creates a function that checks if all of the predicates return truthy when invoked with the arguments
@@ -752,6 +754,10 @@ declare module "../index" {
          * @param predicates The predicates to check.
          * @return Returns the new function.
          */
+        overEvery<A, B extends A, C extends B>(
+            fn1: Refinement<A, B>,
+            fn2: Refinement<B, C>,
+        ): Refinement<A, C>;
         overEvery<T>(...predicates: Array<Many<(...args: T[]) => boolean>>): (...args: T[]) => boolean;
     }
     interface Collection<T> {
--- a/types/lodash/ts3.1/lodash-tests.ts
+++ b/types/lodash/ts3.1/lodash-tests.ts
@@ -6843,6 +6843,13 @@ fp.now(); // $ExpectType number
 // _.overEvery
 // _.overSome
 {
+    interface Named { name: string; }
+    interface Aged { age: number; }
+    const userDefinedTypeGuardNamed: <T>(item: T) => item is T & Named = anything;
+    const userDefinedTypeGuardAged: <T>(item: T) => item is T & Aged = anything;
+
+    _.overEvery(userDefinedTypeGuardNamed, userDefinedTypeGuardAged); // $ExpectType Refinement<{}, Named & Aged>
+
     _.overEvery((number: number) => true); // $ExpectType (...args: number[]) => boolean
     _.overEvery((number: number) => true, (number: number) => true); // $ExpectType (...args: number[]) => boolean
     _.overEvery([(number: number) => true]); // $ExpectType (...args: number[]) => boolean


types/lodash/ts3.1/common/util.d.ts
```diff
@@ -752,6 +754,10 @@ declare module "../index" {
          * @param predicates The predicates to check.
          * @return Returns the new function.
          */
+        overEvery<A, B extends A, C extends B>(
+            fn1: Refinement<A, B>,
+            fn2: Refinement<B, C>,
+        ): Refinement<A, C>;
         overEvery<T>(...predicates: Array<Many<(...args: T[]) => boolean>>): (...args: T[]) => boolean;
```
OliverJAsh: If we're happy with this, I'll go ahead and add the same overload for the other variants.

I can also extend support for the number of allowed functions, if need be.

aj-r: Looks good to me, go ahead and d othe other variants.
Let's stick with just 2 arguments for now, unless a need arises for more arguments.

types/lodash/ts3.1/lodash-tests.ts
```diff
@@ -6843,6 +6843,13 @@ fp.now(); // $ExpectType number
 // _.overEvery
 // _.overSome
 {
+    interface Named { name: string; }
+    interface Aged { age: number; }
+    const userDefinedTypeGuardNamed: <T>(item: T) => item is T & Named = anything;
+    const userDefinedTypeGuardAged: <T>(item: T) => item is T & Aged = anything;
+
+    _.overEvery(userDefinedTypeGuardNamed, userDefinedTypeGuardAged); // $ExpectType Refinement<{}, Named & Aged>
```
aj-r:
```
ERROR: 6851:5  expect  TypeScript@next expected type to be:
  Refinement<{}, Named & Aged>
got:
  <T>(a: T) => a is T & Named & Aged
```

aj-r (changes requested):

types/lodash/ts3.1/lodash-tests.ts
OliverJAsh: Odd, I don't get that error. I do get this error however:

```ts
The 'use-default-type-parameter' rule threw an error in '/Users/oliverjash/Development/DefinitelyTyped/types/lodash/lodash-tests.ts':
TypeError: Cannot read property 'default' of undefined
    at checkArgsAndParameters (/Users/oliverjash/Development/DefinitelyTyped/node_modules/tslint/lib/rules/useDefaultTypeParameterRule.js:63:19)
    at cb (/Users/oliverjash/Development/DefinitelyTyped/node_modules/tslint/lib/rules/useDefaultTypeParameterRule.js:52:13)
    at visitNode (/Users/oliverjash/Development/DefinitelyTyped/node_modules/typescript/lib/typescript.js:16701:24)
    at Object.forEachChild (/Users/oliverjash/Development/DefinitelyTyped/node_modules/typescript/lib/typescript.js:16947:24)
    at cb (/Users/oliverjash/Development/DefinitelyTyped/node_modules/tslint/lib/rules/useDefaultTypeParameterRule.js:54:19)
    at visitNodes (/Users/oliverjash/Development/DefinitelyTyped/node_modules/typescript/lib/typescript.js:16710:30)
    at Object.forEachChild (/Users/oliverjash/Development/DefinitelyTyped/node_modules/typescript/lib/typescript.js:16936:24)
    at cb (/Users/oliverjash/Development/DefinitelyTyped/node_modules/tslint/lib/rules/useDefaultTypeParameterRule.js:54:19)
    at visitNodes (/Users/oliverjash/Development/DefinitelyTyped/node_modules/typescript/lib/typescript.js:16710:30)
    at Object.forEachChild (/Users/oliverjash/Development/DefinitelyTyped/node_modules/typescript/lib/typescript.js:16936:24)
```

types/lodash/ts3.1/common/util.d.ts
OliverJAsh: I just tried to add the other variants, but realised it's not possible.

IIUC, the chain variants must accept only one parameter. I could add an overload for a tuple?

When I ran the FP generate script, no changes were made 🤷‍♂

aj-r: That's fair. The FP generate probably didn't do anything because it's expecting a 1-argument overload. Adding an overload for tuple would probably fix FP. Maybe to keep things simple we should *only* support the tuple version? Just to prevent things from getting out of hand if we ever add overloads for 3+ arguments.

Is it even possible to do tuples for the chained version? I think in the 3.1 version, you'll just end up with `CollectionChain<T>` which you can't really make into a tuple. If so, feel free to ignore the chains.

aj-r (changes requested):

types/lodash/ts3.1/common/util.d.ts
OliverJAsh: > Maybe to keep things simple we should _only_ support the tuple version?

I agree. However, as soon as I switched to the tuple version, my test failed with:

```
ERROR: 6851:5  expect  TypeScript@next expected type to be:
  Refinement<{}, Named & Aged>
got:
  Refinement<{}, Aged>
```

😕 

I have no idea why that would be happening. In VS Code, with the TS selected to use the workspace version, the return type looks fine

image

OliverJAsh: > Adding an overload for tuple would probably fix FP. 

For some reason, `npm run generate` still produced no changes after I had switched to use a tuple instead.

OliverJAsh: Hi @aj-r, any ideas on the above? I will re-open a new PR once we've decided how to proceed.

aj-r: Sorry, I'm not sure what's up with that - I don't see why tuples would behave differently (but I'm also not an expert on the TS compiler).

You could try returning `Refinement<A, B & C>` - I feel like that should fix it (and is probably more reliable anyway).

```ts
overEvery<A, B extends A, C extends A>(fns: [
    Refinement<A, B>,
    Refinement<B, C>,
]): Refinement<A, B & C>;
```

I'm not sure what's up with `generate-fp`. You may need to debug it :/. I find that conditional breakpoints help a lot. If you're using VS Code, I've created a `launch.json` file that should make debugging work, as long as VS Code's project root is `DefinitelyTyped/types/lodash`.

Also keep in mind that there are now 2 `generate` scripts: one for TS<3.1 and one for TS>=3.1.

PR #36902, opened 2019-07-15
(got) fix: make query more specific

jaulz:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/sindresorhus/got#searchparams

### DIFF
--- a/types/got/index.d.ts
+++ b/types/got/index.d.ts
@@ -202,7 +202,7 @@ declare namespace got {
         baseUrl?: string;
         cookieJar?: CookieJar;
         encoding?: E;
-        query?: string | Record<string, string | number> | URLSearchParams;
+        query?: Record<string, string | number | boolean | null> | URLSearchParams | string;
         timeout?: number | TimeoutOptions;
         retry?: number | RetryOptions;
         followRedirect?: boolean;
--- a/types/got/index.d.ts
+++ b/types/got/index.d.ts
@@ -202,7 +202,7 @@ declare namespace got {
         baseUrl?: string;
         cookieJar?: CookieJar;
         encoding?: E;
-        query?: Record<string, string | number | boolean | null> | URLSearchParams | string;
+        query?: Record<string, any> | URLSearchParams | string;
         timeout?: number | TimeoutOptions;
         retry?: number | RetryOptions;
         followRedirect?: boolean;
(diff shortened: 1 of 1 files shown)


types/got/index.d.ts
```diff
@@ -202,7 +202,7 @@ declare namespace got {
         baseUrl?: string;
         cookieJar?: CookieJar;
         encoding?: E;
-        query?: string | object;
+        query?: string | Record<string, string>;
```
ikokostya: This is breaking change. Now, user can specify object with non-string value (number or array). See https://github.com/sindresorhus/got/issues/638. So, in my opinion, right type is `Record<string, any>`. 

I think, `URLSearchParams` should be also specified here https://github.com/sindresorhus/got/tree/v9.6.0#query

ikokostya (changes requested):

jaulz:
@ikokostya better? :blush:

types/got/index.d.ts
```diff
@@ -202,7 +202,7 @@ declare namespace got {
         baseUrl?: string;
         cookieJar?: CookieJar;
         encoding?: E;
-        query?: string | object;
+        query?: string | Record<string, string | number> | URLSearchParams;
```
ikokostya: As I mentioned before, `Record<string, string | number>` is breaking change. Type in documentation is just a hint, user can still pass any type and it will be converted to string. I think more appropriate type here is `Record<string, any>`. Also, I don't understand  why `number` should be excluded?

For example, `querystring.stringify()` from `Node.js` declaration takes any type: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/46014b89d69bf012809c9190374d883204516193/types/node/querystring.d.ts#L17

I want to hear opinions of other maintainers.

ikokostya (changes requested):

jaulz:
I changed it and used the type that's also used in the recent repository:
https://github.com/sindresorhus/got/blob/e9c01e03b07acb80ef5501dab97e9dd1a2e61af8/source/utils/types.ts#L141

ikokostya:
> I changed it and used the type that's also used in the recent repository:
https://github.com/sindresorhus/got/blob/e9c01e03b07acb80ef5501dab97e9dd1a2e61af8/source/utils/types.ts#L141

`querystring.stringify()` take undefined as well:

```js
> querystring.stringify({x: undefined})
'x=
```

Official repo types are incorrect.

jaulz:
Okay, fine, here we go 😊

ikokostya (approved):

PR #36906, opened 2019-07-15
@types/node global process.exitCode should be optional

kuchta:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/node/globals.d.ts
+++ b/types/node/globals.d.ts
@@ -863,7 +863,7 @@ declare namespace NodeJS {
         emitWarning(warning: string | Error, name?: string, ctor?: Function): void;
         env: ProcessEnv;
         exit(code?: number): never;
-        exitCode: number;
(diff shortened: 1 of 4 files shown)


Flarna (approved):
Usually it's not needed to adapt abandoned versions like v9 and 11 but it's not wrong either. Compared to these two v8 is still alive.

SimonSchick (approved):
Agree on the non LTS node versions, we don't really want to update those, it's fine for now but keep it in mind :)

kuchta:
It's not explicitly mentioned in the documentation [process.exitCode](https://nodejs.org/dist/latest-v12.x/docs/api/process.html#process_process_exitcode), but from using it for a long time, process.exitCode is not defined until process.exit(code?) is called. Detecting if it's set I usually end log running programs after catching process.on("SIGINT", ...) and setting it...

PR #36908, opened 2019-07-15
add @storybook/polymer

ryanis-online:
Please fill in this template.

- [ ] Add or edit tests to reflect the change. (Run with `npm test`.) (cannot get past npm naming issue. will open issue)
- [ ] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present). (cannot get this to run without npm naming issue. will open issue)

If adding a new definition:

### DIFF
--- a/types/storybook__polymer/index.d.ts
+++ b/types/storybook__polymer/index.d.ts
@@ -1,11 +1,11 @@
-// Type definitions for @storybook/polymer 5.0
+// Type definitions for @storybook/polymer 5.1.9
 // Project: https://github.com/storybookjs/storybook, https://github.com/storybookjs/storybook/tree/master/app/polymer
 // Definitions by: Joscha Feth <https://github.com/joscha>
 //                 Anton Izmailov <https://github.com/wapgear>
 //                 Dan Dean <https://github.com/dandean>
 //                 Ryan Coleman <https://github.com/ryuhhnn>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.4
 
 /// <reference types="webpack-env" />
 
--- a/types/storybook__polymer/index.d.ts
+++ b/types/storybook__polymer/index.d.ts
@@ -1,0 +1,48 @@
+// Type definitions for @storybook/polymer 5.0
+// Project: https://github.com/storybookjs/storybook, https://github.com/storybookjs/storybook/tree/master/app/polymer
+// Definitions by: Joscha Feth <https://github.com/joscha>
+//                 Anton Izmailov <https://github.com/wapgear>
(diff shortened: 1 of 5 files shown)


types/storybook__polymer/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "no-unnecessary-generics": false
```
andrewbranch: There shouldn’t be anything in `rules`

types/storybook__polymer/index.d.ts
```diff
@@ -0,0 +1,48 @@
+// Type definitions for @storybook/polymer 5.1
+// Project: https://github.com/storybookjs/storybook, https://github.com/storybookjs/storybook/tree/master/app/polymer
+// Definitions by: Joscha Feth <https://github.com/joscha>
+//                 Anton Izmailov <https://github.com/wapgear>
+//                 Dan Dean <https://github.com/dandean>
+//                 Ryan Coleman <https://github.com/ryuhhnn>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.4
```
andrewbranch: At a glance I don't think there's anything in here that requires TypeScript 3.4? This number should be the lowest possible version of TypeScript that is compatible with these definitions.

andrewbranch (changes requested):

types/storybook__polymer/tslint.json
ryuhhnn: Don't know if this is documented anywhere, but there are a large number of packages with tslint rules. I removed them though

ryuhhnn:
@andrewbranch feedback addressed, feel free to ping me if this requires further changes

types/storybook__polymer/tslint.json
andrewbranch: It’s here: https://github.com/DefinitelyTyped/DefinitelyTyped#lint

At the time we started linting DefinitelyTyped, there were already thousands of packages. Fixing them all to conform would have been a herculean effort with a questionable rate of return, so we auto-generated tslint configs disabling the smallest number of rules possible for each package. That‘s why you see existing exceptions—in a perfect world, there would be none.

andrewbranch:
It looks like `lit-element` uses the `unknown` type, which is available starting in TypeScript 3.0—that should be the TypeScript version specified in your header.

andrewbranch:
Ugh

```
Error in storybook__polymer
Error: Errors in typescript@3.0 for external dependencies:
node_modules/lit-html/lib/parts.d.ts(16,89): error TS2304: Cannot find name 'bigint'.
```

`bigint` was added in 3.2... sorry Ryan 🤦‍♂

andrewbranch (approved):

PR #36920, opened 2019-07-16
Node vm module additions

mvcbox:
Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://nodejs.org/api/vm.html#vm_class_vm_sourcetextmodule>

If removing a declaration:


types/node/vm.d.ts
```diff
@@ -19,16 +22,26 @@ declare module "vm" {
          */
         columnOffset?: number;
     }
-    interface ScriptOptions extends BaseOptions {
-        displayErrors?: boolean;
-        timeout?: number;
-        cachedData?: Buffer;
+
+    interface ScriptOptionsObject extends BaseOptions {
+        cachedData?: Buffer | TypedArray | DataView;
```
SimonSchick: Buffer is included in `TypedArray`

types/node/vm.d.ts
```diff
@@ -1,7 +1,10 @@
 declare module "vm" {
+    import TypedArray = NodeJS.TypedArray;
```
SimonSchick: Hu, TIL you can import from namespaces!

Anyhow, this is not consistent to the rest of the code base, please use the namespace directly for now.

types/node/vm.d.ts
```diff
@@ -19,16 +22,26 @@ declare module "vm" {
...
-        displayErrors?: boolean;
-        timeout?: number;
-        cachedData?: Buffer;
+
+    interface ScriptOptionsObject extends BaseOptions {
+        cachedData?: Buffer | TypedArray | DataView;
+        /** @deprecated This option is deprecated in favor of script.createCachedData() */
         produceCachedData?: boolean;
+        /** This option is part of the experimental API for the --experimental-modules flag, and should not be considered stable. */
+        importModuleDynamically?: (specifier: string, module: SourceTextModule) => Object | SourceTextModule;
```
SimonSchick: Prefer to declare as method eg. `importModuleDynamically?(...): R`

SimonSchick (changes requested):

types/node/vm.d.ts
mvcbox: Done (```TypedArray```  removed)

types/node/vm.d.ts
mvcbox: This line has been removed.

types/node/vm.d.ts
mvcbox: Done

types/node/vm.d.ts
mvcbox: Done

types/node/vm.d.ts
mvcbox: Done

mvcbox:
Requested changes were made

SimonSchick (approved):

types/node/vm.d.ts
```diff
@@ -79,13 +90,51 @@ declare module "vm" {
...
         runInNewContext(sandbox?: Context, options?: RunningScriptOptions): any;
         runInThisContext(options?: RunningScriptOptions): any;
         createCachedData(): Buffer;
     }
+
+    /**
+     * This feature is only available with the --experimental-vm-modules command flag enabled.
+     * https://nodejs.org/api/vm.html#vm_class_vm_sourcetextmodule
+     */
+    class SourceTextModule {
```
weswigham: What produces this? I don't see a definition for `vm.createSourceTextModule`.

weswigham:
I'm not sure we should be including experimental flag APIs in the normal node declaration files, since those members aren't normally there (and will consequently throw at runtime) - maybe if we made a prerelease tag version?

types/node/vm.d.ts
mvcbox: An instance of this class is created using
```
const instance = new createSourceTextModule('source code...');
```

mvcbox:
> I'm not sure we should be including experimental flag APIs in the normal node declaration files, since those members aren't normally there (and will consequently throw at runtime) - maybe if we made a prerelease tag version?

It appeared in Node.js since version 9.6.0
I am also sure that the one who will use this functionality knows about the need to use special options for Node.js

weswigham:
Right, but it's been behind a flag for that entire duration, too... And it's exactly the people that might just stumble into it and think it might be useful for them whom wouldn't know a flag was needed, and I can't believe that using the experimental flags is all that common...

mvcbox:
> Right, but it's been behind a flag for that entire duration, too... And it's exactly the people that might just stumble into it and think it might be useful for them whom wouldn't know a flag was needed, and I can't believe that using the experimental flags is all that common...

Yes, I understand what you mean.
But nevertheless, I tend more to the fact that if a person sees this for the first time, he will still have to consult the documentation to see how to use it.
Personally, my opinion is that the presence of this functionality will do more good than harm.

types/node/vm.d.ts
weswigham: Right... there's no definition for that here, though - the class is just an orphaned type.

jessetrinity:
It sounds like there is a disagreement here on including the types behind the experimental flag. Are we still waiting on the opinion of a third contributor?

SimonSchick:
Imo if it's in node, it should be in the typings, experimental features in node often have lifespans across multiple major versions.
We can always remove them later if they are removed in node.

sandersn:
New DT week, new maintainer here. Let me see if I can summarise the blocking disagreement: whether or not experimental features in node that are currently behind the flag `--experimental-vm-modules` should have types in `@types/node`.

@rbuckton What is your opinion on this?

types/node/vm.d.ts
```diff
@@ -19,16 +20,26 @@ declare module "vm" {
...
         produceCachedData?: boolean;
+        /** This option is part of the experimental API for the --experimental-modules flag, and should not be considered stable. */
+        importModuleDynamically?(specifier: string, module: SourceTextModule): Object | SourceTextModule;
     }
+
+    /**
+     * If options is a string, then it specifies the filename.
+     * https://nodejs.org/api/vm.html#vm_constructor_new_vm_script_code_options
+     */
+    type ScriptOptions = ScriptOptionsObject | string;
```
rbuckton: I think it would make more sense to leave the interface name `ScriptOptions` and add the `| string` union to the methods that possibly accept strings.

types/node/vm.d.ts
```diff
@@ -19,16 +20,26 @@ declare module "vm" {
...
-        displayErrors?: boolean;
-        timeout?: number;
-        cachedData?: Buffer;
+
+    interface ScriptOptionsObject extends BaseOptions {
+        cachedData?: NodeJS.TypedArray | DataView;
+        /** @deprecated This option is deprecated in favor of script.createCachedData() */
         produceCachedData?: boolean;
+        /** This option is part of the experimental API for the --experimental-modules flag, and should not be considered stable. */
+        importModuleDynamically?(specifier: string, module: SourceTextModule): Object | SourceTextModule;
```
rbuckton: I don't think `Object` is what you want here. I assume `object` isn't available in the minimum TypeScript version for Node?

types/node/vm.d.ts
rbuckton: The `SourceTextModule` class *is* exported by `vm` when `--experimental-vm-modules` is enabled.

rbuckton (changes requested):

rbuckton:
Would it be feasible to make the types for experimental features possibly `undefined`, since they may not be present?

types/node/vm.d.ts
rbuckton: The `SourceTextModule` class *is* exported by `vm` when `--experimental-vm-modules` is enabled.

NOTE: There is no `createSourceTextModule` function in Node 12, as far as I can see.

Flarna:
Regarding experimental APIs in Node: It's quite common to have them included for more then a major Node.JS version. An example is HTTP2 - that time it was just added and adapted as needed.

Another option would be to create a dedicated type definition package independent from Node typings for the experimental module. But I'm not sure if this is even supported by the typescript/dt tooling.

types/node/vm.d.ts
mvcbox: Such an approach would be counterintuitive, since when importing "ScriptOptions" type, the developer expects to contain all available configuration options.
I solved this problem as follows:
```typescript
type ScriptOptions = {
        cachedData?: NodeJS.TypedArray | DataView;
        /** @deprecated This option is deprecated in favor of script.createCachedData() */
        produceCachedData?: boolean;
        /** This option is part of the experimental API for the --experimental-modules flag, and should not be considered stable. */
        importModuleDynamically?(specifier: string, module: SourceTextModule): any | SourceTextModule;
    } & BaseOptions | string;
```

types/node/vm.d.ts
mvcbox: I think that at the moment we can define this as ```any```, since the return type is quite specific

PR #36928, opened 2019-07-16
[@types/mongoose] Enhance aggregate function with proper signature

hilzfa:
This change is needed to give a proper information about type of the aggregate result in the callback function. I use this coding in my own application already and everything is working. I use the `aggregate` function with the following syntax:

```typescript
Question.aggregate([{ $sample: { size: 5 } }], (err, questions) => {
        // use questions fully typed with Array<IQuestion>
});
```

Without my changes `questions` is typed as `any` and i have to explicitly type `questions` like 

```typescript
Question.aggregate([{ $sample: { size: 5 } }], (err:any, questions:IQuestion[]) => {
        
});
```

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: It's just a stronger typed function

### DIFF
--- a/types/mongoose/index.d.ts
+++ b/types/mongoose/index.d.ts
@@ -25,6 +25,7 @@
 //                 Vlad Melnik <https://github.com/vladmel1234>
 //                 Jarom Loveridge <https://github.com/jloveridge>
 //                 Grimmer Kang <https://github.com/grimmer0125>
+//                 Fabian Hilz <https://github.com/locke21>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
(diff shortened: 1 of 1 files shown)


Alorel (changes requested):
I don't think this PR should be merged as aggregations can project properties not defined on the model interface. Consider this aggregation:

```typescript
  UserModel
    .aggregate()
    .project({
      foo: 1,
      bar: 1,
      computedPropertyThatsNotOnTheModel: {
        $cond: [{$gte: ['$bar', 100]}, true, false]
      }
    });
```

The output type wouldn't have the computed property and it doesn't make sense to include that property on the model interface if it's only part of this one aggregation.

Locke21:
@Alorel Thnaks for your quick response. I see your point and to be honest, i did not think about that case. Totally agree.
I'm fine, when the PR is not merged. :)

Have a nice day!

PR #36930, opened 2019-07-16
Added some more Accounts methods and fixed types for findOne and DDPStatic

perbergland:
Please fill in this template.

  * For `Accounts._checkPassword`: https://github.com/meteor/meteor/blob/devel/packages/accounts-password/password_server.js
  * For `Accounts.registerLoginHandler` https://github.com/meteor/meteor/blob/devel/packages/accounts-base/accounts_server.js
  * For `Accounts.callLoginMethod` https://github.com/meteor/meteor/blob/devel/packages/accounts-base/accounts_client.js

- [-] Increase the version number in the header if appropriate.


perbergland:
This is my first PR so pointer so what I should have done but didn’t would be appreciated. I find it hard to understand how to test the changes, for instance.

types/meteor/globals/accounts-base.d.ts
```diff
@@ -146,3 +146,23 @@ declare module Accounts {
 declare module Accounts {
     function onLogout(func: (user: Meteor.User, connection: Meteor.Connection) => void): void;
 }
+
+declare module Accounts {
```
perbergland: I chose to continue with the existing style of adding a new module for every change

types/meteor/ddp.d.ts
```diff
@@ -2,8 +2,8 @@ declare module "meteor/ddp" {
     module DDP {
         interface DDPStatic {
             subscribe(name: string, ...rest: any[]): Meteor.SubscriptionHandle;
-            call(method: string, ...parameters: any[]): void;
```
perbergland: This makes it possible to use connections in unit tests

types/meteor/mongo.d.ts
```diff
@@ -159,7 +159,7 @@ declare module "meteor/mongo" {
                 fields?: FieldSpecifier;
                 reactive?: boolean;
                 transform?: Function | null;
-            }): T;
+            }): T | undefined;
```
perbergland: The only relevant change in this file

types/meteor/globals/accounts-base.d.ts
```diff
@@ -146,3 +146,23 @@ declare module Accounts {
...
+    }
+
+    type Password = string | {
+        digest: string;
+        algorithm: "sha-256"
+    }
+
+    function callLoginMethod(options: LoginMethodOptions): void;
+
+    function registerLoginHandler(name: string, handler: Function): void;
```
andrei-markeev: There is a very good description of those methods available in the original JS files. It would be great if this description is also added here (with /** ... */ style comments) so that it is shown in code completion hints. Also, I think handler definition could be more strict?

andrei-markeev (changes requested):
- Would be great if you could add some tests with examples of usage of those methods. It really helps to understand what they are supposed to do and streamlines the review process.
- Instead of "Function", use a more strict definition whenever possible, please.
- Would be great if you could add some descriptions to those methods, especially since they are available in original files

types/meteor/globals/accounts-base.d.ts
perbergland: Absolutely - I wasn’t sure how much typings you preferred (I like lots of types)

perbergland: Seems I also made the change in the wrong file (in "globals")

perbergland: OK I think I got a better version now. It seems like someone previously hadn’t run scripts/generate-globals because the files in globals picked up some changes I hadn’t done

types/meteor/globals/accounts-base.d.ts
```diff
@@ -146,3 +146,75 @@ declare module Accounts {
...
+     * This function takes care of:
+     * - Updating the Meteor.loggingIn() reactive data source
+     * - Calling the method in 'wait' mode
+     * - On success, saving the resume token to localStorage
+     * - On success, calling Accounts.connection.setUserId()
+     * - Setting up an onReconnect handler which logs in with
+     *   the resume token
+     *
+     * Options:
+     * - methodName: The method to call (default 'login')
```
andrei-markeev: Options descriptions here can be moved right into the LoginMethodOptions, e.g.

```ts
interface LoginMethodOptions {
    /** The method to call (default 'login') */
    methodName?: string;

    // ...
}
```

andrei-markeev (changes requested):
This looks so much better than before, thanks a lot for addressing those!

One last really small thing, but I am sure it can improve the user experience significantly and it's easy to do...

types/meteor/globals/accounts-base.d.ts
perbergland: Done!

andrei-markeev (approved):

PR #36934, opened 2019-07-16
[office-js-preview] (Outlook) Update SharedProperties props

ElizabethSamuel-MSFT:
Please fill in this template.

### DIFF
--- a/types/office-js-preview/index.d.ts
+++ b/types/office-js-preview/index.d.ts
@@ -16905,13 +16905,13 @@ declare namespace Office {
          * The REST API's base URL (currently https://outlook.office.com/api).
          * Use with targetMailbox to construct REST operation's URL.
          * 
-         * Example usage: targetRestUrl + "/{api_version}/users/" + targetMailbox + "/{REST_operation}"
+         * Example usage: `targetRestUrl + "/{api_version}/users/" + targetMailbox + "/{REST_operation}"`
          */
         targetRestUrl: String;
         /**
          * The target/owner's mailbox. Use with targetRestUrl to construct REST operation's URL.
          * 
-         * Example usage: targetRestUrl + "/{api_version}/users/" + targetMailbox + "/{REST_operation}"
+         * Example usage: `targetRestUrl + "/{api_version}/users/" + targetMailbox + "/{REST_operation}"`
          */
         targetMailbox: String;
         /**
--- a/types/office-js-preview/index.d.ts
+++ b/types/office-js-preview/index.d.ts
@@ -16902,9 +16902,14 @@ declare namespace Office {
          */
         owner: String;
         /**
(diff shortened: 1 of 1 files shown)


types/office-js-preview/index.d.ts
```diff
@@ -16902,9 +16902,18 @@ declare namespace Office {
...
+         * The REST API's base URL (currently https://outlook.office.com/api).
+         * Use with targetMailbox to construct REST operation's URL.
+         * 
+         * Example usage: targetRestUrl + "/{api_version}/users/" + targetMailbox + "/{REST_operation}"
+         */
+        targetRestUrl: String;
+        /**
+         * The target/owner's mailbox. Use with targetRestUrl to construct REST operation's URL.
+         * 
+         * Example usage: targetRestUrl + "/{api_version}/users/" + targetMailbox + "/{REST_operation}"
```
AlexJerabek: Can we put these examples in code ticks?

AlexJerabek (approved):

types/office-js-preview/index.d.ts
```diff
@@ -16902,9 +16902,18 @@ declare namespace Office {
...
+         * 
+         * Example usage: `targetRestUrl + "/{api_version}/users/" + targetMailbox + "/{REST_operation}"`
+         */
+        targetRestUrl: String;
+        /**
+         * The target/owner's mailbox. Use with targetRestUrl to construct REST operation's URL.
+         * 
+         * Example usage: `targetRestUrl + "/{api_version}/users/" + targetMailbox + "/{REST_operation}"`
          */
-        restUrl: String;
```
andrewbranch: Seems like this reflects a breaking change—does the version number need to be bumped?

types/office-js-preview/index.d.ts
```diff
@@ -16902,9 +16902,18 @@ declare namespace Office {
...
+        targetMailbox: String;
```
andrewbranch: You definitely want `string` and not `String`, but I see the rest of the interface matches. It might be worth a follow-up PR to refactor all `String` -> `string`.

ElizabethSamuel-MSFT: I'll update that now.

types/office-js-preview/index.d.ts
ElizabethSamuel-MSFT: Will discuss with my team today and get back to you.

andrewbranch: The version is just supposed to reflect the npm version of the JS library. So if this property was removed from the JS in a version bump, the version of the type defs should match that. If it was incorrectly there to begin with, the version string need not be updated.

ElizabethSamuel-MSFT: This is a preview API which is being refined so we think this is ok.

PR #36937, opened 2019-07-16
[react-table] Updated getTfootTdProps and getThProps

whytspace:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/tannerlinsley/react-table/pull/1387, https://github.com/tannerlinsley/react-table/pull/1399
- [ ] Increase the version number in the header if appropriate. <br />*I don't know. Is it appropriate to increase the version number in this case?*
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. <br />*no substantial changes*

### DIFF
--- a/types/react-table/index.d.ts
+++ b/types/react-table/index.d.ts
@@ -326,11 +326,7 @@ export interface ComponentDecoratorProps {
     getTdProps: ComponentPropsGetterRC | ComponentPropsGetterR;
     getTfootProps: ComponentPropsGetter0;
     getTfootTrProps: ComponentPropsGetter0;
-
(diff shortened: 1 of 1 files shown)


Gelio (commented):
The code looks fine :+1:

The only thing I am worried about is the version bump. I am not well informed on whether the version should be increased in this case. This seems like a breaking change when it comes to typing, even though code with the previous typing would not work. Therefore, in my opinion, this should result in a major version increase.

Let's wait for others to chime in before changing it :slightly_smiling_face:

PR #36941, opened 2019-07-16
[@types/node] added missing 'highWaterMark' option for fs.createWriteStream

DKorosec:
Please fill in this template.

  * Even though it's [not documented on fs.createWriteStream](https://nodejs.org/dist/latest-v8.x/docs/api/fs.html#fs_fs_createwritestream_path_options), a [stream.Writable can take highWaterMark as an option](https://nodejs.org/dist/latest-v8.x/docs/api/stream.html#stream_constructor_new_stream_writable_options). Checking the [source code](https://github.com/nodejs/node/blob/master/lib/internal/fs/streams.js#L231), you can see that the options you pass to createWriteStream are passed to stream.Writable. I've been using it from node 6+ and it's included in [their tests](https://github.com/nodejs/node/blob/39141426d46a7e55d93ad2e8efa12ed86d223522/test/parallel/test-file-write-stream2.js#L67)

### DIFF
--- a/types/node/v7/base.d.ts
+++ b/types/node/v7/base.d.ts
@@ -2965,6 +2965,7 @@ declare module "fs" {
         mode?: number;
         autoClose?: boolean;
         start?: number;
+        highWaterMark?: number;
     }): WriteStream;
     export function fdatasync(fd: number, callback: Function): void;
     export function fdatasyncSync(fd: number): void;
--- a/types/node/fs.d.ts
+++ b/types/node/fs.d.ts
@@ -1724,6 +1724,7 @@ declare module "fs" {
         mode?: number;
         autoClose?: boolean;
         start?: number;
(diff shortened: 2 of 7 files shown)


types/node/v7/base.d.ts
```diff
@@ -2965,6 +2965,7 @@ declare module "fs" {
         mode?: number;
         autoClose?: boolean;
         start?: number;
+        highWaterMark?: number;
```
SimonSchick: You can skip updating 6 and 7 and 9 since they are not LTS.

SimonSchick (approved):

types/node/v7/base.d.ts
DKorosec: Should I revert the changes to 6, 7 and 9, or let them be?

PR #36944, opened 2019-07-17
add types: react-native-easy-upgrade

taoqf:
Please fill in this template.

If adding a new definition:


taoqf:
@typescript-bot Great!

types/react-native-easy-upgrade/index.d.ts
```diff
@@ -0,0 +1,82 @@
+// Type definitions for react-native-easy-upgrade 0.0
+// Project: https://github.com/react-native-org/react-native-easy-upgrade#readme
+// Definitions by: taoqf <https://github.com/taoqf>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
```
andrewbranch: Are the few instances of `unknown` here really worth restricting the version to 3.0 and higher? Callback return types are very frequently ignored, so `void` or `any` is common. `any` also seems appropriate for the `lookupInfo` dictionary. If you did that, I think you could take the version all the way down to 2.1, so potentially a lot more people could use these typings.

types/react-native-easy-upgrade/index.d.ts
```diff
@@ -0,0 +1,82 @@
...
+	downloadTitle: string;
+	downloadDescription: string;
+	downloadDestDirectory: string;
+	downloadApkName: string;
+	downloadApkEnd(path: string): unknown;
+	shouldCheckApkHasDownloaded: boolean;
+	onError(err: Error): unknown;
+}
+
+declare class RNEasyUpgrade {
```
andrewbranch: This _appears_ to be named incorrectly, according to https://unpkg.io/react-native-easy-upgrade@0.0.9/index.js. The structure looks right but the name is `AppUpgrade`. `RNEasyUpgrade` appears to be something else, although I don‘t entirely understand what.

andrewbranch (changes requested):
Fix the name (or if I’m wrong, explain what I missed), and give the version some consideration.

andrewbranch (approved):

PR #36952, opened 2019-07-17
(apimocker:new) new type definition for apimocker library

uchilaka:
Please fill in this template.

- [Y] Use a meaningful title for the pull request. Include the name of the package modified.
- [Y] Test the change in your own code. (Compile and run.)
- [Y] Add or edit tests to reflect the change. (Run with `npm test`.)
- [Y] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [Y] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [Y] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If adding a new definition:
- [Y] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration`
- [Y] If this is for an NPM package, match the name. If not, do not conflict with the name of an NPM package.
- [Y] Create it with `dts-gen --dt`, not by basing it on an existing project.
- [Y] `tslint.json` should be present, and `tsconfig.json` should have `noImplicitAny`, `noImplicitThis`, `strictNullChecks`, and `strictFunctionTypes` set to `true`.


types/apimocker/index.d.ts
```diff
@@ -0,0 +1,49 @@
+// Type definitions for apimocker 1.1
+// Project: https://www.npmjs.com/package/apimocker
+// Definitions by: Uchenna <https://github.com/uchilaka>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.5
+
+import { RequestHandler, Application } from 'express';
+
+export interface configOptions {
```
andrewbranch: Nit: conventionally, type names (including interfaces) are PascalCase (`ConfigOptions`)

types/apimocker/index.d.ts
```diff
@@ -0,0 +1,49 @@
+// Type definitions for apimocker 1.1
+// Project: https://www.npmjs.com/package/apimocker
+// Definitions by: Uchenna <https://github.com/uchilaka>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.5
```
andrewbranch: This should be the lowest possible version in which these definitions work. I don’t see any bleeding-edge new TypeScript features in here, so you should be safe to use whatever `express` uses—[2.2](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c79180dd81e7e726ef7763320ddbdc099eaec995/types/express/index.d.ts#L5). That will let people stuck on older TypeScript versions still benefit from these typings 👍

types/apimocker/package.json
```diff
@@ -0,0 +1,3 @@
+{
+    "private": true
```
andrewbranch: You don’t need to add a package.json unless you have external dependencies to put in here—you can simply delete the file.

types/apimocker/index.d.ts
```diff
@@ -0,0 +1,49 @@
...
+    middlewares: RequestHandler[];
+    setConfigFile: (file: string) => ApiMocker;
+    loadConfigFile: () => void;
+    setRoutes: (webServices: any) => void;
+    /**
+     * Set the route for express, in case it was not set yet
+     */
+    setRoute: (options: any) => void;
+    /**
+     * @TODO inspect Express.Application.start
```
andrewbranch: Did you mean to leave TODO comments in the JSDoc? People will see this in their editor completions 😄

andrewbranch (changes requested):

types/apimocker/index.d.ts
uchilaka: Woops! I was leaving notes for myself. I'll circle back and flesh out the interface more, but this covers the use case in the basic examples for initializing apimocker via node. I think this should be good to ship pending any other feedback until I can add more detail or (wishful!!) re-factor for TS.

uchilaka (commented):
All done - applied updates in a couple of commits, and re-ran lint. Looking good.

uchilaka:
@andrewbranch Hey - want to make sure I'm not missing anything... I figure the bot will eventually flip the switches to indicate the changes were made. I marked the convos as resolved. For now, looks like the review is still showing "1 change requested".

andrewbranch (approved):
Looks good, thanks!

PR #36959, opened 2019-07-17
[react-native] Added Linking.openSettings()

danilobuerger:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://facebook.github.io/react-native/docs/linking.html#opensettings

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -7531,6 +7531,11 @@ export interface LinkingStatic extends NativeEventEmitter {
     canOpenURL(url: string): Promise<boolean>;
 
     /**
+     * Open the Settings app and displays the app’s custom settings, if it has any.
+     */
+    openSettings(): Promise<any>;
+
+    /**
      * If the app launch was triggered by an app link with, it will give the link url, otherwise it will give null
      * NOTE: To support deep linking on Android, refer http://developer.android.com/training/app-indexing/deep-linking.html#handling-intents
      */


Skn0tt:
I made the exact same PR this morning: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36953  - so I approve of the contents, but sadly it's a duplicate.

Skn0tt (approved):

danilobuerger:
@Skn0tt Ah, I see. However, in your PR the return type is given as `Promise<void>` which does not match the flow typings:

https://github.com/facebook/react-native/blob/master/Libraries/Linking/Linking.js#L76

And indeed, it does resolve to non-void results:

https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/modules/intent/IntentModule.java#L165

Skn0tt:
I see your point. However, the Flow typings don't seem to be very strict about that, since the referenced `NativeLinking.openSettings()` [is annotated with `Promise<void>`](https://github.com/facebook/react-native/blob/a43615a38a8594217602e2dc91640e8902d288a2/Libraries/Linking/NativeLinking.js#L22) as its return type.

Using `Promise<any>` makes people wonder about what `any` means in this context. This is why I think that in this case, `Promise<void>` is the type to use. The real (implementation) type (which is `Promise<true>`) does not offer any benefits, but will confuse users; it's the same with `Promise<void>`.

andrewbranch:
If the resolved value isn’t supposed to be looked at, `void` is preferable. From the original Java it looks like there’s no _reason_ to inspect the value, since it’s currently always `true`. Making the TypeScript typings say `void` would prevent people from (for some odd reason) checking the resolved value for truthiness, which would leave more room for react-native to change the underlying implementation without breaking people.

All other things equal, I’d lean toward closing this one and merging @Skn0tt’s version, but will defer to the experts here. What do you all think?

alloy:
Looks like we have a verdict 👍 

Thanks, though, @danilobuerger! 🙌

PR #36969, opened 2019-07-17
Add new npm-naming exemptions

sandersn:
These packages fail the upcoming requirement that the types package version match a version that exists for the original package.

In other words, these type packages have a version, like 0.0 or 1.0, that their original package doesn't have. The "npm-naming" lint rule will soon prevent this, so these packages need to be exempt from this rule.


sandersn:
Wow, some of these packages are really stale. They require redux but have a require TS version that is older than it requires.

redux-sentry-middleware, tinajs__tina-redux, wepy-redux are the packages.

PR #36975, opened 2019-07-18
[relay-runtime]: RecordSource exposes a toJSON method too

maraisr:
With this PR I aim to expose the `toJSON` as a method on the RecordStore.

example:

```ts
const environment = new Environment(...);

environment
    .getStore()
    .getSource()
    .toJSON()
``` 

@voxmatt @alloy @ckknight  Did one of you guys mind reviewing this one?

### DIFF
--- a/types/relay-runtime/index.d.ts
+++ b/types/relay-runtime/index.d.ts
@@ -422,6 +422,7 @@ interface RecordSource {
     has(dataID: DataID): boolean;
     load(dataID: DataID, callback: (error: Error | null | undefined, record: Record | null | undefined) => void): void;
     size(): number;
+    toJSON(): Record;
 }
 export { RecordSource as IRecordSource };
 
@@ -1027,6 +1028,7 @@ declare class RelayInMemoryRecordSource implements MutableRecordSource {
     remove(dataID: DataID): void;
     set(dataID: DataID, record: Record): void;
     size(): number;
+    toJSON(): Record;
 }
 export { RelayInMemoryRecordSource as RecordSource };
 
--- a/types/relay-runtime/index.d.ts
+++ b/types/relay-runtime/index.d.ts
@@ -14,14 +14,9 @@ export interface ConnectionMetadata {
     count: string | null | undefined;
 }
 
-// ./handlers/connection/RelayConnectionInterface
-export interface Record {
-    [key: string]: unknown;
-}
-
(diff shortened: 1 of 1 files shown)


alloy (changes requested):
Thanks and yeah we'd definitely appreciate such changes!

However, I don’t think this is currently correct. I believe `toJSON` will give you an JS object, for example see [here](https://github.com/facebook/relay/blob/1a94841b1d7809836ef5b71854d3ade2b5cb4dde/packages/relay-runtime/store/RelayRecordSourceObjectImpl.js#L78), whereas `JSON.stringify` returns a `string`.

maraisr:
@alloy thank you so much for taking the time review this. Sorry about that misunderstanding.

Looking into the types more, it looks like this method returns the type `Record` (which is what I used) - do you think there is a appetite to just use the [built in type from TypeScript](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkt)?

The `toJSON` method expose the type `Record<DataID, unknown>` effectively, with the exception of the `__Root` typename, which for me is `client:root`.

Or is the Record type being implemented here, because of old versions of tsc?

Sure; DataID === string, and Record is a object of string keys. So its much of a muchness. 🤷🏼‍♂️

Thank you again - looking forward to this being merged! 🕺

alloy:
@maraisr Oh, I like that–yeah that makes sense to me 👍

alloy:
@typescript-bot Hold your horses, bottie, the PR is still marked as “change requested”.

andrewbranch (changes requested):
Requesting changes just because @typescript-bot got confused; this should unconfuse him (her? I don’t want to make assumptions). I learned the other day that if you comment with a thumbs-up emoji, it’s interpreted as an approval 😬

andrewbranch:
BTW, the pending CI checks actually completed (and succeeded)—it seems like Azure DevOps had some transient issues updating status checks, as another PR around the same time had the same problem.

maraisr:
Sorry friends I'm a little bit confused with I'm actually needing to still do.

cc @alloy @andrewbranch

alloy:
@maraisr Sorry for the confusion, I gave the thumbs-up to your suggestion to drop our custom `Record` and just use TS’ version.

alloy:
Great, thanks!

PR #37008, opened 2019-07-19
Add className for component Link

polRk:
https://github.com/molefrog/wouter/pull/61

### DIFF
--- a/types/wouter/index.d.ts
+++ b/types/wouter/index.d.ts
@@ -1,7 +1,8 @@
-// Type definitions for wouter 2.0
+// Type definitions for wouter 2.1
 // Project: https://github.com/molefrog/wouter#readme
 // Definitions by: Tolkunov Alexander <https://github.com/StrayFromThePath>
 //                 Maksim Karelov <https://github.com/Ty3uK>
+//                 Vladislav Polyakov <https://github.com/polRk>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
@@ -30,6 +31,7 @@ export const Route: FunctionComponent<RouteProps>;
 export interface LinkProps {
     to?: string;
     href?: string;
+    className?: string;
     children: ReactElement;
     onClick?: () => void;
 }


Ty3uK:
@polRk [#61](https://github.com/molefrog/wouter/pull/61) is not yet merged

Ty3uK:
@polRk [#61](https://github.com/molefrog/wouter/pull/61) is not yet merged and discussed. I think it's to early make PR to DefinitelyTyped. Also at new version typings will be moved to wouter's repo.

polRk:
I do not think so, you need to start somewhere. Types that in the repository do not work. Works only from DefinitelyTyped

Ty3uK:
@polRk they're works. New version with types not released yet, but it will be work if you build wouter and link it to your project folder

polRk:
I don't see the types in the node_modules/wouter folder

Ty3uK:
> I don't see the types in the node_modules/wouter folder

> New version with types not released yet

Also, simple adding `className` to `Link` is not an option, I think. Also, at Preact class field may be named both `class` and `className`.

That discussion we need to do at [#61](https://github.com/molefrog/wouter/pull/61).

polRk:
@Ty3uK, see an example https://github.com/molefrog/wouter#link-hrefpath-

Ty3uK:
@polRk I know that. What about `id`, `target`, etc. attributes? You posted

> Or is it better to expand component types with types React.AnchorHTMLAttributes ?

That kind of discussion I mean. I think you should use `JSX.HTMLAttributes` in Preact and it's analogue in React.

molefrog:
Hey @Ty3uK @polRk thanks for pointing out at this type mismatch. The latest version has indeed not yet been released, I'm preparing some final type tweaks to settle down before the release. 

I'm not completely sure that adding `className` to the hash would solve this, however. Links might have more attributes than just `className`, so it's better to find a way to tell TypeScript that this component extends Link's interface.

Anyway, let's move the convo to #61

weswigham:
Looks like yall have merged https://github.com/molefrog/wouter/pull/64 instead - I'll close this for now.

PR #37012, opened 2019-07-19
node-forge: Reference `jsbn.BigInteger` in favor of own types

TimWolla:
While node-forge technically does not use the `jsbn` package as its
dependency for all intents and purposes it is using the jsbn library,
no need to maintain these typings separately. Especially not when they
are clearly non-complete.

In fact the `@types/jsbn` typings already contains forge's special
`data` property which does not exist in stock `jsbn`.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/digitalbazaar/forge/blob/master/lib/jsbn.js and https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a1f6013ddd509d0275acf72d1b202c9e432738eb/types/jsbn/index.d.ts#L19

### DIFF
--- a/types/node-forge/index.d.ts
+++ b/types/node-forge/index.d.ts
@@ -16,6 +16,8 @@
 
 /// <reference types="node" />
 
+import { BigInteger as JsbnBigInteger } from 'jsbn';
+
 declare module "node-forge" {
     type Byte = number;
     type Bytes = string;
@@ -26,12 +28,7 @@ declare module "node-forge" {
     type Encoding = "raw" | "utf8";
 
     namespace jsbn {
-        class BigInteger {
-            data: number[];
-            t: number;
-            s: number;
-            toString(): string;
-        }
+        type BigInteger = JsbnBigInteger;
     }
 
     namespace pem {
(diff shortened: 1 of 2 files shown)


andersk:
I don’t think you can generally pass `forge.jbsn` numbers to `jsbn` or vice versa, so it doesn’t seem like a good idea for the types to allow this.

Perhaps you should ask `forge` upstream whether it wants to using the real `jsbn` library?

(See also #33931.)

andersk (changes requested):
I don’t think you can generally pass `forge.jbsn` numbers to `jsbn` or vice versa, so it doesn’t seem like a good idea for the types to suggest that you can.

Perhaps you should ask `forge` upstream whether it wants to switch to using the real `jsbn` library?

(See also #33931.)

TimWolla:
> I don’t think you can generally pass `forge.jbsn` numbers to `jsbn` or vice versa, so it doesn’t seem like a good idea for the types to suggest that you can.

Ugh, yeah. While `forge.jsbn` behaves like JSBN when interfacing with Forge it is not able to interoperate with stock JSBN.

PR #37017, opened 2019-07-20
[ramda] fix R.flatten 

swwind:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#ordering

------

Old code matches a wrong type unexpectedly:

```js
R.flatten([[1, 2], [3, 4]]); // typeof number[][]
```

I think that use `R.flatten` to flatten a two-dimensional array is more general than one, thus I swapped the order of two overloaded functions.

Now it could match the type correctly as we need.

```js
R.flatten([[1, 2], [3, 4]]); // typeof number[]
R.flatten([1, 2, 3, 4]); // typeof number[]
```


types/ramda/ramda-tests.ts
```diff
@@ -925,8 +925,8 @@ interface Obj {
 };
 
 () => {
-    R.flatten([1, 2, [3, 4], 5, [6, [7, 8, [9, [10, 11], 12]]]]);
-    // => [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]
+    R.flatten([[1, 2], [2, 3], [3, 4]]); // => [1, 2, 3, 4, 5, 6]
```
Nemo108: wrong
`R.flatten([[1, 2], [2, 3], [3, 4]]);` will give you `[1, 2, 2, 3, 3, 4]`

types/ramda/index.d.ts
```diff
@@ -1405,7 +1405,7 @@ declare namespace R {
          * Returns a new list by pulling every item out of it (and all its sub-arrays) and putting
          * them in a new array, depth-first.
          */
-        flatten<T>(x: ReadonlyArray<T> | ReadonlyArray<T[]> | ReadonlyArray<ReadonlyArray<T>>): T[];
+        flatten<T>(x: ReadonlyArray<T[]> | ReadonlyArray<ReadonlyArray<T>> | ReadonlyArray<T>): T[];
```
Nemo108: what case does this swap solves?

swwind: Maybe I need more tests?

Nemo108: I've ran some. Current typings aren't the best, but it's not because of your changes.
Your contribution is good

Nemo108 (approved):

PR #37024, opened 2019-07-20
[@types/react] make optional createContext defaultValue argument

### LINKED ISSUE #37023: [@types/react] createContext – An argument for 'defaultValue' was not provided.
Is there any reason of why the argument `defaultValue` is not optional here?

```tsx
import * as firebase from 'firebase/app';

interface AuthProps {
  children: any;
}
interface AuthState {
  user: firebase.User | null;
}
interface AuthContextProps extends AuthState {
  updateUser(user: firebase.User | null): void;
}

export const AuthContext = React.createContext<AuthContextProps>(); // <-- ERROR HERE!

// export const AuthConsumer = AuthContext.Consumer;
export const AuthProvider = (props: AuthProps) => {
  const [state, setState] = React.useState<AuthState>({
    user: null,
  });

  const updateUser = (user: firebase.User | null) => setState({ user });

  return (
    <AuthContext.Provider
      value={{
        updateUser,
        user: state.user,
      }}
    >
      {props.children}
    </AuthContext.Provider>
  );
};

```

Docs url: https://reactjs.org/docs/context.html#reactcreatecontext

I have a matching Provider above in the tree, so I should be able to create a context without set a default value or I'm wrong?

exwyezed:
This PR fixes #37023

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -334,7 +334,7 @@ declare namespace React {
         displayName?: string;
     }
     function createContext<T>(
-        defaultValue: T,
+        defaultValue?: T,
         calculateChangedBits?: (prev: T, next: T) => number
     ): Context<T>;
 


types/react/index.d.ts
```diff
@@ -334,7 +334,7 @@ declare namespace React {
         displayName?: string;
     }
     function createContext<T>(
-        defaultValue: T,
+        defaultValue?: T,
```
eps1lon: This means you could write `createContext<object>();` and get no compile errors when accessing the context unguarded.

What we want is probably an overload. Something like
```ts
function createContext<T>(): Context<T | undefined>;
```

eps1lon (changes requested):

types/react/index.d.ts
alfonmga: Looks good to me, should I apply the suggested change `function createContext<T>(): Context<T | undefined>;` and we can LGTM?

eps1lon: Sure. But don't forget to add a test.

alfonmga:
I'm not 100% sure about this PR, I'm going to close this and let others decide because maybe I'm wrong as I don't have enough experience with TypeScript.

PR #37033, opened 2019-07-21
leaflet: update L.Browser constants

simon04:
Ref: https://leafletjs.com/reference-1.5.0.html#browser

Follow-up to #36858


alejo90 (approved):
👍

PR #37057, opened 2019-07-22
Prepare for TS #32460

ahejlsberg:
Prepare `react-dom` and `react-test-renderer` for microsoft/TypeScript#32460.

### DIFF
--- a/types/react-test-renderer/index.d.ts
+++ b/types/react-test-renderer/index.d.ts
@@ -65,9 +65,8 @@ export function create(nextElement: ReactElement, options?: TestRendererOptions)
  * @see https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#testing-hooks
  */
 // the "void | undefined" is here to forbid any sneaky "Promise" returns.
-// the actual return value is always a "DebugPromiseLike",
-// but having an "| {}" makes it harder to accidentally use.
-export function act(callback: () => void | undefined): DebugPromiseLike | {};
+// the actual return value is always a "DebugPromiseLike".
+export function act(callback: () => void | undefined): DebugPromiseLike;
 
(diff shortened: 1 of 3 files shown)


types/react-test-renderer/index.d.ts
```diff
@@ -65,9 +65,8 @@ export function create(nextElement: ReactElement, options?: TestRendererOptions)
  * @see https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#testing-hooks
  */
 // the "void | undefined" is here to forbid any sneaky "Promise" returns.
-// the actual return value is always a "DebugPromiseLike",
-// but having an "| {}" makes it harder to accidentally use.
-export function act(callback: () => void | undefined): DebugPromiseLike | {};
+// the actual return value is always a "DebugPromiseLike".
+export function act(callback: () => void | undefined): DebugPromiseLike;
```
jgoz: @Jessidhia Without `| {}`, you could call `act().then(...)` without a cast - is that correct? Is there some other way to make it harder to use this (presumably internal) API?

ahejlsberg: It is still *really* hard to call `act.then(...)` because you get a compile error unless you pass two callbacks that each return `never`.

The type `{} | DebugPromiseLike` is definitely suspicious. Ordinarily the compiler would just reduce it to `{}` (since `DebugPromiseLike` is a subtype of `{}`), but for reasons of efficiency and backwards compatibility we don't always reduce unions. With the changes in microsoft/TypeScript#32460 we now infer `{}` for `T` when inferring from `{} | DebugPromiseLike` to `T | PromiseLike<T>`, which certainly is a better inference than the previous `{} | DebugPromiseLike`.

PR #37061, opened 2019-07-22
feat(react-transition-group): types for v4.2 (SwitchTransition)

setsun:
### Description
This adds the `<SwitchTransition />` component, and the `modes` export from `react-transition-group` to `@types/react-transition-group`: https://github.com/reactjs/react-transition-group/blob/master/src/SwitchTransition.js

This also bumps the version to `v4.2` to match: https://github.com/reactjs/react-transition-group/releases/tag/v4.2.0

Tests have been added to test that the `<SwitchTransition />` accepts the expected types.

### DIFF
--- a/types/react-transition-group/SwitchTransition.d.ts
+++ b/types/react-transition-group/SwitchTransition.d.ts
@@ -6,21 +6,18 @@ export enum modes {
 }
 
 declare namespace SwitchTransition {
-
   interface SwitchTransitionProps {
     /**
      * Transition modes.
      * `out-in`: Current element transitions out first, then when complete, the new element transitions in.
      * `in-out: New element transitions in first, then when complete, the current element transitions out.`
-     *
-     * @type {'out-in'|'in-out'}
      */
-    mode?: modes.out | modes.in
+    mode?: modes.out | modes.in;
 
     /**
      * Any `Transition` or `CSSTransition` component
      */
-    children: React.ReactElement
+    children: React.ReactElement;
   }
 }
 
--- a/types/react-transition-group/SwitchTransition.d.ts
+++ b/types/react-transition-group/SwitchTransition.d.ts
@@ -1,0 +1,54 @@
+import { Component } from "react";
+
+export enum modes {
+  out = 'out-in',
+  in = 'in-out'
+}
+
+declare namespace SwitchTransition {
+
+  interface SwitchTransitionProps {
+    /**
+     * Transition modes.
+     * `out-in`: Current element transitions out first, then when complete, the new element transitions in.
+     * `in-out: New element transitions in first, then when complete, the current element transitions out.`
+     *
+     * @type {'out-in'|'in-out'}
+     */
+    mode?: modes.out | modes.in
+
+    /**
+     * Any `Transition` or `CSSTransition` component
+     */
+    children: React.ReactElement
+  }
+}
+
+/**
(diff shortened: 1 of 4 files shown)


types/react-transition-group/SwitchTransition.d.ts
```diff
@@ -0,0 +1,51 @@
...
+}
+
+declare namespace SwitchTransition {
+  interface SwitchTransitionProps {
+    /**
+     * Transition modes.
+     * `out-in`: Current element transitions out first, then when complete, the new element transitions in.
+     * `in-out: New element transitions in first, then when complete, the current element transitions out.`
+     */
+    mode?: 'out-in' | 'in-out';
```
ybiquitous: It seems possible to use `enum modes` here instead of string literals. What about?

```suggestion
    mode?: modes
```

types/react-transition-group/SwitchTransition.d.ts
```diff
@@ -0,0 +1,51 @@
...
+  out = 'out-in',
+  in = 'in-out'
+}
+
+declare namespace SwitchTransition {
+  interface SwitchTransitionProps {
+    /**
+     * Transition modes.
+     * `out-in`: Current element transitions out first, then when complete, the new element transitions in.
+     * `in-out: New element transitions in first, then when complete, the current element transitions out.`
```
ybiquitous: I notice the position of the end backquote is incorrect. But the original is same too. 😅 

```suggestion
     * `in-out`: New element transitions in first, then when complete, the current element transitions out.
```


Could you please modify it? I open a PR to the original repository in order to fix the original code.
https://github.com/reactjs/react-transition-group/pull/524

ybiquitous (changes requested):
Thanks, it looks almost good to me! I left some small comments, so please check them. 😃

types/react-transition-group/SwitchTransition.d.ts
Setsun: Hey there 👋 

I did try this, but TypeScript is pretty strict when I change it to this, and will require that you always use the `enum modes` export when you pass in the React prop. 

So the following example would be a type error:
```tsx
import SwitchTransition from 'react-transition-group/SwitchTransition';

// results in type error
<SwitchTransition mode="in-out">
  {...}
</SwitchTransition>
```

Whereas this would not result in an error:
```tsx
import SwitchTransition, { modes } from 'react-transition-group/SwitchTransition';

// we're using the enum directly, so this is 👌🏼
<SwitchTransition mode={modes.in}>
  {...}
</SwitchTransition>
```

ybiquitous: I see👍

ybiquitous (approved):
LGTM!👍

types/react-transition-group/SwitchTransition.d.ts
Epskampie: If I can weigh in on this, I think it's much better to allow the string ( `<SwitchTransition mode="in-out">` ), since it will 

* be more in line with the official docs
* be easier to type (no need to figure out the import)
* have the same type safety (invalid values still not allowed)

Setsun: Agreed. I've left this as is, so the current state of this PR implements the behavior you're describing 👍

Epskampie: Ah, ok, good to hear!

PR #37067, opened 2019-07-23
swagger-ui-dist additional typing

brycematheson1234:
Extended the fairly barebones typing of `swagger-ui-dist`. This is far from a full typing of `swagger-ui-dist` but is much more complete than it was before, and has `[k: string]: any;` sprinkled throughout definitions to account for the definitions that are still not in there.

### DIFF
--- a/types/swagger-ui-dist/index.d.ts
+++ b/types/swagger-ui-dist/index.d.ts
@@ -1,22 +1,46 @@
 // Type definitions for swagger-ui-dist 3.x
 // Project: https://github.com/swagger-api/swagger-ui#readme
 // Definitions by: Haowen <https://github.com/haowen737>
+//                 Bryce <https://github.com/brycematheson1234>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-/* =================== USAGE ===================
 
-    import * as SwaggerUIDist from "swagger-ui-dist"
-    const swaggerUiAssetPath = SwaggerUIDist.getAbsoluteFSPath()
-
- =============================================== */
-
-export as namespace SwaggerUIDist;
-
-/**
- * get an absolute path to swagger ui for static file serving
- */
-export function getAbsoluteFSPath(): string;
-export function absolutePath(): string;
-
-export const SwaggerUIBundle: any;
-export const SwaggerUIStandalonePreset: any;
+declare module 'swagger-ui-dist' {
+    /**
+     * get an absolute path to swagger ui for static file serving
+     */
+    export function getAbsoluteFSPath(): string;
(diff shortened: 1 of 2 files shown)


haowen737 (commented):
Thanks for these additional changes!

haowen737 (changes requested):
These changes can be merged after  [dtslint test error](https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/562452368?utm_source=github_status&utm_medium=notification) fixed 😉

Thanks again! looking forward to this being merged!

brycematheson1234:
@haowen737 fixed linting issues 😄

haowen737 (approved):

types/swagger-ui-dist/index.d.ts
```diff
@@ -18,5 +13,33 @@ export as namespace SwaggerUIDist;
...
-export const SwaggerUIBundle: any;
+interface SwaggerUIBundle {
+  (a: {
+    deepLinking: boolean;
+    dom_id: string;
+    presets: any[];
+    plugins: any;
+    urls: Array<[string, string]>;
+    layout: string;
+  }): any;
```
leoiii12: Some params are missing. e.g. spec, url


Parameter name | Docker variable | Description
-- | -- | --
configUrl | CONFIG_URL | String. URL to fetch external configuration document from.
dom_id | DOM_ID | String, REQUIRED if domNode is not provided. The ID of a DOM element inside which SwaggerUI will put its user interface.
domNode | Unavailable | Element, REQUIRED if dom_id is not provided. The HTML DOM element inside which SwaggerUI will put its user interface. Overrides dom_id.
spec | SPEC | Object={}. A JavaScript object describing the OpenAPI definition. When used, the url parameter will not be parsed. This is useful for testing manually-generated definitions without hosting them.
url | URL | String. The URL pointing to API definition (normally swagger.json or swagger.yaml). Will be ignored if urls or spec is used.
urls | URLS | Array. An array of API definition objects ([{url: "<url1>", name: "<name1>"},{url: "<url2>", name: "<name2>"}]) used by Topbar plugin. When used and Topbar plugin is enabled, the url parameter will not be parsed. Names and URLs must be unique among all items in this array, since they're used as identifiers.
urls.primaryName | URLS_PRIMARY_NAME | String. When using urls, you can use this subparameter. If the value matches the name of a spec provided in urls, that spec will be displayed when Swagger UI loads, instead of defaulting to the first spec in urls.

leoiii12: https://github.com/swagger-api/swagger-ui/blob/master/docs/usage/configuration.md

PR #37076, opened 2019-07-23
[@types/slate] Update for version 0.47

kaydelaney:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/ianstormtaylor/slate/tree/67e397100e2f4cc6bcd4b8f98d7a85dcf30c72fd/packages/slate/src

Making this as a draft PR since there are a lot of changes and I'd like to get some more opinions before making a proper PR.

Besides updating the definitions to match version 0.47 of Slate, I've also added some definitions that were missing before, and changed definitions that weren't totally correct.


benjiro:
@kaydelaney Thanks, I should have some spare time on the weekend to review it.

Cheers,
Ben

kaydelaney:
@benjiro  awesome, thanks!

types/slate/index.d.ts
```diff
@@ -34,10 +35,12 @@ export interface ObjectAndType {
 }
 
 export interface Rules {
+    rules?: Array<({ match: ObjectAndType | ObjectAndType[] } & Rules)>;
```
fuafa: Should this go to `SchemaProperties`?

kaydelaney: Yes it should, nice catch, thanks!

benjiro (approved):
LGTM 
Thanks for all the work @kaydelaney, the changes look good to me.

benjiro:
Don't merge this yet, please.
@fuafa Once you've finished reviewing and ok with the changes, we will merge this in.

benjiro (changes requested):
#88 RenderMark has now become  RenderAnnotation & RenderDecoration
https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/Changelog.md

```
The renderMark middleware is no longer used for decorations! Previously renderMark would be used for rendering both types of decorations. Now there are separate renderAnnotation and renderDecoration middleware functions to use instead.
```

kaydelaney:
> #88 RenderMark has now become RenderAnnotation & RenderDecoration
> https://github.com/ianstormtaylor/slate/blob/master/packages/slate-react/Changelog.md
> 
> ```
> The renderMark middleware is no longer used for decorations! Previously renderMark would be used for rendering both types of decorations. Now there are separate renderAnnotation and renderDecoration middleware functions to use instead.
> ```

@benjiro Just pushed a commit (hopefully) fixing this. Also added utility methods to the Editor class.

benjiro (approved):
Thanks for that change, looks good now. I'm right for this to be merged

kaydelaney:
@benjiro awesome!

jessetrinity:
Before I merge this, do you want to be updating the types version to 0.47 (currently at 0.22) to match the library?

kaydelaney:
The package `slate` is at version 0.47, but `slate-react` is at 0.22 as far as I know, and the versions in the respective `index.d.ts` files reflect that I think unless I'm missing something?

jessetrinity:
Nope, I was looking at the title that says @types/slate." The file indeed says slate-react.

DinMon:
Hey just wanna check the type SlateNode which is derived from Node does not match with function getEventTransfer(event) return type which is an object containting { files, fragment, html, node, rich, text }. Please could you check if the type SlateNode should be the returning data type of getEventTransfer?

PR #37089, opened 2019-07-23
Add options argument to fs-extra.readdir

sukima:
This is documented here: https://nodejs.org/api/fs.html#fs_fs_readdir_path_options_callback

Please fill in this template.

### DIFF
--- a/types/fs-extra/index.d.ts
+++ b/types/fs-extra/index.d.ts
@@ -14,6 +14,7 @@
 
 import * as fs from "fs";
 import Stats = fs.Stats;
+import Dirent = fs.Dirent;
 
 export * from "fs";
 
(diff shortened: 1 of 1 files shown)


sukima:
Look like these changes are Node version `10.10.x`

ffflorian:
Shouldn't these be exported with the line
```ts
export * from 'fs'
```
?

sukima:
@ffflorian I think fs-extra wraps fs functions in a promise wrapper hence why they need to define their own? TBH I have no idea how this stuff actually works. I just got tired of my programs breaking because someone forgot to allow options in this file.

ffflorian:
> I must be the worlds worst TS programmer!

Don't worry, you're not! 🙂 I can have a better look at it tomorrow.

Maybe in the meantime @alan-agius4 @midknight41 @shiftkey @mees- @jrockwood @sangdth can chime in. 🔔

shiftkey:
@sukima 

> Per CI I screwed the whole thing up. 

```
ERROR: 198:48  unified-signatures  This overload and the one on line 197 can be combined into one signature with an optional parameter.
```

I think this CI test is suggesting that these two APIs:

```ts
export function readdir(path: string | Buffer): Promise<string[]>;
export function readdir(path: string | Buffer, options: { encoding?: string, withFileTypes?: false }): Promise<string[]>;
```

Could be merged into one, because they share the same function name and return type and only differ by the second parameter being options.

I think this is what it's asking for:

```ts
export function readdir(path: string | Buffer, options?: { encoding?: string, withFileTypes?: false }): Promise<string[]>;
```

Also, are you able to add some tests to [`types/fs-extra/fs-extra-tests.ts`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/fs-extra/fs-extra-tests.ts) for these new functions to assert they aren't regressed in the future?

sukima:
@shiftkey First, thank you

> Could be merged into one, because they share the same function name and return type and only differ by the second parameter being options.

I was only copying what other lines in this file were doing. I am curious why the rest of the file can get away with this but I cannot.

> Also, are you able to add some tests to `types/fs-extra/fs-extra-tests.ts` for these new functions to assert they aren't regressed in the future?

I have no idea how. Heck for that matter I don't even know how to TypeScript. This stuff doesn't make sense to me. I just wanted to use `fs-extra` in my TS project. Sigh. This is really hard, is TS supposed to be this hard?

If I have some free time I'll maybe look into it more. But if there is anybody anyone who knows more then me (which is a good bet) please go and change/do whatever ;cause I would not trust me to get things done with this PR.

ffflorian:
> is TS supposed to be this hard?

No, but you are digging really deep here. :smiley: 

Just delete line 197 and the tests should be good.

PR #37094, opened 2019-07-24
redux-first-router: Make individual Query params optional

adam1658:
URL Query parameters are naturally optional - a user could navigate to one of your URLs without providing any parameters, or different parameters to what you expect. eg: you expect `query` to be `{a?: string, b?: string}`, and the user navigates to `/url?a=test` (b is missing).

My route actions include a query payload of type `{a?: string, b?: string}` to reflect this reality, but this is not assignable to redux-first-router's `interface Query = {[key: string]: string}` (added in #35500).
I'd like to change the definition to `interface Query = {[key: string]: string | undefined}` to support this use case.

Please fill in this template.

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: <<url here>> **See description above**

### DIFF
--- a/types/redux-first-router-restore-scroll/index.d.ts
+++ b/types/redux-first-router-restore-scroll/index.d.ts
@@ -2,7 +2,7 @@
 // Project: https://github.com/faceyspacey/redux-first-router-restore-scroll#readme
 // Definitions by: icopp <https://github.com/icopp>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.4
+// TypeScript Version: 2.8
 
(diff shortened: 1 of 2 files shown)


adam1658:
I've updated the supported TS version for redux-first-router to 2.8, as even before my changes it didn't test or lint on anything lower. This indicates that it hasn't worked on 2.7 or below for a while - therefore the supported TS version was incorrect, which I've rectified.

redux-first-router-restore-scroll also has had it's supported TS version updated, as it depends on redux-first router and cannot be behind in terms of TS support.

Tests and linting runs fine now

icopp (approved):

PR #37098, opened 2019-07-24
Update README.md

fa93hws:

Update the readme only.


neilbryson:
Isn't this only TS 3.4+ syntax ?

fa93hws:
> Isn't this only TS 3.4+ syntax ?

Oh right... The backward compatibility is important here...

PR #37104, opened 2019-07-24
feat(toposort): second element of array can be undefined

prokopsimek:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/marcelklehr/toposort/blob/master/index.js#L83 (It doesn't matter if the edge is empty string or undefined. Toposort returns empty string even the undefined in the tree so you have to wrap it e.g. in `_.compact(...)` to filter out undefined and it's much worse filter out empty string.)

### DIFF
--- a/types/toposort/index.d.ts
+++ b/types/toposort/index.d.ts
@@ -1,7 +1,8 @@
 // Type definitions for toposort 2.0
 // Project: https://github.com/marcelklehr/toposort
 // Definitions by: Daniel Byrne <https://github.com/danwbyrne>
+//                 Prokop Simek <https://github.com/prokopsimek>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-declare function toposort(graph: ReadonlyArray<[string, string]>): ReadonlyArray<string>;
+declare function toposort(graph: Array<[string, string | undefined]>): string[];
 export = toposort;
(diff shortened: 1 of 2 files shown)


prokopsimek:
@danwbyrne We may consider changing the params interface to `[any, any]`. In the documentation of toposort is written:
> vertices needn't be strings but can be of any type

src: https://github.com/marcelklehr/toposort#toposortedges

But I am afraid of its behavior after that change.

danwbyrne:
Sorry I never looked at this PR, is there a reason you removed the `readonly` arrays? @prokopsimek

prokopsimek:
@danwbyrne Hi :) I have an array of objects and I did a `.map` on a property of an object, which isn't by default `readonly` and casting to `readonly` didn't make sense for me. It was the reason. Of course I could make it, but it's more code for nothing.

forivall:
When a function's parameters are `readonly`, there's no requirement to cast your array to readonly, it just indicates that the function isn't going to modify the array.

prokopsimek:
@forivall I'm not sure... I wanted to modify the result from the function and typescript didn't allow it? Is it possible, when the result is `readonly`? I can't remember what problem it was.

forivall:
I'm just talking about the parameters being readonly. Ah. I'm now looking at the older version - it returns a ReadonlyArray. I'm about to submit a PR where only the parameters are `readonly`, but the result is mutable.

prokopsimek:
It's related to #37510

forivall:
ah, I didn't see that one. My PR is #37511 37511

PR #37114, opened 2019-07-24
Update index.d.ts

vinnik-dmitry07:
An example where it works wrong:
                    shape: {
                        type: 'poly',
                        coords: [1, 1, 1, 20, 18, 20, 18, 1]
                    }
PyCharm underlines it as wrong, likely because of strongly seted 4 elements.

### DIFF
--- a/types/googlemaps/index.d.ts
+++ b/types/googlemaps/index.d.ts
@@ -995,13 +995,6 @@ declare namespace google.maps {
         text: string;
     }
 
-    interface MarkerShapePolyCoords extends Array<number> {
-        0: number;
-        1: number;
-        2: number;
-        3: number;
-    }
-
     interface MarkerShapeCircle {
         type: 'circle';
         /**
@@ -1027,7 +1020,7 @@ declare namespace google.maps {
          * Coords is **[x1,y1,x2,y2...xn,yn]** where each x,y pair contains the
          * coordinates of one vertex of the polygon.
          */
-        coords: MarkerShapePolyCoords;
+        coords: Array<number>;
     }
 
     /**


demensky (changes requested):
Please provide an error. I do not have errors in PyCharm.
image

life777:
Hi @3Pi0FomTatuin

MarkerShapePolyCoords interface extends Array<number> it means that 'coords' property is an simple array of numbers. It can have any length. But For polygon shape there should be at least 4 coordinates and that's why MarkerShapePolyCoords has second part that says that 'coords' property should be an array of numbers with at least 4 numbers.

Screenshot from 2019-07-25 04-15-05

3Pi0FomTatuin:
> Please provide an error. I do not have errors in PyCharm.
> image



> Hi @3Pi0FomTatuin
> 
> MarkerShapePolyCoords interface extends Array it means that 'coords' property is an simple array of numbers. It can have any length. But For polygon shape there should be at least 4 coordinates and that's why MarkerShapePolyCoords has second part that says that 'coords' property should be an array of numbers with at least 4 numbers.
> 
> Screenshot from 2019-07-25 04-15-05

Probably it happens because of PyCharms Type/Java-Script compatibility (I use js).
With shape:
2019-07-25 15_07_42-nefar  D__nefar  -  _activities_static_activities_js_engine js  nefar  - PyCha
Without:
2019-07-25 15_08_17-nefar  D__nefar  -  _activities_static_activities_js_engine js  nefar  - PyCha

demensky:
@3Pi0FomTatuin

demensky:
@3Pi0FomTatuin 
There are no problems with JavaScript either.

image

What config do you use for `tsconfig.json`?

3Pi0FomTatuin:
Sorry me, I updated PyCharm from 2019.1.2 to 2019.2 and It have solved the problem.

demensky:
@3Pi0FomTatuin, please close the PR.

PR #37115, opened 2019-07-24
Fixed minor missed changes

brycematheson1234:
Sorry missed two minor things in my last PR:
- Got the typing for `urls` slightly off
- SwaggerRequest wasn't exported when it should be as it's a param for a request interceptor. 

Fixed both up in this PR - double-checked it is correct in our environment and it looks good 👍

### DIFF
--- a/types/swagger-ui-dist/index.d.ts
+++ b/types/swagger-ui-dist/index.d.ts
@@ -27,7 +27,7 @@ interface SwaggerUIBundle {
     presets?: any[];
     plugins?: any;
     spec?: any;
-    url?: Url;
+    url?: string;
     urls?: Url[];
     layout?: string;
   }): any;
--- a/types/swagger-ui-dist/index.d.ts
+++ b/types/swagger-ui-dist/index.d.ts
@@ -20,15 +20,19 @@ interface Url {
 
 interface SwaggerUIBundle {
   (a: {
-    deepLinking: boolean;
-    dom_id: string;
-    presets: any[];
-    plugins: any;
-    urls: Url[];
-    layout: string;
+    configUrl?: string;
+    deepLinking?: boolean;
+    dom_id?: string;
+    domNode?: string;
+    presets?: any[];
+    plugins?: any;
+    spec?: any;
+    url?: Url;
+    urls?: Url[];
+    layout?: string;
   }): any;
-  presets: any;
-  plugins: any;
+  presets?: any;
+  plugins?: any;
   [k: string]: any;
 
   getConfigs(): SwaggerConfigs;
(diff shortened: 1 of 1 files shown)


types/swagger-ui-dist/index.d.ts
```diff
@@ -13,13 +13,18 @@ export as namespace SwaggerUIDist;
...
+}
+
 interface SwaggerUIBundle {
   (a: {
     deepLinking: boolean;
     dom_id: string;
     presets: any[];
     plugins: any;
-    urls: Array<[string, string]>;
+    urls: Url[];
```
haowen737: Didn't notice the document before🤔

Parameter name | Docker variable | Description
-- | -- | --
configUrl | CONFIG_URL | String. URL to fetch external configuration document from.
dom_id | DOM_ID | String, REQUIRED if domNode is not provided. The ID of a DOM element inside which SwaggerUI will put its user interface.
domNode | Unavailable | Element, REQUIRED if dom_id is not provided. The HTML DOM element inside which SwaggerUI will put its user interface. Overrides dom_id.


It seems we can do a little more on this commit:
most parameters are optional actually, (except `domNode` and `dom_id`, but we can still make it optional though),

`deepLinking?: boolean;
    dom_id?: string;	    dom_id?: string;...` should be fine

Thanks for this fix anyway, I was just about to make the same commit😁

haowen737 (changes requested):

types/swagger-ui-dist/index.d.ts
brycematheson1234: I didn't notice that document before either :/ 

I just added the parameters from there and made them all optional: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37115/commits/aacfe09ae0a57b5d151f075d862c3fa8104e364c

I left `domNode` and `dom_id` optional as well, because they are `REQUIRED if...` and that depends on a condition, so they're not technically "required" in the traditional sense.

types/swagger-ui-dist/index.d.ts
```diff
@@ -20,15 +20,19 @@ interface Url {
...
-    urls: Url[];
-    layout: string;
+    configUrl?: string;
+    deepLinking?: boolean;
+    dom_id?: string;
+    domNode?: string;
+    presets?: any[];
+    plugins?: any;
+    spec?: any;
+    url?: Url;
```
haowen737: type `url` should be a `string` by documentation

url | String. The URL pointing to API definition (normally swagger.json or swagger.yaml). Will be ignored if urls or spec is used.
-- | --

brycematheson1234: Thanks great catch 👍 Fixed here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37115/commits/f2ea4bfc2dbbb46ed784255eb961d3ecf1f416b8

haowen737 (approved):

PR #37119, opened 2019-07-25
css-tree: WalkContext.atrule can be null

utkarshkukreti:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/csstree/csstree/blob/c5470408616d95a1926031ac0fc413785ea8faf2/lib/walker/create.js#L168

Note: While trying to commit, I got this error from a pre commit hook. I'm not sure what the problem is but I don't think my commit introduced this error.

```
husky > pre-commit (node v11.13.0)
✔  precise-commits: 2 modified file(s) found
✔  [1/2] Processing file: types/css-tree/css-tree-tests.ts
✖  precise-commits: An Error occurred

{ SyntaxError: ';' expected. (1:25)
> 1 | enter(node, item, list) {
    |                         ^
  2 |         this.stylesheet; // $ExpectType StyleSheet
  3 |         node; // $ExpectType CssNode
  4 |         item; // $ExpectType ListItem<CssNode>
```

Please let me know if I missed anything.

### DIFF
--- a/types/css-tree/index.d.ts
+++ b/types/css-tree/index.d.ts
@@ -517,14 +517,14 @@ export function generate(ast: CssNode, options?: GenerateOptions): string;
 
 export interface WalkContext {
     root: CssNode;
-    stylesheet: StyleSheet;
+    stylesheet: StyleSheet | null;
     atrule: Atrule | null;
-    atrulePrelude: AtrulePrelude;
-    rule: Rule;
-    selector: SelectorList;
-    block: Block;
-    declaration: Declaration;
-    function: FunctionNode | PseudoClassSelector | PseudoElementSelector;
+    atrulePrelude: AtrulePrelude | null;
+    rule: Rule | null;
+    selector: SelectorList | null;
+    block: Block | null;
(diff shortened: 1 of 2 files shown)


types/css-tree/index.d.ts
```diff
@@ -518,7 +518,7 @@ export function generate(ast: CssNode, options?: GenerateOptions): string;
 export interface WalkContext {
     root: CssNode;
     stylesheet: StyleSheet;
-    atrule: Atrule;
+    atrule: Atrule | null;
```
erik-kallen: Seems to me that all those members can be null, except for root. In that case, we should change all of them

erik-kallen (changes requested):

types/css-tree/index.d.ts
utkarshkukreti: Fixed, thanks for the suggestion!

erik-kallen:
LGTM

erik-kallen (approved):

PR #37132, opened 2019-07-25
Make possible to use Child Worlds

albertossilva:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/cucumber/cucumber-tests.ts
+++ b/types/cucumber/cucumber-tests.ts
@@ -1,5 +1,5 @@
 import * as assert from "power-assert";
-import { setDefinitionFunctionWrapper, setWorldConstructor, defineParameterType, After, AfterAll, Before, BeforeAll, Given, When, Then } from "cucumber";
+import { setDefinitionFunctionWrapper, setWorldConstructor, defineParameterType, After, AfterAll, Before, BeforeAll, Given, When, Then, World } from "cucumber";
 import cucumber = require("cucumber");
 
 type Callback = cucumber.CallbackStepDefinition;
@@ -7,15 +7,13 @@ type Table = cucumber.TableDefinition;
 type HookScenarioResult = cucumber.HookScenarioResult;
 const Status = cucumber.Status;
 
-// You can optionally declare your own world properties
-declare module "cucumber" {
-    interface World {
-        visit(url: string, callback: CallbackStepDefinition): void;
-        toInt(value: string): number;
-    }
+class CustomWorld {
+    myCustomFunction() {}
 }
 
 function StepSampleWithoutDefineSupportCode() {
+    setWorldConstructor(CustomWorld)
+
     setWorldConstructor(function({attach, parameters}) {
         this.attach = attach;
         this.parameters = parameters;
@@ -151,6 +149,14 @@ function StepSampleWithoutDefineSupportCode() {
         const actual: Array<{ [colName: string]: string }> = table.hashes();
         assert.deepEqual(actual, expected);
     });
+
+    Given('Common world constructor', function () {
+        assert.equal(typeof this.attach, 'function')
+    })
+
+    Given<CustomWorld>('Common world constructor', function () {
+        assert.equal(typeof this.myCustomFunction, 'function')
+    })
 
     defineParameterType({
         regexp: /particular/,
--- a/types/cucumber/cucumber-tests.ts
+++ b/types/cucumber/cucumber-tests.ts
@@ -12,7 +12,7 @@ class CustomWorld {
 }
 
 function StepSampleWithoutDefineSupportCode() {
-    setWorldConstructor(CustomWorld)
+    setWorldConstructor(CustomWorld);
 
     setWorldConstructor(function({attach, parameters}) {
         this.attach = attach;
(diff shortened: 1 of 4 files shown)


albertossilva:
@DonJayamanne, can you review please, just mentioned here as suggested by the README.md

jessetrinity:
Pinging owners. @AbraaoAlves @jan-molak @isoung @BendingBender @ErikSchierboom @petermorlion @DonJayamanne 

Any interest? Thanks!

albertossilva:
Thanks @jessetrinity

sandersn (changes requested):
The code change looks good, but can you add some tests? I'd like to see how the new type parameter gets inferred from usage.

albertossilva:
@sandersn After struggling with how to import `node` type and lint... finally the PR is ready.
So I created for the world following: https://github.com/cucumber/cucumber-js/blob/master/docs/support_files/world.md.

And I just realized that World can be any type.

Now... tests add thank for the review.

types/cucumber/index.d.ts
```diff
@@ -42,46 +55,46 @@ export interface TableDefinition {
     hashes(): Array<{ [colName: string]: string }>;
 }
 
-export type StepDefinitionCode = (this: World, ...stepArgs: any[]) => any;
+export type StepDefinitionCode<T = World> = (this: T, ...stepArgs: any[]) => any;
```
sandersn:
```suggestion
export type StepDefinitionCode = (this: World, ...stepArgs: any[]) => any;
```

types/cucumber/cucumber-tests.ts
```diff
@@ -152,6 +150,14 @@ function StepSampleWithoutDefineSupportCode() {
         assert.deepEqual(actual, expected);
     });
 
+    Given('Common world constructor', function() {
```
sandersn: in this case it looks like there is no inference source for `T` and it'll just default to `World`, right? If there's no usage of it afterward (like in the return type) then you can accomplish the same thing by using a `this: World` instead of `this: T` in `Given`. Then for the custom case, you just put `this: CustomWorld`.

I noticed that the no-unnecessary-generics rule is turned off in tslint.json for cucumber. If it were turned on it would have errored on this change.

types/cucumber/cucumber-tests.ts
```diff
@@ -152,6 +150,14 @@ function StepSampleWithoutDefineSupportCode() {
...
+        assert.equal(typeof this.attach, 'function');
+    });
+
+    Given<CustomWorld>('Common world constructor', function() {
```
sandersn:
```suggestion
    Given('Common world constructor', function(this: CustomWorld) {
```

types/cucumber/index.d.ts
sandersn: The undoes all (most) of the changes below.

sandersn (changes requested):
Also please remove the `no-unnecessary-changes: "false"` override in cucumber's tslint.json.

types/cucumber/cucumber-tests.ts
albertossilva: Done

types/cucumber/cucumber-tests.ts
albertossilva: Done

types/cucumber/cucumber-tests.ts
sandersn: You might still want to have the `this: CustomWorld` test to make sure it continues to work.

types/cucumber/index.d.ts
```diff
@@ -7,8 +7,11 @@
 //                 ErikSchierboom <https://github.com/ErikSchierboom>
 //                 Peter Morlion <https://github.com/petermorlion>
 //                 Don Jayamanne <https://github.com/DonJayamanne>
+//                 Alberto Silva <https://github.com/albertossilva>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.4
+// TypeScript Version: 3.2
```
sandersn: Once you get rid of the default type parameter on setWorldConstructor, I think you can set this back to 2.4.

types/cucumber/index.d.ts
```diff
@@ -77,7 +90,7 @@ export function Given(pattern: RegExp | string, options: StepDefinitionOptions,
 export function setDefaultTimeout(time: number): void;
 export function setDefinitionFunctionWrapper(fn: ((fn: () => void) => (...args: any[]) => any) | ((fn: () => void, options?: {[key: string]: any}) => (...args: any[]) => any)): void;
 // tslint:disable-next-line ban-types
-export function setWorldConstructor(world: ((this: World, init: {attach: Function, parameters: {[key: string]: any}}) => void) | {}): void;
+export function setWorldConstructor<T = World>(world: ((this: T, init: {attach: AttachmentFunction, parameters: WorldParameters}) => void) | { new (): T }): void;
```
sandersn: `T` should be removed in favour of `World` here too, just like all the others, because the intent of the type is (as far as I can tell) to allow (1) an init function from a World object or (2) a World constructor function. But `T` isn't used aside from that.

sandersn (changes requested):

DonJayamanne:
@albertossilva Sorry I missed this.

types/cucumber/index.d.ts
albertossilva: I think `setWorldConstructor` accepts either a `constructor function` or a class.
but... anyway what should be the signature in the case you mentioned?

albertossilva:
That is ok @DonJayamane.

types/cucumber/index.d.ts
sandersn: to Typescript, a class `World` is just an object with `new(): World`, so you can pass either a constructor function or a class.

Here is what I would suggest:

```suggestion
export function setWorldConstructor(world: ((this: World, init: {attach: AttachmentFunction, parameters: WorldParameters}) => void) | { new (): World }): void;
```

This works the same because `CustomWorld` is a subtype of `World`, so it's fine to pass to `setWorldConstructor`. `setWorldConstructor` only cares that it receives *some* `World`, so it doesn't need `T` to capture a particular subtype of it.

PR #37135, opened 2019-07-25
Make ResizeObserver global as it's a browser API

wffurr:
Following the guidance in the TypeScript handbook for declaration files for global APIs.  Instead of pretending that ResizeObserver is a module, declare it as a global and also export it on the window.  Ideally this would be in the DOM typings, but it's still in draft and only supported by Chrome.

I haven't tested because "npm run test" appears to be broken at head.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-d-ts.html


wffurr:
Not sure where else to post this, but I made a SO question about the out of date types-publisher module that I'm getting: https://stackoverflow.com/questions/57209407/npm-install-is-downloading-old-version-of-types-publisher-from-github

types/resize-observer-browser/index.d.ts
```diff
@@ -1,25 +1,23 @@
 // Type definitions for non-npm package resize-observer-browser 0.1
 // Project: https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver, https://developers.google.com/web/updates/2016/10/resizeobserver, https://wicg.github.io/ResizeObserver/
-// Definitions by: Chives <https://github.com/chivesrs>
+// Definitions by: Chives <https://github.com/chivesrs>, William Furr <https://github.com/wffurr>
```
chivesrs: Could you use the multline syntax here please?
https://github.com/DefinitelyTyped/DefinitelyTyped#edit-an-existing-package

types/resize-observer-browser/test/resize-observer-browser-module.tests.ts
```diff
@@ -1,5 +1,3 @@
-import { ResizeObserver } from "resize-observer-browser";
```
chivesrs: It looks like we don't need a module test at all, given there is no import here.
Let's rename this file to `resize-observer-browser-tests.ts` (as per https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package), and remove the global tests file.

Then we should merge in the `window.ResizeObserver` test into here.

types/resize-observer-browser/test/resize-observer-browser-module.tests.ts
```diff
@@ -1,5 +1,3 @@
...
-
 function resizeObserverCreates(): void {
     const resizeObserver: ResizeObserver = new ResizeObserver((entries) => {
```
chivesrs: We should add a test in the scenario where the callback is complex and so is its own function.

Something like

```
function callback(entries: TheType): void {
  // no-op
}

const resizeObserver = new ResizeObserver(callback);
```

This way we know `TheType` is usable without needing to be imported.

chivesrs (changes requested):

chivesrs:
The lint command seems to work for me, I'm using `npm run lint resize-observer-browser` from the root directory (the one which contains `README.md`).

`npm test` seems to not work for me, but I've found the lint command does a much better job anyway.

types/resize-observer-browser/test/resize-observer-browser-module.tests.ts
chivesrs: Oh, I think the file should be moved out of the `tests/` subdirectory as well if we do this.

chivesrs (changes requested):

types/resize-observer-browser/index.d.ts
```diff
@@ -1,25 +1,23 @@
...
-declare global {
-    interface Window {
-        ResizeObserver: typeof ResizeObserver;
-    }
+interface Window {
+    ResizeObserver: typeof ResizeObserver;
 }
 
-export class ResizeObserver {
+declare class ResizeObserver {
```
chivesrs: Oh, `.d.ts` files implicitly add declare before every single item. Do you need this here? Or does everything work without the declare keyword?

chivesrs (changes requested):

types/resize-observer-browser/index.d.ts
wffurr: If I remove "declare" from the class declaration, I get the following error:

index.d.ts:11:1 - error TS1046: A 'declare' modifier is required for a top level declaration in a .d.ts file.

wffurr: All of the handbook examples include declare on namespace, class, and var.

types/resize-observer-browser/index.d.ts
wffurr: Done.

types/resize-observer-browser/test/resize-observer-browser-module.tests.ts
wffurr: Done.

types/resize-observer-browser/test/resize-observer-browser-module.tests.ts
wffurr: Done.

types/resize-observer-browser/index.d.ts
chivesrs: Huh, ok. I thought `.d.ts` implied declare automatically. Guess I was wrong.

types/resize-observer-browser/resize-observer-browser-tests.ts
```diff
@@ -0,0 +1,23 @@
+function resizeObserverCreates(): void {
+    const resizeObserver: ResizeObserver = new ResizeObserver(entries => {
```
chivesrs: nit: could you make the parameter `(entries)` like it was before?

chivesrs (approved):

types/resize-observer-browser/resize-observer-browser-tests.ts
wffurr: Done.

I wish github would show this as a move.  `git show` shows it as a rename + changes.

wffurr: Now it shows the move + diff!  I don't know what I did.

PR #37139, opened 2019-07-25
[c3] Update all c3 types to meet latest documentation and source code

iansan5653:
This involves significant changes across the entire type definition for the [c3.js](https://c3js.org/) library. I've updated all public interfaces to match both the documentation and the source code (where they conflict, I checked the source code myself). I also updated JSDocs for almost every definition, fixing formatting and adding tags where appropriate. Finally, anywhere possible I have updated to use newer and more specific TypeScript features.


types/c3/index.d.ts
```diff
@@ -1,26 +1,57 @@
...
-export type FormatFunction = (v: any, id: string, i: number, j: number) => void;
+export function generate(config: ChartConfiguration): ChartAPI;
 
-export interface TargetIds {
-    ids: ArrayOrString;
-}
+export const version: string;
 
+export type Primitive = string | boolean | number | null;
+export type PrimitiveArray = Array<string | boolean | number | null>;
```
denyo: Should use `Primitive` from the previous line.

types/c3/index.d.ts
```diff
@@ -343,171 +439,247 @@ export interface Data {
...
+    /**
+     * Convert data IDs with this function before creating chart.
+     * @param id The original ID string.
+     * @returns The converted ID string.
+     */
+    idConverter?(id: string): string;
 }
 
-export interface Axis {
+export interface AxesOptions {
```
denyo: Typo? `Axis` with an `i`.

iansan5653: That was intentional; "axes" is plural of "axis", so `AxesOptions` contains configuration for each axis. It helps clarify the difference between `AxesOptions` and `AxisConfiguration`.

types/c3/index.d.ts
iansan5653: Fixed with the latest commits.

iansan5653:
Can anyone explain what "Subtype cache size" is in @typescript-bot's performance metrics?

types/c3/index.d.ts
```diff
@@ -1,26 +1,57 @@
...
 // Definitions by: Marc Climent <https://github.com/mcliment>
 //                 Gerin Jacob <https://github.com/gerinjacob>
 //                 Bernd Hacker <https://github.com/denyo>
 //                 Dzmitry Shyndzin <https://github.com/dmitryshindin>
 //                 Tim Niemueller <https://github.com/timn>
 //                 Nate Mara <https://github.com/natemara>
+//                 Ian Sanders <https://github.com/iansan5653>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+// TypeScript Version: 3.5
```
jessetrinity: Do you intend for this to only work in TypeScript 3.5+?

jessetrinity:
@denyo are you waiting on any changes?

denyo:
> @denyo are you waiting on any changes?

No, honestly I haven't been using these typings in over a year and since its basically a rewrite it's too time consuming to check it with any docs or source code. I guess releasing them as a new major version and waiting for feedback is the way to go.

jessetrinity:
@iansan5653 do you know how to [update the major version](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package)? (I am wondering if you intended to do this instead of updating the typescript version).

types/c3/index.d.ts
iansan5653: That should be 3.0, because I used `unknown`. Thanks for the heads up

iansan5653:
@jessetrinity The latest version of c3 is 0.7.3, which this has been updated to match; should I exceed that in these type declarations?

jessetrinity:
@iansan5653 no, you only need to match the major.minor version, so 0.7 should be fine if it matches the library. 

Maybe I misinterpreted what @denyo meant - are you recommending the old types be moved to another folder?

denyo:
What I meant is that the latest npm package has a version of `0.7.1` with 22k weekly downloads. Since this PR isn't backwards compatible it would break those installations if they update to the next version in case it is going to be released as `0.7.x`.

iansan5653:
@denyo That's true, but the current version is not compatible with the current version of c3, so those installations need to be fixed anyway. I think the c3 library is not using semver properly with recent updates and so breaking changes to the library were published with only patch-level version increases. Not sure what the protocol is to handle that.

sandersn:
From reading the discussion so far it sounds like

1. The current types are out of date for 0.7.3, and this PR brings them up to date.
2. c3 has released breaking changes on patch versions, meaning that there's no good way to match the version without breaking people who depend on these types.
3. But those people were likely broken a lot anyway since the current types don't match 0.7.3 in the first place.
4. The best thing is probably just to ship this change and watch for feedback.

Sound accurate @denyo @iansan5653 ? If so then I'll merge after I hear back from you.

denyo:
@sandersn that sums it up.
I guess most people where casting props to `any` when the typings were wrong anyways. 
So it shouldn't be a big deal.

iansan5653:
Sounds right to me as well.

PR #37145, opened 2019-07-26
Update mapbox-gl to v0.53.0

LiamAttClarke:
https://github.com/mapbox/mapbox-gl-js/releases/tag/v0.52.0
https://github.com/mapbox/mapbox-gl-js/releases/tag/v0.53.0

### DIFF
--- a/types/mapbox-gl/index.d.ts
+++ b/types/mapbox-gl/index.d.ts
@@ -1,3 +1,5 @@
+import { Feature } from "geojson";
+
 // Type definitions for Mapbox GL JS v0.53.0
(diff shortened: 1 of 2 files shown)


dmytro-gokun (changes requested):
1. Please add tests for the newly added methods/properties
2.  I think it makes sense to create an interface named FeatureIdentifier and use that as a type for getFeatureState, removeFeatureState, setFeatureState + may be some other places. There is no need to repeat that object definition again & again.

dmytro-gokun (approved):

PR #37148, opened 2019-07-26
[@types/braintree] Added validation error types.

MatejSkrbis:
The error object does not seems to be list of strings. I've updated types accordingly.

Please fill in this template.

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/braintree/braintree_node/blob/master/lib/braintree/validation_errors_collection.js

### DIFF
--- a/types/braintree/index.d.ts
+++ b/types/braintree/index.d.ts
@@ -50,7 +50,7 @@ declare namespace braintree {
 
     interface ValidatedResponse<T> {
         success: boolean;
-        errors: () => string[];
+        errors: ValidationErrorsCollection;
         message: string;
         params: Record<string, any>;
         address: T extends Address ? Address : never;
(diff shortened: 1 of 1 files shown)


types/braintree/index.d.ts
```diff
@@ -1685,4 +1685,31 @@ declare namespace braintree {
...
+
+    export interface ErrorCollectionsObject {
+        [key: string]: ValidationErrorsCollection;
+    }
+
+    export interface ValidationErrorsObject {
+        [key: string]: ValidationError[];
+    }
+
+    export interface ValidationErrorsCollection {
```
smrubin: Could add `length` to the interface as well for this PR.
```
length: number
```

types/braintree/index.d.ts
```diff
@@ -1685,4 +1685,31 @@ declare namespace braintree {
...
+        validationErrors: ValidationErrorsObject;
```
smrubin: @MatejSkrbis Where do `validationErrors` and `errorCollections` come into play? Couldn't find them from the docs  and was mostly referencing this page for the PR: https://developers.braintreepayments.com/reference/general/validation-errors/overview/node

smrubin (commented):
@MatejSkrbis Thanks so much for building these out. Had a question on two of the props and it'd be awesome to add length as well. Then the PR should be good to go.

types/braintree/index.d.ts
MatejSkrbis: I've tried to execute code so that I got those validation errors and they had no length even if documentation said so. Looking at the source code I don't see how as this is object and not an array. https://github.com/braintree/braintree_node/blob/master/lib/braintree/validation_errors_collection.js

types/braintree/index.d.ts
MatejSkrbis: You've got the point. They should probably be private as they are accessed with the help of `for` and `on` methods. I've removed them.

types/braintree/index.d.ts
smrubin: Thanks for doing some sleuthing and following up. I agree.

smrubin (approved):

PR #37150, opened 2019-07-26
Add (missing) inputContainerStyle props

nazrdogan:
Add  inputContainerStyle


mchappell (approved):
Added style used inputContainerStyle props from react-native-material-textfield property **inputContainerStyle** https://github.com/n4kz/react-native-material-textfield#properties

PR #37156, opened 2019-07-26
[react-native-incall-manager] New types definition

KarlosQ:
Please fill in this template.

If adding a new definition:
- [x If this is for an NPM package, match the name. If not, do not conflict with the name of an NPM package.

### DIFF
--- a/types/react-native-incall-manager/index.d.ts
+++ b/types/react-native-incall-manager/index.d.ts
@@ -15,50 +15,50 @@ export interface StopSetup {
 }
 
 export default class InCallManager {
-    static start(setup?: StartSetup): void;
+    start(setup?: StartSetup): void;
 
-    static stop(setup?: StopSetup): void;
+    stop(setup?: StopSetup): void;
 
-    static turnScreenOff(): void;
+    turnScreenOff(): void;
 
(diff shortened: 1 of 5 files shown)


types/react-native-incall-manager/index.d.ts
```diff
@@ -0,0 +1,64 @@
...
+    media?: string;
+    auto?: boolean;
+    ringback?: string;
+}
+
+export interface StopSetup {
+    busytone?: string;
+}
+
+export default class InCallManager {
```
weswigham: [Looks like the default is an _instance_ of the class, not the class itself.](https://github.com/zxcpoiu/react-native-incall-manager/blob/master/index.js#L187) - i'd remove the `export default` from this class, remove the `static` modifier from all the methods, then add
```ts
declare const _default: InCallManager;
export default _default;
```

weswigham (changes requested):

types/react-native-incall-manager/index.d.ts
KarlosQ: Thanks you for the info 👍

weswigham (approved):

PR #37176, opened 2019-07-27
[@mapbox/mapbox-sdk] Added types for mapbox-sdk-js

jeffbdye:
Hey, thanks for checking out this PR. I'm looking to add types for the mapbox javascript SDK here: https://github.com/mapbox/mapbox-sdk-js

The types didn't generate particularly from `dts-gen`, however the jsdoc comments helped in straightening out types originally presented by @mikeomeara1.

This is my first PR for type definitions, so please let me know if there's anything amiss.

If adding a new definition:

### DIFF
--- a/types/mapbox__mapbox-sdk/index.d.ts
+++ b/types/mapbox__mapbox-sdk/index.d.ts
@@ -8,11 +8,11 @@
 declare module '@mapbox/mapbox-sdk/lib/classes/mapi-client' {
     import { MapiRequest, DirectionsApproach } from '@mapbox/mapbox-sdk/lib/classes/mapi-request';
     export default class MapiClient {
-                       constructor(config: SdkConfig);
-                       accessToken: string;
-                       origin?: string;
-                       createRequest(requestOptions: any): MapiRequest;
-                   }
+        constructor(config: SdkConfig);
+        accessToken: string;
+        origin?: string;
+        createRequest(requestOptions: any): MapiRequest;
+    }
 
     interface SdkConfig {
         accessToken: string;
@@ -1219,7 +1219,9 @@ declare module '@mapbox/mapbox-sdk/services/tokens' {
     export default function Tokens(config: SdkConfig | MapiClient): TokensService;
 
     interface TokensService {
-        /**List your access tokens. */
+        /**
+         * List your access tokens.
+         */
         listTokens(): MapiRequest;
         /**
          * Create a new access token.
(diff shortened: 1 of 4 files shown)


jessetrinity:
Hi @jeffbdye, thanks for your PR! Any reason you chose to include all of your modules in one file rather than breaking it up?

jeffbdye:
Hey @jessetrinity, no particular reason for having all of the modules in one file. I'm guessing I should have separate declaration files in directories matching the mapbox-sdk library structure, like what's documented here? https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#library-file-layout

So, splitting out this single `index.d.ts` file into into files such as 'mapbox__mapbox-sdk/lib/classes/mapi-client/index.d.ts` and `mapbox__mapbox-sdk/services/directions/index.d.ts` with just the respective module in each?

jessetrinity:
@jeffbdye right, but looking at the related (?) mapbox types, it looks like they use the same convention as you. I'm fine merging as is if you are happy with it.

jeffbdye:
@jessetrinity good point, I'd say let's go ahead and merge it then. Thanks again for the review!

wcauchois:
Hi, thanks for these typings!

I was wondering, how am I supposed to `import` the client?

It seems that if I import it as `import * as mapboxGeocoding from '@mapbox/mapbox-sdk/services/geocoding';`, then I get the correct object at runtime, but I get a type error when I try to call mapboxGeocoding() - `Cannot invoke an expression whose type lacks a call signature.`

If I instead `import mapboxGeocoding from '@mapbox/mapbox-sdk/services/geocoding';` then it typechecks but fails at runtime -- mapboxGeocoding is `undefined` at runtime.

I have `"lib": ["es6", "dom", "esnext"], "module": "commonjs",` in my `tsconfig.json`.

Help would be appreciated!

jeffbdye:
Hey @wcauchois!

The second import statement you have should work correctly:
`import mapboxGeocoding from '@mapbox/mapbox-sdk/services/geocoding';`

Not sure I'm following how it'd be `undefined` there, I was able to get a geocode request successfully sent in a small demo project using the types, so be sure you've also installed the '@mapbox/mapbox-sdk' package correctly too.

Also, just to clarify compared to the docs over in mapbox-sdk-js: `mapboxGeocoding` is the default export and should be treated like what you'd get if you were to do `let mapboxGeocoding = require('@mapbox/mapbox-sdk/services/geocoding');`, so in order to get a client you'd need to use `mapboxGeocoding` as a clientFactory: `let geocodeService: GeocodeService = mapboxGeocoding(<pass in your sdkConfig or existing MapiClient>)`.

So, the full import statement you'd have after all that would look something like this:
`import mapboxGeocoding, { GeocodeService, GeocodeRequest, GeocodeResponse } from '@mapbox/mapbox-sdk/services/geocoding';`. 

The test file included in the types package here should also give some hints on the intended usage too. Hopefully that helps!

PR #37177, opened 2019-07-27
[yargs] choices accepts numbers

nikolay-borzov:
According to [docs](https://github.com/yargs/yargs/blob/master/docs/api.md#choiceskey-choices)

> If this method is called multiple times, all enumerated values will be merged together. Choices are generally strings or numbers, and value matching is case-sensitive.


JimiC:
The correct documentation is https://github.com/yargs/yargs/blob/v13.3.0/docs/api.md. Referencing the `master` branch is not accurate.

JimiC (approved):

PR #37182, opened 2019-07-27
[@types/node] Allow to extends Http2ServerResponse and MyHttp2ServerRequest

chyzwar:
Please fill in this template.

This makes constructor public for Http2ServerRequest and Http2ServerResponse. According to http2 docs, it is possible to extend these classes. 

@see https://nodejs.org/api/http2.html#http2_http2_createsecureserver_options_onrequesthandler

There will be a need for more changes in http2 to fully support custom req, res classes.

### DIFF
--- a/types/node/http2.d.ts
+++ b/types/node/http2.d.ts
@@ -583,7 +583,7 @@ declare module "http2" {
     }
 
     export class Http2ServerRequest extends stream.Readable {
-        private constructor();
+        public constructor();
 
         readonly aborted: boolean;
         readonly authority: string;
@@ -651,7 +651,7 @@ declare module "http2" {
     }
 
     export class Http2ServerResponse extends stream.Stream {
-        private constructor();
+        public constructor();
 
         addTrailers(trailers: OutgoingHttpHeaders): void;
         readonly connection: net.Socket | tls.TLSSocket;
--- a/types/node/http2.d.ts
+++ b/types/node/http2.d.ts
@@ -583,7 +583,7 @@ declare module "http2" {
     }
 
     export class Http2ServerRequest extends stream.Readable {
-        public constructor();
+        constructor();
 
         readonly aborted: boolean;
         readonly authority: string;
@@ -651,7 +651,7 @@ declare module "http2" {
     }
 
     export class Http2ServerResponse extends stream.Stream {
-        public constructor();
+        constructor();
 
         addTrailers(trailers: OutgoingHttpHeaders): void;
         readonly connection: net.Socket | tls.TLSSocket;
--- a/types/node/http2.d.ts
+++ b/types/node/http2.d.ts
@@ -583,7 +583,7 @@ declare module "http2" {
     }
 
     export class Http2ServerRequest extends stream.Readable {
-        constructor(stream: ServerHttp2Stream, headers: IncomingHttpHeaders, options: stream.ReadableOptions, rawHeaders: string);
+        constructor(stream: ServerHttp2Stream, headers: IncomingHttpHeaders, options: stream.ReadableOptions, rawHeaders: string[]);
 
         readonly aborted: boolean;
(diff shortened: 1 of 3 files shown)


types/node/http2.d.ts
```diff
@@ -583,7 +583,7 @@ declare module "http2" {
     }
 
     export class Http2ServerRequest extends stream.Readable {
-        private constructor();
+        constructor();
```
Flarna: The arguments are missing, see https://github.com/nodejs/node/blob/0481a7f3b883862cadf963314d23490adc6656a4/lib/internal/http2/compat.js#L274
As long as they were private it was not needed to specify the arguments as it was anyway not possible to call it but if you extend it you have to pass the correct args here.

types/node/http2.d.ts
```diff
@@ -651,7 +651,7 @@ declare module "http2" {
     }
 
     export class Http2ServerResponse extends stream.Stream {
-        private constructor();
+        constructor();
```
Flarna: The arguments are missing, see https://github.com/nodejs/node/blob/0481a7f3b883862cadf963314d23490adc6656a4/lib/internal/http2/compat.js#L428

Flarna (changes requested):
Please add also some tests. I expect the use case is to extend these classes and use it as option to `createServer()`. Would be nice to see this in a test to verify that complete usecase works as expected.

types/node/http2.d.ts
chyzwar: This is now done.

types/node/http2.d.ts
chyzwar: I added type for the first argument but second "options" I am not sure what should be used. It seems options are passed to superclass Stream. Stream class do not define specific constructor options ?

chyzwar:
I am will look into tests but I am not sure if these would complete without making Http2Server generic.

types/node/http2.d.ts
```diff
@@ -583,7 +583,7 @@ declare module "http2" {
     }
 
     export class Http2ServerRequest extends stream.Readable {
-        private constructor();
+        constructor(stream: ServerHttp2Stream, headers: string[], options: stream.ReadableOptions, rawHeaders: string);
```
Flarna: I think `headers` should be of type `IncomingHttpHeaders` and `rawHeaders` should be `string[]`. According to source `options` may be also `undefined`.

Flarna:
> I am will look into tests but I am not sure if these would complete without making Http2Server generic.

You could use the test in http as sample.
I agree that a full solution (correct typed events,...) would require a generic `Http2Server` but as far as I remember this is not really possible with typescript 2.1. At least I failed to do this for http; I think it was because of missing default types for generics (and maybe something else).

types/node/http2.d.ts
chyzwar: thanks, fixed.

types/node/http2.d.ts
chyzwar: it is bug in node? this constructor should be empty ?

chyzwar:
Thanks for review. Added tests for custom req, res classes. I added them as a separate section. I will try to make generic Http2Server next PR and If successful I will add tests for requestHandler.

types/node/http2.d.ts
Flarna: why empty? see https://github.com/nodejs/node/blob/0481a7f3b883862cadf963314d23490adc6656a4/lib/internal/http2/compat.js#L743 the first argument is a `ServerHttp2Stream`. I think the second arg can be omitted.

chyzwar: It is already like that. Thanks for confirming the second arg.

Flarna (approved):

chyzwar:
Thank you everyone.

PR #37185, opened 2019-07-28
[passport-local] Set general boolean as type for 'passReqToCallback'

bephrem1:
Resolve type error when setting 'passReqToCallback' to true when creating the local strategy.


sandersn (changes requested):
Those types look like they were explicitly set to `true` and `false` to prevent incorrect values being passed in. Can you add a test that shows why they are too specific? For example, do they prevent correct code from compiling?

PR #37196, opened 2019-07-29
[@types/mongodb] Mark `_id` as optional for `insert` and `insertMany` methods

### LINKED ISSUE #33740: Letting mongodb pick the _id is not possible anymore when typing collections.
- [x] I tried using the `@types/mongodb` package and had problems.
  - @CaselIT
  - @alanmarcell
  - @bitjson
  - @dante-101
  - @mcortesi
  - @EnricoPicci
  - @AJCStriker
  - @julien-c
  - @daprahamian
  - @denys-bushulyak
  - @BastienAr
  - @sindbach
  - @geraldinelemeur
  - @jishi
  - @various89
  - @angela-1
  - @lirbank
  - @hector7
  - @floric
  - @erikc5000
  - @Manc

Since [this PR](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29245)
`Collection.insert{,One,Many}` are typed with TSchema from `Collection<TSchema>`
same as the return type of the query methods. This makes it a bit hard to use
when you want mongodb to pick the _id of an inserted document for you:

Version with `_id` being part of the TSchema:
```js
import { Collection, MongoClient, ObjectId } from 'mongodb'

interface Person {
  _id: ObjectId
  name: string 
  age: number
}

MongoClient.connect().then(async client => {
  const db = client.db()
  const personCollection: Collection<Person> = db.collection('persons');
  
  const writeResult = await personCollection.insertOne({ name: "Foo Bar", age: 30 }) // type error, `_id` missing
  
  const person = await personCollection.findOne({ _id: writeResult.insertedId })
  
  if (person) {
    console.log(person._id);
  }
})
```

Version without `_id` being part of the TSchema:
```js
import { Collection, MongoClient, ObjectId } from 'mongodb'

interface Person {
  name: string 
  age: number
}

MongoClient.connect().then(async client => {
  const db = client.db()
  const personCollection: Collection<Person> = db.collection('persons');
  
  const writeResult = await personCollection.insertOne({ name: "Foo Bar", age: 30 }) 
  
  const person = await personCollection.findOne({ _id: writeResult.insertedId })
  
  if (person) {
    console.log(person._id); // type error, _id is not defined
  }
})
```

In either case it's not possible to sensibly type collections anymore :(

I think it would make sense to make the `_id` property optional for the inserts.
I can make a PR I just wanted to ask for feedback first.

smeijer:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://mongodb.github.io/node-mongodb-native/3.2/api/Collection.html#insert

---

```ts
interface User {
  _id: string;
  name: string;
}

const doc: Partial<User> = {
  name: 'john',
};


db.users.insert(doc); // db.users = Collection<User>
```

This would throw an error, as `_id` is required. We could define `_id` as optional to fix this. But in every other case than `insert`, `_id` is always present. When sharing types between mongo and graphql, for example with [`graphql-code-gen`](https://graphql-code-generator.com), marking `_id` as optional is not really a serious option. But even besides that, optional _id's are often unwanted. As every user entity should have an `_id` anyway.

This PR marks the `_id` property as `optional any` for all `Collection<T>.insert` statements.

---

resolves #33740

### DIFF
--- a/types/mongodb/index.d.ts
+++ b/types/mongodb/index.d.ts
@@ -995,19 +995,19 @@ export interface Collection<TSchema = Default> {
     initializeUnorderedBulkOp(options?: CommonOptions): UnorderedBulkOperation;
     /** http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#insertOne */
     /** @deprecated Use insertOne, insertMany or bulkWrite */
-    insert(docs: TSchema, callback: MongoCallback<InsertOneWriteOpResult>): void;
+    insert(docs: Omit<TSchema, '_id'>, callback: MongoCallback<InsertOneWriteOpResult>): void;
     /** @deprecated Use insertOne, insertMany or bulkWrite */
-    insert(docs: TSchema, options?: CollectionInsertOneOptions): Promise<InsertOneWriteOpResult>;
+    insert(docs: Omit<TSchema, '_id'>, options?: CollectionInsertOneOptions): Promise<InsertOneWriteOpResult>;
     /** @deprecated Use insertOne, insertMany or bulkWrite */
-    insert(docs: TSchema, options: CollectionInsertOneOptions, callback: MongoCallback<InsertOneWriteOpResult>): void;
+    insert(docs: Omit<TSchema, '_id'>, options: CollectionInsertOneOptions, callback: MongoCallback<InsertOneWriteOpResult>): void;
     /** http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#insertMany */
-    insertMany(docs: TSchema[], callback: MongoCallback<InsertWriteOpResult>): void;
-    insertMany(docs: TSchema[], options?: CollectionInsertManyOptions): Promise<InsertWriteOpResult>;
-    insertMany(docs: TSchema[], options: CollectionInsertManyOptions, callback: MongoCallback<InsertWriteOpResult>): void;
+    insertMany(docs: Omit<TSchema, '_id'>[], callback: MongoCallback<InsertWriteOpResult>): void;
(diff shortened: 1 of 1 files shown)


HosseinAgha:
@caselIT could you please merge this PR as the current behaviour is so annoying.

types/mongodb/index.d.ts
```diff
@@ -995,19 +995,19 @@ export interface Collection<TSchema = Default> {
     initializeUnorderedBulkOp(options?: CommonOptions): UnorderedBulkOperation;
     /** http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#insertOne */
     /** @deprecated Use insertOne, insertMany or bulkWrite */
-    insert(docs: TSchema, callback: MongoCallback<InsertOneWriteOpResult>): void;
+    insert(docs: Omit<TSchema, '_id'> & { _id?: any }, callback: MongoCallback<InsertOneWriteOpResult>): void;
```
HosseinAgha: I think defining a separate type function like the one described in [this comment](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33740#issuecomment-508139851) makes the code much more compact.

CaselIT:
@HosseinAgha it's not up to me.

I think the problem is that `Omit` was added in typescript 3.5. The current typing require typescript 2.4.  
I think it's too early to require v3.5, but we could define `Omit` as 
`type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>` that would require v2.8. [Ref](https://stackoverflow.com/questions/48215950/exclude-property-from-type)

HosseinAgha:
@CaselIT @smeijer I'm going to create a new PR then. and bump the required typescript version to v2.8.

smeijer:
@HosseinAgha, sounds awesome! Sorry for not responding sooner.

PR #37211, opened 2019-07-29
[got] correctly type statusCode & statusMessage

LinusU:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9838f610fb841f77ad00f7c60d56b709035fb1a1/types/node/http.d.ts#L198-L205
- [n/a] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

-----

`got`'s `Response` is always a "response obtained from http.ClientRequest", therefore these two properties are not `| undefined`

### DIFF
--- a/types/got/index.d.ts
+++ b/types/got/index.d.ts
@@ -289,6 +289,10 @@ declare namespace got {
         fromCache: boolean;
         redirectUrls?: string[];
         retryCount: number;
+
+        // got's Response is always a "response obtained from http.ClientRequest", therefore these two properties are never undefined
+        statusCode: number;
+        statusMessage: string;
     }
 
     type GotPromise<B extends Buffer | string | object> = Promise<Response<B>> & { cancel(): void };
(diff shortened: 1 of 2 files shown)


types/got/index.d.ts
```diff
@@ -289,6 +289,8 @@ declare namespace got {
         fromCache: boolean;
         redirectUrls?: string[];
         retryCount: number;
+        statusCode: number;
```
ikokostya: nit: Maybe add some comment for this? For example

> got's Response is always a "response obtained from http.ClientRequest", therefore these two properties are not `undefined`

ikokostya (approved):

types/got/index.d.ts
LinusU: Fixed 👌

PR #17684, opened 2017-07-02
fix: `node` in module `readline`, `cursorTo` should allow only x coordinate

a-peyrard:
fix: `node` in module `readline`, `cursorTo` should allow only x coordinate

Like this we are easily able to move in the current line with absolute x-axis coordinate.

This is unfortunately not described in the `readline` documentation (https://nodejs.org/api/readline.html), but this is used in the `readline` source code:
- https://github.com/nodejs/node/blob/master/lib/readline.js#L346
- https://github.com/nodejs/node/blob/master/lib/readline.js#L359
And we can see that a normal behaviour is made when `y` is `undefined`: 
https://github.com/nodejs/node/blob/master/lib/readline.js#L1071

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<url here>>
- [ ] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #37217, opened 2019-07-29
feat(node): v12.7

### LINKED ISSUE #37035: [@types/node] Bug in dns.lookup
Hey @OJezu, @RyanCavanaugh, @lifeiscontent, @armanio123, and @rbuckton 😊 

I think that there is a bug in the definition of the `lookup` function of the `dns` module. It's about [this line](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e8d6fb67f97b50b357f317da6867dfb7a995fee6/types/node/dns.d.ts#L34). Here it says:

```typescript
function __promisify__(hostname: string, options: LookupAllOptions): Promise<{ address: LookupAddress[] }>;
```

Instead, this should be:

```typescript
function __promisify__(hostname: string, options: LookupAllOptions): Promise<LookupAddress[]>;
```

In the section for native promise support (see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e8d6fb67f97b50b357f317da6867dfb7a995fee6/types/node/dns.d.ts#L298), this is correct.

SimonSchick:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#12.7.0

Also:
Fixes #37035

### DIFF
--- a/types/node/dns.d.ts
+++ b/types/node/dns.d.ts
@@ -31,9 +31,9 @@ declare module "dns" {
 
     // NOTE: This namespace provides design-time support for util.promisify. Exported members do not exist at runtime.
     namespace lookup {
-        function __promisify__(hostname: string, options: LookupAllOptions): Promise<{ address: LookupAddress[] }>;
-        function __promisify__(hostname: string, options?: LookupOneOptions | number): Promise<{ address: string, family: number }>;
-        function __promisify__(hostname: string, options?: LookupOptions | number): Promise<{ address: string | LookupAddress[], family?: number }>;
+        function __promisify__(hostname: string, options: LookupAllOptions): Promise<LookupAddress[]>;
+        function __promisify__(hostname: string, options?: LookupOneOptions | number): Promise<LookupAddress[]>;
+        function __promisify__(hostname: string, options?: LookupOptions | number): Promise<LookupAddress | LookupAddress[]>;
     }
 
     function lookupService(address: string, port: number, callback: (err: NodeJS.ErrnoException | null, hostname: string, service: string) => void): void;
--- a/types/node/dns.d.ts
+++ b/types/node/dns.d.ts
@@ -32,8 +32,8 @@ declare module "dns" {
     // NOTE: This namespace provides design-time support for util.promisify. Exported members do not exist at runtime.
     namespace lookup {
         function __promisify__(hostname: string, options: LookupAllOptions): Promise<LookupAddress[]>;
-        function __promisify__(hostname: string, options?: LookupOneOptions | number): Promise<LookupAddress[]>;
-        function __promisify__(hostname: string, options?: LookupOptions | number): Promise<LookupAddress | LookupAddress[]>;
+        function __promisify__(hostname: string, options?: LookupOneOptions | number): Promise<LookupAddress>;
+        function __promisify__(hostname: string, options: LookupOptions): Promise<LookupAddress | LookupAddress[]>;
     }
 
     function lookupService(address: string, port: number, callback: (err: NodeJS.ErrnoException | null, hostname: string, service: string) => void): void;
--- a/types/node/http.d.ts
+++ b/types/node/http.d.ts
@@ -158,6 +158,16 @@ declare module "http" {
         writeHead(statusCode: number, headers?: OutgoingHttpHeaders): this;
     }
 
+    interface InformationEvent {
+        statusCode: number;
+        statusMessage: string;
+        httpVersion: string;
+        httpVersionMajor: number;
+        httpVersionMinor: number;
+        headers: IncomingHttpHeaders;
+        rawHeaders: string[];
+    }
+
     // https://github.com/nodejs/node/blob/master/lib/_http_client.js#L77
     class ClientRequest extends OutgoingMessage {
         connection: Socket;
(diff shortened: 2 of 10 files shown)


types/node/dns.d.ts
```diff
@@ -31,9 +31,9 @@ declare module "dns" {
 
     // NOTE: This namespace provides design-time support for util.promisify. Exported members do not exist at runtime.
     namespace lookup {
-        function __promisify__(hostname: string, options: LookupAllOptions): Promise<{ address: LookupAddress[] }>;
-        function __promisify__(hostname: string, options?: LookupOneOptions | number): Promise<{ address: string, family: number }>;
-        function __promisify__(hostname: string, options?: LookupOptions | number): Promise<{ address: string | LookupAddress[], family?: number }>;
+        function __promisify__(hostname: string, options: LookupAllOptions): Promise<LookupAddress[]>;
+        function __promisify__(hostname: string, options?: LookupOneOptions | number): Promise<LookupAddress[]>;
```
Flarna: I'm quite sure that the `LookupOneOptions` and `LookupOptions ` variant is correct, see the last doc line for lookup at https://nodejs.org/dist/latest-v12.x/docs/api/dns.html#dns_dns_lookup_hostname_options_callback
Maybe the `| number` for the `LookupOptions` variant should be removed as this is better covered by the `LookupOneOptions` variant.

types/node/http.d.ts
```diff
@@ -172,6 +182,21 @@ declare module "http" {
         setTimeout(timeout: number, callback?: () => void): this;
         setNoDelay(noDelay?: boolean): void;
         setSocketKeepAlive(enable?: boolean, initialDelay?: number): void;
+
+        addListener(event: 'information', listener: (info: InformationEvent) => void): this;
+        addListener(event: string | symbol, listener: (...args: any[]) => void): this;
```
Flarna: Doesn't this hide the `addListener`,... variants from base class? sure, there is the generic variant so nothing is broken but now only the `information` event is typed.
On the other hand copy pasting all the other definitions is also not nice...

types/node/dns.d.ts
SimonSchick: I'm not following, please post the suggested diff.

types/node/http.d.ts
SimonSchick: Well, that's a problem, not only in this place but already existing ones as well.
I will use declaration merging instead.

SimonSchick: Which.. apparently doesn't work anymore, this is going to become annoying.

SimonSchick:
@rbuckton any suggestions on how to handle overloads with inheritance?

types/node/dns.d.ts
Flarna: First variant as you suggested, second as it was before and third a combination of both:
```
function __promisify__(hostname: string, options: LookupAllOptions): Promise<LookupAddress[]>;
function __promisify__(hostname: string, options?: LookupOneOptions | number): Promise<{ address: string, family: number }>;
function __promisify__(hostname: string, options: LookupOptions ): Promise<{ address: string, family?: number } | LookupAddress[]>;
```

types/node/http.d.ts
Flarna: With newer typescript versions this could be improved to define lists of types and reuse them at several places... but we have to stick on 2.0/2.1...

SimonSchick:
@flarna I will be n/a for a day or 2, I gave you push rights to my fork for now, feel free to fix things as you see fit.

Flarna:
I committed some fixes. I tried also to add `Readable.readableObjectMode` and `Readable.readableEncoding` but this resulted in other problems so I removed it.

I have not touched the events of `ClientRequest`, maybe @ or @ have a good idea. See also https://github.com/microsoft/TypeScript/issues/30612

weswigham:
> any suggestions on how to handle overloads with inheritance?

Copy them all, tbh. Overloads don't inherit well - you either get all of them or none.

SimonSchick:
@weswigham that's not really a sustainable solution, it solves the problem for now but probably creates more later and I'm curious if the TS/DT team knows a solution to that :)

weswigham:
>  it solves the problem for now but probably creates more later and I'm curious if the TS/DT team knows a solution to that :)

Yeah, I know. You can do something like `methodName: BaseType["methodName"] & {(type: "newOverload"): void}`, but that's kinda ugly, too.

rbuckton:
I've had a number of discussions with @DanielRosenwasser about ways to improve typing for NodeJS (and DOM) events, but we haven't settled on a good solution yet.

Flarna (changes requested):
events from base class of `ClientRequest` need to be copied.

SimonSchick:
@flarna I think one option here would be to extract all event into separate interfaces and then decl merge them together, this way we could keep those reusable, still kinda hacky.

Flarna:
@SimonSchick I just copy pasted once again and added also the other events documented for `ClientRequest`. I think it's better to cleanup the event stuff in a dedicated PR once for all but I fear the limitation to support TS 2.0 will render this anyway to fail.

Flarna (commented):
.

Flarna (approved):

Flarna:
I have the feeling that I self approved this PR now but I have no idea how to avoid this...

BrunoScheufler (approved):

sandersn:
@Flarna so is this ready to go? I got confused by your "self-approved" comment whether you meant to sign off or not.

Flarna:
@sandersn Yes, it's ready. Thanks!

types/node/test/readline.ts
```diff
@@ -76,8 +76,8 @@ const rl: readline.ReadLine = readline.createInterface(new stream.Readable());
     const x = 1;
     const y = 1;
 
-    readline.cursorTo(strm, x);
```
stephank: This line was in place to test the change in #17684, which has now been undone.

Current Node.js still appears to accept `y` as an optional parameter, but they also never updated the docs. (Doesn't look like the PR was ever created.)

Flarna: I think we should start with a PR for node otherwise such mismatches will confuse people again and again. Maybe I find some time the next days to do this.

astorije: For the record, I have just opened https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37630. Hope this helps!

Flarna: I created a PR for node: https://github.com/nodejs/node/pull/29128

PR #37220, opened 2019-07-29
Modify linkifyjs LinkifyOptions interface to allow extra return type for 'format'

cheshire137:
I noticed that I'm able to pass a `format` function that returns either a React component or a string, and Linkify renders either one. This modifies the type declaration for the `format` function so you can return React components in addition to strings. See https://soapbox.github.io/linkifyjs/docs/2.1.html#reactjs-interface for info about the React interface.

### DIFF
--- a/types/linkifyjs/index.d.ts
+++ b/types/linkifyjs/index.d.ts
@@ -17,7 +17,7 @@ export type PossiblyByType<T> = T | { [type: string]: T | React.ReactElement | u
 
 // NOTE: According to the linkifyjs implementation, `format` can be just a
 // string, so we include it (even though this is not mentioned in the docs).
-export type FormatFunc = (value: string, type: string) => string | React.ReactElement;
+export type FormatFunc = (value: string, type: string) => React.ReactNode;
 
 export type EventHandler = (e: HTMLElement) => void;
 
--- a/types/linkifyjs/index.d.ts
+++ b/types/linkifyjs/index.d.ts
@@ -2,6 +2,7 @@
 // Project: https://github.com/SoapBox/linkifyjs#readme
 // Definitions by: Sean Zhu <https://github.com/szhu>
 //                 Ovidiu Bute <https://github.com/ovidiubute>
+//                 Sarah Vessels <https://github.com/cheshire137>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
@@ -10,6 +11,8 @@ import * as React from "react";
 export type PossiblyFuncOfHrefAndType<T> =
     | T
     | ((href: string, type: string) => T);
+
+export type ElementOrString = string | Element;
 
 // There's always a possibility that values at a string key can always be
 // undefined; i.e., if the key does not exist.
@@ -94,7 +97,7 @@ export interface LinkifyOptions {
      * NOTE: According to the linkifyjs implementation, `format` can be just a
      * string, but this is not mentioned in the docs, so we exclude it.
      */
-    format?: PossiblyByType<(value: string, type: string) => string>;
+    format?: PossiblyByType<(value: string, type: string) => ElementOrString>;
     //
 
     /**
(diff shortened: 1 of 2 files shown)


cheshire137:
cc @szhu and @ovidiubute 🙇

types/linkifyjs/linkifyjs-tests.tsx
```diff
@@ -68,6 +68,19 @@ describe("linkifyjs/html", () => {
...
+            if (type === 'url') {
+                return <span>{value}</span>;
+            }
+            return (
+                <span>
+                    {value} ({type})
+                </span>
+            );
+        },
+    });
```
szhu: This doesn't seem to test that `format()` can return a string.

types/linkifyjs/index.d.ts
```diff
@@ -94,7 +94,7 @@ export interface LinkifyOptions {
      * NOTE: According to the linkifyjs implementation, `format` can be just a
      * string, but this is not mentioned in the docs, so we exclude it.
```
szhu:
```suggestion
      * string, so we include it (even though this is not mentioned in the docs).
```

But actually maybe move this sentence to just above where `FormatFunc` is defined?

types/linkifyjs/index.d.ts
```diff
@@ -13,7 +13,9 @@ export type PossiblyFuncOfHrefAndType<T> =
 
 // There's always a possibility that values at a string key can always be
 // undefined; i.e., if the key does not exist.
-export type PossiblyByType<T> = T | { [type: string]: T | undefined };
+export type PossiblyByType<T> = T | { [type: string]: T | React.ReactElement | undefined };
+
+export type FormatFunc = (value: string, type: string) => string | React.ReactElement;
```
szhu: Can you test whether returning `undefined`/`null`/`number`/an array of the above is allowed? If so, the return type can simply be `React.ReactNode`.

szhu (changes requested):

szhu (commented):
Thanks for contributing this change! Feel free to let me know if you have any questions

types/linkifyjs/linkifyjs-tests.tsx
cheshire137: There's an existing `format` test above that returns only strings, so I had this one returning only elements. I can also add a test where `format` returns strings and elements.

types/linkifyjs/index.d.ts
cheshire137: Good idea! I just tried it, using `export type FormatFunc = (value: string, type: string) => React.ReactNode;` in my type declaration in my project. Returning `null` or `undefined` works and results in an `<a>` tag with no children, as you might expect. Returning a number works and results in that number being the clickable text of the link. Returning an array also works; I tried returning `['cat', 'dog']` and the link text became `catdog`.

types/linkifyjs/index.d.ts
```diff
@@ -13,7 +13,11 @@ export type PossiblyFuncOfHrefAndType<T> =
 
 // There's always a possibility that values at a string key can always be
 // undefined; i.e., if the key does not exist.
-export type PossiblyByType<T> = T | { [type: string]: T | undefined };
+export type PossiblyByType<T> = T | { [type: string]: T | React.ReactElement | undefined };
```
szhu: What was this modification for? `PossiblyByType` is sort of a type that deals mostly with logic. I don't think `React.ReactElement` belongs here.

If there's a specific reason you were trying to use `React.ReactElement` for but you're not sure how to do it, feel free to explain and I can see if I can come up with a way to solve it as well.

types/linkifyjs/linkifyjs-tests.tsx
szhu: Oh, I totally forgot that `value` is a string -- that's why I didn't think any of the functions above were returning strings. Thanks for explaining!

orta:
👋 - just going through old PRs, I think these changes are now integrated in the current types for linkify. Want to double check and close this if so?

cheshire137:
Thank you, @orta!

PR #37239, opened 2019-07-30
Update net.d.ts

mlibanori:
the address () object was always returning a string instead of an addressinfo object


Flarna (changes requested):
If you connect via a unix domain socket or windows pipe the return value is a `string`. Therefore I don't think this change is correct. see also https://nodejs.org/dist/latest-v12.x/docs/api/net.html#net_server_address

SimonSchick (changes requested):
The [https://nodejs.org/dist/latest-v12.x/docs/api/net.html#net_server_address](docs) (which you failed to link to btw) contradict your change request.

SimonSchick:
God damnit, sniped again @Flarna

PR #37240, opened 2019-07-30
[@types/react-native-actionsheet] Add missing ActionSheetCustom

kristerkari:
Update `@types/react-native-actionsheet` to support [ActionSheetCustom](https://github.com/beefe/react-native-actionsheet#use-actionsheetcustom-directly).

The default `ActionSheet` class should match the options schema from `react-native-actionsheet`: https://github.com/beefe/react-native-actionsheet/blob/master/lib/options.js

The `ActionSheetCustom` class has the same options, but with these differences:

- `title` is a `ReactNode` instead of `string`.
- `options` is a `ReactNode[]` instead of `string[]`.
- `style` object can be given.

I have added a test that matches the example given in `react-native-actionsheet` project's README. I have also tested that the typings are working in my own project where I use the library.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/beefe/react-native-actionsheet#use-actionsheetcustom-directly

### DIFF
--- a/types/react-native-actionsheet/index.d.ts
+++ b/types/react-native-actionsheet/index.d.ts
@@ -7,9 +7,19 @@
 import * as React from 'react';
 
 export interface ActionSheetProps {
-    options: React.ReactNode[];
+    options: string[];
     onPress: (index: number) => void;
     title?: string;
+    message?: string;
+    tintColor?: string;
+    cancelButtonIndex?: number;
+    destructiveButtonIndex?: number;
+}
+
+export interface ActionSheetCustomProps {
+    options: React.ReactNode[];
+    onPress: (index: number) => void;
+    title?: React.ReactNode;
     message?: string;
     tintColor?: string;
     cancelButtonIndex?: number;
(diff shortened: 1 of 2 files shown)


types/react-native-actionsheet/index.d.ts
```diff
@@ -7,16 +7,30 @@
...
+    options: string[];
     onPress: (index: number) => void;
     title?: string;
     message?: string;
     tintColor?: string;
     cancelButtonIndex?: number;
     destructiveButtonIndex?: number;
+}
+
+export interface ActionSheetCustomProps {
```
ian-rudge: how about `buttonUnderlayColor`?

kristerkari: Good catch! It's undocumented, so I did not even notice it.

kristerkari: I have added it now.

kristerkari: @ian-rudge can you please re-review?

ian-rudge (approved):
looks good to me!

PR #37245, opened 2019-07-30
react-autosuggest: include state types

samijaber:
Adding the `State` types and `this.input`

  - https://github.com/moroshko/react-autosuggest/blob/d626cef720df0ed5a4c7df4d539836b86bda6afa/src/Autosuggest.js#L101-L108
  - https://github.com/moroshko/react-autosuggest/blob/d626cef720df0ed5a4c7df4d539836b86bda6afa/src/Autosuggest.js#L120

### DIFF
--- a/types/react-autosuggest/index.d.ts
+++ b/types/react-autosuggest/index.d.ts
@@ -20,7 +20,7 @@ declare class Autosuggest<T = any> extends React.Component<
      * Autosuggest exposes these class properties to the parent component.
      * They can be accessed through `ref`.
      */
-    input: HTMLInputElement | null;
+    input: HTMLInputElement | undefined;
 }
 
 export = Autosuggest;
--- a/types/react-autosuggest/index.d.ts
+++ b/types/react-autosuggest/index.d.ts
@@ -12,7 +12,10 @@
 
 import * as React from 'react';
 
-declare class Autosuggest<T = any> extends React.Component<Autosuggest.AutosuggestProps<T>, Autosuggest.AutosuggestState<T>> {
+declare class Autosuggest<T = any> extends React.Component<
+    Autosuggest.AutosuggestProps<T>,
+    Autosuggest.AutosuggestState<T>
+> {
     /**
      * Autosuggest exposes these class properties to the parent component.
      * They can be accessed through `ref`.
--- a/types/react-autosuggest/index.d.ts
+++ b/types/react-autosuggest/index.d.ts
@@ -13,11 +13,11 @@
 import * as React from 'react';
 
 declare class Autosuggest<T = any> extends React.Component<Autosuggest.AutosuggestProps<T>, Autosuggest.AutosuggestState<T>> {
-  /**
-   * Autosuggest exposes these class properties to the parent component.
-   * They can be accessed through `ref`.
-   */
-  input: HTMLInputElement | null;
+    /**
+     * Autosuggest exposes these class properties to the parent component.
+     * They can be accessed through `ref`.
+     */
+    input: HTMLInputElement | null;
 }
 
 export = Autosuggest;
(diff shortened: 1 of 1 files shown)


cdeutsch:
I'm not sure the goal of typings is to expose the inner workings of components.

`input` isn't listed in the docs  
https://github.com/moroshko/react-autosuggest/tree/d626cef720df0ed5a4c7df4d539836b86bda6afa#props

I could see the small use cases for needing access to `input` (even though I'd consider it a React anti-pattern), but not sure why you'd want to know the types for its state. 

Are there use-cases for accessing its state?

I'm not sure what the DefinitelyTyped policy is, but I feel like forcing users to do a `((ref.autoComplete as any).input as HTMLInputElement)` to get to the `input` would be a good thing. A bit of a warning that you're taking a risk accessing internal properties.

types/react-autosuggest/index.d.ts
```diff
@@ -12,7 +12,16 @@
...
-declare class Autosuggest<T = any> extends React.Component<Autosuggest.AutosuggestProps<T>> {}
+declare class Autosuggest<T = any> extends React.Component<
+    Autosuggest.AutosuggestProps<T>,
+    Autosuggest.AutosuggestState<T>
+> {
+    /**
+     * Autosuggest exposes these class properties to the parent component.
+     * They can be accessed through `ref`.
+     */
+    input: HTMLInputElement | null;
```
cdeutsch: Is `undefined` more accurate? Or will it actually be `null`?

The Typescript convention is to prefer undefined for most things (unless it's literally going to be a `null`)

samijaber:
> I'm not sure the goal of typings is to expose the inner workings of components.

See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26587 , where I did the exact same thing for `@types/react-autocomplete` (we're migrating to a non-deprecated library instead 🙂)

> I could see the small use cases for needing access to `input` (even though I'd consider it a React anti-pattern), but not sure why you'd want to know the types for its state.

Our use-case is having a button that, when clicked, will focus the `input`. There's no way I see to do that without doing `input.focus()`.

> Are there use-cases for accessing its state?

Our use-case is that we want to hide suggestions when the user hits the UP key on the _first_ suggestion. Can't quite do that using the library's helpers, so we read `highlightedSuggestionIndex` inside a `onKeyDown` handler.

> I'm not sure what the DefinitelyTyped policy is, but I feel like forcing users to do a `((ref.autoComplete as any).input as HTMLInputElement)` to get to the `input` would be a good thing. A bit of a warning that you're taking a risk accessing internal properties.

Using `refs` is a fair escape hatch built into React. Obfuscating/not exposing the types would be a roundabout way of discouraging people from using `ref`, and forcing them to guess the types they're dealing with.

types/react-autosuggest/index.d.ts
samijaber: `undefined` is more accurate. Mistake on my part.

types/react-autosuggest/index.d.ts
```diff
@@ -199,4 +208,13 @@ declare namespace Autosuggest {
...
         theme?: Theme;
     }
+
+    interface AutosuggestState<TSuggestion> {
+        isFocused: boolean;
+        isCollapsed: boolean;
+        highlightedSectionIndex: number | null;
+        highlightedSuggestionIndex: number | null;
+        highlightedSuggestion: TSuggestion | null;
+        valueBeforeUpDown: TSuggestion | null;
```
cdeutsch: I'd make these `undefined` as well, unless you're sure they'll actually be `null`'s

samijaber: These are indeed initialized as `null` in the codebase. See https://github.com/moroshko/react-autosuggest/blob/d626cef720df0ed5a4c7df4d539836b86bda6afa/src/Autosuggest.js#L101-L108

cdeutsch:
> They list "Integrating with third-party DOM libraries" as a good example of when to use them

The `input.focus` I can understand. 

I think reaching into a component to get its state is a bad practice (breaking the one-way data flow philosophy), and you'd be better off forking the library or submitting a PR to get those values via an event, but its your code base and having the types there doesn't harm anything so 👍

PR #37255, opened 2019-07-31
[lscache] Update version to v.1.3.0 

c-py:
Please fill in this template.

- [ X] Add or edit tests to reflect the change. (Run with `npm test`.)
- [X ] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [ X] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [ X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/pamelafox/lscache/blob/master/lscache.js
- [ X] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.


types/lscache/index.d.ts
```diff
@@ -1,6 +1,7 @@
-// Type definitions for lscache v1.0.7
+// Type definitions for lscache v1.3.0
```
sandersn: Definitely Typed doesn't retain patch versions, so you should just put `1.3` here.

types/lscache/index.d.ts
```diff
@@ -1,6 +1,7 @@
...
 // Project: https://github.com/pamelafox/lscache
 // Definitions by: Chris Martinez <https://github.com/Chris-Martinezz>
+//               : c-py <https://github.com/c-py>
```
sandersn: I don't think this is the accepted format for this field. Check something like react with lots of authors for an example.

sandersn (changes requested):
It looks like the dt-header and npm-naming lint rules are disabled in tslint.json. Since you're changing the header, can you please re-enable at least dt-header? Preferably both unless npm-naming fails -- but it shouldn't, because lscache 1.3.0 is on npm.

c-py:
Thanks @sandersn ! I have updated the PR with the suggested changes.

sandersn (approved):

PR #37261, opened 2019-07-31
[yargs] update definition of .completion() to allow hiding it

royra:
Although the docs only [mention it in a different context](https://github.com/yargs/yargs/blob/master/docs/api.md#commandmodule), passing `false` for a description hides it from the `help` output.


royra:
Also - a relevant issue from the project itself: https://github.com/yargs/yargs/issues/1328#issuecomment-484336198

JimiC (approved):
LGTM

PR #37263, opened 2019-07-31
add @types/express-form-data

NomNes:


types/express-form-data/index.d.ts
```diff
@@ -0,0 +1,37 @@
+// Type definitions for express-form-data 2.0
+// Project: https://github.com/ortexx/express-form-data#readme
+// Definitions by: nomnes <https://github.com/NomNes>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
+// INFO: without disabling 'no-single-declare-module' and 'no-declare-current-package' it not working correctly
+// tslint:disable-next-line no-single-declare-module no-declare-current-package
```
jessetrinity: Instead of disabling these rules, could you follow the guidance given in the rules documentation?
https://github.com/microsoft/dtslint/blob/master/docs/no-declare-current-package.md
https://github.com/microsoft/dtslint/blob/master/docs/no-single-declare-module.md

jessetrinity (changes requested):

types/express-form-data/index.d.ts
NomNes: @jessetrinity  I've tried writing by the rules, but when tested in node_module/@types directly, typescript doesn't see the declared functions. Looks like I need some help with that.

jessetrinity: Could you describe what you did a bit more. For example, does this not work:

```typescript

import { NextHandleFunction } from 'connect';
import { FormOptions } from 'multiparty';

export interface FormDataOptions extends FormOptions {
    autoClean?: boolean;
    autoFields?: boolean;
    autoFiles?: boolean;
    maxFields?: number;
    maxFieldsSize?: number;
    maxFilesSize?: number;
    uploadDir?: string;
    encoding?: string;

    [key: string]: any;
}

export function parse(options?: FormDataOptions): NextHandleFunction;

export function format(): NextHandleFunction;

export function stream(): NextHandleFunction;

export function union(): NextHandleFunction;
```

NomNes: WebStorm
```
import express, { Application } from "express";
import formData from "express-form-data";

const app: Application = express();

app.use(formData.parse({autoClean: true}));
app.use(formData.format());
app.use(formData.union());
```
Basically, it understands that these functions are declared. But cmd+B transition works fine only for union, other functions lead to random third packages.

But it works fine:
```
import { format, parse, stream, union } from "express-form-data";
...
app.use(parse({autoClean: true}));
app.use(format());
app.use(union());
app.use(stream());
```

NomNes: Most likely it is a bug in webstorm

jessetrinity: I'm not sure what else you have installed locally so I can't really comment on the 'third party' content you are seeing, but you do not have an `export default` so you should not be getting anything from the second import `import formData from "express-form-data";`.

If you do not want to do `import { format, parse, stream, union } from "express-form-data";` you can instead use `import * as formData from "express-form-data";`

Your tests will then ideally look like [the example usage from the library](https://www.npmjs.com/package/express-form-data#example):

```typescript
app.use(formData.parse(options));
app.use(formData.format());
app.use(formData.stream());
app.use(formData.union());
```

jessetrinity:
In the future I recommend filling out the PR template appropriately. This will help reviewers and maintainers to review your changes.

NomNes:
Is there anything else you need from me? The change request is made

jessetrinity:
Sorry - looks good. Thanks!

PR #37291, opened 2019-08-01
[segment-analytics] support analytics.js-core library

MiroslavPetrik:
Add typings for the `@segment/analytics.js-core` library.

The segment-analytics and @segment/analytics.js-core are the same, as the first one just configures the later one with integrations.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/segmentio/analytics.js-core/blob/master/lib/analytics.js

### DIFF
--- a/types/segment-analytics/index.d.ts
+++ b/types/segment-analytics/index.d.ts
@@ -1,6 +1,7 @@
 // Type definitions for Segment's analytics.js
 // Project: https://segment.com/docs/libraries/analytics.js/
 // Definitions by: Andrew Fong <https://github.com/fongandrew>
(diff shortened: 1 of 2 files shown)


fongandrew:
Does the main analytics.js object have those new methods defined, or are they only on the core one? I don't see them defined on Segment's docs, so wondering if we should make an interface for core that extends the one used by the main analytics object?

MiroslavPetrik:
@fongandrew yes, for example [here](https://github.com/segmentio/analytics.js/blob/master/analytics.js#L283) is the `init` method.

fongandrew (approved):
Cool, LGTM then.

PR #37319, opened 2019-08-02
[openid-client] Add more detailed types to introspection response as per RFC7662

RMHonor:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://tools.ietf.org/html/rfc7662#section-2.2

### DIFF
--- a/types/openid-client/index.d.ts
+++ b/types/openid-client/index.d.ts
@@ -92,7 +92,7 @@ export interface EndSessionUrlParameters {
 
 export interface IntrospectionResponse {
     active: boolean;
-    scope?: string[];
+    scope?: string;
     client_id?: string;
     username?: string;
     token_type?: string;
--- a/types/openid-client/index.d.ts
+++ b/types/openid-client/index.d.ts
@@ -2,6 +2,7 @@
 // Project: https://github.com/panva/node-openid-client
 // Definitions by: ulrichb <https://github.com/ulrichb>
 //                 Brandon Shelton <https://github.com/YangusKhan>
+//                 Richard Honor <https://github.com/RMHonor>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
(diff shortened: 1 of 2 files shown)


types/openid-client/index.d.ts
```diff
@@ -89,6 +90,22 @@ export interface EndSessionUrlParameters {
...
+    username?: string;
+    token_type?: string;
+    exp?: number;
+    iat?: number;
+    nbf?: number;
+    sub?: string;
+    aud?: string;
+    iss?: string;
+    jti?: string;
+    [key: string]: any;
```
ulrichb: I would prefer `{} | null | undefined` instead of `any` for the indexer.

This forces the callers to think about the result and do an explicit cast (except for the props you added above).

ulrichb (changes requested):

types/openid-client/index.d.ts
```diff
@@ -89,6 +90,22 @@ export interface EndSessionUrlParameters {
     readonly state?: string;
 }
 
+export interface IntrospectionResponse {
+    active: boolean;
+    scope?: string[];
```
ulrichb: Why an array?
image

types/openid-client/index.d.ts
ulrichb: ... also please add `readonly` modifiers to all members (favor immutability)

types/openid-client/index.d.ts
RMHonor: Not quite sure why I thought this was an array, I've updated the type to be a string

types/openid-client/index.d.ts
RMHonor: `any` is probably a bit too forgiving here. 

How would you feel about using `unknown` instead? It would still require explicit casting, and would be more indicative of the intent behind the response.

ulrichb: Ah ... we're already at TS 3.0 ... yes, use `unknown`

ulrichb (approved):

ulrichb:
@RMHonor Many thanks for the PR!

RMHonor:
@ulrichb no problem, thanks for guidance!

PR #37337, opened 2019-08-03
Add types for bayes-classifier

jasonharrison:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/bayes-classifier/index.d.ts
+++ b/types/bayes-classifier/index.d.ts
@@ -4,12 +4,12 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.5
 
-export interface Classifications {
+interface Classifications {
     label: string;
     value: number;
 }
 
-export default class bayes_classifier {
+declare class BayesClassifier {
     constructor();
 
     addDocument(doc: string, label: string): void;
@@ -24,3 +24,5 @@ export default class bayes_classifier {
 
     train(): void;
 }
+
+export = BayesClassifier;
--- a/types/bayes-classifier/index.d.ts
+++ b/types/bayes-classifier/index.d.ts
@@ -1,0 +1,26 @@
+// Type definitions for bayes-classifier 0.0
+// Project: https://github.com/miguelmota/bayes-classifier
+// Definitions by: Jason Harrison <https://github.com/jasonharrison>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.5
+
+export interface Classifications {
+    label: string;
+    value: number;
+}
+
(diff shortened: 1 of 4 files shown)


types/bayes-classifier/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6"
+        ],
+        "allowSyntheticDefaultImports": true,
```
sandersn: older code bases will not have this set by default, so you should remove it.

types/bayes-classifier/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+// Definitions by: Jason Harrison <https://github.com/jasonharrison>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.5
+
+export interface Classifications {
+    label: string;
+    value: number;
+}
+
+export default class bayes_classifier {
```
sandersn: The BayesClassifier type in the source project is exported with `module.exports = BayesClassifier`, so you need to use the equivalent TS syntax: `export = BayesClassifier`.

sandersn (changes requested):
Need to fix problems with export.

types/bayes-classifier/index.d.ts
jasonharrison: I made this change but had to edit `tsconfig.json` and enable `esModuleInterop = true` to make things pass.  Can you check things over to make sure my changes were correct?  Thanks!

types/bayes-classifier/tsconfig.json
jasonharrison: I removed `allowSyntheticDefaultImports`, but left `"lib": ["es6"]` because I would get errors if it was removed.  Was this what you wanted?

sandersn: Yep!

types/bayes-classifier/index.d.ts
sandersn: Sorry, I forgot to mention that the test needs to use a commonjs import to match the commonjs export when `esModuleInterop: "false"`: `import BayesClassifier = require('bayes-classifier')`.

Note that *most* users will have `esModuleInterop: "true"` these days, and thus be able to use `import BayesClassifier from 'bayes-classifier'`, but for backward compatibility, DT tests should have `esModuleInterop: "false"` and use the old syntax in order to make sure it works for old code bases.

jasonharrison: Thank you very much for the guidance, @sandersn.  I am relatively new to TS/JS and this is my first type definition contribution.  Please check now.

sandersn (approved):

PR #37340, opened 2019-08-03
[@types/marked] Renderer instance method parameter types.

kenchan0130:
* Types of Renderer instance methods are different from the reality

Please fill in this template.

### DIFF
--- a/types/marked/index.d.ts
+++ b/types/marked/index.d.ts
@@ -95,7 +95,7 @@ declare namespace marked {
         html(html: string): string;
         heading(text: string, level: number, raw: string, slugger: Slugger): string;
         hr(): string;
-        list(body: string, ordered: boolean, start: number): string;
+        list(body: string, ordered: boolean, start: number | ''): string;
         listitem(text: string): string;
         paragraph(text: string): string;
         table(header: string, body: string): string;
@@ -109,8 +109,8 @@ declare namespace marked {
         codespan(code: string): string;
         br(): string;
         del(text: string): string;
-        link(href: string, title: string, text: string): string;
-        image(href: string, title: string, text: string): string;
+        link(href: string, title: string | null, text: string): string;
+        image(href: string, title: string | null, text: string): string;
         text(text: string): string;
     }
 


worr:
Thanks! Your commit made me realize that we should probably turn on `strictNullChecks` for this package, since it somehow was never turned on.

types/marked/index.d.ts
```diff
@@ -95,7 +95,7 @@ declare namespace marked {
         html(html: string): string;
         heading(text: string, level: number, raw: string, slugger: Slugger): string;
         hr(): string;
-        list(body: string, ordered: boolean, start: number): string;
+        list(body: string, ordered: boolean, start: number | ''): string;
```
worr: I'm not sure that we'd want to do this? I see that `startatt` (https://github.com/markedjs/marked/blob/master/lib/marked.js#L990) may possibly be assigned the eventual value of `''`, but it doesn't seem like this interface is meant to take the empty string.

types/marked/index.d.ts
```diff
@@ -109,8 +109,8 @@ declare namespace marked {
         codespan(code: string): string;
         br(): string;
         del(text: string): string;
-        link(href: string, title: string, text: string): string;
-        image(href: string, title: string, text: string): string;
+        link(href: string, title: string | null, text: string): string;
+        image(href: string, title: string | null, text: string): string;
```
worr: Here, `href` can definitely be nullable, but not sure we should allow `null` in terms of `title` here.

worr (changes requested):

types/marked/index.d.ts
kenchan0130: @worr
We can get `null` when we pass a markdown like

```md
[This is text](http://example.com)
```

If we explicitly pass the title as follows, it will not be null.

```md
[This is text](http://example.com "This is title")
```

types/marked/index.d.ts
kenchan0130: Unfortunately, There's a possibility that the `start` type is string.
https://github.com/markedjs/marked/blob/master/lib/marked.js#L327

For stack:
1. https://github.com/markedjs/marked/blob/master/lib/marked.js#L1259
2. https://github.com/markedjs/marked/blob/master/lib/marked.js#L1253
3. https://github.com/markedjs/marked/blob/master/lib/marked.js#L1158
4 https://github.com/markedjs/marked/blob/master/lib/marked.js#L331

PR #37354, opened 2019-08-04
[@types/node] generics on stream API on ts3.2

EduardoRFS:
### Current
- Added generic support for every API used on stream
- Duplex and Transform is <Write, Read> because it seems to make sense
- Fix every mismatch that I found between the doc and the type definition
- Pipe on writable isn't valid, because it will always throw
- Is a breaking change, but tried to make it really compatible

### Suggestions
- Make the api more strictly in general
- Use unknown instead of any in listeners and on generic default use the right type
- Remove pipe from writable, it exists but even the doc doesn't say it

### Missing
- `pipeline()` but I will do it, if the proposal will be accepted

### Why breaking change?
```javascript
// this is still valid
new Writable({
    write(chunk, enc, cb) {
        // $ExpectType any
        chunk;
    },
    writev(chunks, cb) {
        // $ExpectType { chunk: any; encoding: string; }[]
        chunks;
    }
});

// because of type inference, this isn't valid anymore
new Writable({
    write(chunk: string, enc, cb) {
        // $ExpectType string
        chunk;
    },
    writev(chunks, cb) {
        // will fail, because now chunk is string
        // $ExpectType { chunk: any; encoding: string; }[]
        chunks;
    }
});

// it's easy to make it compatible again
new Writable<any>({
    write(chunk: string, enc, cb) {
        // $ExpectType string
        chunk;
    },
    writev(chunks, cb) {
        // $ExpectType { chunk: any; encoding: string; }[]
        chunks;
    }
});
```

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: (https://nodejs.org/api/stream.html)

### DIFF
--- a/types/node/base.d.ts
+++ b/types/node/base.d.ts
@@ -27,7 +27,6 @@
 /// <reference path="querystring.d.ts" />
 /// <reference path="readline.d.ts" />
 /// <reference path="repl.d.ts" />
-/// <reference path="stream.d.ts" />
 /// <reference path="string_decoder.d.ts" />
 /// <reference path="timers.d.ts" />
 /// <reference path="tls.d.ts" />
--- a/types/node/globals.d.ts
+++ b/types/node/globals.d.ts
@@ -626,31 +626,6 @@ declare namespace NodeJS {
         eventNames(): Array<string | symbol>;
     }
 
-    interface ReadableStream extends EventEmitter {
-        readable: boolean;
-        read(size?: number): string | Buffer;
-        setEncoding(encoding: string): this;
-        pause(): this;
-        resume(): this;
-        isPaused(): boolean;
-        pipe<T extends WritableStream>(destination: T, options?: { end?: boolean; }): T;
-        unpipe(destination?: WritableStream): this;
-        unshift(chunk: string | Uint8Array, encoding?: BufferEncoding): void;
-        wrap(oldStream: ReadableStream): this;
-        [Symbol.asyncIterator](): AsyncIterableIterator<string | Buffer>;
-    }
-
-    interface WritableStream extends EventEmitter {
-        writable: boolean;
-        write(buffer: Uint8Array | string, cb?: (err?: Error | null) => void): boolean;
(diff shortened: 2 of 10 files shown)


SimonSchick:
I'm 100% for this but this is very hard to push through without breaking probably hundreds of packages already on DT.

EduardoRFS:
@SimonSchick It will break only the case where it is strictly wrong(the example on , but my PR closed after a force push, so I can't run the CI again, and for whatever reason I can only run dtslint on my machine, running npm test doesn't work.

But I could also set the default ChunkType to any, so it should accept any package.

SimonSchick:
Problem is that this will likely also break packages/projects that depend on the previous definitions.
Also the fact that this only works with 3.1+ is another problem.
@rbuckton did you ever come up with a plan to update node typings to 3.x? 😅

EduardoRFS:
It doesn't break projects that uses the previous definition, because it's fully compatible, but it does break dtslint `$ExpectType` that's why I opened a issue there, and will try to write a PR for it also. Because $ExpectType comparing based on string doesn't allow to extends types with generics + default, and it is a easy solution to improve typings without fully break the entire ecossystem

https://github.com/microsoft/dtslint/issues/242

EduardoRFS:
After running with a [PR for dtslint](https://github.com/microsoft/dtslint/pull/243) remains:
- cloneable-readable
- csv2json
- node
- pngjs2
- readable-stream

`cloneable-readable` is because it is just a string hack, but I'm trying to make it with the right tool.

`csv2json` and `pngjs2` are trying to use pipe from writable, I don't know if it is a hack or definition problem, but it's easy fixable or here by allowing pipe on writable or just fixing both.

`node` is just replacing Array<> by []

`readable-stream` this one I'm not certain, but doesn't seem hard, just that the error message seems like a pyramid

So we can merge it if it's accepted at dtslint(probably a better implementation),  a breaking change, but every case that is valid, it's still valid, and some invalid cases that was valid now they're invalid.

EduardoRFS:
Waiting https://github.com/microsoft/dtslint/issues/242

PR #37364, opened 2019-08-04
airtable: update SelectOptions to match airtable docs

jonathonlui:
Without these additional options the only way to call `.select()` without type errors was to cast to `any`:
```ts
table.select({ maxRecords: 10 } as any);
```

I tried use something like `<keyof TFields>` on the `fields` and `sort` options so those option values can be type checked (and for nicer autocomplete) but since `FieldSet` uses dynamic keys I wasn't successful in getting anything except `string` as the field type.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: API docs are behind login-wall. See attachement

select-docs

### DIFF
--- a/types/airtable/airtable-tests.ts
+++ b/types/airtable/airtable-tests.ts
@@ -12,7 +12,27 @@ const base = airtable.base('app id');
 const table = base('table name') as Airtable.Table<Row>;
 
 async () => {
-    const query = table.select();
+    const query = table.select({
+        fields: [
+            'field1',
+        ],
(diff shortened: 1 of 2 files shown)


bvalosek (approved):
awesome thanks for the contribution 👍 

I don't use these typings anymore so I don't have the recent first-hand experience to pull from, but there might be a way to go about having better types on the fields like you mentioned, but it may require a bit of re-tooling on how this is setup.

PR #37375, opened 2019-08-05
[ecore] - Add types definition.

michael-whi:
Please fill in this template.

If adding a new definition:


types/ecore/index.d.ts
```diff
@@ -0,0 +1,151 @@
...
+        parse: (data: EObject) => void;
+        elements: (type?: string | EClass) => EObject[];
+        toJSON: () => any;
+    }
+    namespace ResourceSet {
+        function create(): ResourceSet;
+    }
+}
+
+export default Ecore;
```
sandersn: ecore's source package doesn't have a default export as far as I can tell. I think the correct syntax is the commonjs export: `export = Ecore;`

sandersn (changes requested):

types/ecore/index.d.ts
michael-whi: When I do it like that, I get errors during tests.

Error: F:/git/types/DefinitelyTyped/types/ecore/ecore-tests.ts:1:10
ERROR: 1:10   expect                 TypeScript@next compile error: 
Module '"../../../DefinitelyTyped/types/ecore"' has no exported member 'Ecore'.

F:/git/types/DefinitelyTyped/types/ecore/index.d.ts:151:1
ERROR: 151:1  export-just-namespace  Instead of `export =`-ing a namespace, use the body of the namespace as the module body. See: https://github.com/Microsoft/dtslint/blob/master/docs/export-just-namespace.md

sandersn: I'm sorry, I forgot to mention that the import also needs to match the export: `import Ecore = require('ecore')`.

`import { Ecore } from 'ecore'` would technically work but only because the package source does `Ecore.Ecore = Ecore` (or its equivalent).

michael-whi: Thank you! I've fixed the first error, but still, the second error is not gone. What can I do with that?

```
Error: F:/git/types/DefinitelyTyped/types/ecore/index.d.ts:151:1
ERROR: 151:1  export-just-namespace  Instead of `export =`-ing a namespace, use the body of the namespace as the module body. See: https://github.com/Microsoft/dtslint/blob/master/docs/export-just-namespace.md
```

sandersn: Oops. I didn't read carefully enough to see that error. That lint rule is asking you to delete `declare namespace Ecore` and `export = Ecore`, dedent the body of the namespace, and put the `export` keyword in front of anything that should be exported (which is probably everything). They are equivalent, plus it allows people to directly import a few symbols from the module if they want.

michael-whi: Well, I did some research. As far as I can understand I had to get rid of a namespace declaration, cause in case of 'require' it would look strange like Ecore.Ecore as you said. I don't know whether it an appropriate way or not.

sandersn: Looks good to me now.

sandersn (approved):

types/ecore/index.d.ts
michael-whi: Thank you for your help and kindness!

PR #37377, opened 2019-08-05
types/qunit-dom: Fix type errors and update to latest API spec

LucasHill:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/simplabs/qunit-dom/blob/master/API.md>>

### DIFF
--- a/types/qunit-dom/index.d.ts
+++ b/types/qunit-dom/index.d.ts
@@ -1,10 +1,12 @@
-// Type definitions for qunit-dom 0.6
+// Type definitions for qunit-dom 0.9
 // Project: https://github.com/simplabs/qunit-dom#readme
 // Definitions by: Don Denton <https://github.com/happycollision>
+//                 Lucas Hill <https://github.com/lucashill>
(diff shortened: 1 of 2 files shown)


happycollision (changes requested):
Hey there. This PR looks good, but qunit-dom has shipped with its own types since > 0.6.3.

So actually, the types as they are already is good for people who need types on an old version.

So I am not really requesting changes, per se, but that we don’t update these.

If there is a better way to direct people to the actual package, I’d be willing to do that.

LucasHill:
@happycollision 
So I see that qunit-dom is written in typescript. For whatever reason however, if you do not have @types/qunit-dom in your project, it is unable to infer the types on the qunit assert object when you use "assert.dom". Maybe this is because qunit-dom is essentially patching the qunit assert type? Thoughts?

happycollision:
@LucasHill I think no matter what the cause of the problem you pointed out, the solution should be upstream.

People on the latest versions of qunit-dom shouldn’t be using these types at all.

PR #37381, opened 2019-08-05
Fix up Request.authInfo/user declarations in Passport

vaskevich:
Changing the type from `any` to an interface so that it can be merged as needed by users. I believe this was the intention as there was already a test that was declaring a `User` interface within Express.

### DIFF
--- a/types/passport/index.d.ts
+++ b/types/passport/index.d.ts
@@ -11,15 +11,18 @@
 
 declare global {
     namespace Express {
+        interface AuthInfo {}
+        interface User {}
+
         interface Request {
-            authInfo?: any;
-            user?: any;
+            authInfo?: AuthInfo;
+            user?: User;
 
             // These declarations are merged into express's Request type
-            login(user: any, done: (err: any) => void): void;
-            login(user: any, options: any, done: (err: any) => void): void;
-            logIn(user: any, done: (err: any) => void): void;
-            logIn(user: any, options: any, done: (err: any) => void): void;
+            login(user: User, done: (err: any) => void): void;
+            login(user: User, options: any, done: (err: any) => void): void;
+            logIn(user: User, done: (err: any) => void): void;
+            logIn(user: User, options: any, done: (err: any) => void): void;
 
             logout(): void;
             logOut(): void;
--- a/types/passport/index.d.ts
+++ b/types/passport/index.d.ts
@@ -12,7 +12,9 @@
 
 declare global {
     namespace Express {
+        // tslint:disable-next-line:no-empty-interface
         interface AuthInfo {}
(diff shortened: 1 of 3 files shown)


types/passport/tslint.json
```diff
@@ -3,6 +3,7 @@
     "rules": {
         "no-unnecessary-generics": false,
         "no-void-expression": false,
+        "no-empty-interface": false,
```
vaskevich: Without this, there is always a tslint error, saying to use either an explicit type or an interface.

types/passport/index.d.ts
```diff
@@ -6,20 +6,24 @@
...
 
 declare global {
     namespace Express {
+        interface AuthInfo {}
+        interface User {}
+
         interface Request {
-            authInfo?: any;
-            user?: any;
+            authInfo?: AuthInfo;
```
sheetalkamat: https://github.com/DefinitelyTyped/DefinitelyTyped#common-mistakes
This shouldnt be interface without properties. It needs to be either object or any

sheetalkamat (changes requested):

types/passport/index.d.ts
vaskevich: As mentioned in the PR description, this is very much intentional. The `User` and `AuthInfo` properties can be augmented by users depending on how Express is used, and this can only be done in a type safe matter with an interface. If it's `any` or an `object`, it cannot be extended. For more details, see: https://www.typescriptlang.org/docs/handbook/declaration-merging.html

For our use case, this is leading to type conflicts because we're setting `user` to have an explicit type when using Passport, and the current declaration of `any` is conflicting with it, because it does not support declaration merging. IMO a third-party package updating Express type declarations should play nicely with it, not be greedy with common properties like `user` and `authInfo` by setting it them to something that cannot be extended.

types/passport/tslint.json
sheetalkamat: Do not disable it for the whole package, fix it at the location by adding tslint disable comment. Refer https://github.com/DefinitelyTyped/DefinitelyTyped#lint

sheetalkamat (changes requested):

types/passport/tslint.json
vaskevich: Done.

vaskevich:
If anyone runs into issues from this, see https://github.com/DefinitelyTyped/DefinitelyTyped/commit/91c229dbdb653dbf0da91992f525905893cbeb91#commitcomment-34812715

shilangyu:
Thank you for linking a solution @vaskevich

msrumon:
I wished that the `Request` interface would be a generic... 😥

PR #37382, opened 2019-08-05
Add typings for Figma Plugin API

rudi-c:
This PR adds typings for the the Figma Plugin API that was released last week (I work at Figma).

Note that this is not a typings file for a traditional NPM package with source code on Github. These are the typings file for an API exposed to plugin code when they run inside a special sandbox inside the Figma design tool, on our website.

Note also that there already exists an NPM package with the name `figma`. That package is currently empty and owned by the CTO of Figma @evanw, so that's not a problem even though the two packages are unrelated.

If adding a new definition:

### DIFF
--- a/types/figma/index.d.ts
+++ b/types/figma/index.d.ts
@@ -3,6 +3,11 @@
 // Definitions by: Figma <https://www.figma.com>
 //                 Rudi Chen <rudi@figma.com>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+//
+// Talk to the Figma team regarding changes to the typings. While
+// changes to the types aren't breaking changes to how plugins run,
+// it is preferrable to coordinate break type changes with breaking
+// API changes.
 
 // Global variable with Figma's plugin API.
 declare const figma: PluginAPI
--- a/types/figma/index.d.ts
+++ b/types/figma/index.d.ts
@@ -1,0 +1,661 @@
+// Type definitions for Figma API v1.0.0
+// Project: https://www.figma.com/plugin-docs/intro
+// Definitions by: Figma <https://www.figma.com>
+//                 Rudi Chen <rudi@figma.com>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+// Global variable with Figma's plugin API.
+declare const figma: PluginAPI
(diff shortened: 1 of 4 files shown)


types/figma/tslint.json
```diff
@@ -0,0 +1,10 @@
+{
+    "extends": "dtslint/dt.json" ,
+    "rules": {
+        "dt-header": false,
+        "npm-naming": false,
+        "no-any-union": false,
```
sandersn: New packages should not override any lint rules. You explained why dt-header and npm-naming don't apply (thanks! it was a very clear explanation), but please remove the other overrides.

types/figma/index.d.ts
```diff
@@ -0,0 +1,666 @@
...
+
+    createImage(data: Uint8Array): Image
+    getImageByHash(hash: string): Image
+
+    group(nodes: ReadonlyArray<BaseNode>, parent: BaseNode & ChildrenMixin, index?: number): FrameNode
+    flatten(nodes: ReadonlyArray<BaseNode>, parent?: BaseNode & ChildrenMixin, index?: number): VectorNode
+}
+
+interface ClientStorageAPI {
+    getAsync(key: string): Promise<any | undefined>
```
sandersn: `unknown` is better than `any | undefined` here, unless you want to explicitly exclude `null`, in which case `{} | undefined` is the shortest type you could write. But `object | string | number | boolean | symbol | bigint | undefined` might be easier to read.

sandersn (changes requested):
Need to remove some lint rule overrides.

chrisdrackett:
@rudi-c do these types cover the Figma Rest API as well? If not, are their plans to officially publish these? I was thinking of submitting https://github.com/jongold/figma-js/blob/master/src/figmaTypes.ts to DefinitelyTyped but will hold off if there are official plans!

rudi-c:
Ah my bad, I made the changes but then pushed the wrong branch. Re-opened here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37748

@chrisdrackett No, not sure if it makes sense to have both the plugin API types and the REST API types in the same package. It doesn't seem like anyone on the infra team is currently planning on publishing types for the API.

PR #37400, opened 2019-08-06
Simplify .editorconfig & .prettierrc.json

demensky:
Move some prettier options to editorconfig, because [prettier supports editorconfig].

[prettier supports editorconfig]: https://prettier.io/blog/2017/12/05/1.9.0.html#add-editorconfig-support-3255-https-githubcom-prettier-prettier-pull-3255-by-josephfrazier-https-githubcom-josephfrazier

### DIFF
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,20 +1,6 @@
 {
     "$schema": "http://json.schemastore.org/prettierrc",
-
-    "parser": "typescript",
-    "printWidth": 120,
-    "semi": true,
     "singleQuote": true,
-    "tabWidth": 4,
-    "trailingComma": "es5",
-
-    "overrides": [
-        {
-            "files": ["*.md"],
-            "options": {
-                "tabWidth": 2
-            }
-        }
-    ]
+    "trailingComma": "es5"
 }
 
--- a/.prettierrc.json
+++ b/.prettierrc.json
@@ -1,6 +1,6 @@
 {
     "$schema": "http://json.schemastore.org/prettierrc",
     "singleQuote": true,
-    "trailingComma": "es5"
+    "trailingComma": "all"
 }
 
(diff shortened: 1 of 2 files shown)


.prettierrc.json
```diff
@@ -1,20 +1,6 @@
 {
     "$schema": "http://json.schemastore.org/prettierrc",
-
-    "parser": "typescript",
-    "printWidth": 120,
-    "semi": true,
```
demensky: [Default is `true`](https://prettier.io/docs/en/options.html#semicolons).

.prettierrc.json
```diff
@@ -1,20 +1,6 @@
...
     "singleQuote": true,
-    "tabWidth": 4,
-    "trailingComma": "es5",
-
-    "overrides": [
-        {
-            "files": ["*.md"],
-            "options": {
-                "tabWidth": 2
```
demensky: Why for `.md` we use 2 spaces and not 4?

.prettierrc.json
```diff
@@ -1,20 +1,6 @@
 {
     "$schema": "http://json.schemastore.org/prettierrc",
-
-    "parser": "typescript",
-    "printWidth": 120,
-    "semi": true,
     "singleQuote": true,
-    "tabWidth": 4,
-    "trailingComma": "es5",
```
demensky: Let's use `"all"`?

demensky:
@sandersn, what do you think about this?

sandersn:
Tagging @orta since he knows prettier.

.prettierrc.json
orta: "all" required a newer version of node to run by defautl - but given these are all parsed by TS, that seems fine 👍

.prettierrc.json
orta: Probably a left-over from the original template. Seems fine to me to just be the same as everything else.

orta (approved):
Cool, yep - this makes sense to me.

demensky:
@orta. Should I update by your comments?

PR #37403, opened 2019-08-06
[@types/randomColor] Make return type smarter

nixxquality:

This makes the return type `string` by default and automatically turns it into `string[]` if a count argument is supplied.
This is in accordance with the behaviour of the package.
```
> f = require("randomcolor")
[Function: randomColor] { randomColor: [Circular] }
> f()
'#efa977'
> f({count:1})
[ '#f90602' ]
> f({count:2})
[ '#edda61', '#47d170' ]
```

I've also modified the tests to test this behaviour.


nixxquality:
@andrewbranch Is the bot not going to mark this as awaiting reviewer feedback now?
It's pointing out the 🚨but there's no way to say that yes, it is ok?

andrewbranch:
@nixxquality the labeling process runs on a schedule; the perf comment runs at the end of a build and doesn’t contribute to the labeling.

It looks like the 🚨 is because the subtype cache size increased by _Infinity%_—from 0 to 8 😄 

> don’t panic if I say something crazy!

I’ll add it to my list of bugs 😁

feitzi (approved):
Looks fine. Good idea!

PR #37409, opened 2019-08-06
[ramda] Removed utility types in `tools.d.ts` to integrate with "ts-toolbelt" instead

millsp:
Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/pirix-gh/DefinitelyTyped/blob/86a43f47d8212707b2305f0eaca462f0d4c774d5/types/ramda/index.d.ts#L1418
I also brought out curried types for `R.flip` like @imcotton wanted to do

If removing a declaration:

### DIFF
--- a/types/ramda/index.d.ts
+++ b/types/ramda/index.d.ts
@@ -543,6 +543,8 @@
 /// <reference path="./src/zipWith.d.ts" />
 /// <reference path="./src/includes.d.ts" />
 
+import {A, F, T} from "ts-toolbelt"
+
 declare let R: R.Static;
 
 declare namespace R {
--- a/types/ramda/index.d.ts
+++ b/types/ramda/index.d.ts
@@ -543,7 +543,7 @@
 /// <reference path="./src/zipWith.d.ts" />
 /// <reference path="./src/includes.d.ts" />
 
-import {A, F, T} from "ts-toolbelt"
+import { A, F, T } from "ts-toolbelt";
 
 declare let R: R.Static;
 
--- a/types/ramda/index.d.ts
+++ b/types/ramda/index.d.ts
@@ -30,7 +30,7 @@
 //                 Aram Kharazyan <https://github.com/nemo108>
 //                 Jituan Lin <https://github.com/jituanlin>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 3.3
+// TypeScript Version: 3.5
 
 /// <reference path="./es/add.d.ts" />
 /// <reference path="./es/addIndex.d.ts" />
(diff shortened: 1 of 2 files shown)


pirix-gh:
Hi @sandersn,

Travis won't allow me to publish, says "Dependency ts-toolbelt not in whitelist". But it is.

Any insights?

sandersn:
Argh, travis' cache must not understand github references in package.json any more. I reset the caches for now, but I need to find out whether it's a bug in Travis or intended behaviour.

pirix-gh:
Couldn't find any info on this topic, it's most likely not working as intended. Thanks!

pirix-gh:
Hey @andrewbranch, what does "CV" stand for in "Mean CV" for the ts bot?

Nemo108 (approved):

andrewbranch:
@pirix-gh [Coefficient of variation](https://en.wikipedia.org/wiki/Coefficient_of_variation). It’s the _mean_ because we take multiple samples at every identifier location (time allowing—we limit each run to an hour), so the ideal case is something like

- there are 5287 identifiers
- there are 5 measurements of how long it takes to get completions for each of those
- in a perfect world, every measurement in any set of 5 would be identical, because we were measuring the exact same thing
- it's not a perfect world, so we take the CV of those 5 samples to see how imperfect the world is
- at the end, we have 5287 CV values that represent how internally consistent our measurements were, so we average them and report it

It’s a metric that’s entirely about whether or not we can trust the other numbers; it has nothing to do with the performance of the package itself.

_However_—we _did_ run out of time on these measurements, so for most of those 5287 identifiers, we only have _one_ sample, which means a CV of zero.

There are two things highly suspicious about this reporting, which I have only just noticed, thanks to your question. It is either a _remarkable_ coincidence, or much more likely, a bug, that the number of identifiers in tests is exactly the number of samples we took both for master and for the PR. The ideal case would be `samples = identifiers * 5`, but if we hit the time limit, I would expect to see a random number between 0 and `identifiers * 5`. There’s absolutely no reason why it should be `samples == identifiers`, so very likely either we’re printing the wrong number to the table, or something is weird with the actual measurement algorithm and it’s not starting a second pass.

The other suspicious thing is that the CV _isn’t_ zero, despite the fact that it essentially told us that every identifier got exactly one measurement. It sounds more like _most_ identifiers got one sample, and some got two.

Combined, this leads me to believe that we’re _probably_ just mis-reporting either the number of samples taken, which is actually an important metric to have to understand whether or not Mean CV means anything.

Thanks for the question; I’m adding this to my list of bugs to fix soon! 😄

PR #37421, opened 2019-08-07
Add type definitions for `react-inspector`

rogerclotet:
Type definitions for https://github.com/storybookjs/react-inspector

If adding a new definition:

### DIFF
--- a/types/react-inspector/index.d.ts
+++ b/types/react-inspector/index.d.ts
@@ -2,7 +2,7 @@
 // Project: http://formatjs.io/react/, https://github.com/yahoo/react-intl
 // Definitions by: Roger Clotet <https://github.com/rogerclotet>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 3.5
+// TypeScript Version: 2.8
 
 import * as React from 'react';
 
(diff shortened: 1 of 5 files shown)


types/react-inspector/index.d.ts
```diff
@@ -0,0 +1,81 @@
+// Type definitions for react-inspector 3.0
+// Project: http://formatjs.io/react/, https://github.com/yahoo/react-intl
+// Definitions by: Roger Clotet <https://github.com/rogerclotet>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.5
```
sandersn: could this work on lower versions of typescript? It would make this package accessible to people who aren't on new versions of TS.

The CI will tell you if the version you set doesn't work, or you can run `npm run lint react-inspector`. The version can go back to 2.0, but I'm pretty sure React itself requires a higher version than that.

sandersn (changes requested):

types/react-inspector/index.d.ts
rogerclotet: It actually requires 3.5 because I used `Omit`. I will try to define its types omitting `Omit` ;)

rogerclotet: Defined a new interface `DomInspectorProps` to be able to remove the `Omit` usage while keeping the `data` type correct.

sandersn (approved):

PR #37426, opened 2019-08-07
react, react-test-renderer: Update React types for 16.9.0

Jessidhia:
For now I only quickly went through the changes in the changelog at https://github.com/facebook/react/pull/16254. Will check for more things on the entry points later.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/react/pull/16254

### DIFF
--- a/types/react-test-renderer/index.d.ts
+++ b/types/react-test-renderer/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for react-test-renderer 16.8
+// Type definitions for react-test-renderer 16.9
 // Project: https://facebook.github.io/react/
 // Definitions by: Arvitaly <https://github.com/arvitaly>
 //                 Lochbrunner <https://github.com/lochbrunner>
@@ -60,6 +60,20 @@ export function create(nextElement: ReactElement, options?: TestRendererOptions)
  * more closely by executing pending `useEffect`s before returning. This also
  * reduces the amount of re-renders done.
  *
+ * @param callback An asynchronous, void callback that will execute as a single, complete React commit.
+ *
+ * @see https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#testing-hooks
+ */
+// the "void | undefined" is here to forbid any sneaky return values
+// tslint:disable-next-line: void-return
+export function act(callback: () => Promise<void | undefined>): Promise<undefined>;
+/**
+ * Wrap any code rendering and triggering updates to your components into `act()` calls.
+ *
+ * Ensures that the behavior in your tests matches what happens in the browser
+ * more closely by executing pending `useEffect`s before returning. This also
+ * reduces the amount of re-renders done.
+ *
  * @param callback A synchronous, void callback that will execute as a single, complete React commit.
  *
  * @see https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#testing-hooks
@@ -71,6 +85,9 @@ export function act(callback: () => void | undefined): DebugPromiseLike;
 // Intentionally doesn't extend PromiseLike<never>.
 // Ideally this should be as hard to accidentally use as possible.
 export interface DebugPromiseLike {
-    // the actual then() in here is 0-ary, but that doesn't count as a PromiseLike.
-    then(onfulfilled: (value: never) => never, onrejected: (reason: never) => never): never;
+  // the actual then() in here is 1-ary, but that doesn't count as a PromiseLike.
+  then(
+    onfulfilled: (value: never) => never,
+    onrejected: (reason: never) => never,
+  ): never;
 }
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for React 16.8
+// Type definitions for React 16.9
 // Project: http://facebook.github.io/react/
 // Definitions by: Asana <https://asana.com>
 //                 AssureSign <http://www.assuresign.com>
@@ -349,13 +349,6 @@ declare namespace React {
 
         /** A fallback react tree to show when a Suspense child (like React.lazy) suspends */
         fallback: NonNullable<ReactNode>|null;
-
-        // I tried looking at the code but I have no idea what it does.
-        // https://github.com/facebook/react/issues/13206#issuecomment-432489986
-        /**
-         * Not implemented yet, requires unstable_ConcurrentMode
-         */
-        // maxDuration?: number;
     }
     /**
      * This feature is not yet available for server-side rendering.
@@ -382,7 +375,7 @@ declare namespace React {
         onRender: ProfilerOnRenderCallback;
     }
 
-    const unstable_Profiler: ExoticComponent<ProfilerProps>;
+    const Profiler: ExoticComponent<ProfilerProps>;
 
     //
     // Component API
@@ -2199,6 +2192,7 @@ declare namespace React {
         playsInline?: boolean;
         poster?: string;
         width?: number | string;
+        disablePictureInPicture?: boolean;
     }
 
     // this list is "complete" in that it contains every SVG attribute
--- a/types/react/test/tsx.tsx
+++ b/types/react/test/tsx.tsx
@@ -321,7 +321,7 @@ const ForwardRef3 = React.forwardRef(
 <ForwardRef3 ref={divFnRef}/>;
 <ForwardRef3 ref={divRef}/>;
 
-const Profiler = React.unstable_Profiler;
+const { Profiler } = React;
 
 // 'id' is missing
 <Profiler />; // $ExpectError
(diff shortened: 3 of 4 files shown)


types/react-test-renderer/index.d.ts
```diff
@@ -53,6 +53,20 @@ export interface TestRendererOptions {
...
+ * more closely by executing pending `useEffect`s before returning. This also
+ * reduces the amount of re-renders done.
+ *
+ * @param callback An asynchronous, void callback that will execute as a single, complete React commit.
+ *
+ * @see https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#testing-hooks
+ */
+// the "void | undefined" is here to forbid any sneaky return values
+// tslint:disable-next-line: void-return
+export function act(callback: () => Promise<void | undefined>): Promise<undefined>;
```
Jessidhia: `() => Promise<void>` would essentially accept `Promise<any>`, so it's `|`ed with `undefined` to enforce no values are returned. Just `Promise<undefined>` would reject `void` functions.

types/react/index.d.ts
```diff
@@ -349,13 +349,6 @@ declare namespace React {
 
         /** A fallback react tree to show when a Suspense child (like React.lazy) suspends */
         fallback: NonNullable<ReactNode>|null;
-
-        // I tried looking at the code but I have no idea what it does.
-        // https://github.com/facebook/react/issues/13206#issuecomment-432489986
-        /**
-         * Not implemented yet, requires unstable_ConcurrentMode
-         */
-        // maxDuration?: number;
```
Jessidhia: Both this and unstable_ConcurrentMode were removed. unstable_ConcurrentMode was never added to the types anyway.

types/react/index.d.ts
```diff
@@ -382,7 +375,7 @@ declare namespace React {
         onRender: ProfilerOnRenderCallback;
     }
 
-    const unstable_Profiler: ExoticComponent<ProfilerProps>;
+    const Profiler: ExoticComponent<ProfilerProps>;
```
Jessidhia: Just renamed.

types/react/index.d.ts
```diff
@@ -2199,6 +2192,7 @@ declare namespace React {
         playsInline?: boolean;
         poster?: string;
         width?: number | string;
+        disablePictureInPicture?: boolean;
```
Jessidhia: https://github.com/facebook/react/pull/15334/files#diff-e19827754462ba391be5fc99b7a3ca09R297

Added as a `boolean` property.

types/react/test/tsx.tsx
```diff
@@ -357,8 +357,14 @@ const Profiler = React.unstable_Profiler;
 </Profiler>;
 
 type ImgProps = React.ComponentProps<'img'>;
-// $ExpectType "async" | "auto" | "sync" | undefined
-type ImgPropsDecoding = ImgProps['decoding'];
+const imgProps: ImgProps = {};
+// the order of the strings in the union seems to vary
+// with the typescript version, so test assignment instead
```
Jessidhia: Sometimes it's `"async" | "auto" | "sync" | undefined`, sometimes it's `"auto" | "async" | "sync" | undefined`. Not sure why. It doesn't matter for type checking, but dtslint cares about it.

Jessidhia:
I thought of adding this as an additional file inside the project, but I couldn't find any way that works well. dtslint doesn't like module augmentation within the same module, and when I get something that does appear to work with `@next`, it breaks with certain versions of typescript.

```ts
// Definitions for unstable APIs present in React
// Current to 16.9.0-rc.0

declare namespace React {
    //#region unflagged features
    export interface UnstableSuspenseListPropsCommon {
        // ReactFiberSuspenseComponent:SuspenseListTailMode
        // ReactFiberBeginWork:validateTailOptions
        tail?: 'collapsed' | 'hidden';
    }
    export interface UnstableSuspenseListPropsUnordered extends UnstableSuspenseListPropsCommon {
        // ReactFiberBeginWork:validateRevealOrder
        revealOrder?: 'together';
        // ReactFiberBeginWork:validateSuspenseListChildren
        children?: ReactNode;
    }
    export interface UnstableSuspenseListPropsOrdered extends UnstableSuspenseListPropsCommon {
        // children are only validated when the revealOrder is either of these values

        // ReactFiberBeginWork:validateRevealOrder
        // ReactFiberBeginWork:validateSuspenseListChildren
        revealOrder: 'forwards' | 'backwards';
        // ReactFiberBeginWork:validateSuspenseListChildren
        // Array works too but this accepts any Iterable, including Array
        children?: Iterable<ReactChild | undefined | null | false> | null | false;
    }

    export type UnstableSuspenseListProps = UnstableSuspenseListPropsOrdered | UnstableSuspenseListPropsUnordered;

    export const unstable_SuspenseList: ExoticComponent<UnstableSuspenseListProps>;

    export interface UnstableSuspenseConfig {
        timeoutMs: number;
        busyDelayMs?: number;
        busyMinDurationMs?: number;
    }

    // https://github.com/facebook/react/pull/15593
    export function unstable_withSuspenseConfig(
        scope: () => void | undefined,
        config?: UnstableSuspenseConfig | null,
    ): void;
    //#endregion

    //#region enableFlareAPI
    // https://github.com/facebook/react/blob/v16.9.0-rc.0/packages/shared/forks/ReactFeatureFlags.www.js#L73
    // NOTE: the actual internal implementations are on the reconciler,
    // so this uses the react-dom reconciler as the reference right now.
    // only HostComponent (i.e. DOM nodes) can receive listeners
    // TODO: requires declaring react-events package
    //#endregion

    //#region enableFundamentalAPI
    // https://github.com/facebook/react/blob/v16.9.0-rc.0/packages/shared/forks/ReactFeatureFlags.www.js#L75
    // https://github.com/facebook/react/pull/16049
    // "Note: this is a completely private API for internal experimentation use only",
    // and its flag is turned off, so not defined.
    //#endregion

    //#region enableJSXTransformAPI
    // https://github.com/facebook/react/blob/v16.9.0-rc.0/packages/shared/forks/ReactFeatureFlags.www.js#L77
    // implementation of https://github.com/reactjs/rfcs/pull/107
    export function jsx<C extends ElementType>(
        type: C,
        config: JSX.LibraryManagedAttributes<C, ComponentPropsWithRef<C>>,
        maybeKey?: Key,
    ): ReactElement<ComponentPropsWithRef<C>, C>;
    // TODO: implement the difference between jsx and jsxs
    // (there isn't one, really, but jsxs is more optimized when multiple children are passed)
    export { jsx as jsxs };
    /**
     * NOTE: undefined in production builds
     */
    export function jsxDEV<C extends ElementType>(
        type: C,
        config: JSX.LibraryManagedAttributes<C, ComponentPropsWithRef<C>>,
        maybeKey?: Key,
        source?: { fileName: string; lineNumber: number },
        self?: C,
    ): ReactElement<ComponentPropsWithRef<C>, C>;
    //#endregion

    //#region enableSuspenseCallback
    // https://github.com/facebook/react/blob/v16.9.0-rc.0/packages/shared/forks/ReactFeatureFlags.www.js#L81
    export interface SuspenseProps {
        callback?(queuedUpdates: ReadonlySet<PromiseLike<any>>): void;
    }
    //#endregion
}
```

Jessidhia:
It looks like `npm run test` just fails everywhere because of a race condition _within_ `npm`.

types/react-test-renderer/index.d.ts
ferdaber: Does this affect the test renderer runtime at all if the Promise resolves to a defined value? I think some developers like to combine early returns with an evaluation (still not caring about the return value)

```ts
if (somethingThatMakesMeReturnEarly) {
  return doSomethingElse()
}
```

sandersn:
I restarted the build for now, but I probably need to ship the make-npm-install-single-threaded fix from dtslint-runner to types-publisher as well.

Jessidhia:
16.9.0 is released so releasing this :tada:

types/react-test-renderer/index.d.ts
diegohaz: What's the problem with `Promise<any>`?

I'm incurring into an error when trying to do something like this:
```js
await act(() =>
  expect(result.current.validate()).rejects.toEqual({ a: "error" })
);
```
`expect(somePromise).rejects.toEqual()` will return the same type returned by `somePromise`. Test passes, but TypeScript is complaining: `Type 'Promise<SomePromiseReturn>' is not assignable to type 'void'.`

eps1lon: @diegoazh Please open an issue with the full error message. May also be an issue with the jest typings. 

In any case: I would keep it simply an just `await` `result.current.validate()` `toThrow` and wrap it in a callback that's just `void`. It's not obvious anyway that `expect.rejects` returns a promise which makes this `act` harder to read than necessary.

PR #37437, opened 2019-08-07
Slonik - add support for json token type

ivank:
Add JSON helper support.
It's been available for quite some time but wasn't present in the types.
I've opted to use "any" instead of the more rigid Serializable, as json can serialize a very diverse set of objects.
What do you guys think? @gajus @sebald @mmkal 

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/gajus/slonik/tree/d511ede91da9466864eafaaaa3efbc7e9a22fde1#slonik-query-building-sql-json

### DIFF
--- a/types/slonik/index.d.ts
+++ b/types/slonik/index.d.ts
@@ -26,18 +26,7 @@ export type LogicalBooleanOperatorType = 'AND' | 'OR';
 // EXPRESSIONS AND TOKENS
 // ----------------------------------------------------------------------
 
-export type SerializableValueType = 
-    | string 
-    | number 
-    | boolean 
-    | SerializableValueObject 
-    | SerializableValueArray;
-
-interface SerializableValueObject {
-    [x: string]: SerializableValueType;
-}
-
-interface SerializableValueArray extends Array<SerializableValueType> { }
+export type SerializableValueType = any;
 
 export interface IdentifierTokenType {
     names: ReadonlyArray<string>;
@@ -119,6 +108,7 @@ export type SqlTokenType =
     AssignmentListTokenType |
     IdentifierTokenType |
     IdentifierListTokenType |
+    JsonSqlTokenType |
     RawSqlTokenType |
     SqlSqlTokenType<any> |
     TupleListSqlTokenType |
--- a/types/slonik/index.d.ts
+++ b/types/slonik/index.d.ts
@@ -26,7 +26,21 @@ export type LogicalBooleanOperatorType = 'AND' | 'OR';
 // EXPRESSIONS AND TOKENS
 // ----------------------------------------------------------------------
 
-export type SerializableValueType = any;
+export type SerializableValueType =
+    | string
+    | number
+    | boolean
+    | null
+    | object
+    | SerializableValueObject
+    | SerializableValueArray;
+
+export interface SerializableValueObject {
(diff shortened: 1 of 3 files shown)


gajus:
This is how JSON payload is defined in Slonik:

```js
type SerializableValueType = string | number | boolean | null | {+[key: string]: SerializableValueType} | $ReadOnlyArray<SerializableValueType>;

```

ivank:
Yeah I've noticed that, with TS its a a bit weird, but I've found how that can be expressed with it here https://github.com/microsoft/TypeScript/issues/3496#issuecomment-128553540 certainly doable.

However it would forbid putting interfaces without index types (which are indeed totally ok for `JSON.stringify()`) as well as disallowing a lot of other types as well (like Date), which would be stringified without any issue.

So I opted for `any` since that's the type that's actually used on `JSON.stringify()` officially. 
What do you think is it a valid assumption? 

I did try it out with https://github.com/microsoft/TypeScript/issues/3496#issuecomment-128553540 for my own code and it wasn't actually ergonomic

mmkal (changes requested):
This should have at least a test with the sample usage from the docs:

```typescript
await connection.query(sql`
  SELECT (${sql.json([1, 2, 3])})
`);
 ```

~~But also, according to [the docs](https://www.npmjs.com/package/slonik#sqljson), passing `undefined` in will throw an error. So `any` probably isn't the right choice. We could instead use a union type of the values that are allowed (something like what's done in https://www.npmjs.com/package/unknown-ts). There should also be a test for `undefined` not being allowed.~~

Edit: I see there was a bit of discussion of whether we can do better than `any` already. I'm curious what was meant by this:

> However it would forbid putting interfaces without index types 

This is another case where a test would be helpful. Something that https://github.com/microsoft/TypeScript/issues/3496#issuecomment-128553540 can't handle.

**Comment by ivank** [comment]
Yeah this is totally TS's fault I think.  For example typescript would not allow us to do this, if we used https://github.com/microsoft/TypeScript/issues/3496#issuecomment-128553540

```typescript
interface StatementData {
  id: string;
  other: number;
}
const data: StatementData = { id: '123', other: 12 }

await connection.query(sql`SELECT (${sql.json(data)})`);
```

Because `StatementData` doesn't have an index signature. Would need to change it everywhere to:

```
interface StatementData {
  id: string;
  other: number;
  [id: string]: string | number;
}
```
Which looks a bit silly to me.

I'm up for adding tests of course, but I don't think you can do so within DefinitlyTyped right?

mmkal:
@ivank if you add ` | object` to the `type JSONValue = ...` definition, your example will work fine. The example would be a good test, as well as a negative test for `undefined`. The test file is called `slonik-test.ts`, it's next door to the `index.d.ts` file you've modified. DefinitelyTyped will run the type checker and linter on anything you put in there. There are some examples in there for invalid code producing typescript errors, e.g. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/00b85c2f5d62a22b667060a7d98bad8e18d18fb5/types/slonik/slonik-tests.ts#L67-L69

ivank:
🤦‍♂ 10x @mmkal

mmkal (approved):
Looks good!

PR #37443, opened 2019-08-07
lodash: infer return type of flattenDeep when using _.chain

schmod:
Provides the correct return type when `_.chain(value)` and `_(value)` are used with `flattenDeep`.  

```ts
const a: number[][][][] = [[[[1]]]];
const b = _.chain(a)
  .flattenDeep()
  .value(); // currently typed as number[][][][] – should be number[]
```

Note that this only works with up to 4 levels of nesting, due to an internal TS design constraint that prevents types from recursing more than 5 levels deep: https://github.com/microsoft/TypeScript/issues/23024#issuecomment-377711022

---

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://lodash.com/docs/4.17.15#flattenDeep


types/lodash/ts3.1/common/array.d.ts
```diff
@@ -488,6 +488,7 @@ declare module "../index" {
          */
         flatten(): T extends Many<infer U> ? CollectionChain<U> : CollectionChain<T>;
     }
+    type Flat<T> = (T extends List<infer U> ? never : T);
```
aj-r: Why is this needed for chains but not other versions?

Also don't `infer U` if you never use it. Instead just use `any`.

types/lodash/ts3.1/lodash-tests.ts
```diff
@@ -497,12 +497,12 @@ _.chain([1, 2, 3, 4]).unshift(5, 6); // $ExpectType CollectionChain<number>
     _.flattenDeep<number>({0: 1, 1: [2, [3, [4, 5]]], length: 2}); // $ExpectType number[]
 
     _([1, 2, 3]).flattenDeep(); // $ExpectType Collection<number>
-    _([1, [2, [3, [4, 5]]]]).flattenDeep(); // $ExpectType Collection<number | (number | (number | number[])[])[]>
-    _({0: 1, 1: [2, [3, [4, 5]]], length: 2}).flattenDeep(); // $ExpectType Collection<number | (number | (number | number[])[])[]>
+    _([1, [2, [3, [4, 5]]]]).flattenDeep(); // $ExpectType Collection<number>
```
aj-r: It might be nice to test some other cases too. Specifically I'm worried about a case like `_([[[1, 2]]])` resulting in `Collection<never>`

aj-r (changes requested):

types/lodash/ts3.1/lodash-tests.ts
schmod: I've added this test case.

types/lodash/ts3.1/common/array.d.ts
schmod: I've switched `infer U` to `any`.  

I didn't realize that the static version wasn't working until I noticed that the tests were all asserting the final type via a generic.  I've added this workaround to the `LodashStatic` variant of `_.flattenDeep()` and updated the tests to add cases where the explicit generics are not provided.

types/lodash/ts3.1/lodash-tests.ts
```diff
@@ -492,19 +492,30 @@ _.chain([1, 2, 3, 4]).unshift(5, 6); // $ExpectType CollectionChain<number>
 
 // _.flattenDeep
 {
+    _.flattenDeep([1, 2, 3]); // $ExpectType number[]
+    _.flattenDeep([[1, 2, 3]]); // $ExpectType number[]
+    _.flattenDeep([1, [2, [3, [4, 5]]]]); // $ExpectType number[]
+    _.flattenDeep([1, [2, [3, [4, 5]]]]); // $ExpectType number[]
```
aj-r: Duplicate?

aj-r (changes requested):
Looks good, just 1 nitpick

aj-r (approved):

stocksr:
Not sure if this is the best place for this but this release broke my code when using arrays of strings 
I am getting a return type of `never[]`  from a input type of `(string | string[])[]`

Looking at this PR I think adding and fixing these test cases might be the way forward.
```
_.flattenDeep(['x']); // $ExpectType string[]
_.flattenDeep(['x',['y']]); // $ExpectType string[]
```

aj-r:
Weird... maybe it's confused because `string` also implements `ArrayLike<string>`? If that's the case, then we may need to add a special overload for strings.

In the meantime, does this workaround work?

`_.flattenDeep<string>(['x',['y']]);`

aj-r:
Ah I think I see the problem. We probably need to change:

` type Flat<T> = (T extends List<any> ? never : T);`

to:

` type Flat<T> = T extends string ? T : (T extends List<any> ? never : T);`

stocksr:
Hi,  Finally got back around to looking at this again.

> In the meantime, does this workaround work?
> 
> `_.flattenDeep<string>(['x',['y']]);`

No sorry did not work

> Ah I think I see the problem. We probably need to change:
> 
> ` type Flat<T> = (T extends List<any> ? never : T);`
> 
> to:
> 
> ` type Flat<T> = T extends string ? T : (T extends List<any> ? never : T);`


made this change in my local node_modules folder and this works great.

is there an issue tracker I should move this to in-order to get it changed?

aj-r:
Sure, you can create a bug in DefinitelyTyped with "lodash" in the title. Or better yet, create a PR that implements this change (it will get fixed much faster that way).

PR #37447, opened 2019-08-07
[react-select] Export types from src/types.ts for use by package consumers

### LINKED ISSUE #37444: [react-select] Useful types are not exported
Authors: @claasahl @jonfreedman @Methuselah96

Types like `ValueType`, `OptionProps`, `Theme` are not exported from the top level package.  Importing them from the inner packages causes some issues because the JavaScript code tries to import those paths as well.

The fix is to figure out which types should be exported for use and add them to the index.d.ts

dobesv:
The types in src/types are useful to users of react-select who want to write custom `Option` components or reuse those types and interfaces in their own props for any reason.  The main purpose of this PR is to export those types from the top level of the package so they can be used safely.

Importing them from a subpath in the package confuses non-typescript processors who are looking at the implementation files instead of the typings files.

This also has a fix for `createFilter` to make it generic to option type, and merges two slightly different definitions of `OptionProps` into one shared definition.

Fixes https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37444

### DIFF
--- a/types/react-select/index.d.ts
+++ b/types/react-select/index.d.ts
@@ -10,6 +10,33 @@ import { StateManager } from './src/stateManager';
 
 export default StateManager;
 
-export { createFilter } from './src/filters';
+export { createFilter, Config as CreateFilterConfig, Option as CreateFilterOption } from './src/filters';
 export { components } from './src/components/index';
-export { mergeStyles } from './src/styles';
+export { mergeStyles, Styles, StylesConfig, GetStyles } from './src/styles';
+export {
+    ActionMeta,
+    ActionTypes,
+    ClassNameList,
+    ClassNamesState,
+    CommonProps,
+    FocusDirection,
+    FocusEventHandler,
+    GroupedOptionsType,
+    GroupType,
+    InnerRef,
+    InputActionMeta,
+    InputActionTypes,
+    KeyboardEventHandler,
+    MenuPlacement,
+    MenuPosition,
+    MouseEventHandler,
+    OptionProps,
+    OptionsInnerProps,
+    OptionStateProps,
+    OptionsType,
+    PropsWithInnerRef,
+    PropsWithStyles,
+    Theme,
+    ThemeSpacing,
+    ValueType,
+} from './src/types';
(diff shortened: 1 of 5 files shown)


Methuselah96 (changes requested):
I'm not sure what you mean by: "Importing them from a subpath in the package confuses non-typescript processors who are looking at the implementation files instead of the typings files." Can you give an example or a sample repo that has this issue? I use `import { ValueType } from 'react-select/src/types';` without any problems using a pretty standard webpack setup.

dobesv:
I would prefer to import from "react-select", it feels less like I'm using
some internal detail of how the typescript package source tree is
organized. But if you're against it just close the PR.  I eventually found
okay workarounds using the import you suggested and typeof.


On Aug 7, 2019 at 8:06 PM, Nathan Bierema <notifications@github.com> wrote:

*@Methuselah96* requested changes on this pull request.

I'm not sure what you mean by: "Importing them from a subpath in the
package confuses non-typescript processors who are looking at the
implementation files instead of the typings files." Can you give an example
or a sample repo that has this issue? I use import { ValueType } from
'react-select/src/types'; without any problems using a pretty standard
webpack setup.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37447?email_source=notifications&email_token=AACQBGKF7FEU67Q3DJWFERTQDOERXA5CNFSM4IKEWU5KYY3PNVWWK3TUL52HS4DFWFIHK3DMKJSXC5LFON2FEZLWNFSXPKTDN5WW2ZLOORPWSZGOCA5VGEQ#pullrequestreview-272323346>,
or mute the thread

Methuselah96:
@dobesv My main push back is that we are trying to mirror the Flow types that are included in `react-select` as much as possible. Since there is also a [`react-select/src/types.js`](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/types.js) that includes Flow types, this is mirrored similarly in the TypeScript definitions.

I'm not able to close this PR since I'm not the author of the PR, but I don't think I would be willing to approve it unless you have some more compelling arguments for wanting this.

PR #37448, opened 2019-08-07
[react-select] Use a common interface for OptionProps

dobesv:
Previously there were two types named `OptionProps`, this merges them
into one.

Otherwise you might run into problems where you want to create a component that forwards props into `OptionProps` but the types don't match up. e.g.

```
import { components } from 'react-select';
import { OptionProps } from 'react-select/src/types';

const MyCustomOption: ComponentType<OptionProps> = props => {
    return (
      <components.Option {...props} label={bundle.title}>
          Hello {props.data.name}        
      </components.Option>
    );
  };
```

In this case you get a type error because `components.Option` accepts a different kind of `OptionProps` than the one you can import from.

The workaround (without this fix) is to use `props as any` when passing to `components.Option` but it's not ideal.

### DIFF
--- a/types/react-select/src/components/Option.d.ts
+++ b/types/react-select/src/components/Option.d.ts
@@ -1,44 +1,10 @@
-import { ComponentType, ReactNode, MouseEventHandler } from 'react';
+import { CSSProperties, ComponentType } from 'react';
+import { OptionProps, OptionStateProps } from '../types';
 
-import { colors, spacing } from '../theme';
-import { CommonProps, PropsWithStyles, InnerRef } from '../types';
-
-interface State {
-  /** Whether the option is disabled. */
-  isDisabled: boolean;
-  /** Whether the option is focused. */
-  isFocused: boolean;
-  /** Whether the option is selected. */
-  isSelected: boolean;
-}
-interface InnerProps {
-  id: string;
-  key: string;
-  onClick: MouseEventHandler<HTMLDivElement>;
-  onMouseMove: MouseEventHandler<HTMLDivElement>;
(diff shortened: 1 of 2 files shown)


dobesv:
Another workaround is to declare the component as `typeof components.Option`, e.g. 

```
import { components } from 'react-select';
import { OptionProps } from 'react-select/src/types';

const MyCustomOption: typeof components.Option = props => {
    return (
      <components.Option {...props} label={bundle.title}>
          Hello {props.data.name}        
      </components.Option>
    );
  };
```

PranavSenthilnathan:
Can one of the owners please take a look: @claasahl @jonfreedman @Methuselah96

Methuselah96:
I'll take a look at it tomorrow night if I don't get to it tonight.

Methuselah96 (changes requested):
@dobesv Can we keep the common interface in `Option.d.ts`? The one that is in `Option.d.ts` is the one that is more correct because it mirrors the Flow types in [`Option.js`](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/components/Option.js#L23) which is the ultimate source of truth. The type in [`types.js`](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/types.js#L111) is outdated and should be removed from the original Flow types anyway. That way when the Flow types update, the location of the types remains mirrored and it's easy to update these type definitions without having to determine which type a file was moved into.

dobesv:
Hi, I don't have time to work on this any more - I don't need this change.  Maybe it'll be resurrected if someone else runs into issues.

PR #37451, opened 2019-08-08
[chai] Provide type definitions for chai.use

### LINKED ISSUE #29922: @types/chai No types for Chai's internal api (parameters of chai.use have type any)
- [x] I tried using the `@types/chai` package and had problems.
  - Authors: @jedmao @Bartvds @AGBrown @olivr70 @mwistrand @JoshuaKGoldberg @shaunluttin @Naktibalda @micksatana @ErikSchierboom

`Chai.ChaiStatic.use` is typed as `use(fn: (chai: any, utils: any) => void): ChaiStatic`.

The parameters `chai` and `utils` do have types, but they are part of the "private" API of chai, see for example `chai.Assertion` in [chaijs/chai](https://github.com/chaijs/chai/blob/master/lib/chai/assertion.js#L59). I'm trying to extend Chai, which means I would like to have the type of (amongst others) `chai.Assertion`.

Is there a reason why the internal API is not typed? This would be useful for Chai plugin developers.

9999years:
Provide types for the following APIs used in plugin development, as requested in #29922:
- `chai.util`
- `chai.use()`
- `chai.Assertion`

Closes #29922.

## Version number

This PR includes breaking changes which will cause code that was incorrect (but
type-checked without errors) to no longer compile by aligning the type
definitions with Chai's actual public interface.

The only members with a (potentially) breaking change, however, is `chai.use`, which takes a `Chai.ChaiPlugin` rather than a `(chai: any, utils: any) => void` function.

This puts us in the uncomfortable position of needing to release a breaking change *while* not getting out of sync with Chai's major version. I'm not entirely sure what the best way to resolve this is, but I'm open to suggestions. One thing we could do is look for packages that depend on `@types/chai` and see if this change would break them.

## TypeScript version bumps

**Note:** Version bumps to TypeScript 3.0 for better tuple type matching; a lot of Chai's plugin APIs accept an array which is used as arguments to `chai.Assertion.prototype.assert`; that function has several optional parameters, so the last few elements of the tuple are optional. Therefore, the tuple is variable-length, which wasn't supported until TypeScript 3.0.

As a consequence, this PR bumps the minimum TypeScript version for a **lot** of packages. Sorry about that.

---

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.chaijs.com/api/plugins/
- [ ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header. **N/A**
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. **N/A:** Already exists.

### DIFF
--- a/types/chai/chai-tests.ts
+++ b/types/chai/chai-tests.ts
@@ -999,17 +999,91 @@ function _throw() {
 }
 
 function use() {
-    chai.use((_chai) => {
-        _chai.can.use.any();
-    });
+                   // Modified from:
+                   //   https://www.npmjs.com/package/chai-subset
+                   // But with added type annotations and some modifications to make the
+                   // linter whine less, and also a method (chai.assert.containSubset) removed
+                   // because we can't take advantage of namespace-declaration-merging here.
+                   function chaiSubset(chai: Chai.ChaiStatic, utils: Chai.ChaiUtils) {
+                       const Assertion = chai.Assertion;
+                       const assertionPrototype = Assertion.prototype;
 
-    const expect = chai
-        .use((_chai, util) => {
-        })
-        .use((_chai, util) => {
-        })
-        .expect;
-}
+                       Assertion.addMethod('containSubset', function(expected: Object) {
+                           const actual = utils.flag(this, 'object');
+                           const showDiff = chai.config.showDiff;
+
+                           assertionPrototype.assert.call(
+                               this,
+                               compare(expected, actual),
+                               'expected #{act} to contain subset #{exp}',
+                               'expected #{act} to not contain subset #{exp}',
+                               expected,
+                               actual,
+                               showDiff
+                           );
+                       });
+
+                       Assertion.addMethod('containSubset', function(expected: Object) {
+                           const actual = utils.flag(this, 'object');
+                           const showDiff = chai.config.showDiff;
+
+                           assertionPrototype.assert.call(
+                               this,
+                               compare(expected, actual),
+                               'expected #{act} to contain subset #{exp}',
+                               'expected #{act} to not contain subset #{exp}',
+                               expected,
+                               actual,
+                               showDiff
+                           );
+                       });
+
+                       function compare(expected: Object, actual: Object): boolean {
+                           if (expected === actual) {
+                               return true;
+                           }
+                           if (typeof actual !== typeof expected) {
+                               return false;
+                           }
+                           if (typeof expected !== 'object' || expected === null) {
+                               return expected === actual;
+                           }
+                           if (!!expected && !actual) {
+                               return false;
+                           }
+
+                           if (Array.isArray(expected) && Array.isArray(actual)) {
+                               if (typeof actual.length !== 'number') {
+                                   return false;
+                               }
+                               const aa = Array.prototype.slice.call(actual);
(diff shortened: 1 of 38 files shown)


types/chai/chai-tests.ts
```diff
@@ -999,17 +999,91 @@ function _throw() {
...
-    });
-
-    const expect = chai
-        .use((_chai, util) => {
-        })
-        .use((_chai, util) => {
-        })
-        .expect;
-}
+                   // Modified from:
```
jedmao: Why is this indented so much?

jednano (approved):
Looks good, so long as the build passes.

types/chai/chai-tests.ts
9999years: Unsure. Vim weirdness. Fixed, though.

jednano (approved):

facetious:
What happened here?

`node_modules/@types/chai/index.d.ts:121:9 - error TS8020: JSDoc types can only be used inside documentation comments.
121         any?, // actual value
            ~~~~
node_modules/@types/chai/index.d.ts:122:9 - error TS8020: JSDoc types can only be used inside documentation comments.
122         boolean? // showDiff
            ~~~~~~~~
node_modules/@types/chai/index.d.ts:126:16 - error TS2370: A rest parameter must be of an array type.
126         assert(...args: AssertionArgs): void;`

This is with the new @types/chai@4.2.0.

jimlindeman:
This change to use chai.chaiPlugin broke the prior backward compatibility with @types/chai@4.1.7. 
If try to go to @types/chai@4.2.0, it's not compatible with the typescript 2.9.x we're using and fails.  
@facetious see the output from "npm info @types/chai", that's why it failed for you.
```
dist-tags:
latest: 4.2.0  ts2.1: 4.1.7   ts2.3: 4.1.7   ts2.5: 4.1.7   ts2.7: 4.1.7   ts2.9: 4.1.7   ts3.1: 4.2.0   ts3.3: 4.2.0   ts3.5: 4.2.0   
ts2.0: 4.1.7   ts2.2: 4.1.7   ts2.4: 4.1.7   ts2.6: 4.1.7   ts2.8: 4.1.7   ts3.0: 4.2.0   ts3.2: 4.2.0   ts3.4: 4.2.0   ts3.6: 4.2.0   
```
Guess we'll just hardcode to @type/chai-as-packaged@7.1.1 for now.  Was there a reason this was a fix version bump and not at least a minor or major version bump?

facetious:
Thanks, @jimlindeman. This was a typescript version issue, like you said. Guess I'll update to 3+.

leonard-thieu:
> Was there a reason this was a fix version bump and not at least a minor or major version bump?

Version numbers for `@types` packages are aligned with the major and minor versions of the library package they represent.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library

9999years:
Yeah, I'm not the biggest fan of that policy (as I noted in the original PR description). That said, I'd love to help fix any packages this might've broken, because this situation is less than ideal to say the least. There's not a ton I can do about breaking compatibility with TypeScript@<3.0, though.

PR #37453, opened 2019-08-08
[pathval] Add types for pathval

9999years:

If adding a new definition:

### DIFF
--- a/types/pathval/index.d.ts
+++ b/types/pathval/index.d.ts
@@ -4,26 +4,18 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.2
 
-declare namespace PathVal {
-    interface PathInfo {
-        parent: object;
-        name: string;
-        value?: any;
-        exists: boolean;
-    }
-
-    type Property = string | symbol | number;
-
-    interface PathValStatic {
-        hasProperty(obj: object | undefined | null, name: Property): boolean;
-        getPathInfo(obj: object, path: string): PathInfo;
-        getPathValue(obj: object, path: string): object | undefined;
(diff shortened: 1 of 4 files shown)


types/pathval/tslint.json
```diff
@@ -0,0 +1,15 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
```
sandersn: New packages should not override any lint rules. They should only extend `dtslint/dt.json`.

types/pathval/index.d.ts
```diff
@@ -0,0 +1,29 @@
...
+        hasProperty(obj: object | undefined | null, name: Property): boolean;
+        getPathInfo(obj: object, path: string): PathInfo;
+        getPathValue(obj: object, path: string): object | undefined;
+        setPathValue(obj: object, path: string, val: any): object;
+    }
+}
+
+declare const pathval: PathVal.PathValStatic;
+
+declare module "pathval" {
```
sandersn: 1. Please use the modern UMD syntax to create the global object name: `export as namespace pathval`. 
2. This means you can delete `declare const pathval`.
3. Also delete `declare namespace PathVal` and dedent its contents. Prefix the types with `export`. 
4. Also delete `interface PathValStatic`. Dedent its contents and prefix them with `export function`.

sandersn (changes requested):
Needs a few structural changes, but looks good otherwise.

9999years:
OK, made the requested changes. Should `export as namespace pathval` be at the start of the file rather than the end?

PTAL 🙂

sandersn:
I don't have a preference. I usually put it at the end.

sandersn (approved):

PR #37455, opened 2019-08-08
[mongoose-delete] New definition

ndunks:
Please fill in this template.

If adding a new definition:


types/mongoose-delete/index.d.ts
```diff
@@ -0,0 +1,125 @@
+// Type definitions for mongoose-delete 0.5
+// Project: https://github.com/dsanel/mongoose-delete
+// Definitions by: Mochamad Arifin <https://github.com/ndunks>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.2
+
+import mongoose = require('mongoose');
+
+declare global {
```
sandersn: why is this in the global namespace? The instructions for mongoose-delete always use commonjs require. They don't mention browser usage.

types/mongoose-delete/tslint.json
```diff
@@ -0,0 +1,7 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "only-arrow-functions": false,
```
sandersn: New packages should not override any lint rules. They should only extend `dtslint/dt.json`.

sandersn (changes requested):

types/mongoose-delete/index.d.ts
ndunks: Because I don't know how to export interfaces and a function (when the module just use `module.exports = function()`. But I need to exports defined interfaces too. Can you send me example or guide to solve this? I'm new in typing definition.

sandersn: You got it almost right. =) `namespace mongoose_delete` will merge with `function mongoose_delete` and `export = mongoose_delete` will export both. You just need to delete the enclosing `declare global`.

ndunks: I’ll try and test it.. thanks for your attention..

sandersn (approved):

PR #37471, opened 2019-08-08
Add errors property to submissionError

waynebrantley:
@iamandrewluca small missing property, has been around for a long time.

### DIFF
--- a/types/redux-form/lib/SubmissionError.d.ts
+++ b/types/redux-form/lib/SubmissionError.d.ts
@@ -1,8 +1,8 @@
 import { FormErrors } from "../index";
 
 export declare class SubmissionError<FormData = {}, ErrorType = string> extends Error {
-  constructor(errors?: FormErrors<FormData, ErrorType>);
-  public errors: FormErrors<FormData, ErrorType> | undefined;
+  constructor(errors: FormErrors<FormData, ErrorType>);
+  errors: FormErrors<FormData, ErrorType>;
 }
 
 export default SubmissionError;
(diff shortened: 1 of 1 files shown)


types/redux-form/lib/SubmissionError.d.ts
```diff
@@ -2,6 +2,7 @@ import { FormErrors } from "../index";
 
 export declare class SubmissionError<FormData = {}, ErrorType = string> extends Error {
   constructor(errors?: FormErrors<FormData, ErrorType>);
+  public errors: FormErrors<FormData, ErrorType> | undefined;
```
bwlt: [Travis reported](https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/569426943#L369): `member-access  'public' is implicit.` Please fix.
Also in my opinion it should not be nullable: from [source code](https://github.com/erikras/redux-form/blob/master/src/SubmissionError.js) the `errors` member is always set up.

bwlt (changes requested):

types/redux-form/lib/SubmissionError.d.ts
iamandrewluca: In this case the constructor maybe also needs fixing.

waynebrantley: I made the changes requested.  The 'public' was required when I did it locally be linter/typescript - however, I removed it as requested.

unrevised6419 (approved):

bwlt (approved):

PranavSenthilnathan:
Please create PRs using the template in the future to help make the review process easier.

PR #37405, opened 2019-08-06
[react-transition-group] Unify to standard module syntax

ybiquitous:
Use `import` and `export` syntax instead of `namespace`. `Transition.d.ts` has already used the standard syntax. This aims to be a refactoring.

See also https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37395#pullrequestreview-271338324

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/reactjs/react-transition-group>
- [ ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #37474, opened 2019-08-08
Change `react-transition-group` exports so the `Props` definitions ar…

cdeutsch:
…e still exported.

Ref:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37405#issuecomment-519579770

Please fill in this template.

### DIFF
--- a/types/react-transition-group/index.d.ts
+++ b/types/react-transition-group/index.d.ts
@@ -6,7 +6,34 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
-export { default as Transition } from './Transition';
-export { default as CSSTransition } from './CSSTransition';
-export { default as TransitionGroup } from './TransitionGroup';
-export { default as SwitchTransition } from './SwitchTransition';
+export {
+    default as Transition,
+    EndHandler,
+    EnterHandler,
+    ExitHandler,
+    TransitionActions,
+    TransitionChildren,
+    TransitionProps,
+    TransitionStatus,
+    UNMOUNTED,
+    EXITED,
+    ENTERING,
+    ENTERED,
+    EXITING,
+} from './Transition';
+export {
+    default as CSSTransition,
+    CSSTransitionClassNames,
+    CSSTransitionProps,
+} from './CSSTransition';
+export {
+    default as TransitionGroup,
+    IntrinsicTransitionGroupProps,
+    ComponentTransitionGroupProps,
+    TransitionGroupProps,
+} from './TransitionGroup';
+export {
+    default as SwitchTransition,
+    modes,
+    SwitchTransitionProps,
+} from './SwitchTransition';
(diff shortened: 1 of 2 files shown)


ybiquitous:
The reason to use `export default` instead of just `export` is that `react-transition-group` original code uses the former.

https://github.com/reactjs/react-transition-group/blob/b9b43f9e37615f7d4aa05f0770a89c13b57bae4f/src/Transition.js#L553

```jsx
export default Transition
```

Does not this PR cause a compile error as follows?

```tsx
import Transition from "react-transition-group/Transition"
//=> OK

import { Transition } from "react-transition-group/Transition"
//=> NG
```

cdeutsch:
> Does not this PR cause a compile error as follows?
> 
> ```tsx
> import Transition from "react-transition-group/Transition"
> //=> OK
> 
> import { Transition } from "react-transition-group/Transition"
> //=> NG
> ```

I suppose it would, but I don't know why you'd do the import like that, nor do you test for that.

I've never taken into consideration internal, nested file exports when doing a type definition, but I'll change it to work for that scenario.

ybiquitous:
Please suppose the following case, which is completely valid. TypeScript can compile JavaScript file(s) with the `allowJs: true` option.

```json5
// tsconfig.json
{
  "compilerOptions": {
    "allowJs": true
    // other options...
  }
}
```

```jsx
// app.jsx
import Transition from "react-transition-group/Transition";

const App = (
  <Transition in timeout={0}>
    <div />
  </Transition>
);
```

If some users would write the above code, this PR change would cause the following error.

```
node_modules/@types/react-transition-group/Transition.d.ts:187:1 - error TS1128: Declaration or statement expected.

187 export Transition;
    ~~~~~~
```

types/react-transition-group/index.d.ts
```diff
@@ -6,7 +6,34 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
-export { default as Transition } from './Transition';
-export { default as CSSTransition } from './CSSTransition';
-export { default as TransitionGroup } from './TransitionGroup';
-export { default as SwitchTransition } from './SwitchTransition';
+export {
```
Nibblesh: could you not do this?

```
export { default as Transition } from './Transition';
export * from './Transition';
```

cdeutsch: No

types/react-transition-group/index.d.ts
```diff
@@ -6,7 +6,34 @@
...
+} from './CSSTransition';
+export {
+    default as TransitionGroup,
+    IntrinsicTransitionGroupProps,
+    ComponentTransitionGroupProps,
+    TransitionGroupProps,
+} from './TransitionGroup';
+export {
+    default as SwitchTransition,
+    modes,
```
ybiquitous: Does `export modes` work properly? Because `export modes` is not included in [the original `index.js`](https://github.com/reactjs/react-transition-group/blob/b9b43f9e37615f7d4aa05f0770a89c13b57bae4f/src/index.js#L1).

```js
import { modes } from 'react-transition-group/SwitchTransition';
//=> OK

import { modes } from 'react-transition-group';
=> ???
```

Please see also [the original `SwitchTransition.js`](https://github.com/reactjs/react-transition-group/blob/b9b43f9e37615f7d4aa05f0770a89c13b57bae4f/src/SwitchTransition.js#L23
).

cdeutsch:
I can't spend any more time on this.

I'll just keep my package at `2.9.2` where the actual types are still exported

PR #37481, opened 2019-08-08
[@types/react] Make defaultValue of createContext<T>(...) optional

davidkpiano:
The `defaultValue` of `createContext<T>(...)` should be optional. According to [the docs](https://reactjs.org/docs/context.html#reactcreatecontext):

Please fill in this template.

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -334,7 +334,7 @@ declare namespace React {
         displayName?: string;
     }
     function createContext<T>(
-        defaultValue: T,
+        defaultValue?: T,
         calculateChangedBits?: (prev: T, next: T) => number
     ): Context<T>;
 


ferdaber:
The first time we created the types for `createContext` we deliberately made the default value non-optional so that consumers will be able to consume the context without needing to null check every time.

Here is some more context (hehe) on the change and Dan's thoughts on it as well: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/24509#issuecomment-382213106

ferdaber:
Also here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34854

types/react/index.d.ts
```diff
@@ -334,7 +334,7 @@ declare namespace React {
         displayName?: string;
     }
     function createContext<T>(
-        defaultValue: T,
+        defaultValue?: T,
```
eps1lon: Wouldn't this mean I can create it without a value and every consumer gets `T` while it actually is `T | undefined`?

I think overloading it with `function createContext<T>(): Context<T | undefined>` would be the sound solution. But I guess the complaint is then that you have to check for it every time when consuming?

I'd suggest using `createContext<T>(undefined as any)` which at least makes it obvious that you're using the type-system in an unsound way. Moving this "unsoundness"? into the library definition  is a recipe for confusion.

eps1lon:
If we reject this again we should add a comment with links to past work. Because this will come up again.

davidkpiano:
I disagree with the decision (especially since it seems like many others have ran into this issue) especially for the very common use-case that `createContext(...)` will always be used with a `<Provider value={...}>` (whose `value` is a required prop anyway), but I'll respect your decisions.

Providing an unfortunate work-around here for anyone who also has this issue and is unable to provide a default value (which is my use-case):

```tsx
const SomeContext = createContext<SomeType>({} as any);
```

types/react/index.d.ts
eps1lon: @davidkpiano I would appreciate if you could clear up what part of my statement was confusing to you. We had this PR a second time now and I would like to know how I could explain the current approach better.

ferdaber:
I would say that the opt in to type unsound was is not that bad, as you
could do `createContext<TypeYouWant>(undefined!)`. Type unsoundness should
always be opt in and not opt out, in my opinion.

types/react/index.d.ts
davidkpiano: @eps1lon I'll answer below.

davidkpiano:
I disagree with the arguments for making `defaultValue` required, but will respect the decision to keep it.

The main reason I disagree is that most use-cases for context assume that it will be used with a `<Provider value={...}>`, where the `value` is a required prop anyway.

See for yourself - `createContext()` is mostly called with either no arguments or an empty object `createContext({})` or `createContext(null)`: https://github.com/search?l=JavaScript&o=desc&q=react+createcontext&s=indexed&type=Code

Even in TypeScript, either throwaway "empty-content" objects that match the type, or coercions, or taking advantage of `strictNullChecks` being `false` are used to work around the issue:

```tsx
export const StoreContext = createContext<Store>(null as any);

const ToastContext = React.createContext<ToastContextValue>(null);

export const ComponentContext = createContext<HelloContext>(null);

export const MapDispatchContext = createContext((() => 0) as React.Dispatch<Store.Action>);

const SinouiFormStateContext = React.createContext<SinouiFormState>(
  // eslint-disable-next-line @typescript-eslint/no-explicit-any
  null as any,
);

export const RuleModeContext = React.createContext<RuleMode>(null)

export const FormContext = React.createContext<IFormContext>(null);

const Context = createContext<ContextType>(null)

const Context = createContext<ContextType>(undefined)

export const globalContext = createContext<GlobalContext>({} as GlobalContext);

export const StatusBarContext = createContext(({} as any) as ContextProps);
```

And many more examples. Regardless of what the React team says, it does not reflect reality, as it does not reflect the most common use-cases of `createContext()`.

PR #37486, opened 2019-08-09
[@types/react-native] AccessibilityInfo needs forward port to RN0.60 API

### LINKED ISSUE #1125: Accessibility code needs forward-port to RN60 APIs
There is a related issue in #893 about some deprecated APIs, but more importantly, an API in use now is completely gone from RN59 to RN60 - RN.AccessibilityInfo.fetch()

React-Native added support for more accessibility features, and they broke the API out into multiple feature-specific APIs.

Here's the documentation update https://github.com/facebook/react-native-website/pull/835

The API in use is here: https://github.com/microsoft/reactxp/blob/master/src/native-common/Accessibility.ts#L39

The corresponding API (for screen-reader only, to make ReactXP work with RN60) would be to use [RN.AccessibilityInfo.isScreenReaderEnabled()](https://facebook.github.io/react-native/docs/accessibilityinfo#isscreenreaderenabled) and to register the listener for 'screenReaderChanged' event vs 'changed'

I will propose a PR for that specific change as ReactXP will be a breaking-change anyway so it's okay for it not to work on RN59, but future enhancements should expand the Accessibility to take advantage of the rest of the new features. If that is desired I could propose a second PR that offers complete coverage of the enhanced API though I may need some hand-holding in how to deal with iOS/Android-specific differences as parts of the API are available only on one platform or another

Cross-linking with #1123 for visibility there since I mentioned it prior.

mikehardy:
AccessibilityInfo was expanded between RN59 and RN60 to include lots
of accessibility feature information beyond just screen reader status.

This updates the typings to match the docs: https://facebook.github.io/react-native/docs/accessibilityinfo

I have trouble running both `npm test` and `npm run lint react-native` for some reason, it complains of naming issues, but...I have not created a new package I am just editing an existing one? It definitely works for me in testing though, and is required to fix https://github.com/microsoft/reactxp/issues/1125

[Updated after determining the old upstream API exists but was merely deprecated]: The old APIs are deprecated but still exist, so this PR preserves them and just adds the new APIs that should be used going forward

Please fill in this template.

- [!] Add or edit tests to reflect the change. (Run with `npm test`.)
- [!] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://facebook.github.io/react-native/docs/accessibilityinfo

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -6910,7 +6910,15 @@ export interface ShareStatic {
     dismissedAction: "dismissedAction";
 }
 
-type AccessibilityEventName = "change" | "announcementFinished";
+type AccessibilityEventName =
+    "change" | // deprecated, maps to screenReaderChanged
+    "boldTextChanged" | // iOS-only Event
+    "grayscaleChanged" | // iOS-only Event
+    "invertColorsChanged" | // iOS-only Event
+    "reduceMotionChanged" |
+    "screenReaderChanged" |
+    "reduceTransparencyChanged" | // iOS-only Event
+    "announcementFinished"; // iOS-only Event
 
 type AccessibilityChangeEvent = boolean;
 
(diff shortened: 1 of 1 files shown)


mikehardy:
I looked into the DefinitelyTyped.DefinitelyTyped check failure, and it seems unrelated to my change?

mikehardy:
A reviewer in the related PR https://github.com/microsoft/reactxp/pull/1126 just pointed out that the old API still existed, it was just deprecated, so I'm going to push a change here shortly that recognizes that fact by keeping the old fetch() API and "change" event name but marks them as deprecated

types/react-native/index.d.ts
```diff
@@ -6916,22 +6924,61 @@ type AccessibilityEvent = AccessibilityChangeEvent | AccessibilityAnnoucementFin
...
     /**
      * Query whether a screen reader is currently enabled.
-     * Returns a promise which resolves to a boolean. The result is true when a screen reader is enabled and false otherwise.
      */
-    fetch: () => Promise<boolean>;
+    isScreenReaderEnabled: () => Promise<boolean>;
+
+    /**
+     * Query whether a screen reader is currently enabled.
+     *
```
alloy: The [Travis CI failure](https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/570049616#L560) indicates that there’s trailing whitespace here.

```suggestion
     *
```

alloy (changes requested):
Looks good to me 👍

However…

> I looked into the DefinitelyTyped.DefinitelyTyped check failure, and it seems unrelated to my change?

…the Azure one seems unrelated to your changes indeed, but the Travis one is a legit linter error.

> The types already indicate they are for react-native 0.60, but this API change is required to actually support react-native 0.60, they were missed in the update I suppose?

Updates are made opportunistically. When a change is made that requires a new version, the constraint is bumped, but that doesn’t mean anybody goes through all the types.

types/react-native/index.d.ts
mikehardy: 🤦‍♂  sorry!

mikehardy:
Dang - the trailing whitespace was in my second push, and I had not had time to check it out. I committed the suggestion so Travis should go green - hopefully Azure does as well

When I noticed that the old APIs (which are no longer documented) were still present that made sense that the typings hadn't been moved yet - probably just no one needed them yet and the old stuff wasn't actually broken, only looked that way via the docs.

Anyway - the typings for the AccessibilityInfo object will now completely match the javascript in the react-native repo assuming CI passes :-).

Thanks for the review

mikehardy:
Hi @alloy! Thanks again for the review. Not sure how to kick the typescript-bot's status, but I believe the requested change is made, and the CI is all green now. I think this is good to go? Cheers

mikehardy:
@RyanCavanaugh - just a heads-up that the TypescriptBot does not appear to re-examine status if a reviewer actually makes a "suggested change", requests changes, and then the submitter accepts the change (thus resolving the issue, pending review again). This may be a known issue? Unknown :-)

mikehardy:
I'm not sure what else I can do here to move this forward? I'm afraid the auto-status setting has killed it, maybe I should propose it fresh?

Skn0tt:
If there’s something I can do (e.g. creating a new PR), please let me know.

PR #37489, opened 2019-08-09
export react-custom-scrollbars as namespace because it is a UMD module

ryprice:
Without doing this, tsc complains that node_modules/@types/react-custom-scrollbars/index.d.ts is not a module
This change also allows importing Scrollbars as the module default `import Scrollbars from` or as a named export `import {Scollbars} from` because either work.
Documentation that this is a UMD module: https://github.com/malte-wessel/react-custom-scrollbars/commit/f632d08e725f30ac40cdb6257752ad36b40c6fef

### DIFF
--- a/types/react-custom-scrollbars/v3/index.d.ts
+++ b/types/react-custom-scrollbars/v3/index.d.ts
@@ -6,53 +6,55 @@
 
 import * as React from "react";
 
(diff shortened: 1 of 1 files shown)


kittimiyo (approved):
looks good to me

PR #37492, opened 2019-08-09
[readable-stream] asyncIterator doesn't represent every possible case

EduardoRFS:

`AsyncIterator` was too strict, chunk can be anything, not only string

### DIFF
--- a/types/readable-stream/index.d.ts
+++ b/types/readable-stream/index.d.ts
@@ -91,7 +91,7 @@ declare namespace _Readable {
         on(ev: string | symbol, fn: (...args: any[]) => void): this;
 
         _undestroy(): void;
-        [Symbol.asyncIterator](): AsyncIterableIterator<string | Buffer>;
+        [Symbol.asyncIterator](): AsyncIterableIterator<any>;
         // end-Readable
 
         constructor(options?: DuplexOptions);


types/readable-stream/index.d.ts
```diff
@@ -91,18 +91,14 @@ declare namespace _Readable {
         on(ev: string | symbol, fn: (...args: any[]) => void): this;
 
         _undestroy(): void;
-        [Symbol.asyncIterator](): AsyncIterableIterator<string | Buffer>;
+        [Symbol.asyncIterator](): AsyncIterableIterator<any>;
```
TeamworkGuy2: This looks good.

types/readable-stream/index.d.ts
```diff
@@ -91,18 +91,14 @@ declare namespace _Readable {
...
         constructor(options?: DuplexOptions);
     }
 
     // ==== _stream_passthrough ====
-    class PassThrough extends Transform implements stream.PassThrough {
-        constructor(options?: TransformOptions);
-
-        _transform<T>(chunk: T, encoding: string | null | undefined, callback: (error: any, data: T) => void): void;
-    }
+    class PassThrough extends stream.PassThrough {}
```
TeamworkGuy2: I believe the purpose of the PassThrough override is to allow users to get stronger typing in their callbacks via the generic `T` used by both the `chunk` and `callback(..., 'data')` parameter.

EduardoRFS: The reason of this PR is because this override breaks any change in `@types/node` stream module.

I'm trying to pass another PR, but because of this override it breaks, who will have a really stronger typing. Then we could extend it from here.

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37354#issuecomment-519326105

EduardoRFS: The reason of this PR is because this override breaks any change in `@types/node` stream module.

I'm trying to pass another PR, but because of this override it breaks, who will have a really stronger typing. Then we could extend it from here.

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37354#issuecomment-519326105

EduardoRFS: Forgot it, in fact It was a bug on node type definition side, I did fixed that on the other PR, sorry for wasting your time.

TeamworkGuy2 (approved):

TeamworkGuy2:
LGTM 👍

PR #37497, opened 2019-08-09
react-window: update to 1.8.5

ddrozdov:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/bvaughn/react-window/blob/master/CHANGELOG.md#182


martynaskadisa (approved):

types/react-window/index.d.ts
```diff
@@ -264,6 +264,10 @@ export interface GridProps extends CommonProps {
      * Called when the grid scroll positions changes, as a result of user scrolling or scroll-to method calls.
      */
     onScroll?: (props: GridOnScrollProps) => any;
+    /**
+     * @deprecated since version 1.8.2, please use the overscanColumnCount
```
jgoz: Nitpick: could remove the word “the”

jgoz (approved):

types/react-window/index.d.ts
ddrozdov: Yes, sorry, that was a copy-paste from the code... https://github.com/bvaughn/react-window/pull/229/files#diff-3845a69a9768c7a65c701ae9243d1bc9R276

PR #37501, opened 2019-08-09
add deletedCount for DeleteWriteOpResultObject interface

nacimgoura:
Hello,
In the package mongoose (with mongodb), a parameter is missing for `DeleteWriteOpResultObject` response.
https://mongoosejs.com/docs/api.html#query_Query-deleteMany

### DIFF
--- a/types/mongoose/index.d.ts
+++ b/types/mongoose/index.d.ts
@@ -3255,9 +3255,9 @@ declare module "mongoose" {
       callback?: (err: any, res: T) => void): Promise<T>;
 
     /** Removes documents from the collection. */
-    remove(conditions: any, callback?: (err: any) => void): Query<mongodb.DeleteWriteOpResultObject['result']> & QueryHelpers;
-    deleteOne(conditions: any, callback?: (err: any) => void): Query<mongodb.DeleteWriteOpResultObject['result']> & QueryHelpers;
-    deleteMany(conditions: any, callback?: (err: any) => void): Query<mongodb.DeleteWriteOpResultObject['result']> & QueryHelpers;
+    remove(conditions: any, callback?: (err: any) => void): Query<mongodb.DeleteWriteOpResultObject['result'] & { deletedCount?: number }> & QueryHelpers;
+    deleteOne(conditions: any, callback?: (err: any) => void): Query<mongodb.DeleteWriteOpResultObject['result'] & { deletedCount?: number }> & QueryHelpers;
+    deleteMany(conditions: any, callback?: (err: any) => void): Query<mongodb.DeleteWriteOpResultObject['result'] & { deletedCount?: number }> & QueryHelpers;
 
     /**
      * Same as update(), except MongoDB replace the existing document with the given document (no atomic operators like $set).


CaselIT:
The current type is correct according to the documentation http://mongodb.github.io/node-mongodb-native/3.2/api/Collection.html#~deleteWriteOpResult

is the documentation wrong there?

nacimgoura:
> The current type is correct according to the documentation http://mongodb.github.io/node-mongodb-native/3.2/api/Collection.html#~deleteWriteOpResult
> 
> is the documentation wrong there?

I realized the wrong typing when I tested, the method returned the field `DeletedCount`. Typescript generated an error when trying to use it.
I use mongoose so show this documentation : http://mongodb.github.io/node-mongodb-native/3.2/api/Collection.html#~deleteWriteOpResult
We use @types/mongo for mongoose

CaselIT:
maybe the typing in mongoose is wrong then. for mongodb the current definition seems to match the documentation

nacimgoura:
mongoose uses mongodb typing for the `DeleteMany` method.

CaselIT:
I guess that either the documentation of mongoose is not correct or it returns a different object, so the typing is wrong

I've just tried it. This is the value returned by mongodb. The current typing is correct
```js
coll.deleteMany({}).then(data => console.log(data.result)); 
{ n: 0, ok: 1 }
coll.deleteMany({}).then(console.log)
{
  result: { n: 0, ok: 1 },
  connection: '....'
  message: '...'
  deletedCount: 0
}
```

types/mongodb/index.d.ts
```diff
@@ -1464,11 +1464,11 @@ export interface DeleteWriteOpResultObject {
         ok?: number;
         //The total count of documents deleted.
         n?: number;
+        //The number of documents deleted.
+        deletedCount?: number;
```
PranavSenthilnathan: I think the old version was correct. According to the documentation `deletedCount` should be one level higher:
http://mongodb.github.io/node-mongodb-native/3.1/api/lib_collection.js.html#line842

PranavSenthilnathan (changes requested):

types/mongodb/index.d.ts
CaselIT: indeed. i've posted the log from mongodb here https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37501#issuecomment-521194331

PranavSenthilnathan:
@nacimgoura [Here](https://github.com/Automattic/mongoose/blob/5393bb190c9ba2c1dfdfc62f54aebc02b30e2910/lib/queryhelpers.js#L317) is where `mongoose` is transforming the result.

You'll need to change the types for that package from `mongodb.DeleteWriteOpResultObject['result']` to `mongodb.DeleteWriteOpResultObject['result'] & { deletedCount?: number }`

nacimgoura:
Okay, thanks for your help 😃

PranavSenthilnathan (approved):

PranavSenthilnathan:
I'm ok with this, can one of the owners also take a look?

@horiuchi @lukasz-zak @Alorel @jendrikw @ethanresnick @vologab @jussikinnula @ondratra @alfirin @idandrd @various89 @Fazendaaa @NormanPerrin @danmana @stablio @emmanuelgautier @frontendmonster @mingchen @penumbra1 @orblazer @HughKu @niuware @Vladmel1234 @jloveridge @grimmer0125 @richarddd @ToucheSir @borfig

jendrikw (approved):

PR #37502, opened 2019-08-09
Express: improve type of `Request['params']` aka `req.params`

OliverJAsh:
**Migration guide**: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37502#issuecomment-522155668

This will be a breaking change for many people, I'm sure, but it's the correct change to make I think? We could batch this with other breaking changes, if people feel that's necessary.

Users will now have to narrow the union. See an example of this in the tests.

---

Please fill in this template.

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: https://expressjs.com/en/api.html#req.params

### DIFF
--- a/types/express-http-proxy/express-http-proxy-tests.ts
+++ b/types/express-http-proxy/express-http-proxy-tests.ts
@@ -95,7 +95,7 @@ proxy("www.google.com", {
 
 const proxyOptions: proxy.ProxyOptions = {};
 
-app.use("/proxy/:port", proxy(req => "localhost:" + req.params.port));
+app.use("/proxy/:port", proxy(req => "localhost:" + (Array.isArray(req.params) ? {} : req.params)['port']));
 
 proxy("www.google.com", {
     filter: (req, res) => {
--- a/types/express-http-proxy/express-http-proxy-tests.ts
+++ b/types/express-http-proxy/express-http-proxy-tests.ts
@@ -95,7 +95,7 @@ proxy("www.google.com", {
 
 const proxyOptions: proxy.ProxyOptions = {};
 
-app.use("/proxy/:port", proxy(req => "localhost:" + (Array.isArray(req.params) ? {} : req.params)['port']));
+app.use("/proxy/:port", proxy(req => "localhost:" + (Array.isArray(req.params) ? {} : req.params).port));
 
 proxy("www.google.com", {
     filter: (req, res) => {
--- a/types/express-redis-cache/express-redis-cache-tests.ts
+++ b/types/express-redis-cache/express-redis-cache-tests.ts
@@ -28,7 +28,7 @@ app.get('/user/:userid',
     // middleware to define cache name
     (req, res, next) => {
         // set cache name
-        res.express_redis_cache_name = 'user-' + req.params.userid;
+        res.express_redis_cache_name = 'user-' + (Array.isArray(req.params) ? {} : req.params).userid;
         next();
     },
     // cache middleware
--- a/types/express-serve-static-core/index.d.ts
+++ b/types/express-serve-static-core/index.d.ts
@@ -181,9 +181,9 @@ export interface RequestRanges extends RangeParserRanges { }
 
 export type Errback = (err: Error) => void;
 
-type Dictionary<T> = { [key: string]: T };
+export interface Dictionary<T> { [key: string]: T; }
 export type ParamsDictionary = Dictionary<string>;
-export type ParamsArray = Array<string>;
+export type ParamsArray = string[];
 export type Params = ParamsDictionary | ParamsArray;
 
 export interface Request extends http.IncomingMessage, Express.Request {
--- a/types/express-serve-static-core/index.d.ts
+++ b/types/express-serve-static-core/index.d.ts
@@ -181,6 +181,11 @@ export interface RequestRanges extends RangeParserRanges { }
 
 export type Errback = (err: Error) => void;
 
+type Dictionary<T> = { [key: string]: T };
+export type ParamsDictionary = Dictionary<string>;
+export type ParamsArray = Array<string>;
+export type Params = ParamsDictionary | ParamsArray;
+
 export interface Request extends http.IncomingMessage, Express.Request {
     /**
      * Return request header.
@@ -433,7 +438,7 @@ export interface Request extends http.IncomingMessage, Express.Request {
 
     method: string;
 
-    params: any;
+    params: Params;
 
     /** Clear cookie `name`. */
     clearCookie(name: string, options?: any): Response;
--- a/types/express-ws/express-ws-tests.ts
+++ b/types/express-ws/express-ws-tests.ts
@@ -60,7 +60,7 @@ router.ws(
     '/:id',
     (ws, req, next) => { next(); },
     (ws, req, next) => {
-        ws.send(req.params.id);
+        ws.send((Array.isArray(req.params) ? {} : req.params).id);
 
         ws.on('close', (code, reason) => {
             console.log('code:', code);
--- a/types/express/express-tests.ts
+++ b/types/express/express-tests.ts
@@ -111,7 +111,8 @@ namespace express_tests {
         });
 
     router.get('/user/:id', (req, res, next) => {
-        if (Number(req.params.id) === 0) next('route');
+        const paramsDictionary: ParamsDictionary = Array.isArray(req.params) ? {} : req.params;
+        if (Number(paramsDictionary.id) === 0) next('route');
         else next();
     }, (req, res, next) => {
         res.render('regular');
@@ -158,7 +159,7 @@ namespace express_tests {
  *                         *
  ***************************/
 import * as http from 'http';
-import { RequestRanges } from 'express-serve-static-core';
+import { RequestRanges, ParamsDictionary } from 'express-serve-static-core';
 
 namespace node_tests {
     {
--- a/types/express/express-tests.ts
+++ b/types/express/express-tests.ts
@@ -1,4 +1,5 @@
 import express = require('express');
+import { RequestRanges, ParamsDictionary } from 'express-serve-static-core';
 
 namespace express_tests {
     const app = express();
@@ -159,7 +160,6 @@ namespace express_tests {
  *                         *
  ***************************/
 import * as http from 'http';
-import { RequestRanges, ParamsDictionary } from 'express-serve-static-core';
 
 namespace node_tests {
     {
--- a/types/i18n/i18n-tests.ts
+++ b/types/i18n/i18n-tests.ts
@@ -205,7 +205,7 @@ app.get('/ar', (_req: Express.Request, res: Express.Response) => {
     i18n.setLocale(res, 'ar');
     i18n.setLocale(res.locals, 'ar');
 
-    i18n.setLocale([req, res.locals], req.params.lang);
+    i18n.setLocale([req, res.locals], (Array.isArray(req.params) ? {} : req.params).lang);
     i18n.setLocale(res, 'ar', true);
 });
 


niekvb:
I was missing the following link to the documentation to verify these changes: https://expressjs.com/en/5x/api.html#req.params

These changes look fine to me, altough I can't test them at this moment.

OliverJAsh:
@niekvb I included that link in the PR description, in the URL field.

PranavSenthilnathan:
Could someone take a look to make sure this change is good to go: @borisyankov @ulrichb @Danscho @carboneater @niekvb @jsinglet @lindsaykwardell @ajliv @19majkel94 @kacepe @micksatana @samijaber @aereal @SomaticIT @FindQ

niekvb:
> Could someone take a look to make sure this change is good to go

These changes should be good to go, but as the impact might be huge for existing projects, it should be reviewed by other contributors as well before merging.

Danscho (approved):

niekvb (approved):

nediamond:
this pr breaks `req.params.X`

tx-steven:
Can confirm that this broke our build. Found this PR looking for a solution. Does this mean we now have to coerce params every time we want to call or de-structure a property off of it?

SupernaviX:
This also broke our build. This new change is going to be a nuisance, we have to throw around a lot of `(req.params as { [key: string]: string })` to fix type errors that provide no value for us.

Is there a way to make it less onerous? Maybe add overloads for `IRouterMatcher` that enforce the type of `req.params` if the first argument is a regex, and leave it as `any` otherwise?

zmays:
+1 broke our builds as well

OliverJAsh:
Any time a type is switched from `any` to its correct type, breaking changes are inevitable.

The new type errors you're seeing are useful—they are alerting you to a potential type mismatch, which could result in runtime exceptions.

If you were previously assuming `req.params` was always a dictionary, you can use the following type guard

``` ts
(req, res) => {
  const { foo } = req.params;
}
```

``` ts
const getParam = (params: Params , key: string): string | undefined =>
  Array.isArray(req.params) === false ? req.params[key] : undefined;

getParam('foo')

**Comment by OliverJAsh** [comment]
Any time a type is switched from `any` to its correct type, breaking changes are inevitable. Such is the case here.

The new type errors you're seeing are useful—they are alerting you to a potential type mismatch, which could result in runtime exceptions.

`req.params` can be a dictionary object or an array. It's not safe to pick a property or read an index from `req.params` without first narrowing the type of `req.params` to either a dictionary object or array.

#### Migration guide

If you were previously assuming `req.params` was always a dictionary object, e.g. you were using it like this:

``` ts
import { RequestHandler } from 'express';

const requestHandler: RequestHandler = (req, res) => {
    const foo = req.params.foo; // any
}
```

Then you can update your code to use the following type guard:

``` ts
import { RequestHandler } from 'express';
import { Params } from 'express-serve-static-core';

const getParamByKey = (params: Params, key: string): string | undefined =>
    Array.isArray(params) ? undefined : params[key];

const requestHandler: RequestHandler = (req, res) => {
    const foo = getParamByKey(req.params, 'foo'); // string | undefined
};
```

If you were previously assuming `req.params` was always an array, e.g. you were using it like this:

``` ts
import { RequestHandler } from 'express';

const requestHandler: RequestHandler = (req, res) => {
    const foo = req.params[0]; // any
}
```

Then you can update your code to use the following type guard:

``` ts
import { RequestHandler } from 'express';
import { Params } from 'express-serve-static-core';

const getParamByIndex = (params: Params, index: number): string | undefined =>
    Array.isArray(params) ? params[index] : undefined;

const requestHandler: RequestHandler = (req, res) => {
    const foo = getParamByIndex(req.params, 0); // string | undefined
};
```

It's more code, but unlike what went before, this is correct and safe.

Borvik:
This broke our builds as well.

While I agree with the general principal - this is going to be very problematic for a large portion of developers especially since `@types/express` is using a dependency of `"@types/express-serve-static-core": "*"` forcing a switch to the latest.

I second the approach SupernaviX suggested if possible.

tzellman:
Broken (random/unrelated) builds here as well! Unfortunately, I think the real problem is that `express` is trusting that the latest `@types/express-serve-static-core` typings will be sound and backwards-compatible for users. This will break ANY codebase that uses `req.params.someKey` in their code...

SupernaviX:
The `req.params` property can be a dictionary object or an array, but in practice it can be statically detected which it actually is. If your routes are a string (which is what every tutorial I've seen uses), `req.params` is definitely a dictionary object, and accessing any property is safe. It's only an array if the route handler uses a regex (which I didn't even know was supported until I started debugging all my broken builds).

Even ignoring all the impending breakages, I think this change is going to be a big obstacle for newcomers to typescript. Express tutorials for vanilla Javascript are going to say to use `req.params.id`, and newcomers will copy that line into their repos and have to debug some weird compilation error about arrays.

If you're really concerned about correctness over conciseness/user-friendliness, the ParamsDictionary would be more accurately typed as `{ [key: string]: string | undefined }` anyway. Users would have to null-check or add assertions to their code everywhere, but it would also be more correct and safe.

nediamond:
It's right about fork o'clock

colorado-nate:
For future reference, as a breaking change, this should have been a major version bump, not a patch version bump. Sadly, even that wouldn't help as @types/express says it needs `"@types/express-serve-static-core": "*"`, so it still would have picked up this version.  2 bugs make for a very long day for a lot of developers.

jmoniatte:
+1 to what @colorado-nate said.

PranavSenthilnathan:
@colorado-nate @jmoniatte The major.minor version for the @types package will be the same version as the package the types are for (this is for express 4.16 so the @type major.minor also matches it). So we only rev up patch versions for types updates until the types are updated for a new major.minor version of the package.

jwalton:
This is a huge breaking change for everyone with very little benefit.  There is a real type safety problem here, but I don't think this PR even addresses it.  The problem is:

```ts
app.get('/foo/:id',
    function(req: Request, res: Response, next: NextFunction) {
        // req.params will *always* be a dictionary here.
    }
);

app.get(/foo/(.*)/,
    function(req: Request, res: Response, next: NextFunction) {
        // req.params will *always* be an array here.
    }
);
```

The real problem here is you shouldn't be allowed to pass a handler than is expecting params to be a dictionary to a `.get()` that uses a regex, and you shouldn't be allowed to pass a handler expecting the array to a `.get()` that uses a string for the URL.  If you really wanted to fix this from a type safety perspective, what you'd want to do is redefine `IRouteMatcher` to do something along the lines of:

```ts
export interface Request<P> {
    ...
    params: P;
}

export interface RequestHandler<P> {
    (req: Request<P>, res: Response, next: NextFunction): any;
}

export interface IRouterMatcher<T> {
    (path: string | Array<string>, ...handlers: RequestHandler<Dictionary<string>>[]): T;
    (path:  string | Array<string>, ...handlers: RequestHandlerParams<Dictionary<string>>[]): T;
    (path: RegExp | Array<RegExp>, ...handlers: RequestHandler<string[]>[]): T;
    (path: RegExp | Array<RegExp>, ...handlers: RequestHandlerParams<string[]>[]): T;
    (path: Array<string | RegExp>, ...handlers: RequestHandler<Dictionary<string> | string[]>[]): T;
    (path: PathParams, subApplication: Application): T;
}
```

Then in most cases you wouldn't need to do any type narrowing in the route handler, because TypeScript would already know what type `params` was supposed to be based on what you passed in to `get()`.  Then instead of adding a bunch of senseless type narrowing code to our projects, we'd just have to change `Request` to `Request<Dictionary<string>>`.

Even that is a terrible idea, though.  If you were going to do this, the sensible thing to do would be to make this new `Request<P>` be a `RequestParams<P>`, and leave the existing `Request` alone.  Then existing code wouldn't break, and users could use the new types in new code and/or migrate their code over at their leisure.  You could possible make the definition `Request<P=any>` so it defaults to "any", that would work too, and then effectively the type safety would be optional for the people who want it.

Either way people could migrate their code gradually, instead of breaking everyone's build on a Friday night and getting us to work over the weekend.  👎

jwalton:
You could even do `interface Request<P=Dictionary<string>>`, which would still be a breaking change, but it would only break in cases where people use regexes for routes, so like 90% of code wouldn't have to change at all.

OliverJAsh:
@SupernaviX

> The `req.params` property can be a dictionary object or an array, but in practice it can be statically detected which it actually is. If your routes are a string (which is what every tutorial I've seen uses), `req.params` is definitely a dictionary object,

@jwalton 

> you shouldn't be allowed to pass a handler expecting params-as-an-array to a .get() that uses a string for the URL. 

I'm afraid this isn't the case. See the docs:

> When you use a regular expression for the route definition, capture groups are provided in the array using req.params[n], where n is the nth capture group. This rule is applied to unnamed wild card matches with **string routes** such as /file/*:

https://expressjs.com/en/api.html#req.params

Example of two string routes with different type params:

``` ts
app.get('/1/:foo', (req, res) => {
    req.params // dictionary
});

app.get('/2/*', (req, res) => {
    req.params // array
});
```

AFAIK there's no way to statically enforce the correct type of params depending on the value of the path string. Happy for someone to show me I'm wrong though!

@jwalton 

> Either way people could migrate their code gradually, instead of breaking everyone's build on a Friday night and getting us to work over the weekend. 👎

Why do you have to upgrade immediately? 😕

dragomirtitian:
No way to check for wild cards in the `string` in the type system unfortunately. Perhaps force the user to opt into either string paths or wild card string paths ?

```ts
app.get('/1/:foo' as StringPath, (req, res) => {
    req.params // dictionary
});

app.get('/2/*' as WildCardPath, (req, res) => {
    req.params // array
});
```
Forcing people to use the casts would still be a breaking change but at least it would:
1. Not force a useless runtime check on every usage of req.params
2. Make clear the reason for the difference between the two variants. 

Keeping the union behavior for `string` paths would probably be best.

tzellman:
@OliverJAsh

Unfortunately we *have* to address this since any new builds we have will fetch the latest release of this due to the wildcard dependency from express.

Our CI builds started failing from unrelated changes, and I had to track this down on a Friday afternoon. Not complaining necessarily but was not planning on having an issue like this.

I would vote revert until a better solution is drafted.

OliverJAsh:
Don't lock files help?

jwalton:
@OliverJAsh

> I'm afraid this isn't the case. See the docs

Well, that sucks.  :P  This is literally what I came up with after only 5 seconds of thinking about the problem, though.  There has to be a better way to fix this.  I really don't like adding a useless runtime check to every single route handler in the world.

Here's another problem with your solution.  Let's have a look at your migration guide:

```
    const foo = req.params.foo as string | undefined;

    // vs

    const foo = getParamByKey(req.params, 'foo'); // string | undefined
```

(This isn't *quite* the example you provided, but you'll see where I'm going with this...)

So, in the second case, where it's "type safe", there are three possible cases:

1) `req.params` is a dictionary with the key `foo`.  In this case `foo` will be a string.
2) `req.params` is a dictionary without the key 'foo'.  In this case `foo` will be undefined.
3) `req.params` is an array of strings, in which case we'll do a runtime check and see that it's an array, and `foo` will end up undefined.

In the first case:

1) `req.params` is a dictionary with the key `foo`.  In this case `foo` will be a string.
2) `req.params` is a dictionary without the key 'foo'.  In this case `foo` will be undefined.
3) `req.params` is an array of strings, in which `foo` will end up undefined.

Notice that, despite the fact that your version is "type safe", and my version "isn't", we both ended up with exactly the same result?  So what exactly are we protecting against here?

Now, that's slightly unfair; if instead of "foo" we had "length", then your version would win out.  If someone does `Object.prototype.foo = 7` then both our solutions will break.  :P  You also could make it so `getParamByKey()` throws an error if the type is wrong, but you can also do that without your PR.

Here's an alternative fix for you, though:

```
type Params = {
    [key: string]: string;
    [index: number]: string;
} & {[K in keyof string[]]: string[][K] | string};
```

This works pretty much exactly the way yours does, but it's also won't break the vast majority of use cases.  It'll be a breaking change for you if you're doing `req.params.length`, because this could be a string or a number, but that very much limits the scope of this breaking change.  It'd also be a breaking change if you were trying to do `x: number = req.params.x`, but that probably should break.

You could also do:

```
type Params = {
    [key: string]: string | undefined;
    [index: number]: string | undefined;
} & {[K in keyof string[]]: string[][K] | string | undefined};
```

...  Meh.  This is perhaps more technically correct.  The problem is, if your route is "/foo/:bar", then you *know* that `req.params.bar` is going to be present, even if typescript doesn't know that.  So again, it's kind of a senseless guard in most cases.  You might be able to talk me into making that change if this were a new project, or one that wasn't used much, but it's a hugely breaking change with very little upside so it's probably not worth it.

One other thing you could do (in addition to the above):

```
export interface Request<P=Params> {
    ...
    params: P;
}
```

The idea being when you declare a route handler you could do:

```
function handler(req: Request<{userName: string}>, ...) {
    console.log(req.params.username);  // This will be an error.
}
```

Again, you can opt in, and you get much better type safety than a generic `Params` can provide.  You could even do something like:

```
interface MyRouteParams {
    foo: string
};

function handler(req: Request<MyRouteParams>) {
    console.log(req.params.foo);
}

app.get<MyRouteParams>('/route/:foo',
  handler,
  otherHandler,
  yetAnotherHandler
);
```

which I can see being useful in many cases.

jwalton:
@dragomirtitian

This is clever, but unfortunately it won't work.  Typescript is a "duck typed" language;

```
type StringPath = string;
type WildCardPath = string;

const x: StringPath = 'foo' as StringPath;
cosnt y: WildCardPath = x; // No error here, because StringPath and WildCardPath have the same interface.
```

dragomirtitian:
@jwalton I was thinking of a branded type approach, sorry I didn't make that explicit, sometimes I forget to write all that goes on in my head 😅
```ts
type StringPath = string & { __stringPath?: "stringPathBrand" };
type WildCardPath = string& { __wildCardPath?: "wildCardPathBrand" };

declare function get(s: StringPath, cb: (req: { params: Record<string, string>}) => void): void;
declare function get(s: WildCardPath, cb: (req: { params: string[]}) => void): void;
declare function get(s: string, cb: (req: { params: string[] | Record<string, string>}) => void): void;

get('/1/:foo' as StringPath, (req) => {
    req.params // dictionary
});

get('/2/*' as WildCardPath, (req) => {
    req.params // array
});


get("", (req) => { // No assertions, do your own checks
    req.params // array | dictionary
});

```
[Play](https://www.typescriptlang.org/play/#code/C4TwDgpgBAysBOBLAdgcwAoENgAsoF4oBnBFVKAMigG8oB9OkpNLXAfgC4oAiJs1nACF4mZABNuUAL4BuAFChIUAOqIANmIDCmeGIEFipNFVoMA7uq0692HJx4WN23QOGiJ0+XLEQAxmp1oADMAV2RfYEQAe2QoVAhgAAoiLjhmDFsAGihfACMuRPgIAEcuWjAdTABbFKgAJT8o3QAePjRsttQAPikASgIuqAA3KMQxXq4Rsfkff0CoUPDImLiE5K5VJ2sBbLyCotKaKAqRGq5OgG0AXT6B4dHxyYeZvwCihbCI6Nj4pNrO3b5KCFEplY6VM6GdLXKAAH3qjRaAKhZB6-Xwgymj3u0zkcl+iQA5AB6ACMxI4QSiUUJUEwRFgRgyuGyIOK6MG1DkUB5UAOADoTtUGcTiVAxIgvjEdCA5H0vASSQAmYkAKlp9JUlmcNhZwIOHJo3N5AqFNSgorp8BEsvlePxa243FZBrutEtADkonSiEQIPBlsgiNkxN6QFEQvAoFEzLFfDg-ABrIjGnmmiEisU6G1w8WSwMyuW9GRAA)

Similar solution to what ts uses to make sure strings are not used as paths without checks. See [here](https://github.com/microsoft/TypeScript/blob/b57b5fe5ee31a334a1e2b63eced4ba5a046bff61/src/compiler/types.ts#L4)

jwalton:
@dragomirtitian   Oooh.  That is very clever.  Thanks for teaching me a new trick.  :)

jwalton:
@OliverJAsh 

> Don't lock files help?

1) A lot of people think you shouldn't use a lockfile for a "library" project (as opposed to an "application" project).  You're locking dependencies when you're developing in the library, but these are probably not the same dependencies the consumers of your library will end up with, so what's the point?

2) If you use Dependabot, they recommend not using a lockfile for a library project, too.  If you have a lockfile, then Dependabot will open a PR every time any dependency in your entire lockfile changes.  Any reasonably sized project will have a huge graph of dev dependencies, so this leads to a ton of PRs that aren't important.  This is one case where I think Greenkeeper wins out hugely over Dependabot, because it doesn't open a PR unless there's something for you to actually do.

All this to say, lots of valid reasons why people don't use lock files.

jwalton:
Yet another way to solve this would be to fix this in Express.  Right now the API is ambiguous; you pass in a string or a regex, and you get back a `req.params` which is sometimes an array and sometimes an object.  This doesn't lend itself nicely to TypeScript, and maybe there isn't really an awesome solution to this on the type def side (which is why it's `any` right now).

So, we could go write a PR for express; add a `req.getParam(value: string | number): string | undefined`, and then there's a typesafe option to use in place of `req.params`.  Or we could put in a PR over there which makes this wholly unambiguous (although you'd have to come up with something clever to make it worthwhile, I suspect).

OliverJAsh:
@jwalton 

> Notice that, despite the fact that your version is "type safe", and my version "isn't", we both ended up with exactly the same result? So what exactly are we protecting against here?

Here's the behaviour we had before this PR:

``` ts
type Params = any;
declare const params: Params;

params.length; // any
const r1 = params[0]; // any
const r2 = params['foo']; // any
const r3 = params.foo; // any
```

Here's the behaviour which I desired and that led me to create this PR:

``` ts
params.length; // number | string
const r1 = params[0]; // string
const r2 = params['foo']; // string
const r3 = params.foo; // string
```

(Note these index lookups should really return `string | undefined` for complete type safety, but I see that as separate issue for TS to deal with itself: https://github.com/microsoft/TypeScript/issues/13778. For now we just have to accept that this is how TS works.)

We're trying to avoid leaking `any`s in our code, because they completely remove type safety. E.g.

``` ts
const foo = req.params.foo; // any

foo.i.do.not.exist; // runtime exception!
```

How we achieve the above desired behaviour (i.e. avoid params from returning `any`s), I really don't mind.

With this PR, we've achieved the following:

``` ts
type Params = { [key: string]: string } | string[];
declare const params: Params;

const getParam = (
    params: Params,
    indexOrKey: string | number,
): string | undefined =>
    Array.isArray(params)
        ? typeof indexOrKey === 'number'
            ? params[indexOrKey]
            : undefined
        : params[indexOrKey];

params.length; // string | number
const r1 = getParam(params, 0); // string | undefined
const r2 = getParam(params, 'foo'); // string | undefined
```

Not quite as nice as the example with vanilla index lookups, but at least we're not leaking `any`s.

If we try to use vanilla index lookups, we get errors. This doesn't imply that we've defined the wrong type for `params`—this is just how TS works when you try to index into a union of type `Dictionary | Array`.

``` ts
params.length; // string | number
const r1 = params[0]; // error, should be string
const r2 = params['foo']; // error, should be string
const r3 = params.foo; // error, should be string
```

If we use an intersection instead of a union (as suggested by @jwalton [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37502#issuecomment-522233891)), we lose the correct type for `params.length`:

``` ts
type Params = { [key: string]: string } & string[];
declare const params: Params;
params.length; // number, should be string | number
```

jwalton:
> If we use an intersection instead of a union (as suggested by @jwalton here), we lose the correct type for params.length:

```
type Params = {
    [key: string]: string;
    [index: number]: string;
} & {[K in keyof string[]]: string[][K] | string | undefined};
```

fixes some of your concerns.  Here:

```
params.forEach(() => {}) // probably a runtime exception
```

will tell you that `params.forEach` could be undefined.  But, this is getting super hacky, because you run into weird corner cases like:

```ts
    const foo = req.params.foo; // This is a string
    const index = req.params.index; // This is a string | number | undefined
    if (Array.isArray(req.params)) {
        const index = req.params.index; // But this works, at least
    }
```

What about:

```
interface Request<P extends Dictionary<string> | string[] = Dictionary<string>> {
    ...
    params: P
}
```

Then, at least we're not breaking anywhere near as much?  My goal is to make it so this works ouot of the box:

```ts
app.get('/user/:id', function(req: Request, res: Response) {
    db.User.findOne({_id: req.params.id})
        .then(user => res.json(user));
}
```

Or something very close to that.  I agree that it would be nice if params were strings instead of anys, but I'd like to do it in a way that breaks as little as possible in existing code.

OliverJAsh:
It took me awhile to understand what you were doing there, but I think I understand now. I expanded on your example to try and explain what's going on. Note I also discovered some more edge cases with this solution:

``` ts
type ParamsArray = string[];

// Like a normal array, but each array property/method includes the `string`
// type, so that we get the correct return types when indexing into the
// params union.
type ParamsArrayInUnion = {
    [K in keyof ParamsArray]: ParamsArray[K] | string
};

type ParamsDictionary = {
    [key: string]: string;
};

type Params = ParamsDictionary & ParamsArrayInUnion;

declare const params: Params;
params.length; // number | string
const r1 = params[0]; // string
const r2 = params['foo']; // string
const r3 = params.foo; // string

if (Array.isArray(params)) {
    params.length; // number
    const r4 = params[0]; // any, expected string
    const r5 = params['foo']; // string, expected error
    const r6 = params.foo; // string, expected error
} else {
    params.length; // string | number, expected string
    const r4 = params[0]; // string
    const r5 = params['foo']; // string
    const r6 = params.foo; // string
}
```

With regards to the idea of using generics—if we were to default to a dictionary type, I guess we would still be breaking array usages. Those users would have to require an explicit array generic to `Request` somehow. But I guess that's a minority of users, so that would be better than breaking all usages for all users. Is that what you're thinking?

jwalton:
Yes, exactly; instead of breaking it for 100% of use cases, break it for 1%.

jwalton:
Also, the generic version doesn't need a runtime check, which is nice.

OliverJAsh:
@jwalton I'm giving this idea a go here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37718

How does that look? Would appreciate your thoughts and any help you can provide.

The biggest problem I've faced so far is having to bump Express and all its dependants from TS 2.2 to 2.3, in order to use default generics. Hopefully that won't wreak more havoc…

OliverJAsh:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37718 has now been merged/released.

antonpod:
How does one arrive at a decision to break all Express + Typescript projects like that?

transparentCF:
@antonpod Everything has been sorted out now, you should update your deps

mikalai-ramashka:
Hey guys, please revert these breaking changes. There are many use cases with `params`, not just parsing URL,  you have broken tons of express+typescript applications.

kacepe:
@mikalai-ramashka Can you give any example?

mikalai-ramashka:
@kacepe 
1. Type cast:
```
router.param('id', (req, res, next, id) => {
	id = Number(id);
	if (!Number.isInteger(id)) {
		return next(new ServerError(400));
	}
	req.params.id = id;
...
```
2. Pass model to next routes:
```
router.param('id', async (req, res, next, id) => {
	req.params.model= await service.getById(id);
...
router.put(':id'...
router.delete(':id'
 ...
```

It's cool that we can define type of `params`, but it's not required for many cases. Much better default type of params could be { [key: string]: any }

silesky:
thank you guys for having the courage to make this change! I understand that there is blowback, but you've made many typescript projects a lot safer (one `any` at a time!).

krzysdz:
> [`req.params` can be a dictionary object _or_ an array](https://expressjs.com/en/api.html#req.params). It's not safe to pick a property or read an index from `req.params` without first narrowing the type of `req.params` to either a dictionary object or array.

I know that the documentation says that, but it's not true. Since Express 4.0.0 `req.params` is always an object (see https://github.com/expressjs/expressjs.com/issues/2100 or test it yourself).

Sorry for resurrecting this old thread. I have reported it in https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/74006, but I think some of you may want to know it.

PR #37513, opened 2019-08-09
Add type definitions for dialog-polyfill

rubenswieringa:
Types definitions for https://github.com/GoogleChrome/dialog-polyfill

----

_(Adding a new definition:)_


types/dialog-polyfill/index.d.ts
```diff
@@ -0,0 +1,9 @@
+// Type definitions for dialog-polyfill 0.5
+// Project: https://github.com/GoogleChrome/dialog-polyfill
+// Definitions by: Ruben Swieringa <https://github.com/rubenswieringa>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare namespace dialogPolyfill {
+    function registerDialog(element: HTMLElement): void;
+}
+export default dialogPolyfill;
```
sandersn: the .js code that dialog-polyfill [actually ships](https://unpkg.com/dialog-polyfill@0.5.0/dist/dialog-polyfill.js) doesn't have an `export default` in it. It boils down to

```js
module.exports = {
  registerDialog: ...
}
```

So you should delete `export default` and `declare namespace`. Then `export function registerDialog` and in the tests `import * as dialogPolyfill from "dialog-polyfil"`. 

People who have `esModuleInterop: true` in their tsconfig can still use `import dialogPolyfill from "dialog-polyfill"` but for backward compatibility the tests on DT should assume `esModuleInterop: false`.

sandersn (changes requested):

types/dialog-polyfill/index.d.ts
rubenswieringa: Excellent, thanks for the suggestions, code is updated now!

sandersn (approved):

PR #37518, opened 2019-08-09
[hippie] added type definitions

ghost:
Please fill in this template.

If adding a new definition:

If removing a declaration:


types/hippie/index.d.ts
```diff
@@ -0,0 +1,62 @@
...
+
+interface EndCb {
+  (err: undefined|Err|Error, res: http.IncomingMessage, body?: object|string): void;
+}
+
+interface Opts {
+  [key: string]: string|{};
+}
+
+declare function hippie(app?: any): typeof hippie;
```
PranavSenthilnathan: I don't think this is what you want to export. Specifically, this allows users to directly access functions off the module:
```
var hippie = require('hippie');
hippie.end(...)
```
One easy way to fix this is to rename this `hippie` function to something else and export it (and still keep the namespace named `hippie`). Another option is to do something similar but use interfaces to avoid the `typeof`s:

```
interface hippie {
    del(url: string): hippie;
}

declare function hippieConstructor(): hippie;
export = hippieConstructor;
```

types/hippie/index.d.ts
```diff
@@ -0,0 +1,62 @@
...
+    function header(key: string, val: string): typeof hippie;
+    function base(url: string): typeof hippie;
+    function app(app: (req: http.ClientRequest, res: http.ServerResponse) => {}): typeof hippie;
+    function auth(user: string, pass: string): typeof hippie;
+    function opts(fn: (opts: Opts) => void): typeof hippie;
+    function form(): typeof hippie;
+    function json(): typeof hippie;
+    function qs(obj: {}): string;
+    function send(data: any): typeof hippie;
+    function expect(fn: (result: http.IncomingMessage, body: string|{}, next: (err?: Err|Error) => void) => void): typeof hippie;
```
PranavSenthilnathan: As discussed [here](http://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html), it's not recommended to have optional parameters in callbacks.

PranavSenthilnathan (changes requested):

types/hippie/index.d.ts
nicomuschert: Hi Pranav, thank you for review.

Hippie is not my baby. I would like to express how it is expected to work with it based on vesln's (the author) examples. But i think i can change the definition without having side effects in this way:

`err: null|Err|Error`

Would you recommend it with a that null-type?

Best,
Nico

PranavSenthilnathan: You should be able to just make it non-optional. The user can still pass in a callback that takes no arguments so there shouldn't be any change in usage (but has the benefits described in the link above).

```suggestion
    function expect(fn: (result: http.IncomingMessage, body: string|{}, next: (err: Err|Error) => void) => void): typeof hippie;
```

PR #37527, opened 2019-08-10
@types/utm - Make northern optional in toLatLon method

JonasTriki:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [index.es.js#L36](https://github.com/TimothyGu/utm/blob/master/index.es.js#L36) requires `northern` to be undefined to not throw error.


types/utm/index.d.ts
```diff
@@ -19,7 +20,7 @@ export function toLatLon(
     northing: number,
     zoneNum: number,
     zoneLetter: string,
-    northern: boolean,
+    northern?: boolean,
```
PranavSenthilnathan: The js code seems to indicate that exactly one of either `zoneLetter` or `northern` should be set. So this should probably be an overload:
```
export function toLatLon(<args>, zoneLetter: string, northern?: undefined, strict?: boolean);
export function toLatLon(<args>, zoneLetter: undefined <| any other falsy values>, northern: boolean, strict?: boolean);
```

PranavSenthilnathan (changes requested):

PranavSenthilnathan:
Since this is adding an author, @civilizeddev can you take a look if you get a chance?

PR #37529, opened 2019-08-10
[underscore.string] `isBlank` accepts null like objects.

ketan:
Please fill in this template.

  - https://epeli.github.io/underscore.string/#isblank-string-gt-boolean
  - https://github.com/epeli/underscore.string/blob/master/tests/isBlank.js#L11-L15

### DIFF
--- a/types/underscore.string/index.d.ts
+++ b/types/underscore.string/index.d.ts
@@ -1,6 +1,7 @@
 // Type definitions for underscore.string
 // Project: https://github.com/epeli/underscore.string
 // Definitions by: Ry Racherbaumer <https://github.com/rygine>
+//                 Ketan Padegaonkar <https://github.com/ketan>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
@@ -55,7 +56,7 @@ declare module 'underscore' {
          * Determine if a string is 'blank.'
          * @param str
          */
-        isBlank(str: string): boolean;
+        isBlank(str: string | undefined | null | boolean | number): boolean;
 
         /**
          * Removes all html tags from string.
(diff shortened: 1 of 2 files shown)


types/underscore.string/index.d.ts
```diff
@@ -55,7 +56,7 @@ declare module 'underscore' {
          * Determine if a string is 'blank.'
          * @param str
          */
-        isBlank(str: string): boolean;
+        isBlank(str: string | undefined | null | boolean | number): boolean;
```
rygine: since the `makeString` helper will convert non-strings (except objects) to string versions of these types (e.g. `undefined` => `"undefined"`, i don't think it makes sense to allow anything other than a string here. it seems like this function is specifically to check if a string is blank. i believe the type checker, in this case, replaces the need for the string type coercion that's occurring in `makeString`.

ketan: From what I can see in [`makeString`](https://github.com/epeli/underscore.string/blob/master/helper/makeString.js#L5):

```
module.exports = function makeString(object) {
  if (object == null) return '';
  return '' + object;
};
```

the condition `if (object == null)` evaluates to `true` if `object` is `undefined`, so `undefined` is actually evaluated to an empty string, so is null.

Boolean, on the other hand will be converted to a string, and will not be blank.

Screen Shot 2019-08-12 at 12 12 58 AM

rygine: agreed. however, just because it _can_ accept something other than a string, doesn't mean we should promote doing so. IMO, `makeString` is just a guard against passing in a non-string so that the underlying regex doesn't fail.

PranavSenthilnathan: The signature and usage in the documentation seems to imply only strings should be passed in here, but the tests do have cases for non-strings. I can see this going either way. You could even add an overload `isBlank(str: any): false;` after the existing signature. Let me know where you guys want to go with this PR. @rygine @ketan

PranavSenthilnathan (changes requested):

types/underscore.string/index.d.ts
ketan: I did consider adding an `any` type, however since the tests explicitly tested specific types, I decided against it. But if you feel strongly about it, I'm happy to change it to `any`.

PR #37540, opened 2019-08-11
storybook__addon-storyshots: Options are optional

denkristoffer:
`initStoryshots()` can be called without options, as seen in [the Storyshots README](https://github.com/storybookjs/storybook/tree/a9038842ff8c8ad3adba748ac7fc969d9e464722/addons/storyshots/storyshots-core#configure-storyshots-for-html-snapshots).

### DIFF
--- a/types/storybook__addon-storyshots/index.d.ts
+++ b/types/storybook__addon-storyshots/index.d.ts
@@ -52,8 +52,10 @@ export function renderWithOptions(options?: SnapshotOptions): Test;
 
 export function getSnapshotFileName(context: StoryContext): string;
 
-// tslint:disable-next-line no-unnecessary-generics
-export default function initStoryshots<Rendered>(options: InitOptions<Rendered>): void;
+export default function initStoryshots<Rendered>(
+  // tslint:disable-next-line no-unnecessary-generics
+  options?: InitOptions<Rendered>,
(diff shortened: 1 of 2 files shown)


Yama-Tomo (approved):
LGTM 👍

types/storybook__addon-storyshots/index.d.ts
```diff
@@ -52,8 +52,10 @@ export function renderWithOptions(options?: SnapshotOptions): Test;
 
 export function getSnapshotFileName(context: StoryContext): string;
 
-// tslint:disable-next-line no-unnecessary-generics
-export default function initStoryshots<Rendered>(options: InitOptions<Rendered>): void;
+export default function initStoryshots<Rendered>(
+  // tslint:disable-next-line no-unnecessary-generics
```
PranavSenthilnathan: Regarding this lint rule - is it possible to get rid of the generic type and make `options` be of type `InitOptions<any>`?

denkristoffer: Isn't it necessary to be able to pass the renderer's return type?

PranavSenthilnathan: Ah, I didn't see that test case.

PR #37554, opened 2019-08-12
feat(react-dom): Add async act

eps1lon:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: Ports #37426 to `react-dom/test-utils`
- ~[ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.~

### DIFF
--- a/types/react-dom/test-utils/index.d.ts
+++ b/types/react-dom/test-utils/index.d.ts
@@ -291,8 +291,8 @@ export function createRenderer(): ShallowRenderer;
  * @see https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#testing-hooks
  */
 // the "void | undefined" is here to forbid any sneaky "Promise" returns.
-// the actual return value is always a "DebugPromiseLike".
 export function act(callback: () => void | undefined): DebugPromiseLike;
+export function act(callback: () => Promise<void | undefined>): Promise<void>;
 
 // Intentionally doesn't extend PromiseLike<never>.
 // Ideally this should be as hard to accidentally use as possible.
--- a/types/react-dom/test-utils/index.d.ts
+++ b/types/react-dom/test-utils/index.d.ts
@@ -292,7 +292,9 @@ export function createRenderer(): ShallowRenderer;
  */
 // the "void | undefined" is here to forbid any sneaky "Promise" returns.
 export function act(callback: () => void | undefined): DebugPromiseLike;
-export function act(callback: () => Promise<void | undefined>): Promise<void>;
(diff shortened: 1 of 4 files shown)


types/react-dom/test-utils/index.d.ts
```diff
@@ -291,8 +291,10 @@ export function createRenderer(): ShallowRenderer;
  * @see https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#testing-hooks
  */
 // the "void | undefined" is here to forbid any sneaky "Promise" returns.
-// the actual return value is always a "DebugPromiseLike".
 export function act(callback: () => void | undefined): DebugPromiseLike;
+// the "void | undefined" is here to forbid any sneaky return values
+// tslint:disable-next-line: void-return
+export function act(callback: () => Promise<void | undefined>): Promise<undefined>;
```
davidje13: I don't think this needs multiple signatures; `DebugPromiseLike` is promise-like, after all:

```ts
export function act(callback: () => Promise<void | undefined> | void | undefined): DebugPromiseLike;
```

should be fine. And I'm not sure if it even needs the whole `void | undefined` thing now that it doesn't need to catch promises in a special way. Maybe getting rid of the `undefined`s and just using `void`s is fine now? I don't really understand how that bit was working before. It could also be changed to simply `() => any`, but that doesn't get the intended semantics across so well.

eps1lon: `DebugPromiseLike` is here to throw type errors because it will trigger react warnings if you await the sync act and more importantly can cause subtle scheduling bugs.

`() => any` is not valid either because react will warn if you return anything but a thenable from it.

davidje13: Ah, in that case I think the first signature should become a regular promise, because the recommendation now is to `await` even synchronous `act` (if you give it a synchronous function, it may still wait for asynchronous tasks). So maybe combine them with a `Promise<void>` return type?

eps1lon: > because the recommendation now is to await even synchronous act 

Do you have a link for that? I remember a PR where the author for async act explicitly told me this can cause bugs and I'm fairly certain this still issues warnings from react.

davidje13: Hmm, looks like I made it up, which is weird because I distinctly remember seeing that from a maintainer in the [original tracker](https://github.com/facebook/react/issues/14769). Maybe it was an early thought which got edited out. Anyway doesn't seem to be the case, so I guess it does have to be 2 separate signatures.

PR #37557, opened 2019-08-12
more vuelidate properties

mcshaz:
Please fill in this template.

Vuelidate updated validator $v type definitions.

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: https://vuelidate.netlify.com/#sub-v-values

### DIFF
--- a/types/vuelidate/index.d.ts
+++ b/types/vuelidate/index.d.ts
@@ -1,4 +1,5 @@
-// Type definitions for vuelidate 0.7
+// Type definitions for vuelidate 0.7.3
+
 // Project: https://github.com/monterail/vuelidate
 // Definitions by: Jan Esser <https://github.com/janesser>
 //                 Jubair Saidi <https://github.com/jubairsaidi>
--- a/types/vuelidate/index.d.ts
+++ b/types/vuelidate/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for vuelidate 0.7.3
+// Type definitions for vuelidate 0.7
 
 // Project: https://github.com/monterail/vuelidate
 // Definitions by: Jan Esser <https://github.com/janesser>
--- a/types/vuelidate/vuelidate.d.ts
+++ b/types/vuelidate/vuelidate.d.ts
@@ -1,4 +1,4 @@
-import { ValidationParams } from './lib/validators'
+import { ValidationParams, Params } from './lib/validators';
 import Vue, { PluginFunction } from 'vue'
 
 /**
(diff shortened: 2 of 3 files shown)


types/vuelidate/index.d.ts
```diff
@@ -1,4 +1,5 @@
-// Type definitions for vuelidate 0.7
+// Type definitions for vuelidate 0.7.3
```
janesser: It was conscious decision to go with major releases and not include the hot fix version.

Is there some API referred which has been introduced by 0.7.3?

types/vuelidate/vuelidate.d.ts
```diff
@@ -8,20 +8,30 @@ import Vue, { PluginFunction } from 'vue'
...
+interface EachValidation extends Validation {
+  $iter: Validation;
+}
 
-    // const validationMethods
-    $touch(): never
-    $reset(): never
-    $flattenParams(): ValidationParams[]
+export interface IterableValidation extends Validation {
+  $each: EachValidation;
```
janesser: Either all Validation have $iter or null or you come here be unchecked casting. What speaks against putting each on all Validation?

janesser (changes requested):
Can you check the tests being green? I am far away from computer atm.

types/vuelidate/index.d.ts
mcshaz: no - I did want to differentiate from the previous version because changing ```$params: { [attr: string]: any }``` to ```$params: { [attr: string]: Params | null }```, it may mean some existing typescript code will have to change to allow for possible null - i.e. ```$this.$v.$params.between!.min``` - my testing would indicate this is however correct - this value will be null if a non-leaf node higher in a nested validation. I am new to submitting type definitions, and while more technically correct, I am not sure how to indicate this is a 'breaking' type definition

types/vuelidate/vuelidate.d.ts
mcshaz: It is better to put it as possibly undefined on a single interface as you suggest - change added for your perusal.

types/vuelidate/index.d.ts
janesser: Indeed there might be a use-case that isn't Params nor null. I had such a typing here and one of the initial reviewers gave reason to go for any. I need some time back at my computer to trace that back.

Types versioning isn't very strong, didn't came across any so called best practice about it.

types/vuelidate/vuelidate.d.ts
```diff
@@ -1,27 +1,34 @@
...
-    readonly $pending: boolean
-    readonly $params: { [attr: string]: any }
-    $model: any
+export interface Validation {
+  readonly $anyDirty: boolean;
+  readonly $anyError: boolean;
+  readonly $dirty: boolean;
+  readonly $error: boolean;
+  readonly $invalid: boolean;
+  readonly $params: { [attr: string]: Params | null };
```
janesser: Besides this change, the PR looks safe to merge.

janesser (changes requested):

PR #37561, opened 2019-08-12
@types/marko - Simplify/improve `marko/express` definitions

darkwebdev:
These changes use the `open interface` pattern of TS defs. This implementation is much easier to use and more correct.

Please fill in this template.

### DIFF
--- a/types/marko/express.d.ts
+++ b/types/marko/express.d.ts
@@ -1,20 +1,10 @@
 import * as express from 'express';
 import Template from './src/runtime/html/Template';
-import { IRouterMatcher as ExpressRouterMatcher, NextFunction, PathParams } from 'express-serve-static-core';
 
-declare function m(): m.Application;
-
-declare namespace m {
-    interface Response extends express.Response {
+declare module 'express-serve-static-core' {
+    interface Response {
         marko(template: Template, data?: any): void;
     }
-
-    interface Application extends express.Application {
-        get: ((name: string) => any) & ExpressRouterMatcher<this> & MarkoRouterMatcher;
-    }
-
-    type MarkoRouterMatcher = (path: PathParams, ...handlers: RequestHandler[]) => Application;
-    type RequestHandler = (req: express.Request, res: Response, next?: NextFunction) => any;
 }
 
-export = m;
+export = express;
--- a/types/marko/marko-tests.ts
+++ b/types/marko/marko-tests.ts
@@ -1,9 +1,9 @@
 import { createServer } from 'http';
 import { createWriteStream } from 'fs';
-import express = require('express');
-import asyncWriter = require('async-writer');
+import * as express from 'express';
+import * as asyncWriter from 'async-writer';
 
-import markoExpress = require('marko/express');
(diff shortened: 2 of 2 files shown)


types/marko/express.d.ts
```diff
@@ -1,20 +1,10 @@
...
-    interface Application extends express.Application {
-        get: ((name: string) => any) & ExpressRouterMatcher<this> & MarkoRouterMatcher;
-    }
-
-    type MarkoRouterMatcher = (path: PathParams, ...handlers: RequestHandler[]) => Application;
-    type RequestHandler = (req: express.Request, res: Response, next?: NextFunction) => any;
 }
 
-export = m;
+export = express;
```
PranavSenthilnathan: Is this line needed? As far as I understand, this definition is a module augmentation on `express-serve-static-core` and no functions/values are exported directly.

darkwebdev: According to the (Marko.js docs)[https://markojs.com/docs/express/] we need to do this:
```
var express = require("express");
var markoExpress = require("marko/express")

var app = express();
app.use(markoExpress()); // <- this is the reason I export `express`
```

I have this example in the tests and I don't know how to make it better, do you have any ideas?

PranavSenthilnathan: Oops, I didn't read at the docs carefully enough. Your change looks correct.

redonkulus:
@darkwebdev @PranavSenthilnathan I just updated my app to version 4.6.2 (from 4.6.1) and I get the following error from tsc:

```
14:48:42   src/app.ts:27:9 - error TS2349: Cannot invoke an expression whose type lacks a call signature. Type 'typeof e' has no compatible call signatures.
14:48:42  
14:48:42   27 app.use(markoExpress());
14:48:42              ~~~~~~~~~~~~~~
14:48:42  
14:48:42     node_modules/@types/marko/express.d.ts:1:1
14:48:42       1 import * as express from 'express';
14:48:42         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
14:48:42       Type originates at this import. A namespace-style import cannot be called or constructed, and will cause a failure at runtime. Consider using a default import or import require here instead.
```

My app is just doing:

```
import markoExpress from 'marko/express';
app.use(markoExpress());
```

Info:
- typescript@3.5.3
- marko@4.18.11
- @types/marko@4.6.2

darkwebdev:
@redonkulus try `import * as markoExpress from 'marko/express';`
as far as I know that's the default for modules exported using old-school commonJS style like `marko/express` which does:

```
module.exports = function markoAppMiddleware() {
...
```

redonkulus:
Thanks @darkwebdev. Your suggestion wasn't working for me, but `import { default as markoExpress } from 'marko/express';` seems to work. 

I'm surprised this isn't breaking anyone else.

darkwebdev:
@redonkulus looks like usage depends on your `tsconfig`:
- if you have `"allowSyntheticDefaultImports": true` or `"esModuleInterop": true` then use `import { default as ...}`
- otherwise use `import * as ...`

Issue #9120, opened 2016-04-26
chai definitions break tsx


arolson101:
- [X] I tried using the latest `chai/chai.d.ts` file in this repo and had problems.
- [X] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [X] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [X] I want to talk about `chai/chai.d.ts`.
  - The authors of that type definition are cc/ @Bartvds 

The bit at the end of chai.d.ts:

``` ts
interface Object {
    should: Chai.Assertion;
}
```

breaks tsx support because it thinks that all objects must have a `should` property (specifically, " Property 'should' is missing in type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<...").  It should be:

``` ts
interface Object {
    should?: Chai.Assertion;
}
```

nikolaymatrosov:
This Object interface expansion also cause another bug. Typescript compiler adds 
```/// <reference types="chai">``` to all d.ts generated based on sources containing `Object`. (E.g. https://github.com/angular-redux/ng2-redux/pull/209/files)
I think it is better to eliminate polution Object interface with `should` property and make separate version of typings for those who need `should` assertion syntax.

**Comment by villasv** [comment]
It also breaks my compilation process because I have typings for Elasticsearch Search body requests. It has an internal interface

```
export interface Bool {
  should?: Query | Query[];
  filter?: Query | Query[];
  must?: Query | Query[];
  must_not?: Query | Query[];
}
```

And when I try to build an object with
``` 
const b: Bool = { should: query }
```

I get a type error saying that `Query` and `Assertion` are incompatible.

**Comment by villasv** [comment]
I agree with @nikolaymatrosov that we should not pollute the Object interface by default and I'm going to make a PR removing that last bit.

**Comment by villasv** [comment]
I couldn't manage a feasible PR because I tried to make it happen in the existing @types/chai package, but now I think we should instead create an alternatives types library that is exactly the same but doesn't contain the Object declaration.

**Comment by bevacqua** [comment]
This `interface Object { should: Chai.Assertion }` thing is causing TS to infer that an `{}` requires a `should` property when my type contains `should`, but then completely fails to reconcile the two. So when I update Cypress, which has a dependency on Chai, then `tsc` throws tens of errors.

Contrived example:

```
interface QueryContainer {
[...]

PR #37571, opened 2019-08-12
[@types/chai] Remove should from Object interface

villasv:
Avoids Object pollution and solves #9120

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: *change not related to the library itself*

### DIFF
--- a/types/chai/index.d.ts
+++ b/types/chai/index.d.ts
@@ -1828,7 +1828,3 @@ declare const chai: Chai.ChaiStatic;
 declare module "chai" {
     export = chai;
 }
-
-interface Object {
-    should: Chai.Assertion;
-}


villasv:
Hang on, I realize that there are many other type libs that currently depends on `should` being there so I'm going to propose a separate definition without it.

jednano (approved):
This is not a great situation. There's no way you can know if they actually depended on `should` being there or not. At the same time, these changes are definitely improving the accuracy of the types.

Considering the latter, I think I'm personally OK with merging this PR. It might break some implementations, but it will eventually create a better, more accurate set of types.

villasv:
Should I work on removing all tests that depend on .should? Or should I just make it an alternative chai typing?

jedmao:
Are you 100% sure that `should` style is not included in `chai`? I'm looking at https://www.chaijs.com/guide/styles/#should and I don't see any installation docs for installing `should`.

villasv:
`should` IS included in chai, but because it violates our type system, we don't use it. We only use the `expect` form, which is equivalent and doesn't mess with the Object Prototype.

So it's not wrong to include that Object interface in the @types/chai lib, this is not a bugfix. But this is an improvement, because TypeScript folks should avoid using the `should` syntax. Of course this decision is opinionated and will break lots of code, which is why I'm saying that it might be better to release this as a parallel typing library.

villasv:
I don't mind the PR getting close, as long as we come to a conclusion and the discussion doesn't end here. I still want to see some typed chai lib that doesn't have the Object pollution.

villasv:
Tracked by https://github.com/DefinitelyTyped/DefinitelyTyped/issues/9120 already

jwalton:
> should IS included in chai, but because it violates our type system, we don't use it.

```
$ node.js
> chai = require('chai');
> x = {};
> x.should
undefined
```

Not sure if Chai polluted this prototype before, but it doesn't seem to now.

jwalton:
```
$ node.js
> chai = require('chai');
> chai.should();
> x = {};
> typeof x.should
'object'
```

-_-

PR #37579, opened 2019-08-13
Declare namespace instead of simple export

hpbABD:
Just a proposal. YouTube video types are declared in a namespace so PHPStorm and other IDEs can use the types easier.

It has to be tested with current implementations that uses the removed "export default Player".


hpbABD:
Some tests failing so I have to check why.

PR #37581, opened 2019-08-13
Adding namespace to support IDE (PhpStorm)

hpbABD:
Do it a little bit like the Google way to support the IDEs like PhpStorm.


denisyilmaz:
Hm, we had this before I think. The official player.js only exposes a Player class without the namespace. So the typings should reflect that and not export it in an additional namespace. See https://github.com/vimeo/player.js/blob/master/src/player.js

hpbABD:
Ah okay, so its identical to the original. 

Hm, but still a little bit strange without namespace in the original. So if there is another player with the same "Player" class name then its confusing.

Thanks for feedback :)

PR #37582, opened 2019-08-13
Add types for raf-schd

adambrgmn:

Adding a new definition:

### DIFF
--- a/types/raf-schd/index.d.ts
+++ b/types/raf-schd/index.d.ts
@@ -4,7 +4,7 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.1
 
-interface Schedule<T extends (...args: any[]) => any> {
+interface Schedule<T extends (...args: any[]) => void> {
     (...args: Parameters<T>): void;
     cancel(): void;
 }
--- a/types/raf-schd/index.d.ts
+++ b/types/raf-schd/index.d.ts
@@ -1,0 +1,14 @@
+// Type definitions for raf-schd 4.0
+// Project: https://github.com/alexreardon/raf-schd (Does not have to be to GitHub, but prefer linking to a source code repository rather than to a project website.)
+// Definitions by: Adam Bergman <https://github.com/adambrgmn>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.1
+
+interface Schedule<T extends (...args: any[]) => any> {
+    (...args: Parameters<T>): void;
+    cancel(): void;
+}
+
+declare function rafSchd<T extends (...args: any[]) => void>(fn: T): Schedule<T>;
+
+export = rafSchd;
--- a/types/raf-schd/index.d.ts
+++ b/types/raf-schd/index.d.ts
@@ -1,5 +1,5 @@
 // Type definitions for raf-schd 4.0
-// Project: https://github.com/alexreardon/raf-schd (Does not have to be to GitHub, but prefer linking to a source code repository rather than to a project website.)
+// Project: https://github.com/alexreardon/raf-schd
 // Definitions by: Adam Bergman <https://github.com/adambrgmn>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.1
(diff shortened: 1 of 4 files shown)


types/raf-schd/index.d.ts
```diff
@@ -0,0 +1,14 @@
+// Type definitions for raf-schd 4.0
+// Project: https://github.com/alexreardon/raf-schd
+// Definitions by: Adam Bergman <https://github.com/adambrgmn>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.1
+
+interface Schedule<T extends (...args: any[]) => any> {
```
PranavSenthilnathan:
```suggestion
interface Schedule<T extends (...args: any[]) => void> {
```

types/raf-schd/index.d.ts
```diff
@@ -0,0 +1,14 @@
...
+    (...args: Parameters<T>): void;
+    cancel(): void;
+}
+
+declare function rafSchd<T extends (...args: any[]) => void>(fn: T): Schedule<T>;
+
+export = rafSchd;
```
PranavSenthilnathan: It seems like the [underlying library](https://www.npmjs.com/package/raf-schd) is using `export default`, so this .d.ts file should as well.
```suggestion
export default rafSchd;
```

PranavSenthilnathan (changes requested):

types/raf-schd/index.d.ts
adambrgmn: I tried this at first. But running `npm run lint raf-schd` with that code (and the tests updated) makes `dtslint` complain:
```sh
ERROR: 14:1  npm-naming  The types for raf-schd specify 'export default' but the source does not mention 'default' anywhere.
Here is the source:
## Source removed but found here https://runpkg.com/?raf-schd@4.0.2/dist/raf-schd.cjs.js
See: https://github.com/Microsoft/dtslint/blob/master/docs/npm-naming.md
```

The source code refered to is the `main`-file defined in package.json. But bundlers will most likely use the [`module`-file](https://runpkg.com/?raf-schd@4.0.2/dist/raf-schd.esm.js) from package.json which uses `export default...`

Is there a way to either force dtslint to look for the module file instead? Or in worst case skip that rule? The link mentioned in the error message does not exists, so no luck there.

adambrgmn: I took a look at #35165 which seemed to encounter the same problem. They decided to stay on the `export = rafSchd`-format but adding `"esModuleInterop": true` to tsconfig.json. What do you think?

sandersn: The code should use `export = rafSchd`; usage from node looks like this:

```
✔ test  [master L|✚ 11…7]$ npm install raf-schd
✔ test  [master L|✚ 11…7]$ node
> var r = require('raf-schd')
undefined
> r
[Function: rafSchd]
> r.default
undefined
```


This means that raf-schd's README is wrong, by the way, since it claims commonjs usage needs to append `.default`.

Both this PR and #35165 should leave off esModuleInterop since DT's tests need to simulate the least-common-denominator environment. esModuleInterop is only going to be present in recently-created (2.9?) codebases.

types/raf-schd/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+        "strictFunctionTypes": true,
+        "strictNullChecks": true,
+        "baseUrl": "../",
+        "typeRoots": [
+            "../"
+        ],
+        "types": [],
+        "noEmit": true,
+        "forceConsistentCasingInFileNames": true,
+        "esModuleInterop": true
```
sandersn: Remove this line.

types/raf-schd/raf-schd-tests.ts
```diff
@@ -0,0 +1,12 @@
+import rafSchd from 'raf-schd';
```
sandersn: Use the commonjs syntax `import rafSchd = require('raf-schd')`.

Consumers of the package who want to use `import rafSchd from 'raf-schd'` will need to add `"esModuleInterop": true`, but Definitely Typed tests need to be backward compatible to show that the types will work with old code bases.

sandersn (changes requested):

PR #37466, opened 2019-08-08
Add loading attribute to img in React - React

charliearlie:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [ ] Add or edit tests to reflect the change. (Run with `npm test`.) - were any tests testing this?
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://web.dev/native-lazy-loading & https://css-tricks.com/native-lazy-loading/
- [ ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #37590, opened 2019-08-13
[@types/react] add img loading props

id0Sch:
add typings for the new native image lazy loading feature
Reference:
	https://www.chromestatus.com/feature/5645767347798016

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.chromestatus.com/feature/5645767347798016

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -1924,6 +1924,7 @@ declare namespace React {
         crossOrigin?: "anonymous" | "use-credentials" | "";
         decoding?: "async" | "auto" | "sync";
         height?: number | string;
+        loading?: "lazy" | "eager" | "auto" | "";
         sizes?: string;
         src?: string;
         srcSet?: string;
--- a/types/react/test/tsx.tsx
+++ b/types/react/test/tsx.tsx
@@ -365,6 +365,14 @@ imgProps.decoding = 'auto';
 imgProps.decoding = 'sync';
 // $ExpectError
 imgProps.decoding = 'nonsense';
+
+imgProps.loading = '';
+imgProps.loading = 'auto';
+imgProps.loading = 'eager';
+imgProps.loading = 'lazy';
+// $ExpectError
+imgProps.loading = 'nonsense';
+
 type ImgPropsWithRef = React.ComponentPropsWithRef<'img'>;
 // $ExpectType ((instance: HTMLImageElement | null) => void) | RefObject<HTMLImageElement> | null | undefined
 type ImgPropsWithRefRef = ImgPropsWithRef['ref'];


eps1lon (approved):

ferdaber (changes requested):
@eps1lon feel free to re-approve if you disagree but I don't really think this should be merged. The `lazy` attribute isn't standard yet and it's not in the official spec. It's just a Chrome feature right now.

See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37466#issuecomment-519540060 also.

id0Sch:
> @eps1lon feel free to re-approve if you disagree but I don't really think this should be merged. The `lazy` attribute isn't standard yet and it's not in the official spec. It's just a Chrome feature right now.
> 
> See [#37466 (comment)](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37466#issuecomment-519540060) also.

oh I'm sorry, I couldn't find that issue before submitting this PR.

The audience for my product is Chrome only so we want to support lazy loading, having TS errors until this is a standard / overriding the default typings seems wrong. WDYT?
Thanks for your time!

ferdaber:
You can augment the React defs in your own codebase. If you create a `react.d.ts` (just has to end in `.d.ts`, filename can be anything) file where your `tsconfig.json` will pick it up, and add this in, it will augment the current type definitions:
```ts
// react.d.ts
import 'react'

declare module 'react' {
  interface ImgHTMLAttribute<T> extends HTMLAttributes<T> {
    loading?: "lazy" | "eager" | "auto" | "";
  }
}
```

id0Sch:
> You can augment the React defs in your own codebase. If you create a `react.d.ts` (just has to end in `.d.ts`, filename can be anything) file where your `tsconfig.json` will pick it up, and add this in, it will augment the current type definitions:
> 
> ```ts
> // react.d.ts
> import 'react'
> 
> declare module 'react' {
>   interface ImgHTMLAttribute<T> extends HTMLAttributes<T> {
>     loading?: "lazy" | "eager" | "auto" | "";
>   }
> }
> ```
@ferdaber, sounds like a good solution for now.
Thanks again :)

eps1lon (changes requested):
@ferdaber You're right module augmentation is the better solution. 

There's probably an argument to be made to allow any excess attributes since react allows the same and the platform is always moving anyway. But I guess this would allow invalid values for existing attributes?

ferdaber:
Yah if we go with the route of type safety first, it's not a great idea. TS's JSX functionality already allows kebab-case freeform attributes (`data-*` or `aria-*`), so that usually covers the vast majority of other cases.

kachkaev:
@ferdaber there is a small typo in your [code example](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37590#issuecomment-520846460), which I noticed after copy-pasting:

```diff
-   interface ImgHTMLAttribute<T> extends HTMLAttributes<T> {
+   interface ImgHTMLAttributes<T> extends HTMLAttributes<T> {
```

Thanks for sharing the the tip!

ferdaber:
Thanks, I fixed it

kachkaev:
This feature should be hopefully merged in #38390 given a recent attribute acceptance.

kachkaev:
Starting with `@types/react` 16.9.20, @ferdaber's [workaround](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37590#issuecomment-520846460) is no longer needed 🎉

PR #37591, opened 2019-08-13
Adding locales types to plotly

Domino987:
This adds the locale types to plotly. This enables [localisation](https://github.com/plotly/plotly.js/blob/master/dist/README.md#to-include-localization) as described in the docs.

Please fill in this template.


types/plotly.js/lib/locales/zh-tw.d.ts
```diff
@@ -0,0 +1,3 @@
+import { Locale } from '../core';
+
+export default Locale;
```
sheetalkamat: This needs `export =` instead of `export default`. Refer to https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default

```js
const x = require("plotly.js/lib/locales/af");
console.log(x.default);
```
```
> node a.js
undefined
```

sheetalkamat (changes requested):

types/plotly.js/lib/locales/zh-tw.d.ts
Domino987: I changed it, thank you.

types/plotly.js/lib/locales/am.d.ts
```diff
@@ -0,0 +1,3 @@
+import { Locale } from '../core';
+
+export = Locale;
```
sheetalkamat: This needs to be
```ts
const locale: Locale;
export = locale;
```
Since it exports a value of type Locale rather than the type Locale

sheetalkamat (changes requested):

types/plotly.js/lib/locales/am.d.ts
Domino987: This does not work for me. Only this:

`import { Locale } from '../core';

const locale: Locale;

export { Locale };`

types/plotly.js/lib/locales/af.d.ts
```diff
@@ -1,3 +1,6 @@
 import { Locale } from '../core';
 
-export = Locale;
+const locale: Locale;
+
+export { Locale };
```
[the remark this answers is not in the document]
Domino987: But this wont let me import it like this: 'deDictionary' only refers to a type, but is being used as a value here.
If I import it like this: import deDictionary from 'plotly.js/lib/locales/de.js';

PR #37592, opened 2019-08-13
Add types for scroll

RomiC:
Please fill in this template.

If adding a new definition:


types/scroll/tslint.json
```diff
@@ -0,0 +1,80 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
```
PranavSenthilnathan: Could you remove these rules?

types/scroll/index.d.ts
```diff
@@ -0,0 +1,61 @@
+// Type definitions for scroll 3.0.1
```
PranavSenthilnathan: This should be just `3.0` (one of the lint rules will catch this when it's reenabled).

PranavSenthilnathan (changes requested):

types/scroll/index.d.ts
RomiC: So, am I right that that in case of minor changes type definitions should remain the same?

PranavSenthilnathan: You can update the type definitions any time to keep up with patches/revisions of the underlying package as you wish. However, the idea is that patches shouldn't affect the type definitions that much (i.e. breaking changes should be major/minor version changes).

PranavSenthilnathan: I might have misread your question - minor version changes in a package means that you should also rev up the minor version in the definition. So when 3.1.2 (for example) gets released, then you can update the definition to 3.1.

RomiC: @PranavSenthilnathan 👍 Thank you for an explanation!

PR #37594, opened 2019-08-13
[draft-js] Fix DraftBlockType core definition

iRON5:
Screen Shot 2019-08-13 at 19 10 08

`string` is more abstract type than `CoreDraftBlockType`
This way final `DraftBlockType` type always will ignore `CoreDraftBlockType`
Make `DraftBlockType` as generic to allow use both custom and core types

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/iRON5/DefinitelyTyped/blob/981cc6d51eb213943a3606c14ad75d65ea6b360f/types/draft-js/index.d.ts#L330>>

### DIFF
--- a/types/draft-js/index.d.ts
+++ b/types/draft-js/index.d.ts
@@ -36,7 +36,10 @@ declare namespace Draft {
 
             import DraftBlockRenderConfig = Draft.Model.ImmutableData.DraftBlockRenderConfig;
 
-            type DraftBlockRenderMap = Immutable.Map<DraftBlockType, DraftBlockRenderConfig>;
+            type DraftBlockRenderMap<CustomBlockType = void> = Immutable.Map<
+                DraftBlockType<CustomBlockType>,
+                DraftBlockRenderConfig
+            >;
 
             type DraftStyleMap = {
                 [styleName: string]: React.CSSProperties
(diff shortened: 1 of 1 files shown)


MunifTanjim:
I think this unnecessarily complicates thing without providing any real benefits.

For example:

 `RichTextEditorUtil.getCurrentBlockType` can return any `string`. So, even if you write this:

```ts
const blockType = RichTextEditorUtil.getCurrentBlockType(editorState)
```
The type of `blockType` is not gurranteed to be `CoreDraftBlockType` (but your typings suggests so). It can be any string (and you can't change the implementation of that function). So, your typings would be wrong.

Similar things can be said for the other changes.

types/draft-js/index.d.ts
```diff
@@ -920,7 +923,7 @@ declare namespace Draft {
 
             class RichTextEditorUtil {
                 static currentBlockContainsLink(editorState: EditorState): boolean;
-                static getCurrentBlockType(editorState: EditorState): DraftBlockType;
+                static getCurrentBlockType<CustomBlockType = void>(editorState: EditorState): DraftBlockType<CustomBlockType>;
```
MunifTanjim: `RichTextEditorUtil.getCurrentBlockType` can return any string. So, even if you write this:

```ts
const blockType = RichTextEditorUtil.getCurrentBlockType(editorState)
```

The type of `blockType` isn't guaranteed to be `CoreDraftBlockType` (but your typings suggests so). It can be any string (and you can't change the implementation of that function). So, your typings is actually be wrong.

Similar things can be said for other changes too.

orta:
I'm going to close this as idle, if you want to give it another shot - you're very welcome to @iRON5

PR #37605, opened 2019-08-14
[async] add Promise return values

yume-chan:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: caolan/async#1572

----

### To reviewers:

I have several questions when updating this definition:

`AsyncIterableIterator`:

The `IterableCollection<T>` type can also be `AsyncIterableIterator<T>`, but this type is in ES2018 and I don't know if we can use it here.

`doWhilst`:

The `iteratee` parameter will be called with `(callback: (err: any, ...results: any[]) => void)`, and the `test` parameter will be called with `(...results: any[], callback: (err: any, truth: boolean) => void)`, passing the `results` got from `iteratee`.

However, without type recursion, it's impossible to correctly typing both `iteratee` and `test` parameters. So I use `Function` for `test` here.

`race`: 

The result of each task in `tasks` does not necessarily have the same type.

However, the compiler is not able to infer the result type (should be a union type) `T` from the `tasks` parameters, the user must specify it manually.

Also, the official API documentation is missing Promise support for this function.

type alias:

Some function definitions are aliases to others. For example `eachSeries` is an alias to `each`. But they just have the same signature by luck but they work differently. Shall we use aliases for different things but with the same signature?

`series`:

Functions like `series` can accept both Array and Object, but because the return value will also be Array or Object respectively, we can't use the universal `IterableCollection<T>` here.

Currently, the Array overload on these functions contains Array only. Shall we split `IterableCollection<T>` to Array and Object (Object already have `Dictionary<T>`)? What name is suited?

`waterfall`:

Functions like `waterfall`, `seq` and `compose` takes multiple functions, and pass the result from one to the next.

There is no way to correctly typing the relation between these parameter functions, and the type of the return value, right?

Also, the official API documentation is missing Promise support for this function.

`AsyncFunction`:

With `AsyncFunction` become a union type (can be either callback style or Promise style), the compiler can't infer the generic type `T` any more.
I believe we should change this, but how?

Testing:

I bought a test case from caolan/async here, added some type declarations, fixed some issues, so I can pass the typing test with TypeScript 3.0+.

But with the usage of argument spreading and other features introduced in TypeScript 3.0, this file won't pass compilation with TypeScript 2.3~2.9.

Currently I excluded this file from `tsconfig.json`. Is there a way to fix it (without upgrade type definition to use TypeScript 3.0)?

**EDIT:** The CI doesn't like unused files, so I removed it. You can review it at https://github.com/DefinitelyTyped/DefinitelyTyped/blob/21fe63ca69081b6a2d775bd7928587a727f191f9/types/async/test/awaitable.ts


yume-chan:
Related #37325

sheetalkamat:
This PR has questions in it. Popular package definitely needs to be reviewed by definition owners

rbuckton:
@yume-chan: If you want to leverage newer TypeScript features but maintain compatibility with older versions of TypeScript, we have some guidance here: https://github.com/DefinitelyTyped/DefinitelyTyped#i-want-to-use-features-from-typescript-31-or-above

uniqueiniquity (dismissed):
Marking as changes requested until Ron's comment is addressed (so that the bot can handle this appropriately).

yume-chan:
@rbuckton I have already read that, but the definition itself doesn't use TypeScript 3.0+ features, I just want to test new features against the definition file.

rbuckton:
My comment was in response to the question in the PR description, not the changes in the PR itself. I concur with @sheetalkamat that we need a definition owner to weigh in on these changes.

uniqueiniquity:
Re-tagging @borisyankov @Kern0 @Penryn @fenying @pascalmartin @Dmitri1337 @erossignon @Juliiii for review of these changes - we don't feel that it's a good idea to merge without your input.

levyitay:
Is this being maintained?

Robula:
What's the hold-up on this? Can this be merged and we can worry about using newer Typescript features later?

alvis:
Retagging @borisyankov @Kern0 @Penryn @fenying @pascalmartin @Dmitri1337 @erossignon @Juliiii.

I just have a review on the changes. As an async user, I can confirm that the changes look good to me. In fact, besides adding the missing promise types, the author also fix some wrong return types such as `unsaturated` in `AsyncQueue` which has been changed to a Promise since 3.0. (see http://caolan.github.io/async/v3/docs.html#QueueObject)

As this PR has been held up for half a year already, and async 3 has already been released for almost a year. It should be time to move on and merge.

PR #37613, opened 2019-08-14
[leaflet] Add missing TileLayerOptions properties

Wkalmar:
`TitleLayerOptions` definition was missing properties from [getting started guide](https://leafletjs.com/examples/quick-start/)

### DIFF
--- a/types/leaflet/index.d.ts
+++ b/types/leaflet/index.d.ts
@@ -486,6 +486,8 @@ export class GridLayer extends Layer {
 export function gridLayer(options?: GridLayerOptions): GridLayer;
 
 export interface TileLayerOptions extends GridLayerOptions {
+    id?: string;
+    accessToken?: string;
     minZoom?: number;
     maxZoom?: number;
     maxNativeZoom?: number;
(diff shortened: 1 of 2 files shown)


ronikar (approved):
Good catch :)

Nilesh-Neftex:
@Wkalmar As you are working on the definition of the TileLayerOptions, any chance you can modify TileLayer.WMS options? According to [docs](https://leafletjs.com/reference-1.5.0.html#tilelayer-wms-option) any other vendor specific options can specified in options, whereas currently it is restricted to WMSParams. eg

```
        this.leafMap.eachLayer((leafLyr:L.TileLayer.WMS)=>{
            const cqlFilter:string = this.getCqlFilter(lyrName)
            leafLyr.setParams({cql_filter:cqlFilter}, false)
        })
```

In this snippet typescript complains about cql_filter does not exist in type WMSParams

ronikar:
@Nilesh-Neftex, Because `cql_filter` is custom options, you should do it locally. 
Try to use `as any`: `leafLyr.setParams({cql_filter:cqlFilter} as any, false)`

PR #37614, opened 2019-08-14
[chart.js] Added scale property, used by radio chart

ElianCordoba:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: <https://www.chartjs.org/docs/latest/charts/radar.html#scale-options>

### DIFF
--- a/types/chart.js/index.d.ts
+++ b/types/chart.js/index.d.ts
@@ -17,6 +17,7 @@
 //                 Adrián Caballero <https://github.com/adripanico>
 //                 wertzui <https://github.com/wertzui>
 //                 Martin Trobäck <https://github.com/lekoaf>
+//                 Elian Cordoba <https://github.com/ElianCordoba>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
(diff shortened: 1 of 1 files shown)


ElianCordoba:
Screen Shot 2019-08-14 at 11 10 14
Screen Shot 2019-08-14 at 11 09 33
 
Result of using
```typescript
scale: {
  ticks: {
    display: false
  }
}
``` 

Chart generated with ng2-charts

types/chart.js/index.d.ts
```diff
@@ -239,6 +240,7 @@ declare namespace Chart {
         animation?: ChartAnimationOptions;
         elements?: ChartElementsOptions;
         layout?: ChartLayoutOptions;
+        scale?: ChartScales;
```
lekoaf: `scale` and `scales`, which is correct? Can the either of them be removed?

ElianCordoba: Both are used, ´scale´ is used in the radio chart while `scales` is used, for example, in the bar chart.

adripanico: It is true, both ```scale``` and ```scales``` exist, but I think they don't implements the same interface. According to the documentation you provided, there is only one property:

```
options = {
    scale: {
        // Hides the scale
        display: false
    }
};
```

ElianCordoba: You are right, now I updated it.

PR #37621, opened 2019-08-14
react-router - make generatePath accept Params type

MKrupauskas:
`generatePath` helper function now takes in a Params type for stronger typing.

Align the types API to other helper functions. The types of `generatePath` are now similar to `matchPath`.

Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/ReactTraining/react-router/blob/master/packages/react-router/docs/api/generatePath.md

If removing a declaration:

### DIFF
--- a/types/react-router/index.d.ts
+++ b/types/react-router/index.d.ts
@@ -20,6 +20,7 @@
 //                 Wesley Tsai <https://github.com/wezleytsai>
 //                 Sebastian Silbermann <https://github.com/eps1lon>
 //                 Nicholas Hehr <https://github.com/HipsterBrown>
+//                 Mykolas Krupauskas <https://github.com/mkrupauskas>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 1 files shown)


MKrupauskas:
image
As we can see here it validated the params object taking into account the params you pass it. In this example it made `section` required.

grieblius (approved):
Great feature, will let build more robust code!

types/react-router/index.d.ts
```diff
@@ -129,7 +130,7 @@ export type Omit<T, K extends keyof T> = T extends any ? Pick<T, Exclude<keyof T
 
 export function matchPath<Params extends { [K in keyof Params]?: string }>(pathname: string, props: string | RouteProps, parent?: match<Params> | null): match<Params> | null;
 
-export function generatePath(pattern: string, params?: { [paramName: string]: string | number | boolean | undefined }): string;
+export function generatePath<Params extends { [K in keyof Params]?: string }>(pattern: string, params?: Params): string;
```
eps1lon: We could use `Params extends Record<string, string | number | boolean | undefined>` which is more obvious and restores the previous value type. The current typing has this circular reference i.e. Params extends keyof Params which is always harder to reason about.

eps1lon (changes requested):
Thank you for your contribution! The implementation  is essentially good but removes some accepted value types for which no context is provided in the PR description.

PR #37634, opened 2019-08-14
Update VariantArgs interface to support inline variants added in styled-system v5.1.0

benjamingeorge:
styled-system v5.1.0 added a new property on the VariantArgs interface 

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/styled-system/index.d.ts


sheetalkamat (changes requested):
Please provide info as to why this change is correct. (eg. change that added this property in the original package)

PR #37638, opened 2019-08-15
 [@types/map-to-obj] New type definition

JasonHK:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/map-to-obj/index.d.ts
+++ b/types/map-to-obj/index.d.ts
@@ -1,0 +1,13 @@
+// Type definitions for map-to-obj 0.2
+// Project: https://github.com/egoist/map-to-obj#readme
+// Definitions by: Jason Kwok <https://github.com/JasonHK>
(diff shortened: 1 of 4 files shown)


types/map-to-obj/index.d.ts
```diff
@@ -0,0 +1,13 @@
...
+// Project: https://github.com/egoist/map-to-obj#readme
+// Definitions by: Jason Kwok <https://github.com/JasonHK>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.9
+
+/**
+ * Convert a `Map` instance into a plain `object`.
+ * @param map The target to be converted
+ */
+declare function MapToObj<K extends string | number | symbol, V>(map: Map<K, V>): Record<K, V>;
```
PranavSenthilnathan: Also possible is

```suggestion
declare function MapToObj<K extends keyof any, V>(map: Map<K, V>): Record<K, V>;
```

PR #37643, opened 2019-08-15
Add types for ccapture.js

samalexander:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/ccapture.js/index.d.ts
+++ b/types/ccapture.js/index.d.ts
@@ -1,0 +1,67 @@
+// Type definitions for ccapture.js 1.1
+// Project: https://github.com/spite/ccapture.js#readme
+// Definitions by: Sam Alexander <https://github.com/samalexander>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare class CCapture {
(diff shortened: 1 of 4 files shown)


types/ccapture.js/index.d.ts
```diff
@@ -0,0 +1,67 @@
...
+         */
+        startTime?: number;
+        /**
+         * Path to the gif worker script
+         */
+        workersPath?: string;
+    }
+}
+
+export = CCapture;
```
PranavSenthilnathan: Is this actually a module? I can't find any examples besides in a global scope. If this is only a global library, you don't need the exports. See [this template](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-d-ts.html) for a starting point and [this](https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#dependencies-on-global-libraries) is how to consume it in tests.

PranavSenthilnathan (changes requested):

types/ccapture.js/index.d.ts
samalexander: As suggested, I have removed the exports.

PR #37647, opened 2019-08-15
fix(jest-environment-puppeteer): Class & Exports

favna:
This makes the typings actually compatible to create your own test environment and not just as part of the bigger jest-puppeteer package. How to do this is pretty much what the test covers. One of the reasons to do this is to write a handler where Puppeteer takes screenshots after each failed test, as discussed [here](https://github.com/smooth-code/jest-puppeteer/issues/131)

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/smooth-code/jest-puppeteer/blob/master/packages/jest-environment-puppeteer/src/PuppeteerEnvironment.js

### DIFF
--- a/types/jest-environment-puppeteer/index.d.ts
+++ b/types/jest-environment-puppeteer/index.d.ts
@@ -93,4 +93,4 @@ declare global {
     const jestPuppeteer: JestPuppeteer;
 }
 
-export default PuppeteerEnvironment;
+export = PuppeteerEnvironment;
--- a/types/jest-environment-puppeteer/index.d.ts
+++ b/types/jest-environment-puppeteer/index.d.ts
@@ -13,7 +13,7 @@ import { ModuleMocker } from 'jest-mock';
 import { Browser, Page, BrowserContext } from 'puppeteer';
 import { Script, Context } from 'vm';
 
-export interface JestPuppeteer {
+interface JestPuppeteer {
     /**
      * Reset global.page
      *
@@ -50,13 +50,13 @@ export interface JestPuppeteer {
     debug(): Promise<void>;
 }
 
-export interface Timer {
+interface Timer {
     id: number;
     ref: () => Timer;
     unref: () => Timer;
 }
 
-export interface Global extends GlobalType.Global {
+interface Global extends GlobalType.Global {
     browser: Browser;
     context: Context;
     page: Page;
--- a/types/jest-environment-puppeteer/index.d.ts
+++ b/types/jest-environment-puppeteer/index.d.ts
@@ -2,12 +2,18 @@
 // Project: https://github.com/smooth-code/jest-puppeteer/tree/master/packages/jest-environment-puppeteer
 // Definitions by: Josh Goldberg <https://github.com/joshuakgoldberg>
 //                 Ifiok Jr. <https://github.com/ifiokjr>
+//                 Jeroen Claassens <https://github.com/favna>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.4
 
+import { JestEnvironment } from '@jest/environment';
+import { JestFakeTimers as FakeTimers } from '@jest/fake-timers';
+import { Circus, Global as GlobalType, Config } from '@jest/types';
+import { ModuleMocker } from 'jest-mock';
 import { Browser, Page, BrowserContext } from 'puppeteer';
+import { Script, Context } from 'vm';
 
-interface JestPuppeteer {
+export interface JestPuppeteer {
     /**
      * Reset global.page
      *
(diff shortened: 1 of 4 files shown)


types/jest-environment-puppeteer/index.d.ts
```diff
@@ -2,12 +2,18 @@
 // Project: https://github.com/smooth-code/jest-puppeteer/tree/master/packages/jest-environment-puppeteer
 // Definitions by: Josh Goldberg <https://github.com/joshuakgoldberg>
 //                 Ifiok Jr. <https://github.com/ifiokjr>
+//                 Jeroen Claassens <https://github.com/favna>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.4
```
ifiokjr: Why the version change? This would be a breaking change.

Favna: > this would be a breaking change

Hardly, if [this extremely small set of breaking changes made by TS v3.0](https://github.com/Microsoft/TypeScript/wiki/Breaking-Changes#typescript-30) is an issue then you should update your code base rather than resist upgrading. 

But to answer your question, before I bumped it the tests failed, something to do with the imports the new tests and typings have to use.

ifiokjr (approved):

types/jest-environment-puppeteer/index.d.ts
```diff
@@ -44,11 +50,47 @@ interface JestPuppeteer {
...
+
 declare global {
     const browser: Browser;
     const context: BrowserContext;
     const page: Page;
     const jestPuppeteer: JestPuppeteer;
 }
 
-export {};
+export default PuppeteerEnvironment;
```
PranavSenthilnathan: It seems like the [underlying library](https://github.com/smooth-code/jest-puppeteer/blob/master/packages/jest-environment-puppeteer/index.js#L1) is not using `export default`, so this .d.ts file should not either. [This](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-modifying-module-d-ts.html) is the template you should be using. 

```suggestion
export = PuppeteerEnvironment;
```

types/jest-environment-puppeteer/index.d.ts
PranavSenthilnathan: It's not always reasonable to ask consumers to upgrade TypeScript versions for their projects especially if it's not necessary. In the case of this PR I think the `esModuleInterop`, which is only used for the tests, is causing the requirement for a higher version. Please remove this flag and use `import JestEnvironmentPuppeteer = require('jest-environment-puppeteer')` instead.

PranavSenthilnathan (changes requested):

types/jest-environment-puppeteer/index.d.ts
Favna: @PranavSenthilnathan I tried bumping it back down to 2.8 and removing `esModuleInterop` and I got the same error again, it comes from the jest-mock library. Here is the error it throws:
```
1> jest-environment-puppeteer failing:
1> Error: Errors in typescript@next for external dependencies:
1> node_modules/@jest/environment/build/index.d.ts(11,8): error TS1259: Module '"E:/dev/DefinitelyTyped/types/jest-environment-puppeteer/node_modules/jest-mock/build/index"' can only be default-imported using the 'esModuleInterop' flag
1> node_modules/@jest/source-map/build/getCallsite.d.ts(8,100): error TS2503: Cannot find namespace 'callsites'.
1>
1>     at E:\dev\DefinitelyTyped\node_modules\dtslint\bin\index.js:190:19
1>     at Generator.next (<anonymous>)
1>     at fulfilled (E:\dev\DefinitelyTyped\node_modules\dtslint\bin\index.js:5:58)
```

Going back to either 3.0 or 3.4 without `esModuleInterop` still has it fail, so instead I tried version 2.8 with `esModuleInterop` and I ended up with another error, once again due to external dependencies, this is a long one so I dumped it on gist: https://gist.github.com/Favna/fb37f696dcc390fa593cedcf8050751b.

Next thing I tried was TypeScript v3.0 with `esModuleInterop` which again threw an error:
```
=== ERRORS ===
Error in jest-environment-puppeteer
Error: Errors in typescript@3.0 for external dependencies:
node_modules/jest-mock/build/index.d.ts(125,162): error TS2304: Cannot find name 'Parameters'.

    at E:\dev\DefinitelyTyped\node_modules\dtslint\bin\index.js:190:19
    at fulfilled (E:\dev\DefinitelyTyped\node_modules\dtslint\bin\index.js:5:58)
Error: The following packages had errors: jest-environment-puppeteer
    at doRunTests (E:\dev\DefinitelyTyped\node_modules\types-publisher\src\tester\test-runner.ts:235:11)
    at processTicksAndRejections (internal/process/task_queues.js:86:5)
error Command failed with exit code 1.
```

Ultimately I ended up going for TypeScript version 3.4 in 53b225ada2c234a28d8d22c2f2f9fc631c02fb45 after all and with `esModuleInterop`, but I maintained the requested `export = ` and `import .. = require('...')`. I also added a few extra ExpectType tests because the latter means you cannot directly import the interfaces and I'm told I can't have other `export` statements when usin `export =`.

sheetalkamat: Its unfortunate that the packages you are depending on are using features from typescript 3+.

PR #37648, opened 2019-08-15
[yup] Use TestOptionsMessage consistently throughout

kalley:
We were looking for a way to use `setLocale` like https://github.com/jquense/yup/issues/159. Which works fine, but the TS defs for it were not great. This makes at least some of the extra options for `TestOptionsMessage` explicit (I'm sure I missed some since we're not using all of them), and makes it so that the `LocaleObject` not only uses the definitions in the `Parameters`, but also does not allow for keys that don't accept a message.

Because of `Required`, I did have to bump the typescript version. If anyone would like me to, I'm more than happy to switch to using https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#i-want-to-use-features-from-typescript-31-or-above

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/jquense/yup/issues/159
- [ ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header. _I didn't change anything specifically related to 0.27_

### DIFF
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -49,8 +49,8 @@ export type AnySchemaConstructor =
     | ObjectSchemaConstructor;
 
 export type TestOptionsMessage<Extra extends Record<string, any> = {}> =
-    | string
-    | ((params: Extra & Partial<TestMessageParams>) => any);
+  | string
+  | ((params: Extra & Partial<TestMessageParams>) => any);
 
 export interface Schema<T> {
     clone(): this;
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -11,6 +11,7 @@
 //                 Dan Rumney <https://github.com/dancrumb>
 //                 Desmond Koh <https://github.com/deskoh>
 //                 Maurice de Beijer <https://github.com/mauricedb>
+//                 Kalley Powell <https://github.com/kalley>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.1
 
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -12,7 +12,7 @@
 //                 Desmond Koh <https://github.com/deskoh>
 //                 Maurice de Beijer <https://github.com/mauricedb>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.1
 
 export function reach<T>(schema: Schema<T>, path: string, value?: any, context?: any): Schema<T>;
 export function addMethod<T extends Schema<any>>(
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -48,9 +48,9 @@ export type AnySchemaConstructor =
     | ArraySchemaConstructor
     | ObjectSchemaConstructor;
 
-export type TestOptionsMessage =
+export type TestOptionsMessage<Extra extends Record<string, any> = {}> =
     | string
-    | ((params: object & Partial<TestMessageParams>) => string);
+    | ((params: Extra & Partial<TestMessageParams>) => any);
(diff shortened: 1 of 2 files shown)


YashdalfTheGray (commented):
@kalley I'm guessing a lot of these changed came from this file getting `prettier`ed? Can you put down where the actual changes are or roll back the `prettier` changes and turn them in separately? I love prettier and would like to keep those changes but there are a lot 😄

kalley:
@YashdalfTheGray, redid both commits. First one does not update formatting (except in the tests, but not that many). Second one is formatting purely from pre-commit hook, so should be easier to read across the board.

YashdalfTheGray:
Just to make it easier for everyone, [here is the right diff](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37648/commits/6a4098eccbbe408d44449431ba9d13a9f3c8052a) that I think we should really be focusing on. The PR diff is still a little jumbled up but the feature diff linked above makes it easy to see. 

All that being said, the changes look good to me. I wanna have another person take a look at this mostly so that I didn't miss anything! Will wait till EOD Monday PST to give enough time to get someone else to 👀on it. Otherwise, I'll approve it. 

@kalley are you working under a stricter timeline or are you okay with that?

sseppola:
Just to chime in here even though I don't entirely get the changes.

What I'd like to see is a test confirming the intended behavior as it would future proof it and show the expected behavior.

Really appreciate the contribution @kalley, I'll support @YashdalfTheGray's decision if he wants to merge it.

kalley:
@sseppola Thanks for the feedback. I added some tests that hopefully illustrate the purpose this PR. Using `object` would not actually open up the params that would get passed into message, so this makes it so each method can pass the specific keys they are adding into that object. It also trims down the available keys in `setLocale` to those that accept a message, but also keeps the exact version of `TestOptionsMessage` that is a parameter in that method.

sseppola:
Cheers @kalley, thanks for adding the tests, now I see how this is useful :D

On [this line](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37648/files#diff-4399881e5520c19aa7b6e59a9aac0ca9R48) you allow `any`, but your tests all use strings, and it's not clear from the docs how this is supposed to be used.

As I'm not familiar with this myself, can you confirm `any` is the most precise type to use here?

kalley:
@sseppola Great question. We're using `i18next` with `yup`, so we actually return an object. I went ahead and added an example that shows that. It may make sense to limit that return to `string | Record<string, any>`, but I wasn't sure how other folks are using it. But maybe they should be responsible for adding it if they use it some other way. I think I will update that to restrict it so it's not opened up to just anything.

sseppola:
So if I understand this correctly, the user can basically do whatever they want here?
Is there any way the type can just be inferred? That way we're not setting up any roadblocks, and still get the more precise type

kalley:
Looking into whether I can get a generic version of the returned function to work or not

sseppola:
This is great, @YashdalfTheGray I'll give you the last word here, but I'm happy with this.

@kalley, thanks for your contribution. Would you like to join us as maintainers of these types? :)

kalley:
@sseppola I would be happy to join you as maintainers!

types/yup/index.d.ts
```diff
@@ -12,18 +12,13 @@
 //                 Desmond Koh <https://github.com/deskoh>
 //                 Maurice de Beijer <https://github.com/mauricedb>
```
sseppola: @kalley just add your name and GH profile url here :)

sseppola:
I'll leave this until tomorrow morning (CET), if Yashdalf has not replied by then I'll approve it.

YashdalfTheGray:
@sseppola @kalley looking at this right now. Should be good with it but taking one last :eyes:

YashdalfTheGray (approved):
🎉

PR #37649, opened 2019-08-15
aws-lambda: capture the two Handler options in the type system

RichardBradley:
Please fill in this template.

- [ X] Add or edit tests to reflect the change. (Run with `npm test`.)
- [ X] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [ X] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [ X] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [ X] Provide a URL to documentation or source code which provides context for the suggested changes:   This is clarifies the types based on the documentation already linked to at http://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html

### DIFF
--- a/types/aws-lambda/aws-lambda-tests.ts
+++ b/types/aws-lambda/aws-lambda-tests.ts
@@ -890,18 +890,12 @@ context.fail(str);
 /* Handler */
 const handler: AWSLambda.Handler = (event: any, context: AWSLambda.Context, cb: AWSLambda.Callback) => { };
 
-/* In node8.10 runtime, handlers may return a promise for the result value, so existing async
- * handlers that return Promise<void> before calling the callback will now have a `null` result.
- * Be safe and make that badly typed with a major verson bump to 8.10 so users expect the breaking change,
- * since the upgrade effort should be pretty low in most cases, and it points them at a nicer solution.
- */
-// $ExpectError
-const legacyAsyncHandler: AWSLambda.APIGatewayProxyHandler = async (
+const callbackStyleHandler: AWSLambda.APIGatewayProxyHandler = (
     event: AWSLambda.APIGatewayProxyEvent,
     context: AWSLambda.Context,
     cb: AWSLambda.Callback<AWSLambda.APIGatewayProxyResult>,
 ) => {
-    cb(null, { statusCode: 200, body: 'No longer valid' });
+    cb(null, { statusCode: 200, body: 'Alternative to async' });
 };
 
 const node8AsyncHandler: AWSLambda.APIGatewayProxyHandler = async (
@@ -909,7 +903,7 @@ const node8AsyncHandler: AWSLambda.APIGatewayProxyHandler = async (
     context: AWSLambda.Context,
     cb: AWSLambda.Callback<AWSLambda.APIGatewayProxyResult>,
 ) => {
-    return { statusCode: 200, body: 'Is now valid!' };
+    return { statusCode: 200, body: 'valid!' };
 };
 
 const inferredHandler: AWSLambda.S3Handler = (event, context, cb) => {
--- a/types/aws-lambda/index.d.ts
+++ b/types/aws-lambda/index.d.ts
@@ -1033,17 +1033,34 @@ export interface LexResult {
 /**
  * AWS Lambda handler function.
  * http://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html
+ */
+export type Handler<TEvent = any, TResult = any> = PromiseHandler<TEvent, TResult> | CallbackHandler<TEvent, TResult>;
+
+/**
+ * AWS Lambda handler function: Promise version (node8.10 or later)
+ * http://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html
  *
  * @param event – event data.
  * @param context – runtime information of the Lambda function that is executing.
- * @param callback – optional callback to return information to the caller, otherwise return value is null.
- * @return In the node8.10 runtime, a promise for the lambda result.
  */
-export type Handler<TEvent = any, TResult = any> = (
+export type PromiseHandler<TEvent = any, TResult = any> = (
+    event: TEvent,
+    context: Context
+) => Promise<TResult>;
+
+/**
+ * AWS Lambda handler function: callback version
+ * http://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html
+ *
+ * @param event – event data.
+ * @param context – runtime information of the Lambda function that is executing.
+ * @param callback – the callback to return information to the caller
+ */
+export type CallbackHandler<TEvent = any, TResult = any> = (
     event: TEvent,
     context: Context,
     callback: Callback<TResult>,
-) => void | Promise<TResult>;
+) => void;
 
 /**
  * Optional callback parameter.
--- a/types/lambda-tester/lambda-tester-tests.ts
+++ b/types/lambda-tester/lambda-tester-tests.ts
@@ -54,7 +54,7 @@ lambdaTesterInstance().expectError((error: TError) => {
     const t: string = error.message;
 });
 
-const apiGateway: APIGatewayProxyHandler = (event, _context) => {};
+const apiGateway: APIGatewayProxyHandler = async (event: APIGatewayProxyEvent, _context: Context) => { };
 
 const s3Lambda = lambdaTester(apiGateway).event({} as APIGatewayProxyEvent);
 


RichardBradley:
Currently, the "Handler" type attempts to capture both options (callback v.s. Promise) in a single type, which makes it hard to understand and also leads to undesirable behaviour, like the "callback" parameter being required when in fact it is optional.

I have split the type in two, to allow the type system to express the choice between a "callback" handler and a "promise" based handler.

There is one minor drawback: the test case that I have removed from the `aws-lambda-tests.ts` file no longer failed, because its incorrect return value of `Promise<null>` gets automatically accepted by a declared return type of `void`, whereas it used to be rejected by a return type of `Promise<TResult> | void`. This seems rather strange to me, but I don't know how to fix it.

RichardBradley:
This fixes a problem for me: I have some unit tests which were invoking my Promise based handlers directly, and TS was suggesting that the "callback" param was required, when it is in fact optional.

simonbuchan (commented):
Hey, sorry for the slow feedback, I somehow missed the notification.

I'm not particularly happy with this as it is, unfortunately, since it seems like it's at least partially motivated by a misunderstanding of typescript, and looks like it loses an important protection against misuse of the API.

Firstly, a required parameter is required for the *caller*, here AWS, not for functions implementing the type. The context and callback parameters are definitely required for AWS to provide, since there are handlers out there using them! Just use `null as any` or something in tests if you don't want to fully mock them.

Second, the $ExpectError test is to catch changes like this, that would allow handlers that will break in node 8.10 due to it now looking at if you are returning a promise, and using that instead of the callback in my tests if there API behavior, as described in the comment for the test.

RichardBradley:
Thanks for taking a look.

> Firstly, a required parameter is required for the caller, here AWS, not for functions implementing the type. ... Just use null as any or something in tests if you don't want to fully mock them.

The interface type is a two way contract. Both functions implementing the interface (i.e. user's Lambdas) and callers invoking the interface (i.e. AWS Lambda runtime or tests that simulate it) need to understand and implement the interface.

The current interface is misleading in that it suggests that all functions should accept (or ignore) a "callback" param, and that they can return "`void | Promise<TResult>`". It is clearer, if the type system can convey it, to show that these two are linked and that functions are either "callback" or "promise" based, as I have done here.

> Second, the $ExpectError test is to catch changes like this, that would allow handlers that will break in node 8.10 due

I addressed that in my comment above. 
AWS support for node 8 is now 16 months old ([release announcement](https://aws.amazon.com/blogs/compute/node-js-8-10-runtime-now-available-in-aws-lambda/)), so perhaps losing this guard is a reasonable tradeoff.

sheetalkamat:
Waiting on @simonbuchan to re-review

simonbuchan:
I'm not convinced. I'm not going to straight up reject this in the face of strong favor from the rest of the community, but I don't think that's actually there, and it is a biggish change to a critical type in order to make it more technically complicated in order to make it subjectively easier to understand. That is, there seems to be a heck of a lot of potential downside, (including potential for breaking change in a patch version!) for only a very tiny, subjective, upside. Preventing that is a big part of why there is a community review on these. (Note, you don't have to have contributed a PR here to review!)

> The current interface is misleading in that it suggests that all functions should accept (or ignore) a "callback" param, and that they can return "`void | Promise<TResult>`". It is clearer, if the type system can convey it, to show that these two are linked and that functions are either "callback" or "promise" based, as I have done here.

Except, it isn't in practice. You can do stuff like this, for example, and it makes sense, and will work fine:

```ts
export const handler: Handler<FooEvent, FooResult> = (event, context, callback) => {
  if (event.type === "foo") {
    return handleFoo(event, context); // returns a promise
  } else if (event.type === "bar") {
    handleBar(event, context, callback); // uses callback
  } else {
    callback(new Error("Unknown event type: " + event.type)); // haven't touched this yet.
  }
};
```

That is, returning *either* `void` or `Promise<T>`, regardless of what parameters you give it is perfectly fine and useful. The problem was returning a promise for something that wasn't the result, e.g.:

```ts
// Works, since the callback is invoked before the returned promise is resolved.
// I deliberately took a breaking change in the major version update in this case to prevent...
export const scary: Handler<FooEvent, FooResult> = async (event, context, callback) => {
  const value = await something(event);
  callback(null, value);
};

// Worked earlier, but broken in 8.10, will resolve the promise with undefined before invoking the callback, meaning the lambda result is `null`
export const broken: Handler<FooEvent, FooResult> = async (event, context, callback) => {
  const value = await first(event);
  second(value, callback);
}
```

So while your type represents intended usage, it's actually making it harder to understand what the actual type and required behavior is: you *don't* have to return a promise if the environment doesn't give you a callback, because it always will. The downside is someone might add the parameters when they don't need to, and then TS will yell at them for having unused parameters, and then they get angry, and then they get embarrassed when they find out that it's fine and they didn't need to add those. Someone might learn the language they are using better is not a particularly scary possibility for me. Tests always have to deal with ugly hacks if they're taking shortcuts, that's nothing new.

RichardBradley:
> I'm not going to straight up reject this in the face of strong favor from the rest of the community, but I don't think that's actually there

You are the only one to have commented, so "the community"s opinion of this patch is pure speculation at this point.

You should reject the patch if you want, I won't be offended. You have made some good points.

> You can have code like this, for example, and it makes sense, and will work fine:

Yes, that's right, and would be forbidden by my change.

Perhaps the type system can't express the contract here. 

Do you have any better suggestions? Perhaps the current types are as close as we can get to the contract.

RichardBradley:
Maybe what I should do is introduce a type like "PromiseHandler" (from this PR) into my own code, then use that for my Handlers, ignoring the provided aws types. This will tighten & clarify my code, simplify my unit tests and may catch the odd error. It should cause no problems at runtime, as all the types get compiled away anyway. It's a shame that I can't use the `@types/aws-lambda` to achieve this though.

jagregory:
Chiming in to say I wrap the @types/aws-lambda handlers to achieve precisely this. I would rather not have to wrap them, though I do appreciate the desire to not cause unnecessary breakages. If there's a way we can provide stricter signatures without breaking backwards compatibility, that would be nice.

There's a meta discussion going on here: should the types reflect the reality of what is technically possible at run time, or should they encourage good patterns?

Amazon's [docs for Nodejs Lambda handlers](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html) makes a clear delineation between async handlers and non-async handlers, where all their examples omit the callback argument if the function is async. They don't explicitly say you can't use the callback, but it's quite clear they're deliberately trying to keep the two approaches separate.

Quoting the documentation (emphasis mine):

> **The runtime passes three arguments to the handler method**. The first argument is the event object, which contains information from the invoker. The invoker passes this information as a JSON-formatted string when it calls Invoke. When an AWS service invokes your function, the event structure varies by service.

So this lends itself to the argument that the signature should support both, but then...

> The third argument, callback, is a function that **you can call in non-async functions** to send a response. [ ... ]
> 
> **For async functions, you return a response, error, or promise to the runtime instead of using callback.**

But that again whilst there isn't a technical limitation against mixing the two, Amazon is clearly documenting them as two separate approaches.

An additional datapoint: I am aware of at least one popular library (serverless-offline) that raises warnings if you mix callbacks and promises in lambdas. So there's some (if unofficial) precedence out there for mixing the two approaches being undesirable.

I personally have no horse in this race. I've already wrapped the types and will continue to do so if this doesn't get merged, but I would happily delete my types if this was merged. For what it's worth, I personally think the types should reflect the intent and not the technicalities; after all, this is about layering types onto something that was built for Nodejs which is inherently going to be less strict than if it was designed from the ground up with types in mind.

jagregory:
Just to extend my comment a little bit, I think the scenarios Simon raised are still valid and we shouldn't prevent them outright. My preference if it's at all possible would be to provide some "higher level" types that are something like CallbackHandler and PromiseHandler, but then also a "lower level" AnythingGoesCallback (name pending...) that is still available if you need to fall back to it.

For me, it's not about misunderstanding the language. I want the language to support me and help catch mistakes that I may be inadvertently making. The vast majority of the time, if I have an async Lambda I do not want to use the callback and vice-versa. If I'm mixing the two, I am only doing it in very specific circumstances, and all other places I want to be warned that I'm about to do something unintentional.

simonbuchan:
> You are the only one to have commented, so "the community"s opinion of this patch is pure speculation at this point.

Of course! In fact that's why I didn't mark my review as "Request Changes", as I'm fully aware this is my preference only. Even the potential breaking change is not that uncommon in typing packages, and so it shouldn't kill this if the fix is simple enough and makes for better cleaner code: I just want to have a good story for why, when the inevitable revert PR comes in! (that said, read below)

> You should reject the patch if you want, I won't be offended. You have made some good points.

More of a DT thing really: marking a review as "Request Changes" means that the bot talks about waiting for changes from you, it gets taken off the review queue etc. Since this immediately went to the YSYL queue, and got manually marked Revision needed not that big of a difference, but eh.

> An additional datapoint: I am aware of at least one popular library (serverless-offline) that raises warnings if you mix callbacks and promises in lambdas. So there's some (if unofficial) precedent out there for mixing the two approaches being undesirable.

Yes, it's certainly undesirable to mix the styles: the question is how much help is this change. Keep in mind, I'm guessing the most popular usage of this package is to just use the event and result types directly or to use the service-specific `AWSGatewayProxyHandler` etc. predefined handler types, so it's not like they are likely to see the split handler types outside of an extra line in error messages.

With some more time, I put together [a test on playground](http://www.typescriptlang.org/play/index.html#code/PTAEFEA8EsGcBdoDsDmp4E8AOBTAXAFAg6RYD2ATvOtjqABICGSAJgDY4UA8AKuAG44k1ALyhmGADSgeAJRywArm1HikGAHygxAChyDheGQKHxpAYzLCS8IwGEr8GxcZs2AI0bmA1vdcevb155JRUNAEptLX4yaBZQAB9QAAUKMgBbOBxghWV4DQBuIjBk2RpcQhJyKnK6JlYObj4DVQlpOVyVbTVNbtSMrPr2Tl4TYXaQvK0ku39PHyHG0ZaJzvyigirKakxcFLTM2BxFkebTbraZSa6xCS1dAlAn0H1TIzPxx+fLa0hbUAcv3gBEiIi0-UO2Q6oXWm1I21qALmgROTTGrXUqxhF3U91AOi+T1ehmMK0JoB+Tj+9kcznJ5mRPj8bnmQWhU0kIKioBicSKWxquzosxZgRy2NuuO6egoaQoAH4jOBZZREqAkMo2GqEBRkChpBQ1oqrmtQdFYix+fCasgnBQAGZeYW0v6gADeAF8CMUZAodrQKRksIxgSwcOY2IxDaB7YokOZEFZQOlGN4oRodOF3kUwxGo3RY-HE0h0Kn0zp+K5FPgZFmeRaNvAy1xUVwdXrpBr0u5OBoMym0y3mMNuO3UJ3FN3exnwuEik3B62x-r1ZOexQ+zoB9kRQEFsOlsuJ1ONzO5wQF9kl-BdePVyfN9uuBDBgeRkf7+vN7ONiAIKRw2oTg5VgUAk2QSxZUA8QKBQWAL2bVsu3XY8v37RC31HG8O0-adM3PS8hwaEZkM4VC8KfXdWWvW8V1I098PnDDiO4ejyIYp8XyOGicPo79zz-KBcATF4VQoUDwPjShDREw0YQQxdMLbbC72XR9mJHZTaPY-imMUlitJwtT0MHKiUSUj8+LPPSrwslSV2Mrdmy4447O03CGJ-b0-x4P1AyBRRXHKPUCB+BBAyghMzKZBg3N4tde2lYkzH8qkUoZUUfDNd16UZbwdA1NxpGDcScAASWEPQWi8j0ijC6hIJk+AAEFYAweMjB4u8+IuNr43xZLsrdclDXgRQKBLEqjgq+AqtMc9atCqxwsawDWva8weFoFhOri7qEo3XqNoGlojGXIaRpwMaJtAKbysqwaikW+qIqa9b4wAWWgSAcB22KDMsg68UYPrzBO0wLBddK8ou55QGge1weEbQxAAIntMgyFRyJhrhuHRvGksAEYijhj0XjYI4crx54CZujK93ywq2GkAAmc8yYIZ7loa6TAIAZUYe0cC+n6-q6uigaSlZUucCkYe5XHngRpHRDRjGsZx8l8auwn9gGI4ADo5LINhBB0ImOeecmcEpuglZpumSwZ1kCs1NmraeL1Fp9XyVqhwK3AwWp4mAyh4JeiC+YTdFot8f7NLYjy8Tm8ZZb+FxMu8WHvjyt2ituqNpoe6qFrqnn4akyL4HRd7zF2gyk5625QdVnOnidwvSpm1P4DLpakHCqPq9r0GttwcWlKbqWW+O3uzvs9vQE7u6e8erny8H6hh6a0eNtF36G8Tg6dKO-re8hoFM8ZpeVd7lHQHRzHsepmnl91m6SfJG27dfx2P+dnnZmHtSbWw3gPIeVdd4tEFsLA+k9G4n2TtLCG6doZZ1voje+Ih1bPy1m-d+10SwuWNgoU25tLagK9hTKmDs8adxdoEfOLNQDsyoaAb2GxI5QMAtceAccj6HnstINSKC06UjlowrKitcpZ2YcVIu91ZqDX7tw1aCY+F10Ee+YRoBRGz3PoNGROsiFd2Lso6qT1N6QPUfATRY9traKwu5fR4hW7zz0YvYxtMAFmKUb3VRFcd68LWHXeBTjDKqXssDdxyVL5pWvqyTBqsH5P01n-ehviv5kxofbbWPjTFSKZu7VhnsOHgLUdHOxaxYEi2+ofBOQiXHRLEelKGiTAjJOwbg9JdCTF6xISbM2OALZlJ-rQ-JHdfFFPkaU9hnCCBAA), of tests of the handler type, and it seems to show up a few pretty critical issues:

- The event type is no longer inferred from a typed `Handler` and will get `any`, erroring on strict: this is actually critical and a huge breaking change, even if it seems like it's a TS bug?
- Adding `PromiseHandler` seems to break return type checking: `PromiseHandler<A, B>` is assignable to `Handler<A, C>` even with incompatible `C`. `incorrectResultAsyncTyped` returns a `number` to a handler with a `string` result type without error. Not sure what's up there either.
- The new type doesn't even error on mixing callback and promise results, as the intention seems to be: neither the `correctAsyncMixed` (the potentially dangerous case I banned earlier) nor the `correctSafeMixed` have any errors.

simonbuchan (changes requested):
See my comment above.

RichardBradley:
I don't know how to fix this; I suspect it's not possible to meet all these requirements in Typescript.

Unless anyone has any better ideas, I'll close this and I'll continue to use the per-project workaround that I and @jagregory describe above.

RichardBradley:
That's a rather short idle timeout for an open-source project!

Sure, fine to close. If anyone else wants to pick this up, that would be great. I'll go with my project-specific workaround for now.

PR #37663, opened 2019-08-16
Switch await-timeout from export default to export =

ghost:
Please fill in this template.

`export default` implied an es module. Unless `esModuleInterop` is enabled, then `import Timeout from 'await-timeout';` will return `undefined`, since it expects a `default` property.

`export =` implies a CJS module, and fixed some linting problems when using `require`. E.g. `const Timeout = require('await-timeout');` no longer complains when trying to use `Timeout` (since it thinks the only property is `default`).

### DIFF
--- a/types/await-timeout/index.d.ts
+++ b/types/await-timeout/index.d.ts
@@ -2,9 +2,9 @@
 // Project: https://github.com/vitalets/await-timeout
 // Definitions by: Sean Zhu <https://github.com/szhu>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.1
+// TypeScript Version: 2.7
 
-export default Timeout;
+export = Timeout;
 
 declare class Timeout {
     static set(delay: number): Promise<undefined>;
--- a/types/await-timeout/tsconfig.json
+++ b/types/await-timeout/tsconfig.json
@@ -4,6 +4,7 @@
         "lib": [
             "es6"
         ],
+        "esModuleInterop": true,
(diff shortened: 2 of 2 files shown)


szhu:
Hmmm, I don't have enough background to know what's best -- I want to make sure we're not breaking setups for anyone. 

- The `esModuleInterop` part makes sense.
- I've heard some stuff about how `export =` might be bad in some cases, but I don't have the full context anymore.

Can we get a DefinitelyTyped maintainer in here? (I don't know how to summon one immediately, but from the comment above, it looks like the bot will add one after a week.)

Let me know if you have any thoughts on this, or if you're sure that these changes won't break anyone's existing setups!

dsommerich:
Yeah breaking existing code is something I'm concerned about. I don't *think* it will break anything.

The only reason I could see it breaking is if someone doesn't have `esModuleInterop` enabled (since that's required for `export =` to be valid), but without it, the module wouldn't work anyway. It might also break in Typescript < 2.7, but `esModuleInterop` didn't exist until then, so I don't think it would have worked anyway. :man_shrugging: 

It is possible to do a version bump? (I'm not too familiar with DT, so I don't know how that works).

szhu (approved):
Okay, I took a look in this repo and got 9k matches for `export =` (including for lodash), so I assume it's fine!

I think dt automatically handles versions and releases, and we don't need to worry about breaking someone's code if they don't upgrade packages.

PranavSenthilnathan:
I'm surprised this even worked before without this change...

PranavSenthilnathan:

PR #37673, opened 2019-08-16
[Node]: String.prototype has additional ES2019 methods

Luxcium:
The properties trimStart and trimEnd are preferred. The trimLeft and trimRight properies are provided principally for compatibility with old code. It is recommended that the trimStart or trimEnd  properties be used in new ECMAScript code.

https://github.com/nodejs/node/commit/12a1b9b8049462e47181a298120243dc83e81c55#diff-5fb85405f1ec035ec82ebead4371e7e8

Please fill in this template.

https://www.ecma-international.org/ecma-262/10.0/index.html#sec-string.prototype.trimend
https://www.ecma-international.org/ecma-262/10.0/index.html#sec-string.prototype.trimstart

B.2.3Additional Properties of the String.prototype Object : 
B.2.3.15String.prototype.trimLeft ( )
B.2.3.16String.prototype.trimRight ( )

trimStart and trimEnd on String.prototype as better-named alternatives to the widely implemented but non-standard String.prototype.trimLeft and trimRight built-ins. 

The initial value of the trimLeft property is the same function object as the initial value of the  String.prototype.trimStart property.
The initial value of the trimRight property is the same function object as the initial value of the  String.prototype.trimEnd property.

21.1.3.28String.prototype.trimEnd ( )
This function interprets a String value as a sequence of UTF-16 encoded code points, as described in 6.1.4.

The following steps are taken:

Let S be this value.
Return ? TrimString(S, "end").
NOTE
The trimEnd function is intentionally generic; it does not require that its this value be a String object. Therefore, it can be transferred to other kinds of objects for use as a method.

21.1.3.29String.prototype.trimStart ( )
This function interprets a String value as a sequence of UTF-16 encoded code points, as described in 6.1.4.

The following steps are taken:

Let S be this value.
Return ? TrimString(S, "start").
NOTE
The trimStart function is intentionally generic; it does not require that its this value be a String object. Therefore, it can be transferred to other kinds of objects for use as a method.

### DIFF
--- a/types/node/globals.d.ts
+++ b/types/node/globals.d.ts
@@ -146,15 +146,22 @@ interface SymbolConstructor {
     readonly observable: symbol;
 }
 
-// Node.js ESNEXT support
+// Node.js ES2019/ESNEXT support
 interface String {
-    /** Removes whitespace from the left end of a string. */
+    /** The property trimStart is preferred. 
+    The trimLeft property is provided principally for compatibility with old code. 
+    It is recommended that the trimStart property be used in new ECMAScript(ES2019) code. */
     trimLeft(): string;
-    /** String.prototype.trimLeft.name === 'trimStart'; // ? true */
-    trimStart(): string;
-    /** Removes whitespace from the right end of a string. */
+    
+    /** The property trimEnd is preferred. 
+    The trimRight property is provided principally for compatibility with old code. 
+    It is recommended that the trimEnd property be used in new ECMAScript(ES2019) code. */
     trimRight(): string;
-    /** String.prototype.trimRight.name === 'trimEnd'; // ? true */
+
+    /** Removes whitespace from the left end of a string. */
+    trimStart(): string;
+
+    /** Removes whitespace from the right end of a string. */
     trimEnd(): string;
 }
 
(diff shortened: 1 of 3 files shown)


islishude (changes requested):
These have been already added to the stdlib

https://github.com/microsoft/TypeScript/blob/master/lib/lib.es2019.string.d.ts

SimonSchick (changes requested):
What @islishude said, use the existing lib.

Luxcium (commented):
The test was missing the test string initialization `const s = 'foo';`

Luxcium:
@SimonSchick what existing lib should I be using?
 I have error: **`Property 'trimStart' does not exist on type 'string'.ts(2339)`**

with in /tsconfig.json:
```json
{
    "compilerOptions": {
    "lib": [
            "dom",
            "esnext",
            "dom.iterable",
            "es2017.string",
            "es2019.string",
            "es2020.string"
        ],
    }
}
```

and with in package.json:
```json
{
    "devDependencies": {
        "@types/node": "^12.7.2",
        "typescript": "^3.5.3",
    }
}
```

Luxcium:
to many commits and corrections will close PR

Luxcium:
to many commits and corrections will close PR

PR #37688, opened 2019-08-16
Adding @testing-library/jest-dom type definitions

weyert:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/testing-library__jest-dom/index.d.ts
+++ b/types/testing-library__jest-dom/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for @testing-library/jest-dom 4.0.0
+// Type definitions for @testing-library/jest-dom 4.0
 // Project: https://github.com/testing-library/jest-dom
 // Definitions by: Damien Senger <https://github.com/hiwelo>
 //                 Łukasz Fiszer <https://github.com/lukaszfiszer>
--- a/types/testing-library__jest-dom/index.d.ts
+++ b/types/testing-library__jest-dom/index.d.ts
@@ -1,0 +1,35 @@
+// Type definitions for @testing-library/jest-dom 4.0.0
+// Project: https://github.com/testing-library/jest-dom
+// Definitions by: Damien Senger <https://github.com/hiwelo>
+//                 Łukasz Fiszer <https://github.com/lukaszfiszer>
+//                 Revath S Kumar <https://github.com/revathskumar>
+//                 Ernesto García <https://github.com/gnapse>
+//                 Weyert de Boer <https://github.com/weyert>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+declare namespace jest {
+    interface Matchers<R> {
+        /**
+         * @deprecated
+         */
+        toBeInTheDOM(container?: HTMLElement | SVGElement): R;
+        toBeInTheDocument(): R;
+        toBeVisible(): R;
+        toBeEmpty(): R;
+        toBeDisabled(): R;
+        toBeEnabled(): R;
+        toBeInvalid(): R;
+        toBeRequired(): R;
+        toBeValid(): R;
+        toContainElement(element: HTMLElement | SVGElement | null): R;
+        toContainHTML(htmlText: string): R;
+        toHaveAttribute(attr: string, value?: any): R;
+        toHaveClass(...classNames: string[]): R;
+        toHaveFocus(): R;
+        toHaveFormValues(expectedValues: { [name: string]: any }): R;
+        toHaveStyle(css: string): R;
+        toHaveTextContent(text: string | RegExp, options?: { normalizeWhitespace: boolean }): R;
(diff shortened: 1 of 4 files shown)


PranavSenthilnathan:
Aren't these types already being exposed through the package with [the d.ts](https://github.com/testing-library/jest-dom/blob/master/extend-expect.d.ts)?

image

gnapse:
@PranavSenthilnathan if you can, checkout the discussion in testing-library/jest-dom#123. It gives you some context about what we're trying to achieve here.

sheetalkamat (changes requested):
This change is not needed. The package can add "types" field to the package and thats the desired way of adding types. Refer to https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/declaration%20files/Publishing.md

sheetalkamat (changes requested):
This change is not needed. The package can add "types" field to the package and thats the desired way of adding types. Refer to https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/declaration%20files/Publishing.md

weyert:
> This change is not needed. The package can add "types" field to the package and thats the desired way of adding types. Refer to https://github.com/Microsoft/TypeScript-Handbook/blob/master/pages/declaration%20files/Publishing.md

Yes, I understand but we decided to add as a dependency, like the other @testing-library packages. I kindly request to reconsider this decision as this would this library to not have type definitions.

@johnnyreilly @sheetalkamat

johnnyreilly:
Hey @sheetalkamat

I didn't quite follow what you meant when you said this:

> This change is not needed. The package can add "types" field to the package and thats the desired way of adding types.

This PR adds types to Definitely Typed which will result in an `@types/...` package being generated. The actual package doesn't want to maintain it's own type definitions and would rather have them maintained via Definitely Typed. To my knowledge the `"types"` solution via `package.json` only works for local type definitions that ship with the library. That doesn't apply in this case.

As far as I can tell, this PR should be safe to merge; I've reopened it. Any objections?

types/testing-library__jest-dom/index.d.ts
```diff
@@ -0,0 +1,35 @@
+// Type definitions for @testing-library/jest-dom 4.0
+// Project: https://github.com/testing-library/jest-dom
+// Definitions by: Damien Senger <https://github.com/hiwelo>
+//                 Łukasz Fiszer <https://github.com/lukaszfiszer>
+//                 Revath S Kumar <https://github.com/revathskumar>
+//                 Ernesto García <https://github.com/gnapse>
+//                 Weyert de Boer <https://github.com/weyert>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
```
sheetalkamat: According to https://github.com/testing-library/jest-dom#usage this you need to have `extend-expect` to be present that contains the globals

types/testing-library__jest-dom/index.d.ts
```diff
@@ -0,0 +1,35 @@
...
+
+declare namespace jest {
```
sheetalkamat: This is not going to work with `import {toBeInTheDocument, toHaveClass} from '@testing-library/jest-dom'` as per https://github.com/testing-library/jest-dom#usage

sheetalkamat (changes requested):

types/testing-library__jest-dom/index.d.ts
weyert: Yes, as this library is extending expect() to have more matchers.

types/testing-library__jest-dom/index.d.ts
weyert: I am not sure how to support this alternative way of using the library with TypeScript. Do we want to support the not typical usage of the library? Any tips how this could be resolved?

types/testing-library__jest-dom/index.d.ts
sheetalkamat: What I meant was there needs to be a file `extend-expect.d.ts` that defines all the global declaraitons.

types/testing-library__jest-dom/index.d.ts
sheetalkamat: Please read https://github.com/DefinitelyTyped/DefinitelyTyped#how-do-i-write-definitions-for-packages-that-can-be-used-globally-and-as-a-module for details on how to write definitions that are available as module or global

weyert:
Hello @sheetalkamat, @johnnyreilly. Looks like there is a little change of plan regarding this type definition. The idea now is to keep the type definition in the actual package `@testing-library/jest-dom` but also define it in DT but the definition living in DT would depend on the `@testing-library/jest-dom` typescript definiton and just import this one.

Would this be acceptable?

weyert:
See response of @sheetalkamat for the above question here:
https://github.com/testing-library/jest-dom/issues/123#issuecomment-523192376

weyert:
Closing pull request as it's not needed

PR #37692, opened 2019-08-16
[@testing-library/cypress] New type definition

simjes:
This is related to testing-library/cypress-testing-library#68, moving typings to `DefinitelyTyped`.

A couple of questions:
- Is the import of jquery for typings correct in `index.d.ts`? I was thinking I would need to add it to `paths` in `tsconfig.json`, but `npm run test` will fail with `In JQuery, unexpected path mapping for JQuery: 'jquery'`.
- Any suggestions to improve the typings to get rid of unnecessary generics? I've currently disabled the rule in linting, but I see this is located under [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes)

Any other feedback to improve this is appreciated

---
- [ ] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes). **See question**

If adding a new definition:
- [x] Create it with `dts-gen --dt`, not by basing it on an existing project. **Inspired by other typings for testing-library__\***

### DIFF
--- a/types/testing-library__cypress/index.d.ts
+++ b/types/testing-library__cypress/index.d.ts
@@ -1,0 +1,698 @@
+// Type definitions for @testing-library/cypress 4.1
+// Project: https://github.com/testing-library/cypress-testing-library
+// Definitions by: Aaron Mc Adam <https://github.com/aaronmcadam>
+//                 Basti Buck <https://github.com/ppi-buck>
+//                 Stefano Magni <https://github.com/NoriSte>
+//                 Justin Hall <https://github.com/wKovacs64>
+//                 Brian Ng <https://github.com/existentialism>
+//                 Airat Aminev <https://github.com/airato>
+//                 Simon Jespersen <https://github.com/simjes>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+import {
+    Matcher,
+    MatcherOptions as DTLMatcherOptions,
+    SelectorMatcherOptions as DTLSelectorMatcherOptions,
+} from '@testing-library/dom'
+
+import * as JQuery from 'jquery'
+
+export interface CTLMatcherOptions {
+    timeout?: number
+}
+
+export type MatcherOptions = DTLMatcherOptions | CTLMatcherOptions
+export type SelectorMatcherOptions = DTLSelectorMatcherOptions | CTLMatcherOptions
+
+declare global {
+    namespace Cypress {
+        interface Chainable<Subject = any> {
+            /**
+             * dom-testing-library helpers for Cypress
+             *
+             * `getBy*` APIs search for an element and throw an error if nothing found
+             * `getAllBy*` APIs search for all elements and an error if nothing found
+             * `queryBy*` APIs search for an element and returns null if nothing found
+             * `queryAllBy*` APIs search for all elements and return empty array if nothing found
+             *
+             * @see https://github.com/testing-library/cypress-testing-library#usage
+             * @see https://github.com/testing-library/dom-testing-library#table-of-contents
+             */
+            queryByPlaceholderText<E extends Node = HTMLElement>(
+                id: Matcher,
+                options?: MatcherOptions,
+            ): Chainable<JQuery<E>>
+            queryByPlaceholderText<K extends keyof HTMLElementTagNameMap>(
+                id: Matcher,
+                options?: MatcherOptions,
+            ): Chainable<JQuery<HTMLElementTagNameMap[K]>>
+
+            /**
+             * dom-testing-library helpers for Cypress
+             *
+             * `getBy*` APIs search for an element and throw an error if nothing found
+             * `getAllBy*` APIs search for all elements and an error if nothing found
+             * `queryBy*` APIs search for an element and returns null if nothing found
+             * `queryAllBy*` APIs search for all elements and return empty array if nothing found
+             *
+             * @see https://github.com/testing-library/cypress-testing-library#usage
+             * @see https://github.com/testing-library/dom-testing-library#table-of-contents
+             */
+            queryAllByPlaceholderText<E extends Node = HTMLElement>(
+                id: Matcher,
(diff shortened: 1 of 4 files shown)


weyert:
@johnnyreilly Could you have a look at it? :)

wKovacs64:
@simjes do you have an example project of these working? I'm running into JQuery conflicts and I think it's because these new types import the latest version of `@types/jquery`, but [Cypress ships its own version](https://unpkg.com/browse/cypress@3.4.1/types/jquery/), maybe?

simjes:
@wKovacs64 check out https://github.com/simjes/cypress-testing-library-example. Had a question on the PR about the JQuery typings but it was merged without any feedback, so I assumed it was correct.

wKovacs64:
@simjes I ran `npx typescript --noEmit --project cypress` in your example repo and got 56 errors (the same JQuery conflicts I got in my own project). Does it come back clean for you?

If it is indeed a `@types/jquery` version conflict, I'm not sure of the best way to handle that. We could change the dependency version in `@types/testing-library__cypress` from `*` to the version Cypress is using (currently 3.3.6) but that means every time Cypress updates theirs, this would need updated as well. A better solution might be to import JQuery from `cypress/types/jquery` instead of from `@types/jquery` but I have no idea if you can do that in DefinitelyTyped.

simjes:
My bad, getting the same errors. I'll look into it, any tips on how to best solve this @johnnyreilly?

wKovacs64:
Not sure who else to ping on something like this. Maybe we should specify the JQuery version for now, so at least the migration path is possible? 🤔

types/testing-library__cypress/tslint.json
```diff
@@ -0,0 +1,13 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "semicolon": [
```
sheetalkamat: Please open new PR removing all the rules here and fix errors. We do not want new definition with custom rules. If the rule is absolutely necessary disable it at the location, instead of for whole package, so we can review correctness of it.

types/testing-library__cypress/index.d.ts
```diff
@@ -0,0 +1,698 @@
...
+            getAllByRole<K extends keyof HTMLElementTagNameMap>(
+                id: Matcher,
+                options?: MatcherOptions,
+            ): Chainable<Array<HTMLElementTagNameMap[K]>>
+        }
+    }
+}
+
+declare const Cypress: Cypress.Chainable
+export default Cypress
```
sheetalkamat: This needs to be fixed asap. There is no default being exported here.. 
```
// a.js
const x = require("@testing-library/cypress");
console.log(x.default);

> node a.js
undefined
```

sheetalkamat:
@simjes We want you to create a PR fixing the additional comments that are our principle guidelines for PRs.
cc: @johnnyreilly Please note that we would want to hold back the PRs that dont meet the principle guidelines.

sheetalkamat:
Also for the issue you are facing please try using cypress as the dependency in the types to see if that helps with the build errors

simjes:
@sheetalkamat thanks for the feedback, I will start fixing.

johnnyreilly:
> cc: @johnnyreilly Please note that we would want to hold back the PRs that dont meet the principle guidelines.

Totally fine @sheetalkamat. It's not clear to me which principle guidelines for PRs that are being breached by this PR? 

I've looked at the guidance offered here:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/.github/PULL_REQUEST_TEMPLATE.md

And here:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#create-a-new-package

But the points mentioned seem to be different to the ones you've been referencing. Have I missed something, or do these guidelines need updating with extra advice?

wKovacs64:
> Also for the issue you are facing please try using cypress as the dependency in the types to see if that helps with the build errors

Oh, depend on `cypress` itself instead of `jquery`? Interesting, hopefully that works! 🤞

simjes:
@sheetalkamat I've made some progress on this, but I'm getting typings error from cypress/jquery when running `dtslint`, as it uses `typescript@next`. If I run `dtslint` with the `--localTs` flag, it completes with no errors using `typescript@3.5.3`.

Do you have any suggestions? Maybe I should move this to stackoverflow or open a new issue instead?

I've got a repo [here](https://github.com/simjes/DefinitelyTyped/tree/fix-testing-cypress/types/testing-library__cypress) if you would like to take a look. I've temporarly added `typescript` as a dependency on `testing-library__cypress` to test `dtslint --localTs`

The error:
```
node_modules/cypress/types/jquery/index.d.ts(8155,87): error TS2344: Type '"button" | "view" | "altKey" | "bubbles" | "cancelable" | "changedTouches" | "ctrlKey" | "detail" | "eventPhase" | "metaKey" | "pageX" | "pageY" | "shiftKey" | "char" | "charCode" | ... 13 more ... | "touches"' does not satisfy the constraint '"repeat" | "button" | "code" | "view" | "y" | "altKey" | "bubbles" | "cancelable" | "changedTouches" | "ctrlKey" | "detail" | "eventPhase" | "metaKey" | "pageX" | "pageY" | "shiftKey" | ... 54 more ... | "DOM_KEY_LOCATION_STANDARD"'.
  Type '"toElement"' is not assignable to type '"repeat" | "button" | "code" | "view" | "y" | "altKey" | "bubbles" | "cancelable" | "changedTouches" | "ctrlKey" | "detail" | "eventPhase" | "metaKey" | "pageX" | "pageY" | "shiftKey" | ... 54 more ... | "DOM_KEY_LOCATION_STANDARD"'.
```

wKovacs64:
@simjes That error is a Cypress issue and I think they need to update the version of `@types/jquery` they are bundling. See https://github.com/cypress-io/cypress/issues/5065. If you have a fix in place for the original jQuery conflicts and this is the only hold-up, we can probably push it through?

PR #37696, opened 2019-08-17
Express: Revert #37502, because it breaks things and doesn't solve the real problem.

jwalton:
Please fill in this template.

This reverts the changes introduced in #37502.  As I mentioned [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37502#issuecomment-522203464), these changes are breaking, and create a lot of fussy make work and add a lot of unnecessary type checks to customer code.  And while there is a real problem the change is trying to address, there are other better approaches to addressing it that don't offload this work onto the caller, and could possibly be done in a non-breaking fashion.

Since this #37502 is a breaking change for an extremely popular project (express), it has HUGE impact.  We should revert it until we can either implement or rule out other possible solutions.

### DIFF
--- a/types/express-http-proxy/express-http-proxy-tests.ts
+++ b/types/express-http-proxy/express-http-proxy-tests.ts
@@ -95,7 +95,7 @@ proxy("www.google.com", {
 
 const proxyOptions: proxy.ProxyOptions = {};
 
-app.use("/proxy/:port", proxy(req => "localhost:" + (Array.isArray(req.params) ? {} : req.params).port));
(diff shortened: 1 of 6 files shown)


OliverJAsh:
@jwalton 

> there are other better approaches to addressing it that don't offload this work onto the caller, and could possibly be done in a non-breaking fashion.

I responded to this idea here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37502#issuecomment-522209897

y21 (approved):

jwalton:
Withdrawing in favor of #37718.

PR #37718, opened 2019-08-18
Express: use generics for params, default to dictionary

OliverJAsh:
In https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37502, I corrected the type of params from `any` to `Dictionary<string> | Array<string>`.

Whilst this change was correct, it [broke a lot of people's code](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37502#issuecomment-522140845), e.g.

``` ts
app.get('/:foo', (req, res) => {
    req.params.foo; // previously `any`, now an error
});
```

This is because the user [was now required to narrow the type of `req.params` from a union to a dictionary or array](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37502#issuecomment-522155668), depending on the path they had provided to Express.

``` ts
app.get('/:foo', (req, res) => {
    if (!Array.isArray(req.params)) {
        req.params.foo; // string
    }
});

app.get('/*', (req, res) => {
    if (Array.isArray(req.params)) {
        req.params.length; // number
        req.params[0]; // string
    }
});
```

We can improve this by using generics to default the params type to a dictionary (which we assume is the most common use case for params), but still allow it to be changed to an array (via an explicit generic).

``` ts
import { ParamsArray } from 'express-serve-static-core';

// Params defaults to dictionary
app.get('/:foo', (req, res) => {
    req.params.foo; // string
    req.params[0]; // string
});

// Params can used as an array
app.get<ParamsArray>('/*', (req, res) => {
    req.params[0]; // string
    req.params.length; // number
});
```

Thanks to @jwalton for the idea: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37502#issuecomment-522315748

Note: I had to bump the TS version of Express and all its dependants because I needed to use default generics in the Express types.

---

Please fill in this template.

### DIFF
--- a/types/api-error-handler/index.d.ts
+++ b/types/api-error-handler/index.d.ts
@@ -2,7 +2,7 @@
 // Project: https://github.com/expressjs/api-error-handler
 // Definitions by: Tanguy Krotoff <https://github.com/tkrotoff>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.2
+// TypeScript Version: 2.3
 
 import * as express from 'express';
 
--- a/types/express-serve-static-core/index.d.ts
+++ b/types/express-serve-static-core/index.d.ts
@@ -7,7 +7,7 @@
 //                 Sami Jaber <https://github.com/samijaber>
 //                 aereal <https://github.com/aereal>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.2
+// TypeScript Version: 2.3
 
 // This extracts the core definitions from express to prevent a circular dependency between express and serve-static
 /// <reference types="node" />
@@ -31,20 +31,25 @@ export interface NextFunction {
     (err?: any): void;
 }
 
-export interface RequestHandler {
+export interface Dictionary<T> { [key: string]: T; }
+export type ParamsDictionary = Dictionary<string>;
+export type ParamsArray = string[];
+export type Params = ParamsDictionary | ParamsArray;
+
+export interface RequestHandler<P extends Params = ParamsDictionary> {
     // tslint:disable-next-line callable-types (This is extended from and can't extend from a type alias in ts<2.2
-    (req: Request, res: Response, next: NextFunction): any;
+    (req: Request<P>, res: Response, next: NextFunction): any;
 }
 
-export type ErrorRequestHandler = (err: any, req: Request, res: Response, next: NextFunction) => any;
+export type ErrorRequestHandler<P extends Params = ParamsDictionary> = (err: any, req: Request<P>, res: Response, next: NextFunction) => any;
 
 export type PathParams = string | RegExp | Array<string | RegExp>;
 
-export type RequestHandlerParams = RequestHandler | ErrorRequestHandler | Array<RequestHandler | ErrorRequestHandler>;
+export type RequestHandlerParams<P extends Params = ParamsDictionary> = RequestHandler<P> | ErrorRequestHandler<P> | Array<RequestHandler<P> | ErrorRequestHandler<P>>;
 
 export interface IRouterMatcher<T> {
-    (path: PathParams, ...handlers: RequestHandler[]): T;
-    (path: PathParams, ...handlers: RequestHandlerParams[]): T;
+    <P extends Params = ParamsDictionary>(path: PathParams, ...handlers: RequestHandler<P>[]): T;
+    <P extends Params = ParamsDictionary>(path: PathParams, ...handlers: RequestHandlerParams<P>[]): T;
     (path: PathParams, subApplication: Application): T;
 }
 
@@ -181,7 +186,7 @@ export interface RequestRanges extends RangeParserRanges { }
 
 export type Errback = (err: Error) => void;
 
-export interface Request extends http.IncomingMessage, Express.Request {
+export interface Request<P extends Params = ParamsDictionary> extends http.IncomingMessage, Express.Request {
     /**
      * Return request header.
      *
@@ -433,7 +438,7 @@ export interface Request extends http.IncomingMessage, Express.Request {
 
     method: string;
 
-    params: any;
+    params: P;
 
     /** Clear cookie `name`. */
     clearCookie(name: string, options?: any): Response;
--- a/types/express-serve-static-core/index.d.ts
+++ b/types/express-serve-static-core/index.d.ts
@@ -181,11 +181,6 @@ export interface RequestRanges extends RangeParserRanges { }
 
 export type Errback = (err: Error) => void;
 
-export interface Dictionary<T> { [key: string]: T; }
-export type ParamsDictionary = Dictionary<string>;
-export type ParamsArray = string[];
-export type Params = ParamsDictionary | ParamsArray;
-
 export interface Request extends http.IncomingMessage, Express.Request {
     /**
      * Return request header.
@@ -438,7 +433,7 @@ export interface Request extends http.IncomingMessage, Express.Request {
 
     method: string;
 
-    params: Params;
+    params: any;
 
     /** Clear cookie `name`. */
     clearCookie(name: string, options?: any): Response;
--- a/types/express-serve-static-core/index.d.ts
+++ b/types/express-serve-static-core/index.d.ts
@@ -49,9 +49,9 @@ export type PathParams = string | RegExp | Array<string | RegExp>;
 export type RequestHandlerParams<P extends Params = ParamsDictionary> = RequestHandler<P> | ErrorRequestHandler<P> | Array<RequestHandler<P> | ErrorRequestHandler<P>>;
 
 export interface IRouterMatcher<T> {
-    // tslint:disable-next-line no-unnecessary-generics
+    // tslint:disable-next-line no-unnecessary-generics (This generic is meant to be passed explicitly.)
     <P extends Params = ParamsDictionary>(path: PathParams, ...handlers: Array<RequestHandler<P>>): T;
-    // tslint:disable-next-line no-unnecessary-generics
+    // tslint:disable-next-line no-unnecessary-generics (This generic is meant to be passed explicitly.)
     <P extends Params = ParamsDictionary>(path: PathParams, ...handlers: Array<RequestHandlerParams<P>>): T;
     (path: PathParams, subApplication: Application): T;
 }
--- a/types/express-serve-static-core/index.d.ts
+++ b/types/express-serve-static-core/index.d.ts
@@ -192,8 +192,10 @@ export type Errback = (err: Error) => void;
 /**
  * @param P  For most requests, this should be `ParamsDictionary`, but if you're
  * using this in a route handler for a route that uses a `RegExp` or a wildcard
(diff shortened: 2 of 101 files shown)


jwalton (approved):
Looks good.  You can even do `Request<{foo: string}>` to get type-safe params!

types/api-error-handler/index.d.ts
```diff
@@ -2,7 +2,7 @@
 // Project: https://github.com/expressjs/api-error-handler
 // Definitions by: Tanguy Krotoff <https://github.com/tkrotoff>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.2
+// TypeScript Version: 2.3
```
OliverJAsh: I had to bump the TS version of Express and all its dependants because I needed to use default generics in the Express types.

daviduzumeri:
Thaaaaaank you for this!!

types/express-serve-static-core/index.d.ts
```diff
@@ -438,7 +441,7 @@ export interface Request extends http.IncomingMessage, Express.Request {
 
     method: string;
 
-    params: Params;
+    params: P;
```
OliverJAsh: This is the main change.

types/express-serve-static-core/index.d.ts
```diff
@@ -31,20 +31,28 @@ export interface NextFunction {
...
 
 export type PathParams = string | RegExp | Array<string | RegExp>;
 
-export type RequestHandlerParams = RequestHandler | ErrorRequestHandler | Array<RequestHandler | ErrorRequestHandler>;
+export type RequestHandlerParams<P extends Params = ParamsDictionary> = RequestHandler<P> | ErrorRequestHandler<P> | Array<RequestHandler<P> | ErrorRequestHandler<P>>;
 
 export interface IRouterMatcher<T> {
-    (path: PathParams, ...handlers: RequestHandler[]): T;
-    (path: PathParams, ...handlers: RequestHandlerParams[]): T;
+    // tslint:disable-next-line no-unnecessary-generics
```
mastermatt: Can a comment be added here as to why this disable is needed please.

mastermatt (changes requested):
I'm a huge fan of this change compared to the other version that went out last week.
But `express-serve-static-core-tests.ts` really needs to be updated as part of this PR to demonstrate the usage.

OliverJAsh:
@mastermatt I added tests to `express`, which seems to be where all the tests have been placed so far, but I've also copied these new tests into `express-serve-static-core`.

mastermatt:
@OliverJAsh Sorry I completely missed the tests added in `express/express-tests.ts`.

Could you add one more test, to either file, demonstrating the use of a known object shape; as @jwalton pointed out.

Something like:
```js
app.get<{foo: string}>('/:foo', (req, res) => {
    req.params.foo; // $ExpectType string
    req.params.bar; // $$ExpectError
});
```

mastermatt (approved):

types/express-serve-static-core/index.d.ts
```diff
@@ -181,12 +189,7 @@ export interface RequestRanges extends RangeParserRanges { }
 
 export type Errback = (err: Error) => void;
 
-export interface Dictionary<T> { [key: string]: T; }
-export type ParamsDictionary = Dictionary<string>;
-export type ParamsArray = string[];
-export type Params = ParamsDictionary | ParamsArray;
-
-export interface Request extends http.IncomingMessage, Express.Request {
+export interface Request<P extends Params = ParamsDictionary> extends http.IncomingMessage, Express.Request {
```
jwalton: Oh, one thing I was going to suggest - when this broke for me, I ctrl-clicked in VSCode to get to the type defition which brought me here.  It might be worthwhile putting a comment in here (or maybe above `interface Request`) giving a brief description of how to use this.  Something along the lines of:

```
/**
 * An Express Request.
 * @param {P} For most requests, this should be `ParamsDictionary`, but if you're
 * using this in a route handler for a route that uses a RegExp or a wildcard path
 * (e.g. "/user/*"), then `req.params` will be an array, in which case you should
 * use `Request<ParamsArray>` instead.
 *
 * @example
 *     app.get(/echo/(.*),
 *         (req: Request<ParamsDictionary>, res: Response) => res.send(req.params[1])
 *     );
 */
```

davidje13: The example snippet has an invalid regex (guessing you intended `/echo\/(.*)/`?) and should use `ParamsArray`, not `ParamsDictionary`

If you add this documentation, I suggest adding another snippet for the more common case, to demonstrate that `ParamsDictionary` does not need to be specified explicitly.

jwalton:
Thanks for the quick turnaround on this, @OliverJAsh.  :)

types/express-serve-static-core/index.d.ts
OliverJAsh: Good idea. That's now done, let me know what you think.

I put the comment on `Request`, although some users might look to `app.get` (and its other methods), so I'm not sure what the best place for this is.

Bene-Graham:
This still break builds due to the fact that it assumes that params will always be strings. This may not be the case when one is using a middleware that validates and sanitizes user inputs.

Something similar like this was tried a couple of years ago #20820, and I believe was reverted due to how big of an impact the change was going to be.

davidje13:
@Bene-Graham since generics are being used here, the only thing stopping it being compatible is the `extends Params` requirement. If that requirement were removed (or made weaker), but still keeping dictionary-of-strings as the default, would that be acceptable? It would mean adding an explicit type to any route which makes use of that feature, but the bonus would be correct types within the function when destructuring (without the need for casts).

Having dictionary-of-strings is quite useful for the majority use-case, since it removes the need to perform explicit casts for every parameter.

OliverJAsh:
I suspect most users don't manipulate/mutate `req.params` to non-`string` types, so in the majority of use cases, these types would be useful/correct. But I agree there could be some flexibility in the generic to allow for more advanced use cases. Or perhaps we should discourage mutation and keep the types as is? 🤷‍♂ 

> This also does not deal with optional params, which would be undefined.

I would consider this a TS bug. See https://github.com/microsoft/TypeScript/issues/13778.

davidje13:
> Or perhaps we should discourage mutation and keep the types as is?

I don't think it's up to the `@types` package to discourage using a feature which is intentionally offered by the main package. If mutation is an available feature, the types should have a way of working with it (though it doesn't need to be *pretty*).

Is restricting the generic parameter necessary? How many real-world mistakes is that restriction likely to catch?

Bene-Graham:
I would personally prefer to be more relaxed.

--------------------

> 
> > This also does not deal with optional params, which would be undefined.
> 
> I would consider this a TS bug. See [microsoft/TypeScript#13778](https://github.com/microsoft/TypeScript/issues/13778).

It is and is not that bug.
It would be that TS bug in this instance
```ts
app.get("/SomeRoute/:userId", async (req: Request, res: Response) => {
        req.params.userID // <--- It would be that TS bug since the property does not exist (Captial D vs lowercase d)
});
```

With the example I used in a previous comment. Express allocates the property on the object, but sets it to undefined.
```ts
app.get("/SomeRoute/:id?", async (req: Request, res: Response) => {
        req.params.id // <--- this would be undefined if one request the URL without the param
});
```

to handled that, on line 36 of the type def you would need to do a union string | undefined
```ts
export type ParamsDictionary = Dictionary<string | undefined>;
```

OliverJAsh:
microsoft/TypeScript#13778 is suggesting that all dictionary lookups return `undefined`—we shouldn't have to add `undefined` to the signature ourselves.

That would be annoying when we know the param is valid according to the route path, but there's no way of statically enforcing that, so it's better to err on the side of caution (always nullable).

OliverJAsh:
We can definitely add `undefined` to the index signature—I was just saying why I (personally) chose not to do that.

Bene-Graham:
> We can definitely add `undefined` to the index signature—I was just saying why I (personally) chose not to do that.

I totally get that. (Really wish TS would fix / add that :smile:)

Bene-Graham:
I wonder if the generic parameters should just default to any. Something similar to what was suggested here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/20820#issuecomment-339724260

This would allow for a couple of options.

1. People who don't want to opt into have types don't have to. 
2. People who more verbose types (like me) could give it an interface of what the type should look like
3. A Dictionary could be passed in(aka. what it is currently doing)

When I tried replacing ```P extends Params = ParamsDictionary``` with ```p = any``` typescript stops inferring the type.

```ts
app.get<{ id: number; }>("/SomeRoute/:id", async (req: Request, res: Response) => {
    req.params.id // <-- params is any, when I would have thought it should be { id: number; }
});
```

davidje13:
@Bene-Graham you should use `unknown`; the behaviour you're seeing is because `any` means "opt out of type checking".

Bene-Graham:
> @Bene-Graham you should use `unknown`; the behaviour you're seeing is because `any` means "opt out of type checking".

That still does not work, the intellisense still think that Request<unknown>

OliverJAsh:
I'm not keen on `any` as the default, because that completely removes type safety. The default should be type safe (i.e. what we currently have), and `any` should be an escape hatch.

I think you need to remove your `req` param annotation—it will be inferred:

``` ts
app.get<{ id: number; }>("/SomeRoute/:id", async (req, res) => {
    req.params.id
});
```

This will only work if you remove the `extends Params` constraint though.

Bene-Graham:
That doesn't seem to work either, request comes out as being any

cookrn:
@OliverJAsh @Bene-Graham I'm stuck on a simple middleware wrapper with a binding function after this upgrade. Previously, I was able to do the following:

```typescript
import { Application, Router } from "express";
import { RequestHandler } from "express-serve-static-core";

type App = Application | Router;

export class Middleware {
  public constructor(private middleware: RequestHandler) {}

  public bind(app: App, path = "/"): App {
    app.use(path, this.middleware);
    return app; // maybe redundant if `use` is fluent anyway and returns App
  }
}
```

Now, `tsc` fails on the call to `use` with an error that looks like the following:

```
  Overload 1 of 3, '(...handlers: RequestHandler<Dictionary<string>>[]): Router', gave the following error.
    Argument of type 'string' is not assignable to parameter of type 'RequestHandler<Dictionary<string>>'.
  Overload 2 of 3, '(...handlers: RequestHandlerParams<Dictionary<string>>[]): Router', gave the following error.
    Argument of type 'string' is not assignable to parameter of type 'RequestHandlerParams<Dictionary<string>>'.
  Overload 3 of 3, '(path: PathParams, subApplication: Application): Router', gave the following error.
    Argument of type 'RequestHandler<Dictionary<string>>' is not assignable to parameter of type 'Application'.
      Type 'RequestHandler<Dictionary<string>>' is missing the following properties from type 'Application': init, defaultConfiguration, engine, set, and 61 more.
```

Do you know how I can satisfy the new type requirements? I'm on TypeScript v3.6.3.

cookrn:
Apologies if I jumped the gun, but this may be due to a TypeScript upgrade in combination with upgrading the type definitions. Would still love your feedback either way!

PR #37724, opened 2019-08-18
Add joi-password-complexity type definitions

nflaig:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/joi-password-complexity/tslint.json
+++ b/types/joi-password-complexity/tslint.json
@@ -1,6 +1,3 @@
 { 
-    "extends": "dtslint/dt.json",
-    "rules": {
-        "no-unnecessary-class": ["allow-constructor-only"]
-    }
+    "extends": "dtslint/dt.json"
 }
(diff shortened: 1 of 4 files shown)


types/joi-password-complexity/tslint.json
```diff
@@ -0,0 +1,6 @@
+{ 
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "no-unnecessary-class": ["allow-constructor-only"]
```
sheetalkamat: Remove the rules here and fix errors if any.. We want new packages to extend the default rules and fix all the errors

sheetalkamat (changes requested):

types/joi-password-complexity/tslint.json
nflaig: @sheetalkamat Thanks for your feedback. I tried to remove that rule but how do I define a type for a class that only has a constructor then? It says I should use a function instead but I am not sure what is meant by that.

sheetalkamat: You can add `// tslint:disable-next-line:no-unnecessary-class` to the line before class declaration instead of disabling it for the whole package.

PR #37726, opened 2019-08-18
koji-tools types

bdjeffyp:
If adding a new definition:

### DIFF
--- a/types/koji-tools/index.d.ts
+++ b/types/koji-tools/index.d.ts
@@ -10,14 +10,43 @@ export as namespace Koji;
 
 declare namespace Koji {
     // index.js methods
+
+    /**
+     * Server-side function that sets file watchers on all .koji customization
+     * files and allows for hot reloading of these properties.
+     */
+    function watch(): void;
+
     /**
      * Sets up Koji.config parameters for each client and handles communication
      * between the Koji live preview iframe and your app.
-     * @param options ???
      */
-    function pageLoad(options?: {}): void;
+    function pageLoad(options?: object): void;
 
-    function getConfig(): any;
+    /**
+     * Wrapper for fetch that takes objects from Koji.routes.
+     */
+    function request(route: object, params?: object): Promise<any>;
+
+    /**
+     * After the 'pwaPromptReady' event has fired, this function will make a popup
+     * installation prompt appear.
+     */
+    function pwaPrompt(): any;
+
+    /**
+     * An auto-generated list of all of the Koji Customization Controls (VCCs) your
+     * application has setup. When Koji.watch() is being used this list updates automatically.
+     */
+    const config: object;
+
+    /**
+     * A auto-generated list of routes based on koji.json files in your project that are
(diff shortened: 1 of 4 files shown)


types/koji-tools/index.d.ts
```diff
@@ -0,0 +1,30 @@
...
+export as namespace Koji;
+
+declare namespace Koji {
+    // index.js methods
+    /**
+     * Sets up Koji.config parameters for each client and handles communication
+     * between the Koji live preview iframe and your app.
+     * @param options ???
+     */
+    function pageLoad(options?: {}): void;
```
sheetalkamat: Missing few properties eg. from [usage](https://github.com/madewithkoji/koji-tools#usage---frontend) `Koji.config.colors.backgroundColor` and `Koji.config.strings.content` is suppose to exist?

types/koji-tools/index.d.ts
```diff
@@ -0,0 +1,30 @@
...
+    function getConfig(): any;
+
+    /**
+     * Registers a callback on a Koji event.
+     * @param event name of the event being called
+     * @param callback method to execute when the event occurs
+     */
+    function on(event: string, callback: () => void): void;
+
+    function callEvent(event: string, params?: []): void;
```
sheetalkamat: Missing more properties as per https://github.com/madewithkoji/koji-tools#api

sheetalkamat (changes requested):

types/koji-tools/index.d.ts
bdjeffyp: The `Koji.config` object is created from various JSON files into one JSON string, as specified in `./tools/buildConfig.js`. This makes any subelements of `Koji.config` and their types unknown. In the example you cited, the config object was built with two JSON files, named `colors.json` and `strings.json`. `buildConfig.js` bundles those JSON objects into the `Koji.config` object.

bdjeffyp: This is why the `getConfig()` method returns an `any` type.

bdjeffyp: I have changed the definition file to use `config` and specified it as an `object` type.

types/koji-tools/index.d.ts
```diff
@@ -0,0 +1,63 @@
+// Type definitions for koji-tools 0.5
+// Project: https://github.com/madewithkoji/koji-tools
+// Definitions by: Jeff Peterson <https://github.com/bdjeffyp>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+// tslint:disable-next-line:export-just-namespace
+export = Koji;
```
sheetalkamat: Do not disable this. Remove namespace declaration and export the declarations inside it directly.

sheetalkamat (changes requested):

types/koji-tools/index.d.ts
bdjeffyp: Done!

PR #37731, opened 2019-08-19
Add typings for tti-polyfill

derduher:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/tti-polyfill/index.d.ts
+++ b/types/tti-polyfill/index.d.ts
@@ -1,0 +1,40 @@
+// Type definitions for tti-polyfill 0.2
+// Project: https://github.com/GoogleChrome/tti-polyfill#readme
+// Definitions by: Patrick Weygand <https://github.com/derduher>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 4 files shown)


types/tti-polyfill/index.d.ts
```diff
@@ -0,0 +1,40 @@
...
+     * DOM elements will create additional network requests. This can be
+     * disabled to improve performance in cases where you know no additional
+     * request-creating DOM elements will be added.
+     */
+    useMutationObserver?: boolean;
+    /**
+     * The lower bound to start forward-searching for the quite window. If no value is set, the default is after the
+     * DOMContentLoaded event.
+     */
+    minValue: number;
```
sheetalkamat: This should be `minValue?: number | null;` as per https://github.com/GoogleChromeLabs/tti-polyfill#configuration-options

sheetalkamat (changes requested):

types/tti-polyfill/index.d.ts
derduher: Good catch! patched.

PR #37732, opened 2019-08-19
[@types/ramda] Update definition for `flatten` function

reeli:
When I use `R.flatten` function, it shows the error as follows:

image

The function type is a mismatch. So I checked the type definition:

```ts
flatten<T>(x: ReadonlyArray<T[]> | ReadonlyArray<ReadonlyArray<T>> | ReadonlyArray<T>): T[];
```

I think we should use type reload to define this function:

```ts
flatten<T>(x: ReadonlyArray<T[]>): T[];
flatten<T>(x:  ReadonlyArray<ReadonlyArray<T>>): T[];
flatten<T>(x: ReadonlyArray<T>): T[];
```

### DIFF
--- a/types/ramda/index.d.ts
+++ b/types/ramda/index.d.ts
@@ -1408,7 +1408,9 @@ declare namespace R {
          * Returns a new list by pulling every item out of it (and all its sub-arrays) and putting
          * them in a new array, depth-first.
          */
-        flatten<T>(x: ReadonlyArray<T[]> | ReadonlyArray<ReadonlyArray<T>> | ReadonlyArray<T>): T[];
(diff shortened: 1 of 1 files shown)


Nemo108:
@reeli what kind of issue does this PR tries to fix?

reeli:
@Nemo108 

For example:

```ts
interface IData {
  label: string;
  value: number[];
}

export const demo = () =>
  R.compose<IData[], number[][], number[]>(
    R.flatten,
    R.map((item: IData) => item.value),
  );
```

It seems that `<T>(x: readonly T[] | readonly T[][] | readonly (readonly T[])[]) => T[]` was considered as a type argument, the error shows as follows:

image

PR #37741, opened 2019-08-19
[@hapi/vision] add .getViewsManager()

lenovouser:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [hapijs/vision/API.md](https://github.com/hapijs/vision/blob/master/API.md)

### DIFF
--- a/types/hapi__vision/hapi__vision-tests.ts
+++ b/types/hapi__vision/hapi__vision-tests.ts
@@ -25,6 +25,10 @@ const provision = async () => {
         path: __dirname + '/templates',
     });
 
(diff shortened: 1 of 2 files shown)


AJamesPhillips:
@sheetalkamat were these changes in a recent version of vision or just missing from 5.5 definitions?  If so we should bump the version number instead of publishing under a minor patch.  Preferably also reverting the patch too if that was the case.
Thanks for the contribution!

lenovouser:
@AJamesPhillips as far as I can see they were just missing from the 5.5 definitions

PR #37748, opened 2019-08-19
Add typings for Figma Plugin API

rudi-c:
This PR adds typings for the the Figma Plugin API that was released last week (I work at Figma).

Note that this is not a typings file for a traditional NPM package with source code on Github. These are the typings file for an API exposed to plugin code when they run inside a special sandbox inside the Figma design tool, on our website.

Note also that there already exists an NPM package with the name `figma`. That package is currently empty and owned by the CTO of Figma @evanw, so that's not a problem even though the two packages are unrelated.

If adding a new definition:

### DIFF
--- a/types/figma/tslint.json
+++ b/types/figma/tslint.json
@@ -1,7 +1,6 @@
 {
     "extends": "dtslint/dt.json" ,
     "rules": {
-        "dt-header": false,
         "npm-naming": false
     }
 }
--- a/types/figma/tslint.json
+++ b/types/figma/tslint.json
@@ -1,0 +1,10 @@
+{
+    "extends": "dtslint/dt.json" ,
+    "rules": {
+        "dt-header": false,
+        "npm-naming": false,
+        "no-any-union": false,
+        "interface-over-type-literal": false,
+        "semicolon": false
+    }
+}
--- a/types/figma/tslint.json
+++ b/types/figma/tslint.json
@@ -2,9 +2,6 @@
     "extends": "dtslint/dt.json" ,
     "rules": {
         "dt-header": false,
-        "npm-naming": false,
-        "no-any-union": false,
-        "interface-over-type-literal": false,
-        "semicolon": false
+        "npm-naming": false
     }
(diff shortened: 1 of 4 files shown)


types/figma/tslint.json
```diff
@@ -0,0 +1,7 @@
+{
+    "extends": "dtslint/dt.json" ,
+    "rules": {
```
sheetalkamat: Please remove all these rules here.. And fix the errors. New packages cannot override the default rules and need to fix all the errors.
Looking at the figma package it only contains package.json... So i am not sure this is correct package name and types for the npm package.

sheetalkamat (changes requested):

types/figma/tslint.json
rudi-c: Thanks for the feedback! Please see the description in the PR, this isn't a typings for a normal NPM package.

I removed the extra lint rules as per review in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37382, but accidentally let the PR close so I had to open a new one.

sheetalkamat: If thats the case you need to change the name of the package so that it doesnt conflict with existing package and fix the header accordingly. Refer to https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package

rudi-c: Even if we create a unique name for this typings, I'm not aware of any mechanism on the NPM side to prevent someone to create a package that conflicts with a DefinitelyTyped typings?

`@types/figma` the most natural name for developers writing Figma plugins. In that sense, the empty `figma` package on NPM is perfect for security purposes. In case someone gets confused and thinks they need to `npm install figma` to make plugins work (they don't), nothing bad will happen.

sheetalkamat: You can keep the `npm-naming` rule to false but please remove `dt-header` disabling from here. Also in the header please add github ids instead of emails to make sure our bot can ping you when things change in the package. Thanks.

PR #37751, opened 2019-08-19
Node: specify `UnhandledRejectionListener` `reason` param type

OliverJAsh:
Same as https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33195, but updates types for v8/v9/10.

---

Please fill in this template.

### DIFF
--- a/types/node/v10/globals.d.ts
+++ b/types/node/v10/globals.d.ts
@@ -656,7 +656,7 @@ declare namespace NodeJS {
     type ExitListener = (code: number) => void;
     type RejectionHandledListener = (promise: Promise<any>) => void;
     type UncaughtExceptionListener = (error: Error) => void;
-    type UnhandledRejectionListener = (reason: any, promise: Promise<any>) => void;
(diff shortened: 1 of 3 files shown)


islishude (approved):

ostkrok:
Is this really correct?

https://nodejs.org/docs/latest-v10.x/api/process.html#process_event_unhandledrejection

OliverJAsh:
Yes, a promise can reject with any type.

ostkrok:
> Yes, a promise can reject with any type.

So shouldn't the old typing (`any`) be correct then?

This PR changes the type to be: an empty object, null or undefined.

OliverJAsh:
`any` disables type checking, e.g.

``` ts
declare const foo: any;
any.whatever.goes.here // runtime exception!
```

Thus, `unknown` should be preferred.

`unknown` shipped in a recent version of TS, but these typings don't yet support that version. The type used here is like a "poor man's `unknown`"—it does the same thing.

ostkrok:
Ahh, that makes sense.

Thanks for explaining!

PR #37754, opened 2019-08-19
Node: Adds pnp to the process.versions field

arcanis:
Context: The PnP hook generated by Yarn injects a `pnp` variable into the `process.versions` object. This field is used by JS packages to branch their behavior if needed. I haven't documented it yet on our website, but a few packages already use it.

This isn't a Node feature per-se, but it's a very small change so I figured why not 🙂 An alternative would be to type the field with an generic index: `[key: string]: string`.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/yarnpkg/berry/blob/master/packages/berry-pnp/sources/loader/applyPatch.ts#L84

### DIFF
--- a/types/node/globals.d.ts
+++ b/types/node/globals.d.ts
@@ -696,6 +696,7 @@ declare namespace NodeJS {
         zlib: string;
         modules: string;
         openssl: string;
+        pnp?: string;
     }
 
     type Platform = 'aix'
--- a/types/node/v10/globals.d.ts
+++ b/types/node/v10/globals.d.ts
@@ -631,6 +631,7 @@ declare namespace NodeJS {
         zlib: string;
         modules: string;
         openssl: string;
+        pnp?: string;
     }
 
     type Platform = 'aix'
--- a/types/node/v11/globals.d.ts
+++ b/types/node/v11/globals.d.ts
@@ -664,6 +664,7 @@ declare namespace NodeJS {
         zlib: string;
         modules: string;
         openssl: string;
+        pnp?: string;
     }
 
     type Platform = 'aix'
--- a/types/node/v8/base.d.ts
+++ b/types/node/v8/base.d.ts
@@ -469,6 +469,7 @@ declare namespace NodeJS {
         zlib: string;
         modules: string;
         openssl: string;
+        pnp?: string;
     }
 
     type Platform = 'aix'
--- a/types/node/v9/base.d.ts
+++ b/types/node/v9/base.d.ts
@@ -606,6 +606,7 @@ declare namespace NodeJS {
         zlib: string;
         modules: string;
         openssl: string;
+        pnp?: string;
     }
 
     type Platform = 'aix'


arcanis:
I think the failed build is an infra fail unrelated to the code change. Will try repushing tomorrow 🤔

islishude (changes requested):
I don't think non-standard functionality should be added.

SimonSchick:
I concur with @islishude

SimonSchick (changes requested):
Non standard/node properties don't belong in these typings, you can augment them via.
```
declare namespace NodeJS {
  interface ProcessVersions {
   ...
  }
}```
in your typings.

**Comment by arcanis** [comment]
Sounds reasonable. What do you think of the indexing (on top of the explicit listing of the common fields)?:

```ts
interface ProcessVersions {
  node: string;
  v8: string;
  //... 
  [key: string]: string;
}
```

The Node documentation defines the field with:

> Will generate an object *similar* to:

Meaning that for all purposes different fields than the ones described there may appear.

SimonSchick:
The problem with adding an index signature is that it effectively breaks property safety as any misspelled set/get, are there any other use cases other than `pnp`?

arcanis:
Not from third-parties afaik, but in general the list is likely missing some. For example in the Node doc you can see references to `brotli`, `llhttp`, `tz`, ...

> The problem with adding an index signature is that it effectively breaks property safety as any misspelled set/get

Not entirely as only the explicit fields would have the `string` type, right? Others would be `string | undefined`.

Flarna:
> Not from third-parties afaik, but in general the list is likely missing some. For example in the Node doc you can see references to `brotli`, `llhttp`, `tz`, ...

If they are missing we should add them - but not others.

Flarna (changes requested):
JavaScript allows to extend/change any object therefore all definitions done here would need an index signature... and `any` as type.... but I don't think that this would be a step forward.
Not sure if it is possible to augment the node type definitions via another definition but this would be the correct approach here.

arcanis:
> If they are missing we should add them - but not others.

The `versions` field depends on Node's dependencies - which may very well be different from a platform to another, as some features may be supported by different libraries. I don't really see why the type definitions should make explicit what those dependencies are, as you don't control them (for all you know I could use a Node that uses `libressl` instead of `openssl`).

Flarna:
I think the definitions here follow the node binaries you get as downloads from https://nodejs.org/ (and the binaries you get if you build it with the same settings).
I agree that you can compile node without openssl or without support for internationalization and some keys disappear (not sure about `libressl` and if this adds itself in key `openssl` or not at all) but I would guess these setups are really exotic. Most likely such customizations of node would have side effects to other APIs also which would be again hard to cover here.

arcanis:
And that's a good case for explicitly listing some fields, but not to prevent the index - the Node documentation itself defines the field as `<Object>`:

> image


It doesn't list the exact fields that should be expected ... so to perfectly match Node (which is the rational originally given regarding why `pnp` wouldn't work) its typing should be a pure open index.

It's obviously not ideal from a DevX perspective so I perfectly understand why you would want to list *some* of them explicitly, but at the very least it should be open to the possible extras without requiring them to be formally defined in a standalone `@types/` package.

This sounds very similar to me to the `navigator.plugins` variable, which is typed to access any possible plugin that the browser might be using, not only the ones used by the default builds.

PR #37756, opened 2019-08-20
add stringify to yarnpkg__lockfile

fa93hws:
Please fill in this template.

- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`

### DIFF
--- a/types/yarnpkg__lockfile/index.d.ts
+++ b/types/yarnpkg__lockfile/index.d.ts
@@ -3,22 +3,19 @@
 // Definitions by: Eric Wang <https://github.com/fa93hws>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export = parse;
-
-declare namespace parse {
-  interface Dependency {
-    [packageName: string]: string;
-  }
-
-  interface FirstLevelDependency {
-    version: string;
-    resolved?: string;
-    dependencies?: Dependency;
-  }
-
-  interface LockFileObject {
-    [packageName: string]: FirstLevelDependency;
-  }
+export interface Dependency {
+  [packageName: string]: string;
 }
 
-declare function parse(file: string): { object: parse.LockFileObject };
+export interface FirstLevelDependency {
+  version: string;
+  resolved?: string;
+  dependencies?: Dependency;
+}
+
+export interface LockFileObject {
+  [packageName: string]: FirstLevelDependency;
+}
+
+export function parse(file: string): { object: LockFileObject };
+export function stringify(json: { object: LockFileObject }): string;
--- a/types/yarnpkg__lockfile/index.d.ts
+++ b/types/yarnpkg__lockfile/index.d.ts
@@ -17,7 +17,17 @@ export interface LockFileObject {
   [packageName: string]: FirstLevelDependency | {};
 }
 
-export type ParseResult = { type: 'success' | 'merge'; object: LockFileObject } | { type: 'conflict'; object: {} };
+interface SuccessLockFileResult {
+  type: 'success' | 'merge';
+  object: LockFileObject;
+}
+
+interface ConflictLockFileResult {
+  type: 'conflict';
+  object: {};
+}
+
+export type ParseResult = SuccessLockFileResult | ConflictLockFileResult;
 
 export function parse(file: string, fileLoc?: string): ParseResult;
 export function stringify(
--- a/types/yarnpkg__lockfile/index.d.ts
+++ b/types/yarnpkg__lockfile/index.d.ts
@@ -17,9 +17,7 @@ export interface LockFileObject {
   [packageName: string]: FirstLevelDependency | {};
 }
 
-export type ParseResult =
-  { type: 'success' | 'merge';object: LockFileObject; }
-  | { type: 'conflict'; object: {}; };
+export type ParseResult = { type: 'success' | 'merge'; object: LockFileObject } | { type: 'conflict'; object: {} };
 
 export function parse(file: string, fileLoc?: string): ParseResult;
 export function stringify(
--- a/types/yarnpkg__lockfile/index.d.ts
+++ b/types/yarnpkg__lockfile/index.d.ts
@@ -17,12 +17,12 @@ export interface LockFileObject {
   [packageName: string]: FirstLevelDependency | {};
 }
 
-interface SuccessLockFileResult {
+export interface SuccessLockFileResult {
   type: 'success' | 'merge';
   object: LockFileObject;
 }
 
-interface ConflictLockFileResult {
+export interface ConflictLockFileResult {
   type: 'conflict';
   object: {};
 }
--- a/types/yarnpkg__lockfile/index.d.ts
+++ b/types/yarnpkg__lockfile/index.d.ts
@@ -18,14 +18,8 @@ export interface LockFileObject {
 }
 
 export type ParseResult =
-  {
-      type: 'success' | 'merge';
-      object: LockFileObject;
-    }
-  | {
-      type: 'conflict';
-      object: {};
-    };
+  { type: 'success' | 'merge';object: LockFileObject; }
+  | { type: 'conflict'; object: {}; };
 
 export function parse(file: string, fileLoc?: string): ParseResult;
 export function stringify(
--- a/types/yarnpkg__lockfile/index.d.ts
+++ b/types/yarnpkg__lockfile/index.d.ts
@@ -17,13 +17,19 @@ export interface LockFileObject {
   [packageName: string]: FirstLevelDependency | {};
 }
 
-export type ParseResult = {
-  type: 'success' | 'merge';
-  object: LockFileObject;
-} | {
-  type: 'conflict';
-  object: {};
-};
+export type ParseResult =
+  | {
+      type: 'success' | 'merge';
+      object: LockFileObject;
+    }
+  | {
+      type: 'conflict';
+      object: {};
+    };
 
 export function parse(file: string, fileLoc?: string): ParseResult;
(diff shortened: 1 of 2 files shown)


types/yarnpkg__lockfile/yarnpkg__lockfile-tests.ts
```diff
@@ -1,10 +1,13 @@
-import parseYarnLock = require('@yarnpkg/lockfile');
+import { parse, stringify, FirstLevelDependency } from '@yarnpkg/lockfile';
 
-function testFirstLevelDependency(obj: parseYarnLock.FirstLevelDependency) {}
+function testFirstLevelDependency(obj: FirstLevelDependency) {}
 
-const { object } = parseYarnLock('');
+const file = '';
```
fa93hws: copied from https://github.com/yarnpkg/yarn/tree/master/packages/lockfile

types/yarnpkg__lockfile/index.d.ts
```diff
@@ -3,22 +3,19 @@
 // Definitions by: Eric Wang <https://github.com/fa93hws>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export = parse;
```
sheetalkamat: Fix the package name in header which is wrong

sheetalkamat (changes requested):

Seally:
Since I've been having issues with the type definitions for this package (and considering all the pull requests, it's in a state of flux at the moment), I'll leave some comments.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/503ff283b769cbbae01a3f591984d03f7e73e123/types/yarnpkg__lockfile/index.d.ts#L21
The type of the `json` argument for this should be just `LockFileObject` or even `object`. The definition as it stands will leave the line `type success` hanging in the stringified output and add additional indentation, which means the output probably won't be a valid Yarn lockfile (doesn't help that the docs for the package also seem to be misleading). Note that if the `typeof obj !== "object"`, the function will throw a `TypeError`. 

An additional problem with the current typings is that the parser is, well, a parser, not a parser+validator, so even `lockfile.parse("blahblahblah:")` will be successfully parsed. This means that there's no guarantee that the value in the `object` key of the parse results will be what has been specified here. 

A more accurate return type for `parse()` might be `{ type: "success" | "merge", object: unknown } | { type: "conflict", object: {} }`. The `unknown` type is to force type casting to something like `LockFileObject`, but it could also be `any` (like what `JSON.parse()` returns). Either way, it might be best to let the programmer know they're taking a leap of faith with the return value and not assume its return value in the type definitions.

Side note: the non-primitive objects under the `object` key will usually have null prototypes, but I'm not sure what we can do about that besides maybe leaving a warning in the comments. 

The function signatures for `parse()` and `stringify()` are also currently incomplete. Here are the full function signatures from the source (typed with Flow):

https://github.com/yarnpkg/yarn/blob/1a57117853afa2901feb541f6db7f064d9c0f8b9/src/lockfile/parse.js#L416

https://github.com/yarnpkg/yarn/blob/1a57117853afa2901feb541f6db7f064d9c0f8b9/src/lockfile/stringify.js#L100

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/503ff283b769cbbae01a3f591984d03f7e73e123/types/yarnpkg__lockfile/index.d.ts#L10-L14

The `optionalDependencies` and `integrity` fields seem to be missing (at the very least, there are more fields but I haven't seen any real examples of them in a lockfile). I personally called this interface `LockManifest` to match the existing Flow typings (it's not _exactly_ for this object, but it's close enough), but I guess `FirstLevelDependency` works too.

Note that there are also exports for `implodeEntry()`, `explodeEntry()`, and the default class export `Lockfile`, though I don't know if they're useful outside of Yarn itself.

fa93hws:
Many thanks for your suggestion @salley.

Regarding the `parse`, it the return type is taged as "success", do you think the `object` type can be `LockFileObject` instead of `unknown`?

Seally:
@fa93hws 

Well, here's what happens in Node REPL:

```
> const lockfile = require("@yarnpkg/lockfile");
> lockfile.parse("blahblahblah:")
{
  type: 'success',
  object: [Object: null prototype] {
    blahblahblah: [Object: null prototype] {}
  }
}
```

So I'm going with "no", you can't guarantee the return type will be of that type, unless the input is _definitely_ a valid Yarn lockfile (which static typing can't guarantee). 

_Note:_ I haven't added the `Lockfile`, `implodeEntry()`, and `explodeEntry()` to my internal type definitions either, since I have the impression they're made for Yarn internals and just happened to included because they were originally from the same file. I think you can skip them. If someone needs them they're welcome to make a PR.

fa93hws:
Hi, if the string is invalid, the `parse` will throw an error.
if the string is semi valid (`parse` won't throw but the we know it's wrong, such as "blahblahblah:"), the result will be
```
{
  blahblahblah: {}
}
```
which is the same shape if we do
```
interface LockFileObject {
  [packageName: string]: FirstLevelDependency | {};
}
```

Seally:
Hmm, then it's fine I guess. I'm still trying to figure out how to prevent something like `lockFileObject.nonexistentPkg.version` from compiling. If it's possible.

fa93hws:
`FirstLevelDependency | {}` can be regarded as
```
export interface FirstLevelDependency {
  version?: string;
  resolved?: string;
  dependencies?: Dependency;
}
```

Seally:
https://github.com/yarnpkg/yarn/blob/1a57117853afa2901feb541f6db7f064d9c0f8b9/src/lockfile/index.js#L42-L65
This is the Flow typings for the `LockfileManifest` from the source. In a proper Yarn lockfile, the version field should always be present and be a _string_ (makes sense since its primary purpose is to keep the installed package versions the same on all installs, something it can't do if the lockfile doesn't have a `version` field). 

In a syntactically correct lockfile, which has similar flexibility to a JSON file, the version could be a string, a number, a boolean, an object, or completely non-existent. The current type definitions promise that it's a string, which might not be always be the case. This is my reasoning for keeping the type as `any` or `unknown` and letting the user decide to cast for better IDE support.

This input will produce an object whose `version` field is a number: `"foobar@^1.0.0:\n  version 15"`.

It's unlikely it will be a big issue either way, considering the most likely use case for the package, but yeah, I think it's slightly inaccurate.

fa93hws:
Personally I feel that if I mark the type as `any` (`unknown` is not available as it need to support low version typescript), it will be less safe and harder to use.
Even if `unknown` is available, it's very hard for the developer to consume if the return type is `unknown` unless there is a validation method that return a correct type.

Regarding the version type, I think the library should do something like `obj.version = v.toString()` to make sure the return type is actually a string.

sheetalkamat:
waiting for discussion to be resolved before merging

Seally:
> Personally I feel that if I mark the type as `any` (`unknown` is not available as it need to support low version typescript), it will be less safe and harder to use.

I don't think it will make it less safe, since the type definition as you proposed would be effectively lying to the library consumer that it's safe. In fact, I'd say it makes it safer because at the very least you know what you're doing is inherently unsafe. The library's `parse()` is as type-safe as calling `JSON.parse(packageJson)` when the only thing the code knows about `packageJson` is that it's read from a file named `package.json`. After all, I can just name any arbitrary JSON to `package.json` to feed the code and it will blow up (putting this into consideration is important for software security, but I digress). 

Yes, doing this does make it harder to use, but I don't see an issue with doing `lockfile.parse(...).object as lockfile.LockFileObject` as long as the developer knows what they're doing. This is beyond what a type definition file can cover. In fact, in the strictest sense we don't have to provide `LockFileObject`. The fact that it's there at all is for convenience, as the strictest type the value of the `object` key can be is, well, `object` (okay, you can make it slightly more restrictive, but it doesn't help much). We know it's definitely not a `string`, `number`, `boolean`, `symbol`, `null`, or `undefined`. But `object` is only a few types away from `any` (to be clear, I now think typing that as `object` is best, and that keyword seems to have been introduced in TS 2.2, so it's fairly old by now). 

`lockfile.parse()` is equivalent to `JSON.parse()` for the Yarn lockfile _format_, not Yarn lockfiles. From inspecting the Yarn codebase, they actually use the exact same function to parse `yarnrc` files (which might explain why they do conversions on positive numbers and booleans, which isn't useful at all for lockfiles).

> Regarding the version type, I think the library should do something like `obj.version = v.toString()` to make sure the return type is actually a string.

Just so you know, I currently have no involvement whatsoever with the development of Yarn. I'm just reading the code as I see it (with some help from Webstorm go-to declaration and quick docs feature, along with some experience in writing parsers). 

_Pedantic note:_ the input is a string, so if it returns something other than a string it means it does type conversion, not the other way around. If it doesn't care, it'll just be sliced from the input string, which will be... (drumroll) a string. 

To end this message, here are some of the lines that are causing us all this trouble (by doing type conversion):

https://github.com/yarnpkg/yarn/blob/1a57117853afa2901feb541f6db7f064d9c0f8b9/src/lockfile/parse.js#L121-L131

And here's the actual type definitions for our `ParseResult` (in Flow, but it's almost like TypeScript with a different syntax anyway):

https://github.com/yarnpkg/yarn/blob/1a57117853afa2901feb541f6db7f064d9c0f8b9/src/lockfile/parse.js#L21-L26

[`Object`](https://flow.org/en/docs/types/objects/#toc-object-type) in Flow is deprecated and in later versions it's aliased to `any`, which seems to be like TS `any`. But the Yarn project's own lockfile and `.flowconfig` reveals that they're using an old version of Flow, which treats `Object` like `{ [key: string]: any }`... which isn't all that different.

fa93hws:
OK, I will then make it any then.

PR #37778, opened 2019-08-20
TabulatorTables: allow columnDefinition and removed "required" from title (in ColumnLayout)

simusr2:
Allow columnDefinition in ColumnLayout: this allows to have "grouped" columns (Example: measurement group contains L W H).
Removed "required" from ColumnLayout.field: as comment says this is not required in icon/button columns, or group column (first edit in this commit). Comments says not required but field has missing '?'.

### DIFF
--- a/types/tabulator-tables/index.d.ts
+++ b/types/tabulator-tables/index.d.ts
@@ -831,10 +831,11 @@ declare namespace Tabulator {
     }
 
     interface ColumnLayout {
+        columns?: ColumnDefinition[];
         /** title - Required This is the title that will be displayed in the header for this column */
         title: string;
         /** field - Required (not required in icon/button columns) this is the key for this column in the data array*/
-        field: string;
+        field?: string;
         /** visible - (boolean, default - true) determines if the column is visible. (see Column Visibility for more details */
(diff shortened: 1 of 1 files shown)


Jojoshua (changes requested):
- Making field optional is fine

- Adding columns? to ColumnLayout needs more thought. If you are trying to add support for Column Grouping as documented here http://tabulator.info/docs/4.3/columns#groups, then this PR doesn't seem to fix that. Please include a working test in tabulator-tables-tests.ts which includes the code snippet similar to what the tabulator site has for column grouping.

simusr2:
Ok, i will prepare a test. However from link you attached i found this example:

```
{//create column group
        title:"Work Info",
        columns:[
        {title:"Progress", field:"progress", align:"right", sorter:"number", width:100},
        {title:"Rating", field:"rating", align:"center", width:80},
        {title:"Driver", field:"car", align:"center", width:80},
    ],
},
```

In ColumnLayout (where title:"Work Info") there is a ColumnDefinition[]. We added columns? according to this example.

When i'm ready I commit test you requested.
Thanks!

PR #37781, opened 2019-08-20
add declaration for serve-handler

simonhaenisch:
Please fill in this template.

If adding a new definition:
- [x] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration` (I tried to add it but it did not get accepted; see https://github.com/zeit/serve-handler/pull/91)

I had to add `/// <reference types="node" />` to the declaration file, otherwise it would not be able to find the node packages (`fs` and `http`) that I need to reference types from. I also tried adding `"types": ["node"]` in the `tsconfig.json` but that didn't work. However I've looked at other packages and they seem to be able to import node packages without a triple slash directive, and with the same tsconfig, so I'm not sure what I'm missing here.

### DIFF
--- a/types/serve-handler/index.d.ts
+++ b/types/serve-handler/index.d.ts
@@ -1,0 +1,217 @@
+// Type definitions for serve-handler 6.1
+// Project: https://github.com/zeit/serve-handler#readme
+// Definitions by: simonhaenisch <https://github.com/me>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/// <reference types="node" />
+
+import { lstat, realpath, createReadStream, readdir } from 'fs';
+import { IncomingMessage, ServerResponse } from 'http';
+
+export default serveHandler;
+
+/**
+ * This package represents the core of `serve`. It can be plugged into any HTTP
+ * server and is responsible for routing requests and handling responses.
+ *
+ * In order to customize the default behaviour, you can also pass custom routing
+ * rules, provide your own methods for interacting with the file system and much
+ * more.
+ */
+declare function serveHandler(
+    request?: IncomingMessage,
+    response?: ServerResponse,
+    /**
+     * If you want to customize the package's default behaviour, you can use
+     * this argument to pass any of the configuration options provided.
+     *
+     * @see https://github.com/zeit/serve-handler#options
+     */
+    config?: Config,
+    /**
+     * If you want to replace the methods the package is using for interacting
+     * with the file system and sending responses, you can pass them as the
+     * fourth argument to the function call. These are the methods used by the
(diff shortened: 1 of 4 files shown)


types/serve-handler/index.d.ts
```diff
@@ -0,0 +1,217 @@
...
+// Project: https://github.com/zeit/serve-handler#readme
+// Definitions by: simonhaenisch <https://github.com/simonhaenisch>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/// <reference types="node" />
+
+import { lstat, realpath, createReadStream, readdir } from 'fs';
+import { IncomingMessage, ServerResponse } from 'http';
+
+export default serveHandler;
```
sheetalkamat: This needs `export =` instead of `export default`. Refer to https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default

sheetalkamat (changes requested):

types/serve-handler/index.d.ts
simonhaenisch: I read that and had that first, however it seemed like it didn't apply in this case. The package has

```js
module.exports = async (/* ... */) => { /* ... */ }
```

So would `import * as serveHandler from 'serve-handler'` really be the correct way of importing this module (without synthetic default imports)?

simonhaenisch: If I do `export =`, and then in my test file (`serve-handler/serve-handler-tests.ts`) I do

```ts
import serveHandler from 'serve-handler'
```

I get

> This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.

If I do 

```ts
import * as serveHandler from 'serve-handler'
```

I get

> This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.

From how I understood it, `esModuleInterop` is supposed to be on by default, is it not?

sheetalkamat: `module.exports =` gets converted to `export = ` syntax in typescript.  

Running code
```ts
const x = require("serve-handler");
console.log(x.default);
```
```
> node a.js
undefined
```

simonhaenisch: Oh my bad, I see now that I'm supposed to import it as

```ts
import serveHandler = require('serve-handler');
```

And it works now to set `allowSyntheticDefaultImports` in `types/serve-handler/tsconfig.json`, and use `import serveHandler from 'serve-handler'`... for some reason that wasn't working before, maybe my VS Code was just hanging 🤷🏻‍♂️.

PR #37786, opened 2019-08-20
[chart.js] Add data-parameter to tooltipOptions itemSort

mmmmmarco:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.chartjs.org/docs/latest/configuration/tooltip.html#sort-callback and https://github.com/chartjs/Chart.js/blob/2954cf8a9928b863fa59c2a8bdc7701c328a027a/src/core/core.tooltip.js#L623

### DIFF
--- a/types/chart.js/chart.js-tests.ts
+++ b/types/chart.js/chart.js-tests.ts
@@ -40,7 +40,7 @@ const chart: Chart = new Chart(ctx, {
             filter: data => Number(data.yLabel) > 0,
             intersect: true,
             mode: 'index',
-            itemSort: (a, b) => Math.random() - 0.5,
+            itemSort: (a, b, data) => Math.random() - 0.5,
             position: 'average',
             caretPadding: 2,
             displayColors: true,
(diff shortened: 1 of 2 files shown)


adripanico (changes requested):
According to the documentation "This function **can** also accept a third parameter that is the data object passed to the chart", so, ```data``` should be optional.

mmmmmarco:
In the source code (https://github.com/chartjs/Chart.js/blob/2954cf8a9928b863fa59c2a8bdc7701c328a027a/src/core/core.tooltip.js#L623), the function itemSort is always called with three parameters. Looks like the docs are incorrect?

adripanico:
> In the source code (https://github.com/chartjs/Chart.js/blob/2954cf8a9928b863fa59c2a8bdc7701c328a027a/src/core/core.tooltip.js#L623), the function itemSort is always called with three parameters. Looks like the docs are incorrect?

No, docs are correct. JavaScript does not verify anything, so you can define a method with parameters but do not pass any parameter to such method.

In TypeScript you need to explicitly mark it as optional.

adripanico (approved):

PR #37792, opened 2019-08-21
Add @testing-library/jest-dom

jgoz:
Please fill in this template.

If adding a new definition:
- [ ] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration` &mdash; moving types out of the source repository (see https://github.com/testing-library/jest-dom/issues/123 for details)


types/testing-library__jest-dom/test/testing-library__jest-dom-global-tests.ts
```diff
@@ -0,0 +1,67 @@
+import { ValidElement } from "@testing-library/jest-dom";
```
sheetalkamat: This is not global test. 

From https://github.com/DefinitelyTyped/DefinitelyTyped#how-do-i-write-definitions-for-packages-that-can-be-used-globally-and-as-a-module
>  The global test file should exercise the definition according to how it would be used in a script loaded on a web page where the library is available on the global scope - in this scenario you should not specify an import statement.

types/testing-library__jest-dom/index.d.ts
```diff
@@ -0,0 +1,75 @@
...
+
+// CONTRIBUTOR NOTE: When adding/updating matchers, ensure you handle
+// them in both the Global section and the Module section
+
+export type ValidElement = HTMLElement | SVGElement;
+
+// Global types (for `import "@testing-library/jest-dom/extend-expect"`)
+
+declare global {
+    namespace jest {
```
sheetalkamat: Move this global declaration change into file `extend-expect.d.ts` since the wayt to handle is using `import "@testing-library/jest-dom/extend-expect"`
Refer to https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#library-file-layout

sheetalkamat (changes requested):

types/testing-library__jest-dom/index.d.ts
jgoz: Will that make the global declarations available to user code even if the `import` statement does not appear in a TypeScript-compiled file?

Typically, this plugin is configured by including that import in a JS file that is only used by Jest (and Jest does not support reading TS files directly for configuration). The purpose of moving these declarations to DT is so that the global type definitions are just "available" without forcing users to add a separate `.d.ts` file that includes the import statement.

types/testing-library__jest-dom/test/testing-library__jest-dom-global-tests.ts
jgoz: Fixed, thank you.

sheetalkamat:
Going to hold of reviewing this for https://github.com/testing-library/jest-dom/issues/123 still has ongoing discussion.

gnapse:
@jgoz can this effort be resumed in light of testing-library/jest-dom#175?

jgoz:
@sheetalkamat I have rebased this branch and updated it to reflect the API that _will be_ published as version 5.0 in `@testing-library/jest-dom` (see the [`next` branch](https://github.com/testing-library/jest-dom/tree/next), specifically [this PR](https://github.com/testing-library/jest-dom/pull/175)).

The plan is as follows:
- Publish `@types/testing-library__jest-dom` global typings to DT (this PR)
- Remove internal typings from `jest-dom`
- Add dependency to DT typings
- Publish version 5.0

However, there is a slight problem that the DT build is failing because I've targeted a non-existent version (5.0). Do you know if it is possible to target pre-release versions in DT headers?

If not, we might want to revise our plan such that merging & publishing this PR is the last step and we add our dependency in 5.0.1 (FYI @gnapse).

eps1lon:
@jgoz 
You don't have to release a breaking change as far as I can tell:
1. target an existing version in `@types/`
2. cut a new release in `jest-dom` with a dependency on `@types/` and no bundled types

This way we don't need any special attention for this PR.

types/testing-library__jest-dom/index.d.ts
```diff
@@ -0,0 +1,36 @@
...
+        toBeEnabled(): R;
+        toBeInvalid(): R;
+        toBeRequired(): R;
+        toBeValid(): R;
+        toContainElement(element: HTMLElement | SVGElement | null): R;
+        toContainHTML(htmlText: string): R;
+        toHaveAttribute(attr: string, value?: any): R;
+        toHaveClass(...classNames: string[]): R;
+        toHaveFocus(): R;
+        toHaveFormValues(expectedValues: { [name: string]: any }): R;
```
eps1lon:
```suggestion
        toHaveFormValues(expectedValues: Record<string, unknown>): R;
```

Since this targets typescript 3.0 we can use `unknown` instead. `{ [name: string]: any }` has a special behavior in that it allows any object type. But this actually does not consider Symbol keys in the implementation. It should throw for those case as well as when passing e.g. functions. It's a niche edge case but for new types we might as well do it right.

See https://github.com/microsoft/TypeScript/wiki/Breaking-Changes#-k-string-unknown--is-no-longer-a-wildcard-assignment-target

types/testing-library__jest-dom/index.d.ts
```diff
@@ -0,0 +1,36 @@
...
+        toBeVisible(): R;
+        toBeEmpty(): R;
+        toBeDisabled(): R;
+        toBeEnabled(): R;
+        toBeInvalid(): R;
+        toBeRequired(): R;
+        toBeValid(): R;
+        toContainElement(element: HTMLElement | SVGElement | null): R;
+        toContainHTML(htmlText: string): R;
+        toHaveAttribute(attr: string, value?: any): R;
```
eps1lon: nit:
```suggestion
        toHaveAttribute(attr: string, value?: unknown): R;
```

jgoz:
> 1. target an existing version in `@types/`

That's the problem. These typings are only valid for the currently un-released version 5.0.

I'm thinking the easiest way to do this will be to release 5.0 without typings, merge this PR, and then add the dependency in 5.0.1.

eps1lon:
> That's the problem. These typings are only valid for the currently un-released version 5.0.

You could "just" copy the existing types and later update for the breaking changes. I don't see a need to address both issues at the same time

> I'm thinking the easiest way to do this will be to release 5.0 without typings, merge this PR, and then add the dependency in 5.0.1.

Would work as well. Basically what I proposed.

johnnyreilly:
Whichever way forward makes most sense; let pragmatism be your guide 🤗

jgoz:
@eps1lon Thanks for the feedback &mdash; applied your suggested changes.

`@testing-library/jest-dom` @ `5.0.0` has been published, so this should now be a straightforward review (@johnnyreilly, @sheetalkamat).

To recap, the library now globally extends Jest's `expect` object via the root import (previously, it was via `jest-dom/extend-expect`). This means the proposed typings in this PR exactly match the library's API.

eps1lon (approved):

ganymedes01:
any chance of this being merged?

gnapse:
Indeed. Right now we've got jest-dom v5.0.0 out but it does not support types at all, because of the chicken-or-the-egg problem we had. And we're blocked not being able to release the fix in v.5.0.1 because it depends on this PR.

jgoz:
I wonder if I should re-open this PR. It seems to be stuck at "Revision Needed" and "Needs Author Attention", so it might simply not be getting looked at. Thoughts?

gnapse:
It'd be a pity because even if a PR needs author attention, the author should be able to ping back the maintainers when that attention has been given. I'm pinging @johnnyreilly and @sheetalkamat who seem to be part of those that answered on the review side of things. But feel free to pursue that other course of action @jgoz if you think we should.

gnapse:
We already have new PRs (at least one so far) ready to be merged in `jest-dom` that would trigger a new release. We were hoping to have this merged sooner so that our very first release after v5.0.0 would be v5.0.1 with the fix of adding back TS support.

I'm deferring merging any new PRs with the hope that we achieve this, but please please, we need this merged ASAP. Or at least to know if it's not mergeable for some reason.

eps1lon:
This was one of the reason why I proposed doing this for version 4 instead of trying to coordinate simultaneous releases across organizations. The DefinitelyTyped repo has PRs that need a "merge[d] ASAP" all the time.

I would propose adding types for v5 to the actual package in the meantime as a patch and remove them until this PR is merged.

gnapse:
> I would propose adding types for v5 to the actual package in the meantime as a patch and remove them until this PR is merged.

I was under the impression that they (DefinitelyTyped) required for the lib to have no types before they accept the PR. In any case I think we'll need to do that 😕

eps1lon:
> > I would propose adding types for v5 to the actual package in the meantime as a patch and remove them until this PR is merged.
> 
> I was under the impression that they (DefinitelyTyped) required for the lib to have no types before they accept the PR. In any case I think we'll need to do that

In that case disregard my comment. Though I think the difference is that in the case of types being introduced into `@types/*` CI doesn't complain. So if you explain your plan CI would've greenlit the PR and the PR could've been merged. This would result in a state where we have redundant types which is far better than having no types.

johnnyreilly:
If I follow correctly, this appears to be safe to merge. I'm happy to merge it - is that good for everyone?

types/testing-library__jest-dom/index.d.ts
```diff
@@ -0,0 +1,36 @@
+// Type definitions for @testing-library/jest-dom 5.0
+// Project: https://github.com/testing-library/jest-dom
+// Definitions by: Ernesto García <https://github.com/gnapse>
+//                 John Gozde <https://github.com/jgoz>
+//                 Seth Macpherson <https://github.com/smacpherson64>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+/// <reference types="jest" />
```
emmenko: I'm wondering why this needs to be a dependency of `@types/jest`, since Jest ships with its own type declarations now. Can this be removed?

jgoz: This was added prior to Jest 25 being released. AFAIK, Jest 24’s types were not included.

I’d have to look at how Jest’s built-in typings work before giving an answer to your question.

PR #37804, opened 2019-08-21
Add missing RequestStub methods to jasmine-ajax

jgonggrijp:
I have code that uses `jasmine.Ajax.stubRequest('/some/url').andCallFunction()` and soon found out that the latter method hadn't been typed yet. So I added `.andCallFunction` to the `RequestStub` interface and while I was at it, also the missing `.andError` and `.andTimeout` methods and the `.url` and `.query` properties.

The `.data` and `.method` properties were needlessly optional. These properties are always created in the constructor, so I made them non-optional.

`.andCallFunction` was newly introduced in jasmine-ajax version 3.3. For this reason, I bumped the version number accordingly. However, I did not check whether the typings completely represent version 3.3 of jasmine-ajax.

As a habit when contributing to DefinitelyTyped, I always try whether there are any linter rule exceptions that can easily be re-enabled. In this case, I found that the `ban-types` rule could be met by a single line change, so I did that.

While adding tests, I noticed that the pre-existing tests seem to test for correct logic rather than correct typing. Those tests might have been copied verbatim from the jasmine-ajax codebase. The tests that I added, however, test for correct typing.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/jasmine/jasmine-ajax/blob/master/src/requestStub.js and https://github.com/jasmine/jasmine-ajax/pull/152
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header. **Not exactly. I'm adding methods that have been missing for a while (`RequestStub.andError` and `.andTimeout`) as well as a method that was first added in version 3.3 (`.andCallFunction`). I did not check whether the definitions are completely up-to-date with version 3.3, but I bumped the version because of the new method.**
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. *Already present, but I removed one of the exceptions.*

### DIFF
--- a/types/jasmine-ajax/index.d.ts
+++ b/types/jasmine-ajax/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for jasmine-ajax 3.1
+// Type definitions for jasmine-ajax 3.3
 // Project: https://github.com/jasmine/jasmine-ajax
 // Definitions by: Louis Grignon <https://github.com/lgrignon>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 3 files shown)


sheetalkamat (changes requested):

jgonggrijp:
@sheetalkamat Could you please have another look? I need the result of this PR for my work.

@minestarks Perhaps @sheetalkamat is unavailable at this time. If this is the case, could you or somebody else take over? I'm asking you because we worked together on a review before.

jgonggrijp:
Thanks @minestarks!

PR #37806, opened 2019-08-21
Add types for protocols

ffflorian:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/protocols/index.d.ts
+++ b/types/protocols/index.d.ts
@@ -1,0 +1,14 @@
+// Type definitions for protocols 1.4
+// Project: https://github.com/IonicaBizau/protocols
+// Definitions by: Florian Keller <https://github.com/ffflorian>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/**
+ * @param input The input url.
+ * @param first If `true`, the first protocol will be returned. If number, it will represent the zero-based index of the protocols array.
+ * @returns The array of protocols or the specified protocol.
+ */
+declare function protocols(input: string, first: true | number): string;
+declare function protocols(input: string, first?: false): string[];
+
+export = protocols;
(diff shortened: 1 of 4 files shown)


types/protocols/index.d.ts
```diff
@@ -0,0 +1,15 @@
...
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/**
+ * @param input The input url.
+ * @param first If `true`, the first protocol will be returned. If number, it will represent the zero-based index of the protocols array.
+ * @returns The array of protocols or the specified protocol.
+ */
+declare function protocols(input: string, first?: boolean | number): string | string[];
+declare function protocols(input: string, first: true | number): string;
+declare function protocols(input: string, first?: false): string[];
```
sheetalkamat: These should be 
```ts
declare function protocols(input: string, first?: false): string[];
declare function protocols(input: string, first: true | number): string;
```

sheetalkamat (changes requested):

types/protocols/index.d.ts
ffflorian: Thanks, fixed!

types/protocols/protocols-tests.ts
```diff
@@ -0,0 +1,6 @@
+import protocols = require('protocols');
+
+protocols('git+ssh://git@some-host.com/and-the-path/name'); // $ExpectType string | string[]
```
sheetalkamat: This should have `$ExpectType string[]`

sheetalkamat (changes requested):

types/protocols/protocols-tests.ts
ffflorian: Fixed, thanks!

types/protocols/protocols-tests.ts
```diff
@@ -0,0 +1,6 @@
+import protocols = require('protocols');
+
+protocols('git+ssh://git@some-host.com/and-the-path/name'); // $ExpectType string
```
sheetalkamat: This is incorrect. Expected type needs to be string[]

sheetalkamat (changes requested):

types/protocols/protocols-tests.ts
ffflorian: Sorry :smiley: Should be good now.

PR #33527, opened 2019-03-01
chore(node): reduce usage of `Function` and `any`

SimonSchick:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/dist/latest-v11.x/docs/api/documentation.html
- [ ] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

This PR aims to improve type safety across the board by eliminating most usage of `Function` and reducing `any` usage by a fair bit.

PR #37812, opened 2019-08-21
[@types/node] Http2ServerResponse extends Stream

trivikr:
This was fixed in newer versions in #33527 but got missed in < v10
Http2ServerResponse extended Stream since it was introduced in v8.4.0 ([code](https://github.com/nodejs/node/blob/f70261fb3089b8acf5f68584ef0285a1c11f54fd/lib/internal/http2/compat.js#L427))

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/node/blob/f70261fb3089b8acf5f68584ef0285a1c11f54fd/lib/internal/http2/compat.js#L427

### DIFF
--- a/types/node/index.d.ts
+++ b/types/node/index.d.ts
@@ -37,6 +37,7 @@
 //                 Jordi Oliveras Rovira <https://github.com/j-oliveras>
 //                 Thanik Bhongbhibhat <https://github.com/bhongy>
 //                 Marcin Kopacz <https://github.com/chyzwar>
+//                 Trivikram Kamat <https://github.com/trivikr>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 // NOTE: These definitions support NodeJS and TypeScript 3.2.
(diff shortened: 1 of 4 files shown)


Flarna (approved):
Usually it's not needed to adapt EOL versions like Node.Js 9 but it's no problem to do it.

SimonSchick (changes requested):
Test pls

trivikr:
@SimonSchick Is there a procedure to add tests for extends?
I don't see type checks in existing http2 tests https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/test/http2.ts

SimonSchick:
Just add something that ensures it extends stream.

trivikr:
@SimonSchick I don't think it's necessary in this PR as:
* none of the derived classes are tested for type of base class in http2 tests
* this is a bug in @types/node < v10 and is already fixed in v11/v12

SimonSchick (approved):
🤷‍♂

PR #32341, opened 2019-01-21
Add default export for angular-ui-bootstrap

Den-dp:
This will allow `import uiBootstrapModule from 'angular-ui-bootstrap';`

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/angular-ui/bootstrap/blob/baf0a567f9f4641848a9a4699695f9cde5f28f70/index.js#L3
The same code can be found in `angular-ui-router`
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ca809fa16fdb3f3c908bac411b0204c501ae71be/types/angular-ui-router/index.d.ts#L11-L12
`angular-material`
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ca809fa16fdb3f3c908bac411b0204c501ae71be/types/angular-material/index.d.ts#L9-L10
and so on.
- [ ] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

Issue #14630, opened 2017-02-15
@types/angular-ui-bootstrap


jsmarsch:
I tried using the types/angular-ui-bootstrap package, and now I have tons of build errors along the lines of:
"Build" duplicate identifier 'angular'
these are being raised from the angular-animate index.d.ts (which was fine before I imported angular-ui-bootstrap)

I also get the error:
Buile: Module 'angular.ui' has no exported member 'bootstrap' these errors are in the actual angular-ui-bootstrap index.d.ts file -- sorry, but I really don't understand what's happening here.

My setup:
we use typings to pull our definintions (we are on typescript 1.8.10

if I uninstall the angular-ui-bootstrap typings, everything works (we have over a dozen or so typings packages installed, and no problems, so I _think_ my configuration is ok. 

Is there something wrong with the typings, or is there something I should look at on my end?

I don't know whether this helps, but typings installed the angular-bootstrap-ui definitions under typings/modules, but all of the other angular-* typings were installed under typings/globals.  I have to confess, I do not understand the rhyme or reason behind what goes into the /globals folder or the /modules folder, so maybe that is part of the problem?

orta:
Hi thread, we're moving DefinitelyTyped to use [GitHub Discussions](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/53377) for conversations the `@types` modules in DefinitelyTyped.

To help with the transition, we're closing all issues which haven't had activity in the last 6 months, which includes this issue. If you think closing this issue is a mistake, please pop into the [TypeScript Community Discord](https://discord.gg/typescript) and mention the issue in the `definitely-typed` channel.

Issue #8394, opened 2016-03-05
angular-ui-bootstrap.d.ts not up-to-date


blujay2013:
- [X ] I want to talk about `xxxx/xxxx.d.ts`.

The latest version of angular-ui-bootstrap.d.ts on DefinitelyTyped is 0.13.3; however, angular-ui-bootstrap's latest version is 1.2.1. Where can I access a more recent version of angular-ui-bootstrap.d.ts? 

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/angular-ui-bootstrap/angular-ui-bootstrap.d.ts
https://angular-ui.github.io/bootstrap/

Thanks!

r-arias:
I second this.

Den-dp:
`angular-ui-bootstrap` was updated to 2.2.0, so yes, we should bump it up.

PR #37813, opened 2019-08-21
[angular-ui-bootstrap] Export module name

Igorbek:
This PR adds exports of the module name `'ui.bootstrap'` from the module. It started to do it from version 1.0.0, so the version is upgraded as well.

Also, `exports = ` does not work with any other exports, so other exports are removed, which is ok because types are available through `ng.ui.bootstrap`. And this is ok since the version is major.


Igorbek:
It might be fixing #14630, partially #8394 and supersedes #32341

tomyam1:
Hey @Igorbek 
May I ask why did you remove all the type aliases?

Before we could do
```typescript
import { IModalInstanceService, IModalScope, IModalSettings } from 'angular-ui-bootstrap'
```

Now we have to do something like:
```typescript
import IModalInstanceService = angular.ui.bootstrap.IModalInstanceService 
import IModalScope = angular.ui.bootstrap.IModalScope 
import IModalSettings = angular.ui.bootstrap.IModalSettings 
```
Or maybe there is another solution?

Igorbek:
Hey @tomyam1 thank you for bringing a question.

So the package starting from 1.0.0 uses commonjs export `module.exports = ` (see https://github.com/angular-ui/bootstrap/blob/1.0.0/index.js)

This means the only way to import it is to write:

```ts
import * as uibootstrap from 'angular-ui-bootstrap';
// this also works with --syntheticDefaultModules
import uibootstrap from 'angular-ui-bootstrap';
```

Named imports (`import { ... } from `) are not correct in this context.
For that reason, TypeScript explicitly restricted to have both `export = ` and other named exports, including types.

Other type packages for angular (for example see type definitions in `@types/angular-route` or `@types/angular-sanitize`) also used this approach.

This was a breaking change, definitely, but the workaround is to use types in `angular.ui.bootstrap`.
For that reason, I bumped to a major version change.

BTW, there's a trick to define a default export along with types. `@types/angular-ui-router` (deprecated) used this approach. But it only works if TS has synthetic modules enebled, otherwise the statement is not correct from runtime perspective.

Also, the fix would be to ask authours of `angular-ui-library` to create a new version with default exports (proper es modules).

Igorbek:
Also, before the change I made a _bit ts query_ to see how many users of these types:
https://bigtsquery.firebaseapp.com/query?selector=ImportDeclaration:has(StringLiteral%5Bvalue%3D%27angular-ui-bootstrap%27%5D)%20ImportSpecifier:nth-child(1)

It seems it is used in a few projects for type imports, so the scope is quite limited.
I understand the inconvenience, sorry but I think it is the right thing to do.

tomyam1:
Thanks @Igorbek for the explanation.

PR #37814, opened 2019-08-21
Added type definitions for leaflet-responsive-popup

stankurek:
Please fill in this template.

- [ v] Use a meaningful title for the pull request. Include the name of the package modified.
- [ v] Test the change in your own code. (Compile and run.)
- [ v] Add or edit tests to reflect the change. (Run with `npm test`.)
- [v ] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [v ] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [ v] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If adding a new definition:
- [v ] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--

### DIFF
--- a/types/leaflet-responsive-popup/index.d.ts
+++ b/types/leaflet-responsive-popup/index.d.ts
@@ -11,7 +11,9 @@ declare module 'leaflet' {
         hasTip?: boolean;
     }
 
-    class responsivePopup extends Popup {
+    class ResponsivePopup extends Popup {
         constructor(options?: ResponsivePopupOptions, source?: Layer);
     }
+
+    function responsivePopup(options?: ResponsivePopupOptions, source?: Layer): ResponsivePopup;
 }
--- a/types/leaflet-responsive-popup/index.d.ts
+++ b/types/leaflet-responsive-popup/index.d.ts
@@ -1,0 +1,17 @@
+// Type definitions for leaflet-responsive-popup 0.6
+// Project: https://github.com/yafred/leaflet-responsive-popup
+// Definitions by: Stan Kurek <https://github.com/stankurek>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
+
+import * as L from 'leaflet';
+
+declare module 'leaflet' {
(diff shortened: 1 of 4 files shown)


types/leaflet-responsive-popup/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6",
+            "dom"
+        ],
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "strictNullChecks": false,
```
sheetalkamat: strictness cannot be turned off, please turn on both flags and fix errors if any

types/leaflet-responsive-popup/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+// TypeScript Version: 2.3
+
+import * as L from 'leaflet';
+
+declare module 'leaflet' {
+    interface ResponsivePopupOptions extends PopupOptions {
+        hasTip?: boolean;
+    }
+
+    class responsivePopup extends Popup {
```
sheetalkamat: Looking at https://unpkg.com/leaflet-responsive-popup@0.6.4/leaflet.responsive.popup.js::
`ResponsivePopup` seems like a class and `responsivePopup` seems like a function so you need both, one class and one function

sheetalkamat (changes requested):

types/leaflet-responsive-popup/tsconfig.json
stankurek: Thank you for your review. I changed the file.

types/leaflet-responsive-popup/index.d.ts
stankurek: Good point. Thanks for taking the time to check it.

sheetalkamat (approved):

PR #37818, opened 2019-08-22
React-select update types of noOptionsMessage and placeholder

ReinoutStevens:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://react-select.com/props

### DIFF
--- a/types/react-select/src/Select.d.ts
+++ b/types/react-select/src/Select.d.ts
@@ -161,7 +161,7 @@ export interface Props<OptionType = { label: string; value: string }> extends Se
   /* Name of the HTML Input (optional - without this, no input will be rendered) */
   name?: string;
   /* Text to display when there are no options */
-  noOptionsMessage?: (obj: { inputValue: string }) => React.ReactNode;
+  noOptionsMessage?: (obj: { inputValue: string }) => string | null;
   /* Handle blur events on the control */
   onBlur?: FocusEventHandler;
   /* Handle change events on the select */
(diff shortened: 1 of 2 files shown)


ReinoutStevens:
The docs state that noOptionsMessage should return a string or null, but works fine with ReactElements

types/react-select/src/Select.d.ts
```diff
@@ -161,7 +161,7 @@ export interface Props<OptionType = { label: string; value: string }> extends Se
   /* Name of the HTML Input (optional - without this, no input will be rendered) */
   name?: string;
   /* Text to display when there are no options */
-  noOptionsMessage?: (obj: { inputValue: string }) => string | null;
+  noOptionsMessage?: (obj: { inputValue: string }) => React.ReactNode;
```
Methuselah96: I'm fine with the `placeholder` change, but I'm not sure I agree with changing the type of `noOptionsMessage`. While it might very well work with any React node right now, there are no guarantees that it will work with a React node in the future since it's not officially stated anywhere that you can return any React node and the Flow types also don't allow for any React node. I would suggest creating a PR on `react-select` to update their [Flow types and documentation](https://github.com/JedWatson/react-select/blob/master/packages/react-select/src/Select.js#L201) and then we can update it here.

Methuselah96 (changes requested):

types/react-select/src/Select.d.ts
ReinoutStevens: Sure, makes sense. Undid that change

Methuselah96 (approved):

PR #37763, opened 2019-08-20
[WIP] Allow for mongoose mapped values & "constructor"

richarddd:
Solves: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37088

- [X] Use a meaningful title for the pull request. Include the name of the package modified.
- [X] Test the change in your own code. (Compile and run.)
- [X] Add or edit tests to reflect the change. (Run with `npm test`.)
- [X] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [X] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Issue #37088, opened 2019-07-23
Can't set constructor on document type in latest @types/mongoose


tomyam1:
The following code fails with TS2345 after 26587aa3aa75e0021180a9a7f9b4cb45e036002b:

Tested on Typescript 3.5.3.

```typescript
import * as Mongoose from 'mongoose'

export interface IPersonDocument extends Mongoose.Document {
    // The failure happens because we have constructor in IPersonDocument
    constructor: IPersonModel

    firstName: string
}

export interface IPersonModel extends Mongoose.Model<IPersonDocument> {

}

let Person: IPersonModel;

/*
Fails with typescript 3.5.3
TS2345: Argument of type '{ firstName: string; }' is not assignable to parameter of type 'Partial<IPersonDocument>'.   Types of property 'constructor' are incompatible.     Type 'Function' is missing the following properties from type 'IPersonModel': watch, translateAliases, bulkWrite, findById, and 56 more.
 */
const john = new Person({
    firstName: 'John',
});
```

@richarddd 
@armanio123

richarddd:
Aha! The tests do not cover this usecase. We should probably handle that as well as a unary type! Could you create a PR?

tomyam1:
`DeepPartial<T> & { constructor?: T}` gives the same error.
But ignoring the constructor solved it `Pick<DeepPartial<T>, Exclude<keyof DeepPartial<T>, 'constructor'>>`.

But then we got to other with this type checking. 

For instance, this valid code fails.

A string is valid source for an ObjectId but DeepPartial won't allow it.

I actually think it's best to remove the type checking from new function since DeepPartial is way too simplistic solution.

```typescript
import * as Mongoose from 'mongoose'

export interface IPersonDocument extends Mongoose.Document {
    father: Mongoose.Types.ObjectId // id of father record
}

export interface IPersonModel extends Mongoose.Model<IPersonDocument> {

}

let Person: IPersonModel;


const john = new Person({
    // TS2559: Type 'string' has no properties in common with type 'DeepPartial<ObjectId>'.
    father: 'valid-object-id'
});
```

richarddd:
> `DeepPartial<T> & { constructor?: T}` gives the same error.
> But ignoring the constructor solved it `Pick<DeepPartial<T>, Exclude<keyof DeepPartial<T>, 'constructor'>>`.
> 
> But then we got to other with this type checking.
> 
> For instance, this valid code fails.
> 
[...]

PR #37823, opened 2019-08-22
[@types/mongoose] Revert source type of model constructor to any

tomyam1:
The type checking added to the model constructor by @richarddd in

26587aa3aa75e0021180a9a7f9b4cb45e036002b
22761c884a9542171583f83ccd178acd0c1e3b11

Causes valid code not to pass type checking with not way of allowing it

As described in issue #37088:

1. It doesn't allow conversion made by Mongoose from source data types to the data types in the document: e.g. from [string to ObjectID](https://github.com/Automattic/mongoose/blob/5.6.11/lib/cast/objectid.js#L25), from [multiple string to boolean](https://mongoosejs.com/docs/schematypes.html#booleans), etc.
2. It doesn't allow settings the constructor type of a document (this is useful so you can access the model of a document using `doc.consutrctor` [1](https://stackoverflow.com/a/17685260/1705056).
3. It doesn't allow [custom schema types](https://mongoosejs.com/docs/customschematypes.html) which could have any arbitrary cast function

All of this are valid use cases for mongoose and the type checking should not
prevent them.

PR #37763 by richarddd tries to solve some of these issues but it only addresses some of them and doesn't even do it [successfully](As shown in https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37088#issuecomment-522889112). 

Stricter type checking for the source of a document can be a good thing, but the
current solution by @richarddd is incorrect.

Please allow this revert back to `any`.

Please fill in this template.

The lists of casts not supported by the current solution by @richarddd is endless, but here are some references:

string as source for numbers:
https://mongoosejs.com/docs/api.html#mongoose_Mongoose-Number

Custom casts
https://mongoosejs.com/docs/customschematypes.html

From source:
Casts made by mongoose by default:
https://github.com/Automattic/mongoose/tree/5.6.10/lib/cast

### DIFF
--- a/types/mongoose/index.d.ts
+++ b/types/mongoose/index.d.ts
@@ -68,16 +68,6 @@ declare module "mongoose" {
   import mongoose = require('mongoose');
 
   /**
-   * Allows for nested objects and arrays to be partial
-   */
-  export type DeepPartial<T> = {
-    [P in keyof T]?:
-    T[P] extends (infer U)[] ? DeepPartial<U>[] :
-    T[P] extends object ? DeepPartial<T[P]> :
-    T[P];
-  };
-
-  /**
    * Gets and optionally overwrites the function used to pluralize collection names
    * @param fn function to use for pluralization of collection names
    * @returns the current function used to pluralize collection names (defaults to the `mongoose-legacy-pluralize` module's function)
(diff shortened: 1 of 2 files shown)


orblazer (approved):
I approve for now but i think its a good way to have `DeepPartial` of @richarddd but this need more flexibility

**EDIT** : I have try and the `DeepPartial` break all `Ref<...>` from typegoose (declared here https://github.com/szokodiakos/typegoose/blob/master/src/prop.ts#L430).

PR #37826, opened 2019-08-22
Add typings for jsonpointer

WesselKuipers:
These typings are based on [jsonpointer.d.ts](https://github.com/janl/node-jsonpointer/blob/master/jsonpointer.d.ts). No NPM releases have been created since the typings were added, which is over 2 years ago. These types can be removed again once [this PR](https://github.com/janl/node-jsonpointer/pull/43) has been merged.

I modified the type of the `JSONPointer` interface based on the return value of [`compile()`](https://github.com/janl/node-jsonpointer/blob/master/jsonpointer.js#L79).

### DIFF
--- a/types/jsonpointer/index.d.ts
+++ b/types/jsonpointer/index.d.ts
@@ -4,7 +4,7 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.2
 
-interface JSONPointer {
+export interface JSONPointer {
     /**
      * Looks up a JSON pointer in an object
      */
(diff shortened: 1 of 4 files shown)


types/jsonpointer/index.d.ts
```diff
@@ -0,0 +1,36 @@
...
+     */
+    function set(object: object, pointer: string, value: any): void;
+
+    /**
+     *  Builds a JSONPointer instance from a pointer value.
+     */
+    function compile(pointer: string): JSONPointer;
+}
+
+export = JSONPointer;
```
sheetalkamat: Instead of using `export =` i suggest:
```ts
export interface JSONPointer {
    /**
     * Looks up a JSON pointer in an object
     */
    get(object: object): any;

    /**
     * Set a value for a JSON pointer on object
     */
    set(object: object, value: any): void;
}

    /**
     * Looks up a JSON pointer in an object
     */
    export function get(object: object, pointer: string): any;

    /**
     * Set a value for a JSON pointer on object
     */
    export function set(object: object, pointer: string, value: any): void;

    /**
     *  Builds a JSONPointer instance from a pointer value.
     */
    export function compile(pointer: string): JSONPointer;
```

sheetalkamat (changes requested):

types/jsonpointer/index.d.ts
WesselKuipers: Sure thing. Done!

sheetalkamat (approved):

PR #37832, opened 2019-08-22
jira-client -- add missing method getIssueWorklogs

tiger-seo:
https://github.com/jira-node/node-jira-client/blob/master/src/jira.js#L1146


orta:
Cool, thanks!

PR #37842, opened 2019-08-22
[@types/sanitize-html] Adds escapeDisallowedTags

martinmckenna:
## Purpose

This PR accompanies https://github.com/apostrophecms/sanitize-html/pull/169 which allows a boolean option called `escapeDisallowedTags` in the sanitize-html package.

The new option will display blacklisted HTML attributes as a string, rather than removing them entirely.

### DIFF
--- a/types/sanitize-html/index.d.ts
+++ b/types/sanitize-html/index.d.ts
@@ -53,6 +53,7 @@ declare namespace sanitize {
     allowedClasses?: { [index: string]: string[] } | boolean;
     allowedIframeHostnames?: string[];
     allowIframeRelativeUrls?: boolean;
+    escapeDisallowedTags?: boolean;
     allowedSchemes?: string[] | boolean;
     allowedSchemesByTag?: { [index: string]: string[] } | boolean;
     allowedSchemesAppliedToAttributes?: string[];


YuJianrong:
@martinmckenna But this PR have not been merged into the master branch. I think you'd better create this PR after the PR in apostrophecms/sanitize-html merged.

martinmckenna:
@amcasey @YuJianrong Library PR is merged. This one should be good to go.

martinmckenna:
accidentally closed this

martinmckenna:
Looks like the blocking PR got merged and then reverted. I'll post again when I have an update.

RyanCavanaugh:
Is this OK to merge?

afshin:
https://github.com/apostrophecms/sanitize-html/pull/169#issuecomment-533556574

The PR that added this parameter has been reverted, so I don't think this should be merged, here.

RyanCavanaugh:
Should it be closed instead?

afshin:
I don't know the background. But I would say yet, because it's small and can be re-opened when necessary.

RyanCavanaugh:
Thanks!

PR #37843, opened 2019-08-22
[@types/stripe] createSource accepts bank account arguments

danlite:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: (https://stripe.com/docs/api/customer_bank_accounts/create?lang=node)

### DIFF
--- a/types/stripe/stripe-tests.ts
+++ b/types/stripe/stripe-tests.ts
@@ -635,6 +635,41 @@ stripe.customers.createSource(
     }
 );
 
+stripe.customers.createSource(
+    "cus_5rfJKDJkuxzh5Q",
+    {
+        source: {
+            country: 'US',
+            currency: 'USD',
+            account_holder_name: 'Account Holder',
+            account_holder_type: 'individual',
+            account_number: '000123456789',
+            routing_number: '110000000'
+        }
+    },
+    (err, bankAcc) => {
+        // asynchronously called
+        const obj: Stripe.IBankAccount = bankAcc as Stripe.IBankAccount;
+    }
+);
+stripe.customers.createSource(
+    "cus_5rfJKDJkuxzh5Q",
+    {
+        source: {
+            country: 'US',
+            currency: 'USD',
+            account_holder_name: 'Account Holder',
+            account_holder_type: 'individual',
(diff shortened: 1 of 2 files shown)


types/stripe/stripe-tests.ts
```diff
@@ -635,6 +635,41 @@ stripe.customers.createSource(
...
+            currency: 'USD',
+            account_holder_name: 'Account Holder',
+            account_holder_type: 'individual',
+            account_number: '000123456789',
+            routing_number: '110000000'
+        }
+    }).then(
+    (bankAcc) => {
+        // asynchronously called
+        const obj: Stripe.IBankAccount = bankAcc as Stripe.IBankAccount;
```
vaskevich: Can you change this to `bankAcc; // $ExpectType IBankAccount`? The current casts won't do a ton of good.

vaskevich (approved):
LGTM

danlite:
I made the recommended change, and then it turns out the method signature I intended (provide bank account source, return `IBankAccount`) wasn't being enforced, so good catch.

However, I can't get tests to pass now, with the error:
```
ERROR: 922:9  expect  TypeScript@next expected type to be:
  IList<ICard | IBankAccount>
got:
  IList<IBankAccount | ICard>
```

If I switch the order of the union, the (inverse) error occurs for TypeScript@3.0... I'll submit my broken commit in the hopes that someone can chime in.

danlite:
For reference:
https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/575551389#L320

vaskevich:
Uh oh. Looks like they started alphabetizing union types in `typescript@next`. Is there a reason that it's not returning just `IBankAccount`, though? Provided the parameters passed in are correct, TS should be able to infer the return type as just `IBankAccount` or `ICard`, depending on the parameters.

danlite:
The weird thing is, it wasn't failing on the method I modified (`stripe.customers.createSource`), but rather on `stripe.accounts.retrieve` 🤔 

In any case, I've created a helper type that seems to resolve that issue. It feels a little hacky, so I'm open to other suggestions.

PR #37845, opened 2019-08-22
[@types/node] Update Buffer.from single argument

kevin-greene-ck:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_buffer

The definitions for Node Buffer.from when taking a single argument don't behave as expected.

Example (this doesn't compile):
```
function test(value: Buffer | string): Buffer {
    return Buffer.from(value)
}
```

However, this does compile:
```
function test(value: Buffer | string): Buffer {
    if (typeof value === 'string') {
        return Buffer.from(value)
    } else {
        return Buffer.from(value)
    }
}
```

These changes allow either to compile.

All other changes in this commit are from auto formatting.

### DIFF
--- a/types/node/index.d.ts
+++ b/types/node/index.d.ts
@@ -38,6 +38,7 @@
 //                 Thanik Bhongbhibhat <https://github.com/bhongy>
 //                 Marcin Kopacz <https://github.com/chyzwar>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.1
 
 // NOTE: These definitions support NodeJS and TypeScript 3.2.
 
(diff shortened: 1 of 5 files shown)


Flarna:
Could you please remove the format changes as this makes it hard to review.

types/node/index.d.ts
```diff
@@ -38,6 +38,7 @@
 //                 Thanik Bhongbhibhat <https://github.com/bhongy>
 //                 Marcin Kopacz <https://github.com/chyzwar>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.1
```
Flarna: This will for sure cause a lot fails with other modules as definition depending on node (>1500) would need the same min requirement.

smac89:
Kinda difficult to see the changes you've made amidst all the format changes. This problem seems to be related to this typescript issue:

https://github.com/microsoft/TypeScript/issues/14107

PR #37850, opened 2019-08-22
react: function type as a state in useState

zepatrik:
The issue is that you have to provide a `lazyInitialState` to have a state of a function type in order to work.

``` ts
// this is accepted by current type definition but react will think it is a lazy state
const [callback, setCallback] = useState<() => void>(() => undefined)

// this should be forced by the type definition
const [callback, setCallback] = useState<() => void>(() => () => undefined)
```

I'm having a hard time typing this correctly as for some reason typescript narrows down the types when relying on inference with a conditional type. Also tried to use overloading which did not work either. Not sure if that is a bug in typescript or if I am just to stupid.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://reactjs.org/docs/hooks-reference.html#lazy-initial-state

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -833,6 +833,7 @@ declare namespace React {
      * @version 16.8.0
      * @see https://reactjs.org/docs/hooks-reference.html#usestate
      */
+    // function useState<S extends Function>(lazyState: () => S): [S, Dispatch<SetStateAction<S>>];
     function useState<S>(initialState: S extends Function ? () => S : (() => S) | S): [S, Dispatch<SetStateAction<S>>];
     // convenience overload when first argument is ommitted
     /**
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -833,7 +833,7 @@ declare namespace React {
      * @version 16.8.0
      * @see https://reactjs.org/docs/hooks-reference.html#usestate
      */
-    function useState<S>(initialState: S | (() => S)): [S, Dispatch<SetStateAction<S>>];
+    function useState<S>(initialState: S extends Function ? () => S : (() => S) | S): [S, Dispatch<SetStateAction<S>>];
     // convenience overload when first argument is ommitted
     /**
      * Returns a stateful value, and a function to update it.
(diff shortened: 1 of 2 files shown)


ferdaber:
React doesn’t support bare functions as state, if you do want that you should wrap it inside an object so that it’s a method of the object.

ferdaber:
The required typing is a bit complex, here's a past PR that broke some things for past reference: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36080

zepatrik:
Right, thanks for the reference.
So what is the desired way to go? Either the typing should disallow function as state (technically not right) or force you to use lazy initial state and dispatch form of setState.

ferdaber:
If you can make the tests work with your changes, I'd say go for it. It may be a rare use case but we should still support it since the runtime allows it.

types/react/index.d.ts
```diff
@@ -833,7 +833,8 @@ declare namespace React {
      * @version 16.8.0
      * @see https://reactjs.org/docs/hooks-reference.html#usestate
      */
-    function useState<S>(initialState: S | (() => S)): [S, Dispatch<SetStateAction<S>>];
+    // function useState<S extends Function>(lazyState: () => S): [S, Dispatch<SetStateAction<S>>];
+    function useState<S>(initialState: S extends Function ? () => S : (() => S) | S): [S, Dispatch<SetStateAction<S>>];
```
Jessidhia: `Function` is not allowed by the linter; and it seems that using a conditional type here is causing the compiler to infer stricter / const-like types for the input 😞

I haven't tested it, but try this:

```ts
function useState<T>(initializer:
  | () => T
  | Exclude<T, (...args: any[]) => any>
): [T, Dispatch<SetStateAction<T>>]
```

You also need to do the same change on `SetStateAction`.

zepatrik: Right, it would work if the compiler wouldn't infer stricter. I could not find any resources on why this is happening.

zepatrik:
So closing this as I am not able to do this right. Inference with conditional types in typescript behaves quite weird.

PR #37854, opened 2019-08-23
[selenium-webdriver] Fix missing/incorrect webdriver.Builder methods

dbjorge:
I went through all the `Builder` methods on the implementation and corrected any differences I saw.

Some of these changes are breaking, but would only break cases that would already be broken at runtime.

Summary of changes:
* Adds typings for the following webdriver.Builder methods that were missing: `getChromeOptions`, `setChromeService`, `setEdgeService`, `getFirefoxOptions`, `setFirefoxService`, `setIeService`, `getSafariOptions`, `getHttpAgent`
  * The `getEdgeOptions` and `setSafariService` methods are missing intentionally, since they are missing in the [implementation](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/index.js)
* Adds missing `ie.ServiceBuilder` and `ie.Level` types
* Widens the return type of `Builder.getWebDriverProxy` to allow for it to return `null`, per the implementation and docs
* Makes `Builder` and `Capabilities`' `setAlertBehavior` parameter optional per the implementation.
* (BREAKING) fixes the `ie.Options.setLogLevel` signature to match the [implementation](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/ie.js), which uses an IE-specific log level and not the general logging.Level type.
* (BREAKING) replaces the incorrectly-named `Builder.setSafari` method with `Builder.setSafariOptions` to match the [implementation](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/index.js)
* (BREAKING) updates the type signature on `ie.Driver.createSession` to match the [implementation](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/ie.js)

---

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [selenium-webdriver/index.js](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/index.js), [selenium-webdriver/ie.js](https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/ie.js)
- [n/a] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

### DIFF
--- a/types/selenium-webdriver/ie.d.ts
+++ b/types/selenium-webdriver/ie.d.ts
@@ -20,7 +20,7 @@ export class Driver extends webdriver.WebDriver {
    * @return {!Driver} A new driver instance.
    */
   static createSession(
-    options: webdriver.Capabilities|Options,
+    options?: webdriver.Capabilities|Options,
     opt_service?: remote.DriverService): Driver;
 
   /**
--- a/types/selenium-webdriver/ie.d.ts
+++ b/types/selenium-webdriver/ie.d.ts
@@ -1,4 +1,11 @@
 import * as webdriver from './index';
+import * as remote from './remote';
+
+/**
+ * IEDriverServer logging levels.
+ * @enum {string}
+ */
+export type Level = 'FATAL' | 'ERROR' | 'WARN' | 'INFO' | 'DEBUG' | 'TRACE';
 
 /**
  * A WebDriver client for Microsoft's Internet Explorer.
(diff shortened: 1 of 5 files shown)


cnishina:
I am on vacation and won't be able to review this

beta-vulgaris (approved):
looks good to me!

types/selenium-webdriver/ie.d.ts
```diff
@@ -7,11 +14,14 @@ export class Driver extends webdriver.WebDriver {
...
-   * @param {(capabilities.Capabilities|Options)=} opt_config The configuration
-   *     options.
+   * @param {(Capabilities|Options)=} options The configuration options.
+   * @param {(remote.DriverService)=} opt_service The `DriverService` to use
+   *   to start the IEDriverServer in a child process, optionally.
    * @return {!Driver} A new driver instance.
    */
-  static createSession(opt_config?: webdriver.Capabilities|Options): Driver;
+  static createSession(
+    options: webdriver.Capabilities|Options,
```
rbuckton: Is `options` no longer optional? The JSDoc comment above and the [implementation](https://github.com/SeleniumHQ/selenium/blob/5fe20e7658148a4fff90f612556853746bca1cad/javascript/node/selenium-webdriver/ie.js#L392-L393) seem to indicate it is still optional.

dbjorge: Good catch, thanks! I assumed it was non-optional based on the implementation using the name `options` instead of the `opt_config` the other browsers' equivalents use, but I agree that it should be optional based on the JSDoc and implementation. Updated.

PR #37857, opened 2019-08-23
[lodash] Add missing multiply import

henrikra:
## Background

While importing multiply like this works `import { multiply } from 'lodash';` the import like this doesn't work `import multiply from 'lodash/multiply';`.

Then I just noticed that for some reason file that makes this import correct for typescript was missing so I added it.

Here is also picture from lodash docs that shows you can import specific functions like this:
image

What do you guys think? :)

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://lodash.com/

### DIFF
--- a/types/lodash.multiply/index.d.ts
+++ b/types/lodash.multiply/index.d.ts
@@ -1,0 +1,10 @@
+// Type definitions for lodash.multiply 4.9
+// Project: http://lodash.com/
+// Definitions by: Brian Zengel <https://github.com/bczengel>, Ilya Mochalov <https://github.com/chrootsu>, Stepan Mikhaylyuk <https://github.com/stepancar>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+// Generated from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/lodash/scripts/generate-modules.ts
+
+import { multiply } from "lodash";
+export = multiply;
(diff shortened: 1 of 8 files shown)


aj-r (changes requested):
Great catch! I'm amazed this went unnoticed for so long.

Please do the same thing in the `ts3.1` folder.

henrikra:
@aj-r Can you be more specific what you mean? This is my first PR to Lodash types :D

henrikra:
@aj-r I think I got it right. Can you check? :)

aj-r (changes requested):
Yep, that's exactly what I meant :).
I just remembered that there are actaully 2 versions of this that you can import:

- `import multiply from 'lodash/multiply';` (this is the one you just added)
- `import multiple from 'lodash.multiply';` (this is actually outside of the lodash folder, and is a separate npm package)

We actually have a script that generates both of these: `lodash/scripts/generate-modules.ts`. In theory, all you need to do is modify that script (add `multiply` into `allModuleNames()`) and run it.

To run the script:

1. Open a terminal in `lodash/scripts`.
2. `npm install`
3. `ts-node generate-scripts` (requires `ts-node` to be installed globally - alternatively you can use the local version at `lodash/scripts/node_modules/.bin/ts-node`)

I haven't run this script in a very long time so let me know if you have issues.

henrikra:
@aj-r I think I got it now. Can you check again? :D

aj-r (approved):

PR #37861, opened 2019-08-23
[node] Update promisified version of readdir.

JiangJie:
- [x] [node] Add a `__promisify__` declaration for namespace `readdir`, to support single parameter `path` when using `util.promisify`.

### DIFF
--- a/types/node/fs.d.ts
+++ b/types/node/fs.d.ts
@@ -867,13 +867,6 @@ declare module "fs" {
          * @param path A path to a file. If a URL is provided, it must use the `file:` protocol.
          * @param options The encoding (or an object specifying the encoding), used as the encoding of the result. If not provided, `'utf8'` is used.
          */
-        function __promisify__(path: PathLike, options?: { encoding: BufferEncoding | null; withFileTypes?: false } | BufferEncoding | null): Promise<string[]>;
-
-        /**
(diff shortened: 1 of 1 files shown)


types/node/fs.d.ts
```diff
@@ -889,6 +889,12 @@ declare module "fs" {
          * @param options If called with `withFileTypes: true` the result data will be an array of Dirent
          */
         function __promisify__(path: PathLike, options: { encoding?: string | null; withFileTypes: true }): Promise<Dirent[]>;
+
+        /**
+         * Asynchronous readdir(3) - read a directory.
+         * @param path A path to a file. If a URL is provided, it must use the `file:` protocol.
+         */
+        function __promisify__(path: PathLike): Promise<string[]>;
```
Flarna: Whats wrong with the first overload on line 870? it requires `path: PathLike` as first argument, the second one is optional and the return type is `Promise<string[]>]`.
If I'm wrong then please add a test for the new supported usecase.

Flarna (changes requested):

types/node/fs.d.ts
JiangJie: Test case
```javascript
    fn: F

const files = preaddir('/');
```

PR #37689, opened 2019-08-16
Make pg.QueryResult generic, increasing type safety

inad9300:
`any` is used as default type for backwards compatibility.

PR #37892, opened 2019-08-24
[@types/pg] Generify query methods

HoldYourWaffle:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: *not applicable*
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header. *(not applicable)*

<hr>

I generified the `query` methods as well as the corresponding interfaces. `R` will be the return type of the query, `I` a tuple for the parameters.

A usage example:
```ts
interface Person {
    name: string;
    naughty: boolean;
}

/* 
   The types are defined here:
   - The return type R will be the 'Person' interface defined above
   - The parameters (I for input) will be a string and boolean (in order)
*/
client.query<Person, [string, boolean]>(
    'SELECT * FROM user WHERE name = $1 AND naughty = $2',
    [ 'Johnny', true ], // this array will be typechecked
    (err, result) => console.log(result.rows[0].name) // this will now have a 'string' type
);
```
These additional type checks are aimed at users who have interfaces for their database types (for example auto-generated by [schemats](https://github.com/sweetiq/schemats)).
It is **100% opt-in and backwards compatible** with the use of generic defaults added in TS 2.3.

The usefulness `R` parameter (for the return type) pretty much speaks for itself, it enables intellisense/typechecking without the use of annoying type assertions.

The `I` parameter (the inputs) might be too repetitive/over-safe for some people, which is why it comes after the (more useful) returntype generic. For my use-case I'm automatically generating parts of these method calls, and an extra sanity check is very much welcome in that case. It's also a very welcome safeguard when refactoring database types, although that probably doesn't happen too often for most people.

Because these generics are meant to check the parameters supplied directly the `no-unnecessary-generics` rule doesn't apply here. The `dtslint` page says:

> Generic type parameters allow you to relate the type of one thing to another; if they are used only once, *they can be replaced with their type constraint.*

The emphasized part doesn't apply here, because the whole idea of these generics is to have users (optionally) manually provide an extra layer of security by specifying *their own types*. `any` is a perfectly valid type constraint here, the generics enable users to narrow it down to their own types if they so wish.

### DIFF
--- a/types/pg/index.d.ts
+++ b/types/pg/index.d.ts
@@ -88,7 +88,7 @@ export interface QueryResult<R extends QueryResultRow = any> extends QueryResult
     rows: R[];
 }
 
-export interface QueryArrayResult<R = any[]> extends QueryResultBase {
+export interface QueryArrayResult<R extends any[] = any[]> extends QueryResultBase {
     rows: R[];
 }
 
@@ -162,10 +162,10 @@ export class Pool extends events.EventEmitter {
 
     query<T extends Submittable>(queryStream: T): T;
     // tslint:disable:no-unnecessary-generics
-    query<R = any[], I extends any[] = any[]>(queryConfig: QueryArrayConfig<I>, values?: I): Promise<QueryArrayResult<R>>;
+    query<R extends any[] = any[], I extends any[] = any[]>(queryConfig: QueryArrayConfig<I>, values?: I): Promise<QueryArrayResult<R>>;
     query<R extends QueryResultRow = any, I extends any[] = any[]>(queryConfig: QueryConfig<I>): Promise<QueryResult<R>>;
     query<R extends QueryResultRow = any, I extends any[] = any[]>(queryTextOrConfig: string | QueryConfig<I>, values?: I): Promise<QueryResult<R>>;
-    query<R = any[], I extends any[] = any[]>(queryConfig: QueryArrayConfig<I>, callback: (err: Error, result: QueryArrayResult<R>) => void): void;
+    query<R extends any[] = any[], I extends any[] = any[]>(queryConfig: QueryArrayConfig<I>, callback: (err: Error, result: QueryArrayResult<R>) => void): void;
     query<R extends QueryResultRow = any, I extends any[] = any[]>(queryTextOrConfig: string | QueryConfig<I>, callback: (err: Error, result: QueryResult<R>) => void): void;
     query<R extends QueryResultRow = any, I extends any[] = any[]>(queryText: string, values: I, callback: (err: Error, result: QueryResult<R>) => void): void;
     // tslint:enable:no-unnecessary-generics
@@ -182,10 +182,10 @@ export class ClientBase extends events.EventEmitter {
 
     query<T extends Submittable>(queryStream: T): T;
     // tslint:disable:no-unnecessary-generics
-    query<R = any[], I extends any[] = any[]>(queryConfig: QueryArrayConfig<I>, values?: I): Promise<QueryArrayResult<R>>;
+    query<R extends any[] = any[], I extends any[] = any[]>(queryConfig: QueryArrayConfig<I>, values?: I): Promise<QueryArrayResult<R>>;
     query<R extends QueryResultRow = any, I extends any[] = any[]>(queryConfig: QueryConfig<I>): Promise<QueryResult<R>>;
     query<R extends QueryResultRow = any, I extends any[] = any[]>(queryTextOrConfig: string | QueryConfig<I>, values?: I): Promise<QueryResult<R>>;
-    query<R = any[], I extends any[] = any[]>(queryConfig: QueryArrayConfig<I>, callback: (err: Error, result: QueryArrayResult<R>) => void): void;
+    query<R extends any[] = any[], I extends any[] = any[]>(queryConfig: QueryArrayConfig<I>, callback: (err: Error, result: QueryArrayResult<R>) => void): void;
     query<R extends QueryResultRow = any, I extends any[] = any[]>(queryTextOrConfig: string | QueryConfig<I>, callback: (err: Error, result: QueryResult<R>) => void): void;
     query<R extends QueryResultRow = any, I extends any[] = any[]>(queryText: string, values: any[], callback: (err: Error, result: QueryResult<R>) => void): void;
     // tslint:enable:no-unnecessary-generics
@@ -217,7 +217,7 @@ export interface PoolClient extends ClientBase {
     release(err?: Error): void;
 }
 
-export class Query<R, I extends any[]> extends events.EventEmitter implements Submittable {
+export class Query<R extends QueryResultRow = any, I extends any[] = any> extends events.EventEmitter implements Submittable {
     constructor(queryTextOrConfig?: string | QueryConfig<I>, values?: I);
     submit: (connection: Connection) => void;
     on(event: "row", listener: (row: R, result?: ResultBuilder<R>) => void): this;
--- a/types/pg/index.d.ts
+++ b/types/pg/index.d.ts
@@ -1,7 +1,8 @@
 // Type definitions for pg 7.11
 // Project: http://github.com/brianc/node-postgres
-// Definitions by: Phips Peter <https://github.com/pspeter3>
+// Definitions by: Phips Peter <https://github.com/pspeter3>, Ravi van Rooijen <https://github.com/HoldYourWaffle>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
 
 /// <reference types="node" />
 
@@ -48,17 +49,17 @@ export interface PoolConfig extends ClientConfig {
     Promise?: PromiseConstructorLike;
 }
 
-export interface QueryConfig {
+export interface QueryConfig<I extends any[] = any[]> {
     name?: string;
     text: string;
-    values?: any[];
+    values?: I;
 }
 
 export interface Submittable {
     submit: (connection: Connection) => void;
 }
 
-export interface QueryArrayConfig extends QueryConfig {
+export interface QueryArrayConfig<I extends any[] = any[]> extends QueryConfig<I> {
     rowMode: 'array';
 }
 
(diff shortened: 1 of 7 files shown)


HoldYourWaffle:
It seems like `pg-copy-streams` has a lower minimum TS version causing the tests to fail, should I update its minimum version?

types/pg/index.d.ts
```diff
@@ -1,7 +1,8 @@
 // Type definitions for pg 7.11
 // Project: http://github.com/brianc/node-postgres
-// Definitions by: Phips Peter <https://github.com/pspeter3>
+// Definitions by: Phips Peter <https://github.com/pspeter3>, Ravi van Rooijen <https://github.com/HoldYourWaffle>
```
pspeter3: Do you want to review PRs?

pspeter3 (commented):
How does this relate to #37689?

types/pg/index.d.ts
HoldYourWaffle: I don't mind doing it

HoldYourWaffle:
Seems like #37689 is a reduced implementation of this since it only types the return value. I'm not 100% sure what the additional `RowBase` interface is supposed to be though.

Interesting to see that someone had this idea too. I've had this changes running in production for a while now, I honestly just forgot to send a PR out 😅

inad9300:
`RowBase` was there to constraint the generic types you can pass around. For instance, we can tell at compile time that `query<number>(...)` is a bug, as no SQL query can be written in the context of pg that results in a number (nor an array of numbers) being returned.

types/pg/index.d.ts
```diff
@@ -79,12 +80,12 @@ export interface QueryResultBase {
...
 
-export interface QueryResult extends QueryResultBase {
-    rows: any[];
+export interface QueryResult<R = any> extends QueryResultBase {
+    rows: R[];
 }
 
-export interface QueryArrayResult extends QueryResultBase {
-    rows: any[][];
+export interface QueryArrayResult<R = any> extends QueryResultBase {
```
inad9300: Wouldn't this need to follow an approach similar to that of `QueryConfig`, where a tuple type is expected as the generic type? Otherwise, I don't see the generic type (`R`) as being useful at all, as all you can express is a matrix full of numbers, or one full of strings, and so on, effectively not allowing to express resultsets with columns of different types.

HoldYourWaffle:
That makes a lot of sense, I'll incorporate something similar in this PR.
Would it make sense to add a similar constraint type to the `I` generic?

types/pg/index.d.ts
HoldYourWaffle: Good catch, that's definitely a mistake.

Would changing it to this make sense, I'm too tired to tell right now:
```ts
export interface QueryArrayResult<R extends any[] = any[]> extends QueryResultBase {
    rows: R[];
}
```

inad9300:
I believe that for `I` is enough with enforcing that it must extend `any[]`. You could possibly go further and say something like `extends (null | boolean | number | string | Date | Uint8Array)[]` or such, but you never know which types might be returned by the `query()` methods due to node-postgres' support for [custom parsers](https://node-postgres.com/features/types).

types/pg/index.d.ts
inad9300: It is also somewhat late on this part of the planet, but it looks legit :wink:

HoldYourWaffle:
> but you never know which types might be returned by the query() methods due to node-postgres' support for custom parsers.

I'm not sure what the return type of `query` has to do with `I`, `I` is for the input parameters.

> and the types can't be defined once and used for multiple sets of values either, I believe, which makes the type declaration quite redundant.

I'm not sure what you mean by this, could you clarify? What do you want to re-use?

<br>

> It feels to me as saying let s: string = '' (as opposed to just let s = '') and never using s. Could you please try to elaborate on why you think I could be useful?

In a way it's definitely like `let s: string = ''`. I already put my rationale for adding it in my original post:
```
The I parameter (the inputs) might be too repetitive/over-safe for some people, which is why it comes after the (more useful) returntype generic.
For my use-case I'm automatically generating parts of these method calls, and an extra sanity check is very much welcome in that case.
It's also a very welcome safeguard when refactoring database types, although that probably doesn't happen too often for most people.
```
To elaborate on my use-case, I have a setup where database type definitions are [generated for me](https://github.com/sweetiq/schemats). So I'm writing my queries like this:
```ts
await database.query<void, [UserFields.id]>(
    'UPDATE "user" SET date_last_seen = now() WHERE id=$1',
    [ user.id ]
);
```
As you see I'm not literally reusing the same type twice. The `user` variable could come from anywhere and its internal representation might differ from the database, this is where the extra type-check comes in handy.
This is of course a *very* simple example, but it's enough to demonstrate my use-case.
I recognize that not everyone will find this extra check useful, which is why it's 100% opt-in. Embedded SQL queries indeed inherently type-unsafe, but with this extra check hopefully a little less so.

inad9300:
> I'm not sure what the return type of `query` has to do with `I`, `I` is for the input parameters.

I intended to refer to custom serializers, as opposed to parsers. Sorry for the confusion.

> I'm not sure what you mean by this, could you clarify? What do you want to re-use?

If you could somehow declare the types of the values once and have them checked against multiple sets of values, I thought that would somehow be a possible justification for declaring them, but it doesn't seem to be the case. To illustrate what I meant, see the following example which allows exactly that, although through an extra function:

```ts
// Types declared once
function getSomething(values: [string, number]) {
  pg.query(`select * from something where s = $1 and n = $2`, values)
}

// Type-checking performed multiple times
getSomething(['0', 0])
getSomething(['1', 1])
```

> To elaborate on my use-case (...)

Thanks for the additional explanation. I still find it somewhat weird to specify the types of your own input, over which you have complete control. But I'm fine with the change.

HoldYourWaffle:
> Thanks for the additional explanation. I still find it somewhat weird to specify the types of your own input, over which you have complete control.

> But I'm fine with the change.

That's great to hear 😄 I think the bot wants you to push some approve button somewhere in that case.
In the meantime, how should I handle the travis failure? Should I raise the minimum typescript version for `pg-copy-streams`?

HoldYourWaffle:
@pspeter3 What do you think of these changes?

@inad9300 Can you double check for me if I implemented your `RowBase`-like constraint correctly?

inad9300 (approved):

inad9300:
`QueryResultRow` should be applied to `QueryResult` and *not* to `QueryResultArray`, as the latter represents (mostly) a matrix of primitives, and and `QueryResultRow` models a key-value structure. As a basic check that the code is correct, I believe the expression `extends QueryResultRow[]` should appear nowhere in the code.

HoldYourWaffle:
Whoops that's a pretty dumb mistake, no idea how I missed that. I think I've fixed it now, but it's too early for me to tell...

types/pg/index.d.ts
```diff
@@ -79,12 +80,16 @@ export interface QueryResultBase {
...
+	[column: string]: any;
 }
 
-export interface QueryArrayResult extends QueryResultBase {
-    rows: any[][];
+export interface QueryResult<R extends QueryResultRow = any> extends QueryResultBase {
+    rows: R[];
+}
+
+export interface QueryArrayResult<R = any[]> extends QueryResultBase {
```
inad9300: I would type this similar to `QueryArrayConfig`, i.e. `QueryArrayResult<R extends any[] = any[]>`.

types/pg/index.d.ts
```diff
@@ -208,12 +217,12 @@ export interface PoolClient extends ClientBase {
     release(err?: Error): void;
 }
 
-export class Query extends events.EventEmitter implements Submittable {
-    constructor(queryTextOrConfig?: string | QueryConfig, values?: any[]);
+export class Query<R, I extends any[]> extends events.EventEmitter implements Submittable {
```
inad9300: `ResultBuilder` requires `R` to extend `QueryResultRow`. Therefore, since `Query` receives an `R` and passes it to `ResultBuilder`, it should also enforce it to extend `QueryResultRow`. I believe the current code leads to a compilation error, but even if it doesn't, I would suggest to establish the aforementioned type restriction.

types/pg/index.d.ts
```diff
@@ -156,12 +161,14 @@ export class Pool extends events.EventEmitter {
...
 
     query<T extends Submittable>(queryStream: T): T;
-    query(queryConfig: QueryArrayConfig, values?: any[]): Promise<QueryArrayResult>;
-    query(queryConfig: QueryConfig): Promise<QueryResult>;
-    query(queryTextOrConfig: string | QueryConfig, values?: any[]): Promise<QueryResult>;
-    query(queryConfig: QueryArrayConfig, callback: (err: Error, result: QueryArrayResult) => void): void;
-    query(queryTextOrConfig: string | QueryConfig, callback: (err: Error, result: QueryResult) => void): void;
-    query(queryText: string, values: any[], callback: (err: Error, result: QueryResult) => void): void;
+    // tslint:disable:no-unnecessary-generics
+    query<R = any[], I extends any[] = any[]>(queryConfig: QueryArrayConfig<I>, values?: I): Promise<QueryArrayResult<R>>;
```
inad9300: Note that once `QueryArrayResult` enforces `R` to extend `any[]`, `query()` will have to declare the same restriction. Note as well that this comment applies to 4 cases in total, 2 in the context of `Pool` and 2 in the context of `Client`.

types/pg/index.d.ts
HoldYourWaffle: Good catch, I really should stop doing this stuff when I'm tired

types/pg/index.d.ts
HoldYourWaffle: Strangely enough it doesn't generate a compile error, maybe that's due to the `any` default type? But you're right that constraint should be added (and I have now).

I also completely forgot to add defaults here...

types/pg/index.d.ts
HoldYourWaffle: I've definitely fixed that before, no idea what happened to it. Maybe git is just being weird again...

HoldYourWaffle:
@inad9300 I have addressed your concerns.

I have also updated the minimum typescript version for @types/pg-copy-streams to avoid travis failing.

HoldYourWaffle:
Finally got travis to pass.
@inad9300 I think the only thing left now is for you to press some kind of "approve" button somewhere

mainnika (approved):
in general, it's a good PR 👍🏻
pg-pool is ok about these changes.

inad9300 (approved):

rbuckton:
Thanks for the contribution!

PR #37896, opened 2019-08-24
[express-serve-static-core] Add optional generic to Send

Georift:
`Send` has a type `any` passed in, I wanted a way to specify the type
I'm expecting to pass in so I get typechecking (eg.
`res.send<SomeVar>(test)` will assert `test` is typeof SomeVar)

This is optional and if no generic is provided it will fallback to the
existing `any` type.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [See Assertion considered harmful final code example](https://basarat.gitbooks.io/typescript/docs/types/type-assertion.html)

### DIFF
--- a/types/express-serve-static-core/index.d.ts
+++ b/types/express-serve-static-core/index.d.ts
@@ -488,7 +488,8 @@ export interface MediaType {
     subtype: string;
 }
 
-export type Send = (body?: any) => Response;
+// tslint:disable-next-line no-unnecessary-generics
+export type Send = <T = any>(body?: T) => Response;
 
 export interface Response extends http.ServerResponse, Express.Response {
     /**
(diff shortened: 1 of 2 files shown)


Georift:
@borisyankov I'm having second thoughts about if this was the correct way to do this. Having to disable the `no-unused-generics` rule gives my the feeling this mightn't be the most correct way to address this issue.

Perhaps we could have the optional generic strung through the Response interface down to the Send type? Do you have any ideas how to be approch this?

types/express-serve-static-core/index.d.ts
```diff
@@ -488,7 +488,8 @@ export interface MediaType {
     subtype: string;
 }
 
-export type Send = (body?: any) => Response;
+// tslint:disable-next-line no-unnecessary-generics
+export type Send = <T = any>(body?: T) => Response;
```
amcasey: I can see why you'd want to make this sort of type check, but I'm not sure it ought to be associated with a particular function (i.e. `Send`).  Perhaps you'd be better off with a general purpose helper like 
```ts
function checkType<T>(t: T): T { return t; }
```

amcasey (changes requested):

PR #37908, opened 2019-08-25
[react-loadable-visibility] add types for react-loadable-visibility

dbartholomae:
Do not merge this PR yet, there is still one issue I would appreciate help with.

This is my first try at definition files (I usually write TypeScript directly) and my first PR to DefinitelyTyped, so a good eye from a reviewer would be appreciated. Specifically I am not sure how to best type the pass-through of default exporting a function in `react-loadable.d.ts` and would need an expert to help before this PR can be merged. I cannot `import Loadable from 'react-loadable'` since it is declared via `export =`, but I neither can `import * as Loadable from 'react-loadable'` since `Loadable` is a function and module imports can not be called. What to do?

Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/react-loadable-visibility/tsconfig.json
+++ b/types/react-loadable-visibility/tsconfig.json
@@ -1,0 +1,31 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6"
+        ],
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "strictFunctionTypes": true,
+        "strictNullChecks": true,
+        "baseUrl": "../",
+        "jsx": "react",
+        "typeRoots": [
+            "../"
+        ],
+        "types": [],
+        "noEmit": true,
+        "forceConsistentCasingInFileNames": true,
+        "paths": {
+            "@loadable/component": [
+                "loadable__component"
+            ]
+        }
+    },
+    "files": [
+        "index.d.ts",
+        "loadable-components.d.ts",
+        "react-loadable.d.ts",
+        "react-loadable-visibility-tests.tsx"
+    ]
+}
--- a/types/react-loadable-visibility/index.d.ts
+++ b/types/react-loadable-visibility/index.d.ts
@@ -1,0 +1,8 @@
+// Type definitions for react-loadable-visibility 3.0
+// Project: https://github.com/stratiformltd/react-loadable-visibility#readme
+// Definitions by: Daniel Bartholomae <https://github.com/dbartholomae>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+import Loadable from './loadable-components';
+export default Loadable;
--- a/types/react-loadable-visibility/loadable-components.d.ts
+++ b/types/react-loadable-visibility/loadable-components.d.ts
@@ -1,0 +1,2 @@
+import Loadable from '@loadable/component';
+export default Loadable;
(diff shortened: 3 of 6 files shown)


dbartholomae:
Concerning my original issue there is one more idea I am toying with, that looks a bit hacky, but seems to work:
```typescript
import { Loadable } from 'react-loadable';
declare const LoadableExport: Loadable;
export default LoadableExport;
```
This is trying to mimick the type structure used by `react-loadable`.

Furthermore it seems my way to include a fixture file (`test-component.tsx`) is not the intended way. I would like to add in a fixture to mimick the use case as closely as possible (react-loadable helps to asynchronously load files) but if this is preferred I could also ditch the file and try to add a `Promise` to mock the file import. I am not perfectly sure what that promise should look like, but I think I would be able to figure it out :)

dbartholomae:
I have added a [PR to types-publisher](https://github.com/microsoft/types-publisher/pull/651) for the dependency @loadable/component

dbartholomae:
I now also tested the change in my own code (the remaining tick box from the template). I have furthermore repushed the same code in a new commit to trigger another build since the PR to types-publisher was merged.

dbartholomae:
Stuck again because Travis is failing while locally it is not:
```
Could not find a declaration file for module '@loadable/component'. '/home/travis/build/DefinitelyTyped/DefinitelyTyped/types/react-loadable-visibility/node_modules/@loadable/component/dist/loadable.cjs.js' implicitly has an 'any' type.
  Try `npm install @types/loadable__component` if it exists or add a new declaration (.d.ts) file containing `declare module '@loadable/component';`
```
Not sure what to do about this, as I cannot add a `@types` dependency and if I do I get a message that I should not do that.

sandersn:
I tried to repro this myself and got it to fail locally with `npm run lint react-loadable-visibility` from the DT root. I'm looking into what's wrong.

sandersn:
Hm. The error definitely comes from `@loadable/component` not being typed. You don't need the package.json to depend on it -- instead you should depend on the one inside Definitely Typed.

I got that to work with `import Loadable from 'loadable__component'`. I found that a couple of other packages work this way (`@material/*` for example), but the majority of packages successfully use the unmangled name. I'll compare the tsconfig of `@babel/*` to `@material/*` to see what's different. It's likely that the material ones are wrong.

types/react-loadable-visibility/tsconfig.json
```diff
@@ -0,0 +1,31 @@
...
+        "noEmit": true,
+        "forceConsistentCasingInFileNames": true
+    },
+    "files": [
+        "index.d.ts",
+        "loadable-components.d.ts",
+        "react-loadable.d.ts",
+        "react-loadable-visibility-tests.tsx"
+    ],
+    "paths": {
```
sandersn: got it. this should be inside `"compilerOptions"` for it to work. Not sure why we don't give an error for this.

types/react-loadable-visibility/package.json
```diff
@@ -0,0 +1,6 @@
+{
+    "private": true,
+    "dependencies": {
+        "@loadable/component": "^5.10.1"
```
sandersn: this points to the original `@loadable/component` which is untyped. You should delete all of package.json since you only need to depend on packages within DT.

sandersn (changes requested):

types/react-loadable-visibility/package.json
dbartholomae: Done. Does it make sense to revert the PR that added `@loadable/component` to the whitelist?
https://github.com/microsoft/types-publisher/pull/651

types/react-loadable-visibility/tsconfig.json
dbartholomae: Done. I have also created a PR to improve the error message that lead me here:
https://github.com/microsoft/types-publisher/pull/653

sandersn (approved):

PR #37915, opened 2019-08-26
[aws-lambda] Use optional properties instead of union type with null

exoego:
Because optional properties allow to omit keys:
```typescript
const o: AttributeValue = {
    S: "string"
};
```

But ` | null` do not (if I understand correctly):
```typescript
const o: AttributeValue = {
    S: "string",
    B: null,
    BS: null,
    BOOL: null,
    L: null,
    M: null,
    NS: null,
    NULL: null,
    SS: null
};
```

Please fill in this template.

### DIFF
--- a/types/aws-lambda/aws-lambda-tests.ts
+++ b/types/aws-lambda/aws-lambda-tests.ts
@@ -92,19 +92,19 @@ strOrUndefined = apiGwEvtReqCtx.domainName;
 strOrUndefined = apiGwEvtReqCtx.eventType;
 strOrUndefined = apiGwEvtReqCtx.extendedRequestId;
 str = apiGwEvtReqCtx.httpMethod;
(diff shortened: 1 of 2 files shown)


simonbuchan (changes requested):
This will break anyone using `null` right now. Further, these seem to mostly be input properties: there they should reflect what AWS are actually sending, which from a quick test seems to be "include the property with a value of `null`" for API Gateway proxy events, e.g. the current type is correct in that case (I'm not re-testing all these events!)

Further, several of these properties *already* were declared as optional, but also allowed `null`, and you've only removed the ability to use `null`. It's *possible* that this is correct, e.g. if you're pretty damn certain that AWS won't ever send null for an input, and will fail if you send it null as an output, but I'm going to guess that you just removed all the `null`s.

Was there something in particular?

exoego:
Sorry, closing since this is breaking change and controversial.

PR #37922, opened 2019-08-26
[ldapjs] class OperationsError added

ManfredSteiner:
The package does not already provide class OperationsError, although it is used in [docu](http://ldapjs.org/server.html#errors) for ldapjs.

Workaround as long this pull request is not merged: 
 `return next(new (<any>ldapjs).OperationsError('...'));`

### DIFF
--- a/types/ldapjs/index.d.ts
+++ b/types/ldapjs/index.d.ts
@@ -349,7 +349,9 @@ export class NoSuchAttributeError {
 export class ProtocolError {
 	constructor(error?: string);
 }
-
+export class OperationsError {
+	constructor(error?: string);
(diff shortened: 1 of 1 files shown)


mscottnelson (approved):
Definitely a missing type. Appears that Error types in general could use some more work. I'll take a look at them in a future PR

mscottnelson:
Just commenting to inform @ManfredSteiner that I fleshed out all of the other errors in a branch that I expect to merge within the next week, as well as updating their error types to match the actual error types emitted by ldapjs. I also re-established the target library version as v1.0 in that PR.

PR #32883, opened 2019-02-08
Update backbone-relational to version 0.10.0

jgonggrijp:
Nowadays, backbone-relational follows the recommended practice in the TypeScript documentation, to publish your functionality in your own module instead of extending somebody else's module (Backbone in older versions). This pull request updates the type definitions accordingly.

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/PaulUithol/Backbone-relational/commit/6ce19db76bf4cf45048c8859982ebc3228379a54#diff-c99214ca1078164501c576035b7bd1a0
- [x] Increase the version number in the header if appropriate.
- [x] ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.~ (already present)

@eirikhm

PR #37926, opened 2019-08-26
Add the missing Store instance to backbone-relational

jgonggrijp:
backbone-relational exports a `Store` class. It also exports a default instance of this class, called `store` (note the lowercase `s`), but the latter was missing in the typings. So I added it.

The tests that were intended to test for the `store` export, revealed that `Store.unregister` was typed incorrectly. I fixed that as well.

Fixing that, I ran into an unwarranted linter error. `types/backbone/index.d.ts` exports `Model` and `Collection<TModel>`, where `TModel` defaults to `Model`. `types/backbone-relational/index.d.ts` imports these classes as `BModel` and `Collection`, respectively, and defines a `Model` class of its own that extends `BModel`. This confused the `use-default-type-parameter` linter rule when encountering `Collection<Model>` on line 166. I disabled the rule on that line.

- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present). *This is nowadays automatically included with the tests.*
    - http://backbonerelational.org/#Store
    - https://github.com/DefinitelyTyped/DefinitelyTyped/pull/32883
    - https://github.com/PaulUithol/Backbone-relational/blob/4e2f5c005e798364e1d1b4fcb7dfa4cd5d1c6989/backbone-relational.js#L518-L522
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header. *Not applicable.*

### DIFF
--- a/types/backbone-relational/backbone-relational-tests.ts
+++ b/types/backbone-relational/backbone-relational-tests.ts
@@ -110,3 +110,8 @@ var theirHouse = new House({ id: 'house-2' });
 paul.set({ 'livesIn': theirHouse });
 
 alert('theirHouse.occupants=' + theirHouse.get('occupants').pluck('name'));
+
+BackboneRel.store.removeModelScope(window);
+BackboneRel.store.addModelScope(window);
+BackboneRel.store.unregister(Person);
+BackboneRel.store.reset();
--- a/types/backbone-relational/index.d.ts
+++ b/types/backbone-relational/index.d.ts
@@ -162,6 +162,7 @@ export class Store extends EventsMixin implements Events {
 
     update(model:Model):void;
 
+    // tslint:disable-next-line use-default-type-parameter
     unregister(type: Model | Collection<Model> | typeof Model): void;
 
     reset():void;
--- a/types/backbone-relational/index.d.ts
+++ b/types/backbone-relational/index.d.ts
@@ -162,7 +162,7 @@ export class Store extends EventsMixin implements Events {
 
     update(model:Model):void;
 
-    unregister(model:Model, collection:Collection<BModel>, options:any):void;
+    unregister(type: Model | Collection<Model> | typeof Model): void;
 
     reset():void;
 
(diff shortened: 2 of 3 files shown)


adblanc:
Hi, trying to declare a class like in the types test :
```ts
import Backbone from "backbone";
import * as BackboneRel from "backbone-relational";

export default class Guild extends BackboneRel.Model {

};

const guild = new Guild();
```
gives me this error : `Uncaught TypeError: Object.setPrototypeOf: expected an object or null, got undefined`.

When console logging `BackboneRel`, its gives me an empty object.

To fix this, I was able to do something like so :

```ts
import Backbone from "backbone";
import "backbone-relational";

export default class Guild extends Backbone.RelationalModel {

};

const guild = new Guild();
```
and in a separate declaration file : 
```ts
declare module Backbone {
// all declarations from @types/backbone-relational
  export class RelationalModel extends Backbone.Model {
// ...
 }
}
```
In my understanding, backbone-relational module modifies backbone module, so why would we put backbone-relational types somewhere else.
If someone could explain it to me or tell me how I could handle this case differently, I'd appreciate, thanks :smile:

jgonggrijp:
@adblanc Thanks for reporting this. It seems I made a mistake in a previous PR: I updated the types to the latest version on `master`, which still carries the 0.10 version number in its source code but which has quite advanced since the 0.10 that was actually published to NPM. Sorry about that.

I'm not sure how to fix this. Downgrade the module interface again? Make a hybrid to cater to different users? Bump the version to "master"? Asking @sandersn for guidance.

sandersn:
I'm confused about the state of `backbone-relational` itself: it's still at 0.10, so shouldn't `@types/backbone-relational` be at 0.10 as well? And it hasn't updated in 5 years.

Oh, I think I see. This PR provides types for an unpublished version of `backbone-relational`. In that case, yeah you should revert the changes to match 0.10. DT only supports types for published packages.

PR #37940, opened 2019-08-26
Add missing FormState element to redux-form

HendrikRoehm:
According to the [example](https://redux-form.com/8.2.2/examples/submitvalidation/) on the redux-form page, it is possible to have a general error field for the whole form. I confirmed this behaviour with redux-form 8.2.6.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://redux-form.com/8.2.2/examples/submitvalidation/

I am not sure about this. I searched in the redux-form source code where the error field was added, but could not find it.

### DIFF
--- a/types/redux-form/lib/reducer.d.ts
+++ b/types/redux-form/lib/reducer.d.ts
@@ -19,11 +19,11 @@ export interface FormStateMap {
 
 export interface FormState {
     registeredFields: RegisteredFieldState[];
-    fields?: {[name: string]: FieldState};
+    fields?: { [name: string]: FieldState };
     values?: { [fieldName: string]: any };
     active?: string;
     anyTouched?: boolean;
-    error: string;
+    error?: any;
     submitting?: boolean;
     submitErrors?: { [fieldName: string]: string };
     submitFailed?: boolean;
--- a/types/redux-form/lib/reducer.d.ts
+++ b/types/redux-form/lib/reducer.d.ts
@@ -23,6 +23,7 @@ export interface FormState {
     values?: { [fieldName: string]: any };
     active?: string;
     anyTouched?: boolean;
+    error: string;
     submitting?: boolean;
     submitErrors?: { [fieldName: string]: string };
     submitFailed?: boolean;


bwlt:
@HendrikRoehm did you mean [this prop](https://redux-form.com/8.2.2/docs/api/props.md/#-code-error-any-code-)?
according to the documentation it could be `any` value

HendrikRoehm:
Thanks @bwlt for mentioning the api documentation. The error component of the injected props is set correctly in reduxForm.d.ts line 82. I am dealing with another problem: The error key is missing in the (redux) state. 
The error element is set in [createReducer.js](https://github.com/erikras/redux-form/blob/master/src/createReducer.js) line 585 and 604. This is missing in the types currently. Nevertheless, you are right that the type should be any and not string. I have changed that.

What is not clear to me: The injectedFieldProps uses a variable ErrorType, which defaults to string. Shouldn't this also be used in the redux state definition to be consistent?

bwlt:
> What is not clear to me: The injectedFieldProps uses a variable ErrorType, which defaults to string. Shouldn't this also be used in the redux state definition to be consistent?

good point! I thought that is could be safe to add, but there are some issues:
Before add it it will be nice to have a check (or add a test) where the type parameter will be added to the `FormState` interface, but this type parameter should be also be present on other types (because you don't use `FormState` directly).
The points where to add the type parameter (and pass it down to `FormState`/`FormStateMap`):
1. [`interface FormStateMap`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/714e727c8fb8a9de041ec4eaea39ba9e5c76293a/types/redux-form/lib/reducer.d.ts#L16)
2. [`interface FormReducer`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/714e727c8fb8a9de041ec4eaea39ba9e5c76293a/types/redux-form/lib/reducer.d.ts#L4)
3. [`function formValueSelector`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/714e727c8fb8a9de041ec4eaea39ba9e5c76293a/types/redux-form/lib/formValueSelector.d.ts#L3)
4. [`type GetFormState`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/714e727c8fb8a9de041ec4eaea39ba9e5c76293a/types/redux-form/lib/reduxForm.d.ts#L23)

In my opinion adding it on `interface FormStateMap` (point 1) is not safe, because the interface have an index signature. This means that you can not describe correctly the ErrorType of each form you have: once you set an ErrorType it must be valid for all the forms.
For this reason I think that is not the best solution

uniqueiniquity (changes requested):
Marking as changes requested until @bwlt's concern is addressed.

bwlt:
@uniqueiniquity in my opinion the current changes are ok for now.
The proposal described [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37940#issuecomment-527347552) is not feasible for now

uniqueiniquity (approved):

uniqueiniquity:

PR #37941, opened 2019-08-27
[node] Reorder promisified version of readdir.

JiangJie:
Put
```javascript
function __promisify__(path: PathLike, options?: { encoding: BufferEncoding | null; withFileTypes?: false } | BufferEncoding | null): Promise<string[]>;
```
at the end of `readdir`; otherwise the following test case will fail.

Test case
```javascript
    fn: F

const files = preaddir('/');
```


JiangJie:
This is a better version compare to #37861.

JiangJie:
This will test fail.
Closed.

PR #37943, opened 2019-08-27
Add promisified version of readdir.

JiangJie:
Reopen #37861 with test case
```javascript
    fn: F

const files = preaddir('/');
```

### DIFF
--- a/types/node/fs.d.ts
+++ b/types/node/fs.d.ts
@@ -889,6 +889,12 @@ declare module "fs" {
          * @param options If called with `withFileTypes: true` the result data will be an array of Dirent
          */
         function __promisify__(path: PathLike, options: { encoding?: string | null; withFileTypes: true }): Promise<Dirent[]>;
+
+        /**
+         * Asynchronous readdir(3) - read a directory.
+         * @param path A path to a file. If a URL is provided, it must use the `file:` protocol.
+         */
+        function __promisify__(path: PathLike): Promise<string[]>;
     }
 
     /**


SimonSchick (changes requested):
Please add the test case, also see my previous comment.

JiangJie:
> But `function __promisify__(path: PathLike, options?: { encoding: BufferEncoding | null; withFileTypes?: false } | BufferEncoding | null): Promise<string[]>;` already covers this case?

Yes, if put this overload at the end of `readdir`, this case will pass, but other test cases will fail.
Refer to #37941.

JiangJie:
> Reopen #37861 with test case
> 
> ```js
> type UnPromisify<T> = T extends Promise<infer U> ? U : never;
> function putToResult<F extends (...args: any) => any>(
>     fn: F
> ): (...args: Parameters<F>) => Promise<{ result: UnPromisify<ReturnType<F>> }>;
> 
> const readdir = util.promisify(fs.readdir);
> const preaddir = putToResult(readdir);
> const files = preaddir('/');
> ```

@SimonSchick What you need.

JiangJie:
> This already works fine without your modification?
> 
> image
> 
> You also need to include the test case IN your PR.

Without my modification, this will fail.
image

Flarna:
Tests for fs module are currently in file types\node\node-tests.ts but feel free to move them into a dedicated file in subfolder tests.

This seems to be more a problem with the way how typescript handles overload resolution and not with the definitions here itself. Changing the sequence may solve your problem but raise problems for others.

@rbuckton Do you have any hints regarding overload resolution related to `__promisify__`?

PR #37948, opened 2019-08-27
add optional property `config` to StylelintPlugin options

arzyu:
Add some type definitions, so we can pass configuration to `stylelint` directly, like this:

```js
// webpack config
{
  plugins: [
    new StylelintPlugin({
      config: {
        extends: [
          "stylelint-config-recommended",
          "stylelint-config-css-modules"
        ],
        rules: {
          // ...
        }
      }
    })
  ]
}
```

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/webpack-contrib/stylelint-webpack-plugin#options>

### DIFF
--- a/types/stylelint-webpack-plugin/index.d.ts
+++ b/types/stylelint-webpack-plugin/index.d.ts
@@ -23,7 +23,17 @@ declare namespace StylelintWebpackPlugin {
 
     type Formatter = (messages: Message[], source: string) => string;
 
+    interface Config {
+        rules?: object;
+        extends?: string | string[];
+        plugins?: string[];
+        processors?: string[];
+        ignoreFiles?: string | string[];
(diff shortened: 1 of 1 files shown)


types/stylelint-webpack-plugin/index.d.ts
```diff
@@ -23,7 +23,17 @@ declare namespace StylelintWebpackPlugin {
 
     type Formatter = (messages: Message[], source: string) => string;
 
+    interface Config {
+        rules?: object;
+        extends?: string | string[];
+        plugins?: string[];
```
amcasey: I'm surprised this isn't `string | string[]`, but it matches the docs.

amcasey (approved):

PR #37957, opened 2019-08-27
react-syntax-highlighter: Update styles folder for v11.0.2

giuband:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/conorhastings/react-syntaax-highlighter/commit/115c16386384bc0dcdb49853b2865de7d074b361#diff-04c6e90faac2675aa89e2176d2eec7d8

### DIFF
--- a/types/react-syntax-highlighter/index.d.ts
+++ b/types/react-syntax-highlighter/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for react-syntax-highlighter 11.0.2
+// Type definitions for react-syntax-highlighter 11.0
 // Project: https://github.com/conorhastings/react-syntax-highlighter
 // Definitions by: Ivo Stratev <https://github.com/NoHomey>
 //                 Aimee Gamble-Milner <https://github.com/ajgamble-milner>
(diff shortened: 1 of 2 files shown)


guoyunhe (approved):
Good for me. Thanks for contributing and sorry for late review!

types/react-syntax-highlighter/index.d.ts
```diff
@@ -106,824 +106,824 @@ declare module "react-syntax-highlighter/prism" {
     > {}
 }
 
-declare module "react-syntax-highlighter/dist/styles/hljs" {
+declare module "react-syntax-highlighter/dist/esm/styles/hljs" {
```
rbuckton: This seems like a breaking change for anyone still using v10.

rbuckton (commented):
If there is a possibility that the v10 definitions may need to be changed in the future, I would recommend you create a `v10` folder with a copy of the 10.2 declarations per the instructions here: https://github.com/DefinitelyTyped/DefinitelyTyped#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package.

If you do not believe that there is a need to maintain a separate v10 version for future updates, let me know and I will merge this PR.

guoyunhe:
I think we probably won't have volunteers to maintain the types for v10. And the last v10 type is good enough for most usage. So I think we don't need to keep a v10 back up folder.

types/react-syntax-highlighter/index.d.ts
stof: actually, this was a mistake in v10 types AFAICT, as the changelog of the lib mentions that as a breaking change between 9 and 10.

and this is missing some of the updates, as languages also moved.

PR #37970, opened 2019-08-28
added definition for leaflet-loading

BePo65:
added definition for leaflet plugin leaflet-loading


rbuckton:
Thanks for the contribution!

PR #37973, opened 2019-08-28
new @types/flexmonster version update

Uaman:
report object extended with "version" and "creationDate" properties


rbuckton:
Thanks for the contribution!

PR #37987, opened 2019-08-28
update version matcher for openpgp version 3.2

woss:
fix the version matcher. 
this breaks if the version is 4

### DIFF
--- a/types/openpgp/package.json
+++ b/types/openpgp/package.json
@@ -2,7 +2,7 @@
     "private": true,
     "types": "index",
     "typesVersions": {
-        ">=3.2.0-0": {
+        "<=3.2.0-0": {
             "*": [
                 "ts3.2/*"
             ]


po5i (approved):
Looks good to me, Thanks!

woss:
don't get why travis is failing.

errietta:
@woss is it meant to be `<4`? What you're doing will match `<=3.2` and use the types meant for 3.2. I don't understand what it's really meant to be doing

woss:
if i understood correctly that field, it should say, if `<=3.2` then load the file under `ts3.2` folder since the `openpgp` version is at least 3.2 

In my case is version 4.6 and loaded file  should be `index.d.ts` 

Also, i've been working today and i couldn't get the `key.getKeyId()` to be a string, it seems that in the typings return value of that method is a [`string`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/openpgp/index.d.ts#L1786)  but in actual implementation is of [this](https://github.com/openpgpjs/openpgpjs/blob/master/src/type/keyid.js#L35)  signature 
```
/**
 * Parsing method for a key id
 * @param {Uint8Array} input Input to read the key id from
 */
Keyid.prototype.read = function(bytes) {
  this.bytes = util.Uint8Array_to_str(bytes.subarray(0, 8));
};

Keyid.prototype.write = function() {
  return util.str_to_Uint8Array(this.bytes);
};

Keyid.prototype.toHex = function() {
  return util.str_to_hex(this.bytes);
};
```

errietta:
@woss that's the typescript version, not the openpgp version. Probably these types are out of date and need changing for the latest one, I don't think changing package.json is the actual solution here

SardineFish:
Thanks for contribution.
Since the current type definition were generated from the `jsdoc` in souce code of `openpgp@4.4.10`, and corrected manually to fix issues with the generator. If some type definitions miss matched, you can open an issue or make a PR to correct it manually following the `jsdoc` in the source code of the latest version of `openpgp` but not the official documentation, since the official documentation is seems to be out date.

woss:
oh. i am terribly sorry for taking your time. This PR is wrong and let's close it.

I'll look into the OpenPGP and its docs. thanks, guys

PR #38008, opened 2019-08-29
add authors mention helper to issue template

kerolloz:
add a link to a simple website that helps to mention all the authors of a package. very helpful when there are many authors.


amcasey:
@kerolloz, thanks for your contribution!  Unfortunately, I don't think that's sufficiently general-purpose for inclusion in the main readme.

kerolloz:
@amcasey 
would you please explain why?

> I don't think that's sufficiently general-purpose for inclusion in the main readme.

- the main and only purpose of the website is to help you mention the authors.

minestarks:
@kerolloz I think the website is really nifty and it really feels a need - thanks for building it! But I agree about not adding the link to the issue template. I don't think it's a good idea to have links out from DefinitelyTyped to users' personal projects. It might appear as an endorsement, or even an extension of DefinitelyTyped, and we as the maintainers of DefinitelyTyped have a responsibility to the DT community to ensure the quality of the tools we're building and endorsing. At the moment we simply don't have the resources to commit to signing off the quality of an external website we don't have control of.

You did point out a real need though -- it is a pain to find and tag authors manually. Maybe @typescript-bot should go ahead and tag authors on issues automatically. I passed this on to the team members who work on it :)

PR #38011, opened 2019-08-29
add LDAPError and various LDAPError types, as well as update with project-wide linting rules

mscottnelson:
- [ x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/ldapjs/node-ldapjs/blob/master/lib/errors/index.js>>


types/ldapjs/index.d.ts
```diff
@@ -1,88 +1,81 @@
...
 // Definitions by: Charles Villemure <https://github.com/cvillemure>, Peter Kooijmans <https://github.com/peterkooijmans>, Pablo Moleri <https://github.com/pmoleri>, Michael Scott-Nelson <https://github.com/mscottnelson>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-
 /// <reference types="node" />
 
-import { EventEmitter } from "events";
+import { EventEmitter } from 'events';
+
+export class LDAPError extends Error {
+    name: 'LDAPError' | string;
```
minestarks: `'LDAPError' | string` is the same type as `string`, `80 | number` is the same type as `number`, and `any | LDAPError` is the same thing as `any`. In short, having the narrower type in the union isn't actually adding anything from a type checking perspective. To prevent confusion, it's best to simply use `string`, `number` and `any`.

types/ldapjs/index.d.ts
```diff
@@ -1,88 +1,81 @@
...
+    code: 80 | number;
+    message: string;
+    dn: string;
 
-export interface Error {
-	code: number;
-	name: string;
-	message: string;
+    new(message?: string, dn?: string, caller?: any | LDAPError): Error;
```
minestarks: did you mean `constructor` by any chance?

types/ldapjs/index.d.ts
```diff
@@ -258,142 +253,286 @@ export function createServer(options?: ServerOptions): Server;
...
+    baseObject: string;
+    scope: 'base' | 'one' | 'sub';
+    derefAliases: number;
+    sizeLimit: number;
+    timeLimit: number;
+    typesOnly: boolean;
+    filter: any;
+    attributes?: any;
+}
+export class OperationsError extends LDAPError {
```
minestarks: Could you add some tests for `LDAPError` and these new types?

minestarks (changes requested):

Issue #37999, opened 2019-08-28
[@types/prop-types] IsOptional should not care about nulls


davidje13:
Authors: @DovydasNavickas @ferdaber @eps1lon

The definition of IsOptional currently reads:

```ts
export type IsOptional<T> = undefined | null extends T ? true : undefined extends T ? true : null extends T ? true : false;
```

As far as I can tell, this was added long ago in ffc83511271dfdb1646e22e69be3b243cd5b8e26 but with no reasoning for the complexity of the check.

It breaks for me, because I [explicitly distinguish between `null` and `undefined`](https://github.com/davidje13/prop-types-nullable/blob/master/index.js) in my types (and I am not the only one; https://github.com/facebook/react/issues/3163). This is necessary to ensure correct JSON representations.

This means that when I define PropTypes for an object which can have a nullable (but not undefinable) field, `InferProps` incorrectly claims that the field can be undefined.

Example: I use code similar to the following to ensure the various representations of types in my project do not go out-of-sync:

```ts
import PropTypes from 'prop-types';
import nullable from 'prop-types-nullable';

/// prop-types-nullable declares the following:
// interface NullableRequireable<T> extends Validator<T | null> {
//   isRequired: Validator<Exclude<T | null, undefined>>;
// }
// function nullable<T>(
//   type: Validator<T>,
// ): NullableRequireable<T | undefined>;

interface MyType {
  requiredString: string;
  nullableString: string | null;
}

const propTypesMyType: PropTypes.Requireable<MyType> = PropTypes.shape({
  requiredString: PropTypes.string.isRequired,
  nullableString: nullable(PropTypes.string).isRequired,
});

// Type 'Requireable<InferProps<{ requiredString: Validator<string>; nullableString: Validator<string | null>; }>>' is not assignable to type 'Requireable<MyType>'.
//   Type 'InferProps<{ requiredString: Validator<string>; nullableString: Validator<string | null>; }>' is not assignable to type 'MyType'.
//     Types of property 'nullableString' are incompatible.
//       Type 'string | null | undefined' is not assignable to type 'string | null'.
//         Type 'undefined' is not assignable to type 'string | null'.
```

---

I believe the declaration could be changed to:

```ts
export type IsOptional<T> = undefined extends T ? true : false;
```

which is simpler and appears to be more correct.

eps1lon:
This is unfortunately how `prop-types` defines optional types (which originates from flow I think).

```jsx
[...]

PR #38015, opened 2019-08-29
[prop-types] Do not consider null without undefined as optional

davidje13:
Please fill in this template.

  - some failing tests remain in `airbnb-prop-types`; should I fix those in the same PR?

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37999

Authors: @DovydasNavickas @ferdaber @eps1lon

---

<s>This change works fine with my own code, though due to a bug (?) in the react types I haven't been able to run the formal test suite on my machine. "Error: Errors in typescript@next for external dependencies: ../react/index.d.ts(28,22): error TS2307: Cannot find module 'csstype'.". Hopefully the PR process will run tests on a CI machine without this issue.</s>

This both simplifies the `IsOptional` check and makes it behave better. Since the `Validator` type already includes `null | undefined` anyway, this won't make any difference to the built-in validators (they will still be considered optional unless marked as `isRequired`), but custom validators which consider `null` and `undefined` separately are now inferred correctly by `shape`.

<s>I would like to add a test for this use-case as well, but since I can't run the tests myself and I'm not familiar with this testing framework, I want the first commit to just ensure the existing tests do not break.</s>

### DIFF
--- a/types/prop-types/prop-types-tests.ts
+++ b/types/prop-types/prop-types-tests.ts
@@ -33,6 +33,8 @@ interface Props {
         baz?: any;
     };
     optionalNumber?: number | null;
+    nullableNumber: number | null,
+    undefinableNumber?: number;
     customProp?: typeof uniqueType;
     component: PropTypes.ReactComponentLike;
 }
@@ -75,6 +77,8 @@ const propTypes: PropTypesMap = {
     objectOf: PropTypes.objectOf(PropTypes.number.isRequired).isRequired,
     shape: PropTypes.shape(innerProps).isRequired,
     optionalNumber: PropTypes.number,
+    nullableNumber: (() => null) as PropTypes.Validator<number | null>,
+    undefinableNumber: (() => null) as PropTypes.Validator<number | undefined>,
     customProp: (() => null) as PropTypes.Validator<typeof uniqueType | undefined>,
     component: PropTypes.elementType.isRequired
 };
@@ -102,6 +106,8 @@ const propTypesWithoutAnnotation = {
     objectOf: PropTypes.objectOf(PropTypes.number.isRequired).isRequired,
     shape: PropTypes.shape(innerProps).isRequired,
     optionalNumber: PropTypes.number,
+    nullableNumber: (() => null) as PropTypes.Validator<number | null>,
+    undefinableNumber: (() => null) as PropTypes.Validator<number | undefined>,
     customProp: (() => null) as PropTypes.Validator<typeof uniqueType | undefined>,
     component: PropTypes.elementType.isRequired
 };
--- a/types/prop-types/prop-types-tests.ts
+++ b/types/prop-types/prop-types-tests.ts
@@ -33,7 +33,7 @@ interface Props {
         baz?: any;
     };
     optionalNumber?: number | null;
-    nullableNumber: number | null,
+    nullableNumber: number | null;
     undefinableNumber?: number;
     customProp?: typeof uniqueType;
     component: PropTypes.ReactComponentLike;
--- a/types/prop-types/prop-types-tests.ts
+++ b/types/prop-types/prop-types-tests.ts
@@ -20,7 +20,7 @@ interface Props {
     instanceOf: TestClass;
     oneOf: 'a' | 'b' | 'c';
     oneOfType: string | boolean | {
-        foo?: string;
+        foo?: string | null;
         bar: number;
     };
     numberOrFalse: false | number;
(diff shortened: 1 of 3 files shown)


davidje13:
Hmm, the remaining test failures are strange. I'm wondering if this is a typescript issue, or something I don't understand in how these types are being inferred;

If I explicitly cast the optional type validators as `PropTypes.Validator<(whatever) | undefined | null>`, they work, whereas `PropTypes.Requireable<(whatever)>` fails (somehow the `undefined` possibility is being lost). But `Requireable<T> extends Validator<T | undefined | null>`

Nothing seems to look for `extends Requireable` explicitly, so the above 2 cases should be identical. Any ideas?

davidje13:
OK so the tests are now all passing (apologies for the back-and-forth in the commit history; for the reasons explained in the first comment I haven't been able to run the tests properly on my local machine)

I had to switch to a `nominalTypeHack` style which didn't interfere with `undefined`. I notice that's an exported symbol; if anybody is using it from outside it will break, but should it really be exported?

I also added tests for nullable-only and undefinable-only validators, which was the original point here.

The remaining test failures are all in `airbnb-prop-types` which is making rather precise assertions on `Requireable<number | null>`, etc. which have now become `Requireable<number | null | undefined>`. Obviously the actual behaviour is unchanged, because `Requireable` has always added `null | undefined` to the type.

What's the usual procedure when changes are needed in a downstream project's tests? Do I just fix those tests in this PR too?

eps1lon (commented):
New types look way friendlier. Thanks :+1: I'll let @ferdaber decide if this is ok since he's is the original code owner.

@dt-bot should mention downstream package collaborators. If they don't respond then it should be ok.

types/prop-types/prop-types-tests.ts
```diff
@@ -20,7 +20,7 @@ interface Props {
     instanceOf: TestClass;
     oneOf: 'a' | 'b' | 'c';
     oneOfType: string | boolean | {
-        foo?: string;
+        foo?: string | null;
```
ferdaber: Why did this and `shape.bar` below it need to be changed to both be nullable?

ferdaber:
The changes in two of the test suites' types concern me that this would be a breaking change. I remember that there were TS devs who were strongly opinionated that they be able to define their props interfaces more precisely, but still have propTypes for JS users (e.g. library authors). So they want to be able to keep their types a bit more precise for consumption within their own codebase, but have prop types for consumers outside their codebase, if that makes sense.

types/prop-types/prop-types-tests.ts
davidje13: To be honest I'm not sure why this wasn't needed before. It came up as an error so I happily patched it, because `| null` is the more correct type (`PropTypes.string` would allow `null` through). I understand that it may not be desirable from a practical standpoint, but maybe the better way to resolve that is to remove `|null` from the `Requireable` type? At least that way it's implemented in a consistent way (notice that the root-level optional type *already* had to include `null`; the change now is that nested types match the same behaviour)

davidje13:
@ferdaber that makes sense; see my response to your inline comment above.

The problem with the old code is that it's inconsistent; the top-level `optionalNumber` already needed `| null`, so it's strange that the sub-properties `foo` and `bar` didn't, when they're defined the same way.

I don't know why that inconsistency happened in the old code (I can't see any logical reason for it in the types, and my intuition would be that `| null` would be needed).

There's no great way to relax this while also maintaining correctness for those users who want correctness. Presumably the most common issue would be types marked as optional where the library author doesn't want to bother considering the `null` case. That could be the source of a bug (if they guard against `undefined` explicitly, `null`s would slip through), but most of the time guards will be against falsey values, so it just adds noise to the type.

One option might be to introduce a `LaxNulls<T extends Validator>` or similar, which has the sole purpose of removing `| null` from all properties in `T`. Another option is to remove `| null` from `Requireable` (though that will negatively affect those who want correct types).

ferdaber:
The most important functionality is to preserve type inference for JavaScript users, as all of these revolve around the `LibraryManagedAttributes` interface in the `JSX` namespace. As long as, for a JS user, the correct types are inferred via prop types, I'm fine with whatever changes— just be wary that you may have some angry TS users coming at you after this change goes through 😜

ferdaber (approved):

davidje13:
Yes I expect that even this small change in a popular library like prop-types will generate some anger (sadly the semver strategy doesn't work for DefinitelyTyped packages)

It is change in the right direction though (i.e. encourages safer, more correct code); maybe somebody else can suggest a better way to allow lax null checking on an opt-in basis.

Also worth noting that this only affects those with StrictNullChecks enabled in tsconfig (which presumably biases towards developers who care about nullability anyway), and those who have nested properties (since top-level properties already did this).

rbuckton:
Thanks for the contribution!

PR #38016, opened 2019-08-29
Fix testing-library__cypress by exposing @testing-library/dom#configure function

sebinsua:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/testing-library/cypress-testing-library/pull/81#issuecomment-526326736

### DIFF
--- a/types/testing-library__cypress/tslint.json
+++ b/types/testing-library__cypress/tslint.json
@@ -1,10 +1,6 @@
 {
     "extends": "dtslint/dt.json",
     "rules": {
-        "semicolon": [
-            true,
-            "never"
-        ],
         "whitespace": [
             false
         ],
--- a/types/testing-library__cypress/index.d.ts
+++ b/types/testing-library__cypress/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for @testing-library/cypress 4.1
+// Type definitions for @testing-library/cypress 4.2
 // Project: https://github.com/testing-library/cypress-testing-library
 // Definitions by: Aaron Mc Adam <https://github.com/aaronmcadam>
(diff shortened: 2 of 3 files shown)


types/testing-library__cypress/tslint.json
```diff
@@ -1,10 +1,6 @@
 {
     "extends": "dtslint/dt.json",
     "rules": {
-        "semicolon": [
-            true,
-            "never"
-        ],
```
sebinsua: Apologies for removing the lint rule that disables semicolons, but since [the 14th of June](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35672 ) `prettier` has been automatically formatting changes to include semicolons (so subsequently it would fail the lint rules that had been setup) and I didn't think it was a good idea to have a lint rule which fought the global configuration of `prettier`.

kentcdodds (approved):
LGTM

kentcdodds (approved):

types/testing-library__cypress/tslint.json
airato: sounds good! just one minor non-blocking comment - would be great to have the prettier config change and prettifying the code as 2 separate commits (and maybe in separate PR). And the `configure` export change on top of that. This would make the reviewing faster and easier 👍

airato (approved):

types/testing-library__cypress/tslint.json
sebinsua: @airato It's two separate commits within this PR now.

airato (approved):

types/testing-library__cypress/tslint.json
airato: thank you!

PR #38017, opened 2019-08-29
[opossum] v4.0 updates.

mastermatt:
The lib now exposes a constructor directly, so the export has been
changed to use the `export =` with `declare` style instead of `export default`.

- `promisify` removed.
- References to `hystrix` were removed.
- The primary class now has generics that pulls the argument and return
  types from the constructor then passes them to the result of `fire`
  and some of the event listeners.
- Event listeners were added to give users types in the specific callbacks.
- Doc updates for the methods.
- The `action` function must now return a Promise.
-----
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/nodeshift/opossum/blob/master/CHANGELOG.md>

### DIFF
--- a/types/opossum/index.d.ts
+++ b/types/opossum/index.d.ts
@@ -1,20 +1,17 @@
-// Type definitions for opossum 1.10
+// Type definitions for opossum 4.0
 // Project: https://github.com/nodeshift/opossum, https://nodeshift.dev/opossum
 // Definitions by: Quinn Langille <https://github.com/quinnlangille>
 //                 Willy Zhang <https://github.com/merufm>
 //                 Lance Ball <https://github.com/lance>
 //                 Matt R. Wilson <https://github.com/mastermatt>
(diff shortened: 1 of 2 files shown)


rbuckton:
Is there a possibility that updates will be needed for the v1 typings for this package? If so you should add a `v1` directory with the existing version of the types as mentioned [here](https://github.com/DefinitelyTyped/DefinitelyTyped#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package). If not, let me know and I can merge this.

mastermatt:
@rbuckton v1 isn't being maintained anymore, so I don't see why we'd ever need to update the types for them.

lance (approved):
@mastermatt thank you for maintaining these!

mastermatt:
@rbuckton anything else you think this PR needs?

PR #38028, opened 2019-08-30
Reference the toplevel index.d.ts from the extensions.

dirk-olmes:
cometd: fix imports of the extensions 

When using an extension in code the old 'import ..' does not work. Other modules explicitly import ../index which is what should be used here, too.

### DIFF
--- a/types/cometd/AckExtension/index.d.ts
+++ b/types/cometd/AckExtension/index.d.ts
@@ -1,4 +1,4 @@
-import * as m from '..';
+import * as m from '../index';
 
 declare class AckExtension implements m.Extension {
     constructor();
--- a/types/cometd/BinaryExtension/index.d.ts
+++ b/types/cometd/BinaryExtension/index.d.ts
@@ -1,4 +1,4 @@
-import * as m from '..';
+import * as m from '../index';
 
 declare class BinaryExtension implements m.Extension {
     constructor();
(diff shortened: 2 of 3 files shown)


amcasey:
@dirk-olmes I'm not sure I understand - what was the issue with the old imports?

dirk-olmes:
> @dirk-olmes I'm not sure I understand - what was the issue with the old imports?

Using the following import `import AckExtension from 'cometd/AckExtension';` I get a tsc compile error `import AckExtension from 'cometd/AckExtension';`.

I looked around how other typedefs handle imports from a subdir to the parent definition and they do it in suggested way.

hagl:
In our codebase we use exactly the same import without problems.

@dirk-olmes Could this StackOverflow post solve your import issue: https://stackoverflow.com/questions/44028806/should-the-index-ts-be-resolved-by-typescript-as-default-module-file

minestarks (changes requested):
Agree that this is probably an issue with the end user's configuration rather than an issue with the typings themselves. Marking "request changes" to prevent merging by accident

dirk-olmes:
> In our codebase we use exactly the same import without problems.
> 
> @dirk-olmes Could this StackOverflow post solve your import issue: https://stackoverflow.com/questions/44028806/should-the-index-ts-be-resolved-by-typescript-as-default-module-file

The main difference between your code base and mine is the module system that's used. The SO article suggests using `commonjs` but my project is set to `es2015`.

I'll try to put together something to dig deeper but that won't happen right now as I am super busy. Stay tuned ...

PR #38032, opened 2019-08-30
[@types/tabulator-tables] copyToClipboard should accept 'selected' in…

mstlaurent-paradigm:
…stead of 'selection', as well as undefined

Please fill in this template.

* 'selected' selector: https://github.com/olifolkerd/tabulator/blob/cb39652ac7fcc0063eb23cecfe303a439f8ca4e4/dist/js/modules/clipboard.js#L863
* 'undefined' parameter for copyToClipboard: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/13ad568eaaa359a9f4586b695229a1738ed8a189/types/tabulator-tables/index.d.ts#L1606

### DIFF
--- a/types/tabulator-tables/index.d.ts
+++ b/types/tabulator-tables/index.d.ts
@@ -1604,7 +1604,7 @@ declare class Tabulator {
     The first argument is the copy selector, you can choose from any of the built in options or pass a function in to the argument, that must return the selected row components.
 
     If you leave this argument undefined, Tabulator will use the value of the clipboardCopySelector property, which has a default value of table */
-    copyToClipboard: (type: 'selection' | 'table') => void;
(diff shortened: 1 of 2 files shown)


Jojoshua (changes requested):
Ok. I was looking at the hardcoded example in the documentation here which apparently was invalid. http://tabulator.info/docs/4.3/clipboard#copy

However, lets include all the type options and instead of including undefined as an option, just make type optional. Something like this would work.

copyToClipboard: (type?: 'selected' | 'table' | 'active') => void;

mstlaurent-paradigm:
Changes pushed! I will open an issue in Tabulator about their documentation.

Jojoshua (approved):
Thanks!

PR #38042, opened 2019-08-30
Add type declarations for [bristol] logger

ElliottCampbellJHA:

If adding a new definition:

### DIFF
--- a/types/bristol/bristol-tests.ts
+++ b/types/bristol/bristol-tests.ts
@@ -1,0 +1,19 @@
+import * as bristol from 'bristol';
+
+const log = new bristol.Bristol();
+
+log.addTarget('console');
+log.info('Things be working.', {});
+
+log.addTarget('console').withFormatter('human');
+log.debug('Hello, world', { code: 404 });
+
+const uhoh: bristol.LogData = {
+  code: 500,
+  error: {
+    message: 'Something broke',
+    reason: 'idk why though',
+  },
+};
+
+log.error('Something went wrong', uhoh);
--- a/types/bristol/bristol-tests.ts
+++ b/types/bristol/bristol-tests.ts
@@ -1,14 +1,12 @@
-import * as bristol from 'bristol';
+import bristol = require('bristol');
 
-const log = new bristol.Bristol();
+bristol.addTarget('console');
+bristol.info('Things be working.', {});
 
-log.addTarget('console');
-log.info('Things be working.', {});
+bristol.addTarget('console').withFormatter('human');
+bristol.debug('Hello, world', { code: 404 });
 
-log.addTarget('console').withFormatter('human');
-log.debug('Hello, world', { code: 404 });
-
-const uhoh: bristol.LogData = {
+const uhoh = {
   code: 500,
   error: {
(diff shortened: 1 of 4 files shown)


amcasey (changes requested):
Unfortunately, the underlying library has complicated exports that can't be perfectly captured in a .d.ts file.  In particular, the module it exports is effectively both a class instance and a namespace.  Of the two, the former is arguably more important (e.g. see the sample usage on the GH page).  To make that work, you probably want to remove your existing `export` keywords, `declare const bristol: Bristol;`, and then `export = bristol;`.  Then `import bristol = require('bristol');` will work as expected, except that `bristol.Bristol` and `bristol.Bristol.logUtil` will be unavailable.

If you'd like to get fancier and build a better approximation, let me know.

eheikes:
@amcasey I see what you're saying. Unfortunately, for our situation we need/use the `Bristol` constructor rather than the pre-constructed instance. But I also agree that the class instance is probably more common, so we can modify the PR for that case if it's not possible to include both.

ElliottCampbellJHA:
I think I'd rather not get too fancy at the moment I've uploaded the suggested changes.

amcasey:
@eheikes I think supporting the common case is the key consideration, so I'm inclined to merge this version as is (thanks for the changes @ElliottCampbellJHA !).  However, I believe a more complicated version could address your use case as well and I'd be happy to work with you to create a new PR.

eheikes:
I'm fine with merging this PR for now.

> However, I believe a more complicated version could address your use case as well and I'd be happy to work with you to create a new PR.

That would be great! How do you want to do that -- do you want to start work on it, or should we give it a shot?

amcasey (approved):

amcasey:
@heikes, cool, let me draft a skeleton for you and you can flesh it out in a separate PR.

amcasey:
@eheikes, here you go:
```ts
import logUtil2 = require("./logUtil");

interface LogError {
    message?: string;
    reason?: any;
    stack?: any;
}

interface LogData {
    code?: number;
    id?: string;
    path?: string;
    error?: LogError;
    data?: any;
}

export class Bristol {
    addTarget(target: any, opts?: any): any;
    withFormatter(formatter: string): any;
    withLowestSeverity(severity: string): any;
    info(message: string, data: LogData): any;
    warn(message: string, data: LogData): any;
    error(message: string, data: LogData): any;
    debug(message: string, data: LogData): any;
}

export namespace Bristol {
    const logUtil: typeof logUtil2;
}

export function addTarget(target: any, opts?: any): any;
export function addTarget(target: any, opts?: any): any;
export function withFormatter(formatter: string): any;
export function withLowestSeverity(severity: string): any;
export function info(message: string, data: LogData): any;
export function warn(message: string, data: LogData): any;
export function error(message: string, data: LogData): any;
export function debug(message: string, data: LogData): any;
```

Two caveats:
 1. The module object won't actually be an instance of the `Bristol` class for nominal typing purposes
 2. All members of the class are declared twice

I'll ask around in case there's a clever way to eliminate the duplication, but I think we're stuck with the first restriction.

PR #38049, opened 2019-08-31
stripe checkout.sessions.create 

### LINKED ISSUE #685: Type error while adding to subscription_data to a session.
When creating a new session for a subscription plan I get a static error on items:

Type '{ plan: string; }[]' is missing the following properties from type 'IList<ISubscriptionItem>': object, data, has_more, url;

index.d.ts(6467, 13): The expected type comes from property 'items' which is declared here on type 'ISubscription'
Code below:

        import * as Stripe from 'stripe';
        const stripe = new Stripe(process.env.STRIPE_SECRET_KEY);

          stripe.checkout.sessions.create(
            {
              success_url: 'https://example.com/success',
              cancel_url: 'https://example.com/cancel',
              payment_method_types: ['card'],
              customer_email: dummy@example.com 
              subscription_data: {
                 items: [ {plan: subcriptionInfo.name,}],
             },
            function(err, session) {
              // asynchronously called
            }
          );

I was able to get around the error by going to "interface ICheckoutCreationOptions" and changing line 2108 to:
subscription_data?: subscriptions.ISubscription | any;

Is there another way to do this?

cdock1029:
Fix stripe/stripe-node#685

### DIFF
--- a/types/stripe/index.d.ts
+++ b/types/stripe/index.d.ts
@@ -2158,7 +2158,7 @@ declare namespace Stripe {
                 /**
                  * Use instead of @param line_items when using a subscription
                  */
-                subscription_data?: Partial<subscriptions.ISubscription>;
+                subscription_data?: subscriptions.ISubscriptionCustCreationOptions;
             }
 
             interface ICheckoutLineItems {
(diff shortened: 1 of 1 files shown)


pushplay (approved):
This looks right.  In the future please follow the PR template.  Linking to relevant documentation saves a bit of reviewing work.

cdock1029:
@pushplay ok will do, sorry about that. Thanks for approving it

PR #37212, opened 2019-07-29
Add express-boom typings

nox-404:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If adding a new definition:
- [x] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration`
- [x] If this is for an NPM package, match the name. If not, do not conflict with the name of an NPM package.
- [ ] Create it with `dts-gen --dt`, not by basing it on an existing project.
- [x] `tslint.json` should be present, and `tsconfig.json` should have `noImplicitAny`, `noImplicitThis`, `strictNullChecks`, and `strictFunctionTypes` set to `true`.

Notes: Based on Sandeep K Nair gist: https://gist.github.com/sandeepsuvit/2486c99c0346db87de24539472f34451

PR #38056, opened 2019-09-01
Add express-boom typings

nox-404:
Typings for express-boom again :)

If adding a new definition:

Notes: Based on Sandeep K Nair gist: https://gist.github.com/sandeepsuvit/2486c99c0346db87de24539472f34451
Reopened from #37212 (was away sorry)


types/express-boom/index.d.ts
```diff
@@ -0,0 +1,353 @@
...
+// Definitions by: Aurelien Gaston <https://github.com/Nox-404>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
+
+/**
+ * Type definition for the middleware itself
+ */
+
+import express = require('express');
+declare namespace e { }
```
amcasey: What's the effect of merging in this empty namespace?

types/express-boom/index.d.ts
```diff
@@ -0,0 +1,353 @@
...
+ * Type augmentation for express
+ */
+
+/**
+ * Based on Sandeep K Nair gist: https://gist.github.com/sandeepsuvit/2486c99c0346db87de24539472f34451
+ * And https://github.com/hapijs/boom
+ */
+declare global {
+  namespace Express {
+    interface Boom {
```
amcasey: I feel like I'm missing something obvious, but how are these types intended to be used?  Perhaps some example code could be added as a test?

amcasey (changes requested):
A couple of questions.

types/express-boom/index.d.ts
Nox-404: These types augment express typings. When using express boom middleware, res gets augmented with the boom property.
There is an example already.

amcasey: It looks like the augmentation is adding the `boom` property `Express.Response` - is that correct?  The other types just describe the shape of the new property?

It might be nice to test more than one member (i.e. `forbidden`) of the augmentation.

Nox-404:
I'll do that before Tuesday then

types/express-boom/index.d.ts
Nox-404: @amcasey That's exactly what you said. I added a few other examples.

types/express-boom/index.d.ts
amcasey: As far as I know, this does nothing, so I don't strongly object to leaving it.

types/express-boom/index.d.ts
amcasey: Thanks!

amcasey (approved):

types/express-boom/index.d.ts
Nox-404: I removed it, it seems to still does the trick so better getting rid of it !

amcasey: I didn't see this before I merged.  If you tag me on a follow-up PR, I'll fast-track it.

PR #38058, opened 2019-09-01
Change order of index signature - express-session

bondz:
Before this change, I was getting

```ts
Duplicate string index signature.

[key: string]: any;
```

This fixes that.

### DIFF
--- a/types/express-session/index.d.ts
+++ b/types/express-session/index.d.ts
@@ -20,8 +20,8 @@ declare global {
     }
 
     interface SessionData {
-      [key: string]: any;
       cookie: SessionCookieData;
+      [key: string]: any;
     }
 
     interface SessionCookieData {


johnanisere (approved):

amcasey:
@bondz I'm not sure I follow - are you saying that the order of interface member declarations determines whether or not you get a compilation error?

bondz:
@amcasey yes. I get this error. 
image

But not when I change the order.

bondz:
Hmm, actually even when I restart tsc with this change the error comes back so I don't think this completely fixes it.

When I run

```
tsc -w
```

I get that error. If I change the order in node_modules directly, the error goes away until I restart vsCode. I think this is a different problem due to the type resolution.

amcasey:
@bondz FWIW, when I install `@types/connect-redis`, I don't see `@types/express-session` within it.  Same for the underlying packages.

bondz:
@amcasey, I realise that now... Had to remove my `yarn.lock` file to get the dependencies to install correctly. Thanks :)

PR #38063, opened 2019-09-02
Add types for lightpick.

akwiatek:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/lightpick/index.d.ts
+++ b/types/lightpick/index.d.ts
@@ -4,14 +4,17 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.4
 
-import * as moment from 'moment';
+import moment = require('moment');
+
+export = Lightpick;
+export as namespace Lightpick;
 
 declare class Lightpick {
     constructor(options: Lightpick.Options);
 
     gotoToday(): void;
 
-    gotoDate(date: Lightpick.InputDate): void;
+    gotoDate(date?: Lightpick.InputDate): void;
 
     gotoMonth(month: number): void;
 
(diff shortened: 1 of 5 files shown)


types/lightpick/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
```
amcasey: It looks like the underlying library is UMD, so you might want to have a look at this template: https://github.com/microsoft/TypeScript-Handbook/blob/master/pages/declaration%20files/templates/module-class.d.ts.md

types/lightpick/index.d.ts
```diff
@@ -0,0 +1,384 @@
+// Type definitions for Lightpick 1.3
+// Project: https://wakirin.github.io/Lightpick
+// Definitions by: Adam Kwiatek <https://github.com/akwiatek>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.4
+
+import * as moment from 'moment';
```
amcasey: Generally speaking, you can't instantiate a symbol imported in this way.  Maybe you meant `import moment = require('moment');`?

amcasey (changes requested):

types/lightpick/index.d.ts
kwiatek-kainos: Done.

types/lightpick/tsconfig.json
kwiatek-kainos: Thanks. I've made adjustments according to that template and type definitions for Lodash and jQuery.

I'm not entirely sure though, why you put a comment against tsconfig.json . Should I change anything here? Configs for other UMD types look similar...

amcasey: Sorry, I blanked.  I was thinking that the module type should be "umd", but that doesn't matter for .d.ts files.

amcasey (approved):
Thanks for the changes!

PR #38067, opened 2019-09-02
add cloudevents-sdk/v03 types

andycmaj:
Typescript types for [`cloudevents-sdk/v03`](https://github.com/cloudevents/sdk-javascript).

# Notes

* only `v03` is supported here, since that's the recommended version in cloudevents-sdk docs.
* `no-declare-current-package` and `no-single-declare-module` are ignored. i did try to break out the `/v03` module into a `v03.d.ts` file, but VSCode seemed to understand better when everything was in `index.d.ts`. Also, [recompose types](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/362a5990fa39dc7981746fd5f32f5145795a228d/types/recompose/index.d.ts#L415) use the same pattern.

# Checklist

If adding a new definition:

### DIFF
--- a/types/cloudevents-sdk/index.d.ts
+++ b/types/cloudevents-sdk/index.d.ts
@@ -5,6 +5,17 @@
 
 // tslint:disable-next-line no-declare-current-package no-single-declare-module cloudevents-sdk/v03 is the documented way to import the cloudevent builder
 declare module 'cloudevents-sdk/v03' {
+    // v03 CloudEvent payload
+    // https://github.com/cloudevents/sdk-javascript/blob/master/lib/specs/spec_0_3.js#L11
+    interface CloudEvent {
+        id: string;
+        datacontenttype: 'application/json';
+        time: string;
+        type: string;
+        source: string;
+        data: {};
+    }
+
     // Cloudevent fluent builder
     // https://github.com/cloudevents/sdk-javascript/blob/master/lib/cloudevent.js
     interface CloudEventBuilder {
@@ -21,20 +32,5 @@ declare module 'cloudevents-sdk/v03' {
 
     // Cloudevent.event() static constructor
     // https://github.com/cloudevents/sdk-javascript#an-easy-way-to-create-events
-    const Cloudevent: {
-        event: () => CloudEventBuilder;
-    };
-
-    export default Cloudevent;
-}
-
-// v03 CloudEvent payload
-// https://github.com/cloudevents/sdk-javascript/blob/master/lib/specs/spec_0_3.js#L11
-interface CloudEvent {
-    id: string;
-    datacontenttype: 'application/json';
-    time: string;
-    type: string;
-    source: string;
-    data: {};
+    function event(): CloudEventBuilder;
 }
--- a/types/cloudevents-sdk/index.d.ts
+++ b/types/cloudevents-sdk/index.d.ts
@@ -32,7 +32,5 @@ declare module 'cloudevents-sdk/v03' {
 
     // Cloudevent.event() static constructor
     // https://github.com/cloudevents/sdk-javascript#an-easy-way-to-create-events
-    export default {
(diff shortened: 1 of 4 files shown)


types/cloudevents-sdk/index.d.ts
```diff
@@ -0,0 +1,40 @@
...
+    const Cloudevent: {
+        event: () => CloudEventBuilder;
+    };
+
+    export default Cloudevent;
+}
+
+// v03 CloudEvent payload
+// https://github.com/cloudevents/sdk-javascript/blob/master/lib/specs/spec_0_3.js#L11
+interface CloudEvent {
```
rbuckton: This introduces an interface in the global scope. It seems like this should be included in the 'cloudevents-sdk/v03' module, above.

types/cloudevents-sdk/index.d.ts
```diff
@@ -0,0 +1,40 @@
...
+        toString: () => string;
+    }
+
+    // Cloudevent.event() static constructor
+    // https://github.com/cloudevents/sdk-javascript#an-easy-way-to-create-events
+    const Cloudevent: {
+        event: () => CloudEventBuilder;
+    };
+
+    export default Cloudevent;
```
rbuckton: [This is not how the package is exported](https://github.com/cloudevents/sdk-javascript/blob/0ac40bc5b0e9c68b0686cff9973cf6501dea8d62/v03/index.js#L19-L27), as it does not use `exports.default = `.

The best way to define exports for this package is to replace L24-L28 with just the following:

```ts
    function event(): CloudEventBuilder;
```

This way, the `CloudEventBuilder` interface (and `CloudEvent` interface once it is moved into the module), can all be imported via `import { } from 'cloudevents-sdk/v03';`.

rbuckton (changes requested):

types/cloudevents-sdk/index.d.ts
andycmaj: ah so the idea is to mirror the exact import you'd have to use when using the module's exports. that way the types become available with the imports you're already using? 

is that the idea?

andycmaj: so typescript compilation failed with:

```bash
Error in cloudevents-sdk
Error: /Workspace/github/DefinitelyTyped/types/cloudevents-sdk/cloudevents-sdk-tests.ts:1:8
ERROR: 1:8  expect  TypeScript@next compile error:
Module '"cloudevents-sdk/v03"' has no default export.

    at /Workspace/github/DefinitelyTyped/node_modules/dtslint/bin/index.js:190:19
    at fulfilled (/Workspace/github/DefinitelyTyped/node_modules/dtslint/bin/index.js:5:58)
Error: The following packages had errors: cloudevents-sdk
    at doRunTests (/Workspace/github/DefinitelyTyped/node_modules/types-publisher/src/tester/test-runner.ts:240:11)
    at process._tickCallback (internal/process/next_tick.js:68:7)
npm ERR! Test failed.  See above for more details.
```

But as i understand, using `allowSyntheticDefaultImports` 'fixes' this warning. would you recommend using that in my `tsconfig.json`? There DO seem to be a number of other DT packages that are using this compiler option... so i've submitted that change here.

andycmaj:
@rbuckton should i squash before merging? didn't see any specific advice about that in the contrib section of readme...

andycmaj:
@rbuckton this should be ready for a re-review... thanks! appreciate your time.

andycmaj:
ping... anyone? @rbuckton

andycmaj:
new year, new ping :) can i get someone to take another look at this?

rbuckton (approved):

kostap13:
@rbuckton Seems like this commit https://github.com/DefinitelyTyped/DefinitelyTyped/commit/540cea2e2bdae0ee5bcd110952cdb5cedd76aa37
broke builds

PR #38096, opened 2019-09-03
@types/ssh2: Added shell() "env" option

cowwoc:
I copied the existing "env" definition from the `ExecOptions` interface into `ShellOptions`. I did not add any tests as the former was not tested either and I am new to TypeScript.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/mscdex/ssh2/blob/master/lib/client.js#L823

### DIFF
--- a/types/ssh2/index.d.ts
+++ b/types/ssh2/index.d.ts
@@ -494,7 +494,7 @@ export interface ExecOptions {
 }
 
 export interface ShellOptions {
-    /** An environment to use for the execution of the command. */
+    /** An environment to use for the execution of the shell. */
     env?: any;
     /** Set either to `true` to use defaults, a number to specify a specific screen number, or an object containing x11 settings. */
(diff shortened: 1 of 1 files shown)


types/ssh2/index.d.ts
```diff
@@ -494,6 +494,8 @@ export interface ExecOptions {
 }
 
 export interface ShellOptions {
+    /** An environment to use for the execution of the shell. */
+    env?: any;
```
wrboyce: I think this should be

```suggestion
    env: NodeJS.ProcessEnv;
```

cowwoc: Fixed.

cowwoc: Fixed.

wrboyce (approved):

rbuckton:
Thanks for the contribution!

PR #38105, opened 2019-09-03
fabric event types

cancerberoSgx:
fabricjs types

 * better types for events

I have more changes to propose, but first I want to know if you agree with this. Thanks

### DIFF
--- a/types/fabric/fabric-impl.d.ts
+++ b/types/fabric/fabric-impl.d.ts
@@ -142,15 +142,79 @@ interface IDataURLOptions {
 	withoutShadow?: boolean;
 }
 
+/**
+ * A general Event type that represents all events mixed in a single type.
+ */
 interface IEvent {
-	e: Event;
+  /**
+   * The native Event instance, for example, a DOM MouseEvent if the user selected or 
+   * modified fabric.js objects using the mouse in the browser. 
+   */
+  e: Event;
 	target?: Object;
-    subTargets?: Object[],
+  subTargets?: Object[];
 	button?: number;
+  selected?: Object[];
 	isClick?: boolean;
+  deselected: Object[];
 	pointer?: Point;
 	absolutePointer?: Point;
-    transform?: { corner: string, original: Object, originX: string, originY: string, width: number };
+  transform?: { corner: string, original: Object, originX: string, originY: string, width: number };
+}
+
+/**
+ * Emitted when an existing selection is cleared. Any selection with one or more elements are considered. 
+ */
+interface ISelectionClearedEvent extends IEvent {
+  /** 
+   * Elements in the cleared selection. 
+   */
+  deselected: Object[];
+}
+
+/**
+ * Emitted when an existing selection is cleared.
+ * @deprecated this event is deprecated. use selection:created
+ */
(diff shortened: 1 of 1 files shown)


cancerberoSgx:
@typescript-bot Got it give me time and I will fix the linter errors. I couldn't run the tests nor in mac-os or linux with latest node/npm/yarn so don't get too exigent please. Probably I have to downgrade node.js or use docker ? Could you please tolerate this opened PR for the following 1 or 2 weeks ?

This is my nth time trying to collaborate with this project and I don't want to fail, again. I often author TypeScript declaration for libraries that are important for me and Always ended maintaining them in my own repos since pushing here is too time consuming .  these are a couple of examples: 

 * opencv.js : https://github.com/cancerberoSgx/mirada/tree/master/mirada/src/types/opencv - https://cancerberosgx.github.io/demos/mirada-opencv-api-html/modules/_types_opencv_mat_.html
 * [blessed](https://cancerberosgx.github.io/demos/accursed/api/accursed/modules/_blessedtypes_.html) 

Thanks

NotWoods:
Hi @cancerberoSgx! For now you might want to push another work-in-progress commit to make the bot happy. If the PR does end up getting closed you can probably just re-open it.

The failing issues are all just due to lint errors. It seems that some lines you added have extra unneeded spaces at the end. The travis logs point out the lines with issues.

Screen Shot 2019-09-07 at 8 40 24 PM

types/fabric/fabric-impl.d.ts
```diff
@@ -291,13 +355,33 @@ interface ICollection<T> {
 }
 
 interface IObservable<T> {
+  /**
+	 * Observes the selection of one or more objects.
+	 */
+	on(eventName: 'selection:cleared', handler: (e: ISelectionClearedEvent) => void): T;
```
NotWoods: I think these events should probably be on `Canvas` which extends `IObservable`.

types/fabric/fabric-impl.d.ts
```diff
@@ -142,15 +142,79 @@ interface IDataURLOptions {
...
+   * modified fabric.js objects using the mouse in the browser. 
+   */
+  e: Event;
 	target?: Object;
-    subTargets?: Object[],
+  subTargets?: Object[];
 	button?: number;
+  selected?: Object[];
 	isClick?: boolean;
+  deselected: Object[];
```
NotWoods: Since these properties aren't on every single event, they should only be present in the subclasses you wrote.

If you have an object that's typed as an `IEvent` and need a property from a subclass you would cast it instead.
```ts
let event: IEvent = ...
(event as ISelectionClearedEvent).deselected
```

types/fabric/fabric-impl.d.ts
```diff
@@ -142,15 +142,79 @@ interface IDataURLOptions {
 	withoutShadow?: boolean;
 }
 
+/**
+ * A general Event type that represents all events mixed in a single type.
```
NotWoods: The extra comments are great, thank you!

NotWoods (changes requested):

PR #38106, opened 2019-09-03
[react-datepicker@2.9.4] formatWeekday parameter is a string, not a date object

enagy27:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/Hacker0x01/react-datepicker/blob/de392f68bb08f63676edbbb0f28f17f0595c44a7/src/date_utils.js#L346

### DIFF
--- a/types/react-datepicker/index.d.ts
+++ b/types/react-datepicker/index.d.ts
@@ -50,7 +50,7 @@ export interface ReactDatePickerProps {
 	filterDate?(date: Date): boolean;
 	fixedHeight?: boolean;
 	forceShowMonthNavigation?: boolean;
-	formatWeekDay?(date: Date): string;
+	formatWeekDay?(formattedDate: string): string;
 	formatWeekNumber?(date: Date): string | number;
 	highlightDates?: Array<HighlightDates|Date>;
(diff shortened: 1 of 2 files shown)


Ky6uk (approved):

minestarks (changes requested):
I'm not seeing how the source link shows this parameter is a string, it still seems like `Date` to me. Any reason why this function is singled out?

enagy27:
@minestarks yes, I believe it's a `string`, and in my use of `react-datepicker` it looks like it's been passing me a `string`. The `formatWeekDay` function is the `formatFunc` in the provided snippet and it is supplied the result of `date-fns`'s `format` function (via `formatDate`), which will in turn provide a `string` (https://github.com/date-fns/date-fns/blob/ff7b6e1b15fa19e3b5613b1fb68c93abb7d99a1d/src/format/index.js#L314)

minestarks:
@enagy27 , I see it now, my bad. Thanks for the additional pointer!

minestarks (approved):

enagy27:
@minestarks absolutely! Thanks for your review! 😄

PR #38119, opened 2019-09-04
Add types for audio-play and audio-context

bdjeffyp:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/audio-context/index.d.ts
+++ b/types/audio-context/index.d.ts
@@ -1,0 +1,29 @@
+// Type definitions for audio-context 1.0
+// Project: https://github.com/audiojs/audio-context
+// Definitions by: Jeff Peterson <https://github.com/bdjeffyp>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/**
+ * Options for your audio context:
+@param sampleRate if specified, will set the context sampleRate.
+@param latencyHint if specified, will control latency. One of 'balanced', 'playback', 'interaction' (default) or number.
+@param offline if specified, will create OfflineAudioContext.
+@param length if specified, will set number of frames for offline context.
+@param channels if specified, will set number of channels for offline context.
+@param contextAttributes any other options for the context.
+ */
+export interface Options {
+    sampleRate?: number;
+    latencyHint?: string | number;
+    offline?: boolean;
+    length?: number;
+    channels?: number;
+    contextAttributes?: object;
+}
+
+/**
+ * Gets an audio context from your web browser.
+ * @param options Takes an Options object or just provide a sample rate.
+ * @returns the audio context or null if there was an error or not a web browser.
+ */
+export function getContext(options?: Options | number): AudioContext | null;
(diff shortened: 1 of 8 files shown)


types/audio-context/index.d.ts
```diff
@@ -0,0 +1,30 @@
...
+    channels?: number;
+    contextAttributes?: object;
+}
+
+/**
+ * Gets an audio context from your web browser.
+ * @param options Takes an Options object or just provide a sample rate.
+ * @returns the audio context or null if there was an error or not a web browser.
+ */
+export function getContext(options?: Options | number): AudioContext | null;
```
amcasey: It looks like, to match the underlying library, you want to drop all `export` keywords and then add `exports = getContext;` at the bottom.

types/audio-play/audio-play-tests.ts
```diff
@@ -0,0 +1,6 @@
+import { audioPlay } from "audio-play";
```
amcasey: The import corresponding to the change I suggested above (and, as far as I can tell, the underlying library) is `import audioPlay = require("audio-play");`.

types/audio-play/index.d.ts
```diff
@@ -0,0 +1,41 @@
...
+ * @param start The timestamp at which to start the audio. Can be negative to start from end. (Default: 0)
+ * @param end The timestamp the audio ends at. (Default: length of audio buffer)
+ * @param autoplay Plays back the audio immediately upon loading. (Default: false)
+ * @param loop Continuously loops the buffer until paused. (Default: false)
+ * @param context Handle to an audio context. If not provided, one is provided for you.
+ * @param volume (not implemented) Playback the audio at a percentage of full volume. (Default: 1)
+ * @param detune (not implemented) Percentage of fine-tuning. (Default: 0)
+ * @param rate (not implemented) Playback rate, in percent, of the audio. (Default: 1)
+ */
+export interface Options {
```
amcasey: Same comment.

amcasey (changes requested):

types/audio-context/index.d.ts
bdjeffyp: I had thought about that, but doing so affects the accessibility of the Options interface. I wouldn't be able to export the interface then because of the `export = getContext'`. I feel that the interface is important for users to have since it defines exactly what the `getContext()` method is expecting to get.
I couldn't find any examples in my searches for a solution to that problem. Do you have any recommendations?

types/audio-play/audio-play-tests.ts
bdjeffyp: I can also use `import audioPlay from "audio-play";` since using `export = audioPlay;` is essentially a default export?

types/audio-context/index.d.ts
bdjeffyp: Nevermind. I figured out what I was doing wrong after another look at the module-function template. 🤦‍♂️😅

types/audio-play/audio-play-tests.ts
bdjeffyp: Same as above. The module-function template had all the answers.

amcasey: I believe it only functions as a default if you enable synthetic default exports in your tsconfig file.

amcasey:
@bdjeffyp thanks for the changes!  Looks like you just need to remove some redundant `export` modifiers on members of already-exported namespaces.

bdjeffyp:
Lol whoops! That should do it. I appreciate the review!

amcasey (approved):

PR #38120, opened 2019-09-04
[@types/react-native-canvas] Update type definitions and test cases.

chubillkelvin:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/iddan/react-native-canvas/issues/148>>

### DIFF
--- a/types/react-native-canvas/index.d.ts
+++ b/types/react-native-canvas/index.d.ts
@@ -1,21 +1,45 @@
 // Type definitions for react-native-canvas 0.1
 // Project: https://github.com/iddan/react-native-canvas#readme
 // Definitions by: hmajid2301 <https://github.com/hmajid2301>
+//                 Kelvin Chu <https://github.com/ragebill>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 3.1
+// TypeScript Version: 3.5
 
-import * as React from "react";
-import { StyleProp, ViewStyle } from "react-native";
+import * as React from 'react';
+import { StyleProp, ViewStyle } from 'react-native';
 
 export interface CanvasProps {
     baseUrl?: string;
     originWhitelist?: string[];
-    ref: (canvas: Canvas) => any;
+    ref: ((canvas: Canvas) => any) | React.RefObject<Canvas>;
     style?: StyleProp<ViewStyle>;
 }
 
+export interface DOMMatrix2DInit {
+    a?: number;
+    b?: number;
+    c?: number;
+    d?: number;
+    e?: number;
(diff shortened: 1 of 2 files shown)


hmajid2301 (approved):

amcasey:
@iddan expressed an interest in reviewing this.

types/react-native-canvas/index.d.ts
```diff
@@ -165,6 +136,7 @@ export default class Canvas extends React.Component<CanvasProps> {
 }
 
 export class Image {
+    constructor(canvas: Canvas, height: number, width: number);
```
minestarks: This is redundant with the already existing constructor declaration, isn't it?

types/react-native-canvas/index.d.ts
```diff
@@ -1,21 +1,45 @@
 // Type definitions for react-native-canvas 0.1
 // Project: https://github.com/iddan/react-native-canvas#readme
 // Definitions by: hmajid2301 <https://github.com/hmajid2301>
+//                 Kelvin Chu <https://github.com/ragebill>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 3.1
+// TypeScript Version: 3.5
```
minestarks: Why was the TS version upgrade needed?

minestarks (changes requested):

types/react-native-canvas/index.d.ts
RageBill: The main reason is because I personally used and tested this on Typescript 3.5. Although this change is not needed, I tend to state the latest version of software that this definition has been tested on since people tend to adopt newer version of Typescript in their projects. Do you think it would be better to keep it as 3.1?

types/react-native-canvas/index.d.ts
RageBill: You are right! This seems to be a mistake when I moved the code around. Thanks for spotting!

minestarks (approved):

types/react-native-canvas/index.d.ts
minestarks: Yeah this creates a lot of confusion, and yours was a reasonable way to interpret it. But this version states the _minimum required_ TypeScript version to be able to use these typings - so you'd only want to increment it if you're using newer TypeScript features in your typings, or depending on other packages that do. Keeping it on 3.1 means users on older versions of TS don't miss out on your changes.

RageBill: Didn't get my head around to realize it was the minimum required version! Thanks for the clarification here :)

PR #38125, opened 2019-09-04
[react-instantsearch-native:hotfix] avoid to set type conflicted prop

tkow:
react-instantsearch-core remove algoliaClient prop and, we shouldn't permit that auth information props and clientSearch prop  coexist. This PR is adapted.
I save backward compatibility.

https://github.com/algolia/react-instantsearch/blob/642ba0b207deb8534c2d97bcde2f2b1fedd7e48c/packages/react-instantsearch-dom/src/core/createInstantSearchServer.js#L147-L160

Please fill in this template.

If adding a new definition:
- [ ] `tslint.json` should be present and it shouldnt have any additional or disabling of rules. Just content as `{ "extends": "dtslint/dt.json" }`. If for reason the some rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]`  and not for whole package so that the need for disabling can be reviewed.

If removing a declaration:

### DIFF
--- a/types/react-instantsearch-native/index.d.ts
+++ b/types/react-instantsearch-native/index.d.ts
@@ -63,6 +63,7 @@ export interface UsingSearchClientProps extends InstantSearchBaseProps {
 export interface UsingManualInfoProps extends InstantSearchBaseProps {
     apiKey: string;
     appId: string;
+    algoliaClient?: any;
 }
 
 export type InstantSearchProps = UsingSearchClientProps | UsingManualInfoProps;
--- a/types/react-instantsearch-native/index.d.ts
+++ b/types/react-instantsearch-native/index.d.ts
@@ -60,10 +60,9 @@ export interface UsingSearchClientProps extends InstantSearchBaseProps {
     searchClient: any;
 }
 
-export interface UsingManualInfoProps extends InstantSearchBaseProps  {
+export interface UsingManualInfoProps extends InstantSearchBaseProps {
     apiKey: string;
     appId: string;
-    algoliaClient?: any;
 }
 
 export type InstantSearchProps = UsingSearchClientProps | UsingManualInfoProps;
--- a/types/react-instantsearch-native/index.d.ts
+++ b/types/react-instantsearch-native/index.d.ts
@@ -56,17 +56,17 @@ interface InstantSearchBaseProps {
     };
 }
 
-export interface UsingSearchClient extends InstantSearchBaseProps {
+export interface UsingSearchClientProps extends InstantSearchBaseProps {
     searchClient: any;
 }
 
-export interface UsingManualInfo extends InstantSearchBaseProps  {
+export interface UsingManualInfoProps extends InstantSearchBaseProps  {
     apiKey: string;
     appId: string;
     algoliaClient?: any;
 }
 
-export type InstantSearchProps = UsingSearchClient | UsingManualInfo;
+export type InstantSearchProps = UsingSearchClientProps | UsingManualInfoProps;
 /**
  * <InstantSearch> is the root component of all React InstantSearch implementations. It provides all the connected components (aka widgets) a means to interact with the searchState.
  *
--- a/types/react-instantsearch-native/index.d.ts
+++ b/types/react-instantsearch-native/index.d.ts
@@ -66,7 +66,7 @@ export interface UsingManualInfo extends InstantSearchBaseProps  {
     algoliaClient?: any;
 }
 
-export type InstantSearchProps = UsingSearchClient | UsingManualInfo
+export type InstantSearchProps = UsingSearchClient | UsingManualInfo;
 /**
  * <InstantSearch> is the root component of all React InstantSearch implementations. It provides all the connected components (aka widgets) a means to interact with the searchState.
  *
--- a/types/react-instantsearch-native/index.d.ts
+++ b/types/react-instantsearch-native/index.d.ts
@@ -57,13 +57,13 @@ interface InstantSearchBaseProps {
 }
 
 export interface UsingSearchClient extends InstantSearchBaseProps {
-    searchClient: any
+    searchClient: any;
 }
 
 export interface UsingManualInfo extends InstantSearchBaseProps  {
     apiKey: string;
-    appId:string;
-    algoliaClient?: any
+    appId: string;
+    algoliaClient?: any;
 }
 
 export type InstantSearchProps = UsingSearchClient | UsingManualInfo
--- a/types/react-instantsearch-native/index.d.ts
+++ b/types/react-instantsearch-native/index.d.ts
@@ -42,13 +42,8 @@ export { connectStats } from 'react-instantsearch-core';
 export { connectToggleRefinement } from 'react-instantsearch-core';
 
 // Native
-export interface InstantSearchProps {
-    apiKey: string;
-    appId: string;
+interface InstantSearchBaseProps {
     indexName: string;
-
-    algoliaClient?: any;
-    searchClient?: any;
     createURL?: (...args: any[]) => any;
     searchState?: any;
     refresh?: boolean;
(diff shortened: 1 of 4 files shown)


types/react-instantsearch-native/index.d.ts
```diff
@@ -60,6 +55,18 @@ export interface InstantSearchProps {
...
 }
+
+export interface UsingSearchClient extends InstantSearchBaseProps {
+    searchClient: any;
+}
+
+export interface UsingManualInfo extends InstantSearchBaseProps  {
+    apiKey: string;
+    appId: string;
+    algoliaClient?: any;
```
Haroenv: we are removing these in v6, so will be easy to update!

Haroenv (approved):
could be nice if this change is done in dom too, but no big worry

types/react-instantsearch-native/index.d.ts
```diff
@@ -60,6 +55,18 @@ export interface InstantSearchProps {
...
+}
+
+export type InstantSearchProps = UsingSearchClient | UsingManualInfo;
```
amcasey: I could be missing something obvious, but it looks like this makes accessing `apiKey` on an object of type `InstantSearchProps` an error.  Won't that break existing consumers?

amcasey (changes requested):
Not blocking, just changing status for visibility.

types/react-instantsearch-native/index.d.ts
tkow: Does it mean you use this prop type? 
In the case, we may use UsingManualInfoProps or type confinement. This PR is stricter than ever, so someone may meet breaking change but it's reasonable as I thought it.

types/react-instantsearch-native/index.d.ts
tkow: I see. if that release is stable, I'll remove them asap.

types/react-instantsearch-native/index.d.ts
tkow: Or I'm sorry if I misunderstand what you mean. If you mentioning when initializing IntantSearch components, we need either props
```
 {
  apiKey: string;
  appId: string;
  algoliaClient?: any;
}
``` 
or
```
 {
   searchClient: any;
 }
```
as required. So you need apiKey with appId parameters, too. Could I answer it?

types/react-instantsearch-native/index.d.ts
tkow: Oh, I misunderstood it, it is no need to input forever. OK I remove it,

types/react-instantsearch-native/index.d.ts
Haroenv: I agree with @tkow

gburgett: Another option is to add `searchClient: undefined` to the `UsingManualInfo` definition, and vice versa.

types/react-instantsearch-native/index.d.ts
gburgett: Please don't remove it yet!  Or if you do, then you should copy these types into a v5 folder as described here: https://github.com/DefinitelyTyped/DefinitelyTyped#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package

This is important for all the people still using version 5.x of the react-instantsearch library who want to get upgrades or contribute back to the type definitions.

gburgett (changes requested):
Please do not make breaking API changes to support v6.0 without bumping the major version of the types as well!

types/react-instantsearch-native/index.d.ts
amcasey: Sorry, I think I was unclear.  Suppose you had 
```ts
type T = { a: number } | { b: number };
let t: T;
```
Then `t` would have either a property called `a` or a property called `b`, which I think is what you're going for.  However, as written, both `t.a` and `t.b` are errors because the union type has neither of those properties.

gburgett: @amcasey yes I've run into just this problem in my typescript projects.  You are correct!

Supposing your example, you could not do this:

```ts
let t: T = ...
console.log(t.a)  // ERROR!
```

You have to do this:

```ts
let t: T = ...
if ('a' in t) {
  console.log(t.a)
}
```

However, if you define the types like this:
```ts
type T = { a: number, b?: undefined } | { a?: undefined, b: number };
```

then you can do this:
```ts
let t: T = ...
if (t.a) {
  console.log(t.a)
}
```

types/react-instantsearch-native/index.d.ts
tkow: @gburgett

https://github.com/algolia/react-instantsearch/commit/174cc28a6ed8ef09da1b5128c5d2157182303e3a

algoliaClient parameter deprecated from v5.1.0. If users need backward compatibility less than this version, I keep it, but I wonder if it's possible to avoid this by versioning as you described.

types/react-instantsearch-native/index.d.ts
tkow: @gburgett 
You describe clearly what I want to tell @amcasey. And as I see it, InstantSearchProps are useless type about exporting your project in most case, because, when you want to use this props' parameter in your callback args or your components' props,   you can achieve  UsingSearchClientProps or UsingManualInfoProps instead of InstantSearchProps in your components. It's more preferable to before because typescript will not know which props used when initialize in these scope. It depends on developers's choice. So, I don't have any idea about case we need UsingSearchClientProps type except for you want to pass the props from upper parent keeping ambiguous initialize props. In this context, we don't need  https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38125#discussion_r321316445 change as I thought it.

types/react-instantsearch-native/index.d.ts
tkow: And if it kept, I need some modification  because apiKey and appId exclusively independent on  algoliaClient. Please let me know if anyone has opinions about this.

Haroenv:
@gburgett, don’t worry, this is merely fixing an error in the typings, passing no props was already not allowed before

types/react-instantsearch-native/index.d.ts
gburgett: @tkow Thank you for pointing this out!  Yes it is deprecated, but deprecated is not the same thing as removed.  If we remove it from the types, anyone who consumes the types for v5 and uses the deprecated API will experience a *breaking change*.  Deprecation in general is used to avoid breaking changes, and warn the user that they will experience a breaking change *in the future*.

Since v6 of the react-instantsearch library is not out yet (current version on npmjs is 5.7.0), we should not make a separate v5 and v6 types folder.  The current types should continue to faithfully represent the public API of react-instantsearch 5.x.x, *including* deprecated properties.

If you would be willing to separate out the type definition then please do!  I am happy with wherever we land on this, so long as the end user can upgrade their type definitions and still use the deprecated API.

gburgett:
@haroenv sure, the breaking API change I'm referring to is the removal of the deprecated parameter.  Please see above!

Haroenv:
Oh yes definitely!

types/react-instantsearch-native/index.d.ts
```diff
@@ -42,13 +42,8 @@ export { connectStats } from 'react-instantsearch-core';
 export { connectToggleRefinement } from 'react-instantsearch-core';
 
 // Native
-export interface InstantSearchProps {
-    apiKey: string;
-    appId: string;
+interface InstantSearchBaseProps {
     indexName: string;
-
-    algoliaClient?: any;
```
Haroenv: Algoliaclient still needs to be possible with v5 :)

Haroenv (changes requested):

types/react-instantsearch-native/index.d.ts
tkow: @grurgett OK. I fixed it. This patch focus type mismatch so, backword compatibility type directory another PR.

gburgett (approved):

gburgett:
Thanks @tkow!

Haroenv (approved):

PR #35274, opened 2019-05-08
Remove definitions for deprecated MapView

methodbox:
Related to:  https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35239

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If removing a declaration:
- [x] If a package was never on DefinitelyTyped, you don't need to do anything. (If you wrote a package and provided types, you don't need to register it with us.)

This is related to a conflict with the now-separate module "MapView" / `react-native-maps`, which is also a deprecated and unsupported module from `react-native` (these are two separate modules, `react-native-maps` is now community supported and maintained). The existence of these types conflicts with that modules index.d.ts.

I have a PR to get them to remove that if we can also move all types into an @types/react-native-maps module for better maintenance and consistency.

I cannot fix the conflict in my original PR on @types/react-native without this change.

PR #35239, opened 2019-05-06
Remove MapView definitions from @types/react-native - create new react-native-maps module

methodbox:
initialRegion is a valid prop in `react-native-maps` but it's definition is missing in the @types.

This contains the same properties as the `region` and is optional, and therefore should be defined as 

```js
export interface MapViewProps extends ViewProps {
...
initialRegion?: MapViewRegion;
...
}
```
Which is the same as `region`.

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
You can test this simply by placing the following code in a `.tsx` file and note that `MapView` is flagged for the `initialRegion` with no definition:
```js
import React, { Component } from 'react';
import { MapView } from 'expo';  //  or react-native-maps

export default class MapTest extends Component {
  render() {
    return (
      <MapView
        style={{ flex: 1 }}
        initialRegion={{
          latitude: 37.78825,
          longitude: -122.4324,
          latitudeDelta: 0.0922,
          longitudeDelta: 0.0421,
        }}
      />
    );
  }
}
```
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: 
This is defined in their component API docs here:
https://github.com/react-native-community/react-native-maps/blob/master/docs/mapview.md

PR #38128, opened 2019-09-04
[react-native] Minor UIManager update, and removal of dead types

acoates-ms:
Over time various types have been deprecated from react-native, and never removed from the types. -- Most of this is removal of types that are no longer part of react-native.

I also added a few deprecation attributes to types that will be removed soon, and throw runtime deprecation warnings if used.

Also added the getViewManagerConfig function to UIManager, since that was missing.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/react-native

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -8855,17 +8855,6 @@ declare global {
     var console: Console;
 
     /**
-     * Navigator object for accessing location API
-     * @see https://facebook.github.io/react-native/docs/javascript-environment.html#polyfills
-     */
-    interface Navigator {
-        readonly product: string;
-        readonly geolocation: Geolocation;
-    }
-
-    const navigator: Navigator;
-
-    /**
      * This contains the non-native `XMLHttpRequest` object, which you can use if you want to route network requests
      * through DevTools (to trace them):
      *
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -22,6 +22,7 @@
 //                 Ceyhun Ozugur <https://github.com/ceyhun>
 //                 Mike Martin <https://github.com/mcmar>
 //                 Theo Henry de Villeneuve <https://github.com/theohdv>
+//                 Eli White <https://github.com/TheSavior>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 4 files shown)


Esemesek (approved):
lgtmt

Titozzz (approved):

types/react-native/index.d.ts
```diff
@@ -8500,6 +7860,12 @@ export interface UIManagerStatic {
         success: (item: string, index: number | undefined) => void
     ): void;
 
+    getViewManagerConfig: (
+        name: string,
+      ) => {
+        Commands: {[key: string]: number};
```
TheSavior: I think this might be optional, but it is typed as `Object` in the RN flow type, so...this is at least better than that...

I'm fine with this if you are.

elicwhite (approved):

types/react-native/index.d.ts
```diff
@@ -22,6 +22,7 @@
 //                 Ceyhun Ozugur <https://github.com/ceyhun>
 //                 Mike Martin <https://github.com/mcmar>
 //                 Theo Henry de Villeneuve <https://github.com/theohdv>
+//                 Eli White <https://github.com/TheSavior>
```
alloy: @TheSavior :wave:

alloy (approved):
Great stuff 👌 

Looks like this is breaking the current styled-components typings, but in reality it should probably be updated to no longer use WebView from RN but from https://github.com/react-native-community/react-native-webview instead, which ships with an ambient declaration file.

acoates-ms:
> Great stuff 👌
> 
> Looks like this is breaking the current styled-components typings, but in reality it should probably be updated to no longer use WebView from RN but from https://github.com/react-native-community/react-native-webview instead, which ships with an ambient declaration file.

Do you know if that break will block the PR?  I'm not that familiar with making changes on DefinitelyTyped.  - It does seem like the issue is on styled-component at this point though.

alloy:
With #35239 & #35274 there already is precedent for removing deprecated native bits from the styled-components typings. As such, I feel it would be fair to include a change in this PR that would remove `WebView` coverage from the current styled-components typings.

/cc @JKillian @Igorbek

orta:
I've pushed a commit which I think should fix CI, removing the webview from styled components - should unblock this PR

alloy:
@orta Yup, that’s it—thanks!

alloy:
I actually wanted to wait to hear from the maintainers of the styled-components typings before merging, but I should have said so explicitly. My bad.

PR #38131, opened 2019-09-04
Added types for html-webpack-tags-plugin

maxbogus:
Please fill in this template.

If adding a new definition:
- [ x] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration`

If removing a declaration:

### DIFF
--- a/types/html-webpack-tags-plugin/index.d.ts
+++ b/types/html-webpack-tags-plugin/index.d.ts
@@ -1,6 +1,8 @@
 // Type definitions for html-webpack-tags-plugin 2.0
 // Project: https://github.com/jharris4/html-webpack-tags-plugin
-// Definitions by: // Definitions by: 贺师俊 <https://github.com/hax>, Vladimir Grenaderov <https://github.com/VladimirGrenaderov>, Max Boguslavskiy <https://github.com/maxbogus>
+// Definitions by: 贺师俊 <https://github.com/hax>
+//                 Vladimir Grenaderov <https://github.com/VladimirGrenaderov>
+//                 Max Boguslavskiy <https://github.com/maxbogus>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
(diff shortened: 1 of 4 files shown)


types/html-webpack-tags-plugin/index.d.ts
```diff
@@ -0,0 +1,17 @@
+// Type definitions for html-webpack-tags-plugin 2.0
+// Project: https://github.com/jharris4/html-webpack-tags-plugin
+// Definitions by: // Definitions by: 贺师俊 <https://github.com/hax>, Vladimir Grenaderov <https://github.com/VladimirGrenaderov>, Max Boguslavskiy <https://github.com/maxbogus>
```
rbuckton: You have duplicated the `// Definitions by:` part of the comment. I would also suggest this:

```suggestion
// Definitions by: 贺师俊 <https://github.com/hax>
//                 Vladimir Grenaderov <https://github.com/VladimirGrenaderov>
//                 Max Boguslavskiy <https://github.com/maxbogus>
```

types/html-webpack-tags-plugin/index.d.ts
```diff
@@ -0,0 +1,17 @@
...
+
+import { Plugin } from 'webpack';
+
+export interface Options {
+    tags: string | string[];
+    files?: string | string[];
+    append?: boolean;
+}
+
+export default class HtmlWebpackTagsPlugin extends Plugin {
```
rbuckton: According to the source, this is exported with `module.exports =` not `exports.default =`, so this import would not work at runtime unless additional compiler options are specified. I would recommend using the `export =` syntax to emulate the actual behavior, and move `Options` into an `HtmlWebpackTagsPlugin` namespace that merges with the class definition:

```ts
class HtmlWebpackTagsPlugin extends Plugin {
  constructor(options: HtmlWebpackTagsPlugin.Options);
}
namespace HtmlWebpackTagsPlugin {
  interface Options { ... }
}
export = HtmlWebpackTagsPlugin;
```

rbuckton (changes requested):

maxbogus:
@rbuckton , please review my changes.

maxbogus:
@rbuckton , please review my changes.

PR #38142, opened 2019-09-05
feat(aws-lambda) add S3BatchEvent

pwmcintyre:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://docs.aws.amazon.com/AmazonS3/latest/dev/batch-ops-invoke-lambda.html
- [ ] (Not applicable) If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ ] (Not applicable) If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

### DIFF
--- a/types/aws-lambda/index.d.ts
+++ b/types/aws-lambda/index.d.ts
@@ -271,7 +271,7 @@ export interface S3BatchEventJob {
 export interface S3BatchEventTask {
     taskId: string;
     s3Key: string;
-    s3VersionId?: string;
+    s3VersionId: string | null;
     s3BucketArn: string;
 }
 
--- a/types/aws-lambda/index.d.ts
+++ b/types/aws-lambda/index.d.ts
@@ -257,6 +257,7 @@ export type S3CreateEvent = S3Event; // old name
  * S3 Batch Operations event
  * https://docs.aws.amazon.com/AmazonS3/latest/dev/batch-ops-invoke-lambda.html
  */
+
 export interface S3BatchEventRequest {
     invocationSchemaVersion: string;
     invocationId: string;
@@ -269,10 +270,10 @@ export interface S3BatchEventJob {
 }
 
 export interface S3BatchEventTask {
-    taskId: string;
-    s3Key: string;
-    s3VersionId: string | null;
-    s3BucketArn: string;
+  taskId: string;
+  s3Key: string;
+  s3VersionId: string | null;
+  s3BucketArn: string;
 }
 
 export interface S3BatchEventResponse {
@@ -1166,5 +1167,8 @@ export type FirehoseTransformationHandler = Handler<FirehoseTransformationEvent,
 
 export type CustomAuthorizerHandler = Handler<CustomAuthorizerEvent, CustomAuthorizerResult>;
 export type CustomAuthorizerCallback = Callback<CustomAuthorizerResult>;
+
+export type S3BatchHandler = Handler<S3BatchEventRequest, S3BatchEventResponse>;
+export type S3BatchCallback = Callback<S3BatchEventResponse>;
 
 export as namespace AWSLambda;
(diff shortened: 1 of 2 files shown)


pwmcintyre:
seeking feedback / review from recent contributors: @juancarbonel @redlickigrzegorz @skyzenr @loikg @dalen

skyzenr (approved):
Looks good to me!

amcasey:
@andrewbranch, this looks good to me but the completion perf number is superficially concerning.  What is that testing?  Would it be sensitive to the fact that there are now more symbols sharing a prefix?

types/aws-lambda/index.d.ts
```diff
@@ -252,6 +253,46 @@ export interface S3Event {
...
+}
+
+export interface S3BatchEventJob {
+    id: string;
+}
+
+export interface S3BatchEventTask {
+    taskId: string;
+    s3Key: string;
+    s3VersionId?: string;
```
simonbuchan: This being optional makes me a little curious, as it isn't mentioned in the docs: is it completely not present for non-versioned buckets?

simonbuchan (changes requested):
This looks pretty great, but it's missing the specialized handler and context type, down the bottom of `index.d.ts`. I should probably clean this up at some point to make things like that more obvious....

types/aws-lambda/index.d.ts
pwmcintyre: yeah that's a really good point -- all my buckets are actually version enabled, let me whip something up

pwmcintyre: @simonbuchan 

okay the results are in -- and you're correct, it's `null` -- i've committed a fix

actual payload:
```
            {
                "taskId": "AAAAAAAAAAHLSJmnbBR1fLJ5Y9/truncated",
                "s3BucketArn": "arn:aws:s3:::ommitted",
                "s3Key": "47600731-b366-4444-b0f4-24255c08c6e3.json",
                "s3VersionId": null
            }
```

pwmcintyre:
@simonbuchan -- addressed, see latest commit

> This looks pretty great, but it's missing the specialized handler and context type, down the bottom of `index.d.ts`. I should probably clean this up at some point to make things like that more obvious....

types/aws-lambda/index.d.ts
simonbuchan: Thanks for checking!

simonbuchan:
Thanks for the investigation and fixes!

Sorry for the pedentry, but I also just noticed looking at the handler types you're using `...EventRequest` and `...EventResponse`, the convention is that the actual handler input and result types are `...Event` and `...Result`, so for this `S3BatchEvent` and `S3BatchResult` (and similar updates for the child types).

There are probably some `...EventRequest` and `...EventResponse` types in there for `request` and `response` properties of `...Event`, so I'd like to keep these consistent to avoid confusion. It does unfortunately suggest the `results` property of `S3BatchResult` should have type `S3BatchResultResult[]`, though 🤷‍♂️

pwmcintyre:
@simonbuchan -- totally fair points, please review latest commits. It does get a little quirky with the `S3BatchResultResultCode` ... but it does hold true to the convention, so I think that's how it should stay.

> Thanks for the investigation and fixes!
> 
> Sorry for the pedentry, but I also just noticed looking at the handler types you're using `...EventRequest` and `...EventResponse`, the convention is that the actual handler input and result types are `...Event` and `...Result`, so for this `S3BatchEvent` and `S3BatchResult` (and similar updates for the child types).
> 
> There are probably some `...EventRequest` and `...EventResponse` types in there for `request` and `response` properties of `...Event`, so I'd like to keep these consistent to avoid confusion. It does unfortunately suggest the `results` property of `S3BatchResult` should have type `S3BatchResultResult[]`, though 🤷‍♂️

simonbuchan (approved):
Looks great! Thanks again for the quick turnaround.

PR #38144, opened 2019-09-05
feat(echarts): add types for echarts

theWorkFsz:
+ lineStyle.color can be an string[]
https://echarts.apache.org/zh/option.html#yAxis.splitLine.lineStyle.color

+ cartesianAxis.min/max can be a function
https://echarts.apache.org/zh/option.html#yAxis.min

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://www.echartsjs.com/zh/option.html#yAxis.min>

<https://www.echartsjs.com/zh/option.html#yAxis.splitLine.lineStyle.color>

### DIFF
--- a/types/echarts/index.d.ts
+++ b/types/echarts/index.d.ts
@@ -1170,9 +1170,9 @@ declare namespace echarts {
                  * Moreover, it can be set as negative number, like `-3`.
                  *
                  * @default null
-                 * @see https://ecomfe.github.io/echarts-doc/public/en/option.html#yAxis.min
+                 * @see https://echarts.apache.org/zh/option.html#yAxis.min
                  */
-                min?: number | string;
+                min?: number | string | ((value: { min:number, max:number }) => number);
 
                 /**
                  * The maximum value of axis.
@@ -1187,9 +1187,9 @@ declare namespace echarts {
                  * Moreover, it can be set as negative number, like `-3`.
                  *
                  * @default null
-                 * @see https://ecomfe.github.io/echarts-doc/public/en/option.html#yAxis.max
+                 * @see https://echarts.apache.org/zh/option.html#yAxis.max
                  */
-                max?: number | string;
+                 max?: number | string | ((value: { min:number, max:number } ) => number);
 
                 /**
                  * It is available only in numerical axis, i.e., type: `'value'`.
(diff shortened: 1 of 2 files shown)


types/echarts/index.d.ts
```diff
@@ -1170,9 +1170,9 @@ declare namespace echarts {
                  * Moreover, it can be set as negative number, like `-3`.
                  *
                  * @default null
-                 * @see https://ecomfe.github.io/echarts-doc/public/en/option.html#yAxis.min
+                 * @see https://echarts.apache.org/zh/option.html#yAxis.min
```
tmtron: I recommend that the the link should point to the English version: e.g. `/en/` instead of `/zh/`

types/echarts/index.d.ts
```diff
@@ -1187,9 +1187,9 @@ declare namespace echarts {
                  * Moreover, it can be set as negative number, like `-3`.
                  *
                  * @default null
-                 * @see https://ecomfe.github.io/echarts-doc/public/en/option.html#yAxis.max
+                 * @see https://echarts.apache.org/zh/option.html#yAxis.max
```
tmtron: I recommend that the the link should point to the English version: e.g. `/en/` instead of `/zh/`

types/echarts/index.d.ts
Ovilia: You may use `https://echarts.apache.org/option.html#yAxis.min` because it will redirect to the English or Chinese version based on the language setting of the Web browser.

types/echarts/index.d.ts
Ovilia: Same as before.

Ovilia (changes requested):

types/echarts/index.d.ts
```diff
@@ -1170,9 +1170,9 @@ declare namespace echarts {
                  * Moreover, it can be set as negative number, like `-3`.
                  *
                  * @default null
-                 * @see https://ecomfe.github.io/echarts-doc/public/en/option.html#yAxis.min
+                 * @see https://echarts.apache.org/option.html#yAxis.min
                  */
-                min?: number | string;
+                min?: number | string | ((value: { min:number, max:number }) => number);
```
Ovilia: I think there should be a space before `number`.

Ovilia (changes requested):

Ovilia (approved):

orta:
Let's get this in, the formatting issue isn't critical

PR #38145, opened 2019-09-05
[@type/conflux-web]

liuis:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/conflux-web/cfx/abi.d.ts
+++ b/types/conflux-web/cfx/abi.d.ts
@@ -1,0 +1,47 @@
+export interface ABIDefinition {
+    constant?: boolean;
+    payable?: boolean;
+    stateMutability?: "pure" | "view" | "nonpayable" | "payable";
+    anonymous?: boolean;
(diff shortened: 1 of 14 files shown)


liuis:
I have already modified the additional qualifications to code according to what you said.Main project not contain "type" directory, thank you for your support!

liuis:
I want to know need how long can be merged into master, THX PLS

uniqueiniquity (changes requested):
Thanks for your patience! We've been somewhat overwhelmed with new PRs recently.

Since `conflux-web`, `conflux-web-cfx`, and `conflux-web-utils` are available as separate packages, I suggest creating separate @types packages for each of them as well, instead of combining them all into the same package as you've done here. That will make it much easier to use and review. Thanks!

EDIT: Actually, it seems like there are even more sub-packages than just those. Even though it seems like a lot of duplication, I really recommend having one @types package per "real" package.

liuis:
Thank you for your review, your suggestion is very good and to the point, we will be improved in the next release.

PR #38154, opened 2019-09-05
style(redux-api-middleware): improve generics names

unrevised6419:

### DIFF
--- a/types/redux-api-middleware/index.d.ts
+++ b/types/redux-api-middleware/index.d.ts
@@ -13,13 +13,14 @@ import {
     applyMiddleware
 } from 'redux';
 
+type TypeOrResolver<Arg, Type> = Type | ((arg: Arg) => Type);
+
 /**
  * This module is also a UMD module that exposes a global variable 'ReduxApiMiddleware'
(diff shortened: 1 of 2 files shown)


types/redux-api-middleware/index.d.ts
```diff
@@ -13,13 +13,14 @@ import {
...
 
+type TypeOrResolver<Arg, Type> = Type | ((arg: Arg) => Type);
+
 /**
  * This module is also a UMD module that exposes a global variable 'ReduxApiMiddleware'
  * when loaded outside a module loader environment.
  */
 export as namespace ReduxApiMiddleware;
 
-export type RSAA = '@@redux-api-middleware/RSAA';
```
amcasey: Why eliminate the type?

iamandrewluca: It wasn't used. `redux-api-middleware` exports just this const. And `RSAAAction` uses `RSAA` as a value not a type.
This happens when only type exists.
```
'RSAA' only refers to a type, but is being used as a value here.
```

Mrman (approved):
LGTM

PR #38155, opened 2019-09-05
Initial commit for Akumina types (see akumina.github.io)

MarshHawk:
Please fill in this template.

If adding a new definition:


.github/CODEOWNERS
```diff
@@ -2,6 +2,5926 @@
 # Add yourself to the "Definitions by:" list instead.
 # See https://github.com/DefinitelyTyped/DefinitelyTyped#edit-an-existing-package
 
+<<<<<<< HEAD
```
amcasey: This file is updated automatically by a tool - please don't edit it directly.

amcasey (changes requested):
The underlying package appears to already contain a .d.ts file.  Is the intention to host it here instead?

.github/CODEOWNERS
MarshHawk: I think this is fixed now, thanks

MarshHawk: Also, yes, the intention is to deprecate the akumina-core NPM package and just release the types package for now...

MarshHawk:
Addressed issues, thanks

MarshHawk:
@amcasey Hey, I think we addressed the issues. Just wondering if the wait is due to additional fixes/ our approach or if you're just flat out?...

amcasey:
@MarshHawk Oh no!  I completely missed the ping.  Sorry about that.  Looking now.

types/akumina-core/tslint.json
```diff
@@ -0,0 +1,9 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "npm-naming": false,
```
amcasey: This appears to be unused.

types/akumina-core/index.d.ts
```diff
@@ -0,0 +1,1471 @@
...
+        class Calendar {
+            constructor(options: any);
+        }
+        class CompanyCalendar {
+            constructor(options: any);
+        }
+    }
+
+    namespace Digispace {
+        class ConfigurationContext {
```
amcasey: I assume you know from the lint messages that a class with only static members doesn't make much sense but prefer to keep it this way.

amcasey (approved):

types/akumina-core/index.d.ts
MarshHawk: yes, yes, just deferring that improvement for the moment, thanks.

PR #38156, opened 2019-09-05
[@types/lodash] returned object from defaults cannot be null/undefiend

XVincentX:
According to the docs and the source code, the returned object from `defaults` is never going to be null or undefined.

I can see there are multiple lodash object definitions, but to be honest I'm not too sure which one I should change, so I changed them all

### DIFF
--- a/types/lodash/common/object.d.ts
+++ b/types/lodash/common/object.d.ts
@@ -858,7 +858,7 @@ declare module "../index" {
         defaults(
             object: any,
             ...sources: any[]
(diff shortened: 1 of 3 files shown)


DenisFrezzato (approved):

DenisFrezzato (approved):

XVincentX:
Is anybody going to help with this?

uniqueiniquity:
Thanks for your patience!

PR #38159, opened 2019-09-05
[@types/facepaint] update `DynamicStyleFunction` signature to return array

timswalling:
Updates the signature of the `DynamicStyleFunction` so the returned value is an array of `DynamicStyle` objects, to match [the value returned by the function](https://github.com/emotion-js/facepaint/blob/master/src/index.js#L66).

---

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/emotion-js/facepaint/blob/master/src/index.js#L66


DogPawHat (approved):
Blegh, I forgot I had written this. Seems ok to me.

PR #38160, opened 2019-09-05
[@types/react] Improve type of useState when storing functions

Dremora:
We have a number of use cases of storing functions inside React state. React docs don't mention anywhere that this shouldn't be done under any circumstances. However, storing functions has the following pitfall when it comes to passing `initialState` to the `useState` hook or new state to the `setState` function (the one that comes from the hook): in both cases, React will execute the functions and store their return value instead. In order to store such functions in state, they need to be wrapped into a higher level function.

This PR makes corresponding changes to the type signature of `useState`, making sure that if the value stored inside state is a function, then the only way to update the state or provide the initial value is by using the functional variant of setState or initialState (that is, wrapping the function into another function).

**Help required!** I can't seem to make this work on TS < 3.6. Lower versions break down on conditional types, as they infer the type of primitives as their value instead of their actual type (e.g. the type of value stored inside `useState(true)` will be inferred as `true` instead of `boolean`).

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://reactjs.org/docs/hooks-reference.html#functional-updates

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -21,7 +21,7 @@
 //                 Kanitkorn Sujautra <https://github.com/lukyth>
 //                 Sebastian Silbermann <https://github.com/eps1lon>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.6
 
 /// <reference path="global.d.ts" />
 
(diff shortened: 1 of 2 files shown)


ferdaber:
See this for reference: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37850 (and @Jessidhia 's comment regarding your issue too).

ferdaber:
You're going to have a hell of a time updating the minimum version for React types as it'll need to propagate to every dependent library's types.

PR #38165, opened 2019-09-05
Updated SharePoint SPCR definitions

thespooler:

### DIFF
--- a/types/sharepoint/index.d.ts
+++ b/types/sharepoint/index.d.ts
@@ -25,6 +25,100 @@ declare const enum Sods {
     loaded = 4,
     error = 5
 }
+
+declare function GetThemedImageUrl(image: string): string;
+declare function $urlHtmlEncodeString(string: string): string;
+declare var SPClientRenderer: SPClientRenderer;
+declare var MSOWebPartPageFormName: string;
+declare var HP: HP;
+
+interface SPClientRenderer {
+    AddCallStackInfoToErrors: boolean;
+    AddPostRenderCallback: (context: any, callback: () => void) => void;
+    CoreRender: () => void;
+    GlobalDebugMode: boolean;
+    IsDebugMode: () => boolean;
+    ParseTemplateString: (template: string) => () => string;
+    ParseTemplateStringWorker: (template: string) => () => string;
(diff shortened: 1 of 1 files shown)


types/sharepoint/index.d.ts
```diff
@@ -25,6 +25,100 @@ declare const enum Sods {
     loaded = 4,
     error = 5
 }
+
+declare function GetThemedImageUrl(image: string): string;
```
baywet: shouldn't be this method in the Utility class? https://docs.microsoft.com/en-us/dotnet/api/microsoft.sharepoint.utilities.sputility.getthemedimageurl?view=sharepoint-server

baywet (changes requested):

types/sharepoint/index.d.ts
```diff
@@ -11520,6 +11660,41 @@ declare namespace Srch {
...
+        activate: (prompt: string,
+            searchBoxId: string, searchBoxDivId: string,
+            navButtonId: string, suggestionsListId: string,
+            navListId: string, searchBoxLinkId: string,
+            searchBoxProgressClass: string, searchBoxPromptClass: string) => void;
+    }
+
+    interface RefinementCategory {
+        k: boolean; // Use KQL,
+        m: { [key: string]: string } | null; // token to display value map
```
andrei-markeev: please use comments with two asterisks and place them before the property, so that those comments are visible in the intellisense, for example like this:
```ts
interface RefinementCategory {
        /** use KQL */
        k: boolean;
        /** token to display value map */
        m: { [key: string]: string } | null;
        // ....
}
```

andrei-markeev (changes requested):

types/sharepoint/index.d.ts
thespooler: Not on the declarations I have. It gets pulled from init.js and is not scoped from what I can see. The link you pointed to seems to refer to SSOM, not the JS library.

baywet: ok thanks I was simply double checking!

baywet (approved):

andrei-markeev (approved):

PR #38166, opened 2019-09-05
Add sauronjs types

EricPMulligan:
Please fill in this template.

If adding a new definition:


types/sauronjs/index.d.ts
```diff
@@ -0,0 +1,19 @@
...
+
+interface SauronDOM {
+    update(): void;
+}
+
+interface SauronEvents {
+    dom: SauronDOM;
+}
+
+declare const events: SauronEvents;
```
uniqueiniquity: Instead of doing `declare const events: SauronEvents`, I recommend the following for clarity:
```ts
export namespace events {
  export namespace dom {
    export function update(): void;
  }
}
```

types/sauronjs/util/insert.d.ts
```diff
@@ -0,0 +1 @@
+export function insert(params: HTMLElement): void;
```
uniqueiniquity: From [the source](https://unpkg.com/browse/sauronjs@0.1.2/src/util/insert.js), it looks like this is a callable module.
Therefore, it should be as follows:
```ts
declare function insert(params: HTMLElement): void;
export = insert;
```
What you currently have models `module.exports.insert = insert`, but what the source shows is `module.exports = insert`.

types/sauronjs/util/ready.d.ts
```diff
@@ -0,0 +1 @@
+export function ready(fn: () => void): void;
```
uniqueiniquity: Same thing here as my previous comment - this is a callable module.

types/sauronjs/core/service.d.ts
```diff
@@ -0,0 +1,7 @@
+export class Service {
```
uniqueiniquity: Same thing here - you should `declare` the class and then `export = Service;`.

uniqueiniquity (changes requested):
I've left a couple comments about the module structure of some of your files. I think you've broken this package up really well, but you need to match each file to the underlying library a little better. I suggest taking a look at [this page](http://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html) as well as some of the templates linked there - that will give some more context to some of my suggestions. Thanks!

types/sauronjs/index.d.ts
EricPMulligan: Writing it like this, I get `strict-export-declare-modifiers` tslint error.

EricPMulligan: Entered it as:

```
export namespace events {
  namespace dom {
    function update(): void;
  }
}
```

types/sauronjs/util/insert.d.ts
EricPMulligan: Done.

types/sauronjs/util/ready.d.ts
EricPMulligan: Done.

types/sauronjs/core/service.d.ts
EricPMulligan: Done.

uniqueiniquity (approved):

uniqueiniquity:

PR #38170, opened 2019-09-06
[chart.js] Use RadialLinearScale type for scale option

indigolain:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/chartjs/Chart.js/blob/6632b8ba8475e020fd0533e2630117376a9cbc73/src/controllers/controller.radar.js#L13

### DIFF
--- a/types/chart.js/index.d.ts
+++ b/types/chart.js/index.d.ts
@@ -126,7 +126,7 @@ declare namespace Chart {
 
     type PointStyle = 'circle' | 'cross' | 'crossRot' | 'dash' | 'line' | 'rect' | 'rectRounded' | 'rectRot' | 'star' | 'triangle';
 
-    type PositionType = 'left' | 'right' | 'top' | 'bottom';
+    type PositionType = 'left' | 'right' | 'top' | 'bottom' | 'chartArea';
 
     type InteractionMode = 'point' | 'nearest' | 'single' | 'label' | 'index' | 'x-axis' | 'dataset' | 'x' | 'y';
 
@@ -434,6 +434,7 @@ declare namespace Chart {
 
     interface GridLineOptions {
         display?: boolean;
+        circular?: boolean;
         color?: ChartColor;
         borderDash?: number[];
         borderDashOffset?: number;
@@ -503,6 +504,8 @@ declare namespace Chart {
         display?: boolean;
         color?: ChartColor;
         lineWidth?: number;
+        borderDash?: number[];
+        borderDashOffset?: number;
     }
 
     interface PointLabelOptions {
@@ -511,6 +514,7 @@ declare namespace Chart {
         fontFamily?: string;
         fontSize?: number;
         fontStyle?: string;
+        lineHeight?: number|string;
     }
 
     interface LinearTickOptions extends TickOptions {
@@ -663,11 +667,14 @@ declare namespace Chart {
         minUnit?: TimeUnit;
     }
 
-    interface RadialLinearScale extends LinearScale {
-        lineArc?: boolean;
+    interface RadialLinearScale {
+        animate?: boolean;
+        position?: PositionType;
         angleLines?: AngleLineOptions;
         pointLabels?: PointLabelOptions;
-        ticks?: TickOptions;
+        ticks?: LinearTickOptions;
+        display?: boolean;
+        gridLines?: GridLineOptions;
     }
 
     interface Point {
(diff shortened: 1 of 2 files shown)


indigolain:
This change is required to enable the radar chart to use the ticks option and more.
The file linked in the pull request description defines the scale type of the radar chart to be radialLinear, which is defined below and is already defined in the current index.d.ts .
https://github.com/chartjs/Chart.js/blob/87e44fa360be580dae085d85d74614dfdc3988c5/src/scales/scale.radialLinear.js

related: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37614

adripanico (changes requested):
According to docs, this change does not apply: https://www.chartjs.org/docs/latest/charts/radar.html#scale-options

indigolain:
@adripanico 
Thank you for the review!
I also checked the document, and I believe that what the document is saying in the quoted text is that it takes the singular `scale` and not the plural `scales` option.
> The radar chart supports only a single scale. The options for this scale are defined in the scale property.

I've made a codepen to test out if I can use the `radialLinear` scale options, and it seems to work. It'll be great if you could check it out.
https://codepen.io/indigolain/pen/XWrZydB

indigolain:
I looked into the documentation of Chart.js deeper and found an example where scale options other than `display` was used, will this make this pull request approvable?

https://www.chartjs.org/docs/latest/axes/radial/linear.html#axis-range-settings

types/chart.js/index.d.ts
```diff
@@ -663,6 +664,7 @@ declare namespace Chart {
     }
 
     interface RadialLinearScale extends LinearScale {
+        display?: boolean;
```
adripanico: RadialLinearScale extends LinearScale which extends ChartScales which already define the ```display``` property.

adripanico (changes requested):

indigolain:
I think I can use the RadialChartOptions interface, introduced in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26688 , when defining the option, and then use the option to get the desired result.
Since you could also use this interface for specifying the display option, should we revert the change introduced in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37614 ?

ElianCordoba:
Okay so I just try every configuration that the `scale` have with your change and I found the following:

```
// These are defined in the RadialLinearScale interface
lineArc:     don't know what it does but in your code pen does not change anithing
angleLines:  works
pointLabels  works
ticks:       works
// This come from ChartScales which itself comes from LinearScale
type:        don't work
scaleLabel:  don't work
position:    don't work
type:        don't work
display      works
gridLines    works
xAxes        don't work
yAxes        don't work
```

My fix would be to replate the current type with 
```typescript
// Note I'm not extending from LinearScale
interface RadialLinearScale {
  lineArc?: boolean; // Need confirmation on this one
  angleLines?: AngleLineOptions;
  pointLabels?: PointLabelOptions;
  ticks?: TickOptions;
  display?: boolean;
  gridLines?: GridLineOptions;
}
```

indigolain:
@ElianCordoba 
Thank you for checking out the configuration! From looking at the chart.js implementation [here](https://github.com/chartjs/Chart.js/blob/87e44fa360be580dae085d85d74614dfdc3988c5/src/scales/scale.radialLinear.js), I confirmed that the configs below, which includes the configs you suggested, are valid.
- display
- animate (This may not be working  ref: https://github.com/chartjs/Chart.js/issues/5749)
- position
- angleLines
- gridLines
- ticks
- pointLabels

Also, it seems that `lineArc` config was replaced with angleLines, gridLines, and pointLabels config in the pull request below, so I think we should remove it if we don't have to think about 
backward compatibility.
https://github.com/chartjs/Chart.js/pull/3782

@adripanico 
The change @ElianCordoba suggested will narrow down the invalid config that a user may specify, but I believe that this will not solve the problem you mentioned about the generic options interface used openly. What do you think about this?

indigolain:
I've edited the types based on @ElianCordoba 's idea, and added some tests for the changes that I made.

And I've also added some type definitions based on the references below.

For PositionType
https://github.com/chartjs/Chart.js/blob/db6a6d5989cb33dda38c3ecbe2ce0b4b19b3d013/src/core/core.layouts.js#L208

For GridLineOptions
https://www.chartjs.org/docs/latest/axes/styling.html#grid-line-configuration

For PointLabelOptions
https://www.chartjs.org/docs/latest/axes/radial/linear.html#point-label-options

ElianCordoba:
Looks good to me, @adripanico what do you think?

adripanico (approved):

indigolain:
@adripanico  @ElianCordoba 
Thank you for reviewing my codes!

adripanico:
> @adripanico @ElianCordoba
> Thank you for reviewing my codes!

Thanks to you!

PR #38179, opened 2019-09-06
react-redux: correct `connect` return type

OliverJAsh:
Re. https://github.com/reduxjs/react-redux/releases/tag/v7.0.1:

> #### Changes
> This release should be public-API-compatible with version 6. The only public breaking change is the update of our React peer dependency from 16.4 to 16.8.4.
> 
> > Note: connect now uses React.memo() internally, which returns a special object rather than a function. Any code that assumed React components are only functions is wrong, and has been wrong since the release of React 16.6. If you were using PropTypes to check for valid component types, you should change from PropTypes.func to PropTypes.elementType instead.

---

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/reduxjs/react-redux/releases/tag/v7.0.1


types/react-redux/react-redux-tests.tsx
```diff
@@ -430,15 +430,6 @@ function mapDispatchToProps(dispatch: Dispatch) {
     };
 }
 
-connect(
-    mapStateToProps,
-    mapDispatchToProps
-)(Counter);
-
-@connect(mapStateToProps)
```
OliverJAsh: Because `connect` no longer returns a class, I guess it can't be used as a decorator.

types/react-redux/index.d.ts
```diff
@@ -93,10 +86,10 @@ export type GetProps<C> = C extends ComponentType<infer P> ? P : never;
 
 // Applies LibraryManagedAttributes (proper handling of defaultProps
 // and propTypes), as well as defines WrappedComponent.
-export type ConnectedComponentClass<
+export type ConnectedComponent<
     C extends ComponentType<any>,
     P
-> = ComponentClass<JSX.LibraryManagedAttributes<C, P>> & hoistNonReactStatics.NonReactStatics<C> & {
+> = NamedExoticComponent<JSX.LibraryManagedAttributes<C, P>> & hoistNonReactStatics.NonReactStatics<C> & {
```
OliverJAsh: `NamedExoticComponent` is the return type of `React.memo`, which `connect` uses.

OliverJAsh:
Build failed to some obscure reason, don't think it's related to this PR: https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/581686477#L247. What do I do now?

Nicolas-Vega:
Hi, After this MR on my project I'm having a few problems with some tsx file that have reference to components that are using Proptypes with connect of react-redux on JS. I have narrow it down to this change since v7.1.3 is working fine.

On all those cases I'm getting the same error, for example:
`Type '{ propName: ReactNode; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<RouteComponentProps<any, StaticContext, any>, never>, any, any>> & Readonly<Pick<RouteComponentProps<any, StaticContext, any>, never>> & Readonly<...>'.
  Property 'propName' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<Component<Pick<RouteComponentProps<any, StaticContext, any>, never>, any, any>> & Readonly<Pick<RouteComponentProps<any, StaticContext, any>, never>> & Readonly<...>'.  TS2322`


Example of this:
Detail.tsx:

```import React from "react";
import Header from "./Header";


type Props = {
  propName?: React.ReactNode;
};

class Detail extends React.PureComponent<Props> {
  render() {
    const {
      propName
    } = this.props;
    return (
      <>
        <Header backButton={propName} />
      </>
    );
  }
}

export default Detail;
```

Header.js:

```
import React from "react";
import PropTypes from "prop-types";
import { connect } from "react-redux";

class Header extends React.PureComponent {

  render () {
    return (<>Test</>
  };
}

Header.propTypes = {
  backButton: PropTypes.any,
};

export default withRouter(
  connect(
    state => ({
      user: state.login.user,
    }),
    {
      ...loginActions,
    }
  )(Header)
);
```

mikea:
As mentioned in #38678 this also breaks children property of components

OliverJAsh:
I'm on holiday until 7th October, so I can't take a look until then at the
earliest.

PR #38188, opened 2019-09-06
[@types/react] Allow `null` in HTML element `style` attribute.

calebmer:
React allows `null` in addition to `undefined` for CSS properties (see [`dangerousStyleValue()`](https://github.com/facebook/react/blob/21d79ce04066ade1098a7ae8990e596e01332a65/packages/react-dom/src/shared/dangerousStyleValue.js#L30)) but `csstype` properties are only optional.

I’m adding TypeScript to an untyped/Flow-typed codebase which uses `null` for style attributes a lot.

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -1645,7 +1645,7 @@ declare namespace React {
         placeholder?: string;
         slot?: string;
         spellCheck?: boolean;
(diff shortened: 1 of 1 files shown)


calebmer:
Ok, I thought this would be a trivial change but the tests show otherwise. Closing to rethink approach.

PR #38203, opened 2019-09-07
add @types/async-eventemitter

patarapolw:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/async-eventemitter/index.d.ts
+++ b/types/async-eventemitter/index.d.ts
@@ -17,17 +17,14 @@ import { EventEmitter } from "events";
  * - Interupt the callback chain in async listeners by calling the callback with the error as the first parameter; in sync listeners by throwing an Error.
  * @see https://www.npmjs.com/package/async-eventemitter#important-differences-between-asynceventemitter-the-native-eventemitter
  */
-export type AsyncListener<T, R> = (data?: T, callback?: (result?: R) => void) => Promise<R> | void;
-export interface EventMap {
-    [event: string]: AsyncListener<any, any>;
-}
+export = AsyncEventEmitter;
 
 /**
  * An EventEmitter that supports serial execution of asynchronous event listeners.
  * It also supports event listeners without callbacks (synchronous), as well as
  * interrupting the call-chain (similar to the DOM's e.stopPropagation()).
  */
-export default class AsyncEventEmitter<T extends EventMap> extends EventEmitter {
+declare class AsyncEventEmitter<T extends AsyncEventEmitter.EventMap> extends EventEmitter {
     /**
      * Executes all listeners for the event in order with the supplied data argument.
      * The optional callback is called when all of the listeners are done.
@@ -84,4 +81,11 @@ export default class AsyncEventEmitter<T extends EventMap> extends EventEmitter 
 
     getMaxListeners(): number;
     setMaxListeners(maxListeners: number): this;
+}
+
+declare namespace AsyncEventEmitter {
+    type AsyncListener<T, R> = (data?: T, callback?: (result?: R) => void) => Promise<R> | void;
+    interface EventMap {
+        [event: string]: AsyncListener<any, any>;
+    }
 }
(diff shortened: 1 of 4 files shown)


types/async-eventemitter/index.d.ts
```diff
@@ -0,0 +1,87 @@
...
+export interface EventMap {
+    [event: string]: AsyncListener<any, any>;
+}
+
+/**
+ * An EventEmitter that supports serial execution of asynchronous event listeners.
+ * It also supports event listeners without callbacks (synchronous), as well as
+ * interrupting the call-chain (similar to the DOM's e.stopPropagation()).
+ */
+export default class AsyncEventEmitter<T extends EventMap> extends EventEmitter {
```
uniqueiniquity: It seems like the [underlying library](https://unpkg.com/async-eventemitter@0.2.4/lib/AsyncEventEmitter.js) is not using `export default`, so this .d.ts file should not either. Instead, you should follow the appropriate template from [the TypeScript handbook](http://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-class-d-ts.html). Note that if you want to use `import name from 'package';` in your consuming code, you can use the `esModuleInterop` compiler setting.

types/async-eventemitter/index.d.ts
```diff
@@ -0,0 +1,87 @@
...
+ * that of the native EventEmitter. However there are some important differences which should be noted.
+ * - Data sent to event listeners (eg emit(data)) must always be zero or one argument, and can not be a function.
+ * - Event listeners will always recieve the data object, which may or may not be undefined.
+ * - The second argument can only be a callback, and will only be supplied if the event listener has an arity of two or more (eg function(e, next){}).
+ * - Event listeners with an arity of one or zero (eg without a callback argument specified) will be treated as synchronous.
+ * - Even if all event listeners are synchronous, they will still be executed asynchronously (through setImmediate) and thus code suceeding .emit() will be executed before any event listeners.
+ * - Interupt the callback chain in async listeners by calling the callback with the error as the first parameter; in sync listeners by throwing an Error.
+ * @see https://www.npmjs.com/package/async-eventemitter#important-differences-between-asynceventemitter-the-native-eventemitter
+ */
+export type AsyncListener<T, R> = (data?: T, callback?: (result?: R) => void) => Promise<R> | void;
```
uniqueiniquity: As discussed [here](http://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html), it's not recommended to have optional parameters in callbacks.

uniqueiniquity (changes requested):

types/async-eventemitter/index.d.ts
patarapolw: This will break,

```typescript
'premature-resolve': (data: number, resolve?: (result: any) => void) => Promise<any>;
...
ee.on('premature-resolve', async (data, resolve) => {
  if (resolve) resolve(true);
  'does not reach here';
});
```

That it should not require anything in resolve (default to `undefined`).

uniqueiniquity: Ah, I don't think I was clear.
If I make the following change, then everything works as expected, including the lines you've specified in your comment:
```ts
export type AsyncListener<T, R> = (data: T, callback: (result?: R) => void) => Promise<R> | void;
```

In this case, the callback I'm concerned about is the whole `AsyncListener`. The `?` on `result` is correct.

types/async-eventemitter/index.d.ts
```diff
@@ -0,0 +1,107 @@
...
+  eventNames(): Array<keyof T & string>;
+  listeners<E extends keyof T>(event: E & string): Array<T[E]>;
+  listenerCount(event: keyof T & string): number;
+
+  getMaxListeners(): number;
+  setMaxListeners(maxListeners: number): this;
+}
+
+declare namespace AsyncEventEmitter {
+    type AsyncListener<T, R> = (data?: T, callback?: (result?: R) => void) => Promise<R> | void;
```
uniqueiniquity: Also, the type `Promise<R> | void` doesn't really make sense. `void` only makes sense as the only return type of a function.

This should be the following:
```ts
type AsyncListener<T, R> = ((data: T, callback: (result?: R) => void) => Promise<R>) | ((data: T, callback: (result?: R) => void) => void);
```
That is, either it's listener returns nothing, or it returns something, which is a promise of `R`.

uniqueiniquity (changes requested):

uniqueiniquity (approved):

uniqueiniquity:

Issue #23002, opened 2018-01-17
Should `enum`s be `const enum`s?


TexAgg:
If you know how to fix the issue, make a pull request instead.

- [X] I tried using the `@types/firefox-webext-browser` package and had problems.
- [X] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [X] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [X] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @bomjacob 

If you do not mention the authors the issue will be ignored.

---

The only `enum` I used was [`ContextType`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3a8270c12cbaa59d973faeb95cb069d04e9cdaaa/types/firefox-webext-browser/index.d.ts#L4995), so I don't know if this is an issue with any other of the enums declared, but I had to change it to a `const enum` to get it to work. 
The [WebExtension API](https://developer.mozilla.org/en-US/Add-ons/WebExtensions/API/menus) expects a string for that entry, but the Typescript compiler converts an ordinary enum to an object and not a string literal(for example. `browser.contextMenus.ContextType.all` compiled to `browser.contextMenus.ContextType.all` originally but after I changed it to a const enum it compiled to `'all'`)
According to the [Typescript documentation for enums](https://www.typescriptlang.org/docs/handbook/enums.html), `const enum`s are completely inlined, so I assume they would be compiled into the form that the webextension API expects.

bomjacob:
Hmm, thanks a lot for this request.
I'll see what I can do to fix it (though I'm quite busy, but I'll try to do it sometime next week).

mhegazy:
You should not use `const enums` in a .d.ts **ever**. they break single file compilation scenarios. (i.e. compiling with `--isolatedModules`). 

Enums should be a declaration file iff a value with the same name and shape exists at runtime.

otherwise, use a union of literal types.

luckyrat:
Converting the `ContextType` enums to uppercase fixed it for me without needing to use const enums. Same story for `browser.notifications.TemplateType`.

I see no indication of uppercase naming on MDN or the Chrome docs so maybe this needs fixing in https://github.com/mozilla/webextension-polyfill ?

[...]

PR #38204, opened 2019-09-07
fix @types/googlemaps uses of MapTypeId | string

terencehonles:
MapTypeId as of #38084 is defined as a string enum, but all uses of MapTypeId were accidentally left as ``MapTypeId | string`` which looses type information.

I updated returns to be strictly MapTypeId, but input still seems to need to be ``MapTypeId | string`` because the TypeScript compiler seems to not recognize that a string with the proper value is a MapTypeId without casting it as such. Changing this would break all consumers and does not seem to make sense to do.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://developers.google.com/maps/documentation/javascript/reference/map#MapTypeId>

### DIFF
--- a/types/googlemaps/index.d.ts
+++ b/types/googlemaps/index.d.ts
@@ -49,7 +49,7 @@ declare namespace google.maps {
         getCenter(): LatLng;
         getDiv(): Element;
         getHeading(): number;
-        getMapTypeId(): MapTypeId | string;
+        getMapTypeId(): MapTypeId;
         getProjection(): Projection | null;
         getStreetView(): StreetViewPanorama;
         getTilt(): number;
@@ -149,7 +149,7 @@ declare namespace google.maps {
         /** The initial display options for the Map type control. */
         mapTypeControlOptions?: MapTypeControlOptions;
         /** The initial Map mapTypeId. Defaults to ROADMAP. */
-        mapTypeId?: MapTypeId;
+        mapTypeId?: MapTypeId | string;
         /**
          * The maximum zoom level which will be displayed on the map. If omitted, or
          * set to null, the maximum zoom from the current map type is used instead.


terencehonles:
This is related to #38084 and I'm a little confused to how TypeScript is handling the string enums. I removed the union with `string`, but when testing it against SebastianM/angular-google-maps#1648 I was getting errors that "roadmap" (one of the enum values) was not a `google.maps.MapTypeId`. I tried editing the type definition to use a `const enum`, but that doesn't seem to have helped and I noticed there was this issue: #23002 which suggests they should not be.

This change should make consumers receive a typed `google.maps.MapTypeId` when calling the library, but when specifying a value the compiler will accept a typed enum or a string which is what the code actually expects.

I'm not sure if tracking actual enum values for types in the TypeScript compiler is on the roadmap.

life777:
@terencehonles 
Hi, thanks.
You are right it definitely should be only MapTypeId.
Probably the same fix should be done for setting methods `setMapTypeId` and others for preventing bugs connected with settings misspelled map types. How do you think?

terencehonles:
@life777 I had done that initially, but then I had to cast strings to `MapTypeId` explicitly since it doesn't seem the TypeScript compiler was using the values I had provided in the PR.

As mentioned in my comment I tried using a `const enum` to see if that changed anything, but it didn't. To give an explicit example of what I mean:


```TypeScript
// generates a compiler error that string and MapTypeId are not compatible
const typeId: google.maps.MapTypeId = 'roadmap';

// works, but is cumbersome
const typeId: google.maps.MapTypeId = 'roadmap' as google.maps.MapTypeId;
```

If I were to change the `setMapTypeId` and the `MapOptions` to use strictly `MapTypeId` it would require casting strings wherever the value was used, and would cause unnecessary compilation errors for consumers of this type definition.

life777:
@terencehonles 
Yes, you are right. Even in their documentation they use this type union:
https://developers.google.com/maps/documentation/javascript/reference/map#Map.getMapTypeId 

But I think in this case it's better not to use a string value (where you can make a mistake), but google maps constants:
```javascript
const typeId: google.maps.MapTypeId = google.maps.MapTypeId.ROADMAP;
```
What do you think?

terencehonles:
I agree it would be ideal to use the constant values, but as the google maps package is loaded in an asynchronous manner that would require the constants to be resolved before use and it is likely that's not always common with consumers of the library (it's not the case in the library I was contributing to).

life777 (approved):
Yes, you are right.

orta:
👍 cool

PR #38206, opened 2019-09-08
explicit `any` value for P in React.ReactElement on material-ui

t0rr3sp3dr0:
Previous versions of React would set `any` as default value of `P` in `React.ReactElement` class. Newer versions do not have a default value for `P` and require an explicit value for it.

By replacing `React.ReactElement` with `React.ReactElement<any>` in `index.d.ts`, we can still use `material-ui` library with newer versions of React while keeping backwards compatibility with older versions.


t0rr3sp3dr0:
Actually it was a problem with the React definition that I was using, updating it solve the problem.

PR #38209, opened 2019-09-08
underscore: add CompiledTemplate interface, fix _.template typings

jgonggrijp:
This PR fixes the following problems with the `template` typings in underscore:

- There was no interface to represent the common return type of `_.template('some string')`, `_('some string').template()` and `_.chain('some string').template().value()`; an ad hoc type was written in each code location instead. The type in question is that of a compiled template, which still has to be executed.
- The ad hoc return type was incomplete and inaccurate. `(...arg: any[]) => string` was used while the returned function takes only one argument. The returned function also has a `source` property which wasn't declared.
- The ad hoc return type of `_.chain('some string').template()`, in addition, incorrectly mixed up the semantics of `_.chain` and `_.template`. The old typing assumed that `_.chain('some string').template()` would return a special type of compiled template that would wrap its return value in a new chain. This was probably based on an optimistic interpretation of the following sentences in the documentation of `_.chain`: "Returns a wrapped object. Calling methods on this object will continue to return wrapped objects until value is called". Reading the underscore source, however, makes it very clear that the wrapped object is *always* an instance of `_`. So the actual return value of `_.chain('some string').template()` is a *regular* compiled template, wrapped entirely inside a `_` instance.

My fixes are contained in 002b8784. The other commits introduce additional tests, as well as some linter improvements.

    - https://underscorejs.org/#template
    - https://underscorejs.org/#chain
    - https://underscorejs.org/docs/underscore.html (see bottom of page, scroll a bit up from there for the `_.template` implementation)
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header. *not applicable*
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed. *already present, I re-enabled a few rules.*


DanielRosenwasser:
Thanks!

PR #38213, opened 2019-09-08
@rails/actioncable: create version 6

renchap:
This is a copy of the `actioncable` type definitions. It as been renamed to `@rails/actioncable` with Rails 6. As per [the changelog](https://github.com/rails/rails/blob/6-0-stable/actioncable/CHANGELOG.md), the public API did not change.

I cleaned up a bit the `tsconfig.json`. `tslint.json` also now only disables the rules that are triggered by this type definition (the previous version disabled a lot of other rules).

If adding a new definition:
- [ ] `tslint.json` should be present and it shouldnt have any additional or disabling of rules. Just content as `{ "extends": "dtslint/dt.json" }`. If for reason the some rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]`  and not for whole package so that the need for disabling can be reviewed.

### DIFF
--- a/types/rails__actioncable/index.d.ts
+++ b/types/rails__actioncable/index.d.ts
@@ -1,0 +1,52 @@
+// Type definitions for ActionCable 6.0
+// Project: https://github.com/rails/rails/tree/master/actioncable/
+// Definitions by: Vincent Zhu <https://github.com/zhu1230>
+//                 Jared Szechy <https://github.com/szechyjs>
+//                 Renaud Chaput <https://github.com/renchap>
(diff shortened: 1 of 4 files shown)


uniqueiniquity (changes requested):
It appears from [the source](https://github.com/rails/rails/blob/master/actioncable/app/assets/javascripts/action_cable.js) that this is a UMD module. To make this consumable in browser and module contexts (e.g., Node), please follow [this template](http://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html).

PR #38226, opened 2019-09-08
Seamless Immutable: Add Optional Arguments to Array Map Callback

rharriso:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map#Parameters
- N/A If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- N/A If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

--- 

Example Code:

```js
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array
> const SI = require('seamless-immutable')
> SI.from(['zero', 'one', 'two']).map((word, index,arr) => `${index}: ${word}, ${arr.join('-')}`)
[ '0: zero, zero-one-two',
  '1: one, zero-one-two',
  '2: two, zero-one-two' ]
```

### DIFF
--- a/types/seamless-immutable/index.d.ts
+++ b/types/seamless-immutable/index.d.ts
@@ -108,7 +108,7 @@ declare namespace SeamlessImmutable {
         /** Custom implementation of the array functions, which return Immutable. */
         interface Overrides<T> {
             forEach(callbackfn: (value: Immutable<T>, index: number, array: Immutable<T[]>) => void, thisArg?: any): void;
-            map<TTarget>(mapFuction: (item: Immutable<T>) => TTarget): Immutable<TTarget[]>;
+            map<TTarget>(mapFuction: (item: Immutable<T>, index: number, array: Immutable<T[]>) => TTarget): Immutable<TTarget[]>;
             filter(filterFunction: (item: Immutable<T>) => boolean): Immutable<T[]>;
             slice(start?: number, end?: number): Immutable<T[]>;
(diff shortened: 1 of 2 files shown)


types/seamless-immutable/index.d.ts
```diff
@@ -108,7 +108,7 @@ declare namespace SeamlessImmutable {
         /** Custom implementation of the array functions, which return Immutable. */
         interface Overrides<T> {
             forEach(callbackfn: (value: Immutable<T>, index: number, array: Immutable<T[]>) => void, thisArg?: any): void;
-            map<TTarget>(mapFuction: (item: Immutable<T>) => TTarget): Immutable<TTarget[]>;
+            map<TTarget>(mapFuction: (item: Immutable<T>, index?: number, array?: Immutable<T[]>) => TTarget): Immutable<TTarget[]>;
```
pheromonez: `index` and `array` should not be marked as optional parameters, as function callback parameters are already optional by default. If `index` and `array` are marked as optional then when we use them in the mapping-function their types will be `number | undefined` and `Immutable<T[]> | undefined`, which is not the behaviour we want.

To demonstrate the above behaviour, consider the `forEach` function, which allows us to  omit index and array, without needing to mark those parameters as optional:
```
arr.forEach((value) => console.log({value})
```

pheromonez (approved):

PR #38228, opened 2019-09-09
Fixes #35572, and the problem that using useRef with ref={} currently yields error

### LINKED ISSUE #35572: [@types/react] useRef return type clashes with ref prop type
- [x] I tried using the `@types/16.8.17` package and had problems.

```
Type 'MutableRefObject<HTMLDivElement | undefined>' is not assignable to type 'string | ((instance: HTMLDivElement | null) => void) | RefObject<HTMLDivElement> | null | undefined'.
  Type 'MutableRefObject<HTMLDivElement | undefined>' is not assignable to type 'RefObject<HTMLDivElement>'.
    Types of property 'current' are incompatible.
      Type 'HTMLDivElement | undefined' is not assignable to type 'HTMLDivElement | null'.
        Type 'undefined' is not assignable to type 'HTMLDivElement | null'.
```

staeke:
Fixes #35572, and the problem that using 

```jsx
const a = useRef<HTMLDivElement>(); 
return <div ref={a}/>
```

currently yields error

Please fill in this template.

- [ x ] Provide a URL to documentation or source code which provides context for the suggested changes: https://reactjs.org/docs/hooks-reference.html#useref

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -20,6 +20,7 @@
 //                 Saransh Kataria <https://github.com/saranshkataria>
 //                 Kanitkorn Sujautra <https://github.com/lukyth>
 //                 Sebastian Silbermann <https://github.com/eps1lon>
+//                 Staffan Eketorp <https://github.com/staeke>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
@@ -77,7 +78,7 @@ declare namespace React {
     type Key = string | number;
 
     interface RefObject<T> {
-        readonly current: T | null;
+        readonly current: T | null | undefined;
     }
 
     type Ref<T> = { bivarianceHack(instance: T | null): void }["bivarianceHack"] | RefObject<T> | null;
--- a/types/react/test/hooks.tsx
+++ b/types/react/test/hooks.tsx
@@ -63,11 +63,11 @@ export function App() {
     const birthdayRef = React.useRef<FancyButton>(null);
 
     React.useLayoutEffect(() => {
-        if (birthdayRef.current !== null) {
+        if (birthdayRef.current != null) {
             birthdayRef.current.fancyClick();
         } else {
             // this looks redundant but it ensures the type actually has "null" in it instead of "never"
-            // $ExpectType null
+            // $ExpectType null | undefined
             birthdayRef.current;
         }
     });


eps1lon (changes requested):
My issue with this change is that the overloaded ref type is only encountered before the ref is initialized. With this change the "type lifecycle" becomes `undefined - HTMLDivElement - null (- HTMLDivElement - null)*`. Since most of the use cases will not special case the very first cycle (`undefined`) I think it's fine to require explicit type annotation (`HTMLDivElement | undefined`) if you're interested in it or require `const divRef = useRef<HTMLDivElement>(null)`. 

As it stands right now this would be a breaking change if previous code strictly compared to `null` (`current === null`) since this would only narrow down to `HTMLDivElement | undefined`.

staeke:
Thanks for reviewing @eps1lon .

I agree and disagree. Since the change is potentially breaking I agree we should be bumping the major. I don't think status quo is acceptable though. But I do think we can reach a compromise that provides reasonable results for all use cases

##### 1.
First, and to the point with this PR - I don't think it's acceptable to require actual code changes from users (as opposed to typing changes) to accommodate ts issues - I'm thinking about adding a `null` parameter to `useRef`. That goes against the fundamental idea with typescript. Then, to your first suggestion. Writing:

```tsx
const ref = useRef<HTMLDivElement|undefined>()
return <div ref={ref}/>
```

...yields an error, so I don't see how your proposed solution works. Also, I personally think it's dirty to require `|undefined` for such a common use case to be sprinkled everywhere.

But, I'm all fine with changing `RefAttributes` and `ClassAttributes` instead, like below (`undefined` added): 

```ts
interface RefAttributes<T> extends Attributes {
    ref?: Ref<T | undefined>;
}
interface ClassAttributes<T> extends Attributes {
    ref?: LegacyRef<T | undefined>;
}
```

Would you be ok with that?

##### 2.
Secondly, I don't think it's acceptable to see that a textbook example yields a typescript error, and one that is also not correct. I'm thinking about:

```tsx
const ref = useRef()
return <div ref={ref}/>
```

I think this use case is best solved by having useRef default its type parameter to `any`. Which, I argue, is fundamentally what we know here. Not using a type parameter, should result in weaker typing, but not compile failure. But, not addressed in this PR. I could though.

##### 3.
Third, I think the `null` overload of `useRef` returning a non-mutable ref object is an error. It's just as mutable and we should change it. Do you agree?

staeke:
I have yet another problem with not including `undefined` with `useRef` being used without parameters; it simply doesn't reflect how React works. The output of `useRef` in reality **IS** `undefined` if you don't pass parameters. Hiding this fact is just producing worse typings, and more bugs. I think we have two options, either require `undefined` to be part of the type if no parameters are passed (seems overly unsmart to me). Or, simply, draw the conclusion, that it has to be part of the return type. Take the example:

```ts
const n = useRef<number>()
n.current.toFixed()
```

Currently yields no ts error. Breaks at runtime though

Now that could be fixed by amending `MutableRefObject` like I did, or e.g. change the `useRef` signature. Maybe that's better. How about:

```ts
function useRef<T>(): MutableRefObject<T|undefined>;
```

And, to be clear, and summarize it all, this would my preferred overloads:

```ts
function useRef<T = any>(): MutableRefObject<T|undefined>;
function useRef<T>(initialValue: T|null): MutableRefObject<T;
```

Jessidhia:
There is a reason why the overloads are like they are.

Maybe I need to write a lot more comments in the `.d.ts` files to elaborate; this is something that comes up time and again, and the problem is always the same.

First, `RefObject`  can **never** hold an `undefined` when you create it with `createRef`. It's initialized with `null`, and you are never supposed to write into it.

As for the overloads, here are the current ones:

```ts
function useRef<T>(initialValue: T): MutableRefObject<T>;
function useRef<T>(initialValue: T|null): RefObject<T>;
function useRef<T = undefined>(): MutableRefObject<T | undefined>;
```

The first overload is what you should hit whenever your initial value is sufficient to infer the generic, or when it is directly compatible with the explicit generic. This is what will happen with, for example, `useRef<0>(0)`, as useless as that is. Or `useRef<number>(0)`, which is the same as what's inferred by `useRef(0)`. This is the common case.

The second overload is **the** convenience overload when you want something be passed as a `ref={}` value. `useRef<RefType>(null)`. It's explicitly _not_ a mutable object because its intended use is to be written to by React. From your point of view, `.current` is a read-only value. If you want to make a mutable version of it, then just specify `| null` on your generic like in `useRef<RefType | null>(null)` to hit the first overload.

The third overload is there when you give no arguments (i.e. the "first argument" is `undefined`), and it's what'd be used by `useRef<Foo>()`. This of course prevents you from giving it as a `ref={}` because React will write `null` into it when the ref-ed component unmounts, but never write `undefined`. You can also just specify `useRef<Foo | null>()` here, which does _exactly_ what you ask for in your comments, but if you pass it to a `ref={}` the `.current` will only ever be `undefined` on the initial render pass; React will set it to `Foo | null` exclusively once it has control over the `ref={}`.

As I said in all cases, unless you want to use **the** `ref={}` convenience overload, then include the `| null` in your generic argument. Because React _wants_ to write `null` into it, not specifying the `null` yourself would make the type incorrect. On the other hand, if you let the code assume a `null` can be written there on all cases, like on this PR, then you create an overly-wide and less safe type when you use it as mutable fiber-local storage and not as a `ref={}` argument, and specify an outright unreachable type when used with `createRef()`.

mikew:
@Jessidhia So how does one create a ref object, with a generic argument, an initial value of null, and have TS complain when there's improper usage? I'd expect a TS error in this:

```ts
const ref = useRef<string>(null)
console.log(ref.current.toUpperCase())
```

But it happily compiles: https://codesandbox.io/s/sparkling-firefly-e6fjk

staeke:
@mikew good point. Can you pls add it in the other thread - the open PR at https://github.com/DefinitelyTyped/DefinitelyTyped/issues/38260

eps1lon:
> @Jessidhia So how does one create a ref object, with a generic argument, an initial value of null, and have TS complain when there's improper usage? I'd expect a TS error in this:
> 
> ```ts
> const ref = useRef<string>(null)
> console.log(ref.current.toUpperCase())
> ```
> 
> But it happily compiles: [codesandbox.io/s/sparkling-firefly-e6fjk](https://codesandbox.io/s/sparkling-firefly-e6fjk)

codesandbox has like `strictNullChecks` disabled. The type checker reports and error with `strictNullChecks: true`: http://www.typescriptlang.org/play/index.html#code/JYWwDg9gTgLgBAbzgVwM4FMBK6BmcC+cOUEIcARFOgIYDGM5A3ALABQbOyAdvcBF3AAq6VDAAUASkRs4KDNhwAeUVGBcA5gD4xXZABs9EgHS1kUKlxhGYEAKpgw6KAGFqGSS1ay0WXMpiqGtrGpuboltZ2Dk6u7hKeMnK+Sipq6nAAPihcACa4aug52tx5OAU5IWYWVjb2ji5u6B6JPgr+gelZJflchcEmVeE1UfWxTfFsLfJ+qRqZcLoG2ouGA2ERtdENcZ7e0ykBafMr-aHVkXUxjc2s+GxAA

karimuddin-tech:
`React.useRef<any | null>();` this is work for me

PR #38240, opened 2019-09-09
[yup] Fix locale object conditionals

### LINKED ISSUE #38231: `never` keyword breaks Yup `LocaleObject`

- [ x ] I tried using the `@types/yup` package and had problems.
- [ x ] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [ x ] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/). 
- [ x ] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @....

@kalley 

Thanks for the update to the `@types/yup` type definitions. 

Can you please check the `LocaleObject` type definitions? If one creates a `LocaleObject` by taking the contents of the default `locale.js` as a starter and then tries to apply the `LocalObject` type to it there is a problem with the use of the `never` keyword:

```
export const LocaleValidationMessages: LocaleObject = {
    mixed: {
        default: '${path} is invalid',
        required: '${path} is a required field',
        oneOf: '${path} must be one of the following values: ${values}',
        notOneOf: '${path} must not be one of the following values: ${values}',
        notType: ({ path, type, value, originalValue }) => {
            const isCast = originalValue !== undefined && originalValue !== null && originalValue !== value;
            let msg =
                `${path} must be a \`${type}\` type but the final value was: \`${ValueFormatter.printValue(value, true)}\`` +
                (isCast ? ` (cast from the value \`${ValueFormatter.printValue(originalValue, true)}\`).` : '.');

            if (value === null) {
                msg += `\n If "null" is intended as an empty value be sure to mark the schema as \`.nullable()\``;
            }

            return msg;
        }
    },
    string: {
        length: '${path} must be exactly ${length} characters',
        min: '${path} must be at least ${min} characters',
        max: '${path} must be at most ${max} characters',
        matches: '${path} must match the following: "${regex}"',
        email: '${path} must be a valid email',
        url: '${path} must be a valid URL',
        trim: '${path} must be a trimmed string',
        lowercase: '${path} must be a lowercase string',
        uppercase: '${path} must be a upper case string'
    },
    number: {
        min: '${path} must be greater than or equal to ${min}',
        max: '${path} must be less than or equal to ${max}',
        lessThan: '${path} must be less than ${less}',
        moreThan: '${path} must be greater than ${more}',
        positive: '${path} must be a positive number',
        negative: '${path} must be a negative number',
        integer: '${path} must be an integer'
    },
    date: {
        min: '${path} field must be later than ${min}',
        max: '${path} field must be at earlier than ${max}'
    },
    boolean: {
        // NOOP
    },
    object: {
        noUnknown: '${path} field cannot have keys not specified in the object shape'
    },
    array: {
        min: '${path} field must have at least ${min} items',
        max: '${path} field must have less than or equal to ${max} items'
    }
};
```
Specifically, on `mixed.default` and `string.marches`. This has been working pre the merge of the new changes.

See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37648

Many thanks

JT

kalley:
Fixes #38231

Was going to use `Omit<MixedSchema, 'default'>`, but didn't want to bump the required TS version even higher, which is why I just added the override to the default method.

The `Extract` pulls out the match if there is one in a union, so we can check it. If it just matches, then we're already ready to go. Might at some point make sense to invert some of the logic here at some point. So, we'd have a `LocaleObject` with the keys supported defined, then in each method, the message argument or property (in the case of `string.matches`) could just use that there. If anyone thinks that would be worth doing now, I'm more than happy to update this PR to follow that pattern.

Please fill in this template.

### DIFF
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -13,7 +13,7 @@
 //                 Maurice de Beijer <https://github.com/mauricedb>
 //                 Kalley Powell <https://github.com/kalley>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 3.1
+// TypeScript Version: 2.8
 
 export function reach<T>(schema: Schema<T>, path: string, value?: any, context?: any): Schema<T>;
 export function addMethod<T extends Schema<any>>(
@@ -69,8 +69,8 @@ export interface Schema<T> {
     default(value: any): this;
     default(): T;
     typeError(message?: TestOptionsMessage): this;
-    oneOf(arrayOfValues: Array<T | Ref | null>, message?: TestOptionsMessage<{ values: T | Ref }>): this;
-    notOneOf(arrayOfValues: any[], message?: TestOptionsMessage<{ values: T | Ref }>): this;
+    oneOf(arrayOfValues: Array<T | Ref | null>, message?: MixedLocale['oneOf']): this;
+    notOneOf(arrayOfValues: any[], message?: MixedLocale['notOneOf']): this;
     when(keys: string | any[], builder: WhenOptions<this>): this;
     test(
         name: string,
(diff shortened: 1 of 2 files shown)


types/yup/index.d.ts
```diff
@@ -431,15 +431,21 @@ export interface FormatErrorParams {
 export type LocaleValue = string | ((params: FormatErrorParams) => string);
 
 type MessageFromParameters<P extends unknown[]> = {
-    [K in keyof P]: P[K] extends TestOptionsMessage<any> ? P[K] : never;
+    [K in keyof P]: Extract<P[K], TestOptionsMessage<any>> extends TestOptionsMessage<any>
+        ? Extract<P[K], TestOptionsMessage<any>>
+        : never;
```
sseppola: Any way you can demonstrate this in a test? The difference is not clear to me

kalley: The only reason I needed this was because of the union in `string.matches`. Without the `Extract`, it would fail because that argument doesn't match only `TestOptionsMessage`. I'm not sure how to show that in a test besides somehow negating the change. This is part of why I'm considering inverting the logic and making the message argument in methods pull from `LocaleObject` instead of building `LocaleObject` from the schemas.

sseppola: ok

sseppola:
LGTM, any input @YashdalfTheGray?

kalley:
I went ahead and changed the strategy for defining the `LocaleObject`. I think it's a bit easier to reason about and it makes it so if someone is using TS 2.8, they can still use the definitions.

orta:
Cool, yep, this all looks good to me 👍

PR #38241, opened 2019-09-09
Fix type for hls.js audio track

drax98:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/video-dev/hls.js/blob/master/src/types/track.ts

### DIFF
--- a/types/hls.js/index.d.ts
+++ b/types/hls.js/index.d.ts
@@ -4,6 +4,7 @@
 //                 Johan Brook <https://github.com/brookback>
 //                 Adrián Caballero <https://github.com/adripanico>
 //                 Alexey I. Berezin <https://github.com/beraliv>
+//                 Arkadiusz Babiarz <https://github.com/drax98>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.4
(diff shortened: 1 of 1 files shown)


adripanico (changes requested):
What is the sense of having an interface that is not being used anywhere?

drax98:
> What is the sense of having an interface that is not being used anywhere?

Before this change, there was used native AudioTrack interface from HTML media elements, which is not what hls.js returns. Adding new interface fixed this problem. 

When you view the whole file you can see where this interface is used.

adripanico (approved):

types/hls.js/index.d.ts
```diff
@@ -4,6 +4,8 @@
 //                 Johan Brook <https://github.com/brookback>
 //                 Adrián Caballero <https://github.com/adripanico>
 //                 Alexey I. Berezin <https://github.com/beraliv>
+//                 Alexey I. Berezin <https://github.com/beraliv>
```
minestarks: This line is duplicated.

minestarks (changes requested):

minestarks (approved):

PR #38242, opened 2019-09-09
Updates pnpapi to add extend process.versions

arcanis:
The PnP hook adds a specific entry within `process.versions` to allow userland to detect whether the `pnpapi` package can be required or not. Since `process.versions` doesn't have an index (#37754), it needs to be extended.


arcanis:
How should I reference `process.versions` without using `types`?

PR #38250, opened 2019-09-09
[angular-file-saver] add module name export

Igorbek:
Starting from version 1.0.3, `angular-file-saver` exports module name: https://github.com/alferov/angular-file-saver/commit/da6ae5c2555795f466d646f86ef73aae56aa1ba6

cc @deenairn


deenairn (commented):
Changes look good to me, thanks Igor.

Issue #31065, opened 2018-12-04
[@types/react] RefObject.current should no longer be readonly


bschlenk:
It is now okay to assign to `ref.current`, see example: https://reactjs.org/docs/hooks-faq.html#is-there-something-like-instance-variables

Trying to assign a value to it gives the error: `Cannot assign to 'current' because it is a constant or a read-only property.`

- [x] I tried using the `@types/react` package and had problems.
- [x] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @johnnyreilly @bbenezech @pzavolinsky @digiguru @ericanderson @morcerf @tkrotoff @DovydasNavickas @onigoetz @theruther4d @guilhermehubner @ferdaber @jrakotoharisoa @pascaloliv @Hotell @franklixuefei

ferdaber:
A PR would be appreciated! It should be a very easy one line change 😊

jagatstha:
@ferdaber, 
Hi, I like to pick this issue. As I am new to typescript (just a week) and it will be my first contribution to open source. 

I have gone through node_modules/@types/react/index.d.ts and there is the following code at line no 61 which declare current as readonly.

```
interface RefObject<T> {
        readonly current: T | null;
    }
```
if this is the issue, then I can solve. can you guide for my first PR?
 Thank you for your time :)

ferdaber:
Yes, it's just removing the `readonly` modifier on that line.

pshrmn:
`React.useRef` returns a `MutableRefObject` whose `current` property is not `readonly`. I suppose the question is whether the ref object types should be unified.

ferdaber:
They should be unified, the React runtime has no limitation on the `current` property created by `React.createRef()`, the object itself is sealed but not frozen.

ferdaber:
That's fair, I've been using `React.createRef()` as a helper function to just create a pointer, as React won't manage it unless it's passed in as a `ref` prop, but you can just as well create a simple pointer object with the same structure.

Kovensky:
[...]

PR #38260, opened 2019-09-10
Fixes #35572 Altered useRef for discussions

### LINKED ISSUE #35572: [@types/react] useRef return type clashes with ref prop type

staeke:
Continuation of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38228

Please fill in this template.

- [ x ] Provide a URL to documentation or source code which provides context for the suggested changes: https://reactjs.org/docs/hooks-reference.html#useref


types/react/index.d.ts
```diff
@@ -914,21 +915,20 @@ declare namespace React {
...
      *
-     * Usage note: if you need the result of useRef to be directly mutable, include `| null` in the type
-     * of the generic argument.
-     *
      * @version 16.8.0
      * @see https://reactjs.org/docs/hooks-reference.html#useref
      */
     // TODO (TypeScript 3.0): <T extends unknown>
-    function useRef<T>(initialValue: T|null): RefObject<T>;
+    function useRef<T = any>(initialValue: null): MutableRefObject<T|null>;
```
Jessidhia: Nope, this makes the type safety _worse_ for newer typescript versions.

types/react/index.d.ts
```diff
@@ -942,7 +942,7 @@ declare namespace React {
      * @see https://reactjs.org/docs/hooks-reference.html#useref
      */
     // TODO (TypeScript 3.0): <T extends unknown>
-    function useRef<T = undefined>(): MutableRefObject<T | undefined>;
+    function useRef<T = any>(): MutableRefObject<T | undefined | null>;
```
Jessidhia: The `= undefined` is there for a reason, it's to stop it from defaulting to `unknown`. Making it default to `any` is _much, much worse_.

types/react/index.d.ts
```diff
@@ -92,7 +93,7 @@ declare namespace React {
         ref?: Ref<T>;
     }
     interface ClassAttributes<T> extends Attributes {
-        ref?: LegacyRef<T>;
+        ref?: LegacyRef<T | undefined>;
```
Jessidhia: This makes writing callback refs more difficult because you now have to check for an unreachable value.

Jessidhia (changes requested):
See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38228#issuecomment-529749802 again.

Don't alter the types to produce weaker types, or types that make it _harder_ to write type-safe code, in order to make it easier to port type-unsafe code or to keep type-unsafe code writing habits. Perhaps you want to disable `--strictNullChecks` instead?

The change to the generic default is **especially** dangerous because that is the one implicit `any` that is not flagged by `--noImplicitAny`.

staeke:
So @Jessidhia, this is a continuation of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38228

First, I have not bothered to fix tests yet in this PR. If you lo and behold would accept it, I would of course fix them. But I wanted to continue our discussion, and since you closed the PR I had to create a new one. Please note that the code/branch has changed. 

I felt very rudely treated with you just closing the PR, and your condescending attitude. If you have your arguments, and I have mine (and I'm sure the community has more), let's talk this out without resorting to "This is wrong. Close". Feel free to disagree and not merge, but welcoming feedback and hearing people out is the very foundation to getting community contributions and a better results in the end.

Secondly, the problem isn't that I don't understand you. The problem is that I (and I'm sure others) disagree with you.

If you ask me, the reason this issue keeps popping out is likely not that you have provided too little info in the comments, but that the current typings are confusing.

Anyhow, to the discussion at hand...let's look at some potential real-life examples to talk about the goals/tradeoffs/pros/cons:

For reference, also see Typescript Design Goals - https://github.com/Microsoft/TypeScript/wiki/TypeScript-Design-Goals

```ts
// 1. Breaks with yours. Passes with mine. Along with Typescript's design goals I 
// argue that this should pass. It's a standard textbook legitimate usage. No fishy 
// behavior is caught that we should flag. Now we can't infer types so ideally 
// ref.current should be "any"
const ref = useRef()
const div = <div ref={ref}/>

// 2. Breaks with yours. Passes with mine. I argue that the first thing someone 
// will do after seeing example 1 is to get a little more typing. Thus, the natural 
// thing is to add the element type. 
const ref = useRef<HTMLDivElement>()
const div = <div ref={ref}/>

// 3. Breaks with yours. Passes with mine. I believe the argument that 2. should 
// have an extra | undefined is sound, but not pragmatic. However, if you went 
// that route you'd get below. Again I believe it's crazy that this doesn't pass
const ref = useRef<HTMLDivElement|undefined>()
const div = <div ref={ref}/>

// 4. Breaks with yours. Passes with mine. This is like 1. pretty standard javascript 
// and it's very hard to understand why somebody would fail it. Now, the error 
// you're getting is can't assign the object to null, which probably leads some 
// people to type it like 5, especially since a lot of other typings around ref, seems 
// to add null anyhow
const a = useRef(null)
a.current = {foo:'bar'}

// 5. Breaks with yours. Passes with mine. I don't have a strong issue with requiring
//  | null, other than that I think there is no consistency between how we look at 
// refs. But...the current error is very confusing: Can't assign to read-only property. What?!
const a = useRef<{foo:string}>(null)
a.current = {foo:'bar'}
```

Regarding 5, with the increased usage of refs (not only for elements) with hooks, we cannot simply infer if they should be used as a mutable ref or immutable. Therefore it has to be mutable.

Looking forward to a sane discussion, rather than "I'm right. You're wrong. Closing"

types/react/index.d.ts
staeke: You have to write 

```tsx
const ref = useRef<HTMLDivElement>()
let myVal:HTMLDivElement|undefined|null
const div = <div ref={v => myVal = v}/>
```

That's not very hard, and above all, it's valid, since myVal can be null/undefined. If what you're complaining about is this:

```tsx
const ref = useRef<HTMLDivElement>()
let myVal:HTMLDivElement|null = null
const div = <div ref={v => myVal = v}/>
```

...now getting an error. Then yes, that's a flaw. If that's your beef that seems totally solvable by a better structure around the different ref types (LegacyRef, RefObject, function etc.)

types/react/index.d.ts
staeke: Your arguments are where exactly? I gave mine below

types/react/index.d.ts
staeke: See discussion below around the examples. Mutability is something we can't say anything about given modern usages

staeke:
@Jessidhia You were fast.

So fast I weren't able to write the first comment. Didn't want to pollute the PR text. Anyhow, please see my examples, and tell

staeke:
Very fast @Jessidhia 

I didn't even have the time to provide my first comment. I apologize, I could probably have written it as part of the original description to avoid confusion.

Feel free to address with arguments, and reason around the examples I gave (or add more of your own)

types/react/index.d.ts
staeke: Give me an example where it defaults to `unknown`. I gave some, imo, false positives below

staeke:
That said, I do think you have a point about strict null checks and ideals! If you turn them on, you're basically asking for harder control, so I buy the not overly leaning to any approach

types/react/index.d.ts
staeke: Could be fixed e.g. with 
```ts
type LegacyRef<T> = string | Ref<T> | {current:undefined}
```

eps1lon:
> I felt very rudely treated with you just closing the PR, and your condescending attitude.

@staeke I do not share this impression. @Jessidhia's comment was very detailed trying to give you as much context as needed to understand the current typing. Please be considerate that we don't always know how much knowledge/context a PR author has. If a maintainer added more details than you think you needed then it's never with the intention to belittle you but also because it can be referenced in the future or helps contributors who only read a PR. The reason this was closed so quickly is that this was already discussed before (https://github.com/DefinitelyTyped/DefinitelyTyped/issues/31065) and everybody seemed to be fine with it. Considering this PR allowed weaker types in order to write less closing it immediately might've been a little too fast in hindsight but nothing out of the ordinary.

That being said it would probably help to not discuss this purely theoretically but add concrete examples to the test file that would fail with the new change. This would help explain the behavior to any future PR author. We can then weight those breaking changes against benefits. As far as I understand this change this only allows 

```diff
-const divOrNullishRef = React.useRef<HTMLDivElement>(null);
+const divOrNullishRef = React.useRef<HTMLDivElement>();
```
where the new change has the drawback of widening the type of `ref.current`. It might seem that allowing the second case is "obvious" to you but there are still people who use `undefined` and `null` distinctly and don't use weak equality for `null`. This is especially ironic since react internally does not write `undefined` or compares nullish types weak but strict (`===` instead of `==`).

Jessidhia:
Yeah, closing it might not have been the best call, sorry for the bluntness.

The problem with the widening comes when it's time to actually use the refs, or to actually guard against their misuse when you create a ref for one purpose and accidentally reference it for a different purpose. It's to help catch errors.

Going example by example:

1.

```tsx
// type of `ref.current` becomes `any`
const ref = useRef()
// compiler accepts, type of `food` is non-implicit `any`
const food = ref.current.potato()
// without the `T = any` change, `.current` becomes `unknown` instead,
// which would have caught the error in `ref.current.potato`.
```

2.

```tsx
const ref = useRef<HTMLDivElement>()
const div = <div ref={ref}/>
// will be undefined on initial mount, but then never again
// even if div is unmounted, it'll reset to null, not undefined
const { current } = ref
useEffect(() => {
  const { current } = ref
  if (current === undefined) {
    // unreachable code, but can't be detected anymore,
    // even without passing --allowUnreachableCode
    console.log('unreachable')
  }
  if (current !== null) {
    // current is now HTMLDivElement | undefined, not HTMLDivElement
    // so you have to check for current !== undefined too which, as I said before,
    // should have been unreachable
    current.nodeType // false positive error
  }
}, [])
```

3. is a correct error, and not the argument I had made; my argument was with being explicit about `| null`. The `undefined` is implied by the initial value itself being omitted.

```tsx
const ref = useRef<HTMLDivElement|undefined>()
const div = <div ref={ref}/>
// is undefined on first render pass, fine
// but if you mount and unmount div, it'll become null, and you didn't specify null
const { current } = ref
```

4. is definitely the correct behavior; it doesn't even go into any of the convenience overloads, it's handled by the first one. You need to tell the compiler at definition time what other types of values you expect to see.

```tsx
const a = useRef(null)
a.current = {foo:'bar'}
```

5. is because the absence of the `| null` indicates the use of the convenience overload for `ref={}`s. Now, as for why it's read-only on that overload, it has to do with the concepts of ownership and "out parameters", which you might see discussed in Microsoft/TypeScript as "covariance/contravariance". We don't have a borrow checker so marking as `readonly` from our point-of-view is the safe way to say it's written into by other code, not by us. There's no way to make the fact that something was passed into a `ref={}` modify its behavior like it would in, say, Rust; so the opinionated second overload is the only way to retain some safety. The `| null` you _can_ add when invoking it opts out of the extra safety of the `readonly`; but other than with this specific overload setup, there is no way to optionally _add_ safety.

```tsx
const a = useRef<{foo:string}>(null)
a.current = {foo:'bar'}
```

See this for an example of "out parameters" in another language: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/out-parameter-modifier

---

It's true that some of the pain points of using `useRef` have to do with limitations of TypeScript itself. For example, earlier on, I mentioned covariance/contravariance. TypeScript currently has no way to signal that the `ref` prop, in React, is contravariant from the point of view of the caller.

In simpler terms,

```tsx
const a = useRef<HTMLElement>(null)
const div = <div ref={a}/>
```

is _correct and safe_ code, but TypeScript doesn't know that the point of the `ref={}` argument is for React to write an `HTMLElement | null` into it, it thinks we're trying to pass an `HTMLElement | null` and it doesn't like that because, covariantly, an `HTMLElement` is not an `HTMLDivElement`.

I don't think there's any good solution for this until TypeScript implements some way of indicating "out" parameters or other kinds of covariance relationships. There is a hack that can be done with function refs (the `bivarianceHack` trick you can see in the `.d.ts`) but it's still the wrong type relation, it just doesn't affect DOM types in particular.

staeke:
So thanks for your replies @Jessidhia and @eps1lon

I tend to agree with you on some matters but not others. I understand what happens everywhere, and basically why it's typed the way it is (I think). I just disagree about the current inconvenience and the tradeoffs chosen. 

I understand that some suggestions get buried in different places and threads, but to reiterate - my main issue is something that would be fixed with 

```ts
interface ClassAttributes<T> extends Attributes {
    ref?: LegacyRef<T> | Ref<undefined>;
}
```

I don't see a lot of downsides with that? So, what do you say about that? This would fix 90% of my concerns.

##### Comment regarding 2.
Sure, but you're making it easy for yourself. Why would you only have to deal with types after first render (`useEffect`). Surely you could have code that uses the ref before that. 

### Confusing API
For the last 10%, I seem to not have fully conveyed what I think is my second issue really, namely the (according to me) unintuitive typings:

> 5. is because the absence of the | null indicates the use of the convenience overload

Now I understand that's your goal. However, I think the idea of idea of having `null` passed to indicate "convenience" is non-intuitive and not very convenient at all unless you really study the typings. I understand that refs in practice are intended to be written to either by user code exclusive or React, and although I think it's honorable to try and differ between these use cases, I don't think you can do it with enough precision with typings. Thus you end up with many false positives and workaround hacks. To the point where I think this "convenience" should be removed. But I guess that's where we differ philosophically.

staeke:
Oh and one more comment :)

Since I'm guessing you will probably still disagree with me, can we at least make the comments around `useRef` clearer. I'm thinking something like:

```ts
/**
* This overload is for setting the ref attribute on an element. 
* If you want another use case, please use useRef<MyType|null>(null)
*
* ...
*/
function useRef<T>(initialValue: T|null): RefObject<T>;

/**
*  This overload is NOT intended for for setting the ref attribute on an element.
*  If you want that use case, please pass null: e.g. useRef(null)
*
* ...
*/
function useRef<T = undefined>(): MutableRefObject<T | undefined>;
```

staeke:
Just wanted to clarify another thing: As for defaulting to `any` for `useRef` I think a good comparison is the built-in typing for `Map` (regarding case 1. above). Doing this does **not** yield a typescript error

```ts
const map = new Map()
map.get('a').potato()
```

To me that's analogous to 

```ts
const ref = useRef()
const food = ref.current.potato()
```

Now `Map` could have been typed to not default to `any`, but requiring types. That's a tradeoff between supporting plain javascript and catching all errors. Typescript tries to hit a middle ground. And my objection on that point is that I think these typings (for defaulting to `any`) deviates too much from Typescript philosophy

PR #38275, opened 2019-09-10
Added type definitions for "ineum"

evolkmann:
ineum is the global object used by the Instana Website Monitoring API (https://docs.instana.io/products/website_monitoring/api/) to collect different kinds of information from websites/apps.

Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/ineum/ineum-tests.ts
+++ b/types/ineum/ineum-tests.ts
@@ -1,5 +1,3 @@
-/// <reference types="./index" />
-
 // $ExpectType void
 ineum("key", "key");
 // $ExpectType void
--- a/types/ineum/ineum-tests.ts
+++ b/types/ineum/ineum-tests.ts
@@ -1,4 +1,4 @@
-/// <reference path="index.d.ts" />
+/// <reference types="./index" />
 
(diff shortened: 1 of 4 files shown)


types/ineum/ineum-tests.ts
```diff
@@ -0,0 +1,35 @@
+/// <reference types="./index" />
```
minestarks: I think you can remove this line.

evolkmann: Hi @minestarks, I included it because the `index.ts` is not exporting a module, therefore the testfile did not find the types via the `tsconfig.json`.

When removing the line, the testfile would be marked as full of errors.

minestarks: Strange, I can remove that line and the file compiles fine and the tests pass locally with `npm run lint ineum`. How are you running the tests?

evolkmann: You are right. I checked it again. Must have been an error from mid-development where the error just temporarily occurred. I remove the type reference.

PR #38285, opened 2019-09-11
[react-plotly.js] Use frame type from plotly.js so callbacks work

gricey432:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: See desc below

Changed `Frame` type to be the one already present in Plotly, makes the types more inter-operable.
Added frames to the the `Figure` type, see https://github.com/plotly/react-plotly.js/blob/54bf0b25316079510d28fd3de2ccfcc080842146/src/factory.js#L175

### DIFF
--- a/types/react-plotly.js/index.d.ts
+++ b/types/react-plotly.js/index.d.ts
@@ -3,7 +3,7 @@
 // Definitions by: Jon Freedman <https://github.com/jonfreedman>
 //                 Mitchell Grice <https://github.com/gricey432>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 3.4
+// TypeScript Version: 2.8
 
 import * as Plotly from 'plotly.js';
 import * as React from 'react';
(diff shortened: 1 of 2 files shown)


types/react-plotly.js/index.d.ts
```diff
@@ -1,27 +1,23 @@
 // Type definitions for react-plotly.js 2.2
 // Project: https://github.com/plotly/react-plotly.js#readme
 // Definitions by: Jon Freedman <https://github.com/jonfreedman>
+//                 Mitchell Grice <https://github.com/gricey432>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.4
```
uniqueiniquity: I think you can leave this as 2.8 - this should be the minimum version of TypeScript that is compatible with these type definitions.

uniqueiniquity (changes requested):

types/react-plotly.js/index.d.ts
gricey432: Ah yep, my misunderstanding there

uniqueiniquity (approved):

PR #38288, opened 2019-09-11
add setAgent method annotations for dialogflow v2

jrthib:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://cloud.google.com/dialogflow/docs/reference/rpc/google.cloud.dialogflow.v2#setagentrequest>
<https://github.com/googleapis/nodejs-dialogflow/blob/86839335d2c2de5be68c8e7822082a59b94a7d85/src/v2/agents_client.js#L357>

### DIFF
--- a/types/dialogflow/index.d.ts
+++ b/types/dialogflow/index.d.ts
@@ -960,17 +960,17 @@ export type MatchMode =
 
 // API version for the agent.
 export type ApiVersion =
-  | 'API_VERSION_UNSPECIFIED'
-  | 'API_VERSION_V1'
-  | 'API_VERSION_V2'
-  | 'API_VERSION_V2_BETA_1';
+  | "API_VERSION_UNSPECIFIED"
+  | "API_VERSION_V1"
+  | "API_VERSION_V2"
+  | "API_VERSION_V2_BETA_1";
 
 // Represents the agent tier.
 export type Tier =
-  | 'TIER_UNSPECIFIED'
-  | 'TIER_STANDARD'
-  | 'TIER_ENTERPRISE'
-  | 'TIER_ENTERPRISE_PLUS';
+  | "TIER_UNSPECIFIED"
+  | "TIER_STANDARD"
+  | "TIER_ENTERPRISE"
+  | "TIER_ENTERPRISE_PLUS";
 
 export interface Credentials {
   client_email: string;
--- a/types/dialogflow/index.d.ts
+++ b/types/dialogflow/index.d.ts
@@ -1,8 +1,9 @@
-// Type definitions for dialogflow 0.9
+// Type definitions for dialogflow 0.11
 // Project: https://github.com/googleapis/nodejs-dialogflow
 // Definitions by: Daniel Dyla <https://github.com/dyladan>
 //                 Tom Carrio <https://github.com/tcarrio>
 //                 Khalil Choudhry <https://github.com/khalilchoudhry>
+//                 Joseph Thibeault <https://github.com/jrthib>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.4
 
@@ -17,6 +18,11 @@ export namespace v2 {
     getProjectId(): Promise<string>;
     getProjectId(callback?: (error: Error, id: string) => string): void;
 
+    setAgent(
+      request: SetAgentRequest,
+      options?: gax.CallOptions,
+      cb?: (err: Error, agent: Agent) => void
+    ): Promise<[Agent]>;
     getAgent(
       request: GetAgentRequest,
       options?: gax.CallOptions,
@@ -448,6 +454,11 @@ export namespace longrunning {
   }
 }
 
+export interface SetAgentRequest {
+  agent: Agent;
+  updateMask?: any;
+}
+
 export interface GetAgentRequest {
   parent: string;
 }
@@ -859,6 +870,8 @@ export interface Agent {
   enableLogging?: boolean;
   matchMode?: MatchMode;
   classificationThreshold?: number;
+  apiVersion?: ApiVersion;
+  tier?: Tier;
 }
 
 export interface Context {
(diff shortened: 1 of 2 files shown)


tcarrio:
I'm taking a look over this now. I see the 0.11 landed a few days ago.

types/dialogflow/index.d.ts
```diff
@@ -1,8 +1,9 @@
-// Type definitions for dialogflow 0.9
+// Type definitions for dialogflow 0.11
 // Project: https://github.com/googleapis/nodejs-dialogflow
 // Definitions by: Daniel Dyla <https://github.com/dyladan>
 //                 Tom Carrio <https://github.com/tcarrio>
 //                 Khalil Choudhry <https://github.com/khalilchoudhry>
+//                 Joseph Thibeault <https://github.com/jrthib>
```
tcarrio: Only include yourself if you are dedicating yourself to maintaining this package. This will include code reviews on all future pull requests.

types/dialogflow/index.d.ts
```diff
@@ -17,6 +18,11 @@ export namespace v2 {
     getProjectId(): Promise<string>;
     getProjectId(callback?: (error: Error, id: string) => string): void;
 
+    setAgent(
```
tcarrio: It is possible to pass the callback in place of options. See: https://github.com/googleapis/nodejs-dialogflow/blob/master/src/v2/agents_client.js#L358

types/dialogflow/index.d.ts
```diff
@@ -892,6 +905,36 @@ export type MatchMode =
   | "MATCH_MODE_HYBRID"
   | "MATCH_MODE_ML_ONLY";
 
+// API version for the agent.
+export enum ApiVersion {
```
tcarrio: Enums from TypeScript won't be compatible with the enums returned in the API. Union types of strings can be used instead where enums are used in the Dialogflow API. These provide type-safety, intellisense, and other advantages of TypeScript.

types/dialogflow/index.d.ts
```diff
@@ -892,6 +905,36 @@ export type MatchMode =
...
+
+  // V2 API.
+  API_VERSION_V2,
+
+  // V2beta1 API.
+  API_VERSION_V2_BETA_1,
+}
+
+// Represents the agent tier.
+export enum Tier {
```
tcarrio: Same comment as the `ApiVersion`

types/dialogflow/index.d.ts
tcarrio: For an example of how gRPC enums are handled with types take a look at the Dialogflow docs for AutoExpansionMode[0] and the corresponding Dialogflow types[1].

0: https://cloud.google.com/dialogflow/docs/reference/rpc/google.cloud.dialogflow.v2#google.cloud.dialogflow.v2.EntityType.AutoExpansionMode
1: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e5b9aa9887930b1879bffe09a1ee50a90f2a2e8a/types/dialogflow/index.d.ts#L960

types/dialogflow/index.d.ts
tcarrio: Would be easiest to simply provide some method overloading for `setAgent` rather than union types on the parameters.

tcarrio (changes requested):
Please review the following comments on your updates. Most everything looks good to go.

types/dialogflow/index.d.ts
dyladan: To expand on what tom said, the typings are only meant to reflect what the js already exports. The js does not export an object called ApiVersion with properties API_VERSION_UNSPECIFIED and so on. In methods that expect this "enum", they take strings as arguments and these are simply the possible string values.

eg `export type ApiVersion = "API_VERSION_UNSPECIFIED" | "API_VERSION_V1" ... `

jrthib: Makes sense, thanks for clarifying

types/dialogflow/index.d.ts
jrthib: Updated

types/dialogflow/index.d.ts
jrthib: Makes sense. I will PR to remove myself when I'm no longer able.

types/dialogflow/index.d.ts
jrthib: I'll submit another PR to fix the rest of the methods that use this pattern as well. `getAgent`, `searchAgents`, `deleteAgent` should all use overloading.

types/dialogflow/index.d.ts
```diff
@@ -892,6 +958,20 @@ export type MatchMode =
   | "MATCH_MODE_HYBRID"
   | "MATCH_MODE_ML_ONLY";
 
+// API version for the agent.
+export type ApiVersion =
+  | 'API_VERSION_UNSPECIFIED'
```
tcarrio: Please keep quotation marks consistent with the code - this project follows the TypeScript coding guidelines, using double quotes for strings.

https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#strings

tcarrio (changes requested):
Just an update to be consistent with the rest of the code then you're set.

tcarrio:
Quotation marks aren't enforced through tslint in the case in makes more sense to use single quotes (strings that could heavily containing double-quotes such as HTML). That's the last thing that needs to be done and otherwise these changes look good @jrthib

types/dialogflow/tslint.json
```diff
@@ -1 +1,6 @@
-{ "extends": "dtslint/dt.json" }
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "quotemark": [true, "double", "avoid-escape"]
```
jrthib: Added double quote enforcement via tslint with the `avoid-escape` argument so that it's ok to use a single quote in a string that would cause heavy escaping, such as the HTML case.

tcarrio:
Looks good! Let's get this moving along. Thank you for your contribution 🎉

tcarrio (approved):

types/dialogflow/tslint.json
DanielRosenwasser: Please don't add exceptions to the lint rules (especially since it doesn't seem to be used here). Instead, consider a PR to dtslint.

jrthib: Ok no prob. Just pushed a commit for removal.

orta:
Nice work folks 👍

PR #38300, opened 2019-09-11
Fix Lodash's _.memoize return type in TS3.1

itsh01:
Fix incorrect typing of `_.memoize()`:
Since memoize wraps the original function and do not return it,
the return type cannot be `T` because it won't have the properties defined on the passed function,
e.g:
```typescript
type AnyFunction = (...args: any[]) => any
type CreateFunction = () => AnyFunction & {otherValue: number}
const f = createFunction()
const memoized = _.memoize(f) 
// memoized.otherValue is undefined even though it's type is number
```

### DIFF
--- a/types/lodash/ts3.1/lodash-tests.ts
+++ b/types/lodash/ts3.1/lodash-tests.ts
@@ -3526,14 +3526,8 @@ fp.now(); // $ExpectType number
     memoizedFn.cache = new WeakMap();
     memoizedFn.cache = new Map();
 
-    {
-        type MemoizedResultFn = (a1: string, a2: number) => boolean;
-        let beforeMemoize: MemoizedResultFn & {otherKey: number};
-        let result: MemoizedResultFn;
-
-        beforeMemoize = _.assign(memoizeFn, {otherKey: 1});
-        result = _.memoize(beforeMemoize);
-    }
+    const beforeMemoize = _.assign((a1: string, a2: number): boolean => true, {otherKey: 1}); // $ExpectType ((a1: string, a2: number) => boolean) & { otherKey: number; }
+    const afterMemoize = _.memoize(beforeMemoize); // $ExpectType ((a1: string, a2: number) => boolean) & MemoizedFunction
 }
 
 // _.overArgs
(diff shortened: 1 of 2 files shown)


aj-r (approved):
I worry that this will degrade compiler performance for a very small benefit. But the performance checks seem to be ok.

uniqueiniquity (changes requested):
Please add a test that demonstrates your change.

itsh01:
@uniqueiniquity Added the tests

uniqueiniquity:
@aj-r Now that a test has been added, the performance impact shows up. Is this something you're comfortable with?

types/lodash/ts3.1/lodash-tests.ts
```diff
@@ -3525,6 +3525,15 @@ fp.now(); // $ExpectType number
...
     memoizedFn.cache = new WeakMap();
     memoizedFn.cache = new Map();
+
+    {
+        type MemoizedResultFn = (a1: string, a2: number) => boolean;
+        let beforeMemoize: MemoizedResultFn & {otherKey: number};
+        let result: MemoizedResultFn;
+
+        beforeMemoize = _.assign(memoizeFn, {otherKey: 1});
+        result = _.memoize(beforeMemoize);
```
aj-r: Use $ExpectType to get stricter type checking here. I think this test will currently pass even without your change.

E.g.

`_.memoize(...); // $ExpectType (a1: string, ...`

Note that ExpectType is kind of dumb (it just does a string comparison rather than actually checking if the 2 types are equivalent) so you can't use your MemoizedResultFn type. Unless they improved it recently.

aj-r (changes requested):
I'm not sure why this would cause a performance difference that significant. Try these suggestions and see if they make a difference. If not, maybe Parameters<> has bad performance?

itsh01:
@aj-r It looks better now, WDYT?

types/lodash/ts3.1/lodash-tests.ts
```diff
@@ -3525,6 +3525,9 @@ fp.now(); // $ExpectType number
     const memoizedFn = _.memoize(memoizeFn);
     memoizedFn.cache = new WeakMap();
     memoizedFn.cache = new Map();
+
+    const beforeMemoize = _.assign((a1: string, a2: number): boolean => true, {otherKey: 1}); // $ExpectType ((a1: string, a2: number) => boolean) & { otherKey: number; }
```
aj-r: Remove the `$ExpectType` on this line - you should not test `_.assign()` here.

aj-r (changes requested):

aj-r:
@itsh01 Agreed, performance looks good now.

PR #38304, opened 2019-09-11
Lodash: `negate`: require function to be predicate

OliverJAsh:
Please fill in this template.

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: https://lodash.com/docs/4.17.15#negate

### DIFF
--- a/types/lodash/ts3.1/common/function.d.ts
+++ b/types/lodash/ts3.1/common/function.d.ts
@@ -521,7 +521,7 @@ declare module "../index" {
          * @param predicate The predicate to negate.
          * @return Returns the new function.
          */
-        negate<T extends any[]>(predicate: (...args: T) => any): (...args: T) => boolean;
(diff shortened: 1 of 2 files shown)


aj-r (changes requested):
I worry that this will break things for some people. While the predicate should usually return a boolean, some people might just return a truthy/false value, e.g.

```
const isEmpty = negate((array) => array.length);
```

I'm not sure if anyone does this but I'm not sure we want to drop support for it.

Maybe require the predicate to return `NotVoid` instead?

aj-r:
Sorry for the late response. I think you make a good point, and this would be consistent with some of our other predicates (some, every, filter?). If you recreate this I'll approve it. Make sure you update all of the variations (chains, fp).

OliverJAsh:
@aj-r https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38973

PR #38314, opened 2019-09-11
[d3-scale] Make categorical scales support null and undefined

kristw:
Please fill in this template.

`scaleOrdinal`, `scaleBand` and `scalePoint` all support `null` and `undefined`. The current definition uses `{ toString(): string }` which excludes `null` and `undefined`.

### DIFF
--- a/types/d3-scale/index.d.ts
+++ b/types/d3-scale/index.d.ts
@@ -16,154 +16,154 @@ import { CountableTimeInterval, TimeInterval } from 'd3-time';
 // -------------------------------------------------------------------------------
 
 /**
- * An Interpolator factory returns an interpolator function.
- *
- * The first generic corresponds to the data type of the interpolation boundaries.
- * The second generic corresponds to the data type of the return type of the interpolator.
- */
+  * An Interpolator factory returns an interpolator function.
+  *
+  * The first generic corresponds to the data type of the interpolation boundaries.
+  * The second generic corresponds to the data type of the return type of the interpolator.
+  */
 export interface InterpolatorFactory<T, U> {
-    /**
-     * Construct a new interpolator function, based on the provided interpolation boundaries.
-     *
-     * @param a Start boundary of the interpolation interval.
-     * @param b End boundary of the interpolation interval.
-     */
-    (a: T, b: T): ((t: number) => U);
+  /**
+    * Construct a new interpolator function, based on the provided interpolation boundaries.
+    *
+    * @param a Start boundary of the interpolation interval.
+    * @param b End boundary of the interpolation interval.
+    */
+  (a: T, b: T): (t: number) => U;
 }
 
 /**
- * A helper interface for a continuous scale defined over a numeric domain.
- */
+  * A helper interface for a continuous scale defined over a numeric domain.
+  */
 export interface ScaleContinuousNumeric<Range, Output> {
-    /**
-     * Given a value from the domain, returns the corresponding value from the range, subject to interpolation, if any.
-     *
-     * If the given value is outside the domain, and clamping is not enabled, the mapping may be extrapolated such that the returned value is outside the range.
-     *
-     * Note: The interpolation function applied by the scale may change the output type from the range type as part of the interpolation.
-     *
-     * @param value A numeric value from the domain.
-     */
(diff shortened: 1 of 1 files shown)


kristw:
Could I get a review please?

denisname:
Please, provide examples and link to documentation or source code.

kristw:
```js
import { scaleOrdinal, scalePoint, scaleBand } from 'd3-scale';

const scale1 = scaleOrdinal().range(['red', 'green', 'blue]);
scale1(undefined); // red
scale1(null); // green
```

More live demo:
https://observablehq.com/@kristw/untitled

Documentation:
https://github.com/d3/d3-scale/blob/master/README.md#_ordinal
https://github.com/d3/d3-scale/blob/master/README.md#_point
https://github.com/d3/d3-scale/blob/master/README.md#_band

Source code:
https://github.com/d3/d3-scale/blob/master/src/ordinal.js

gustavderdrache:
I don't think these scales actually support `null` or `undefined` - what you're seeing is the implicit domain construction behavior of ordinal scales. The relevant code is [here](https://github.com/d3/d3-scale/blob/master/src/ordinal.js#L13-L16). When a key is not found in the map, it's simply added if implicit value support exists. The ordinal scale [docs on `.domain()`](https://github.com/d3/d3-scale/blob/master/README.md#ordinal_domain) say this:

> Setting the domain on an ordinal scale is optional if the unknown value is implicit (the default).

You can see this if you run `color.domain()` in your example _before_ calling `color(null)`:

```js
> let s = d3.scaleOrdinal().range(['red', 'green', 'blue'])
> s.domain()
Array []
> s(null)
"red"
> s.domain()
Array [ null ]
```

By this same logic, we could argue that scales "support" function arguments:

```js
> s(() => {})
"green"
> s.domain()
Array [ null, () ]
```

But in reality, it's just the fact that stringification is, generally, well-defined in JavaScript:

```js
> s("() => {}")
"green"
```

We can verify that it calls `toString` by deliberately crashing:

```js
> s({ toString() { throw new Error("toString called"); } })
Error: toString called
```

kristw:
By limiting the domain type of `scaleOrdinal` to `{ toString: () => string }` making it seems like I cannot call `scaleOrdinal()(undefined)`, which is not true. 

This is causing problem in consumer application that may pass `undefined` to color scale (`scaleOrdinal`). Either I add `@ts-ignore` which is kinda ugly workaround or have to catch and try to convert `undefined` to string before calling.

denisname:
@gustavderdrache is right.

> This is causing problem in consumer application that may pass undefined to color scale (scaleOrdinal).

"Just because a function produces a result doesn't mean that result is a meaningful one." (https://github.com/microsoft/TypeScript/issues/4002#issuecomment-124201510)

Your consumer application should have test for `undefined` when writing his code. He did not. Fortunately, typescript with these type definitions spot the error. You can now fix your *bug*.

kristw:
> By this same logic, we could argue that scales "support" function arguments:

Umm, I am not asking the type definition to support functions, only `null` and `undefined` though. 

> "Just because a function produces a result doesn't mean that result is a meaningful one." (microsoft/TypeScript#4002 (comment))

Base on the statement above, I would argue that the result from `scaleOrdinal` for `null` or `undefined` are meaningful. 

`scaleOrdinal` is commonly used for categorical color coding in visualizations. 

```
const data = [{ name: 'Jim', language: 'js' }, { name: 'Jane', language: 'ts' }, { name: 'Jill', language: undefined }];
const color = d3.scaleOrdinal(d3.schemeCategory10);
```

RyanCavanaugh:
The docs say (emph. mine):

> ordinal.domain sets or reads the scale’s domain. (Caveat: **the domain must be composed of numbers or strings, not objects.**)

The examples in the docs never use `null` or `undefined` either.

RyanCavanaugh:
Logged https://github.com/d3/d3-scale/issues/193 for the author to weigh in. 🤷‍♂️

RyanCavanaugh:
From the library author:

> In a type-safe context, I would not consider the scale function or the scale.domain method as accepting null or undefined; only string and array of string respectively.

kristw:
Thank you all reviewers for the consideration and thoughtful discussion.

PR #38315, opened 2019-09-11
[serverless] fix hooks and cloudformation.resources definitions

timtrinidad:
This fixes a couple of inconsistencies with the type declarations:
* `Plugin.hooks` is a dictionary of functions that return a promise, not a dictionary of promises
  * e.g. https://github.com/serverless/serverless/blob/v1.52.0/lib/plugins/plugin/plugin.js#L16-L21
* `serverless.service.provider.compiledCloudFormationTemplate.Resources` is an object rather than an array
  * e.g. https://github.com/serverless/serverless/blob/v1.52.0/lib/plugins/aws/package/compile/events/alb/lib/targetGroups.js#L14-L15
* `awsprovider.naming` definitions
  * e.g. https://github.com/serverless/serverless/blob/v1.52.0/lib/plugins/aws/provider/awsProvider.js#L125

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes
- N/A - If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- N/A -  If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.


types/serverless/classes/Plugin.d.ts
```diff
@@ -1,11 +1,12 @@
 import Serverless = require("../index");
 
-declare abstract class Plugin {
```
timtrinidad: I can't figure out how best to handle this.

With it as an abstract class, If I create a class with 
```ts
class MyPlugin extends Plugin {}
```
I get the correct typescript errors if my class doesn't implement the right constructor or the right `hooks` property. However, when I use that class, I get the following error since `classes/Plugin` [doesn't exist in serverless/serverless](https://github.com/serverless/serverless/tree/v1.52.0/lib/classes):

```
Cannot find module 'serverless/classes/Plugin'
```

With it as an interface, [we can't define the constructor](https://www.typescriptlang.org/docs/handbook/interfaces.html#difference-between-the-static-and-instance-sides-of-classes). Open to other suggestions.

timtrinidad: I also tried this, but TS doesn't complain if your constructor doesn't match this so it doesn't really help:

```ts
declare interface Plugin {
  hooks: {
    [event: string]: () => Promise<any>;
  };
}

declare const Plugin: {
  prototype: Plugin;
  new (serverless: Serverless, options: Serverless.Options): Plugin;
};

export = Plugin;

```

hassankhan: Hi @timtrinidad, sorry for the late reply. Honestly, I'm not quite sure how to approach this ... perhaps a maintainer might have a better suggestion?

uniqueiniquity: Hi @timtrinidad, what was the issue you were seeing with the second example you gave above? I'm not sure I see why that doesn't work.

uniqueiniquity (changes requested):

types/serverless/classes/Plugin.d.ts
timtrinidad: Are you referring to [this](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38315#discussion_r324222081)?

With that, I get the following error if `hooks` is not properly defined in my plugin, which is good:
```
Error:(8, 14) TS2420: Class 'ServerlessPluginExample' incorrectly implements interface 'Plugin'. Property 'hooks' is missing in type 'ServerlessPluginExample' but required in type 'Plugin'.
```

However, typescript won't complain if I define a constructor like this:
```
constructor(numTest: number)
```
even though it is not compatible with the constructor defined in the type declarations. So I'm not sure defining the constructor in the type declarations at all provides any value.

uniqueiniquity: Ah, I think I understand now.
I played with this locally for a bit, and here's where I got.
Seems like the core issue is that you want to provide types for a `Plugin`, but that's not a real class that exists (and hence the `super` call doesn't work as you point out).

```ts
// Plugin.d.ts
import Serverless = require('../index');
export interface Hooks {
    [event: string]: (...rest: any[]) => any;
}

export interface Commands {
    [command: string]: {
        usage?: string;
        lifecycleEvents?: string[];
        commands?: { [command: string]: {} };
        options?: {
            [option: string]: {
                usage?: string;
                required?: boolean;
                shortcut?: string;
            };
        };
    };
}

export interface Plugin {
    hooks: Hooks;
    commands?: Commands;
}

export interface PluginStatic {
    new (serverless: Serverless, options: Serverless.Options): Plugin;
}
```

```ts
// PluginManager.d.ts
import Serverless = require("../index");
import { Plugin, PluginStatic } from "./Plugin";

declare class PluginManager {
    constructor(serverless: Serverless)

    setCliOptions(options: Serverless.Options): void;
    setCliCommands(commands: {}): void;

    addPlugin(plugin: PluginStatic): void;
    loadAllPlugins(servicePlugins: {}): void;
    loadPlugins(plugins: {}): void;
    loadCorePlugins(): void;
    loadServicePlugins(servicePlugins: {}): void;
    loadCommand(pluginName: string, details: {}, key: string): {};
    loadCommands(pluginInstance: Plugin): void;
    spawn(commandsArray: string | string[], options?: any): Promise<void>;

    cliOptions: {};
    cliCommands: {};
    serverless: Serverless;
    plugins: Plugin[];
    commands: {};
    hooks: {};
    deprecatedEvents: {};
}

export = PluginManager;
```
```ts
// serverless-tests.ts
import Serverless from 'serverless';
import { Plugin, Hooks } from 'serverless/classes/Plugin';
import PluginManager from 'serverless/classes/PluginManager';

const options: Serverless.Options = {
    noDeploy: false,
    stage: null,
    region: '',
};

const serverless = new Serverless();

class CustomPlugin implements Plugin {
    commands = {
        command: {
            usage: 'description',
            lifecycleEvents: ['start'],
            options: {
                option: {
                    usage: `description`,
                    required: true,
                    shortcut: 'o',
                },
            },
        },
    };

    customProp = {};

    hooks: Hooks;

    constructor(serverless: Serverless, options: number) {
        this.hooks = {
            'command:start': () => {},
        };
    }
}

const manager = new PluginManager(serverless);
manager.addPlugin(CustomPlugin); //$ExpectError
```
As you point out, this won't give you an error when you try to declare a class with an incorrect constructor.
However, it won't let you pass that to `addPlugin`, so you do still get some checking benefit from this.

uniqueiniquity (changes requested):

types/serverless/classes/Plugin.d.ts
```diff
@@ -1,11 +1,12 @@
...
+interface Plugin {
+    // Typical constructor signature:
+    // constructor(serverless: Serverless, options: Serverless.Options)
+
     hooks: {
-        [event: string]: Promise<any>;
+        [event: string]: () => Promise<any>;
```
a-pavlenko: If hook is synchronous I don't see a reason to wrap it in Promise:
```javascript
[event: string]: () => <any>;
```

hassankhan: Can hooks be synchronous? I thought Serverless required hooks to return Promises - that said my knowledge of Serverless is quite old now so I may be wrong...

a-pavlenko: I didn't find any info in the documentation about Serverless plugins` hooks and whether they should mandatory return promise or not. This [blog post](https://serverless.com/blog/writing-serverless-plugins-2/) provides an insight that hook wrapped in promise will be awaited before proceeding to the next hook:
> As downloading is likely an asynchronous event, your function needs to return a promise. This way, the Serverless Framework will know to wait for this step to finish before starting the following one.

Since Serverless [uses](https://github.com/serverless/serverless/blob/master/lib/classes/PluginManager.js#L505) bluebird's [Promise.reduce](http://bluebirdjs.com/docs/api/promise.reduce.html) to invoke hooks, hence synchronous hooks will work just fine.

So if Serveless does not enforce usage of promises, from my perspective this library shouldn't do this either.

hassankhan: Fair enough, I vaguely remember failing code review (once upon a time) for a Serverless-internal plugin because I didn't return a Promise.

types/serverless/classes/Plugin.d.ts
timtrinidad: That's true. Most plugin devs won't be invoking the `PluginManager` themselves, but no harm in having the extra typing. I'll go ahead and add t.

types/serverless/classes/Plugin.d.ts
timtrinidad: I can't find anything either. [This blog post](https://serverless.com/blog/writing-serverless-plugins/) shows non-async hooks. I'll update the PR.

sashkopavlenko (approved):

uniqueiniquity (approved):

timtrinidad:
Awesome - thanks everyone!

hassankhan:
Thank you @timtrinidad for keeping the types up-to-date 👍

PR #38321, opened 2019-09-12
Update index.d.ts

yaotzin1:
Missing properties in FlatDataItem


dolezel (changes requested):
I would just change `extends TreeNode, TreePath` to `extends NodeData`. No need to add additional fields.

PR #38326, opened 2019-09-12
[react-router] Fix #38271 (`withRouter` fails with `ComponentType` starting in 3.6.2)

### LINKED ISSUE #38271: react-router: `withRouter` fails with `ComponentType` starting in 3.6.2
This worked in 3.5.2, but fails in 3.6.2.

`strict` mode is disabled.

Is this a bug in the typings or TypeScript?

I added failing tests for this here: https://github.com/DefinitelyTyped/DefinitelyTyped/commit/f036856dfc9a7fd140f378402c102206738ec0a4.

``` ts

// Copied from https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5db94ca7fd3570dc23588b404d7bb669a7886a8a/types/react-router/index.d.ts#L146
declare function withRouter<P, C extends React.ComponentType<P>>(
    component: C & React.ComponentType<P>,
): unknown;

declare const Component: React.ComponentType<{}>;
/*
Argument of type 'ComponentType<{}>' is not assignable to parameter of type 'ComponentClass<{}, any> | (ComponentClass<{}, any> & FunctionComponent<{}>)'.
  Type 'FunctionComponent<{}>' is not assignable to type 'ComponentClass<{}, any> | (ComponentClass<{}, any> & FunctionComponent<{}>)'.
    Type 'FunctionComponent<{}>' is not assignable to type 'ComponentClass<{}, any> & FunctionComponent<{}>'.
      Type 'FunctionComponent<{}>' is not assignable to type 'ComponentClass<{}, any>'.
        Type 'FunctionComponent<{}>' provides no match for the signature 'new (props: {}, context?: any): Component<{}, any, any>'.
*/
withRouter(Component);
```

---

  - Authors: @sergey-buturlakin @mrk21 @vasek17 @ngbrown @awendland @KostyaEsmukov @johnnyreilly @LKay @DovydasNavickas @huy-nguyen @grmiade @DaIgeb @egorshulga @rraina @pret-a-porter @t49tran @8enSmith @wezleytsai @eps1lon @HipsterBrown

karol-majewski:
Fixes #38271.

Please fill in this template.

No substantial changes are made here. This PR fixes a type error uncovered by TypeScript 3.6.2 by using 1 type parameter instead of two. The second one is derived with `React.ComponentProps`.

> you should strive to have fewer type parameters in your function signatures

https://twitter.com/SeaRyanC/status/1118632999266861056

### DIFF
--- a/types/react-router/index.d.ts
+++ b/types/react-router/index.d.ts
@@ -143,8 +143,12 @@ export interface WithRouterStatics<C extends React.ComponentType<any>> {
 // they are decorating. Due to this, if you are using @withRouter decorator in your code,
 // you will see a bunch of errors from TypeScript. The current workaround is to use withRouter() as a function call
 // on a separate line instead of as a decorator.
-export function withRouter<P extends RouteComponentProps<any>, C extends React.ComponentType<P>>(
-  component: C & React.ComponentType<P>,
-): React.ComponentClass<Omit<P, keyof RouteComponentProps<any>> & WithRouterProps<C>> & WithRouterStatics<C>;
+export function withRouter<C extends React.ComponentType<RouteComponentProps>>(
+  component: C,
+): React.ComponentClass<
+  Omit<React.ComponentProps<C>, keyof RouteComponentProps> & WithRouterProps<C>,
+  never
+> &
+  WithRouterStatics<C>;
 
 export const __RouterContext: React.Context<RouteComponentProps>;
--- a/types/react-router/test/WithRouter.tsx
+++ b/types/react-router/test/WithRouter.tsx
@@ -9,6 +9,12 @@ const ComponentFunction = (props: TOwnProps) => (
     <h2>Welcome {props.username}</h2>
 );
 
+const FunctionComponent: React.FunctionComponent<TOwnProps> = props => (
+  <h2>Welcome {props.username}</h2>
+);
+
+declare const Component: React.ComponentType<TOwnProps>;
+
 class ComponentClass extends React.Component<TOwnProps> {
     render() {
         return <h2>Welcome {this.props.username}</h2>;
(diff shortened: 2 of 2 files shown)


eps1lon (commented):
This looks fine. Though I don't know what this has to do with the tweet. That one was talking about type parameters not union return types.

Curious why this would result in slower benchmarks. Could you try returning a plain function instead?

karol-majewski:
> No a plain function e.g. (props: Props) => JSX.Element

> Since we never gave the type guarantee that the component is a class we can safely choose the simplest type.

As my colleague @OliverJAsh pointed out, such a change would break the callers relying on features available only on class-based components (like React refs).

This piece of code indicates that not only `withRouter` supports both types of components, but also makes a change to the return type based on the type of the provided component:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/bf75bcd274e6770fa6e34064d8ce3b2d7c57c929/types/react-router/index.d.ts#L134-L136

eps1lon:
> As my colleague @OliverJAsh pointed out, such a change would break the callers relying on features available only on class-based components (like React refs).

Just realized we never returned a ComponentType. Should be fine.

Regardless of this change I would recommend to not rely on the specific component type declared in typescript.

`ref` being available is simply a matter of the props interface not the component type. You can define function components that accept a ref.

eps1lon (approved):

orta:
OK, looks good to merge to me then yep 👍

ulrichb:
Hmmm ... this makes type inference fail when using 

```ts
withRouter(React.memo<RouteComponentProps & {
    readonly someProp: string;
}>( ({ match, someProp }) =>  /* ... */ ) ))
```

:(

ksylvest:
Looks like this definition now fails:

```typescript
const Sample: React.FC<RouteComponentProps & { name: string }> = ({ name }) => (
  <>{name}</>
);

const SampleWithRouter = withRouter(Sample);
```

eps1lon:
@karol-majewski 

Does
```tsx
withRouter<RouteComponentProps & {
    readonly someProp: string;
}>(React.memo( ({ match, someProp }) =>  /* ... */ ) ))
```
work?

Regardless please open for each case a new issue.

eps1lon:
I encourage everyone to open new issues. If we fix your particular issue you have no way of being notified if you only comment in a closed thread. The typescript playground now allows configuring the ts version +  config and is able to download type definitions. Including a link in a new issue helps maintainers a lot. Thank you :pray: 

It seems like this only throws with `strictFunctionTypes` and I have a solution in mind. We'll see if it works out.

artola:
From my point of view the offending new types need to be rolled back, and the newest should have a simple test to verify them.

PR #38331, opened 2019-09-12
pdfmake Add addVirtualFileSystem to base

jkoudys:
Please fill in this template.

- [ -] Add or edit tests to reflect the change. (Run with `npm test`.)
- [ X] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [ X] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).

- [ X] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.

### DIFF
--- a/types/pdfmake/index.d.ts
+++ b/types/pdfmake/index.d.ts
@@ -24,6 +24,8 @@ declare module "pdfmake/build/pdfmake" {
         vfs?: any
     ): TCreatedPdf;
 
+    function addVirtualFileSystem(vfs: any): void;
+
     enum PageSize {
         A0_x_4 = "4A0",
         A0_x_2 = "2A0",


andipaetzold (changes requested):
This doesn't seem to be part of the public api of `pdfmake`, right?
Or could you please provide a link to the docs that mention the behavior of `addVirtualFileSystem`.

andipaetzold (changes requested):
This doesn't seem to be part of the public api of `pdfmake`, right?
Or could you please provide a link to the docs that mention the behavior of `addVirtualFileSystem`.

jkoudys:
Their docs seem to assign directly into `pdfmake.vfs` instead of doing the `addVirtualFileSystem` (which calls `fd.bindFS()`). That's used by their system for extending the `pdfmake` object with some fonts when you load `build/vfs`. I _was_ using this because doing `import()`s introduced a race condition in how it extended pdfmake, but if they're not going to ask people to use that fn (for whatever benefits it may have) then I guess it doesn't need to be a type.

PR #38333, opened 2019-09-12
react-geosuggest: fix invalid types for Suggest.location

FrancisEgan:
Changing definition for Suggest.location. lat and lng should be number, not string. Proper type for these properties (number) can be seen here:
https://github.com/ubilabs/react-geosuggest/blob/master/src/types/location.d.ts


brmenchl (approved):
Nice, thanks!

PR #38341, opened 2019-09-12
Create typings for honeybadger module

rskrz:
Please fill in this template.

- [ X Add or edit tests to reflect the change. (Run with `npm test`.)

If adding a new definition:

### DIFF
--- a/types/honeybadger/index.d.ts
+++ b/types/honeybadger/index.d.ts
@@ -2,46 +2,46 @@
 // Project: https://github.com/honeybadger-io/honeybadger-node
 // Definitions by: Ryan Skrzypek <https://github.com/rskrz>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 3.5
+// TypeScript Version: 2.9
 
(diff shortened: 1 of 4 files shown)


types/honeybadger/index.d.ts
```diff
@@ -0,0 +1,47 @@
...
+  resetContext: (context?: object) => void;
+  factory: (options?: ConfigureOptions) => HoneyBadgerInstance;
+  errorHandler: ErrorRequestHandler;
+  requestHandler: RequestHandler;
+  lambdaHandler: (handler: LambdaHandler) => LambdaHandler;
+  onUncaughtException: (func: (error: Error) => void) => void;
+}
+declare const honeybadger: HoneyBadgerInstance;
+
+export = honeybadger;
```
uniqueiniquity: It looks this is a module that exposes a set of properties but is not itself callable or constructable. In that case, I recommend using [this template](http://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html) for clarity.

uniqueiniquity (changes requested):

PR #38157, opened 2019-09-05
[ramda] split tests to separate modules - functions beginning with 'a'

devrelm:
This is an alternative to #37710, which itself is an alternative to #37636. This PR aims to be easier to review than #37710 by migrating only a small subset of the tests.

I recommend looking at each commit individually. The first commit is the README, which would be good to read to get a feel for why I broke up the tests the way I did. The remaining 15 commits are each fairly small and every line removed from `ramda-tests.ts` should show up in a separate file with the (possible) exception of some duplicate tests that I found along the way.

Unlike #37710, which migrated _all_ tests to separate files, this just migrates tests for ramda functions that begin with the letter 'a'. It also adds a `README.md` to guide contributors towards writing meaningful tests for the `@types/ramda` package.

Unlike #37636, no changes were made to `types/ramda/index.d.ts`, and no additional assertions were added to the tests, nor were existing tests modified beyond being moved to a separate file. This PR just splits out each functions' tests into separate files.

The added readme lays one ground rule for the tests: that each test file should only contain tests where the outermost function called is the one which shares the name of the file. For instance, the following:

```ts
// $ExpectType (x: number) => number
R.pipe(
  R.add(2),
  R.add(3),
  R.add(4),
);
```

tests `R.pipe`, _not_ `R.add`. So it belongs in `test/pipe-tests.ts` rather than `test/add-tests.ts`.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).
- [ ] ~Provide a URL to documentation or source code which provides context for the suggested changes: <<url here>>~
- [ ] ~If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.~
- [ ] ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.~

PR #37710, opened 2019-08-17
[ramda] split tests into one module per ramda function

devrelm:
This is an alternative to #37636 which just splits the `ramda-tests.ts` file and adds a `README.md` to guide contributors towards writing tests for the `@types/ramda` package.

Unlike #37636, no changes were made to `types/ramda/index.d.ts`, and no assertions were added to the tests. The only thing this PR does is split out the tests for each function into a separate file for each function.

The added readme lays one ground rule for the tests: that each test file should only contain tests where the outermost function called is the one which shares the name of the file. For instance, the following:

```ts
R.pipe(
  R.add(2),
  R.add(3),
  R.add(4),
);
```

tests `R.pipe`, _not_ `R.add`. So it belongs in `test/pipe-tests.ts` rather than `test/add-tests.ts`.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).
- [ ] ~Provide a URL to documentation or source code which provides context for the suggested changes: <<url here>>~
- [ ] ~If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.~
- [ ] ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.~

PR #38347, opened 2019-09-12
[ramda] split tests to separate modules: 'filter' - 'head'

devrelm:
#37710 attempted to split the tests for `@types/ramda` into separate files per function, but the PR was too large to be approved on its own. This PR is part of a new piecemeal approach to migrating the tests.

I recommend stepping through the PR commit-by-commit to make it easier to check that no individual tests were dropped or changed substantively.

This PR is the next of many to move the functions over in a more piecemeal fashion. It continues from the following PRs:

* #38157 `[ramda] split tests to separate modules - functions beginning with 'a'`
* #38251 `[ramda] split tests to separate modules - functions beginning with 'b' - 'c'`
* #38274 `[ramda] split tests to separate modules - functions beginning with 'd' - 'e'`

I'm splitting out the tests into separate files per function to make it easier to track what is and is not being tested. Please read the descriptions for #37710 and #38157 for more details.

---

- [ ] ~~Provide a URL to documentation or source code which provides context for the suggested changes: <<url here>>~~
- [ ] ~~If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.~~


types/ramda/test/forEach-tests.ts
```diff
@@ -0,0 +1,26 @@
...
+    console.log(x + 5);
+  }
+
+  R.forEach(printXPlusFive, [1, 2, 3]);
+})();
+
+// (() => {
+//   function printXPlusFive(x, i) { console.log(i + 5); }
+//   R.forEach.idx(printXPlusFive, [{name: 1}, {name: 2}, {name: 3}]);
+// })();
```
drewwyatt: is there reason to keep this around?

drewwyatt (approved):
Added one minor comment, but this looks great! I don't see anything getting dropped. 🎉

types/ramda/test/forEach-tests.ts
devrelm: I think so. This actually tests being able to pass in an array of objects rather than just an array of numbers, so it's slightly different.

I didn't want to uncomment it right now, but I could see how it could be useful to remind us of other test-cases once we get around to writing more later on.

I'm less concerned about this commented-out test, and more concerned with the tests below it (lines 16-26), which appear to be a copy of the first uncommented test-case (lines 3-9) except with the addition of the curried test-case. I had removed other duplicate tests for other functions, so I'm not 100% sure why I left that one behind.

Still, no harm no foul. We can clean them up further later-on.

devrelm:
@minestarks Got another one for you. Thanks for your help!

devrelm:
Bump

PR #38348, opened 2019-09-12
mongodb: support two argument collection call with options

dobesv:
You can call the collection() method with a name and options, but
no callback, if you want to.

https://mongodb.github.io/node-mongodb-native/3.3/api/Db.html#collection

### DIFF
--- a/types/mongodb/index.d.ts
+++ b/types/mongodb/index.d.ts
@@ -621,9 +621,8 @@ export class Db extends EventEmitter {
     /** http://mongodb.github.io/node-mongodb-native/3.1/api/Db.html#admin */
     admin(): Admin;
     /** http://mongodb.github.io/node-mongodb-native/3.1/api/Db.html#collection */
-    collection<TSchema = Default>(name: string): Collection<TSchema>;
-    collection<TSchema = Default>(name: string, callback: MongoCallback<Collection<TSchema>>): Collection<TSchema>;
-    collection<TSchema = Default>(name: string, options: DbCollectionOptions, callback: MongoCallback<Collection<TSchema>>): Collection<TSchema>;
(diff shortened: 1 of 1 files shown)


types/mongodb/index.d.ts
```diff
@@ -622,7 +622,7 @@ export class Db extends EventEmitter {
     admin(): Admin;
     /** http://mongodb.github.io/node-mongodb-native/3.1/api/Db.html#collection */
     collection<TSchema = Default>(name: string): Collection<TSchema>;
-    collection<TSchema = Default>(name: string, callback: MongoCallback<Collection<TSchema>>): Collection<TSchema>;
+    collection<TSchema = Default>(name: string, callbackOrOptions: MongoCallback<Collection<TSchema>> | DbCollectionOptions): Collection<TSchema>;
```
CaselIT: It seems ok, but I think it would be better to use the same pattern used in the other cases, like for the `createCollection` method:

```ts
createCollection<TSchema = Default>(name: string, callback: MongoCallback<Collection<TSchema>>): void;
createCollection<TSchema = Default>(name: string, options?: CollectionCreateOptions): Promise<Collection<TSchema>>;
createCollection<TSchema = Default>(name: string, options: CollectionCreateOptions, callback: MongoCallback<Collection<TSchema>>): void;
```

CaselIT (changes requested):

types/mongodb/index.d.ts
dobesv: I realized that the second and third option in your list can be combined, since this method always returns the same type.  I am assuming you preferred the other way because the argument names were better.

CaselIT (approved):

PR #38351, opened 2019-09-13
[relay-runtime] augment NormalizationSelection

artola:
Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/relay-tools/relay-compiler-language-typescript/issues/131#issuecomment-531004495

If removing a declaration:


types/relay-runtime/index.d.ts
```diff
@@ -728,7 +734,7 @@ interface ReaderCondition {
 }
 
 interface ReaderClientExtension {
-    readonly kind: 'ClientExtension';
+    readonly kind: string; // 'ClientExtension';
```
alloy: Why was this change necessary?

alloy (changes requested):
I think it would make sense to add some test coverage for this, could you add some that shows how you actually use it?

types/relay-runtime/index.d.ts
artola: > Why was this change necessary?

Just to match all the other members in the union `ReaderSelection`. I would prefer to have a more strict type than just `string`, but this means that others should be refactored, and possible some other members added to the union that now match `string`.

artola:
> I think it would make sense to add some test coverage for this, could you add some that shows how you actually use it?

Added test for ConcreteRequest.

types/relay-runtime/relay-runtime-tests.tsx
```diff
@@ -75,3 +76,78 @@ commitLocalUpdate(environment, store => {
...
+        },
+        params: {
+            operationKind: 'query',
+            name: 'FooQuery',
+            id: null,
+            text: 'query FooQuery {\n  __typename\n}\n',
+            metadata: {},
+        },
+    };
+})();
```
alloy: 👌

types/relay-runtime/index.d.ts
alloy: Gotcha :+1:

alloy:
The lint failures about the test can just be silenced, but the one in the declaration file should indeed be fixed.

artola:
> The lint failures about the test can just be silenced, but the one in the declaration file should indeed be fixed.

I agree ... working.

alloy (approved):

alloy:
Perfect, thanks! 🙏

PR #38355, opened 2019-09-13
Add a selector object for all selectors

adimit:
Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nightwatchjs.org/guide/working-with-page-objects/#element-properties

If removing a declaration:

### DIFF
--- a/types/nightwatch/index.d.ts
+++ b/types/nightwatch/index.d.ts
@@ -1,9 +1,10 @@
-// Type definitions for nightwatch 1.1
+// Type definitions for nightwatch 1.2
 // Project: http://nightwatchjs.org
 // Definitions by: Rahul Kavalapara <https://github.com/rkavalap>
 //                 Connor Schlesiger <https://github.com/schlesiger>
 //                 Clayton Astrom <https://github.com/ClaytonAstrom>
 //                 Lukas Beranek <https://github.com/lloiser>
+//                 Aleksandar Dimitrov <https://github.com/adimit>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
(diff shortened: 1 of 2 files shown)


types/nightwatch/index.d.ts
```diff
@@ -1785,7 +1826,13 @@ export interface ElementCommands {
      * @see waitForElementPresent
      * @since v0.4.0
      */
-    waitForElementNotPresent(selector: string, time?: number, abortOnFailure?: boolean, callback?: (this: NightwatchAPI, result: NightwatchCallbackResult<void>) => void, message?: string): this;
+    waitForElementNotPresent(
+      selector: NightwatchSelector,
```
lloiser: I think it goes even further: if you pass a selector object, all arguments are ignored but the very last which is the callback.
So actually we have 2 different function signatures, at least for the `waitForElement*` functions
```js
waitForElementNotVisible(
    selector: string,
    time?: number,
    abortOnFailure?: boolean,
    callback?: (this: NightwatchAPI, result: NightwatchCallbackResult<void>) => void,
    message?: string
): this;
```
and
```js
waitForElementNotVisible(
    selector: NightwatchSelectorObject,
    callback?: (this: NightwatchAPI, result: NightwatchCallbackResult<void>) => void,
): this;
```
right?

Also I'm not 100% sure if the `message?: string` in the first one is even correct because code always uses the last argument as callback:
https://github.com/nightwatchjs/nightwatch/blob/7b2ef1e1f57f7093554cd477910f63da9eeb5a45/lib/element/command.js#L322-L331

lloiser (changes requested):

types/nightwatch/index.d.ts
adimit: Yes, I agree, it seems the functions are "overloaded" — though the way arguments are parsed is very programmatic.

I think the last argument should have the signature `callback | string`, if you look at the function [right above setCallback](https://github.com/nightwatchjs/nightwatch/blob/7b2ef1e1f57f7093554cd477910f63da9eeb5a45/lib/element/command.js#L311). I think that's a bug in nightwatch, to be honest. I'll try to reproduce it and file an issue if it is indeed impossible to specify both a callback and a message.

adimit:
> @adimit I haven't seen anything from you in a while and this PR currently has problems that prevent it from being merged. The PR will be closed tomorrow if there aren't new commits to fix the issues.

PR #38361, opened 2019-09-13
[@hapi/joi] update to v16

RecuencoJones:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://hapi.dev/family/joi/?v=16.0.0, https://github.com/hapijs/joi/issues/2037

Currently work in progress

### DIFF
--- a/types/hapi__joi/index.d.ts
+++ b/types/hapi__joi/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for @hapi/joi 15.0
+// Type definitions for @hapi/joi 16.0
 // Project: https://github.com/hapijs/joi
 // Definitions by: Bart van der Schoor <https://github.com/Bartvds>
 //                 Laurence Dougal Myers <https://github.com/laurence-myers>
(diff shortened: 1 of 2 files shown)


SimonSchick:
I realise this in is WIP (as by your comment).
I tried to work on this as well but, the change log is HUGE!
Do you want to collaborate on this?
I know @marsup is trying to move typings into joi (he pinged me directly) so might be good to have his opinion as well.

SimonSchick:
That said, feel free to copy paste from https://github.com/SimonSchick/DefinitelyTyped/tree/feat/joi-v16
I tried to sort of apply the changes per joi issue but this wasn't very successful since many of them overlap..

Marsup:
Well, if you're nearing the end, it might be wiser to import the result into joi and fix things afterwards. What do you think ?

SimonSchick:
I don't think either of us are near the end 😛

Marsup:
This PR seems pretty advanced to me.

SimonSchick:
Are you talking about *this* PR or my PR?
I can see that *this* one is missing a few things mine includes and vise versa, but even if we merged those I believe there is a good chunk of things missing, such as custom rules.

Marsup:
Oh... I didn't see yours :sweat_smile: Having the definitions included in joi will make it easier for us to manage, even if it's 80/20, it's better than nothing.

RecuencoJones:
Hey @SimonSchick, @Marsup, sorry for joining late to the conversation, didn't expect such a quick response!

> I tried to work on this as well but, the change log is HUGE!
> Do you want to collaborate on this?

I was expecting someone to be working on this already, so quite happy to collaborate of course! 😄

The changelog is indeed huge, I tried to read through it but it was at times confusing, so I went ahead and directly compared with the documentation site of v16.0.0 as it is easier to understand.

So, how can I help? Should we maybe split changes per schema types?

> Having the definitions included in joi will make it easier for us to manage, even if it's 80/20, it's better than nothing.

Is the plan to ship the typings along with the package, but manually build as these ones or are you considering switching to typescript to get them autogenerated?

SimonSchick:
@RecuencoJones I believe the best path forward is you continue your work and maybe incorporate mine, once that's done I will do a first review pass, there's a LOT to go over :)

RecuencoJones:
I'm facing a bit of an issue here, there are some new top level exported functions which have keywords as names (`function` and `in`). Any ideas on how to fix this?

I have been looking for solutions like moving all top level exports to an interface, but then I am not allowed to export that interface without a default export...

types/hapi__joi/index.d.ts
```diff
@@ -19,1378 +19,1883 @@
...
+
+        /**
+         * Requires the array values to be unique.
+         * Remember that if you provide a custom comparator function,
+         * different types can be passed as parameter depending on the rules you set on items.
+         * Be aware that a deep equality is performed on elements of the array having a type of object,
+         * a performance penalty is to be expected for this kind of operation.
+         */
+        unique(comparator?: string, options?: ArrayUniqueOptions): this;
+        unique<T = any>(comparator?: (a: T, b: T) => boolean, options?: ArrayUniqueOptions): this;
```
SimonSchick: While we are here, we might want to consider making both of these `unknown` instead.

types/hapi__joi/index.d.ts
```diff
@@ -19,1378 +19,1883 @@
...
+        path: Array<string | number>;
+        type: string;
+        context?: Context;
+    }
+
+    type ValidationErrorFunction = (errors: ValidationErrorItem[]) => string | ValidationErrorItem | Error;
+
+    interface ValidationResult<T> {
+        error: ValidationError;
+        warnings: any;
```
SimonSchick: Can we narrow this?

types/hapi__joi/index.d.ts
```diff
@@ -19,1378 +19,1883 @@
...
+         */
+        then?: SchemaLike;
+        /**
+         * the alternative schema type if the condition is false. Required if then is missing.
+         */
+        otherwise?: SchemaLike;
+    }
+
+    interface Cache {
+        set(key: any, value: any): void;
```
SimonSchick: Can keys really be `any`?

types/hapi__joi/index.d.ts
```diff
@@ -19,1378 +19,1883 @@
...
+         * @default true
+         */
+        allowUnicode?: boolean;
+
+        /**
+         * Options for TLD (top level domain) validation. By default, the TLD must be a valid name listed on the [IANA registry](http://data.iana.org/TLD/tlds-alpha-by-domain.txt)
+         *
+         * @default { allow: true }
+         */
+        tlds?: {
```
SimonSchick: We can create a common type with the `tlds` property above.

types/hapi__joi/index.d.ts
```diff
@@ -19,1378 +19,1883 @@
...
+         */
+        noDefaults?: boolean;
+        /**
+         * when true, inputs are shallow cloned to include non-enumerables properties. Defaults to false.
+         */
+        nonEnumerables?: boolean;
+        /**
+         * sets the default presence requirements. Supported modes: 'optional', 'required', and 'forbidden'. Defaults to 'optional'.
+         */
+        presence?: 'optional' | 'required' | 'forbidden';
```
SimonSchick: This looks duplicate can we introduce a type for this?

SimonSchick (commented):
Starting to look pretty good, some comments so far!

types/hapi__joi/index.d.ts
RecuencoJones: From [CacheInterface](https://hapi.dev/family/joi/?v=16.0.1#cache-interface) docs:

> Note that key and value can be anything including objects, array, etc.

I will add documentation to it.

types/hapi__joi/index.d.ts
RecuencoJones: The only downside to it is we will need to specify the type when passing the comparator function or else: 
image

What do you think? Enforcing to add the type to the comparator signature seems fine by me

RecuencoJones: Seems like upgrading to TS 3.0 would require other packages to upgrade as well:

https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/585918998#L237
```
Error: hapi__hapi depends on hapi__joi but has a lower required TypeScript version.
```

types/hapi__joi/index.d.ts
```diff
@@ -152,1245 +226,1757 @@ export interface DomainOptions {
...
+
+        /**
+         * Allows the number to be outside of JavaScript's safety range (Number.MIN_SAFE_INTEGER & Number.MAX_SAFE_INTEGER).
+         */
+        unsafe(enabled?: any): this;
+    }
+
+    interface StringSchema extends AnySchema {
+        /**
+         * Requires the string value to only cont
```
jaulz: `pre` seems to be obsolete according to https://github.com/hapijs/joi/blob/master/lib/schemas.js#L76-L113

jaulz: `describe` seems to be obsolete according to `pre` seems to be obsolete according to https://github.com/hapijs/joi/blob/master/lib/schemas.js#L76-L113

jaulz: `coerce` seems to be `coerce?(schema: ExtensionBoundSchema, value: any, helpers:CustomHelpers): any;` now according to https://hapi.dev/family/joi/?v=16.0.1#extensions

jaulz: We should add `flags?: Record<string, Flag>;` here with:
```
    interface Flag {
        setter?: string;
        default?: any
    }
```

(see https://github.com/hapijs/joi/blob/master/lib/schemas.js#L85-L88)

jaulz: We should add flags here for extensions:
```
        /**
         * Flags of current schema
         */
        _flags:Record<string, any>;
```

RecuencoJones: Also missing: 
- `$_terms`
- `$_validate`

jaulz: This should be `args?: (RuleArgs | string)[];` (see https://github.com/hapijs/joi/blob/master/lib/schemas.js#L54-L67)

jaulz: Sorry, having a second look and using the library I think the example is wrong and it's now:
```
    interface ErrorReport extends Error {
        code: string;
        flags: Record<string, Flag>;
        path: string;
        prefs: ValidationOptions;
        messages: LanguageMessages;
        state: State;
        value: any;
    }

    type CoerceFunction = (value: any, helpers:CustomHelpers) => {
        errors?:ErrorReport[]
        value?: any
    } | undefined

        coerce?: {
            from: string | string[],
            method: CoerceFunction
        } | CoerceFunction;
```

See:
https://github.com/hapijs/joi/blob/aa374d7034d8fccc166ad4a80601ce1ddd7ed8db/lib/extend.js#L61-L71  https://github.com/hapijs/joi/blob/bb1b40559f7c9b6abd62c70bcd5c49a5ca4951fe/lib/validator.js#L218-L232
https://github.com/hapijs/joi/blob/14efd782935a56e6b5631fd5fcf37958b570254c/lib/errors.js#L18-L24

types/hapi__joi/index.d.ts
```diff
@@ -152,1245 +226,1907 @@ export interface DomainOptions {
...
+
+        /**
+         * Validates a value using the schema and options.
+         */
+        validate<T>(value: T, options?: ValidationOptions): ValidationResult<T>;
+
+        /**
+         * Validates a value using the schema and options.
+         */
+        validateAsync<T>(value: T, options?: AsyncValidationOptions): Promise<ValidationResult<T>>;
```
jaulz: This should be:
```
validateAsync<T>(value: T, options?: AsyncValidationOptions): Promise<T>;
```

SimonSchick: Technically `value` should be `any`, not `T` since we may validate any object type and the schema returns a different one.

jaulz: Right, though you could overwrite it via validateAsync<any> but actually my point was that the result is not a ValidationResult but the value itself.

RecuencoJones: Good point!

SimonSchick (approved):

jaulz:
Can we get this merged? 😊

types/hapi__joi/index.d.ts
```diff
@@ -152,1245 +226,1922 @@ export interface DomainOptions {
...
+        clone: boolean;
+    }
+
+    interface CustomHelpers<V = any> {
+        schema: ExtensionBoundSchema;
+        state: State;
+        prefs: ValidationOptions;
+        original: V;
+        warn: (code: string, local?: Context) => void;
+        error: (code: string, local?: Context) => ValidationError;
```
jaulz: I think `error` and `message` return `ErrorReport` and not `ValidationReport`.

types/hapi__joi/index.d.ts
```diff
@@ -152,1245 +226,1922 @@ export interface DomainOptions {
...
+    interface ValidationErrorItem {
+        message: string;
+        path: Array<string | number>;
+        type: string;
+        context?: Context;
+    }
+
+    type ValidationErrorFunction = (errors: ValidationErrorItem[]) => string | ValidationErrorItem | Error;
+
+    interface ValidationResult {
```
jaulz: I think we should make this generic: 
``
    interface ValidationResult<T = any> {
        error: ValidationError;
        errors: ValidationError;
        warning: ValidationError;
        value: T;
    }
``

jaulz (changes requested):

types/hapi__joi/index.d.ts
jaulz: Mhm, not sure what's best. 
@SimonSchick what do you think?

SimonSchick: `no-unnecessary-generics` is _usually_ right, as you are effectively substituting casts with generics, which in the end reduces type safety, unless your intent it to make schemas generic and default to `unknown` then this should NOT be generic.

jaulz: But how would it look like then? Am I supposed to cast every validation result like this?
```ts
const result = schema.validate() as {
  error: ValidationError;
  errors: ValidationError;
  warning: ValidationError;
  value: {
    validatedString: string          
  };
}
```

Or is there a simpler way?

jaulz (approved):

armanio123:
Seems like this is ready to be merged. I'll merge it tomorrow. Let me know if I should wait.

types/hapi__joi/index.d.ts
evanshortiss: After reading the comment by @SimonSchick, maybe `unknown` is an OK compromise since it's at least safer than `any`. Supporting a generic still makes sense to me though since it doesn't violate the lint rule since it was used in the params and return value AFAICT.

SimonSchick: That's the thing, the input value may NOT be the return value since joi performs cohersion/conversion.

evanshortiss: Exactly, that's why the generic is required you need to specify the return value type. So I *think* we agree here?

SimonSchick: No, you input `unknown` and you get `unknown`, the generic would do nothing but hide the type assertion, which `no unnecessary generic` attempts to prevent.

evanshortiss: Agreed you input `unknown`, but returning `unknown` is a choice being made to satisfy the lint rule AFAICT. 

We know what'll be returned is either:

* an object with the `value` key with value `T` (since Joi validated it)
* an object with an `error` key and associated value (if validation failed)

Changing the `any` to `unknown` is probably a good call since changing the lint rule is probably a no-go and it would force developers to apply the type cast?

PR #38371, opened 2019-09-13
Add channelgroup type definitions to pubnub types

nathanloyer:
I started to use the pubnub type definitions, but I needed to call the set of methods I have added. I've had a forked copy of this file in my codebase for a few months, so I wanted to merge it in here so I can remove the file from my code. I was referencing pubnub@4.25.2, which is the latest version. I don't know if there are other things still missing from the types here, but I only focused on adding the ones I needed.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/pubnub/javascript/blob/master/src/core/pubnub-common.js
- [ ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header. (I was referencing 4.25.2, but I'm only adding the parts I need and not checking everything.)

### DIFF
--- a/types/pubnub/index.d.ts
+++ b/types/pubnub/index.d.ts
@@ -33,25 +33,30 @@ declare class Pubnub {
 
   publish(
     params: Pubnub.PublishParameters,
-    callback: (status: Pubnub.PublishStatus, response: Pubnub.PublishResponse) => void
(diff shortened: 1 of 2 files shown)


uniqueiniquity:

PR #38376, opened 2019-09-14
Add types npm-registry-fetch, libnpmsearch, and pacote

joelspadin:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/pacote/index.d.ts
+++ b/types/pacote/index.d.ts
@@ -6,14 +6,16 @@
 
 /// <reference types="node" />
 
+import { Readable } from 'stream';
 import npmFetch = require('npm-registry-fetch');
 
-export import extract = require('./extract');
-export import manifest = require('./manifest');
-export import packument = require('./packument');
-export import prefetch = require('./prefetch');
-export import tarball = require('./tarball');
-import { Readable } from 'stream';
+import extract = require('./extract');
+import manifest = require('./manifest');
+import packument = require('./packument');
+import prefetch = require('./prefetch');
+import tarball = require('./tarball');
+
+export { extract, manifest, packument, prefetch, tarball };
 
 /**
  * This utility function can be used to force pacote to release its references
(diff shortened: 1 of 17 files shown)


types/pacote/index.d.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+export interface PackageVersion {
+    name: string;
+    version: string;
+    dependencies?: Record<string, string>;
+    optionalDependencies?: Record<string, string>;
+    devDependencies?: Record<string, string>;
+    peerDependencies?: Record<string, string>;
+    directories: {};
```
ChaosinaCan: [npm's documentation for this field](https://github.com/npm/registry/blob/master/docs/REGISTRY-API.md#version) literally says "???". 🤷‍♀️

In every request I've tried, it's always shown up as an empty object.

types/pacote/index.d.ts
```diff
@@ -0,0 +1,162 @@
...
+// Project: https://github.com/npm/pacote#readme
+// Definitions by: Joel Spadin <https://github.com/ChaosinaCan>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+/// <reference types="node" />
+
+import npmFetch = require('npm-registry-fetch');
+
+export import extract = require('./extract');
```
uniqueiniquity: For readability, I'd suggest doing:
```ts
import extract = require('./extract');
import manifest = require('./mainfest');
...

export { extract, manifest, ... };
```

uniqueiniquity (changes requested):

types/pacote/index.d.ts
ChaosinaCan: Good idea. Updated.

uniqueiniquity (approved):

ChaosinaCan:
Thanks!

PR #38388, opened 2019-09-15
Add typings for aurora.js (av)

cdauth:
Please fill in this template.

If adding a new definition:

If removing a declaration:


types/av/filter.d.ts
```diff
@@ -0,0 +1,16 @@
+/// <reference path="./baseTypes.d.ts" />
+
+declare namespace AV {
```
DanielRosenwasser: I don't think this is the right way to structure this. You're relying on global merging behavior which introduces a global variable named `AV`. Instead just use top-level imports and exports, and use `export *`. Preferably just keep these in the same file.

types/av/filter.d.ts
```diff
@@ -0,0 +1,16 @@
...
+	class Filter {
+		constructor();
+		constructor(context: { }, key: string);
+
+		process(buffer: TypedArray): void;
```
DanielRosenwasser: Can you switch from tabs for 4 spaces?

DanielRosenwasser (changes requested):

types/av/filter.d.ts
cdauth: I'm not sure I completely understand what you mean. I was experimenting with exports before (see [here](https://github.com/cdauth/DefinitelyTyped/commit/52648054c16f96dbffec87a5934f47016a925593)), but couldn't get it to work as I wanted. I was trying to export the AV object like [this](https://github.com/cdauth/DefinitelyTyped/commit/52648054c16f96dbffec87a5934f47016a925593#diff-67eabcd3abfea6a992cc13333d5acf69R20), but then I didn't understand how anyone from the outside could access the types (I was trying to do something like this: `const AV = require('AV'); const buffer: AV.Buffer;`, but it wouldn't let me access the types like that). Do you have a suggestion for that?

 Would you say that all the declarations should be in one file instead of split up as they are now?

cdauth:
@DanielRosenwasser I responded to your review comment, could you have a look at it?

PR #38389, opened 2019-09-15
Update index.d.ts

saeedtabrizi:
Update Orientjs New Api to support latest orientjs v3.0.6


orta:
This is a pretty drastic change (and I'd expect it to break some people's builds) - does is correlate to a major server bump in the dependency itself? Maybe you can update the major version with that also?

saeedtabrizi:
> This is a pretty drastic change (and I'd expect it to break some people's builds) - does is correlate to a major server bump in the dependency itself? Maybe you can update the major version with that also?
Thanks . it's ready to use with orientjs v 3.0.6 and have compatibility with older version .

orta:
OK, cool 👍

PR #38401, opened 2019-09-16
feat(node): v12.10

SimonSchick:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#12.10.0

@RyanCavanaugh (unsure if right person), I'm unsure about adding `Promise.allSettled` as it's one of those more popular functions (why is why I was asked to do this update) and should technically live inside the lib files.
I'm concerned causing incompatibilities due to name/signature conflicts of types.

Main changes:
- Add all changes from node v12.7 - v12.10 and some missing ones from previous versions.
- Move fs tests to own file (no changes, just additions).
- Consolidate some type unions for more clarify.
- Add missing v8 functions.
- Add missing bigint fs stat functions.


Flarna (approved):

BrunoScheufler (approved):

r3nya (approved):
Great job!

SimonSchick:
I'm thinking about de-coupling the `readable-stream` package from the node typings as they like to break on stream changes, thoughts?

Flarna:
`readable-stream` should not refer to latest node typings as they implement streams based on Node.Js 10. See https://www.npmjs.com/package/readable-stream

But I'm not sure about the implications if we change the dependencies there. You may get several versions of node typings in a project but not sure if typescript is able to handle this.

SimonSchick:
@Flarna so should I remove all references to node? 🤔

Flarna:
@SimonSchick No I meant it should refer to v10 typings of node instead latest. This should avoid that changes from latest node require adaptions in `readable-stream` which are not correct there as it should reflect the v10 version of streams to my understanding.

The only risk I see is here is that this may cause problems for project referring to `readable-stream` and latest node.

But removing all references to node is maybe also a good solution as one intention of this project is to have node like streams independent of node...

SimonSchick:
I tried to make them reference node v10 but this doesn't seem to work, via the `tsconfig`'s `path` property, any ideas?

Flarna:
maybe as dependency in package.json?

SimonSchick:
That will not work or pass the linter, worst case it will even install node typings twice in a user install, which is not going to end well.

Flarna:
Technically the correct approach would be to just copy needed stream definitions from v10 to readable-stream but not sure if this results in happy users.
The same is valid for adding v12 properties there which are actually not present there in real world.

In short: I don't know what's best to move forward 😢

PR #38404, opened 2019-09-16
prosemirror-view: upgrade api to 1.11.3

ifiokjr:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/ProseMirror/prosemirror-view/blob/master/CHANGELOG.md#1110-2019-09-16>>

### DIFF
--- a/types/prosemirror-view/index.d.ts
+++ b/types/prosemirror-view/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for prosemirror-view 1.9
+// Type definitions for prosemirror-view 1.11
 // Project: https://github.com/ProseMirror/prosemirror-view
 // Definitions by: Bradley Ayers <https://github.com/bradleyayers>
 //                 David Hahn <https://github.com/davidka>
@@ -234,6 +234,10 @@ export class EditorView<S extends Schema = any> {
    * should not directly interfere with its content.)
    */
   dom: Element;
+  /**
+   * Indicates whether the editor is currently [editable](#view.EditorProps.editable).
+   */
+   editable: boolean;
   /**
    * When editor content is being dragged, this object contains
    * information about the dragged slice and whether it is being
@@ -546,7 +550,7 @@ export interface EditorProps<S extends Schema = any> {
     [name: string]: (
       node: ProsemirrorNode<S>,
       view: EditorView<S>,
-      getPos: () => number,
+      getPos: (() => number) | boolean,
       decorations: Decoration[]
     ) => NodeView<S>;
   } | null;
@@ -683,11 +687,17 @@ export interface NodeView<S extends Schema = any> {
   /**
    * Called when a DOM
    * [mutation](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)
-   * happens within the view. Return false if the editor should
+   * or a selection change happens within the view. When the change is
+   * a selection change, the record will have a `type` property of
+   * `"selection"` (which doesn't occur for native mutation records).
+   * Return false if the editor should re-read the selection or
    * re-parse the range around the mutation, true if it can safely be
    * ignored.
    */
-  ignoreMutation?: ((p: MutationRecord) => boolean) | null;
+  ignoreMutation?: ((p: MutationRecord | {
+    type: 'selection';
+    target: Element;
+  }) => boolean) | null;
   /**
    * Called when the node view is removed from the editor or the whole
    * editor is destroyed.


types/prosemirror-view/index.d.ts
```diff
@@ -546,7 +550,7 @@ export interface EditorProps<S extends Schema = any> {
     [name: string]: (
       node: ProsemirrorNode<S>,
       view: EditorView<S>,
-      getPos: () => number,
+      getPos: () => number | boolean,
```
timjb: You need some parens here:

```typescript
      getPos: (() => number) | boolean,
```

(`() => number | boolean` is the type of functions that do not require any arguments and return a `number` or a `boolean`.)

types/prosemirror-view/index.d.ts
```diff
@@ -683,14 +687,22 @@ export interface NodeView<S extends Schema = any> {
...
    * Called when the node view is removed from the editor or the whole
    * editor is destroyed.
    */
   destroy?: (() => void) | null;
 }
+/**
+ * When the change is a selection change, the record will have a `type` property
+ * of `"selection"` (which doesn't occur for native mutation records).
+ */
+export type ProsemirrorMutationRecordType = 'attributes' | 'characterData' | 'childList' | 'selection';
```
timjb: FYI: The following is also possible:

```typescript
export type ProsemirrorMutationRecordType = MutationRecord['type'] | 'selection';
```

This makes a little bit clearer what's added.

timjb (changes requested):

types/prosemirror-view/index.d.ts
ifiokjr: Good spot 😄

types/prosemirror-view/index.d.ts
ifiokjr: True 👍

types/prosemirror-view/index.d.ts
```diff
@@ -683,14 +687,22 @@ export interface NodeView<S extends Schema = any> {
...
-   * happens within the view. Return false if the editor should
+   * or a selection change happens within the view. When the change is
+   * a selection change, the record will have a `type` property of
+   * `"selection"` (which doesn't occur for native mutation records).
+   * Return false if the editor should re-read the selection or
    * re-parse the range around the mutation, true if it can safely be
    * ignored.
    */
-  ignoreMutation?: ((p: MutationRecord) => boolean) | null;
+  ignoreMutation?: ((p: MutationRecord & { readonly type: ProsemirrorMutationRecordType }) => boolean) | null;
```
timjb: Unfortunately, this doesn't work as intended: the type `MutationRecord & { readonly type: ProsemirrorMutationRecordType }` is equivalent to `MutationRecord`, because all the values of this type must be objects whose `type` property has the intersection type `MutationRecord['type'] & ProsemirrorMutationRecordType`, which is equivalent to `MutationRecord['type']`. (Since in general, the type `A & (A | B)` is equivalent to `A` for any types `A` and `B`.)

A better type to use here would be `MutationRecord | { type: 'selection', target: Element }`. This is more accurate since the object constructed [in this line in ProseMirror's source](https://github.com/ProseMirror/prosemirror-view/blob/cbda274c7942000698a83b1aa8cc5db1a4c43266/src/domobserver.js#L118) has only `type` and `target` set, not other properties of `MutationRecord`.

timjb (changes requested):
Sorry, just found another bug!

types/prosemirror-view/index.d.ts
ifiokjr: Okay that makes sense. I've updated it.

timjb (approved):
Looking good now! :+1:

ifiokjr:
@timjb Any ideas how I can speed up the merge process. Typically it would have been merged by now.

ifiokjr:
@alloy I'm wondering if this PR has somehow slipped under the radar. I'd love to bump it to the front of the queue for getting merged.

alloy:
Unsure, perhaps because it still has the `revision needed` label? /cc @andy-ms

Anyways, looks like the maintainer agrees, so merging now 👍

ifiokjr:
> perhaps because it still has the `revision needed`

@alloy I suspected that might be the reason. Thanks for the quick turnaround 😄

hubgit:
This PR changed the definition of `getPos` for nodeViews  from `getPos: () => number` to `getPos: (() => number) | boolean`, which doesn't seem to be appropriate (at least not with the current version of prosemirror-view) - was that deliberate?

BrianHung:
@hubgit 

From https://prosemirror.net/docs/ref/#view.EditorProps.nodeViews
> For marks, the third argument is a boolean that indicates whether the mark's content is inline.

I think both @types/prosemirror-view and the (pseudo) types on prosemirror.net docs have to be updated.

hubgit:
@BrianHung I've since realised that I was wrong, the third argument can be a boolean if the nodeView is for a mark. It's a bit frustrating that it's not possible to distinguish between "node" and "mark" arguments (here and elsewhere), but it is at least correct.

PR #38419, opened 2019-09-16
ramda: better mergeDeepRight / mergeDeepLeft types

ekilah:
`MergeDeep` doesn't type correctly for array members of to-be-merged objects, since arrays `extend object` too.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [ramda REPL](https://ramdajs.com/repl/?v=0.26.1#?R.mergeDeepRight%28%7B%20name%3A%20%27fred%27%2C%20age%3A%2010%2C%20contact%3A%20%7B%20email%3A%20%27moo%40example.com%27%20%7D%2C%20types%3A%20%5B2%2C%205%5D%7D%2C%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%7B%20age%3A%2040%2C%20contact%3A%20%7B%20email%3A%20%27baa%40example.com%27%20%7D%2C%20types%3A%20%5B1%2C%204%5D%7D%29%3B%0A%2F%2F%3D%3E%20%7B%20name%3A%20%27fred%27%2C%20age%3A%2040%2C%20contact%3A%20%7B%20email%3A%20%27baa%40example.com%27%20%7D%7D)

### DIFF
--- a/types/ramda/index.d.ts
+++ b/types/ramda/index.d.ts
@@ -551,8 +551,14 @@ declare namespace R {
     type Omit<T, K extends string> = Pick<T, Exclude<keyof T, K>>;
 
     type CommonKeys<T1, T2> = keyof T1 & keyof T2;
+
+    type PropsThatAreArrays<T, K extends keyof T> = K extends keyof T ? T[K] extends ReadonlyArray<any> ? K : never : never;
+    type PropsThatAreObjectsButNotArrays<T, K extends keyof T> = K extends keyof T ? T[K] extends ReadonlyArray<any> ? never : T[K] extends object ? K : never : never;
     type PropsThatAreObjects<T, K extends keyof T> = K extends keyof T ? T[K] extends object ? K : never : never;
+
+    type CommonPropsThatAreArrays<T1, T2> = PropsThatAreArrays<T1, keyof T1> & PropsThatAreArrays<T2, keyof T2>;
     type CommonPropsThatAreObjects<T1, T2> = PropsThatAreObjects<T1, keyof T1> & PropsThatAreObjects<T2, keyof T2>;
+    type CommonPropsThatAreObjectsButNotArrays<T1, T2> = PropsThatAreObjectsButNotArrays<T1, keyof T1> & PropsThatAreObjectsButNotArrays<T2, keyof T2>;
 
     type Ord = number | string | boolean | Date;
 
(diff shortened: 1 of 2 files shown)


ekilah:
this could also use some changes for things like `Set`s and other things that Ramda won't merge deeply (just like Arrays), but what that really shows is it would be nice to just identify plain old `{}`s vs things like Arrays/Sets/Maps/etc and I'm not sure if that's possible? `extends Record<any, any>` might do it but maybe not?

Nemo108 (approved):

pirix-gh:
Hey @ekilah, would you mind closing this in favor of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38503

ekilah:
@pirix-gh #38503 looks a lot more complex than this and solves (or at least, promises to solve) a lot more issues with `mergeDeep...` than I tried to here. I've also experienced frustrations with the `| undefined` stuff you're working on there and in `ts-toolbelt`. I'm wiling to close this if / when #38503 merges 👍 

(i haven't tested your changes but it seems like you know the type system a whole lot better than me re: `ts-toolbelt`)

pirix-gh:
thanks @ekilah, we can do that. Do you have some examples where you had problems? I can look at it.
I wrote a type called `Kind` that "untangles" overlapping types (object, array, and function that can all narrow down to object), to resolve the problem you mentioned.

ekilah:
sure, getting a little off-topic, but for the sake of the discussion:

something like: 

```ts

type MyType = {
  foo: string
  bar: number
} 

const defaults: MyType = {foo: 'val', bar: 0}

const defaultsWithOverride: MyType = R.mergeRight<MyType, Partial<MyType>>(
  defaults,
  {foo: 'new value'} // override defaults
)
```

will, with the typings live on `@types` (before your PR #38503 at least), fail type checks because TS thinks the merged type has `foo: string | undefined`.

The only reason I would declare `Partial<MyType>` in the generic for `mergeRight` here is so that I get some guarantee from the type system that there are no typos in the keys of the override object `{foo: 'newValue'}`.

pirix-gh:
Yes, with https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38503, you won't need to worry about override problems anymore, if I understood well what you said. 
And thanks for the tip, it's a nice way to do a `defaultsDeep`.

mrcrane:
We've merged #38503 so this should be closed.

PR #38422, opened 2019-09-16
[koa-generic-session] Add a sameSite attribute to SessionOptions

JackMc:
Seems like koa-generic-session has started supporting SameSite, but the definitions do not contain it.

## Changing an existing definition
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.npmjs.com/package/cookie#samesite
- ~If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.~ This is not a result of a version change, the `cookie` package upon which this is based supports this option.
- ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.~ Not making a substantial change.

### DIFF
--- a/types/koa-generic-session/index.d.ts
+++ b/types/koa-generic-session/index.d.ts
@@ -37,6 +37,7 @@ declare namespace koaSession {
             maxAge?: number;
             secure?: boolean;
             httpOnly?: boolean;
+            sameSite?: string;
         };
         allowEmpty?: boolean;
         defer?: boolean;
--- a/types/koa-generic-session/index.d.ts
+++ b/types/koa-generic-session/index.d.ts
@@ -37,7 +37,7 @@ declare namespace koaSession {
             maxAge?: number;
             secure?: boolean;
             httpOnly?: boolean;
-            sameSite?: string;
+            sameSite?: boolean | "lax" | "none" | "strict";
         };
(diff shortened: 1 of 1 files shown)


JackMc:
cc @nsimmons @Ragg- as code owners

nsimmons (approved):

types/koa-generic-session/index.d.ts
```diff
@@ -37,6 +37,7 @@ declare namespace koaSession {
             maxAge?: number;
             secure?: boolean;
             httpOnly?: boolean;
+            sameSite?: string;
```
orta: Looks like it could be boolean or a string, maybe this is a better option?

```suggestion
            sameSite?: boolean | "lax" | "none" | "strict";
```

JackMc: TIL you can do that in TypeScript :) Thanks, I accepted your suggestion.

orta: Yeah! More on [that topic here](https://typescript-site-60.ortam.now.sh/play/index.html?#example/union-and-intersection-types)

PR #38426, opened 2019-09-16
[pino-http] Add err to ServerResponse

davidduffett:
[pino-http](https://github.com/pinojs/pino-http/blob/master/logger.js#L48) uses the `err` property on `ServerResponse` to pass errors to the error serializer. Without this definition, you cannot set `res.err = err;` to allow the error to be logged.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/pinojs/pino-http/blob/master/logger.js#L48

### DIFF
--- a/types/pino-http/index.d.ts
+++ b/types/pino-http/index.d.ts
@@ -39,4 +39,7 @@ declare module 'http' {
     interface IncomingMessage {
         log: Logger;
     }
+    interface ServerResponse {
+        err: Error;
+    }
 }
--- a/types/pino-http/index.d.ts
+++ b/types/pino-http/index.d.ts
@@ -40,6 +40,6 @@ declare module 'http' {
         log: Logger;
     }
     interface ServerResponse {
-        err: Error;
+        err?: Error;
     }
 }
--- a/types/pino-http/pino-http-tests.ts
+++ b/types/pino-http/pino-http-tests.ts
@@ -9,7 +9,7 @@ const httpLogger = pinoHttp();
 function handle(req: http.IncomingMessage, res: http.ServerResponse) {
   httpLogger(req, res);
   req.log.info('something else');
-  const err: Error = res.err;
+  const err: Error | undefined = res.err;
 }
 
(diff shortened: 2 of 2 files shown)


types/pino-http/index.d.ts
```diff
@@ -39,4 +39,7 @@ declare module 'http' {
     interface IncomingMessage {
         log: Logger;
     }
+    interface ServerResponse {
+        err: Error;
```
jdforsythe: Should this be optional?

davidduffett: Good point, yep - I've updated that. Thanks!

jdforsythe:
Lgtm

jdforsythe (approved):

types/pino-http/index.d.ts
```diff
@@ -39,4 +39,7 @@ declare module 'http' {
     interface IncomingMessage {
         log: Logger;
     }
+    interface ServerResponse {
```
DanielRosenwasser: Instead of module augmentations, why not have local types that extend the base types? e.g.

```
import { IncomingMessage as IncomingMessageBase, ServerResponse as ServerResponseBase } from 'http';

interface IncomingMessage extends IncomingMessageBase {
    log: Logger;
}

interface ServerResponse extends ServerResponseBase {
    err?: Error;
}
```

This way, you don't pollute the global scope.

davidduffett: Sure, I'm happy to do this in this PR if @jdforsythe is on board.  I'm a bit of a n00b at this so forgive the stupid question, but: would these interfaces sit under the `PinoHttp` namespace? I would think this would be a breaking change for users of this package. Or do they simply sit at the top level of this module?

davidduffett: In fact I'm not sure how this change wouldn't be breaking for consumers, given the point of the package is to add a `log` property for use within http handlers. Presumably anyone using this package would have to change their handlers to reference the new interface rather than the `http` one?

PR #38431, opened 2019-09-17
added vue the mask types

domschmidt:
added new types for vue the mask
https://github.com/vuejs-tips/vue-the-mask


uniqueiniquity (changes requested):
Looks like this got pulled into #38435 - I'm fine with you submitting them both in that PR.

PR #38433, opened 2019-09-17
rebass: add Emotion CSSObject

thebuilder:
Please fill in this template.

-  [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/rebassjs/rebass/issues/664

This adds the CSSObject from Emotion to the global css object definition on `BaseProps`. Rebass is using Emotion by default, so the user should expect to be able to use `css` prop from Emotion.

### DIFF
--- a/types/rebass/package.json
+++ b/types/rebass/package.json
@@ -1,6 +1,6 @@
 {
     "private": true,
     "dependencies": {
-        "@emotion/core": "^10.0.17"
+        "@emotion/core": "*"
     }
 }
--- a/types/rebass/package.json
+++ b/types/rebass/package.json
@@ -1,0 +1,6 @@
+{
+    "private": true,
+    "dependencies": {
+        "@emotion/core": "^10.0.17"
+    }
+}
--- a/types/rebass/index.d.ts
+++ b/types/rebass/index.d.ts
@@ -13,6 +13,7 @@ import { ResponsiveStyleValue, SystemStyleObject } from '@styled-system/css';
 import * as React from "react";
 import * as StyledComponents from "styled-components";
 import * as StyledSystem from "styled-system";
+import { CSSObject } from '@emotion/serialize'
 
 export {};
 
@@ -21,6 +22,7 @@ type Omit<T, K> = Pick<T, Exclude<keyof T, K>>;
 export interface BaseProps extends React.RefAttributes<any> {
     as?: React.ElementType;
     css?:
+        | CSSObject
         | StyledComponents.CSSObject
         | StyledComponents.FlattenSimpleInterpolation
         | string;
--- a/types/rebass/index.d.ts
+++ b/types/rebass/index.d.ts
@@ -13,7 +13,7 @@ import { ResponsiveStyleValue, SystemStyleObject } from '@styled-system/css';
 import * as React from "react";
 import * as StyledComponents from "styled-components";
 import * as StyledSystem from "styled-system";
-import { InterpolationWithTheme } from '@emotion/core'
+import { InterpolationWithTheme } from '@emotion/core';
 
 export {};
 
--- a/types/rebass/index.d.ts
+++ b/types/rebass/index.d.ts
@@ -13,7 +13,7 @@ import { ResponsiveStyleValue, SystemStyleObject } from '@styled-system/css';
 import * as React from "react";
 import * as StyledComponents from "styled-components";
 import * as StyledSystem from "styled-system";
-import { CSSObject } from '@emotion/serialize'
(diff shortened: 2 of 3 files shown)


types/rebass/package.json
```diff
@@ -0,0 +1,6 @@
+{
+    "private": true,
+    "dependencies": {
+        "@emotion/core": "*"
```
rhysd: Should we use `*` for version restriction? If supported version of emotion is clear, I think specifying it would be better.

thebuilder: Not sure what's the best practice for requiring these dependencies in DefinitelyTyped. Just saw it like this in another repo - but setting a specific version should also be fine.

rhysd: OK, I was just curious that there might be some version specified clearly in rebass document.

rhysd (approved):

benshell:
I'm glad to see this PR, as I was having trouble with these type definitions too. However, this doesn't entirely solve the problem. In the case of extending a Rebass component, it can be helpful to re-use (and probably extend) the component type. But then I'd be declaring my component props as allowing _any_ of these `css` attribute types, which isn't compatible with Emotion which only allows one type:

```
declare module 'react' {
  interface DOMAttributes<T> {
    css?: InterpolationWithTheme<any>
  }
}
```

So I get this error:

```
Type '{ variant?: string | (string | null)[] | { [key: string]: string; } | undefined; as?: "symbol" | "object" | "a" | "abbr" | "address" | "area" | "article" | "aside" | "audio" | "b" | "base" | "bdi" | ... 164 more ... | undefined; ... 404 more ...; onTransitionEndCapture?: ((event: TransitionEvent<...>) => void) | und...' is not assignable to type 'IntrinsicAttributes'.
  Types of property 'css' are incompatible.
    Type 'string | number | boolean | ComponentSelector | Keyframes | SerializedStyles | ArrayInterpolation<undefined> | ObjectInterpolation<undefined> | ((theme: any) => Interpolation<undefined>) | CSSObject | FlattenSimpleInterpolation | null | undefined' is not assignable to type 'InterpolationWithTheme<any>'.
      Type 'FlattenSimpleInterpolation' is not assignable to type 'InterpolationWithTheme<any>'.
        Type 'FlattenSimpleInterpolation' is not assignable to type 'ObjectInterpolation<undefined>'.
```

The only solution I can think of would be to use generics just about everywhere, so that can I set the expected css prop type when I need to, e.g.

```
  type CSSTypes =
    | InterpolationWithTheme<any>
    | StyledComponents.CSSObject
    | StyledComponents.FlattenSimpleInterpolation
    | string

  export interface BaseProps<U = CSSTypes> extends React.RefAttributes<any> {
    as?: React.ElementType
    css?: U
  }
```

```
  interface BoxKnownProps<U = CSSTypes>
    extends BaseProps<U>,
      StyledSystem.SpaceProps,
      StyledSystem.LayoutProps,
      StyledSystem.FontSizeProps,
      StyledSystem.ColorProps,
      StyledSystem.FlexProps,
      StyledSystem.OrderProps,
      StyledSystem.AlignSelfProps,
      SxProps {
    variant?: StyledSystem.ResponsiveValue<string>
  }
  export interface BoxProps<U = CSSTypes>
    extends BoxKnownProps<U>,
      Omit<React.HTMLProps<HTMLDivElement>, keyof BoxKnownProps> {}
  export const Box: React.FunctionComponent<BoxProps>
```

and then in my component:

```
/** @jsx jsx */
import * as React from 'react'
import { jsx } from '@emotion/core'
import { Box, BoxProps } from 'rebass'
import { InterpolationWithTheme } from '@emotion/core'

const MyBox: React.FC<BoxProps<InterpolationWithTheme<any>>> = props => {
  return <Box {...props}></Box>
}

export default MyBox
```

It's not pretty, so I'd love to hear that there's a better way. Thoughts?

thebuilder:
I'm struggling a bit with getting this working on DefTyped.

I'm thinking the best solution is to just go with `InterpolationWithTheme<any>` as the definition for `css` in `@types/rebass` - Then it might make sense to create a `@†ypes/rebass/styled-components` or similar version.

thebuilder:
When working with components that can handle BoxProps, i usually define them as:

```ts
type Props = BoxProps & {
 customProp: string
}
```

benshell:
> I'm struggling a bit with getting this working on DefTyped.
> 
> I'm thinking the best solution is to just go with `InterpolationWithTheme<any>` as the definition for `css` in `@types/rebass` - Then it might make sense to create a `@†ypes/rebass/styled-components` or similar version.

Agreed... that would be ideal.

PR #38435, opened 2019-09-17
Vue moment types

domschmidt:
added types for vue-moment

https://github.com/brockpetrie/vue-moment

### DIFF
--- a/types/vue-moment/index.d.ts
+++ b/types/vue-moment/index.d.ts
@@ -1,0 +1,32 @@
+// Type definitions for vue-moment 4.0
+// Project: https://github.com/brockpetrie/vue-moment
+// Definitions by: Dominik Schmidt <https://github.com/domschmidt>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.4
+
+import { Moment, MomentFormatSpecification, MomentInput } from 'moment';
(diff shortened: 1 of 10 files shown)


types/vue-the-mask/index.d.ts
```diff
@@ -0,0 +1,21 @@
...
+
+import { Component, DirectiveFunction, PluginObject } from 'vue';
+
+declare namespace VueTheMaskPlugin {
+    interface VueStatic {
+        (): void;
+    }
+}
+
+interface VueTheMaskPlugin extends PluginObject<void> {
```
uniqueiniquity: `void` doesn't really make sense for the type argument for `PluginObject` here because it's used to describe the type of an argument to a function. I'd recommend using `undefined` instead.

types/vue-moment/index.d.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+}
+
+declare module 'vue/types/vue' {
+    interface Vue {
+        $moment: VueMomentPlugin.VueStatic;
+    }
+}
+
+interface VueMoment extends PluginObject<void> {}
```
uniqueiniquity: Same here as my other comment.

uniqueiniquity (changes requested):

uniqueiniquity (approved):

PR #38442, opened 2019-09-17
[@types/prop-types] update `props` type for `Validator`

kaiyoma:
Custom validators that use the `Validator` type need to be able to
access values out of the `props` object, but currently this fails type
checking. The object of props is keyed by strings, so add this simple
type.

Please fill in this template.

### DIFF
--- a/types/prop-types/index.d.ts
+++ b/types/prop-types/index.d.ts
@@ -38,7 +38,7 @@ export type OptionalKeys<V> = Exclude<keyof V, RequiredKeys<V>>;
 export type InferPropsInner<V> = { [K in keyof V]-?: InferType<V[K]>; };
 
 export interface Validator<T> {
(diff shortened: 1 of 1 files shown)


ferdaber (approved):

DanielRosenwasser:
Should it be `{ [prop: string]: unknown }` instead, or { [prop: string]: unknown } & object`? That would mean you still have to probe the structure of these values.

eps1lon:
> Should it be `{ [prop: string]: unknown }` instead, or { [prop: string]: unknown } & object`? That would mean you still have to probe the structure of these values.

We are still on ^2.8 with the types and I'm not sure maintaining multiple type versions is sustainable.

kaiyoma:
In my team's codebase, we would have written this as `{ [prop: string]: unknown }`, but I used `any` in this commit for parity with the rest of the file (`unknown` never appears, but `any` appears 20+ times).

PR #38444, opened 2019-09-17
[react] Use PropsWithChildren for ExoticComponent

danielkcz:
Please fill in this template.

---

Just quick, dirty & small change. I am mostly just curious if it's intended for some reason which I am missing.

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -309,7 +309,7 @@ declare namespace React {
         /**
          * **NOTE**: Exotic components are not callable.
          */
-        (props: P): (ReactElement|null);
+        (props: PropsWithChildren<P>): (ReactElement|null);
         readonly $$typeof: symbol;
     }
 


eps1lon (changes requested):
If anything we want to get rid of implicit children. At some point we want to type them correctly<sup>1</sup>. Both these changes will enable strict types for e.g. render props APIs.

Edit: Bit more about implicit children: `children` is no special prop that is treated differently by react. Component authors should be explicit about `children` support since it is not always obvious where these are put. Basically in every component that returns a fragment the position of the children is ambiguous and maybe they aren't even handled.

<sup>1</sup> They are currently too wide for host components. They accept functions and objects that will throw at runtime.

FredyC:
@eps1lon Thanks, that makes total sense. I was mostly curious if it's an oversight or what.

I did type children explicitly in cases where it makes sense, especially `children?: never`. I even made an interface `NoChildren` to extend from. Perhaps couple of these could be introduced directly in typings.

eps1lon:
@FredyC I'm in the process of drafting a RFC to get rid of implicit children. Your `NoChildren` interface might come in handy. I'll try to remember pinging you when I propose the change.

FredyC:
@eps1lon These are aliases we have crafted and seems to work quite nicely :)

```ts
/**
 * Single React element (not a component).
 */
declare type ReactNode = null | React.ReactElement<any>

/**
 * Single or multiple React elements.
 */
declare type ReactNodes = ReactNode | ReactNode[]

/**
 * Extend props with specific type for children, eg. string.
 * Use Required<WithChildren<T>> if children are required.
 */
declare type WithChildren<T> = {
  children?: T
}

/**
 * Extend props to forbid any child to be passed to component.
 *
 * Most components don't do anything with children. It's a good
 * practice to inform consumer about it to avoid confusion of children
 * not rendered.
 */
declare type NoChildren = WithChildren<never>

/**
 * Extend props to allow only single react element as child.
 *
 * On rare occasions only a single child might be relevant and component
 * should inform about it to avoid accidental ignoration of multiple children.
 *
 * Use `Required<SingleChild>` if child is required.
 */
declare type SingleChild = WithChildren<ReactNode>

/**
 * Extend props to allow for single, multiple or none react elements as children.
 *
 * Most common to use when children are utilized and passed down to nested components.
 * For advanced patterns like render prop, the `WithChildren<T>` is more appropriate.
 *
 * Use `Required<SomeChildren>` if children are required.
 */
declare type SomeChildren = WithChildren<ReactNodes>

```

PR #38446, opened 2019-09-17
[react-calendar-timeline] Updated types to version 0.24.6

RIP21:
- added absent types
- more test cases
- removed old non-existent now props

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [Original repo](https://github.com/namespace-ee/react-calendar-timeline)


RIP21:
@acemac @Ilaiwi @radziksh Please review :)

RIP21:
As tests, I took some of the CodeSandbox examples + some from the documentation itself. All typings are based on documentation + a little on source code `propTypes` and other stuff that I have found. 
This thing is used in some pieces by other engineers in my company, and I saw too many hacks with typings, so I decided to do this job for the benefit of everyone.

@Ilaiwi Please, if you do some changes to API update it :) Or ping me, and I'll try to do my best to update them too :)

Ilaiwi:
@RIP21 could we extend this to include the latest version? I understand that CustomHeaders are a bit complicated but I can definitely help you in that

Ilaiwi:
@RIP21 oh I see you included them. I was tricked by the tile version. It should be `0.26.6`
amazing work! I will review this for sure and get back to you.

types/react-calendar-timeline/index.d.ts
```diff
@@ -1,131 +1,395 @@
...
+        resizeDetector?: (containerResizeDetector: any) => void;
+        verticalLineClassNamesForTime?: (start: number, end: number) => string[] | undefined;
+        horizontalLineClassNamesForGroup?: (group: CustomGroup) => string[];
+
+        // Fields that are in propTypes but not documented
+        stickyHeader?: boolean;
+        headerRef?: React.Ref<any>;
+        showCursorLine?: boolean;
+        headerLabelFormats?: TimelineHeaderLabelFormat;
+        subHeaderLabelFormats?: TimelineHeaderLabelFormat;
```
Ilaiwi: `subHeaderLabelFormats` and `headerLabelFormats` were removed from the package

types/react-calendar-timeline/index.d.ts
```diff
@@ -1,131 +1,395 @@
...
+        itemRenderer?: (props: ReactCalendarItemRendererProps<CustomItem>) => React.ReactNode;
+        groupRenderer?: (props: ReactCalendarGroupRendererProps<CustomGroup>) => React.ReactNode;
+        resizeDetector?: (containerResizeDetector: any) => void;
+        verticalLineClassNamesForTime?: (start: number, end: number) => string[] | undefined;
+        horizontalLineClassNamesForGroup?: (group: CustomGroup) => string[];
+
+        // Fields that are in propTypes but not documented
+        stickyHeader?: boolean;
+        headerRef?: React.Ref<any>;
+        showCursorLine?: boolean;
```
Ilaiwi: this one is removed as well

types/react-calendar-timeline/index.d.ts
```diff
@@ -1,131 +1,395 @@
...
+        ): any;
+        onBoundsChange?(canvasTimeStart: number, canvasTimeEnd: number): any;
+        itemRenderer?: (props: ReactCalendarItemRendererProps<CustomItem>) => React.ReactNode;
+        groupRenderer?: (props: ReactCalendarGroupRendererProps<CustomGroup>) => React.ReactNode;
+        resizeDetector?: (containerResizeDetector: any) => void;
+        verticalLineClassNamesForTime?: (start: number, end: number) => string[] | undefined;
+        horizontalLineClassNamesForGroup?: (group: CustomGroup) => string[];
+
+        // Fields that are in propTypes but not documented
+        stickyHeader?: boolean;
```
Ilaiwi: this one is removed as well

types/react-calendar-timeline/index.d.ts
RIP21: Found in master :) Kept :) But surely I'll remove :)

types/react-calendar-timeline/index.d.ts
RIP21: Will remove :)

types/react-calendar-timeline/index.d.ts
RIP21: Will remove :)

types/react-calendar-timeline/index.d.ts
Ilaiwi: I will make sure to remove them from the source code.

RIP21: @ilaiwi Take a loot at `labelFormat` at DataHeaderProps. Does it have `formatOptions` of this shape still? Or some different?

RIP21:
@Ilaiwi  BTW. What about `getRootProps` in `CustomHeaderPropsChildrenFnProps`
Any specific typings? I wasn't able to find anything about it in docs :)

types/react-calendar-timeline/index.d.ts
Ilaiwi: @rip21 this is intended to be for internal use only. I had it before as a param but decided not to include it in the public API to make things more clear.

RIP21: @ilaiwi so does it mean that
```typescript
    export interface DateHeaderProps<Data> {
        style?: React.CSSProperties;
        className?: string;
        unit?: Unit | 'primaryHeader';
        labelFormat?:
            | string
            | ((
                  [startTime, endTime]: [Moment, Moment],
                  unit: Unit,
                  labelWidth: number,
                  formatOptions: TimelineHeaderLabelFormat, // <== This can be removed?
              ) => string);
        intervalRenderer?: (props?: IntervalRenderer<Data>) => React.ReactNode;
        headerData?: Data;
        children?: (props: SidebarHeaderChildrenFnProps<Data>) => React.ReactNode;
        height?: number;
    }
```

Ilaiwi: yes, it should be removed. `short`, `medium` and `long` are very relative and might change. So removing it would be better

RIP21: @Ilaiwi  Also, this tuple `[startTime, endTime]: [Moment, Moment],` at the beginning feels a bit off. Is it tuple? Or `Interval` object instead?
{ startTime, endTime }
And whether it's Moment or it's actually `number | Date | Moment`

Ilaiwi:
@RIP21 it returns a style object checkout it our [here](https://github.com/namespace-ee/react-calendar-timeline#prop-getters-functions-1)

types/react-calendar-timeline/index.d.ts
RIP21: Okay, I got lost a bit :) 
Can you write typing here for 
```
labelFormat?:
            | string
            | ((
                  [startTime, endTime]: [Moment, Moment],
                  unit: Unit,
                  labelWidth: number,
                  formatOptions: TimelineHeaderLabelFormat, // <== This can be removed?
              ) => string);
```
Because I'm not sure what I need to remove :)

Ilaiwi: yes it is intended to be a tuple. is there a better way in typescript to define it as a tuple?
and yes they are moment object

Ilaiwi:
```
labelFormat?:
            | string
            | ((
                  [startTime, endTime]: [Moment, Moment],
                  unit: Unit,
                  labelWidth: number,
              ) => string);
```

RIP21: I think helper like such may help :) Have no idea why such is not included into the TS itself :)
```
type Tuple<LeftPart, RightPart> = [LeftPart, RightPart]

Tuple<Moment, Moment>
```

Thanks for the clarification :)

types/react-calendar-timeline/index.d.ts
```diff
@@ -1,131 +1,366 @@
...
+        id: number;
+        title: React.ReactNode;
+        rightTitle?: React.ReactNode;
+    }
+
+    export interface TimelineItemBase<DateType> {
+        id: number;
+        group: number;
+        title?: React.ReactNode;
+        start: DateType;
```
RIP21: @Ilaiwi hey man. One question. Do these should be `start_time` and `end_time` or `start` and `end` to work fine?
Because it stopped working for us with that but in CodeSandbox examples `start` and `end` is used and now I'm wondering what is wrong and what is right.

Ilaiwi: the problem here is that the package provides a "keys" object where you can change what keys you provide for items and groups. 
The issue here that this makes typescript support much harder but I would guess maybe a keys genric should be passed or inferred to the component.

RIP21: Ah gotcha, it's overridable.
image

IMO it's a bit useless flexibility :) I would rather write some mapping over some data to match it to the shape (what will be needed anyway) than add some odd keys mapping.

Will change it to `start_time` and `end_time`

Ilaiwi: I agree. legacy code is always an issue

RIP21:
Hey mate.
@RyanCavanaugh
Any chances to merge/review this. Reviewers seem to be absent. Except for the current maintainer of the lib itself that had already done a review and it's good to go as far as I can see.

Ilaiwi (approved):

RIP21:
@RyanCavanaugh anyone will review? :) It's been a week since :)

uniqueiniquity:
Thank you for your patience!

PR #38452, opened 2019-09-18
Add types for rox-browser

g-guirado:
rox-browser is the Javascript browser SDK for our https://rollout.io/ service.

### DIFF
--- a/types/rox-browser/index.d.ts
+++ b/types/rox-browser/index.d.ts
@@ -1,0 +1,273 @@
+// Type definitions for rox-browser 4.8
+// Project: https://rollout.io
+// Definitions by: g-guirado <https://github.com/g-guirado>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+/**
+ *
+ * Official documentation for rox-browser is available here:
+ * https://support.rollout.io/docs/javascript-browser-api
+ *
+ */
+
+export {};
+
+export interface RoxContainer {
+  [key: string]: Flag | Configuration<any> | Variant;
+}
+
+/**
+ * The register function should be called before the call to Rox.setup()
+ *
+ * https://support.rollout.io/docs/javascript-browser-api#section-register
+ */
+export function register(namespace: string, roxContainer: RoxContainer): void;
+
+/**
+ * Set Global Context.
+ * You can think of Global Context as a default context
+ *
+ * https://support.rollout.io/docs/context#section-global-context
+ */
+export function setContext(globalContext: unknown): void;
+
+/**
+ * Initiate connection with Rox servers for the application identified by the application key.
+ * The registered containers will be synced and Rox entities will get the appropriate values.
+ *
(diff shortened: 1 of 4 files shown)


ekeren (approved):

ekeren (approved):

types/rox-browser/index.d.ts
```diff
@@ -0,0 +1,273 @@
...
+
+  /**
+   * full flag name including namespace
+   *
+   * https://support.rollout.io/docs/javascript-browser-api#section--rox-overrides-hasoverride-
+   */
+  hasOverride(nameSpacedFlagName: string): boolean;
+}
+
+export const overrides: RoxOverrides;
```
uniqueiniquity: Instead of doing `export const` and defining some type to describe the available methods, I'd suggest doing the following:
```ts
export namespace overrides {
  ...
  export function clearAllOverrides(): void;
  ...
}
```

types/rox-browser/index.d.ts
```diff
@@ -0,0 +1,273 @@
...
+
+  // Unlock the Configuration value from changes from the last time it was freezed
+  unfreeze(): void;
+}
+
+/**
+ * Ensure that TypeScript properly types things with a basic type.
+ * For example, if T is true, returnedtype shall be boolean, not true
+ */
+type BasicType<T> = T extends boolean ? boolean : T extends number ? number : T extends string ? string : never;
```
uniqueiniquity: Unless you explicitly do not want this to be exported, you can remove the `export {}` on line 14.

types/rox-browser/index.d.ts
```diff
@@ -0,0 +1,273 @@
...
+}
+
+export const overrides: RoxOverrides;
+
+/**
+ * Dynamic API is an alternative to Rollout static API for defining flags on the
+ * different container objects and accessing them from that container object.
+ * https://support.rollout.io/docs/dynamic-api
+ */
+export class DynamicApi {
```
uniqueiniquity: If the `DynamicApi` class isn't intended to be constructed by consumers (that is, if you don't want users to do `new Rox.DynamicApi`) and if you don't expect users to do an `instanceof` check for this class, you may want to consider making this `export namespace dynamicApi { }` as I discussed in my previous comment.

uniqueiniquity: Actually, if you want to be able to do the `instanceof` checks but prevent the class from being constructable, you can add a `private constructor() {}`.

uniqueiniquity (changes requested):

types/rox-browser/index.d.ts
g-guirado: Indeed, thanks!
Change done.

types/rox-browser/index.d.ts
g-guirado: I've remove `export {}`

types/rox-browser/index.d.ts
g-guirado: There are no reasons to treat this differently from overrides above. Changed to `export namespace dynamicApi`

uniqueiniquity:

PR #38453, opened 2019-09-18
fix(react-router): Failing with strictFunctionTypes

eps1lon:
Resolves issues reported in https://github.com/DefinitelyTyped/DefinitelyTyped/issues/38271#issuecomment-532428873 and https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38326#issuecomment-532328360

I'll try to make sense of the current issue later in this text but TL;DR: The introduced bug reports have more upvotes than the fixed bug reports so we revert until we figure out what's happening.

I tried to fiddle around with the `withRouter` declaration but anything I tried created an error in a different version of TS with strictFunctionTypes.

The current issue is a result of `C & ComponentType<P>` which resolves (in the removed test case) to `FunctionComponent<P> | (Functioncomponent<P> & ComponentClass<P>)` which is probably a result of the intermediate distribution of `&` to `(FunctionComponent<P> & FunctionComponent<P>) | (FunctionComponent<P> & ComponentClass<P>) | (ComponentClass<P> & FunctionComponent<P>) | (FunctionComponent<P> & ComponentClass<P>)` which should be simplified to `FunctionComponent<P> | ComponentClass<P> | (FunctionComponent<P> | ComponentClass<P>)` but for some reason the `ComponentClass<P>` is dropped. Maybe TypeScript is using only the first part of the union and resolves the `C` in `C & ComponentType<P>` to `FunctionComponent<P> & ComponentType<P>` which would explain the result.

### DIFF
--- a/types/react-router/index.d.ts
+++ b/types/react-router/index.d.ts
@@ -143,12 +143,8 @@ export interface WithRouterStatics<C extends React.ComponentType<any>> {
 // they are decorating. Due to this, if you are using @withRouter decorator in your code,
 // you will see a bunch of errors from TypeScript. The current workaround is to use withRouter() as a function call
 // on a separate line instead of as a decorator.
-export function withRouter<C extends React.ComponentType<RouteComponentProps>>(
-  component: C,
-): React.ComponentClass<
-  Omit<React.ComponentProps<C>, keyof RouteComponentProps> & WithRouterProps<C>,
-  never
-> &
-  WithRouterStatics<C>;
+export function withRouter<P extends RouteComponentProps<any>, C extends React.ComponentType<P>>(
+  component: C & React.ComponentType<P>,
+): React.ComponentClass<Omit<P, keyof RouteComponentProps<any>> & WithRouterProps<C>> & WithRouterStatics<C>;
 
 export const __RouterContext: React.Context<RouteComponentProps>;
--- a/types/react-router/test/WithRouter.tsx
+++ b/types/react-router/test/WithRouter.tsx
@@ -13,8 +13,6 @@ const FunctionComponent: React.FunctionComponent<TOwnProps> = props => (
   <h2>Welcome {props.username}</h2>
 );
 
-declare const Component: React.ComponentType<TOwnProps>;
-
 class ComponentClass extends React.Component<TOwnProps> {
     render() {
         return <h2>Welcome {this.props.username}</h2>;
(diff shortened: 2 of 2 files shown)


OliverJAsh:
If we merge this can we make sure we re-open or re-file https://github.com/DefinitelyTyped/DefinitelyTyped/issues/38271 so it's not lost?

eps1lon:
> If we merge this can we make sure we re-open or re-file #38271 so it's not lost?

Yes. I probably have another fix for this and suspect it's actually an issue with TypeScript. I'll narrow it down a bit more and keep you posted (likely pinging you in a TS issue).

MayhemYDG:
Shouldn't we add another test to prevent future regressions?
In my case the new types break on a component that looks like this:
```ts
const Component = withRouter(({ a, b, location, ...props }: ComponentOwnProps & RouteComponentProps) => {
  // ...
});
```

OliverJAsh:
Thanks for opening https://github.com/microsoft/TypeScript/issues/33490

eps1lon:
> Shouldn't we add another test to prevent future regressions?
> In my case the new types break on a component that looks like this:
> 
> ```ts
> const Component = withRouter(({ a, b, location, ...props }: ComponentOwnProps & RouteComponentProps) => {
>   // ...
> });
> ```

I'll check it out later. I think main issue that caused this regression was that the tests do not run with `strictFunctionTypes`. Could you check if the regression persists after disabling this option in your tsconfig?

MayhemYDG:
`"strictFunctionTypes": false` does remove the error.
Shouldn't types be tested with `"strict": true` as well?

eps1lon:
> `"strictFunctionTypes": false` does remove the error.
> Shouldn't types be tested with `"strict": true` as well?

The default for `@types/` packages does not include this, no. I'm not sure about the implications but I suspect it would require package consumer to enable this as well to be sure they get working types. For example since all types are tested with `strictNullChecks` they might behave differently without it. Package consumers without that option basically use untested types.

MayhemYDG:
I would have expected the opposite: strict types work with loose settings, while loose types will break with strict settings.

orta:
👍  looks good to me, we should get this fix in. 

I think it might be worth moving to have `strict: true` also, because strict TS is effectively a subset of loose TS which is effectively a subset of loose JS 🍡

eps1lon:
> I think it might be worth moving to have strict: true also, because strict TS is effectively a subset of loose TS which is effectively a subset of loose JS dango

I need to revisit how we implemented the type helper I have in mind but one issue with `strictNullChecks` is that conditionals can behave differently. This means you can create different types with e.g. `undefined extends T ? U : never`. So you have to be very cautious with every conditional you use and consider different configurations.

PR #38462, opened 2019-09-18
add readInlineData

itajaja:
tested locally. should I add tests?

### DIFF
--- a/types/react-relay/test/react-relay-tests.tsx
+++ b/types/react-relay/test/react-relay-tests.tsx
@@ -5,17 +5,19 @@ import * as React from 'react';
 import { Environment, Network, RecordSource, Store, ConnectionHandler } from 'relay-runtime';
 
 import {
-    graphql,
     commitMutation,
     createFragmentContainer,
     createPaginationContainer,
     createRefetchContainer,
-    requestSubscription,
+    graphql,
     QueryRenderer,
     ReactRelayContext,
-    RelayRefetchProp,
+    readInlineData,
     RelayPaginationProp,
     RelayProp,
+    RelayRefetchProp,
+    requestSubscription,
+    _FragmentRefs,
 } from 'react-relay';
 
 // ~~~~~~~~~~~~~~~~~~~~~
@@ -593,6 +595,16 @@ function markNotificationAsRead(source: string, storyID: string) {
         },
     });
 }
+
+// ~~~~~~~~~~~~~~~~~~~~~
+// readInlineData
+// ~~~~~~~~~~~~~~~~~~~~~
+
+function functionWithInline(storyRef: _FragmentRefs<Story_story>): Story_story {
+    return readInlineData(graphql``, storyRef)
+}
+
+functionWithInline({ ' $fragmentRefs': 1 as any });
 
 // ~~~~~~~~~~~~~~~~~~~~~
 // Modern Subscriptions
--- a/types/react-relay/test/react-relay-tests.tsx
+++ b/types/react-relay/test/react-relay-tests.tsx
@@ -600,10 +600,18 @@ function markNotificationAsRead(source: string, storyID: string) {
 // readInlineData
 // ~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


itajaja:
cc @taion @jquense

taion:
- should update version to v6, since this is in relay v6
- does the relay compiler distinguish inline fragments from normal fragments at all? possibly not, i'm sure, but it'd be neat if it did

itajaja:
> should update version to v6, since this is in relay v6

the version is not specified, except here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38430/files#diff-9bf0e4126d019d96bea44f758a214a1aR1

alloy (changes requested):
Awesome ✨ 

Yes, some test coverage would be great, especially around the fragment reference checking.

types/react-relay/test/react-relay-tests.tsx
```diff
@@ -5,17 +5,19 @@ import * as React from 'react';
 import { Environment, Network, RecordSource, Store, ConnectionHandler } from 'relay-runtime';
 
 import {
-    graphql,
     commitMutation,
     createFragmentContainer,
     createPaginationContainer,
     createRefetchContainer,
-    requestSubscription,
+    graphql,
```
itajaja: I just sorted the imports alphabetically

itajaja:
the only question I have is that by using `readInlineFragment`, the consumer who is writing the function will need `_FragmentRefs`, which, given the _ sounds a little internal (even though is exported). I can't think of a better alternative to type that function, so I think that it might be better to rename that type?

types/react-relay/test/react-relay-tests.tsx
alloy: That's fine, prettier was added to DT recently-ish 👍

types/react-relay/test/react-relay-tests.tsx
```diff
@@ -594,6 +596,16 @@ function markNotificationAsRead(source: string, storyID: string) {
...
 
+// ~~~~~~~~~~~~~~~~~~~~~
+// readInlineData
+// ~~~~~~~~~~~~~~~~~~~~~
+
+function functionWithInline(storyRef: _FragmentRefs<Story_story>): Story_story {
+    return readInlineData(graphql``, storyRef)
+}
+
+functionWithInline({ ' $fragmentRefs': 1 as any });
```
alloy: Can you please use a more realistic example and no casting to `any`?

alloy:
Yeah I think you’re right and the consumer will need to use that type, so re-naming sounds good 👍

itajaja:
> Yeah I think you’re right and the consumer will need to use that type, so re-naming sounds good

I acutally changed to use `FragmentOrRegularProp` so no need to rename

types/react-relay/test/react-relay-tests.tsx
```diff
@@ -594,6 +596,18 @@ function markNotificationAsRead(source: string, storyID: string) {
...
 }
 
+// ~~~~~~~~~~~~~~~~~~~~~
+// readInlineData
+// ~~~~~~~~~~~~~~~~~~~~~
+
+function functionWithInline(
+  storyRef: FragmentOrRegularProp<Story_story>,
+): Story_story {
+  return readInlineData<Story_story>(graphql``, storyRef);
```
alloy: In this example, would the `Story_story` fragment be defined in this (empty) `graphql` template string?

(And, if so, can you add it? 🙏)

itajaja: sure, but I don't think that would in any way affect the tests. as `graphql` returns a tagged node whatever the string content is. I just saw other usages of `graphql`` ` so I just used. I can change it for the sake of clarity/documentation though

alloy: My request is about the latter.

alloy (approved):

alloy:
Perfect, thanks for bearing with me! 🙏

PR #38474, opened 2019-09-19
Adds types for version 2.0 of svg-parser

JoelCodes:
Please fill in this template.

If adding a new definition:

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: [NPM Package (with versions)](https://www.npmjs.com/package/svg-parser)

If removing a declaration:

### DIFF
--- a/types/svg-parser/index.d.ts
+++ b/types/svg-parser/index.d.ts
@@ -1,5 +1,5 @@
 // Type definitions for svg-parser 2.0
-// Project: https://gitlab.com/Rich-Harris/svg-parser#README
+// Project: https://github.com/Rich-Harris/svg-parser
 // Definitions by: mrmlnc <https://github.com/mrmlnc>
 //                 Joel Shinness <https://github.com/JoelCodes>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
@@ -11,9 +11,9 @@ export interface TextNode {
 }
 
 export interface ElementNode {
-    type: string;
+    type: 'element';
     tagName?: string;
-    properties?: {};
+    properties?: Record<string, string|number>;
     children: Array<Node | string>;
     value?: string;
     metadata?: string;
@@ -21,4 +21,9 @@ export interface ElementNode {
 
 export type Node = TextNode | ElementNode;
 
-export function parse(source: string): Node;
+export interface RootNode {
+    type: 'root';
+    children: Node;
+}
+
+export function parse(source: string): RootNode;
--- a/types/svg-parser/index.d.ts
+++ b/types/svg-parser/index.d.ts
@@ -1,14 +1,24 @@
-// Type definitions for svg-parser 1.0
+// Type definitions for svg-parser 2.0
 // Project: https://gitlab.com/Rich-Harris/svg-parser#README
 // Definitions by: mrmlnc <https://github.com/mrmlnc>
+//                 Joel Shinness <https://github.com/JoelCodes>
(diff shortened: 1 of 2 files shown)


types/svg-parser/index.d.ts
```diff
@@ -1,14 +1,24 @@
-// Type definitions for svg-parser 1.0
+// Type definitions for svg-parser 2.0
 // Project: https://gitlab.com/Rich-Harris/svg-parser#README
```
mrmlnc: The `svg-parser@2.x.x` now lives on GitHub.

https://github.com/Rich-Harris/svg-parser

types/svg-parser/index.d.ts
```diff
@@ -1,14 +1,24 @@
...
+    type: string;
+    tagName?: string;
+    properties?: {};
+    children: Array<Node | string>;
+    value?: string;
     metadata?: string;
 }
 
-export function parse(content: string): Node;
+export type Node = TextNode | ElementNode;
```
mrmlnc: Unfortunately, now this type is not fully correct.

The second version of the packages is always returns `root` node:

```js
{ type: 'root', children: [] }
```

https://github.com/Rich-Harris/svg-parser/blob/734f11b37397eafed09a4688a505972d9f5613b7/src/index.js#L231-L234
https://github.com/Rich-Harris/svg-parser/blob/master/test/samples/empty-svg/output.json

We can make this place better by always returning the `RootNode` node. Something like it:

```js
export type RootNode {
  type: 'root';
  children: TextNode | ElementNode;
}
```

types/svg-parser/index.d.ts
```diff
@@ -1,14 +1,24 @@
...
-    children: Node[];
+export interface TextNode {
+    type: 'text';
+    value?: string | boolean | number;
+}
+
+export interface ElementNode {
+    type: string;
+    tagName?: string;
+    properties?: {};
```
mrmlnc: We can restrict the types in this location. The `properties` can be `Record<string, string | number>`.

types/svg-parser/index.d.ts
```diff
@@ -1,14 +1,24 @@
...
-    name: string;
-    attributes: Record<string, string | number>;
-    children: Node[];
+export interface TextNode {
+    type: 'text';
+    value?: string | boolean | number;
+}
+
+export interface ElementNode {
+    type: string;
```
mrmlnc: As I understand from the tests, an element always has the `element` type.

https://github.com/Rich-Harris/svg-parser/blob/734f11b37397eafed09a4688a505972d9f5613b7/src/index.js#L74

mrmlnc (changes requested):

types/svg-parser/index.d.ts
JoelCodes: That's a good catch!  Thanks!

types/svg-parser/index.d.ts
JoelCodes: I've changed to that.  I was being a little more general, but I do like making it more specific.

types/svg-parser/index.d.ts
JoelCodes: That follows.  I've updated this to become more specific.

types/svg-parser/index.d.ts
JoelCodes: I hadn't even checked!  Thanks for that!

mrmlnc (approved):

PR #38484, opened 2019-09-19
Fixed a compile error in c3 typings

josundt:
Please fill in this template.

- [**x**] Test the change in your own code. (Compile and run.)
- [**N/A**] Add or edit tests to reflect the change. (Run with `npm test`.)
- [**x**] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: **N/A - compile failed for typings using typescript 3.6.3 - the PR is a bug fix that does not alter the typings any other way than fix this bug**

### DIFF
--- a/types/c3/index.d.ts
+++ b/types/c3/index.d.ts
@@ -739,10 +739,6 @@ export interface YAxisConfigurationWithTime extends YAxisConfiguration {
 
 export interface TickConfiguration {
     /**
-     * A function to format tick value. Format string is also available for timeseries data.
-     */
-    format?: string | ((x: number | Date) => string | number);
-    /**
      * Show x axis outer tick.
      */
     outer?: boolean;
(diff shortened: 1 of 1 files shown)


iansan5653:
What exactly is the compile error? Can you provide more info on that?

iansan5653 (commented):
`YTickConfiguration.format` should definitely have it's own JSDoc description and I'm not sure that it should be required. ~Can you confirm if it works without it?~

Edit: I just confirmed it's definitely not required. Not sure why I missed that before but might as well fix it in this PR.

types/c3/index.d.ts
```diff
@@ -765,31 +761,38 @@ export interface TickConfiguration {
 }
 
 export interface XTickConfiguration extends TickConfiguration {
```
iansan5653: Indentation for this block should be 4 spaces like in the rest of the file.

josundt: @iansan5653: 
I have updated the PR source branch and made the following changes:
* Indentation for the `XTickConfiguration` interface body fixed
* Added JSDoc comment to `YTickConfiguration.format`
* Made `YTickConfiguration.format` function optional and allow `number | string` return value - verified that it works and is correct

> What exactly is the compile error? Can you provide more info on that?

Here's the compile error that caused my build to break originally and made me create this PR:
```
Interface 'YTickConfiguration' incorrectly extends interface 'TickConfiguration'.
  Types of property 'format' are incompatible.
    Type '(x: number) => string' is not assignable to type 'string | ((x: number | Date) => string | number) | undefined'.
      Type '(x: number) => string' is not assignable to type '(x: number | Date) => string | number'.
        Types of parameters 'x' and 'x' are incompatible.
          Type 'number | Date' is not assignable to type 'number'.
            Type 'Date' is not assignable to type 'number'.ts(2430)
```

iansan5653 (approved):
Looks good now, thanks!

uniqueiniquity:

PR #38499, opened 2019-09-20
enzyme-react-intl typings 

mateusz-meller:

If adding a new definition:

### DIFF
--- a/types/enzyme-react-intl/index.d.ts
+++ b/types/enzyme-react-intl/index.d.ts
@@ -1,0 +1,41 @@
+// Type definitions for enzyme-react-intl 2.0
+// Project: https://github.com/joetidee/enzyme-react-intl#readme
+// Definitions by: Mateusz Meller <https://github.com/mateusz-meller>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.1
+
+import { Component, ReactElement } from 'react';
+import { ShallowRendererProps, MountRendererProps, ShallowWrapper, ReactWrapper } from 'enzyme';
+
+// shallow methods
+
+export function shallowWithIntl<C extends Component, P = C['props'], S = C['state']>(
+    node: ReactElement<P>,
+    options?: ShallowRendererProps,
+): ShallowWrapper<P, S, C>; // tslint:disable-line no-unnecessary-generics
+// tslint:disable-next-line no-unnecessary-generics
+export function shallowWithIntl<P>(node: ReactElement<P>, options?: ShallowRendererProps): ShallowWrapper<P, any>;
+// tslint:disable-next-line no-unnecessary-generics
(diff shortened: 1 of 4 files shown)


uniqueiniquity (changes requested):
I noticed that you've written these type definitions in a way that the linter doesn't agree with, but didn't write any tests demonstrating the usage of the features you're aiming for by disabling the linter.

Please either write tests demonstrating why you believe these extra generics are necessary or write them in such a way that the lint rule doesn't need to be disabled. 

For example, unless you have a particular usage in mind, `shallowWithIntl` should be defined as follows:
`export function shallowWithIntl<P>(node: ReactElement<P>, options?: ShallowRendererProps): ShallowWrapper<P>`.

Thank you!

mateusz-meller:
I added unit test related to generic typings. Can you explain me why:
image
constructed is the worst identifier?

uniqueiniquity:
I don't really have a good reason for why that's the case. Likely it's because of how the type of `constructed` depends on the type argument you provide when calling `shallowWithIntl`. It's not really a good pattern in TS to encourage type arguments to be provided at a method call - that's why we have the "no-unnecessary-generics" lint rule. I suspect that by doing it in this way, the performance is worse than if you just expected users to cast the result of `state()`.

types/enzyme-react-intl/index.d.ts
```diff
@@ -0,0 +1,41 @@
...
+import { Component, ReactElement } from 'react';
+import { ShallowRendererProps, MountRendererProps, ShallowWrapper, ReactWrapper } from 'enzyme';
+
+// shallow methods
+
+export function shallowWithIntl<C extends Component, P = C['props'], S = C['state']>(
+    node: ReactElement<P>,
+    options?: ShallowRendererProps,
+): ShallowWrapper<P, S, C>; // tslint:disable-line no-unnecessary-generics
+// tslint:disable-next-line no-unnecessary-generics
```
uniqueiniquity: This `tslint:disable` is unnecessary.

types/enzyme-react-intl/index.d.ts
```diff
@@ -0,0 +1,41 @@
...
+export function shallowWithIntl<P, S>(node: ReactElement<P>, options?: ShallowRendererProps): ShallowWrapper<P, S>;
+
+// mount methods
+
+export function mountWithIntl<C extends Component, P = C['props'], S = C['state']>(
+    node: ReactElement<P>,
+    options?: MountRendererProps,
+): ReactWrapper<P, S, C>; // tslint:disable-line no-unnecessary-generics
+// tslint:disable-next-line no-unnecessary-generics
```
uniqueiniquity: This tslint:disable is unnecessary.

uniqueiniquity (changes requested):

uniqueiniquity (approved):

uniqueiniquity:

PR #38504, opened 2019-09-20
sqlite3-promise type definitions

jonathanlb:
Please fill in this template.

If adding a new definition:
- [ ] `tsconfig.json` should have `noImplicitAny`, `noImplicitThis`, `strictNullChecks`, and `strictFunctionTypes` set to `true`.   The types/sqlite3/tsconfig.json file is copied, as the new package wraps sqlite3 with bluebird promises.

### DIFF
--- a/types/sqlite3-promise/index.d.ts
+++ b/types/sqlite3-promise/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for node-sqlite3-promise 1.0
+// Type definitions for sqlite3-promise 1.0
 // Project: https://github.com/Aminadav/node-sqlite3-promise
 // Definitions by: Jonathan Bredin <https://github.com/jonathanlb>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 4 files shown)


uniqueiniquity (changes requested):
Rather than duplicating the types from `sqlite3` as you have here, it would be much cleaner to do the following:
```ts
import * as sqlite3 from 'sqlite3';
export * from 'sqlite3';

declare module 'sqlite3' {
    export interface Database {
        allAsync(sql: string): Promise<any[]>;
        closeAsync(): Promise<void>;
        eachAsync(sql: string, cb?: (this: sqlite3.Statement, err: Error | null, row: any) => void): Promise<number>;
        eachAsync(sql: string, params: any, cb?: (this: sqlite3.Statement, err: Error | null, row: any) => void): Promise<number>;
        execAsync(sql: string): Promise<sqlite3.Statement>;
        getAsync(sql: string): Promise<any>;
        runAsync(sql: string): Promise<void>;
    }
}
```

This will re-export the actual types from `sqlite3`, so you don't have to maintain those, and will add your new methods to the `Database` class.

uniqueiniquity:

uniqueiniquity (approved):

PR #38507, opened 2019-09-20
react-native: Fix commandArgs type for dispatchViewManagerCommand

cozmo:
The `dispatchViewManagerCommand` method has a subtle but dangerous bug in the types. We define the last argument (`commandArgs`) as optional, but `react-native` is strict about argument count https://github.com/facebook/react-native/blob/84f5ebe4f941e752ba96b4d4a75dc6c6f5faa665/React/Base/RCTModuleMethod.mm#L512-L536. 

This means if you leave off the `undefined` your code will not work. This can be seen in practice here https://github.com/react-native-community/react-native-webview/pull/886 - where trusting the incorrect types caused a breaking change to sneak by.

I tried to run `npm run lint react-native` but got a bunch of errors that weren't caused by this change? 

```
./DefinitelyTyped/types/react-native/index.d.ts:1:1
ERROR: 1:1  npm-naming               d.ts file must have a matching npm package.
```

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/react-native/blob/84f5ebe4f941e752ba96b4d4a75dc6c6f5faa665/React/Base/RCTModuleMethod.mm#L512-L536

### DIFF
--- a/types/react-native/test/ui-manager.tsx
+++ b/types/react-native/test/ui-manager.tsx
@@ -1,0 +1,4 @@
+import { UIManager } from 'react-native';
+
+UIManager.dispatchViewManagerCommand(1, 1, undefined);
+UIManager.dispatchViewManagerCommand(1, 1, ["foo", "bar", "baz", 1, 2, 3]);
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -7879,7 +7879,7 @@ export interface UIManagerStatic {
      * commandID - Id of the native method that should be called.
      * commandArgs - Args of the native method that we can pass from JS to native.
      */
-    dispatchViewManagerCommand: (reactTag: number | null, commandID: number, commandArgs?: Array<any>) => void;
+    dispatchViewManagerCommand: (reactTag: number | null, commandID: number, commandArgs: Array<any> | undefined) => void;
 }
 
 export interface SwitchPropsIOS extends ViewProps {
--- a/types/react-native/tsconfig.json
+++ b/types/react-native/tsconfig.json
@@ -23,6 +23,7 @@
         "test/init-example.tsx",
         "test/ART.tsx",
         "test/legacy-properties.tsx",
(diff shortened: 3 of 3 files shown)


types/react-native/test/ui-manager.tsx
```diff
@@ -0,0 +1,4 @@
+import { UIManager } from 'react-native';
+
+UIManager.dispatchViewManagerCommand(1, 1, undefined);
```
TheSavior: Your previous example in webview used null, not undefined. Are you sure undefined is supported?

TheSavior: For those that look later, this is the type in core: https://github.com/facebook/react-native/blob/91f139b94118fe8db29728ea8ad855fc4a13f743/Libraries/ReactNative/NativeUIManager.js#L45

cozmo: Yea, in fact the hot fix for the issue that was caused by accidentally following these types switched to using undefined. https://github.com/react-native-community/react-native-webview/pull/886

In practice either works but this is a smaller delta in terms of changing the types and also feels more accurate for how that argument is meant to be used.

elicwhite (approved):
Alright, sounds good. :thumbsup:

types/react-native/test/ui-manager.tsx
alloy: `?foo` in Flow can be `null` or `undefined`, so really it should be both, but we haven’t done that consistently, if at all, it looks to be mostly `null` throughout the file. However, in this case `reactTag`, which is part of the same signature, is typed as `number | null`, so I think that in this case it should follow that to at least be consistent with other parameters within the same signature.

alloy (changes requested):
Thanks for the fix! 🙏 

This looks good, although I do think we should go with `null` here.

alloy:
@cozmo Let us know if you want to iterate and re-open this PR.

PR #38510, opened 2019-09-20
[relay-runtime] Expand and start splitting in files

alloy:
Like the subject says.


alloy:
Hmm, something must have gone really wrong. Will, re-open when I've had the time to clean this up.

PR #38511, opened 2019-09-20
Adding a new package: jwk-to-pem

esilkensen:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/jwk-to-pem/index.d.ts
+++ b/types/jwk-to-pem/index.d.ts
@@ -3,7 +3,11 @@
 // Definitions by: Erik Silkensen <https://github.com/esilkensen>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-declare namespace jwkToPem {
+export = jwkToBuffer;
+
+declare function jwkToBuffer(jwk: jwkToBuffer.JWK, opts?: jwkToBuffer.Options): string;
+
+declare namespace jwkToBuffer {
     interface Options {
         private: boolean;
     }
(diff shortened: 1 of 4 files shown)


types/jwk-to-pem/index.d.ts
```diff
@@ -0,0 +1,44 @@
...
+        dq?: string;
+        qi?: string;
+    }
+
+    type JWK = EC | ECPrivate | RSA;
+
+    type jwkToBuffer = (jwk: JWK, opts?: Options) => string;
+}
+
+declare const jwkToPem: jwkToPem.jwkToBuffer;
```
uniqueiniquity: This appears to be a callable module. Instead of declaring this as a const, I recommend following [this template](http://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-function-d-ts.html).

uniqueiniquity (changes requested):

types/jwk-to-pem/index.d.ts
esilkensen: Nice, thank for the pointer!

types/jwk-to-pem/index.d.ts
```diff
@@ -0,0 +1,43 @@
+// Type definitions for jwk-to-pem 2.0
+// Project: https://github.com/Brightspace/node-jwk-to-pem#readme
+// Definitions by: Erik Silkensen <https://github.com/esilkensen>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export = jwkToBuffer;
```
esilkensen: I don't know whether it matters or one name would be better than the other - `jwkToBuffer` vs. `jwkToPem`? (`jwkToBuffer` matches the source definition)

uniqueiniquity: I think matching the source definition is best here. It doesn't really matter, since the name you use here is not externally visible.

uniqueiniquity:

PR #38516, opened 2019-09-21
Added type definitions for react-image-magnify

sumitparakh:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/react-image-magnify/index.d.ts
+++ b/types/react-image-magnify/index.d.ts
@@ -11,7 +11,7 @@ import * as React from 'react';
  */
 export default function ReactImageMagnify(props: ReactImageMagnifyProps & Readonly<{ children?: React.ReactNode }>): React.ReactElement;
 
-interface CommonImageType {
+export interface CommonImageType {
     src: string;
     srcSet?: string;
     sizes?: string;
@@ -19,7 +19,7 @@ interface CommonImageType {
     onError?: () => void;
 }
 
-interface SmallImageType extends CommonImageType {
+export interface SmallImageType extends CommonImageType {
     /**
      * Required if isFluidWidth is not set
      */
@@ -37,7 +37,7 @@ interface SmallImageType extends CommonImageType {
     isFluidWidth: boolean;
 }
 
-interface LargeImageType extends CommonImageType {
+export interface LargeImageType extends CommonImageType {
     width: number;
     height: number;
 
--- a/types/react-image-magnify/index.d.ts
+++ b/types/react-image-magnify/index.d.ts
@@ -1,0 +1,224 @@
+// Type definitions for react-image-magnify 2.7
+// Project: https://github.com/ethanselzer/react-image-magnify
+// Definitions by: Sumit Parakh <https://github.com/sumitparakh>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+import * as React from 'react';
+
+/**
+ * React Image Magnify
+ */
+export default function ReactImageMagnify(props: ReactImageMagnifyProps & Readonly<{ children?: React.ReactNode }>): React.ReactElement;
+
+interface CommonImageType {
+    src: string;
+    srcSet?: string;
+    sizes?: string;
+    onLoad?: () => void;
+    onError?: () => void;
+}
+
+interface SmallImageType extends CommonImageType {
+    /**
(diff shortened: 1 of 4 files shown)


types/react-image-magnify/index.d.ts
```diff
@@ -0,0 +1,229 @@
+// Type definitions for react-image-magnify 2.7
+// Project: https://github.com/gaearon/react-side-effect
```
uniqueiniquity: This seems like the wrong link.

types/react-image-magnify/index.d.ts
```diff
@@ -0,0 +1,229 @@
...
+// Definitions by: Sumit Parakh <https://github.com/sumitparakh>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+/// <reference types="react"/>
+
+declare module 'react-image-magnify' {
```
uniqueiniquity: This module declaration is not needed.
You can just `export default ReactImageMagnify` directly.
If you want the ability to import each of these interfaces as well, you can export them directly as well.

uniqueiniquity (changes requested):

types/react-image-magnify/index.d.ts
sumitparakh: It is updated now.

types/react-image-magnify/index.d.ts
sumitparakh: Done.

sumitparakh:
Hi @uniqueiniquity . I've made required changes as per your feedback. Can you look at it now?

types/react-image-magnify/index.d.ts
```diff
@@ -0,0 +1,228 @@
...
+// Definitions by: Sumit Parakh <https://github.com/sumitparakh>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+import * as React from 'react';
+
+/**
+ * React Image Magnify
+ */
+declare const ReactImageMagnify: (props: ReactImageMagnifyProps & Readonly<{ children?: React.ReactNode }>) =>
```
uniqueiniquity: Instead of declaring this as a const with a function type, I suggest the following:

```ts
export default class ReactImageMagnify extends React.Component<ReactImageMagnifyProps> {}
```

types/react-image-magnify/tsconfig.json
```diff
@@ -0,0 +1,25 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6",
+            "dom"
+        ],
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "strictNullChecks": false,
```
uniqueiniquity: Please make this `true`; otherwise, it will be difficult for consumers of this package to use `strictNullChecks`.

types/react-image-magnify/tslint.json
```diff
@@ -0,0 +1,80 @@
+{
```
uniqueiniquity: This file should just be `{ "extends": "dtslint/dt.json" }`.

uniqueiniquity (changes requested):

types/react-image-magnify/index.d.ts
sumitparakh: Problem is functional declaration is working in both stateless and statefull component.. But it stops working in stateless function component once i declare it like you suggested..  :/

uniqueiniquity: That's odd... can you add a test case demonstrating the issue that you're seeing? This will help make sure no one breaks it by making this change in the future.

sumitparakh: Hi @uniqueiniquity , I've added test cases for both stateless and stateful component.

types/react-image-magnify/tsconfig.json
sumitparakh: Done

types/react-image-magnify/tslint.json
sumitparakh: Done

sumitparakh:
Hi @uniqueiniquity . I've updated changes as per your review.. Can you check now?

uniqueiniquity (approved):

uniqueiniquity:
Sorry for the delay - looks great now!

Issue #38528, opened 2019-09-23
[@type/react-infinite-calendar] locale.locale is not included


qkreltms:
It is related with issue ```locale not apply #220``` (https://github.com/clauderic/react-infinite-calendar/issues/220)

I'm using Typescript with React and I tried 

datePicker.tsx
```js
     import ko from "date-fns/locale/ko";
 //...
<InfiniteCalendar
          locale={{
            locale: ko,
            headerFormat: "YYYY MM DD",
            weekdays: ["월", "화", "수", "목", "금", "토", "일"],
            blank: "날짜를 선택하세요.",
            todayLabel: {
              long: "오늘"
            }
          }}
/>
```

index.d.ts
```diff
    locale?: {
+       locale?: any;
        blank?: string;
        headerFormat?: string;
        todayLabel?: {
            long: string;
        };
        weekdays?: string[];
        weekStartsOn?: 0 | 1 | 2 | 3 | 4 | 5 | 6;
    };
```

It seems ```locale.locale``` is not included in index.d.ts

- [x] I tried using the `@types/react-infinite-calendar` package and had problems.
- [x] I tried using the latest stable version of @types/react-infinite-calendar": "^2.3.3
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @christianchown

orta:
Hi thread, we're moving DefinitelyTyped to use [GitHub Discussions](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/53377) for conversations the `@types` modules in DefinitelyTyped.

To help with the transition, we're closing all issues which haven't had activity in the last 6 months, which includes this issue. If you think closing this issue is a mistake, please pop into the [TypeScript Community Discord](https://discord.gg/typescript) and mention the issue in the `definitely-typed` channel.

PR #38529, opened 2019-09-23
[@type/react-infinite-calendar]update: locale.locale

qkreltms:
This is related with https://github.com/DefinitelyTyped/DefinitelyTyped/issues/38528

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/clauderic/react-infinite-calendar>>

### DIFF
--- a/types/react-infinite-calendar/index.d.ts
+++ b/types/react-infinite-calendar/index.d.ts
@@ -47,6 +47,7 @@ export interface ReactInfiniteCalendarProps {
         todayHelperRowOffset?: number;
     };
     locale?: {
+        locale?: any;
         blank?: string;
         headerFormat?: string;
         todayLabel?: {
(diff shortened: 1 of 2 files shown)


christianchown (changes requested):
Hi - can we be more specific with `locale` than `any`? I'm thinking that it should match the date-fns definition here:

https://github.com/date-fns/date-fns/blob/master/typings.d.ts#L35-L63

bart-lewandowski-conichi:
this was not approved ? how come ?

qkreltms:
I'll do this soon.

qkreltms:
@christianchown I've searched documentation when I found you referenced ```date-fns: 2.x.x```, but calendar uses ```"date-fns": "^1.27.2"``` as a result of this I think we should process this issues when react-infinite-calendar move date-fns version to 2.x.x

bart-lewandowski-conichi:
also using 1.27 version of date-fns fix the issue and all works like it should im sorry i didnt wrote it right after i made it work :<

PR #38530, opened 2019-09-23
[chrome] add the missing content_scripts::match_about_blank

tophf:
This is a trivial change, just one missing property was added.  
No tests needed but I did test it in IDE and observed the change successfully as expected.

URL to documentation or source code:
https://developer.chrome.com/extensions/content_scripts#match_about_blank


fredrikhr (approved):
LGTM :+1:

PR #38536, opened 2019-09-23
[relay-runtime] Expand slightly and start splitting into files.

alloy:
Adds an export here and there and introduces `ReaderInlineDataFragment`, which is a v6 addition.

Also, I've started moving typings into their own files, so that we can export from the main entry point only those things that the actual package exports and we can import selectively from specific files.


alloy:
Going for it, seeing as this is largely a refactor and my relay-compiler changes are dependent on it.

PR #38539, opened 2019-09-23
[relay-compiler] Add requirements for relay-compiler-language-typescript

alloy:
Moves any missing types from https://github.com/relay-tools/relay-compiler-language-typescript/tree/83962d2b059d65a5dfd8a9bf7fa28be62fb7cbc3/types to here.

Also ran prettier.

### DIFF
--- a/types/relay-compiler/lib/core/GraphQLIRVisitor.d.ts
+++ b/types/relay-compiler/lib/core/GraphQLIRVisitor.d.ts
@@ -45,14 +45,17 @@ export type VisitNode =
     | Stream
     | Variable;
 
-type EnterLeave<T> = { readonly enter?: T; readonly leave?: T };
+export interface EnterLeave<T> {
+    readonly enter?: T;
+    readonly leave?: T;
+}
 
 export type VisitFn<T extends VisitNode> = (
     node: T, // node we're visiting
     key?: any, // index/key to node from parent array/object
-    parent?: null | VisitNode | Array<VisitNode>, // Object immediately above node
-    path?: Array<any>, // keys to get from root: [keyForChild, ..., keyForParent]
-    ancestors?: Array<VisitNode | Array<VisitNode>>, // [root, child1, ..., grandparent]
+    parent?: null | VisitNode | VisitNode[], // Object immediately above node
+    path?: any[], // keys to get from root: [keyForChild, ..., keyForParent]
+    ancestors?: Array<VisitNode | VisitNode[]>, // [root, child1, ..., grandparent]
 ) => // Note: ancestors includes arrays which contain the visited node
 // These correspond to array indices in `path`.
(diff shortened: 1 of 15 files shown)


n1ru4l:
@alloy You have to add the files that are not referenced anywhere to the `tsconfig.json`

types/relay-compiler/lib/core/GraphQLIRVisitor.d.ts
```diff
@@ -0,0 +1,109 @@
...
+    | ScalarField
+    | SplitOperation
+    | Stream
+    | Variable;
+
+type EnterLeave<T> = { readonly enter?: T; readonly leave?: T };
+
+export type VisitFn<T extends VisitNode> = (
+    node: T, // node we're visiting
+    key?: any, // index/key to node from parent array/object
```
n1ru4l: I would prefer `unknown`

alloy: Hmm, yeah I get that it would normally be the preferable type, but this is a straight port [from upstream](https://github.com/facebook/relay/blob/v6.0.0/packages/relay-compiler/core/GraphQLIRVisitor.js#L90) where they also went with `any`. I guess I prefer to keep things easily portable, but don’t feel strongly about that, so if you do then I'll update this 👍

alloy:
> You have to add the files that are not referenced anywhere to the `tsconfig.json`

This is the first time I’ve encountered this… thanks for letting me know 😅

types/relay-compiler/lib/core/GraphQLIRVisitor.d.ts
n1ru4l: Hm then it might be okay. How is any from flow different than typescripts any?

alloy:
Hmm. I think that to resolve this last failure we have to add the `dom` types to this package, just because `relay-runtime` relies on those. I'm unsure why that would have succeeded in the past then 🤔 I guess I'll at least _try_ to see if that's the issue.

types/relay-compiler/lib/core/GraphQLIRVisitor.d.ts
alloy: Afaik they are pretty much equivalent. Flow’s `mixed` is similar to TS’ `unknown`.

alloy:
Phew, didn’t think we'd ever see light at the end of the tunnel 😅

n1ru4l (approved):
Legit! 🎉 thank you @alloy 🔥

types/relay-compiler/lib/core/GraphQLIRVisitor.d.ts
n1ru4l: Alright!

PR #38550, opened 2019-09-23
add destroyed flag as of node 8 https://nodejs.org/api/stream.html#stream_writable_destroyed

ed-sparkes:
Adding destroyed flag to ReadableStream and WriteableStream as per the node documentation from v8 onwards. 

- [1 ] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/api/stream.html#stream_writable_destroyed

### DIFF
--- a/types/node/globals.d.ts
+++ b/types/node/globals.d.ts
@@ -750,6 +750,7 @@ declare namespace NodeJS {
         cork(): void;
         uncork(): void;
         destroy(error?: Error): void;
+        destroyed: boolean;
     }
     interface ReadStream extends Socket {
         readonly readableHighWaterMark: number;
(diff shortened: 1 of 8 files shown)


SimonSchick:
Please do not update EOL v7 and v9 typings.

trivikr (approved):

MichaelDeBoey (approved):

ExE-Boss (changes requested):
This needs rebasing.

Also, the main `types⁠/node⁠/base.d.ts` file is now for **Node v13**, so you need to do this for `types⁠/node⁠/v12⁠/base.d.ts` as well.

PR #38557, opened 2019-09-24
react-helmet: remove default export

dobesv:
This package does not have a default export.

```
> typeof require('react-helmet').Helmet
'function'
> typeof require('react-helmet').default
'undefined'
```

### DIFF
--- a/types/react-helmet-async/index.d.ts
+++ b/types/react-helmet-async/index.d.ts
@@ -2,7 +2,7 @@
 // Project: https://github.com/staylor/react-helmet-async#readme
 // Definitions by: forabi <https://github.com/forabi>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.6
+// TypeScript Version: 2.8
 
 import * as React from 'react';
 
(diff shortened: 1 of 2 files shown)


forabi (approved):

dobesv:
How do I fix:

```
Error: react-helmet-async depends on react but has a lower required TypeScript version.
    at checkTypeScriptVersions (/home/travis/build/DefinitelyTyped/DefinitelyTyped/node_modules/types-publisher/src/check-parse-results.ts:51:23)
    at Object.checkParseResults [as default] (/home/travis/build/DefinitelyTyped/DefinitelyTyped/node_modules/types-publisher/src/check-parse-results.ts:19:5)
    at <anonymous>
``` 

?

unindented:
@dobesv You shouldn't have changed the version of the type definition. It's only `react-helmet-async@1.0` that removed the default export. Fixing here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38700

dobesv:
Ah sorry I must have reverted the version number by accident, sorry about that.

antoinerousseau:
`package.json` has `"main": "./lib/Helmet.js",`
and `lib/Helmet.js` has `exports.default = HelmetExport;`
so there actually is a default export, although legacy.

PR #38571, opened 2019-09-24
[@types/ember] add ember__template

scottkidder:
Moved htmlSafe and isHTMLSafe into @ember/template where the official docs say they are.

If adding a new definition:


mike-north (approved):
LGTM

types/ember__template/index.d.ts
```diff
@@ -0,0 +1,9 @@
+// Type definitions for non-npm package @ember/template 3.0
+// Project: https://emberjs.com/api/ember/3.12/modules/@ember%2Ftemplate
+// Definitions by: Mike North <https://github.com/mike-north>
```
jamescdavis: @scottkidder, wanna put your info here?

types/ember__template/index.d.ts
```diff
@@ -0,0 +1,9 @@
...
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+import { SafeString } from '@ember/string/-private/handlebars';
```
jamescdavis: IMHO, we should move `@ember/string/-private/handlebars`  to *this* package (`@ember/template/-private/handlebars`).

jamescdavis (commented):
Looks pretty good, but I think we should move the `SafeString` type def here.

jamescdavis:
Also, see: https://github.com/typed-ember/ember-cli-typescript/issues/784#issuecomment-534840744

types/ember__template/index.d.ts
scottkidder: Ok, I'll go ahead and do that :+1:

scottkidder:
@typescript-bot Please don't merge this yet

scottkidder:
@jamescdavis I think I've properly moved `-template/handlebars.d-ts` into `@ember/template`. I'm not sure if the `@deprecated` comment on the exports in string are actually working.

types/ember__string/index.d.ts
```diff
@@ -1,18 +1,23 @@
...
 // TypeScript Version: 2.8
 
-import { SafeString } from "./-private/handlebars";
+import { SafeString } from "@ember/template/-private/handlebars";
+
+/** @deprecated use import from '@ember/template' instead. */
+export function htmlSafe(str: string): SafeString;
+
+/** @deprecated use import from '@ember/template' instead. */
+export function isHTMLSafe(str: any): str is SafeString;
```
jamescdavis: Instead of redefining these here, IMHO it's better to re-export them from `@ember/template`:
```ts
export { htmlSafe, isHTMLSafe } from '@ember/template';
```
That way we keep a single definition for these functions and you don't have to import `SafeString` from `@ember/template/-private` here. I'm not sure if the deprecated annotation will be picked up by tslint doing it this way, but I'd still add it for documentation purposes.

scottkidder: Yeah I've done that now. I agree, from what I could find I don't think `@deprecated` will picked up this way.

jamescdavis (approved):
Looks great!

types/ember/index.d.ts
```diff
@@ -48,7 +49,7 @@ import { TemplateFactory } from 'htmlbars-inline-precompile';
 import { Registry as ServiceRegistry } from '@ember/service';
 import { Registry as ControllerRegistry } from '@ember/controller';
 import * as EmberStringNs from '@ember/string';
-import * as EmberStringHandlebarsNs from '@ember/string/-private/handlebars';
+import * as EmberStringHandlebarsNs from '@ember/template/-private/handlebars';
```
mike-north:
```suggestion
import * as EmberTemplateHandlebarsNs from '@ember/template/-private/handlebars';
```

types/ember__string/index.d.ts
```diff
@@ -1,18 +1,18 @@
 // Type definitions for non-npm package @ember/string 3.0
 // Project: https://emberjs.com/api/ember/3.4/modules/@ember%2Fstring
 // Definitions by: Mike North <https://github.com/mike-north>
+//                 Scott Kidder <https://github.com/scottkidder>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
-import { SafeString } from "./-private/handlebars";
+/** @deprecated use imports from '@ember/template' instead. */
+export { htmlSafe, isHTMLSafe } from '@ember/template';
```
mike-north: I'm unclear here -- is this

(a) a bug fix in our types, where `import { htmlSafe } from '@ember/template';` no longer works entirely in Ember 3.x

or 

(b) A deprecation in Ember 3.x that currently works, but logs a deprecation warning?

mike-north (changes requested):

types/ember__template/index.d.ts
```diff
@@ -0,0 +1,10 @@
+// Type definitions for non-npm package @ember/template 3.0
+// Project: https://emberjs.com/api/ember/3.12/modules/@ember%2Ftemplate
+// Definitions by: Mike North <https://github.com/mike-north>
+//                 Scott Kidder <https://github.com/scottkidder>
```
mike-north: While anyone can add themselves to these headers in DT, you should be aware of what this means
- your own code changes will be merged in automatically _without a review_, and _in 24 hours, instead of the usual 5-7 days_
- your approvals to incoming PRs will cause them to automatically be merged, and then new releases of these packages will be immediately cut and released.

If things go wrong as a result of your code changes and/or approvals, are you willing to dig in on short notice and untangle things so we can unblock the community?

scottkidder: Hi Mike, given that I'm still novice with TypeScript I think it's better that I remove myself from the headers. Thanks for the information!

types/ember__string/index.d.ts
scottkidder: @mike-north as far as I can tell `import { htmlSafe } from '@ember/string';` still works and is not deprecated. However official [docs](http://api.emberjs.com/ember/3.13/functions/@ember%2Ftemplate/htmlSafe) say to import from `@ember/template`.

I imagine that the import from `@ember/string` should/will be deprecated at some point.

When I used `@deprecated` here I was thinking that I was deprecating the dts import itself. I'm happy to remove `@deprecated` to revisit at a later date since there is no underlying deprecation to match.

jamescdavis: Those were my findings as well. I guess remove `@deprecated`. Here's where they changed the docs: https://github.com/emberjs/ember.js/pull/16887 (not much info there)

types/ember__template/index.d.ts
jamescdavis: @mike-north should we add the Ember TS team so you're not solely responsible?

jamescdavis (changes requested):
In`types/ember/index.d.ts`under `namespace Handlebars` need to:
```diff
- class SafeString extends EmberStringHandlebarsNs.SafeString {}
+ class SafeString extends EmberTemplateHandlebarsNs.SafeString {}
```

types/ember__string/index.d.ts
jamescdavis: Here we go: https://github.com/emberjs/rfcs/blob/master/text/0236-deprecation-ember-string.md

mike-north: We need to align with Ember's public API. If it's not deprecated there, we should not indicate anything is deprecated here.

types/ember__template/index.d.ts
mike-north: @jamescdavis that sounds good to me. Mind opening a separate PR to do this?

scottkidder:
Thanks @jamescdavis! I apologize for the delay here..

types/ember__string/index.d.ts
scottkidder: Got it, makes sense. Thoughts on if it deserves another RFC to deprecate the exports from `@ember/string`?

jamescdavis: It's unclear whether
> 1. Move` htmlSafe` and `isHTMLSafe` to `@ember/template`

(from [the above RFC](https://github.com/emberjs/rfcs/blob/master/text/0236-deprecation-ember-string.md))
would include deprecating the exports from `@ember/string`. Regardless, the current state appears to be that importing from `@ember/string` not yet officially deprecated, just no longer being taught in the guides. Therefore, I agree with @mike-north that we shouldn't indicate deprecation here at this time.

types/ember__template/index.d.ts
jamescdavis: Happy to once these get merged.

jamescdavis (approved):
Looks great! Thanks @scottkidder!

mike-north (approved):
Looks great, thanks!

RyanCavanaugh:
Good to merge then?

mike-north:
> Good to merge then?

Yes

PR #38578, opened 2019-09-24
[react-router] update types for v5.1.0

1000hz:
Please fill in this template.

⚠️tests are currently broken on `master`
```console
Error: In package.json: Dependency vega-typings not in whitelist.
If you are depending on another `@types` package, do *not* add it to a `package.json`. Path mapping should make the import work.
If this is an external library that provides typings,  please make a pull request to types-publisher adding it to `dependenciesWhitelist.txt`.
    at checkPackageJsonDependencies (/Users/csaffary/src/github/DefinitelyTyped/node_modules/types-publisher/src/lib/definition-parser.ts:204:19)
    at combineDataForAllTypesVersions (/Users/csaffary/src/github/DefinitelyTyped/node_modules/types-publisher/src/lib/definition-parser.ts:99:37)
    at <anonymous>
Error: Parsing failed.
```
However, no type errors are present in VSCode for the tests I added.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <[Source](https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/hooks.js) | [Docs](https://github.com/ReactTraining/react-router/blob/3debcd4566bf3ee4c77b60d28f708bfd9222958c/packages/react-router/docs/api/hooks.md)>

### DIFF
--- a/types/react-router/index.d.ts
+++ b/types/react-router/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for React Router 5.0
+// Type definitions for React Router 5.1
 // Project: https://github.com/ReactTraining/react-router
 // Definitions by: Sergey Buturlakin <https://github.com/sergey-buturlakin>
 //                 Yuichi Murata <https://github.com/mrk21>
@@ -148,3 +148,13 @@ export function withRouter<P extends RouteComponentProps<any>, C extends React.C
 ): React.ComponentClass<Omit<P, keyof RouteComponentProps<any>> & WithRouterProps<C>> & WithRouterStatics<C>;
 
 export const __RouterContext: React.Context<RouteComponentProps>;
+
+export function useHistory(): H.History;
+
+export function useLocation(): H.Location;
+
+export function useParams<Params extends { [K in keyof Params]?: string } = {}>(): { [p in keyof Params]: string };
+
+export function useRouteMatch<Params extends { [K in keyof Params]?: string } = {}>(
+    path: string | RouteProps,
+): match<Params> | null;
--- a/types/react-router/index.d.ts
+++ b/types/react-router/index.d.ts
@@ -149,9 +149,9 @@ export function withRouter<P extends RouteComponentProps<any>, C extends React.C
 
 export const __RouterContext: React.Context<RouteComponentProps>;
 
-export function useHistory(): H.History;
+export function useHistory<HistoryLocationState = H.LocationState>(): H.History<HistoryLocationState>;
 
-export function useLocation(): H.Location;
+export function useLocation<S = H.LocationState>(): H.Location<S>;
 
 export function useParams<Params extends { [K in keyof Params]?: string } = {}>(): { [p in keyof Params]: string };
 
--- a/types/react-router-dom/index.d.ts
+++ b/types/react-router-dom/index.d.ts
@@ -50,11 +50,10 @@ export interface HashRouterProps {
 }
 export class HashRouter extends React.Component<HashRouterProps, any> {}
 
-export interface LinkProps<S = H.LocationState>
-  extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
-  to: H.LocationDescriptor<S>;
-  replace?: boolean;
-  innerRef?: React.Ref<HTMLAnchorElement>;
+export interface LinkProps<S = H.LocationState> extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+    to: H.LocationDescriptor<S> | ((location: H.Location<S>) => H.LocationDescriptor<S>);
+    replace?: boolean;
+    innerRef?: React.Ref<HTMLAnchorElement>;
 }
 export class Link<S = H.LocationState> extends React.Component<
   LinkProps<S>,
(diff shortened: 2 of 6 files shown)


types/react-router/index.d.ts
```diff
@@ -148,3 +148,13 @@ export function withRouter<P extends RouteComponentProps<any>, C extends React.C
 ): React.ComponentClass<Omit<P, keyof RouteComponentProps<any>> & WithRouterProps<C>> & WithRouterStatics<C>;
 
 export const __RouterContext: React.Context<RouteComponentProps>;
+
+export function useHistory(): H.History;
```
danielnixon: Can we include a type param for the history state `S` here?

types/react-router/index.d.ts
```diff
@@ -148,3 +148,13 @@ export function withRouter<P extends RouteComponentProps<any>, C extends React.C
...
+
+export function useLocation(): H.Location;
```
danielnixon: And here I think.

types/react-router/index.d.ts
1000hz: I went ahead and named the type param `HistoryLocationState` to match the definition for `H.History`

danielnixon: Nice

razh:
As of 5.1, `<Link>` also now supports passing functions to the `to` prop. This was not mentioned in the release notes:

https://github.com/ReactTraining/react-router/blob/master/packages/react-router-dom/docs/api/Link.md#to-function

1000hz:
@razh I saw your comment before your edit and added exactly the same code 🙂

1000hz:
The only remaining change I need to implement from the [changelog](https://github.com/ReactTraining/react-router/releases/tag/v5.1.0) is the one around supporting `forwardRef` in `<Link>` (and `<NavLink>`).

This... is tricky. Because `Link` is a generic component (it accepts generic parameter `S` for the location state of its `to` prop). Declaring a generic component concretely isn't possible yet (see https://github.com/microsoft/TypeScript/issues/17574) so we can either:
* unroll the typedef for `React.ForwardRefExoticComponent` to manually include the generic param
```tsx
export const Link: {
    <S = H.LocationState>(props: LinkProps<S>): React.ReactElement | null;
    readonly $$typeof: symbol;
    displayName: string;
    defaultProps?: Partial<LinkProps>;
};
```
* leave the typedef as-is and forgo correctness

Does anyone have any other ideas? Any preference?

eps1lon (approved):
> Does anyone have any other ideas? Any preference?

We shouldn't be too concerned with the component type exposed. The consumer should only care about that it's a component, the props and non-react statics. Everything else is an implementation detail unless specified in the docs.

It would be fine to use a function (which allows generic params) and then add the `ref` to the props.

The more complicated issue is that the type depends on the react version (<16.3 will not support forwarding refs). But I wouldn't bother too much with it.

1000hz:
I had a small issue when I tried to update to the new version of `@types/react-router-dom`. 
I tried to use the new hooks but the typedefs weren't being loaded. I looked in my `yarn.lock` and saw the old version of `@types/react-router` was still being used.

`@types/react-router-dom` and `@types/react-router-native` don't specify their dependency versions explicitly so they get published as `*`, which means people will need to either explicitly depend on `@types/react-router` in addition, or manually remove it from their lockfile and reinstall.

Seems like we should include a `package.json` in `@types/react-router-dom` and `@types/react-router-native` that specifies the versions of their dependents to prevent this in the future.

JulianG:
I just came across this 5 minutes ago.
```
➜ npm ls @types/react-router
level-manager-fe@0.1.0 /Users/julian/work/king/ctt/level-manager/ui
└─┬ @types/react-router-dom@5.1.0
  └── @types/react-router@5.0.5 
```

Iniside my node_modules I can see the package.json for @types/react-router-dom declares these:

```
    "dependencies": {
        "@types/history": "*",
        "@types/react": "*",
        "@types/react-router": "*"
    },
```

But I don't know where that package.json is on GitHub?

bennyn:
With @types/react-router-dom v5.1.0 I am getting the following errors:

**Log**

```bash
../../node_modules/@types/react-router-dom/index.d.ts:32:5 - error TS2305: Module '"./node_modules/@types/react-router"' has no exported member 'useHistory'.
../../node_modules/@types/react-router-dom/index.d.ts:33:5 - error TS2305: Module '"./node_modules/@types/react-router"' has no exported member 'useLocation'.
../../node_modules/@types/react-router-dom/index.d.ts:34:5 - error TS2305: Module '"./node_modules/@types/react-router"' has no exported member 'useParams'.
../../node_modules/@types/react-router-dom/index.d.ts:35:5 - error TS2305: Module '"./node_modules/@types/react-router"' has no exported member 'useRouteMatch'.
```

**Screenshot**

image

My dependencies (among others) are:

```json
"react": "16.9.0",
"react-dom": "16.9.0",
"react-router": "5.1.2",
"react-router-dom": "5.1.2",
"@types/react-router": "5.1.1",
"@types/react-router-dom": "5.1.0",
```

Am I holding something wrong?

eps1lon:
Bump your react-router declarations as well (in yarn just remove the lockfile entry for every `@types/react-router` and run `yarn` again).

bennyn:
I removed the "@types/react-router" and "@types/react-router-dom" definitions from my `yarn.lock` file and reran `yarn`. It did the job. Thanks, @eps1lon. 👍

PR #38585, opened 2019-09-25
fix: update type of eureka-js-client config obj

wayjam:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/jquatier/eureka-js-client/blob/master/src/EurekaClient.js>


jpsullivan (approved):
Sorry for not picking this up earlier. I believe this looks fine to me. `shouldUseDelta` at the root level seems like a no-brainer. The logger could be seen as instance-specific but I'm fine with it at the root level as well.

Might want to clarify that this could potentially cause breaking changes for users who are already configuring those fields.

PR #38591, opened 2019-09-25
[@types/simpl-schema] Add 'oneOf()' method to SimpleSchemaStatic.

nicu-chiciuc:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/aldeed/simple-schema-js#multiple-definitions-for-one-key

SimpleSchema has the static `oneOf()` method. The link points to the exact part about that method. The current type definition could be better, by having a common type with the constructor. But at least having the method defined will work, until the definition is more exhaustive.

### DIFF
--- a/types/simpl-schema/index.d.ts
+++ b/types/simpl-schema/index.d.ts
@@ -122,7 +122,7 @@ interface SimpleSchemaStatic {
   addValidator(validator: () => boolean): any;
   pick(...fields: string[]): SimpleSchemaStatic;
   omit(...fields: string[]): SimpleSchemaStatic;
-  oneOf(...types: Array<(SchemaDefinition | StringConstructor | StringConstructor | NumberConstructor | DateConstructor | ArrayConstructor)>): SimpleSchemaStatic;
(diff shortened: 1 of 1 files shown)


types/simpl-schema/index.d.ts
```diff
@@ -122,6 +122,7 @@ interface SimpleSchemaStatic {
   addValidator(validator: () => boolean): any;
   pick(...fields: string[]): SimpleSchemaStatic;
   omit(...fields: string[]): SimpleSchemaStatic;
+  oneOf(...types: Array<(SchemaDefinition | StringConstructor | StringConstructor | NumberConstructor | DateConstructor | ArrayConstructor)>): SimpleSchemaStatic;
```
mrcrane: StringConstructor is listed twice here. Maybe you meant to have BooleanConstructor instead?

mrcrane (changes requested):

types/simpl-schema/index.d.ts
nicu-chiciuc: My bad, I replaced the first one with BooleanConstructor.

mrcrane (approved):

PR #38352, opened 2019-09-13
test(react): Remove usage of React.Attributes

eps1lon:
Spotted this during review and it kind of threw me off. `React.Attributes` only includes `key` with which you'll never deal in props validation nor see in your component implementation.

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

PR #38337, opened 2019-09-12
[react] Add propTypes to ForwardRefExoticComponent

rpokorny:
- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).
- [x] Provide a URL to documentation or source code which provides context for the suggested changes
  - https://github.com/facebook/react/pull/12911
  - https://github.com/facebook/react/blob/v16.9.0/packages/react/src/__tests__/forwardRef-test.js#L77
- [ ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header. **(It doesn't)**
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.  **(N/A)**

PR #38603, opened 2019-09-25
react: Fix test failure from merge conflict.

sandersn:
A test written in #38337 on 12 Sep was merged on 24 Sep. In between in #38352, the Props definition it used changed so that the new test was an error.

This PR just clones the original Props type and gives it a new name. This fixes the error.

### DIFF
--- a/types/react/test/index.ts
+++ b/types/react/test/index.ts
@@ -402,7 +402,13 @@ const ForwardingRefComponent = React.forwardRef((props: {}, ref: React.Ref<RefCo
     return React.createElement(RefComponent, { ref });
 });
 
-const ForwardingRefComponentPropTypes: React.WeakValidationMap<Props> = {};
+interface AttributeProps extends React.Attributes {
+    hello: string;
+    world?: string | null;
+    foo: number;
+}
+
+const ForwardingRefComponentPropTypes: React.WeakValidationMap<AttributeProps> = {};
 ForwardingRefComponent.propTypes = ForwardingRefComponentPropTypes;
 
 function RefCarryingComponent() {


johnnyreilly (approved):

types/react/test/index.ts
```diff
@@ -402,7 +402,13 @@ const ForwardingRefComponent = React.forwardRef((props: {}, ref: React.Ref<RefCo
     return React.createElement(RefComponent, { ref });
 });
 
-const ForwardingRefComponentPropTypes: React.WeakValidationMap<Props> = {};
+interface AttributeProps extends React.Attributes {
```
eps1lon:
```suggestion
interface AttributeProps {
```

`React.Attributes` is not meant to be used that way since it includes props that you never see inside your component or need to validate with `propTypes`. It's technically only a private type.

eps1lon (changes requested):

types/react/test/index.ts
sandersn: The test is broken without `extends React.Attributes` -- `WeakValidationMap<AttributeProps>` doesn't have any properties in common with `WeakValidationMap<RefAttributes<RefComponent>>` from `ForwardRefExoticComponent<RefAttributes<RefComponent>>`

Does that mean the test is broken or the new propTypes property in ForwardRefExoticComponent is wrong? I suppose I could say

```ts
const ForwardingRefComponentPropTypes: React.WeakValidationMap<React.RefAttributes<RefComponent>> = {};
```

in the test instead.

sandersn: I don't know React that well, I'm just trying to get it unblocked for other contributors, so any advice you have is helpful.

eps1lon: Well lets merge this if this block other work. I'll take  look later why we need this.

sandersn: Thanks. I merged it.

@rpokorny Maybe you can take another look too?

PR #38616, opened 2019-09-26
[jsonwebtoken] change verify to return unknown

lautarodragan:
WIP.

[The jsonwebtoken docs](https://github.com/auth0/node-jsonwebtoken#jwtverifytoken-secretorpublickey-options-callback) aren't super clear on this, but `verify` and `decode` [return the payload passed by the user](https://github.com/auth0/node-jsonwebtoken/blob/v8.3.0/verify.js#L196) to `sign` in the first place, which can't be known by the library.

A type guard by the user of the library is always necessary.

Reminder to self: do [this](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#i-want-to-use-features-from-typescript-31-or-above)

### DIFF
--- a/types/jsonwebtoken/index.d.ts
+++ b/types/jsonwebtoken/index.d.ts
@@ -89,7 +89,7 @@ export type VerifyErrors =
     | TokenExpiredError;
 export type VerifyCallback = (
     err: VerifyErrors,
-    decoded: object | string,
+    decoded: unknown,
 ) => void;
 
 export type SignCallback = (
@@ -163,7 +163,7 @@ export function verify(
     token: string,
     secretOrPublicKey: string | Buffer,
     options?: VerifyOptions,
-): object | string;
+): unknown;
 
 /**
  * Asynchronously verify given token using a secret or a public key to get a decoded token
--- a/types/jsonwebtoken/index.d.ts
+++ b/types/jsonwebtoken/index.d.ts
@@ -6,7 +6,8 @@
 //                 Veli-Pekka Kestilä <https://github.com/vpk>,
 //                 Daniel Parker <https://github.com/rlgod>,
 //                 Kjell Dießel <https://github.com/kettil>,
-//                 Robert Gajda <https://github.com/RunAge>
+//                 Robert Gajda <https://github.com/RunAge>,
+//                 Lautaro Dragan <https://github.com/lautarodragan>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
--- a/types/jsonwebtoken/index.d.ts
+++ b/types/jsonwebtoken/index.d.ts
@@ -8,7 +8,7 @@
 //                 Kjell Dießel <https://github.com/kettil>,
 //                 Robert Gajda <https://github.com/RunAge>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.2
+// TypeScript Version: 3.0
 
 /// <reference types="node" />
 
--- a/types/jsonwebtoken/jsonwebtoken-tests.ts
+++ b/types/jsonwebtoken/jsonwebtoken-tests.ts
@@ -14,6 +14,8 @@ interface TestObject {
     foo: string;
 }
 
+const isTestObject = (x: any): x is TestObject => x && typeof x.foo === 'string'
+
 const testObject = { foo: "bar" };
 
 /**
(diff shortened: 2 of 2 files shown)


lautarodragan:
CI is failing with this error:

```
Testing 1 changed packages: jsonwebtoken

Testing 5 dependent packages: jwt-express,passport-jwt,sc-auth,scc-broker-client,socketcluster-server

Error: jwt-express depends on jsonwebtoken but has a lower required TypeScript version.
    at checkTypeScriptVersions (/home/travis/build/DefinitelyTyped/DefinitelyTyped/node_modules/types-publisher/src/check-parse-results.ts:51:23)
    at Object.checkParseResults [as default] (/home/travis/build/DefinitelyTyped/DefinitelyTyped/node_modules/types-publisher/src/check-parse-results.ts:19:5)
    at <anonymous>
```

TODO: try [this](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package)

rlgod:
I'm not totally sure what you mean by 'return the payload passed by the user to sign in the first place'. Are you able to elaborate on this a bit?

I'm concerned by introducing the `unknown` type in this case because people will have written code expecting the types previously specified and their code would just stop working with this change (I imagine) without really gaining them much. In the async case, they should be checking the err argument before handling the decoded argument to their callback, and in the sync case the functions will throw an error anyway so for all intents and purposes they can expect that return type to be `object | string`. 

If I'm missing something please feel free to explain it a bit further so I can understand.

lautarodragan:
> I'm not totally sure what you mean by 'return the payload passed by the user to sign in the first place'. Are you able to elaborate on this a bit?

@rlgod totally. It's a bit hard for people who have been using type casts or the any type to grasp the value `unknown` provides.

I think the changes to the tests already show something: the type cast is no longer needed. When using that type cast, type information is basically lost. You're basically telling the compiler "trust me, I know this is of this type". 

In truth, the returned value by `verify` isn't a `string | object`, the `verify` method can't guarantee it since it doesn't know what's inside the JWT it's decoding. It could be anything. In that sense, `any` would have been more accurate. But `any` could fail at run time. `unknown` is the type-safe counterpart of any.

Here's a working example: https://repl.it/repls/HeavenlyRuddyComputers.

> I'm concerned by introducing the `unknown` type in this case because people will have written code expecting the types previously specified and their code would just stop working with this change (I imagine) without really gaining them much. 

It's possible, but in practice it's most likely nobody expects a `string | object`, and everybody is doing a type cast, as in the tests. In this case, nothing will change for them.

This change actually gains them type safety, I wouldn't say that isn't much.

> In the async case, they should be checking the err argument before handling the decoded argument to their callback, and in the sync case the functions will throw an error anyway so for all intents and purposes they can expect that return type to be `object | string`.

I'm confused by this part, I'm not following how this is related to the `unknown` change...

rlgod:
Ok I see what you're getting at. Thanks for clarifying with that example.

I've done some of my own testing too and I think that this is a good change and a good use of `unknown`.

It may be worth looking through the rest of the types in this project to see if there are additional places where this change makes sense. e.g. `decode`.

Am I right in thinking that people would need to modify their code to add explicit type checking after this change? e.g.
```
if (decoded instanceof object) {
    // Do work
}
```

lautarodragan:
Awesome! Thanks for the feedback.

I'm not sure but I think that, if people upgrade to new type definitions that use `unknown`, 
- To take advantage of the `unknown`, if they had been using a type cast, they'll need to modify the code to use a type guard. `instanceof x`, as you mention, could be one type guard, but I believe `typeof x === 'whatever'` may be better for primitives, and functions that return `x is y` for anything else, as `instanceof` has several caveats which may be particularly problematic for deseralized objects, see [`instanceof` lies](https://medium.com/javascript-scene/javascript-factory-functions-vs-constructor-functions-vs-classes-2f22ceddf33e) for example.
- Users could also upgrade the type definitions but keep their code the same if they were using type casts, in this case nothing should change for them. I've upgraded the test cases to not use type casts, but I'll add one back to exemplify this.
- Finally, in the case a user was actually expecting `verify` to return a `string | object`, their code will break if they don't update it. But this is unlikely, since it's impractical for the payload to be a `string`, and a plain `object` type doesn't really give any type information, you wouldn't even be able to access the properties the library itself adds to the payload, such as the `payload.iat`.

rlgod:
@lautarodragan I'm happy with the change. You'll have to make sure the build succeeds. I'm also thinking that I might leave this for one of the core DefinitelyTyped team to review because despite being happy with your changes, I'm not 100% confident on correct usage of `unknown` and would feel more comfortable with one of the core maintainers looking at it.

lautarodragan:
Totally! Thanks @rlgod for your input. I'll try and get the build working and everything sorted out before asking for anyone's feedback so I don't waste anyone's time.

lautarodragan:
Reminder to self: do [this](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#i-want-to-use-features-from-typescript-31-or-above)

PR #38620, opened 2019-09-26
prosemirror-view: make decorations generic [suggestion]

ifiokjr:
I thought it might be a nice addition to make the decoration spec generic. I've found myself working quite closely with decorations lately and this addition would allow workflows like the following. 

```ts
let decorationSet: DecorationSet; // Defined somewhere in the code.

const isSpecialDecoration = (
  decoration: Decoration,
): decoration is Decoration<{ type: 'special'; value: number }> => 
  decoration.spec.type === 'special';

const decorations = decorationSet.find(); // Get all decorations
const filtered = decorations.filter(isSpecialDecoration);

type A = typeof filtered[number]; // => Decoration<{ type: 'special', value: number }>
```

Please fill in this template.

### DIFF
--- a/types/prosemirror-view/index.d.ts
+++ b/types/prosemirror-view/index.d.ts
@@ -21,11 +21,72 @@ import { EditorState, Selection, Transaction } from 'prosemirror-state';
 import { Mapping } from 'prosemirror-transform';
 
 /**
+ * The `spec` for a widget decoration
+ */
+export interface WidgetDecorationSpec {
+  /**
+   * Controls which side of the document position this widget is
+   * associated with. When negative, it is drawn before a cursor
+   * at its position, and content inserted at that position ends
+   * up after the widget. When zero (the default) or positive, the
+   * widget is drawn after the cursor and content inserted there
+   * ends up before the widget.
+   *
+   * When there are multiple widgets at a given position, their
+   * `side` values determine the order in which they appear. Those
+   * with lower values appear first. The ordering of widgets with
+   * the same `side` value is unspecified.
+   *
+   * When `marks` is null, `side` also determines the marks that
+   * the widget is wrapped in—those of the node before when
+   * negative, those of the node after when positive.
+   */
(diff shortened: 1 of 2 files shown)


ifiokjr:
Don’t merge yet. I’d like the feedback.

mrcrane (changes requested):
Ok, I will mark this as changes requested for now and I will approve when you get some feedback. Thanks!

ifiokjr:
@timjb @patsimm @davidka @bradleyayers any thoughts on this addition?

I'd rather not just merge it without any thoughts. I've been using it for a few days in my own code and it's made working with decorations more intuitive.

types/prosemirror-view/index.d.ts
```diff
@@ -48,87 +109,32 @@ export class Decoration {
    * also directly pass a DOM node. getPos can be used to find the
    * widget's current document position.
    */
-  static widget(
+  static widget<T extends {} = { [key: string]: any }>(
```
timjb: Use `static widget<T extends object = { [key: string]: any }>(` instead. The `extends object` makes sure that one gets a type error if for example one tried to use a string as the `spec` argument:

```typescript
Decoration.widget(0, (x) => x.dom, 'this argument should be an object, not a string!'); // $ExpectError
```

(Same for other functions.)

timjb (changes requested):
One small detail, looks good otherwise!

types/prosemirror-view/index.d.ts
ifiokjr: Thanks @timjb! I never knew this. I've just updated several other projects with this enhancement.

ifiokjr:
> One small detail, looks good otherwise!

Thanks @timjb, I've addressed the request. 

@mrcrane I'm happy to merge the changes now.

timjb (approved):
LGTM!

mrcrane (approved):

PR #38369, opened 2019-09-13
chore(hapi): update hapi ecosystem to ts 3.x

SimonSchick:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/hapijs/hoek/blob/master/API.md https://github.com/hapijs/wreck
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

This touched a few packages across the @hapi namespace to improve type safety.
Core changes are in hapi and hoek.

PR #38623, opened 2019-09-26
[hapi__hapi] Improve types for RequestAuth

p-kuen:
artifacts and credentials are null if the user is not authenticated.
Also the credentials could be any value, therefore I added the `extends Record<string, any>` to AuthCredentials.
This should be considered in the types correctly.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/hapijs/hapi/blob/master/lib/auth.js

### DIFF
--- a/types/hapi__hapi/index.d.ts
+++ b/types/hapi__hapi/index.d.ts
@@ -184,7 +184,7 @@ export interface AppCredentials {
 /**
  * User-extensible type for request.auth credentials.
  */
-export interface AuthCredentials {
+export interface AuthCredentials extends Record<string, any> {
     /**
      * The application scopes to be granted.
      * [See docs](https://github.com/hapijs/hapi/blob/master/API.md#-routeoptionsauthaccessscope)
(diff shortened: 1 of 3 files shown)


saboya (changes requested):
You're supposed to use module augmentation to enhance these types as needed.

Patcher56:
@saboya Oh, thank you. But still it is not possible to set `artifacts` via module augmentation.
The Record<string, any> would fix that.

My code:
```
declare module '@hapi/hapi' {
  interface RequestAuth {
    artifacts: {
      id: number
    }
  }

  interface AuthCredentials {
    id: number
    name: string
  }
}
```
I get the error that `artifacts` from the dt types are not from the same type as my types. The `object` keyword should be the problem here.

saboya:
Maybe those types do need a `null` in there, but nowhere in the documentation it indicates that they should be a `Record`.

Maybe we should focus our efforts in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38369. `artifacts` might be a fit for `unknown`

Patcher56:
Alright. Closing this as the other PR is addressing these issues.

PR #38624, opened 2019-09-26
[react-router-dom] Adding package.json to specify @types/react-router: ^5.1.0 dependency

JulianG:
Adding package.json to specify dependencies.
In particular "@types/react-router": "^5.1.0"

see: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38578#issuecomment-535006348
> "Seems like we should include a package.json in @types/react-router-dom and @types/react-router-native that specifies the versions of their dependencies to prevent this in the future."

also: https://github.com/DefinitelyTyped/DefinitelyTyped#i-notice-some-packages-having-a-packagejson-here

> #### I notice some packages having a package.json here.
> Usually you won't need this. When publishing a package we will normally automatically create a package.json for it. A package.json may be included for the sake of specifying dependencies. Here's an example. We do not allow other fields, such as "description", to be defined manually. Also, if you need to reference an older version of typings, you must do that by adding "dependencies": { "@types/foo": "x.y.z" } to the package.json.

### DIFF
--- a/types/react-router-dom/package.json
+++ b/types/react-router-dom/package.json
@@ -1,0 +1,8 @@
+{
+  "private": true,
+  "dependencies": {
+    "@types/history": "*",
+    "@types/react": "*",
+    "@types/react-router": "^5.1.0"
+  }
+}


JulianG:
I'm not sure what to do with this error message:
```
```

The reason we need a package.json is to be able to force the version of @types/react-router to be `^5.1`. The generated package.json would list the dependency version as "*".
This is not enough. See: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38578#issuecomment-535006348

Is there another way to specify dependencies versions?

1000hz:
This seems like the right thing to do, as recommended in the FAQ. @sandersn any insight as to why this is failing?

types/react-router-dom/package.json
```diff
@@ -0,0 +1,8 @@
+{
+  "private": true,
+  "dependencies": {
+    "@types/history": "*",
```
1000hz: `@types/history` needs to be pinned to `^4.7.2` as well because `History.Location` was not a generic type in older versions.

1000hz (changes requested):
we should add the same `package.json` under `types/react-router-native` as well

types/react-router-dom/package.json
sandersn: DT implicitly uses the newest version of other DT packages.

sandersn (changes requested):
This change is not needed. Definitely Typed implicitly adds the newest version of other DT packages to the published package.json.
You can add explicit redirects to older versions using the `"paths"` entry in tsconfig, but this package uses the current version of all its dependencies.

1000hz:
@sandersn Hmm, that doesn't seem to be the case. See the published dependencies here:
https://unpkg.com/browse/@types/react-router-dom@5.1.0/package.json

Since they're listed as `*`, anyone upgrading to a new version of this package will not pull in the newer dependency versions because the old versions in their lockfile satisfy this loose semver.

sandersn:
Ah, I see the problem. Can you open an issue on microsoft/types-publisher? That's the code that generates package.json (among other things). We should solve this for all packages, not just react-router-dom.

PR #38610, opened 2019-09-25
[Jasmine] Add parameter to SpyObjMethodNames

chivesrs:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: N/A, this just matches the paramterized types of `createSpyObj` with the functions right above it.
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

PR #38629, opened 2019-09-26
Add propertyNames support for Jasmine 3.5

scarletcs:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://jasmine.github.io/api/3.5/jasmine.html#.createSpyObj

---

I can't figure out how to run tests for these definitions right now, sorry. I'm actually concerned the current tests in the Jasmine root might be out of date since Jasmine 3.0 — but I'm not sure how to verify it.

These changes should correspond to being `@types/jasmine` version 3.5.0, but I'm not sure if what I've done is sufficient to mark that, or if that's even something I do myself.


kolodny:
Is this overlapping with #38610?

chivesrs:
You should update the file in `ts3.1` directory as well.

spacemonaut:
@kolodny That pull request does other, related, but non-overlapping/conflicting work.

@chivesrs Thank you, that's been updated.

jh-pq:
Oh, I just saw — propertyNames should be added to that other PR!

chivesrs:
Do you need to rebase to pick up #38412? I guess GitHub isn't complaining about a merge conflict so probably not.

You can ignore #38610 entirely, it's been closed (as the former PR contains the latter's changes).

types/jasmine/index.d.ts
```diff
@@ -247,11 +248,11 @@ declare namespace jasmine {
     function objectContaining<T>(sample: Partial<T>): ObjectContaining<T>;
     function createSpy(name?: string, originalFn?: Function): Spy;
 
-    function createSpyObj(baseName: string, methodNames: SpyObjMethodNames): any;
-    function createSpyObj<T>(baseName: string, methodNames: SpyObjMethodNames<T>): SpyObj<T>;
+    function createSpyObj(baseName: string, methodNames: SpyObjMethodNames, propertyNames?: SpyObjMethodNames): any;
+    function createSpyObj<T>(baseName: string, methodNames: SpyObjMethodNames<T>, propertyNames?: SpyObjMethodNames<T>): SpyObj<T>;
```
zvirja: It's incorrect to use `SpyObjMethodNames<T>` here, as it will kill types for the properties. Instead, you should create a new type for this purpose:

```ts
    type SpyObjPropertyNames<T = unknown> =
        unknown extends T ?
            (ReadonlyArray<string> | { [propertyName: string]: any }) :
            (ReadonlyArray<keyof T> | { [ P in keyof T ]?: T[P] } )
```

types/jasmine/index.d.ts
```diff
@@ -247,11 +248,11 @@ declare namespace jasmine {
...
 
-    function createSpyObj(methodNames: SpyObjMethodNames): any;
-    function createSpyObj<T>(methodNames: SpyObjMethodNames<T>): SpyObj<T>;
+    function createSpyObj(methodNames: SpyObjMethodNames, propertyNames?: SpyObjMethodNames): any;
```
zvirja: Instead of making it optional, I'd suggest to create a new overload and mark it with `@since 3.5.0` attribute. E.g.:

```ts
    /**
     * @since 3.5.0
     */
    function createSpyObj<T>(baseName: string, methodNames: SpyObjMethodNames<T>, propertyNames: SpyObjPropertyNames<T>): SpyObj<T>;
```

Do the same for all the methods.

types/jasmine/jasmine-tests.ts
```diff
@@ -1443,6 +1443,15 @@ describe("createSpyObj", function() {
         expect(spyObj.method2.and.identity()).toEqual('BaseName.method2');
     });
 
+    it("should allow you to provide properties with values to jasmine.createSpyObj()", function() {
```
zvirja: Notice, this code is never executed and it's sole purpose it just to describe types.. Therefore, it would be more clear to write scenarios from this perspective, as it allows to omit some parts (e.g. assertion part is not necessary). Please cover all 4 creation methods you have just introduced. Also ensure to have test cases with non-existing property names, so it errors in that line. For both, array and object syntax.

zvirja (changes requested):
Please take a look at the comments I left.

Please also notice, that there are two test files - one immediately under `jasmine` folder and second under `jasmine/3.1`. Please ensure to keep those files in sync, so they are the same and only differ in places where behavior is different (e.g. errors for higher version, while allows for lower).

Thanks for giving it a go! 👍

types/jasmine/index.d.ts
```diff
@@ -15,6 +15,7 @@
 //                 Stephen Farrar <https://github.com/stephenfarrar>
 //                 Mochamad Arfin <https://github.com/ndunks>
 //                 Alex Povar <https://github.com/zvirja>
+//                 Spacemonaut <https://github.com/spacemonaut>
```
zvirja: Notice, you should add your name here only if you want to become a maintainer of this project and always participate in future project's life. You will be notified each time a PR is created about jasmine, so it's quite a lot.

Otherwise, you could make changes without adding your name here, as it's not mandatory 😉

PR #38635, opened 2019-09-26
Loosen type signature of treeify method asTree

carnesen:
Please fill in this template.

- [x] Add or edit tests to reflect the change. (Note: `npm test` is failing but it does not seem related to any changed I made here)

https://github.com/notatestuser/treeify/blob/master/treeify.js#L64
^^ Where showValues gets used (checks for truthiness only)
https://github.com/notatestuser/treeify/blob/master/treeify.js#L36
^^ Where hideFunctions gets used (checks for truthiness only)

### DIFF
--- a/types/treeify/index.d.ts
+++ b/types/treeify/index.d.ts
@@ -1,27 +1,23 @@
-// Type definitions for treeify 1.0
+// Type definitions for treeify 1.1
 // Project: https://github.com/notatestuser/treeify
 // Definitions by: Mike North <https://github.com/mike-north>
+//                 Chris Arnesen <https://github.com/carnesen>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.1
-
-export interface TreeObject {
-    [k: string]: TreeValue;
-}
-export type TreeValue = string | TreeObject;
+// TypeScript Version: 2.2
 
 export function asTree(
-    treeObj: TreeObject,
-    showValues: boolean,
-    hideFunctions: boolean
+    obj: object,
+    showValues?: boolean,
+    hideFunctions?: boolean
 ): string;
 
 export function asLines(
-    treeObj: TreeObject,
+    obj: object,
     showValues: boolean,
     lineCallback: (line: string) => void
 ): string;
 export function asLines(
-    treeObj: TreeObject,
+    obj: object,
     showValues: boolean,
     hideFunctions: boolean,
     lineCallback: (line: string) => void
--- a/types/treeify/index.d.ts
+++ b/types/treeify/index.d.ts
@@ -3,22 +3,22 @@
 // Definitions by: Mike North <https://github.com/mike-north>
 //                 Chris Arnesen <https://github.com/carnesen>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.2
+// TypeScript Version: 2.1
 
 export function asTree(
-    obj: object,
+    obj: any,
     showValues?: boolean,
     hideFunctions?: boolean
-): string;
+): string | undefined;
 
 export function asLines(
-    obj: object,
+    obj: any,
     showValues: boolean,
     lineCallback: (line: string) => void
-): string;
+): string | undefined;
 export function asLines(
-    obj: object,
+    obj: any,
     showValues: boolean,
     hideFunctions: boolean,
     lineCallback: (line: string) => void
-): string;
+): string | undefined;
--- a/types/treeify/treeify-tests.ts
+++ b/types/treeify/treeify-tests.ts
@@ -2,18 +2,21 @@ import * as treeify from 'treeify';
 
 function log(s: string): void {}
 
+// $ExpectType string
 treeify.asTree([0, { foo: 'bar' }, 2], true);
 // ├─ 0: 0
 // ├─ 1
 // │  └─ foo: bar
 // └─ 2: 2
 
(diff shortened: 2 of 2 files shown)


mike-north (changes requested):
I think this lets through some cases that I would not want to type-check 

example: 

```ts
asTree({ notATree() {} }, true, false);
asTree({ a: undefined }, true, false);
asTree(Promise, true, false);
```

Do you agree that we should see compile errors for the above cases?

carnesen:
@mike-north These two are perfectly legal invocations of the function with sensible results:
```
asTree({ notATree() {} }, true, false);
//└─ notATree: notATree() { }
asTree({ a: undefined }, true, false);
//└─ a: undefined
```
As such they should be allowed by the typings. In general my philosophy is that the types should describe as accurately as possible how the module actually behaves.

As for `asTree(Promise, true, false)`, that returns `undefined` since `typeof Promise` is `'function'`. Probably we should handle that case separately as an overload. Would you be ok with:
```
export function asTree(obj: (...args: any[]) => any): undefined;
export function asTree(obj: object, showValues?: boolean, hideFunctions?: boolean): string;
```
?

That wouldn't actually handle the pretty degenerate case of `asTree(Promise, true, false)` but it would handle the more plausible case of someone passing in a function as the `obj` argument.

mike-north:
I think we should be able to do this via something like the following 

https://www.typescriptlang.org/play/index.html#code/JYOwLgpgTgZghgYwgAgCpQhA8gIwFYQJjIDeAUMpcgNoDWAXMgM5hSgDmAuo+pgGpwANgFcUAH2QB7fITBkAvmVCRYiFLwgBBKFDgBPZBAAekEABMmybbr0AeDQJHipMogD5SigJBljAB0koYjA9P3UMCEdRZABeZlYOZAkNXAIiJLQI630AbjIyGGEQImBJEGQ4Jg0ACgoqVkxUngjU2QyNbL0AGi8qZDrKAHpB5AbsfEZpNLAevoHmAAtJAHcoiCZGHElJQQg4EB75heAzCAAxIpKyjeQtnb2QMgBKRhY2EHZSeYwwYShygDkAIU+UqNRIFUYAJwAPkXVGUFE8PggiYECeeTBEWq1GhAM48NYSOQKLRGLIWMw1RIcEY1SesQ8QLhCOJpPRmKq2Oo9MZyCBBNZEGRQjJnJqIGEgkEhMRwpJoo5FK5VKKpxgoAgZllbMV5KAA

I'll concede that functions are allowed as values, but I believe the whole point of this library is to pass in something like an array or object as the "root" of a hierarchy. We should represent this accurately via types

carnesen:
@mike-north I've just pushed a commit that actually loosens the type definition further to allow `any` input with the return type being `string | undefined`. I don't think the user benefits by imposing restrictions on the input type here given the underlying behavior of the module. If you disagree could you describe such a scenario where the types you propose are saving the user from "shooting themselves in the foot" as the saying goes?

carnesen:
@mike-north Have you had a chance to consider the implications of the looser types proposed in this PR that more closely reflect the behavior of the implementation?

types/treeify/index.d.ts
```diff
@@ -1,28 +1,24 @@
-// Type definitions for treeify 1.0
+// Type definitions for treeify 1.1
 // Project: https://github.com/notatestuser/treeify
 // Definitions by: Mike North <https://github.com/mike-north>
+//                 Chris Arnesen <https://github.com/carnesen>
```
mike-north: As a co-maintainer of these types with me, are you ok with the idea of using the documentation and test-suite as the sole source of truth for the library's public API?

mike-north:
RE: the cases of `asTree(null)` and `asTree(undefined)`, I found some test cases that indicate the author isn't covering these scenarios by accident

https://github.com/notatestuser/treeify/blob/3855e5d6e3f6b7236b1b41d78ecde4a9cb6bb1c7/test/tree-test.js#L259-L260

types/treeify/index.d.ts
carnesen: I'm not sure I understand the question, but I'd argue that the library's source code is the ultimate source of truth for its public API. Imagine for example that the author of the library converted the source code to TypeScript and ran `tsc` with `declarations` set to true. What would the `.d.ts` files look like? That is how I tend to approach `@types`. In this case, because of this line:

https://github.com/notatestuser/treeify/blob/master/treeify.js#L71

sometimes asTree can return `undefined` therefore the type should include `undefined` in its ReturnType.

mike-north: In this case, I can see evidence in the test suite that returning `undefined` is what the author intends. However, when maintaining these `@types/*` packages, it's often not quite so clear. Adding your name to the list of maintainers means you will get your changes merged in w/o a review, so i want to ensure we're alignment beyond this one code change.

If I had the following code
```js
function combine(a, b) {  return a + b; }
```
and my readme consisted of 
```js
add(3, 4); // 7
```
it sounds like you'd want to add types that support things like 
```js
add('foo ', 'bar'); // "foo bar"
add('foo ', 6); // "foo 6"
add(6, ' bar'); // "6 bar"
```
Am I misrepresenting your view here?

carnesen: Yeah I suppose I would! Absent other information I'd say concatenation is a reasonable use of the library that you might expect to see "in the wild". That's my other consideration is thinking about the person who is using the library in JavaScript and now decides to upgrade to TypeScript. Do the types cover their reasonable use case?

PR #38639, opened 2019-09-26
[react-native-vector-icons] fix: Icon style

Naturalclar:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/oblador/react-native-vector-icons/blob/master/lib/create-icon-set.js

`create-icon-set.js` in `react-native-vector-icons` suggests that `Icon` Component is a `Text` Component, so the style should expect a `TextStyle` (With current definition, providing `color` results in TS error, because `color` is not part of `ViewStyle`).
It's true that default prop has a `marginRight: 10` , which is an invalid style for `TextStyle`, and I believe that is an error in `react-native-vector-icon`

### DIFF
--- a/types/react-native-vector-icons/Icon.d.ts
+++ b/types/react-native-vector-icons/Icon.d.ts
@@ -55,7 +55,7 @@ export interface IconButtonProps extends IconProps, TouchableHighlightProps, Tou
    *
    * @default {marginRight: 10}
    */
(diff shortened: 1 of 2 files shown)


iRoachie (approved):
Looks good, maybe we can just remove the default line then

PR #38641, opened 2019-09-26
console-ui now uses Readable/WritableStream from node

sandersn:
Previously, it used process.Readable/Writeable. But its tests show that the constructor is meant to take process.stdin, which is a ReadStream. Until
recently, the two types were mistakenly assignable. Now that they are not, I think the type should be ReadableStream, which is a common
supertype of process.Readable and ReadStream.


dfreeman (approved):
That sounds about right. Thanks!

PR #38648, opened 2019-09-27
[sortablejs] update `onMove()` to 2 arguments & add new properties in `Options`

waynevanson:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/SortableJS/Sortable#options>

### DIFF
--- a/types/sortablejs/index.d.ts
+++ b/types/sortablejs/index.d.ts
@@ -27,7 +27,16 @@ declare class Sortable {
      * @param options Sortable options object.
      */
     static create(element: HTMLElement, options: Sortable.Options): Sortable;
-
+    
+    /**
+     * Mounts a plugin to the `Sortable`
+     * @param plugins One or more of the sortable plugins. 
+     */
+    // todo:
+    // + tests
+    // + declarations for the plugins. Plugins should extend a `Sortable.Plugin` class.
+    static mount(...plugins: any[]): void;
+    
     /**
      * Options getter/setter
      * @param name a Sortable.Options property.
@@ -138,6 +147,15 @@ declare namespace Sortable {
          */
         delay?: number;
         /**
+         * Only delay if user is using touch
+         */
+        delayOnTouchOnly?: boolean;
+        /**
+         * Direction of Sortable
+         * (will be detected automatically if not given)
+         */
+        direction?: 'horizontal' | 'vertical';
+        /**
          * Disables the sortable if set to true.
          */
         disabled?: boolean;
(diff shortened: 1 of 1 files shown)


types/sortablejs/index.d.ts
```diff
@@ -28,6 +29,15 @@ declare class Sortable {
...
     static create(element: HTMLElement, options: Sortable.Options): Sortable;
 
+    /**
+     * Mounts a plugin to the `Sortable`
+     * @param plugins One or more of the sortable plugins.
+     */
+    // todo:
+    // + tests
+    // + declarations for the plugins. Plugins should extend a `Sortable.Plugin` class.
+    static mount(...plugins: any[]): void;
```
MaartenStaa: It seems like a missed opportunity to not also add types now for the plugins. This is better than nothing, but still.

types/sortablejs/index.d.ts
```diff
@@ -137,6 +147,15 @@ declare namespace Sortable {
...
         delay?: number;
+        /**
+         * Only delay if user is using touch
+         */
+        delayOnTouchOnly?: boolean;
+        /**
+         * Direction of Sortable
+         * (will be detected automatically if not given)
+         */
+        direction?: 'horizontal' | 'vertical';
```
MaartenStaa: This isn't right, `direction` can also be a function. Example in the docs:

```javascript
Sortable.create(el, {
	direction: function(evt, target, dragEl) {
		if (target !== null && target.className.includes('half-column') && dragEl.className.includes('half-column')) {
			return 'horizontal';
		}
		return 'vertical';
	}
});
```

types/sortablejs/index.d.ts
```diff
@@ -255,7 +332,7 @@ declare namespace Sortable {
         /**
          * Event when you move an item in the list or between lists
          */
-        onMove?: (event: MoveEvent) => boolean;
+        onMove?: (evt: MoveEvent, originalEvent: Event) => void;
```
MaartenStaa: This return type shouldn't be `void`. The docs say:

```javascript
		// return false; — for cancel
		// return -1; — insert before target
		// return 1; — insert after target
```

Though, reading the actual code, it seems returning `true` would also work. So maybe `=> boolean | -1 | 1`?

MaartenStaa (changes requested):

types/sortablejs/index.d.ts
waynevanson: It does indeed. I would love to but I'm not too familiar with all the plugins.

I plan to add more to this at a later date in another pull.

types/sortablejs/index.d.ts
waynevanson: Ok, I will add this in.

types/sortablejs/index.d.ts
waynevanson: Yes, that is correct.

I will change this. That was a mistake on my part.

waynevanson:
I'm actually going to close this because I just rewrote a lot of new changes, including all the plugins.

I changed all that was requested.

PR #38651, opened 2019-09-27
Update Model.bulkCreate records type

hassanhe:
It should be like `Model.create` `record` type `TCreationAttributes`


todd (approved):
Makes sense to me! 🎉 :shipit:

PR #38653, opened 2019-09-27
New type definition: vue-ls

lexasss:

adding a new definition:

### DIFF
--- a/types/vue-ls/index.d.ts
+++ b/types/vue-ls/index.d.ts
@@ -17,10 +17,22 @@ declare module 'vue/types/vue' {
     }
 }
 
-declare class WebStorage {
+export enum Types {
+    Session = 'session',
+    Local = 'local',
+    Memory = 'memory',
+}
+
+export interface Options {
+    namespace?: string;
+    name?: string;
+    storage?: Types;
+}
+
+export class WebStorage {
     length: number;
 
-    constructor(storage: StorageTypes);
+    constructor(storage: Types);
     get(name: string, def?: any): any;
     set(name: string, value: any, expire?: number): void;
     key(index: number): string | null;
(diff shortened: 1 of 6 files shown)


types/vue-ls/index.d.ts
```diff
@@ -0,0 +1,50 @@
...
+    Session = 'session',
+    Local = 'local',
+    Memory = 'memory',
+}
+
+declare const _default: {
+    install: PluginFunction<StorageOptions>;
+};
+
+export default _default;
```
uniqueiniquity: It seems like the [underlying library](https://unpkg.com/vue-ls@3.2.1/dist/vue-ls.js) is not using `export default`, so this .d.ts file should not either. Instead, you should follow the appropriate template from [the TypeScript handbook](http://www.typescriptlang.org/docs/handbook/declaration-files/templates.html). Note that if you want to use `import name from 'package';` in your consuming code, you can use the `esModuleInterop` compiler setting.

uniqueiniquity (changes requested):

types/vue-ls/index.d.ts
lexasss: You are right, I removed the default export. A new commit was added.

uniqueiniquity (approved):

uniqueiniquity:

types/vue-ls/index.d.ts
lexasss: @uniqueiniquity,
After `default` export was removed and this PR was merged, I figured out that my project where I use `vue-ls` receives a run-time error. After long research and trying out various module templates from  the TypeScript handbook, I found that only the original definition with `export default` works correctly. Neither `declare const VueStorage: ...` nor `declare global { const VueStorage: ... ; }` nor `export const VueStorage: ...` works here even if the TS compiler is happy about it: `VueStorage` is not a valid object when it comes into use.

Given this, and the fact that the `vue-ls` library exports VueStorage into a global space (`(window || self || global).VueStorage = ...`) and also returns this object from the library (`return VueStorage` at the end), is it OK if I open an new PR with `export default` as stated originally?

uniqueiniquity: That's disappointing that the `jsnext:main` and `main` versions of the library don't align - I assume you're using Rollup or Webpack or something of the sort, and getting the version that _does_ have a default export.

In this case, I think it is fine to open a new PR with `export default`. I would also suggest adding a comment explaining that this is because the assumed usage is in an environment that understands ES modules (i.e. Webpack or Rollup).

lexasss: Thanks! Indeed, `vue-ls` is meant to be used in VueJS projects with Webpack running behind (configured automatically by `vue-cli`).

PR #38657, opened 2019-09-27
[onesignal-cordova-plugin] adapt to SDK v 2.6

czabaj:
More accurate definitions for onesignal-cordova-plugin

Public documentation can be found [here](https://documentation.onesignal.com/docs/cordova-sdk), example project is [here](https://github.com/OneSignal/OneSignal-Cordova-Example/blob/master/www/js/index.js).

The documentaion does not contain all SDK methods and some examples in documentation contains error! It is example of misleading documentation and this type definition aims to improve the flaws.

This type definition is based on real-world usage. I did not used all methods though, so it is not guarantied to be 100% accurate, still, better than before.

### DIFF
--- a/types/onesignal-cordova-plugin/index.d.ts
+++ b/types/onesignal-cordova-plugin/index.d.ts
@@ -3,6 +3,7 @@
 // Definitions by: David Broder-Rodgers <https://github.com/broder>
 //                Vaclav Novotny <https://github.com/vaclavnovotny>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.1
 
 interface Window {
 	plugins: CordovaPlugins;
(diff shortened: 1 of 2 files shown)


types/onesignal-cordova-plugin/index.d.ts
```diff
@@ -1,205 +1,364 @@
...
-        Notification = 2
-    }
-
-    const enum OSNotificationPermission {
-        NotDetermined = 0,
-        Authorized = 1,
-        Denied = 2,
-    }
+	interface OneSignalCordovaPlugin {
+		OSInFocusDisplayOption: { None: 0; InAppAlert: 1; Notification: 2 };
```
broder: Are these needed in addition to the enums defined below?

broder (approved):
One small query and some formatting issues from the CI build but otherwise looks good to me, thanks!

types/onesignal-cordova-plugin/index.d.ts
vaclavnovotny: Yes, I think so. Enums are actually defined on SDK object and I cannot find better way of writing this. Setting the property as Enum will force it to be `number` in enums range. As far as I know, `enum`s in `d.ts` files doest not create enum object thus are missing in runtime, since the enums are `const` values should be inlined in `ts` files and it probably works, but we are transpiling our application with *babel* and we are using TS just for type checking, so inlining does not work for us.

PR #38663, opened 2019-09-27
[react-relay] Avoid v6 installs @types/relay-runtime v5

renanmav:

### DIFF
--- a/types/react-relay/package.json
+++ b/types/react-relay/package.json
@@ -1,7 +1,0 @@
-{
-    "private": true,
-    "dependencies": {
-        "@types/react": "*", 
-        "@types/relay-runtime": ">=6.0.3",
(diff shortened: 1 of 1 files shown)


renanmav:
Getting this error
```
For example, if package *P* used to have types on Definitely Typed at @types/P,
but now has its own types, a dependent package *D* will need to use package.json
to refer to @types/P if it relies on old versions of P's types.
```

How to avoid behavior described in the title? 

https://github.com/DefinitelyTyped/DefinitelyTyped#im-writing-a-definition-that-depends-on-another-definition-should-i-use-reference-types--or-an-import

alloy:
Yeah this is very unfortunate behaviour when `*` is used as the version requirement.

PR #38712, opened 2019-09-30
[react-map-gl] Fix callback handler inputs to use ViewportProps instead of ViewState

oturpe:
Based on both react-map-gl documentation and trying things out, both
_ViewportChangeHandler_ and _ViewStrateChangeHandler_ receive a
_ViewportProps_ as input, rather than _ViewState_ as was defined. This
commit fixes that problem.

Fields that were missing include at least 'width' and 'height',
important for working with an auto-resizing map.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/uber/react-map-gl/blob/master/docs/components/interactive-map.md
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header. _Not related to a version upgrade._
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed. _No action._

### DIFF
--- a/types/react-map-gl/index.d.ts
+++ b/types/react-map-gl/index.d.ts
@@ -95,8 +95,10 @@ export interface PositionInput {
 
 export type ViewportChangeHandler = (viewState: ViewportProps) => void;
 
+export type ContextViewportChangeHandler = (viewState: ViewportProps, interactionState: ExtraState, oldViewState: ViewportProps) => void;
+
 export interface MapControllerOptions {
-    onViewportChange?: ViewportChangeHandler;
+    onViewportChange?: ContextViewportChangeHandler;
     onStateChange?: (state: MapState) => void;
     eventManager?: any;
     isInteractive: boolean;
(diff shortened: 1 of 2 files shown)


fnberta (approved):
Thank you, this looks good! Looking at the docs you linked, `onViewportChange` and `onViewStateChange` actually take additional params, namely `interactionState` and `oldViewState`. Maybe you could add these as well?

oturpe:
Adding the extra `interactionState` and `oldViewState` is a good idea. There is the complication that not all the callbacks that are passed the primary `ViewportState` are also passed those two others — I am not sure if that is clear in the documentation, but that's what seems to happen. I am not sure this should be visible in the type definitions or just be a case of unused parameters. I will look into it again and propose something.

oturpe:
Ok, I added commits that add theextra parameters to places where they are actually used. Other places do not use them, so what used to be a single callback function type `ViewStateChangeHandler` split into original `ViewStateChangeHandler` and extended `ContextViewStateChangeHandler`, and similarly for its sibling `ViewportChangeHandler`.

I am not very happy with the name prefix _Context_, since it does not really describe what the difference is. I used it simply because the root of this difference is in a file called `map-context.js`. I could not infer from the upstream what the reason for this difference and thought it is better not to name anything based on a guess.

It would also be better if the original `ViewStateChangeHandler` would be renamed to something like `ControlViewStateChangeHandler` (since all components that use this form are prefixed with _Control_). I did not do that, since that would break backwards compatibility.

I also discovered that `ExtraState` was missing some fields that were used, and was missing optional markings for others that were not always supplied. I fixed this also.

What do you think?

fnberta (approved):
Awesome, thanks a lot! I am also not super happy with `Context` as it is a very overloaded term. But let's keep it for now (I also don't have any better suggestions). Maybe we can think of something more accurate in the future.

PR #38713, opened 2019-09-30
Extend Quill Key.key type

DanielKucal:
It's common to use numeric keycode, especially when we need to access any element of `keyboard.bindings` directly.
Casting to string like:
```
enum KEYS = {
    UP: 38,
    DOWN: 40,
};

keyboard.addBinding({
  key: KEYS.UP.toString(),
  // ...
});
```
does *not* work, so `number` should be allowed as well.


43081j (approved):

43081j:
The docs specify that key can be a number:
https://quilljs.com/docs/modules/keyboard/#configuration

LGTM

PR #38718, opened 2019-09-30
Add types for package "hasbin"

michastreppel:
Please fill in this template.

### DIFF
--- a/types/hasbin/index.d.ts
+++ b/types/hasbin/index.d.ts
@@ -1,0 +1,30 @@
+// Type definitions for hasbin 1.2
+// Project: https://github.com/springernature/hasbin
+// Definitions by: Micha Streppel <https://github.com/michastreppel>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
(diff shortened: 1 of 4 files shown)


RyanCavanaugh:
Why is the tslint rule disabled?

types/hasbin/index.d.ts
```diff
@@ -0,0 +1,30 @@
+// Type definitions for hasbin 1.2
+// Project: https://github.com/springernature/hasbin
+// Definitions by: Micha Streppel <https://github.com/michastreppel>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare function hasbin(bin: string, resultFunction: (result: boolean) => void): void;
```
RyanCavanaugh: This module doesn't create a global. Please follow the standard format for CommonJS packages.

RyanCavanaugh (changes requested):

types/hasbin/index.d.ts
michastreppel: Could you please tell me where I can find this standard format? I cannot seem to find it. Thanks in advance

PR #38724, opened 2019-09-30
remove findById findByPrimary aliases

hassanhe:
Deprecated see [#9372](https://github.com/sequelize/sequelize/issues/9372)


todd (commented):
Hi, thanks for the Pull Request, but please close. These methods were removed in Sequelive V5, but the types in this repository are for V4, where these methods still very much exist.

Sequelize V5 bundles its own types, obviating the need to use these as a dependencies for versions > 5.

PR #38730, opened 2019-09-30
Fix incorrect import/export clashes.

sandersn:
A new error in Typescript 3.7 forbids name clashes like this:

```ts
import { X } from 'y'
export interface X { }
```

Previously they were incorrectly allowed. Typescript 3.7 will have a
beta version in the next day or two. In the meantime you can try
typescript@next -- the nightly build -- to see these errors.

There were generally two fixes

1. Remove the import -- it was unused.
2. Rename the import -- it conflicts but is still used.

I also updated one dependency: ts-toolbelt, which recently shipped a fix for this error.
There are 3 dependencies that are still broken: react-dnd, protractor and three. I'll look at those next and send a PR to those projects if necessary.

### DIFF
--- a/types/azure-sb/lib/models/acstokenresult.d.ts
+++ b/types/azure-sb/lib/models/acstokenresult.d.ts
@@ -14,8 +14,8 @@
 // limitations under the License.
 //
 
-import { Azure } from 'azure-sb';
-import Dictionary = Azure.ServiceBus.Dictionary;
+import { Azure as Az } from 'azure-sb';
+import Dictionary = Az.ServiceBus.Dictionary;
 
 export namespace Azure.ServiceBus.Results {
     export interface AcsTokenResponse extends Dictionary<string | Dictionary<string>> {
--- a/types/actions-on-google/assistant-app.d.ts
+++ b/types/actions-on-google/assistant-app.d.ts
@@ -3,7 +3,7 @@ import * as express from 'express';
 import { BasicCard, BrowseCarousel, BrowseItem, Carousel, ImageDisplays, List, MediaObject,
          MediaResponse, MediaValues, OptionItem, RichResponse, SimpleResponse } from './response-builder';
 import { ActionPaymentTransactionConfig, Cart, GooglePaymentTransactionConfig, LineItem,
-         Location, Order, OrderUpdate, TransactionDecision, TransactionValues } from './transactions';
+         Order, OrderUpdate, TransactionDecision, TransactionValues } from './transactions';
 
 //
 // Note: These enums are exported due to limitations with Typescript and this
--- a/types/azure-sb/lib/models/notificationhubresult.d.ts
+++ b/types/azure-sb/lib/models/notificationhubresult.d.ts
@@ -14,12 +14,11 @@
 // limitations under the License.
 //
 
-import { Azure } from 'azure-sb';
-import Dictionary = Azure.ServiceBus.Dictionary;
+import { Azure as Az } from 'azure-sb';
 
 export namespace Azure.ServiceBus.Results {
     export interface NotificationHubResult {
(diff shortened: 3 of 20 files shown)


sandersn:
Since these are pretty straightforward changes, and the failures cluttering up our DT tests during the 3.7 release, I'm going to merge them today barring any really bad issues.

@andrewbranch, who merged the original PR.

types/azure-sb/lib/models/acstokenresult.d.ts
```diff
@@ -14,8 +14,8 @@
 // limitations under the License.
 //
 
-import { Azure } from 'azure-sb';
-import Dictionary = Azure.ServiceBus.Dictionary;
+import { Azure as Az } from 'azure-sb';
+import Dictionary = Az.ServiceBus.Dictionary;
 
 export namespace Azure.ServiceBus.Results {
```
andrewbranch: Huh. This was _merging_ with `Azure` from `'azure-sb'`, right? It kind of feels like that should still be allowed? The issue that the https://github.com/microsoft/TypeScript/pull/31231 was supposed to fix was when an imported type completely shadowed an exported type. If they merge, and exports from both are accessible, is that a problem?

sandersn: I don't think it *was* merging it. I'll test with an older version of Typescript.

andrewbranch: Wouldn’t it have to in order to resolve `Dictionary`?

sandersn: So, after testing, they do *not* merge:

```ts
// @Filename: second.ts
export namespace Azure.ServiceBus {
    export interface Dictionary<T> {
        [k: string]: T
    }
}

export namespace Azure.ServiceBus.Results {
    export interface AcsTokenResponse extends Dictionary<string | Dictionary<string>> {
        WrapAccessToken: Dictionary<string>;
        WrapAccessTokenExpiresIn: Dictionary<string>;
    }

    export interface AcsTokenResult {
        parse(acsTokenQueryString: string): AcsTokenResponse;
    }
}

// @Filename: first.ts
import { Azure } from './second';
import Dictionary = Azure.ServiceBus.Dictionary;
export namespace Azure.ServiceBus.Results {
    export interface Id {
        d: Dictionary<number>
    }
}

// @Filename: welove.ts
import { Azure as A } from './first'
import { Azure as B } from './second'
let x: A.ServiceBus.Results./*1*/
let y: B.ServiceBus.Results./*2*/
```

At /*1*/, you see only `Id` and at /*2*/ you see only AcsTokenResponse and Result. And you can't import both with the name Azure.

sandersn: Sorry, was replying.

The old behaviour was that references to `Azure` referred to the imported symbol. The module *exports* a namespace named `Azure` too, but that's (1) different (2) not referred to *inside* the module.

andrewbranch: 🤦‍♂ 

I completely missed that the old file had `import Dictionary = Azure.ServiceBus.Dictionary;`. I only saw it in your added changes, so I thought that previously, `Dictionary` was being resolved against the current namespace scope, which would have implied that it was merged somehow. That was the source of my confusion.

This seems fine. Thanks for investigating and unconfusing me.

andrewbranch (approved):

andrewbranch:
BenchmarkPR is going to time out since so many packages changed; feel free to merge any time

sandersn:
Your confusion makes sense; I think the authors of the package thought the namespaces were going to merge too.

guillaume86:
> I just published [`@types/redux-form@8.1.6` to npm](https://www.npmjs.com/package/@types/redux-form).

@RyanCavanaugh is it possible to trigger a v7 publish?

PR #38734, opened 2019-09-30
Update react-dnd-touch-backend dnd-core dependency

sandersn:
dnd-core dependency goes from 4.5 to 9.3.1.
9.3.1 is the first version that fixes an incorrect export that is now detected by Typescript 3.7.

I'm not sure this is the best fix, but I don't see a way to create a new release of dnd-core@4.x either.


sandersn:
Upon further looking around, it seems like react-dnd-touch-backend ships its own types now. I'll just remove it from DT.

sandersn:
eg https://unpkg.com/browse/react-dnd-touch-backend@9.4.0/lib/index.d.ts

sandersn:
Looks like the CI run is confused and is testing the pre-removal version of the code.

sandersn:
Ah, nope, the errors are from react-dnd-multi-backend, which also fails because of dnd-core. But it *not* ship Typescript types, and is *not* up-to-date with 9.x dnd-core. I'm not sure what to do about this one.

PR #38736, opened 2019-09-30
Update react-dnd-multi-backend to version 4.0

sandersn:
Update react-dnd-multi-backend to 4.0

The types haven't changed much since 3.x, and this updates to the version, 9.3.1, of dnd-core that is fixed for use with Typescript 3.7.

The only problem is that it isn't officially available yet, but I don't think 3.x is going to get any more updates since dnd-core 9.x has been out for a while and react-dnd-multi-backend 3.x only supports versions earlier than 9.x. You can find the discussion here: https://github.com/LouisBrunner/dnd-multi-backend/issues/31

### DIFF
--- a/types/react-dnd-multi-backend/index.d.ts
+++ b/types/react-dnd-multi-backend/index.d.ts
@@ -54,6 +54,10 @@ export interface BackendDeclaration {
      */
     backend: BackendFactory;
     /**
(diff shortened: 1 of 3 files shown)


eps1lon (approved):
Blocking #38748

Issue #38656, opened 2019-09-27
[@types/jasminewd2] Error on the last version "cannot find type definition file for 'jasmine/v2'


mkashi:
If you know how to fix the issue, make a pull request instead.

- [ I tried using the `@types/jasminewd2` with the version 2.0.7 package and had problems.
erorr was:

`ERROR in node_modules/@types/jasminewd2/index.d.ts(8,23): error TS2688: Cannot find type definition file for 'jasmine/v2'`

my package.json for information:

```
{
  "name": "ionic-stencil-essential",
  "version": "1.2.0",
  "author": "Ionic Framework",
  "homepage": "http://ionicframework.com/",
  "scripts": {
    "ng": "ng",
    "start": "ng serve",
    "build": "ng build",
    "test": "ng test",
    "lint": "ng lint",
    "e2e": "ng e2e"
  },
  "private": true,
  "dependencies": {
    "@angular/common": "^7.2.2",
    "@angular/core": "^7.2.2",
    "@angular/forms": "^7.2.2",
    "@angular/http": "^7.2.2",
    "@angular/platform-browser": "^7.2.2",
    "@angular/platform-browser-dynamic": "^7.2.2",
    "@angular/router": "^7.2.2",
    "@ionic-native/core": "5.0.0",
    "@ionic-native/splash-screen": "5.0.0",
    "@ionic-native/status-bar": "5.0.0",
    "@ionic/angular": "4.3.0",
    "chart.js": "^2.8.0",
    "cordova-plugin-device": "^2.0.2",
    "cordova-plugin-ionic-keyboard": "^2.1.3",
    "cordova-plugin-ionic-webview": "4.0.0",
    "cordova-plugin-splashscreen": "^5.0.2",
    "cordova-plugin-statusbar": "^2.4.2",
    "cordova-plugin-whitelist": "^1.3.3",
    "core-js": "^2.5.4",
    "jasminewd2": "^2.2.0",
    "ng2-charts": "^1.4.1",
    "node-sass": "^4.12.0",
    "postcss-loader": "^3.0.0",
    "rxjs": "~6.4.0",
    "rxjs-compat": "^6.5.3",
    "tslib": "^1.9.0",
    "zone.js": "^0.8.29"
  },
  "devDependencies": {
    "@angular-devkit/architect": "~0.13.8",
    "@angular-devkit/build-angular": "~0.13.8",
    "@angular-devkit/core": "~7.3.8",
[...]

PR #38737, opened 2019-09-30
[jasminewd2, jasmine] Fix jasminewd2 compatibility with older versions

zvirja:
Please fill in this template.

---

Hotfix for #38656

Fix compatibility with older versions of jasminewd2 to fix horrible regression affecting people. 

The most correct fix should be following:
- update jasminewd2 to depend on jasmine v2 - blocked by https://github.com/microsoft/dtslint/issues/253
- use approach with [path mapping](https://github.com/DefinitelyTyped/DefinitelyTyped#how-do-definitely-typed-package-versions-relate-to-versions-of-the-corresponding-library), but it requires ES imports/exports, which is not how the types are currently written:
    > Also, /// &lt;reference types=".." /> will not work with path mapping, so dependencies must use import.

Current fix solves the issue, but must be revisited going forward.

### DIFF
--- a/types/jasmine/ts3.1/index.d.ts
+++ b/types/jasmine/ts3.1/index.d.ts
@@ -617,9 +617,6 @@ declare namespace jasmine {
     type MatchableArgs<Fn> = Fn extends (...args: infer P) => any ? { [K in keyof P]: P[K] | AsymmetricMatcher<any> } : never;
 
     interface FunctionMatchers<Fn extends Func> extends Matchers<any> {
-        toHaveBeenCalled(): boolean;
-        toHaveBeenCalledBefore(expected: Func): boolean;
-        toHaveBeenCalledTimes(expected: number): boolean;
         toHaveBeenCalledWith(...params: MatchableArgs<Fn>): boolean;
 
         /**
--- a/types/jasminewd2/index.d.ts
+++ b/types/jasminewd2/index.d.ts
@@ -5,7 +5,7 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
-/// <reference types="jasmine/v2" />
+/// <reference types="jasmine" />
 
 declare function it(expectation: string, assertion?: (done: DoneFn) => Promise<void>, timeout?: number): void;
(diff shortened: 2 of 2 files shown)


types/jasminewd2/index.d.ts
```diff
@@ -47,6 +47,14 @@ declare namespace jasmine {
     not: ArrayLikeMatchers<T>;
   }
 
+  // Add definition to be compatible with latest jasmine v3 types.
```
gkalpak: @zvirja, could explain how this helps make jasminewd2 types compatible with jasmine v3 types?

zvirja: Sure! The reason `jasminewd2` is not compatible with the latest jasmine is the `FunctionMatchers<T>` interface I recently added to `jasmine`. Here in `jasminewd2` we are extending `Matchers<T>` interface, so because of that the `FunctionMatchers<T>` in `jasmine` library starts to incorrectly extend the `Matchers<T>` interface (the member with promise becomes missing) and TS linter complains about that.

As you can see around, we extend all the `jasmine` matchers inherited from `Matchers<T>`, e.g. `ArrayLikeMatchers<T>`, so my fix is not special.

Also `FunctionMatchers<T>` is not used for TS versions prior to 3.1, so interface will be there, but will not be actually used.

types/jasmine/ts3.1/index.d.ts
```diff
@@ -617,9 +617,6 @@ declare namespace jasmine {
     type MatchableArgs<Fn> = Fn extends (...args: infer P) => any ? { [K in keyof P]: P[K] | AsymmetricMatcher<any> } : never;
 
     interface FunctionMatchers<Fn extends Func> extends Matchers<any> {
-        toHaveBeenCalled(): boolean;
-        toHaveBeenCalledBefore(expected: Func): boolean;
-        toHaveBeenCalledTimes(expected: number): boolean;
```
abdulrahman-khankan: @zvirja this broke things.
example:
```typescript
expect(someFunctionSpy).toHaveBeenCalledTimes(1);
// Property 'toHaveBeenCalledTimes' does not exist on type 'FunctionMatchers<any>'
```

Issue #34749, opened 2019-04-15
[@types/styled-components] tsserver running at 100% CPU


andrew-humu:
If you know how to fix the issue, make a pull request instead.

- [x] I tried using the `@types/xxxx` package and had problems.
- [ ] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [ ] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @Igorbek @Igmat @lavoaster @Jessidhia

### Versions

* `@types/styled-components@4.1.14`
* `typescript@3.3.3`

### Summary

Upon upgrading to the latest version of `@types/styled-components` and using VS Code, I noticed that editing typescript files was suddenly causing my CPU usage to jump to 100%.  My editor's typescript auto-complete features and lint warnings suddenly took multiple seconds to resolve after typing each character.

On further investigation, I could see that `tsserver.js` was responsible for all of that CPU usage.  Downgrading one version at a time, I was able to isolate the performance slowdown to something that changed between `@types/styled-components@4.0.3` and `@types/styled-components@4.1.0`.

It looks like the two commits added in that version are:

https://github.com/DefinitelyTyped/DefinitelyTyped/commit/5c49d4ddc6f5bd06c5cf04d911398f3db05ead54#diff-9726c25d6953300e0670141c0abac3db

https://github.com/DefinitelyTyped/DefinitelyTyped/commit/59ab5b3aef80fa55010f7d8385f785514d2e0b1a#diff-9726c25d6953300e0670141c0abac3db

### To reproduce

1. In a new typescript project:

```
yarn add typescript@3.3.3 styled-components@4.2.0 @types/styled-components@4.1.0
```

2. Import `styled-components` and begin editing the file.

---

This issue seems similar, but not exactly the same: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34391

deleonn:
I'm experiencing exactly the same issue with:

```js
"typescript": "^3.4.5",
"styled-components": "^4.2.0",
"@types/styled-components": "^4.1.14"
```

Fixed it by downgrading `@types/styled-components` to 4.0.3

ChrisBews:
[...]

PR #38742, opened 2019-09-30
Update type definitions for "faker"

paulovitin:
Please fill in this template.

### DIFF
--- a/types/faker/faker-tests.ts
+++ b/types/faker/faker-tests.ts
@@ -66,13 +66,13 @@ resultDate = faker.date.soon();
 resultDate = faker.date.soon(30);
 resultStr = faker.date.month();
 resultStr = faker.date.month({
-  abbr: true,
-  context: true,
+	abbr: true,
+	context: true
 });
 resultStr = faker.date.weekday();
 resultStr = faker.date.weekday({
-  abbr: true,
-  context: true,
+	abbr: true,
+	context: true
 });
 
 resultStr = faker.finance.account();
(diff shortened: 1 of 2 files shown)


mattbishop:
@andrewbranch I love the PR perf tool but I don't know what value it has for a bunch of typing files? Definitely something I'd use in other projects though.

types/faker/faker-tests.ts
```diff
@@ -65,18 +66,19 @@ resultDate = faker.date.soon();
 resultDate = faker.date.soon(30);
 resultStr = faker.date.month();
 resultStr = faker.date.month({
-	abbr: true,
-	context: true
+    abbr: true,
```
mattbishop: Whitespace changes and trailing commas? Not needed.

mattbishop (changes requested):
Please undo the style changes (whitespace, trailing commas, single quotes) and I will approve.

andrewbranch:
@mattbishop the perf tool measures the impact of these typings on a consumer’s compilation time, memory footprint, and editor responsiveness for services like checking type info and getting statement completion. The reason it exists is that it is possible to write types on DefinitelyTyped that are so complex and difficult for the compiler to reason about that when a user installs them (which might even happen automatically under the hood in VS Code, if you’re a JavaScript user), they essentially DDoS’s the user’s project compilation and/or editor. This is not a hypothetical conjecture; [it happened](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/34749).

types/faker/faker-tests.ts
paulovitin: @mattbishop done.

mattbishop:
Woah @andrewbranch ! Thank you for explaining the reasoning for testing compile times.

mattbishop (approved):
Thanks!

PR #38748, opened 2019-10-01
chore(react): Clarify React.Attributes

eps1lon:
Second attempt at #38352 which had to be reverted due to a merge conflict in #38603.

The problem was that the actual `Props` interface of the component didn't match any properties validated in `propTypes`. While this may be valid js/ts to write it doesn't make a lot of sense. It looks like this passed previously only accidentally due to `React.Attributes` usage.


eps1lon:
The errors originates from `react-dnd@7.7.0`. This particular error is fixed in later versions but would need a backport (https://github.com/react-dnd/react-dnd/issues/1560).

sandersn:
@eps1lon I just merged #38736. I'll look at the other react-dependent two failures next.

sandersn:
#38777 should fix the other two failures.

PR #38640, opened 2019-09-26
[relay-runtime] Split files to better maintenance

renanmav:
I've tried to match typings with source code of [relay-runtime](https://github.com/facebook/relay/tree/master/packages/relay-runtime).

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.

PR #38749, opened 2019-10-01
fix: Improve Relay types some more

taion:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/relay/blob/71d2a70839920b7d4f49ecc6e290d2ad03fd08bb/packages/relay-runtime/store/readInlineData.js#L25-L40


types/react-relay/index.d.ts
```diff
@@ -123,39 +125,28 @@ export { ReactRelayQueryRenderer as QueryRenderer };
...
-}
-
 export type ContainerProps<Props> = MappedFragmentProps<Pick<Props, Exclude<keyof Props, 'relay'>>>;
 
 export type Container<Props> = React.ComponentType<ContainerProps<Props> & { componentRef?: (ref: any) => void }>;
 
 export function createFragmentContainer<Props>(
     Component: React.ComponentType<Props & { relay?: RelayProp }>,
-    fragmentSpec: GraphQLTaggedNode | GeneratedNodeMap
+    fragmentSpec: Record<string, GraphQLTaggedNode>,
```
taion: these only take objects now

types/react-relay/index.d.ts
```diff
@@ -123,39 +125,28 @@ export { ReactRelayQueryRenderer as QueryRenderer };
...
     edges?: ReadonlyArray<any> | null;
     pageInfo?: Partial<PageInfo> | null;
 }
 
-export function readInlineData<T extends _RefType<any>>(
-    fragment: GraphQLTaggedNode,
-    ref: FragmentOrRegularProp<T>,
-): T;
-
-type FragmentVariablesGetter = (prevVars: Variables, totalCount: number) => Variables;
```
taion: inlined

types/react-relay/index.d.ts
```diff
@@ -123,39 +125,28 @@ export { ReactRelayQueryRenderer as QueryRenderer };
...
-    fragmentSpec: GraphQLTaggedNode | GeneratedNodeMap
+    fragmentSpec: Record<string, GraphQLTaggedNode>,
 ): Container<Props>;
 
 interface ConnectionData {
     edges?: ReadonlyArray<any> | null;
     pageInfo?: Partial<PageInfo> | null;
 }
 
-export function readInlineData<T extends _RefType<any>>(
```
taion: this is actually in relay-runtime

types/relay-runtime/lib/store/readInlineData.d.ts
```diff
@@ -0,0 +1,19 @@
+import { GraphQLTaggedNode } from '../query/RelayModernGraphQLTag';
+
+interface _RefType<T> {
+    ' $refType': T;
+}
+interface _FragmentRefs<T> {
+    ' $fragmentRefs': T;
+}
+
+type FragmentRef<T> = T extends _RefType<infer U> ? _FragmentRefs<U> : never;
```
taion: helper type to get something that holds onto the fragment ref (see `react-relay` test)

types/relay-runtime/lib/store/readInlineData.d.ts
```diff
@@ -0,0 +1,19 @@
...
+
+declare function readInlineData<T extends _RefType<any>>(fragment: GraphQLTaggedNode, ref: FragmentRef<T>): T;
+declare function readInlineData<T extends _RefType<any>>(
+    fragment: GraphQLTaggedNode,
+    ref: FragmentRef<T> | null | undefined,
+): T | null | undefined;
+declare function readInlineData(fragment: GraphQLTaggedNode, ref: null | undefined): null | undefined;
```
taion: is this overloading overkill? see https://github.com/facebook/relay/blob/71d2a70839920b7d4f49ecc6e290d2ad03fd08bb/packages/relay-runtime/store/readInlineData.js#L46-L48

i.e. `readInlineData` returns null-y iff it's called with something nully.

though arguably calling it with something statically known to be null-y is silly. what's the right way to express this sort of function that passes through nulls?

taion: or do i just remove the overload that takes `FragmentRef<T> | null | undefined`?

renanmav (commented):
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38749/files#diff-c54878a488116f9a68a889c37d574a2bR648

The type assertion on `viewer` is making CI fail

taion:
Okay, I had to update `RelayStoreTypes` and a few other modules to address some regressions from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38640.

`mixed` was incorrectly rendered as `any` instead of `unknown`, and a number of optionals were not carried over.

Additionally, a few of the types added were actually not in v6.0.0.

types/relay-runtime/lib/store/readInlineData.d.ts
taion: i think we need to export some kind of helper like this to make it reasonably possible for users to type `readInlineData` calls

taion:
Any Flow's `Object` is more like `{[key: string]: any}` than like TS `object`.

types/relay-runtime/index.d.ts
```diff
@@ -62,8 +62,6 @@ export {
     FragmentReference,
     FragmentSpecResolver,
     HandleFieldPayload,
-    LogEvent,
```
renanmav: This types actually exists on `relay-runtime` source code.

Check this https://github.com/facebook/relay/blob/master/packages/relay-runtime/index.js#L119-L120

taion: not in 6.0.0: https://github.com/facebook/relay/blob/b7501c6472b34ae2b0bb8cc164a648bb90749d74/packages/relay-runtime/index.js#L126-L127

renanmav: Fair enough. I was looking at master branch.

taion:
I'm puzzled by the CI failure here. There's a non-null assert on `viewer` there.

types/relay-runtime/lib/store/RelayRecordSource.d.ts
```diff
@@ -2,20 +2,19 @@ import { MutableRecordSource, RecordMap, Record } from './RelayStoreTypes';
 import { DataID } from '../util/RelayRuntimeTypes';
 import { RecordState } from './RelayRecordState';
 
-export class RelayRecordSource {
+export class RelayRecordSource implements MutableRecordSource {
```
renanmav: Why this class implements `MutableRecordSource`?

https://github.com/facebook/relay/blob/v6.0.0/packages/relay-runtime/store/RelayRecordSource.js#L19

types/relay-runtime/lib/store/readInlineData.d.ts
renanmav: honestly, I'm trying to figure it out what `readInlineData` does

taion: see https://relay.dev/docs/en/graphql-in-relay.html#inline

maybe i should must match the overloads in flow

types/relay-runtime/lib/store/RelayRecordSource.d.ts
taion: flow does https://github.com/facebook/relay/blob/v6.0.0/packages/relay-runtime/store/RelayRecordSource.js#L20

but TS doesn't support ctor return types: https://github.com/Microsoft/TypeScript/issues/27594

we _could_ just use pure structural typing but implementing the interface means we get earlier errors if we forget to specify a method

taion:
okay, there was a silly missing non-null assertion in tests

the remaining question here is that the `FragmentRef<T>` API here is new, but it seems like something like this is required to make `readInlineData` work appropriately...

as in, if you have a function that itself calls `readInlineData`, then all you care is that, _given_ a `_RefType`, you get a corresponding `_FragmentRefs`.

i think `FragmentRef` is reasonable. could also be `Fragment` if we want consistency with `FragmentOrRegularProp` in the `react-relay` typings, but that naming is weird/wrong... it's a fragment ref, not a fragment, that's what's passed in

taion:
this is sort of a bugfix, too – currently `commitMutation` is not compatible with the exported `Environment` type

types/relay-runtime/lib/store/readInlineData.d.ts
alloy: Sure, let's export it from the top-level module?

types/relay-runtime/lib/store/readInlineData.d.ts
alloy: I dunno, it seems like the function isn’t useful in any way when passing in `null`, right? There’s runtime code to guard against it, but that doesn’t mean imo that we need to type it up as valid use-cases. Let's just type what is actually useful?

taion: so what _is_ useful is that you can pass in a nullable fragment ref and then get back a nullable piece of data

what's the canonical way to represent that? the point i'm looking to convey is something like:

- if you pass in something non-nullable, you'll get something non-null back
- if you pass in something nullable, you'll get back a nullable fragment

i guess it's probably not worth guarding against someone passing in something statically known to be null? i'm guessing that the typing here is just wrong: https://github.com/facebook/relay/blob/71d2a70839920b7d4f49ecc6e290d2ad03fd08bb/packages/relay-runtime/store/readInlineData.js#L39

taion: i took out the null-only override

types/relay-runtime/lib/store/readInlineData.d.ts
taion: yup, it's exported from both

types/relay-runtime/lib/store/readInlineData.d.ts
alloy: > so what is useful is that you can pass in a nullable fragment ref and then get back a nullable piece of data

Gotcha. The overload seems fine then, or is it not working as expected? Alternatively you could use conditional types to more accurately represent the permutations.

* https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html
* https://artsy.github.io/blog/2018/11/21/conditional-types-in-typescript/

types/relay-runtime/lib/store/readInlineData.d.ts
alloy: But then this module could perhaps just import it from the top-level module?

types/relay-runtime/lib/store/readInlineData.d.ts
taion: nope, this is fine – i'm going to err to the side of following the flow types here

types/relay-runtime/lib/store/readInlineData.d.ts
taion: it doesn't look like we have any other types declared at top level in `relay-runtime`. do you think that makes sense to add / wouldn't lead to circular import issues?

types/relay-runtime/lib/store/readInlineData.d.ts
alloy: That makes sense to me 👍

types/relay-runtime/lib/store/readInlineData.d.ts
alloy: It shouldn’t lead to import issues afaik. I’d rather have them there then introduce a new file that doesn’t exist upstream or defining these trickier types in more than one place.

alloy:
It’s getting a little noisy, so let me know when it’s considered done 👍

taion:
i moved the utility types to the root of `relay-runtime`, and now import from there both in `readInlineData` and `react-relay`. this should be gtg now

alloy:
Great, let’s do this :shipit:

PR #38774, opened 2019-10-01
React - MouseEvent correctly inherits UIEvent properties

lwilkins:
https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent at time of
writing shows that MouseEvent should share properties from UIEvent.

detail and view properties are now available on mouse events.

--------------------

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -1118,7 +1118,7 @@ declare namespace React {
         which: number;
     }
 
-    interface MouseEvent<T = Element, E = NativeMouseEvent> extends SyntheticEvent<T, E> {
+    interface MouseEvent<T = Element, E = NativeMouseEvent> extends UIEvent<T, E> {
         altKey: boolean;
         button: number;
         buttons: number;
@@ -1154,7 +1154,7 @@ declare namespace React {
         touches: TouchList;
     }
 
-    interface UIEvent<T = Element> extends SyntheticEvent<T, NativeUIEvent> {
+    interface UIEvent<T = Element, E = NativeUIEvent> extends SyntheticEvent<T, E> {
         detail: number;
         view: AbstractView;
     }


types/react/index.d.ts
```diff
@@ -1118,7 +1118,7 @@ declare namespace React {
         which: number;
     }
 
-    interface MouseEvent<T = Element, E = NativeMouseEvent> extends SyntheticEvent<T, E> {
+    interface MouseEvent<T = Element, E = NativeMouseEvent> extends UIEvent<T, E> {
```
eps1lon: Reacts synthetic events and native DOM events are two different things. Please include documentation that shows that properties of UIEvent are part of reacts synthetic events. The native mouse events should already inherit from UIEvent.

eps1lon (changes requested):

types/react/index.d.ts
Jessidhia: https://github.com/facebook/react/blob/master/packages/react-dom/src/events/SyntheticMouseEvent.js#L21

Looks like it does extend the (Synthetic)UIEvent.

Jessidhia: `.extend` is itself https://github.com/facebook/react/blob/master/packages/legacy-events/SyntheticEvent.js#L227-L249 which is an ES5 implementation of `class` inheritance.

Jessidhia: (Synthetic)`TouchEvent` and (Synthetic)`KeyboardEvent` also do `.extend` from (Synthetic)`UIEvent` so it looks like this PR is in the right direction, just with a smaller scope than it could be.

lwilkins: Thanks @eps1lon @Jessidhia for looking at the pr.

@Jessidhia you beat me to it! I found the link in the source code last night but didn't have the chance to comment on the pr until now. I did try to look through the documentation otherwise and couldn't find a reference to the direct link there.

Other than the build failure, I'm unclear if I need to make further changes for this pr to be considered ready to merge?

On the build failure front, is there a cut down version of the tests I can run to replicate the problem? I can't see any obvious switches/options to pass to the test runner to do so. On the face of the error, it would appear unrelated to my change, but the feedback loop is very long (test section in Azure is about 26 minutes...!).

eps1lon: > Other than the build failure, I'm unclear if I need to make further changes for this pr to be considered ready to merge?

I think you're extending the native UIEvent. You should extend the SyntheticUIEvent.

Jessidhia: No, it is the locally-declared `UIEvent`. The events here are declared without a prefix (it's `React.UIEvent`, `React.MouseEvent`, etc, but since we're already inside the `React` namespace we don't use the `React.` prefix)

The native events are shadowed and aliased with a `Native` prefix, which is why `E = NativeMouseEvent`.

eps1lon: Then `SyntheticMouseEvent` shouldn't have been there in the first place anymore?

eps1lon (approved):

uniqueiniquity:

danilofuchs:
I am having trouble with this PR.
All of my `onClick` handlers are throwing these errors:
```typescript
const handler = (event?: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
            // ... do stuff
}

return <button onClick={handler} />

// Types of parameters 'event' and 'event' are incompatible.
//          Type 'MouseEvent<HTMLDivElement, MouseEvent>' is missing the following
//          properties from type 'MouseEvent<HTMLButtonElement, MouseEvent>': detail, view
```
How should I type my handlers now?

PR #38779, opened 2019-10-01
Update export style

MaxySpark:
If we change `export = Web3` to `export declare class Web3 { }`
we can use it without `"esModuleInterop": true,` in tsconfig.json

if we use `"esModuleInterop": true,` in tsconfig.json

then, `import * as lodash from 'lodash'`, this type of imports will starts breaking

After changing `export = Web3` to `export declare class Web3 { }`

we can directly use `import { Web3 } from 'web3';`

Please fill in this template.

- [ * ] Use a meaningful title for the pull request. Include the name of the package modified.
- [ * ] Test the change in your own code. (Compile and run.)

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/web3/index.d.ts
+++ b/types/web3/index.d.ts
@@ -53,4 +53,3 @@ export declare class Web3 {
     setProvider(provider: Provider): void;
     utils: Utils;
 }
-
--- a/types/web3/index.d.ts
+++ b/types/web3/index.d.ts
@@ -29,7 +29,7 @@ import { BatchRequest, Net, Personal } from "./eth/types";
 import Utils from "./utils";
 import Eth from "./eth/index";
 
-declare class Web3 {
+export declare class Web3 {
     static providers: Providers;
     static givenProvider: Provider;
     static modules: {
@@ -54,4 +54,3 @@ declare class Web3 {
     utils: Utils;
 }
 
-export = Web3;
--- a/types/web3/web3-tests.ts
+++ b/types/web3/web3-tests.ts
@@ -4,6 +4,7 @@ import { TransactionReceipt, Log } from "web3/types";
 import PromiEvent from "web3/promiEvent";
 import { NEW_ABI_STANDARD, OLD_ABI_STANDARD } from "web3/test/abi-tests";
 import { Provider, JsonRPCResponse } from "web3/providers";
+import { TxSignature } from "web3/eth/accounts";
 
 const contractAddress = "0xde0B295669a9FD93d5F28D9Ec85E40f4cb697BAe";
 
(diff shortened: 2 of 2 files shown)


phra:
try to use

```typescript
import lodash = require('lodash')
```

more info: https://stackoverflow.com/a/29598404/8006810

MaxySpark:
> 
> 
> try to use
> 
> ```ts
> import lodash = require('lodash')
> ```
> 
> more info: https://stackoverflow.com/a/29598404/8006810

ok. but most people use `import` syntex like `import * as ...` this.

And I try to I tried to fixed the ci errors. Fixed 3 but don't know how to fix the remaining two. u can have a look.

```
Error: /home/travis/build/DefinitelyTyped/DefinitelyTyped/types/web3/web3-tests.ts:14:18

ERROR: 14:18  expect  TypeScript@next compile error: 

This expression is not constructable.

  Type 'typeof import("/home/travis/build/DefinitelyTyped/DefinitelyTyped/types/web3/index")' has no construct signatures.

ERROR: 176:1  expect  TypeScript@next: Expected an error on this line, but found none
```

types/web3/index.d.ts
```diff
@@ -53,5 +53,3 @@ declare class Web3 {
     setProvider(provider: Provider): void;
     utils: Utils;
 }
-
-export = Web3;
```
phra: this is compatible only with web3.js 1.x

https://github.com/ethereum/web3.js/blob/1.x/packages/web3/src/index.js

types/web3/index.d.ts
```diff
@@ -29,7 +29,7 @@ import { BatchRequest, Net, Personal } from "./eth/types";
 import Utils from "./utils";
 import Eth from "./eth/index";
 
-declare class Web3 {
+export declare class Web3 {
```
phra: this is only compatible with web3.js 2.x because esm is available:

https://github.com/ethereum/web3.js/blob/2.x/packages/web3/package.json#L12

types/web3/web3-tests.ts
```diff
@@ -174,7 +175,7 @@ const weiBn: BigNumber = web3.utils.toWei(web3.utils.toBN("1"));
 // $ExpectError
 const weiNumber = web3.utils.toWei(100);
 
-const rndHex: string = Web3.utils.randomHex(10);
+const rndHex: string = web3.utils.randomHex(10);
```
phra: maybe these changes should be in a separated PR..

phra (changes requested):

types/web3/index.d.ts
MaxySpark: now I am using like this `import web3 = require('web3')`

`new web3.Web3('ws://localhost:8545')`

I am getting

`TypeError: web3.Web3 is not a constructor`

can u plz help me how to solve this!

phra:
```typescript
import Web3 = require('web3')
const web3 = new Web3()
```

PR #38783, opened 2019-10-01
[@types/express-session] Fix incorrect type for Session.all(callback)

Calvin-LL:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/expressjs/session/blob/master/README.md

According to https://github.com/expressjs/session/blob/master/README.md 

>`store.all(callback)`
>Optional
>
>This optional method is used to get all sessions in the store as an **array**. The callback should be called as `callback(error, sessions)`.

### DIFF
--- a/types/express-session/index.d.ts
+++ b/types/express-session/index.d.ts
@@ -95,7 +95,7 @@ declare namespace session {
     get: (sid: string, callback: (err: any, session?: Express.SessionData | null) => void) => void;
     set: (sid: string, session: Express.SessionData, callback?: (err?: any) => void) => void;
     destroy: (sid: string, callback?: (err?: any) => void) => void;
-    all: (callback: (err: any, obj?: { [sid: string]: Express.SessionData; } | null) => void) => void;
+    all: (callback: (err: any, obj?: Express.SessionData[] | null) => void) => void;
     length: (callback: (err: any, length?: number | null) => void) => void;
     clear: (callback?: (err?: any) => void) => void;
     touch: (sid: string, session: Express.SessionData, callback?: (err?: any) => void) => void;
--- a/types/express-session/index.d.ts
+++ b/types/express-session/index.d.ts
@@ -95,7 +95,7 @@ declare namespace session {
     get: (sid: string, callback: (err: any, session?: Express.SessionData | null) => void) => void;
     set: (sid: string, session: Express.SessionData, callback?: (err?: any) => void) => void;
     destroy: (sid: string, callback?: (err?: any) => void) => void;
-    all: (callback: (err: any, obj?: Express.SessionData[] | null) => void) => void;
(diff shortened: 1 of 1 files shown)


types/express-session/index.d.ts
```diff
@@ -95,7 +95,7 @@ declare namespace session {
     get: (sid: string, callback: (err: any, session?: Express.SessionData | null) => void) => void;
     set: (sid: string, session: Express.SessionData, callback?: (err?: any) => void) => void;
     destroy: (sid: string, callback?: (err?: any) => void) => void;
-    all: (callback: (err: any, obj?: { [sid: string]: Express.SessionData; } | null) => void) => void;
+    all: (callback: (err: any, obj?: Express.SessionData[] | null) => void) => void;
```
armanio123: I don't think this implementation is correct. Documentation doesn't state this is an `array` and based on the code. This seems like an object with ids mapping to the SessionData.

See this for details: https://github.com/expressjs/session/blob/4d253405aca773e3e994c0259a3bc658c22430e0/session/memory.js#L58

I'm not familiar with the library as I don't use it. Could you please confirm if I'm looking at the right place?

armanio123 (changes requested):

types/express-session/index.d.ts
CoolCyberBrain: Good catch. Did not notice that. It seems to be a mismatch of documentation and code. So the store object default to the file you pointed to. But as the documentation states 

> Warning The default server-side session storage, MemoryStore, is purposely not designed for a production environment. It will leak memory under most conditions, does not scale past a single process, and is meant for debugging and developing.

So in production store is set to a third party library listed at the bottom of this library's [README](https://github.com/expressjs/session#compatible-session-stores).

The two most popular [connect-redis](https://github.com/tj/connect-redis) and [connect-mongo](https://github.com/jdesboeufs/connect-mongo) follow the documentation and returns an Array instead of an Object.

See here: https://github.com/jdesboeufs/connect-mongo/blob/master/src/index.js#L365 and
https://github.com/tj/connect-redis/blob/master/lib/connect-redis.js#L101

In an effort to not break anything I have added both the documented type Array and the undocumented default dev-only type Object. I will also create a pull request on the expressjs/session library to change the documentation to include both types.

CoolCyberBrain: Quick follow up in https://github.com/expressjs/session/pull/700#issuecomment-540855551 a member of the team points out that the default MemoryStore.all is undocumented whereas most users of this library will be using this function according to the documentation. Seems reasonable to follow the documentation while allowing the undocumented types.

Thanks for taking your time to review this PR :)

armanio123: Makes sense. Thanks for taking a look at it.

armanio123 (approved):

PR #38796, opened 2019-10-02
[mock-fs] Update type definitions to v4.10.

Seally:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/tschaub/mock-fs/blob/master/changelog.md

`mock-fs` v3 => v4 made significant changes to the API, so in order to avoid having to check for every single change, I rewrote the type definitions from scratch for my (currently) private project. I've also added type definitions for some other files which are referenced and exported by certain functions in the main `index.js` for convenience in typing some of the types in there, even though most users will probably only be interested in the main export (though they can import them and it should match up with the NPM package exports).

Note: Should I remove the v3 maintainers from the list for v4?


RyanCavanaugh:
@Nemo157 @tqkubo care to review?

types/mock-fs/mock-fs-tests.ts
```diff
@@ -1,85 +1,156 @@
+import mock = require('mock-fs');
```
armanio123: Whey most of the mock tests were removed, are they no longer valid?

armanio123 (commented):
I have very limited knowledge on the library so I have just reviewed the types implementation and not the library and LGTM.

Just a comment,

types/mock-fs/mock-fs-tests.ts
Seally: Looking at this again (it's been some time since I've written this), I combined parts of the old tests into one larger one (and added some of my own) to check nesting. 

With that said, looking at it again, I think I'll re-add the v3 tests, but I will strip them out of the functions definitions. I have a separate repo that runs this test file, so the functions need to be called when this file is  `require`'d. I've had a few problems in the past with type definitions typing things that don't exist (not for `mock-fs` though), so I created the runnable version to make sure the functions can be called without errors and work like I think they do.

`mock.fs()` no longer exists in v4 (for now), so the test for that was removed.

I've also realized that I've forgotten to test the second parameter on `mock()`, so I'll add that too. Thanks for this mention since I would've missed that otherwise. The type definitions were part of my private repo and was written a bit independently of the official type definitions, which is why certain things ended up so different.

andrewbranch:
mock-fs has been on v4 for three years now, and it looks like v3 isn't actively maintained. I think it's _probably_ safe to delete the v3 typings.

Seally:
I removed the v3 typings. Not sure if it was the right choice or not, but I'm assuming (completely unverified statement) that because the library messes a bit with Node internals `mock-fs` v3 might not even work anymore in the latest Node. 

Performed a rebase, otherwise `npm test` will fail on missing v3 files.

@tkqubo @Nemo157 Would you still be interested in maintaining this?

PR #38800, opened 2019-10-02
[pnpapi] Updates the PnP API

arcanis:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://next.yarnpkg.com/advanced/pnpapi

### DIFF
--- a/types/pnpapi/index.d.ts
+++ b/types/pnpapi/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for pnpapi 0.0
+// Type definitions for pnpapi 3.0
 // Project: https://yarnpkg.github.io/berry/advanced/pnpapi
 // Definitions by: Maël Nison <https://github.com/arcanis>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
--- a/types/pnpapi/index.d.ts
+++ b/types/pnpapi/index.d.ts
@@ -11,50 +11,61 @@ declare namespace NodeJS {
     }
 }
 
-declare module 'pnpapi' {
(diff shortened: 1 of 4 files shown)


types/pnpapi/index.d.ts
```diff
@@ -1,41 +1,60 @@
...
+    interface ProcessVersions {
+        pnp?: string;
+    }
 }
 
-export interface TopLevelPackageLocator {
-    name: null;
-    reference: null;
-}
+declare module 'pnpapi' {
```
RyanCavanaugh: This is going to substantially worsen the end-user experience of this package. You can separate out the `NodeJS` augmentation into a different file and leave this file's format as-is.

RyanCavanaugh (changes requested):

types/pnpapi/index.d.ts
arcanis: Sounds good, I'll check tomorrow. Why is the declare module approach discouraged btw? What use cases does it break?

arcanis:
Ping @RyanCavanaugh ? 😃

types/pnpapi/tslint.json
```diff
@@ -1 +1,8 @@
-{ "extends": "dtslint/dt.json" }
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "no-single-declare-module": false,
```
weswigham, line 4: In lieu of @RyanCavanaugh 's feedback - you should be able to remove these lint rule adjustments now, right?

weswigham (changes requested):

PR #38801, opened 2019-10-02
Update signature on unstable_scheduleCallback

mauricedb:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: See the current signature here: https://github.com/facebook/react/blob/master/packages/scheduler/src/Scheduler.js#L295

### DIFF
--- a/types/scheduler/index.d.ts
+++ b/types/scheduler/index.d.ts
@@ -20,7 +20,7 @@ export const unstable_NormalPriority = 3;
 export const unstable_IdlePriority = 5;
 export const unstable_LowPriority = 4;
 export function unstable_runWithPriority<T>(priorityLevel: number, eventHandler: () => T): T | undefined;
-export function unstable_scheduleCallback(priorityLevel: number, callback: FrameCallbackType, options?: { delay: number}): CallbackNode;
+export function unstable_scheduleCallback(priorityLevel: number, callback: FrameCallbackType, options?: { delay?: number, timeout?: number}): CallbackNode;
 export function unstable_cancelCallback(callbackNode: CallbackNode): void;
 export function unstable_wrapCallback(callback: FrameCallbackType): () => FrameCallbackType | undefined;
 export function unstable_getCurrentPriorityLevel(): number;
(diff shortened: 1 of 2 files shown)


types/scheduler/index.d.ts
```diff
@@ -20,7 +20,7 @@ export const unstable_NormalPriority = 3;
 export const unstable_IdlePriority = 5;
 export const unstable_LowPriority = 4;
 export function unstable_runWithPriority<T>(priorityLevel: number, eventHandler: () => T): T | undefined;
-export function unstable_scheduleCallback(callback: FrameCallbackType, deprecated_options?: { timeout: number}): CallbackNode;
+export function unstable_scheduleCallback(priorityLevel: number, callback: FrameCallbackType, options?: { delay: number}): CallbackNode;
```
eps1lon: Timeout is still an option: https://github.com/facebook/react/blob/d346d92084316ec10714b9ac4bfb1ebcfdf05fb9/packages/scheduler/src/Scheduler.js#L308 and both of them are optional

eps1lon (changes requested):
Thank you for updating them.

Please don't forget to to update the version header to 0.16

types/scheduler/index.d.ts
mauricedb: @eps1lon Oops, missed that one. The timeout back in now and both timeout and delay are optional.

types/scheduler/test/scheduler.ts
```diff
@@ -1,7 +1,7 @@
-import { unstable_scheduleCallback, unstable_cancelCallback, unstable_now } from 'scheduler';
+import { unstable_scheduleCallback, unstable_cancelCallback, unstable_now, unstable_NormalPriority } from 'scheduler';
 
 // $ExpectType CallbackNode
-const callbackNode = unstable_scheduleCallback(() => {}, { timeout: 100 });
+const callbackNode = unstable_scheduleCallback(unstable_NormalPriority, () => {}, { delay: 100 });
```
eps1lon: Last nitpick: Only adjust this test to pass and add a new one for delay.

eps1lon (approved):

PR #38813, opened 2019-10-02
ReactDOM.createRoot takes an HTMLElement as container

mauricedb:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOM.js#L378

### DIFF
--- a/types/react-dom/index.d.ts
+++ b/types/react-dom/index.d.ts
@@ -112,13 +112,13 @@ export interface SyncRoot {
     unmount(callback?: () => void): void;
 }
 
-export function unstable_createSyncRoot(container: HTMLElement, options?: RootOptions): SyncRoot;
+export function unstable_createSyncRoot(container: Element | null, options?: RootOptions): SyncRoot;
 
 export interface Root extends SyncRoot {
   createBatch(): Batch;
 }
 
 export function unstable_createRoot(
-    container: HTMLElement,
+    container: Element | null,
     options?: RootOptions
 ): Root;
--- a/types/react-dom/index.d.ts
+++ b/types/react-dom/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for React (react-dom) 16.9
+// Type definitions for React (react-dom) 16.10
 // Project: http://facebook.github.io/react/
 // Definitions by: Asana <https://asana.com>
 //                 AssureSign <http://www.assuresign.com>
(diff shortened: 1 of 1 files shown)


Jessidhia:
Rather, it is `Element | Document` instead of `HTMLElement`. You can give it almost any DOM node that is capable of having children (or is a `Document` because of special handling).

https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOM.js#L390-L398

https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOM.js#L176-L184

mauricedb:
@Jessidhia You are right. 

There is a mismatch between the types that causes a compile error. I was assuming that this was the result of an `HTMLElement` from `node_modules\typescript\lib\lib.dom.d.ts` being compared with an `Element` from `node_modules\@types\react\global.d.ts` but it appears I was wrong. It seems the actual issue is that `document.getElementById("root")` can return a `null` so the signature (in strict mode) is `HTMLElement | null` while `unstable_createRoot()` only allows for `Element` and not `null`. So it looks like the correct fix is `unstable_createRoot(container: Element | null, options?: RootOptions): Root;`

Jessidhia:
Also no, because that will cause an `invariant` error at runtime as `null` is not a valid container.

If you really want to give the result of `document.getElementById('root')` directly to `createRoot`, you *actually* need to use postfix `!` on the result, where you promise TypeScript it'll never _actually_ be `null` because you know what it should return at runtime. So either you use `document.getElementById('root')!` as the argument, or you need to actually check the result before using it.

mauricedb:
@Jessidhia I agree that a `null` reference isn't going to work at runtime I would recommend keeping this consistent with `ReactDOM.render()` where the container parameter allows for `null` making the developers life easier. 

Doing an actual `null` check would be the correct thing here.

Jessidhia:
That's a bug in the declarations, then; if the function does not accept `null`, then it should not be declared to accept `null`s (and it indeed does not -- it does the same `isValidContainer` check as `ReactRoot`s do). If you want to pass `null`s in anyway, then assert them away with `!`.

mauricedb:
@Jessidhia Sounds reasonable to me. I will close this PR for now.

I can open another for `ReactDOM.render()` but it would be a breaking change as CRA generates `ReactDOM.render(<App />, document.getElementById('root'));` without the additional `null` check. That could be updated with a PR to CRA but all existing projects would still be affected.

PR #38815, opened 2019-10-02
fix: Parse.File save method enable option for progress callback

max4kaps:
according to official specs (link below), there are those options:
Valid options are:
useMasterKey: In Cloud Code and Node only, causes the Master Key to be used for this request.
progress: In Browser only, callback for upload progress
<<https://parseplatform.org/Parse-SDK-JS/api/master/Parse.File.html>>


RaschidJFR (approved):
LGTM

yagotome (approved):

PR #38824, opened 2019-10-02
[Stripe] Improved Stripe.tokens.IToken

htunnicliff:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://stripe.com/docs/api/tokens/object
----
I refactored the `IToken` interface to a Union type to better reflect the nature of tokens, in that they are either a `card` or `bank_account`. 

### Existing
```ts
const token = stripe.tokens.retrieve("tok_xxxxx");
if (token.type === "card") {
  console.log(token.card.fingerprint)
  // Error: Object is possibly 'undefined'. ts(2532)
}
```

### With Union Type
```ts
const token = stripe.tokens.retrieve("tok_xxxxx");
if (token.type === "card") {
  console.log(token.card.fingerprint)
  // No error!
}
```

### DIFF
--- a/types/stripe/index.d.ts
+++ b/types/stripe/index.d.ts
@@ -6977,10 +6977,10 @@ declare namespace Stripe {
     }
 
     namespace tokens {
-        type IToken = ICardToken | IBankAccountToken
+        type IToken = ICardToken | IBankAccountToken;
 
         interface ICardToken extends ITokenBase {
-            type: "card";
+            type: 'card';
             /**
              * Hash describing the card used to make the charge
              */
@@ -6988,7 +6988,7 @@ declare namespace Stripe {
         }
 
         interface IBankAccountToken extends ITokenBase {
-            type: "bank_account";
+            type: 'bank_account';
             /**
              * Hash describing the bank account
              */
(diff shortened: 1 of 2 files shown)


types/stripe/index.d.ts
```diff
@@ -6977,20 +6977,22 @@ declare namespace Stripe {
     }
 
     namespace tokens {
-        interface IToken extends ICardToken, IBankAccountToken {}
+        type IToken = ICardToken | IBankAccountToken
 
         interface ICardToken extends ITokenBase {
+            type: "card";
```
pushplay: This should be single quotes.  Also there should be an empty line before the next comment.

Same goes for IBankAccountToken.

pushplay (changes requested):
Yup union type is totally the right thing to do here.  Good catch.  Just needs some formatting improvements for consistency and it's good to go.

htunnicliff:
I made formatting changes requested by @pushplay and added some tests as well.

types/stripe/index.d.ts
```diff
@@ -6977,20 +6977,22 @@ declare namespace Stripe {
     }
 
     namespace tokens {
-        interface IToken extends ICardToken, IBankAccountToken {}
+        type IToken = ICardToken | IBankAccountToken;
 
         interface ICardToken extends ITokenBase {
+            type: 'card';
```
pushplay: Still needs a newline here.

pushplay (changes requested):
Thanks for adding the tests!  The newlines still need to be addressed and then it's good to go.

htunnicliff:
Newlines added!

Issue #38342, opened 2019-09-12
[aws-lambda] Handler does not allow a function signature without Callback


duro:
The current definition for a Handler is:

```
export type Handler<TEvent = any, TResult = any> = (
    event: TEvent,
    context: Context,
    callback: Callback<TResult>,
) => void | Promise<TResult>
```

Which correctly notes that handlers can return a promise, but the third argument for Callback is not required if the handler is an Async handler.

Shouldn't this argument be optional?

------

- [X] I tried using the `@types/xxxx` package and had problems.
- [X] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [X] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [X] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @darbio @skarum @StefH @tobyhede @buggy @y13i @wwwy3y3 @OrthoDex @MichaelMarner @daniel-cottone @kostya-misura @coderbyheart @palmithor @daniloraisi @simonbuchan @Haydabase @repl-chris @aneilbaboo @jeznag @louislarry @dpapukchiev @ohookins @trevor-leach @jagregory @dalen @loikg @skyzenr @redlickigrzegorz @juancarbonel

coyoteecd:
@simonbuchan I understand your reasoning, but the casting solution sucks.

That said, I can see AWS cannot modify their API because of backwards compatibility, and the Handler type must reflect that.

On the other hand, what I want - and probably what @duro wants as well - is to be able to say "I do not want the callback, my handler is using the async API". What I did in my project is to declare an AsyncHandler as follows:

```
export type AsyncHandler<TEvent, TResult> = (
  event: TEvent,
  context: Context
) => Promise<TResult>;
```

Then you can use declare your lambda with it:
```
export const handler: AsyncHandler<OrderRequest, APIGatewayProxyResult> = async (event) => {
  // do stuff
}
```

This type is compatible with Handler and does not have the issue you mentioned. I simply declare I do not want the callback, and I guarantee I'm returning a Promise. My function then does not have to bother validation optional parameters, because the parameter is not there at all. Furthermore, if I really need the callback (e.g. ApolloServer project uses it), I can stick to the full Handler definition.

[...]

PR #30820, opened 2018-11-26
[aws-lambda] Make the Handler type callback param optional

J2D2Development:
In the node 8 lambda environment, a `Handler` does not need a callback function, it may return a promise instead.  However, the `callback` parameter in the `Handler` type is not currently optional.

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: <<url here>>
- [ ] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #38828, opened 2019-10-02
The callback of an AWS Lambda handler is optional

jish:
The documentation says that the callback is optional, and if you
provide a handler that only takes two arguments, the lambda runs just
fine.

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/aee73c479bd3f156b1468f3fd7e34811263f9e1a/types/aws-lambda/index.d.ts#L1082

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html

### DIFF
--- a/types/aws-lambda/index.d.ts
+++ b/types/aws-lambda/index.d.ts
@@ -1085,7 +1085,7 @@ export interface LexResult {
 export type Handler<TEvent = any, TResult = any> = (
     event: TEvent,
     context: Context,
-    callback: Callback<TResult>,
+    callback?: Callback<TResult>,
 ) => void | Promise<TResult>;
 
 /**


simonbuchan (changes requested):
Nope, typescript optional parameters are optional for the caller, putting the burden of handling the parameter missing on the implementation.

Dupe of #30820, see also the discussion on #37649.

jish:
Is there anything to be done here?

In the other thread it sounds like there is a workaround, but Definitely Typed doesn’t want to apply it. They want everyone to apply it in their own repos individually.

simonbuchan:
Not sure what you're referring to? The fix is don't write incorrectly typed code?
I actually cover this better in #38342, should have linked that too.

PR #38836, opened 2019-10-03
add typings for @fluent/bundle, @fluent/dedent, @fluent/langneg and @fluent/react

blushingpenguin:
Please fill in this template.

If adding a new definition:


RyanCavanaugh:
Is the listing of authors intentional? @huy-nguyen, @JamesNimlos can you review?

blushingpenguin:
Yes, the packages have been renamed and refactored (the originals are named fluent, fluent-react) and these typings are based on the original typings

huy-nguyen:
It looks like these packages have indeed been namespace-ed under `@fluent`. If all the author did was copying over the definitions from the existing packages I think it should be good to go.

andrewbranch:
I feel like the packages at `fluent-react` etc. should be deleted, but I’m not sure exactly how that should be done... normally when we remove packages, it's because the JS library started shipping its own types. @sandersn any advice?

types/fluent__dedent/index.d.ts
```diff
@@ -0,0 +1,8 @@
+// Type definitions for @fluent/dedent 0.1
+// Project: http://projectfluent.org
+// Definitions by: Mark Weaver <https://github.com/blushingpenguin>, Huy Nguyen <https://github.com/huy-nguyen>, James Nimlos <https://github.com/jamesnimlos>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.3
+
+// tslint:disable-next-line:npm-naming
+export default function ftl(strings: TemplateStringsArray, ...values: any[]): string;
```
andrewbranch: Looks like this shouldn’t be a default export; it’s a UMD wrapping a CommonJS export: https://unpkg.com/browse/@fluent/dedent@0.1.0/index.js

```ts
declare function FluentDedent(/* ... */): string;
export = FluentDedent;
export as namespace FluentDedent;
```

andrewbranch (changes requested):
I noticed most of these libraries have UMD wrappers that aren’t represented here. If people use them as globals, they’d need to add an `export as namespace WhateverTheGlobalNameIs`.

sandersn:
@andrewbranch the right thing is for, eg, fluent-react to be deprecated and point to `@fluent/react`. I *think* the below addition to notNeededPackages might work, although I have not tested it. I haven't looked up asOfVersion, but it needs to be a version that's newer than the currently published `@types/fluent-react` and I think it also has to match a version of `@types/fluent__react`.

```json
        {
            "libraryName": "@types/fluent__react",
            "typingsPackageName": "fluent-react",
            "sourceRepoURL": "https://github.com/DefinitelyTyped/DefinitelyTyped",
            "asOfVersion": "?.?.?"
        },
```

Of course the fluent-react files need to be deleted as well.

types/fluent__dedent/index.d.ts
blushingpenguin: Ok, I see what you mean -- I have fixed the dedent definition and added namespaces to the other modules now.  Also I had originally split the definition files as per the module source which was wrong as the packaged modules are all in one file so I've merged them back into one.
Thanks for the help!

PR #38837, opened 2019-10-03
[react] added missing prop (title) to interface SVGProps

stigelde:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/create-react-app/pull/7118


eps1lon (changes requested):
I think that create-react-app feature is enabled by a webpack plugin. You can use module augmentation to enable this. An example on how to add non-standard props can be found in https://github.com/typescript-cheatsheets/react-typescript-cheatsheet/blob/master/ADVANCED.md#adding-non-standard-attributes

We can't accept this here since we only type standard attributes on native components.

kvalmos:
@eps1lon ok, will try that, thanks

PR #38840, opened 2019-10-03
Add types for sortobject.

SanderDeWaal1992:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/sortobject/index.d.ts
+++ b/types/sortobject/index.d.ts
@@ -1,0 +1,9 @@
+// Type definitions for sortobject 1.2
+// Project: https://github.com/bevry/sortobject
+// Definitions by: Sander de Waal <https://github.com/SanderDeWaal1992>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
+declare function sortObject<T extends object>(obj: T, comparator?: (a: string, b: string) => number): T;
+
+export = sortObject;
--- a/types/sortobject/sortobject-tests.ts
+++ b/types/sortobject/sortobject-tests.ts
@@ -1,0 +1,11 @@
+import sortobject = require("sortobject");
+
+interface ObjectType {
+	some: string;
+}
+
+const sortedObject1: ObjectType = sortobject<ObjectType>({ some: "object" });
+const sortedObject2: ObjectType = sortobject({ some: "object" });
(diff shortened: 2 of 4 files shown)


RyanCavanaugh:
The project attempts to ship its own types: https://github.com/bevry/sortobject#typescript

Maybe you can send them a PR to produce a real .d.ts and include it in the package? Writing a separate .d.ts when the authors have already tried to do seems bad. cc @bevry

SanderDeWaal1992:
> 
> 
> The project attempts to ship its own types: https://github.com/bevry/sortobject#typescript
> 
> Maybe you can send them a PR to produce a real .d.ts and include it in the package? Writing a separate .d.ts when the authors have already tried to do seems bad. cc @bevry

They do suggest to set maxNodeModuleJsDepth to 5, which only works if allowJs is set. Both options are not an option for many projects (including ours).

I don't see a sign about them attempting to add TS typings anywhere and personally I don't like packages that are not written in TS but have TS typings files (and often forget to update typings or even care about them after adding them). DefinitelyTyped is there specifically for that purpose as far as I know and to my opinion it works a lot better than depending on package owners.

Anyways, I believe this addition to DefinitelyTyped has value.

armanio123:
I think what @RyanCavanaugh is trying to say is that the developer of the library has already a solution for providing type information and it is not DefinitelyTyped purpose to compete with authors.

A better approach would be to submit a PR to the library to improve their types or start a discussion with them to about the challenges you're facing with their solution.

Let me know if I'm misinterpreting something @RyanCavanaugh .

andrewbranch:
> I don't see a sign about them attempting to add TS typings anywhere or have I just missed that?

Well, they have JSDoc annotations, which are not exactly TypeScript definitions, but TypeScript does understand them. But I don’t think suggesting people change their maxNodeModuleJsDepth is a good solution.

In TypeScript 3.7 (currently in beta), they could generate declaration files from their JSDoc if they wanted to. Since they seem to have an interest in maintaining good JSDoc annotations and supporting TypeScript with them, that’s probably the best solution. They would need to add TypeScript as a devDependency, add a tsconfig.json file, and run `tsc --declaration -p .` as part of their build process.

If they’re opposed to doing that, then adding the typings here on DefinitelyTyped is totally reasonable.

weswigham:
I think this is perfectly fine until they ship _declaration files_ themselves. Plus, these types are better than their doc-based ones right now, too - they have a generic parameter to reflect how the input and output are related.

PR #38842, opened 2019-10-03
[react-with-styles] Make ThemedStyleSheet an importable value

mohsen1:
A very small fix

### DIFF
--- a/types/react-with-styles/lib/ThemedStyleSheet.d.ts
+++ b/types/react-with-styles/lib/ThemedStyleSheet.d.ts
@@ -11,4 +11,5 @@ declare const ThemedStyleSheet: {
   flush: () => void;
 };
 
+export type ThemedStyleSheet = typeof ThemedStyleSheet;
 export default ThemedStyleSheet;
--- a/types/react-with-styles/lib/ThemedStyleSheet.d.ts
+++ b/types/react-with-styles/lib/ThemedStyleSheet.d.ts
@@ -1,4 +1,4 @@
-interface ThemedStyleSheet {
+declare const ThemedStyleSheet: {
   registerTheme: (theme: any) => void;
   registerInterface: (registerInterface: any) => void;
   create: (makeFromTheme: any, createWithDirection: any) => any;
--- a/types/react-with-styles/lib/ThemedStyleSheet.d.ts
+++ b/types/react-with-styles/lib/ThemedStyleSheet.d.ts
@@ -9,6 +9,6 @@ declare const ThemedStyleSheet: {
   resolveLTR: () => any;
   resolveRTL: () => any;
   flush: () => void;
-}
+};
 
 export default ThemedStyleSheet;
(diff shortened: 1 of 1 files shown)


RyanCavanaugh:
Do you want to add `type ThemedStyleSheet = typeof ThemedStyleSheet;` so that this doesn't break anyone?

types/react-with-styles/lib/ThemedStyleSheet.d.ts
```diff
@@ -1,4 +1,4 @@
-interface ThemedStyleSheet {
+declare const ThemedStyleSheet: {
```
brieb: 👍 corresponds to https://github.com/airbnb/react-with-styles/blob/61b02293f7d91cdc5fe9f5d338cc30fc21e4325b/src/ThemedStyleSheet.js#L95-L106

types/react-with-styles/lib/ThemedStyleSheet.d.ts
```diff
@@ -1,4 +1,4 @@
...
   registerTheme: (theme: any) => void;
   registerInterface: (registerInterface: any) => void;
   create: (makeFromTheme: any, createWithDirection: any) => any;
```
brieb: This gets exported as `create: createLTR`
https://github.com/airbnb/react-with-styles/blob/61b02293f7d91cdc5fe9f5d338cc30fc21e4325b/src/ThemedStyleSheet.js#L98

So the type should be `create: (makeFromTheme: any) => any;`

types/react-with-styles/lib/ThemedStyleSheet.d.ts
```diff
@@ -9,6 +9,7 @@ interface ThemedStyleSheet {
   resolveLTR: () => any;
   resolveRTL: () => any;
   flush: () => void;
-}
+};
 
+export type ThemedStyleSheet = typeof ThemedStyleSheet;
 export default ThemedStyleSheet;
```
brieb: Seems odd that this would be exported as a named and default export.
User of value would do: `import ThemedStyleSheet from '...';`
User of type would do: `import { ThemedStyleSheet } from '...';`

Should we not export the type in this case, and users of the type could do the `typeof ThemedStyleSheet`? 

Thinking longer-term, if this file were written in TS with JS & DTS emitted, it wouldn't get a type export like this anyways.

brieb: Or if the type is commonly used enough we could export it under a different name

mohsen1: This was done based on Ryan's [suggestion](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38842#issuecomment-538157377) for backward compatibility

lencioni (approved):

armanio123:
@brieb, is this good or you still want a comment to be addressed?

types/react-with-styles/lib/ThemedStyleSheet.d.ts
brieb: Ah. Good point. I missed that, thanks!

brieb (approved):

brieb:
LGTM

PR #38846, opened 2019-10-03
[i18n-js] Add missing getFullScope static method to types definition

elyalvarado:
`getFullScope` static method was missing from the types definition

Please fill in this template.

- [x ] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/fnando/i18n-js/blob/dc76dcb1b5d7c0e3e50099c1d94c63a5fc048c1d/app/assets/javascripts/i18n.js#L401

### DIFF
--- a/types/i18n-js/i18n-js-tests.ts
+++ b/types/i18n-js/i18n-js-tests.ts
@@ -17,6 +17,8 @@ I18n.t("noun", { defaultValue: "I'm a {{noun}}", noun: "Mac" });
 I18n.t("some.missing.scope", { defaults: [{ scope: "some.existing.scope" }] });
 I18n.t("some.missing.scope", { defaults: [{ message: "Some message" }] });
 
+I18n.getFullScope("translation", { scope: "some.scoped" });
+
 I18n.fallbacks = true;
 I18n.locales.no = ["nb", "en"];
 I18n.locales.no = "nb";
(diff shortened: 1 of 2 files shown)


types/i18n-js/i18n-js-tests.ts
```diff
@@ -17,6 +17,8 @@ I18n.t("noun", { defaultValue: "I'm a {{noun}}", noun: "Mac" });
 I18n.t("some.missing.scope", { defaults: [{ scope: "some.existing.scope" }] });
 I18n.t("some.missing.scope", { defaults: [{ message: "Some message" }] });
 
+I18n.getFullScope("translation", { scope: "some.scoped" });
```
ypresto: Unfortunately, it looks undocumented internal method, which should not be appear in d.ts .

ypresto (changes requested):
It seems an undocumented internal method (`like lookup()`)

ypresto (changes requested):
(Please see previous comment)

ypresto (changes requested):
Trying to mark as request changes but it seems not working...

types/i18n-js/i18n-js-tests.ts
elyalvarado: It's true it is undocumented, however is exposed and unlike `lookup` is not clearly marked as for internal use only in the source code. I use for example in my `missingTranslation` callback to log the actual scope.

elyalvarado:
It's true it is undocumented, however is exposed and unlike lookup is not clearly marked as for internal use only in the source code. I use for example in my missingTranslation callback to log the actual scope.

ypresto:
> I use for example in my missingTranslation callback to log the actual scope.

Ah, I see, it is calles in default missingTranslation handler. I'll see this later. (Maybe it'd be better to pass full scope from the library.)

PR #38849, opened 2019-10-03
Update react-virtualized-select index.d.ts for `optionHeight` prop

epzilla:
`optionHeight` can take a number or a function that returns a number, so I've updated the type for that prop

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/bvaughn/react-virtualized-select#react-virtualized-select-props

### DIFF
--- a/types/react-virtualized-select/index.d.ts
+++ b/types/react-virtualized-select/index.d.ts
@@ -24,7 +24,7 @@ export interface VirtualizedOptionRenderOptions<T> {
 
 export interface AdditionalVirtualizedSelectProps<TValue> {
     maxHeight?: number;
-    optionHeight?: number | (({ option: any }) => number);
+    optionHeight?: number | ((option) => number);
     optionRenderer?(options: VirtualizedOptionRenderOptions<TValue>): JSX.Element;
     selectComponent?: React.ComponentClass<any> | React.StatelessComponent<any>;
 }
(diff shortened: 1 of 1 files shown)


types/react-virtualized-select/index.d.ts
```diff
@@ -24,7 +24,7 @@ export interface VirtualizedOptionRenderOptions<T> {
 
 export interface AdditionalVirtualizedSelectProps<TValue> {
     maxHeight?: number;
-    optionHeight?: number;
+    optionHeight?: number | ((option: TValue) => number);
```
seansfkelley: Should this actually be `({ option: TValue }) => number`? The docs you linked suggest so, but I'm not sure what `TValue` generally is (is it an object with an `option` field?) as I haven't used this library in years.

epzilla: Yes, you're right. Good catch, thanks! Fixed it.

epzilla: 😕but now the build is failing after that change. Not really sure what to do.

seansfkelley: > Binding element 'TValue' implicitly has an 'any' type

I guess it's asking to name the `{ option: TValue }` type? Maybe you could introduce `interface WrappedValue<TValue>` and use that instead of an inline type?

seansfkelley: I just ran into this same pattern in my own code. The problem is just that it wants an argument name: `(options: { option: TValue }) => ...`.

epzilla: Thanks @seansfkelley! I went ahead and made that change.

seansfkelley (approved):

PR #38850, opened 2019-10-03
fix(wellknow): parse result

igrzhukovich:
fix Type definitions for wellknown
add return object for parse function

### DIFF
--- a/types/wellknown/index.d.ts
+++ b/types/wellknown/index.d.ts
@@ -3,5 +3,8 @@
 // Definitions by: Yair Tawil <https://github.com/yairtawil>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export function parse(input: string): {};
+export function parse(input: string): {
+    type: string,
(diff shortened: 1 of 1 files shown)


types/wellknown/index.d.ts
```diff
@@ -3,5 +3,8 @@
 // Definitions by: Yair Tawil <https://github.com/yairtawil>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export function parse(input: string): {};
+export function parse(input: string): {
```
armanio123: I'm not very familiar with the library.  Doing a quick look in the documentation seems like this method returns a "geometry object" which seem to have different `coordinates` properties based on the type.

Could you confirm this is the correct change? Also as maintainers we appreciate if you could point on the description of the PR the documentation that proves the change.

andrewbranch (changes requested):
Please use the PR template in the future—it has, among other things, a checkbox that instructs you to include a link to the API/documentation that justifies this change, and we can't effectively review without that information.

PR #38851, opened 2019-10-03
Adds missing importance attribute in LinkHTMLAttributes

jigz:
## Changes
* Adds missing `importance` to LinkHTMLAttributes

## Screenshot
Screen Shot 2019-10-04 at 2 14 58 AM

## References
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link
* https://github.com/zeit/next.js/pull/8878

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -2001,6 +2001,7 @@ declare namespace React {
         crossOrigin?: string;
         href?: string;
         hrefLang?: string;
+        importance?: string;             
         integrity?: string;
         media?: string;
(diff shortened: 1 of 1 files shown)


eps1lon (changes requested):
We're only including standard attributes. I could not find any spec that defines it nor implementation status.

If this attribute is polyfilled by `next` you can patch the types locally following https://github.com/typescript-cheatsheets/react-typescript-cheatsheet/blob/master/ADVANCED.md#adding-non-standard-attributes

PR #38855, opened 2019-10-03
[js-cookie] Update CookieAttributes definition to accept SameSite and freeform attributes

### LINKED ISSUE #38798: @types/js-cookie CookieAttributes interface should be kept open
@theodorejb @BendingBender @alepee @yutod 

The `CookieAttributes` interface only accepts `expires`, `path`, `domain` and `secure` properties.
This should be kept open so cookies like this can be emmited:

```ts
Cookies.set('_test', 'value', { secure: true, samesite: 'lax' });
```
Proposed fix:

```ts
interface CookieAttributes {
        expires?: number | Date;
        path?: string;
        domain?: string;
        secure?: boolean;
        [otherProperty: string]: any;
}
```

nreynis:
Fix #38798 
`CookieAttributes` definition should be kept open. The underlying code isn't restricted to known attributes, it will serialize any attribute to the best of it's capacity:
https://github.com/js-cookie/js-cookie/blob/master/src/js.cookie.mjs#L51

I also added definitions for the `SameSite` attribute:
https://tools.ietf.org/html/draft-west-first-party-cookies-07
https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#SameSite_cookies
https://web.dev/samesite-cookies-explained
https://caniuse.com/#feat=same-site-cookie-attribute

### DIFF
--- a/types/js-cookie/index.d.ts
+++ b/types/js-cookie/index.d.ts
@@ -4,6 +4,7 @@
 //                 BendingBender <https://github.com/BendingBender>
 //                 Antoine Lépée <https://github.com/alepee>
 //                 Yuto Doi <https://github.com/yutod>
+//                 Nicolas Reynis <https://github.com/nreynis>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
@@ -32,6 +33,18 @@ declare namespace Cookies {
          * secure protocol (https). Defaults to false.
          */
         secure?: boolean;
+
+        /**
+         * Define the first party cookie policy.
+         * @see: https://tools.ietf.org/html/draft-west-first-party-cookies-07
+         */
+        samesite?: 'Strict' | 'Lax';
+
+        /**
+         * An attribute which will be serialized, conformably to RFC 6265
+         * section 5.2.
+         */
+        [property: string]: any;
     }
 
     interface CookiesStatic<T extends object = object> {
--- a/types/js-cookie/index.d.ts
+++ b/types/js-cookie/index.d.ts
@@ -35,10 +35,11 @@ declare namespace Cookies {
         secure?: boolean;
 
         /**
-         * Define the first party cookie policy.
-         * @see: https://tools.ietf.org/html/draft-west-first-party-cookies-07
+         * Asserts that a cookie must not be sent with cross-origin requests,
+         * providing some protection against cross-site request forgery
+         * attacks (CSRF)
          */
-        samesite?: 'Strict' | 'Lax';
+        sameSite?: 'Strict' | 'Lax' | 'None';
 
         /**
          * An attribute which will be serialized, conformably to RFC 6265
--- a/types/js-cookie/js-cookie-tests.ts
+++ b/types/js-cookie/js-cookie-tests.ts
@@ -9,6 +9,8 @@ Cookies.set('name', 'value', { expires: 7, path: '', domain: '', secure: true })
 Cookies.set('name', 'value', { secure: true });
 Cookies.set('name', 'value', { domain: '' });
 Cookies.set('name', 'value', { path: '' });
+Cookies.set('name', 'value', { samesite: 'Strict' });
+Cookies.set('name', 'value', { custom: 'property' });
 
 // $ExpectType string | undefined
 Cookies.get('name');
--- a/types/js-cookie/js-cookie-tests.ts
+++ b/types/js-cookie/js-cookie-tests.ts
@@ -9,7 +9,7 @@ Cookies.set('name', 'value', { expires: 7, path: '', domain: '', secure: true })
 Cookies.set('name', 'value', { secure: true });
 Cookies.set('name', 'value', { domain: '' });
 Cookies.set('name', 'value', { path: '' });
-Cookies.set('name', 'value', { samesite: 'Strict' });
+Cookies.set('name', 'value', { sameSite: 'Strict' });
 Cookies.set('name', 'value', { custom: 'property' });
 
 // $ExpectType string | undefined


types/js-cookie/index.d.ts
```diff
@@ -32,6 +33,18 @@ declare namespace Cookies {
          * secure protocol (https). Defaults to false.
          */
         secure?: boolean;
+
+        /**
+         * Define the first party cookie policy.
+         * @see: https://tools.ietf.org/html/draft-west-first-party-cookies-07
+         */
+        samesite?: 'Strict' | 'Lax';
```
theodorejb: Consider naming the property `sameSite`, since in the spec it is two words (the official attribute name is "SameSite").

theodorejb: The property should support a value of "None" in addition to Strict and Lax.

types/js-cookie/index.d.ts
```diff
@@ -32,6 +33,18 @@ declare namespace Cookies {
...
+
+        /**
+         * An attribute which will be serialized, conformably to RFC 6265
+         * section 5.2.
+         */
+        [property: string]: any;
```
theodorejb: I'm not sure it's a good idea to add this. What is the use case? Wouldn't it prevent catching mistakes when one of the supported attributes is mistyped?

theodorejb (changes requested):

types/js-cookie/index.d.ts
```diff
@@ -32,6 +33,18 @@ declare namespace Cookies {
          * secure protocol (https). Defaults to false.
          */
         secure?: boolean;
+
+        /**
+         * Define the first party cookie policy.
+         * @see: https://tools.ietf.org/html/draft-west-first-party-cookies-07
```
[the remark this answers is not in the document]
nreynis: Good point,
I could link to both: the MDN page which is more developer friendly and the current RFC version.

https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#page-42

Also the current RFC and MDN says that 'None' is a possible value, while previous draft had a BNF which defined 'Strict' and 'Lax' only ('None' was the default value when the attribute didn't exist and couldn't be passed explicitly).

types/js-cookie/index.d.ts
nreynis: Agreed for the property name, I asked myself the same question and went for lowercase because `secure` was used instead of `Secure`. CamelCase would make more sense though.

For `None` you're right, I was looking at an outdated RFC's BNF (https://tools.ietf.org/html/draft-west-first-party-cookies-07#section-3.1)

types/js-cookie/index.d.ts
nreynis: I don't think it's the role of types definition to validate or constraint values. Sure adding known standardized attributes is a nice added value, but you also have to accuratly represent the code ability.

Right now the library is compatible with `SameSite` and also with other attributes who may be added in the future. The only thing breaking it is too restrictive type definition.

If in a few years navigators decide to add let's say `SpdyOnly` or whatever, the library will be fine, types definition should also.

nreynis: Restricting the properties is an API design decision that should be made in the library. Types definition, IMHO, should be about describing and providing type safety, not adding our own opinions on API design.

theodorejb: Normally I'd agree about it being a library decision, but in this case it's more an implementation detail of the library, and the API design decision is ultimately made by the HTTP Cookie specification implemented by browsers. If some other attribute is hypothetically added in the future, the type definitions should be updated to correctly suggest and constrain this property (just as is being done now for SameSite).

nreynis: I'm sorry to insist, but it's not an implementation detail, it's by design.

1. The RFC has the grammar for it (https://tools.ietf.org/html/rfc6265#section-4.1.1):
```ABNF
 cookie-av         = expires-av / max-age-av / domain-av /
                     path-av / secure-av / httponly-av /
                     extension-av
 expires-av        = "Expires=" sane-cookie-date
 max-age-av        = "Max-Age=" non-zero-digit *DIGIT
 domain-av         = "Domain=" domain-value
 path-av           = "Path=" path-value
 secure-av         = "Secure"
 httponly-av       = "HttpOnly"
 extension-av      = <any CHAR except CTLs or ";">
```
2. There's an issue for it (https://github.com/js-cookie/js-cookie/issues/276)
3. And also a merged PR complete with explicit unit test for unofficial attributes (https://github.com/js-cookie/js-cookie/pull/293)

theodorejb: Okay, I guess it makes sense then. Shouldn't the value at least have a type of `string`, though, instead of `any`?

nreynis: Well you can pass:
* `string` which would be used as is
* `number` which would be casted to string
* boolean `true`, the attribute will be printed but won't have value (the loop will hit the second `continue`)

Other values can technically also be passed but don't really make sense:
* boolean `false`, will skip the attribute (the loop will hit the first `continue`)
* `array` and `function` will be casted to string

Kinda grey area:
* `object` could potentially make sense if the object implements `toString`

theodorejb: If the property can only hold a string value, I think it should be declared as such. Devs can call `toString` themselves on objects/numbers if they want to store them. Otherwise it's confusing if you get a different value out than you put in.

nreynis: So `string | number | boolean` ?

theodorejb: No, only `string` (just like the value parameter on the `set` method has a string type, even though technically in JavaScript you could pass other things to it which would be cast to a string).

nreynis: Accepting only `string` won't do the trick because you need at least `string | boolean` to declare value-less attributes.

Also I just tested it and the generic property declaration have to at least be compatible with all other attributes declaration or else other attributes will throw `ts(2411)` errors.
So the bare minimum is `[property: string]: string | number | boolean | Date | undefined;`... might as well keep `any`

https://www.typescriptlang.org/docs/handbook/interfaces.html#indexable-types

theodorejb (approved):

types/js-cookie/index.d.ts
nreynis: Thank you for your time @theodorejb

PR #38856, opened 2019-10-03
Add type definitions for features added between Dinero.js v1.4 and v1.6.

JorgenVatle:
This pull request addresses features that were added between [Dinero.js](https://github.com/sarahdayan/dinero.js) versions v1.4 and v1.6.
Namely the addition of a [`toJson()` alias](https://github.com/sarahdayan/dinero.js/releases/tag/v1.5.0) for instances of Dinero and [support for promises](https://github.com/sarahdayan/dinero.js/releases/tag/v1.6.0) in the `ExchangeRatesApiOptions` interface.

This should not conflict with the changes made in #38198.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: (see above)


types/dinero.js/index.d.ts
```diff
@@ -72,7 +73,7 @@ declare namespace DineroFactory {
         | 'HALF_AWAY_FROM_ZERO';
 
     interface ExchangeRatesApiOptions {
-        endpoint: string;
+        endpoint: string | Promise<{[key: string]: any}>;
```
andrewbranch: Should it not be `string | Promise<{ rates: { [K in Currency]?: number } }>`?

andrewbranch (approved):

types/dinero.js/index.d.ts
JorgenVatle: The promise can resolve to any object. It does not have any constraints apart from being possible to navigate it using a key path that may or may not include a `{{to}}` or `{{from}}` currency.

The Dinero docs define [`propertyPath`](https://sarahdayan.github.io/dinero.js/module-Dinero.html#~convert) which is used to navigate the resolved object. This allows for use of differently structured exchange rate objects, which may have actual exchange rates deeply nested.

Can't really think of a way to make type stricter than it already is without breaking usage of nested exchange rate objects, like the structure used in [`dinero.js-tests.ts`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f573096ba00fea104202310253b1798586040d95/types/dinero.js/dinero.js-tests.ts#L26). 🤔

JorgenVatle: If you have any ideas, I'd be happy to refactor accordingly! 🤗

andrewbranch: Sounds good, I just only saw one type structure from quickly looking at examples so I was wondering if it could be improved. There’s probably not much we can do with dynamic stuff like `propertyPath`.

andrewbranch (approved):

PR #38858, opened 2019-10-03
fix(pino): DestinationStream can be any object with write()`

ethanresnick:
This is per the [pino docs](https://github.com/pinojs/pino/blob/master/docs/api.md#destination-sonicboom--writablestream--string):

> The destination parameter, at a minimum must be an object with a write method. 

Inspecting the source confirms that only a write method (with no particular return type) is required for basic usage: https://github.com/pinojs/pino/blob/master/lib/proto.js#L153. Calls to other potential methods (flush, flushSync, etc) are always guarded with an existence check, as they're not required.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/pinojs/pino/blob/master/docs/api.md#destination-sonicboom--writablestream--string

### DIFF
--- a/types/pino/index.d.ts
+++ b/types/pino/index.d.ts
@@ -171,7 +171,10 @@ declare namespace P {
         labels: { [level: number]: string; };
     }
     type TimeFn = () => string;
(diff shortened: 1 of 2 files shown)


lummish (approved):
LGTM

PR #38863, opened 2019-10-04
Add types for object-assign-deep

nickbclifford:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/object-assign-deep/index.d.ts
+++ b/types/object-assign-deep/index.d.ts
@@ -1,0 +1,22 @@
+// Type definitions for object-assign-deep 0.4
+// Project: https://github.com/saikojosh/Object-Assign-Deep#readme
+// Definitions by: Nick Clifford <https://github.com/nickbclifford>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.1
+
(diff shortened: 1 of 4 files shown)


types/object-assign-deep/index.d.ts
```diff
@@ -0,0 +1,22 @@
+// Type definitions for object-assign-deep 0.4
+// Project: https://github.com/saikojosh/Object-Assign-Deep#readme
+// Definitions by: Nick Clifford <https://github.com/nickbclifford>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.1
+
+declare const objectAssignDeep: ObjectConstructor['assign'];
```
armanio123: I'm not sure what are you trying to achieve here. Isn't this a `function` with n parameters or am I missing something?

nickbclifford: Yes, but it effectively has the same API as `Object.assign`, so I figured that this might be cleaner than rewriting those types. This is my first new definition and I wasn't really sure about the convention, so I went with my gut on that.

armanio123: Makes sense. Thanks for clarifying.

armanio123 (approved):

PR #38869, opened 2019-10-04
change export declaration to avoid esModuleInterop

stefdelec:
Change declaration of export to avoid use of esModuleInterop in tsconfig of a project using bn.js

### DIFF
--- a/types/bn.js/index.d.ts
+++ b/types/bn.js/index.d.ts
@@ -581,4 +581,4 @@ declare class RedBN extends BN {
     redPow(b: BN): RedBN;
 }
 
-export = BN;
+export default BN;


ExE-Boss (changes requested):
See [FAQ](https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default) for why this is not correct and should be closed.

HenryNguyen5 (changes requested):
As stated before, the current export style is correct and shouldn’t change

stefdelec:
My issue I have, and it happens only with BN.js, is this:

Return type of public method from exported class has or is using name 'BN' from external module "/mypath/node_modules/@types/bn.js/index" but cannot be named.ts(4053)

PR #38876, opened 2019-10-04
[@types/async] Fix async.constant return type

gabmontes:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/caolan/async/blob/v3.1.0/lib/constant.js#L13

### DIFF
--- a/types/async/index.d.ts
+++ b/types/async/index.d.ts
@@ -259,7 +259,7 @@ export function race<T, E = Error>(tasks: Array<AsyncFunction<T, E>>, callback: 
 export function memoize(fn: Function, hasher?: Function): Function;
 export function unmemoize(fn: Function): Function;
 export function ensureAsync(fn: (... argsAndCallback: any[]) => void): Function;
-export function constant(...values: any[]): Function;
+export function constant(...values: any[]): AsyncFunction<any>;
 export function asyncify(fn: Function): (...args: any[]) => any;
 export function wrapSync(fn: Function): Function;
 export function log(fn: Function, ...args: any[]): void;


alloy (commented):
I’m unfamiliar with this library, but I did a quick check. While [the docs](https://github.com/caolan/async/blob/f9ad468f7b464f5ef88116def2399a5c6e59e21d/lib/constant.js#L13) state the return type is an `AsyncFunction`, [the returned function](https://github.com/caolan/async/blob/f9ad468f7b464f5ef88116def2399a5c6e59e21d/lib/constant.js#L44-L47) is not technically of type `AsyncFunction` as it isn’t tagged with the `async` keyword. Perhaps it’s something like `() => Promise<any>`, however looking at [the tests](https://github.com/caolan/async/blob/master/test/constant.js) it appears to be a regular function that is invoked at a later time?

I don’t think I can assist any further, but naively it seems to me that the docs are incorrect here 🤷‍♂  

_Btw, with this function being about constants, it seems like the typing of the function could easily [be improved to reflect the inputs](https://www.typescriptlang.org/play/index.html?ssl=1&ssc=1&pln=5&pc=31#code/GYVwdgxgLglg9mABMOcA8AVRBTAHlbMAEwGdEBBAJ0oEMBPNGsOgPhYAoA6bmygcxIAuRBgCUwrAG8AUIjmJK2KCEpJeA6QF9p0iAhJREuRAF5kqdgBYATABpEAIhRwH9qJRDZRQA)._

gabmontes:
@alloy there library is about managing control flow when you have synchronous functions that use callbacks. It has nothing to do with promises or async/await functions. So yes, the function returned is a plain function that is invoked later as you mentioned. In the library, the type is called “AsyncFunction” and represents a function that receives a callback as its last parameter. And the function is not about constants. It is about generating a function that receives a constant and a callback last and invoked the callback with the constant received. So the change is correct. Please read the docs Abd familiarize with the library so you can eventually assess the change again. Thanks for the comments anyway.

andrewbranch (approved):

types/async/index.d.ts
```diff
@@ -259,7 +259,7 @@ export function race<T, E = Error>(tasks: Array<AsyncFunction<T, E>>, callback:
 export function memoize(fn: Function, hasher?: Function): Function;
 export function unmemoize(fn: Function): Function;
 export function ensureAsync(fn: (... argsAndCallback: any[]) => void): Function;
-export function constant(...values: any[]): Function;
+export function constant(...values: any[]): AsyncFunction<any>;
```
andrewbranch: Actually, it looks like
```ts
export function constant<T>(...values: T[]): AsyncFunctionEx<T>;
```
would be an improvement?

alloy: Ahh, the `Ex` suffix here put me on the right track to notice that `AsyncFunction` was being defined locally, rather than being the canonical `AsyncFunction` type:

```
> async () => {}
[AsyncFunction]
```

gabmontes: @andrewbranch thanks for the observation. In fact, `constant()` can receive arguments of any type. I don't think assuming those will be all of the same type is correct.

andrewbranch: In that case, the well-typed version would be

```ts
interface AsyncFunctionVariadic<T extends any[], E = Error> {
  (callback: (err?: E, ...params: T) => void): void;
}

export function constant<T extends any[]>(...values: T): AsyncFunctionVariadic<T>;
```

However, it would require bumping the minimum TypeScript version to 3.0 (currently it’s set to 2.3). I’m not an `async` user so I don’t have the context to know whether that tradeoff is worth it. It seems that this library could really benefit from these features (e.g. `compose`, `seq`, `applyEach` all suffer from `any`-infected typings that could be fixed with generic tuples like this). But of course, this is outside the scope of what you originally came here to do.

If you’re happy with the status quo here, I’ll go ahead and merge.

gabmontes: @andrewbranch I think we should go ahead and merge.

Any further optimization like the one you mentioned would be great but should be addressed in conjunction with the `async` authors/maintainers IMO.

andrewbranch: I was wondering if you would be inspired to become a maintainer, since none of the maintainers showed up to review your PR 😄

PR #38877, opened 2019-10-04
Add node-easy-cert typings

SomaticIT:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/node-easy-cert/node-easy-cert-tests.ts
+++ b/types/node-easy-cert/node-easy-cert-tests.ts
@@ -1,0 +1,82 @@
+import * as CertManager from 'node-easy-cert';
+
+const options = {
+    rootDirPath: '/the/full/path/of/the/dir',
+    defaultCertAttrs: [
+        { name: 'countryName', value: 'CN' },
+        { name: 'organizationName', value: 'CertManager' },
+        { shortName: 'ST', value: 'SH' },
+        { shortName: 'OU', value: 'CertManager SSL' },
+    ],
+};
+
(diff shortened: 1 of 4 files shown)


types/node-easy-cert/tslint.json
```diff
@@ -0,0 +1,4 @@
+{
+    "extends": "dtslint/dt.json",
+    "strict-export-declare-modifiers": true
```
armanio123: I believe this is true by default. You can remove this line.

types/node-easy-cert/node-easy-cert-tests.ts
```diff
@@ -0,0 +1,82 @@
+import * as CertManager from 'node-easy-cert';
```
armanio123: `CertManager` is being imported as namespace which is not callable. Instead use `import CertManager = require('node-easy-cert');`

armanio123 (changes requested):

types/node-easy-cert/tslint.json
SomaticIT: Fixed in last commit

types/node-easy-cert/node-easy-cert-tests.ts
SomaticIT: I'm not sure that it was needed since `CertManager` is a mix of a function, a class and a namespace.
Maybe in old TypeScript versions...

To avoid confusion, I fixed this in last commit.

armanio123 (approved):

Issue #38295, opened 2019-09-11
Leaflet.js types totally outdated for version 1.5.1


homer-jay:
## Description
I tried using the last version of `@types/leaflet` and found the following problems:
*  Property 'attributionControl' does not exist on type 'Map'
*  Property 'latlng' does not exist on type 'LeafletEvent'
*  Issues with L.polygon: Type 'number[]' is not assignable to type 'LatLngExpression[][]'.
*  'maximunAge' does not exist in type 'LocateOptions'

Clearly some of the types are outdated and do not match the current JS API.

##Authors: 
@alejo90
@atd-schubert
@mcauer
@ronikar
@sanfrisc
@life777

life777:
Hi, @homer-jay 
Thanks. 

1) I couldn't find 'attributionControl' property in [Map class](https://github.com/Leaflet/Leaflet/blob/master/src/map/Map.js)
This property can be found in [MapOptions interface](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/leaflet/index.d.ts#L1016)

2)  LeafletEvent is a parent class interface. It's extended by different event Interfaces. Some of them don't have 'latlng' property. Example: LeafletMouseEvent has latlng property, but LeafletKeyboardEvent doesn't have it because there is no need for it.
![Screenshot from 2019-09-11 15-31-32](https://user-images.githubusercontent.com/891497/64728755-b56ea680-d4a9-11e9-8e2e-288e7b928ff9.png)

3) L.polygon accepts `latlngs: LatLngExpression[] | LatLngExpression[][] | LatLngExpression[][][]` as first parameter.
`LatLngExpression` is `LatLng | LatLngLiteral | LatLngTuple`.
It means that L.polygon accepts either `[number, number][]` or { lat: number; lng: number }[] or LatLng[]
Neither of these types is compatible with number[]. 

4) As far as I can see it's there: [link to type definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/leaflet/index.d.ts#L1263)
![Screenshot from 2019-09-11 15-43-14](https://user-images.githubusercontent.com/891497/64729334-fa470d00-d4aa-11e9-9de7-ed315576a18d.png)

homer-jay:
@life777 

[...]

PR #38885, opened 2019-10-04
[leaflet] match event types with appropriate handler function types

alejo90:
I went over the leaflet documentation and matched event `type` parameters in `Evented` methods with the appropriate event handler type. The change is non-breaking for the overloads that take `type` and `fn` handler. The overloads that take a handler map, however, may break if an event type not defined in leaflet is used. The reason for that is the handler map is an interface and we can't add a `[event: string]: any` declaration without making it impossible to type check. There may be something we could do to prevent that. If someone has an idea, please reach out.
In addition, I had to change the `leaflet-draw` definition to make the events it declares implement `LeafletEvent`. Although it's not clear from the documentation or source that all events inherit from `LeafletEvent`, you can see that it's effectively what happens when Leaflet dispatches events in [`Events.js`](https://github.com/Leaflet/Leaflet/blob/19c8e077ceccf487006e2d84a8db4f13e3991418/src/core/Events.js#L174):

```typescript
var event = Util.extend({}, data, {
	type: type,
	target: this,
	sourceTarget: data && data.sourceTarget || this
});
```

Addresses one of the concerns in #38295 

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://leafletjs.com/reference-1.5.0.html#evented

### DIFF
--- a/types/leaflet/index.d.ts
+++ b/types/leaflet/index.d.ts
@@ -222,10 +222,103 @@ export function bounds(topLeft: PointExpression, bottomRight: PointExpression): 
 
 export function bounds(points: Point[] | BoundsLiteral): Bounds;
 
+// Event handler types
+
 export type LeafletEventHandlerFn = (event: LeafletEvent) => void;
 
+export type LayersControlEventHandlerFn = (event: LayersControlEvent) => void;
+
+export type LayerEventHandlerFn = (event: LayerEvent) => void;
+
+export type ResizeEventHandlerFn = (event: ResizeEvent) => void;
+
+export type PopupEventHandlerFn = (event: PopupEvent) => void;
+
+export type TooltipEventHandlerFn = (event: TooltipEvent) => void;
+
+export type ErrorEventHandlerFn = (event: ErrorEvent) => void;
+
+export type LocationEventHandlerFn = (event: LocationEvent) => void;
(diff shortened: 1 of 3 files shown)


types/leaflet/index.d.ts
```diff
@@ -241,33 +334,99 @@ export abstract class Evented extends Class {
      * keyword will point to). You can also pass several space-separated types
      * (e.g. 'click dblclick').
      */
+    // tslint:disable:unified-signatures
```
Favna: Hm not sure how I feel about this to be honest

favna (commented):
I want to say it looks alright to me but then you have also clearly delved into Leaflet far more than I have myself so I lack confidence to give the green light.

types/leaflet/index.d.ts
alejo90: @Favna unfortunately, it's necessary. Otherwise it pushes to merge the `type, fn, ctx?` overloads with the `eventMap` overload, which would be inaccurate because `eventMap` overloads do not take optional `ctx`.

favna (commented):
As said before, LGTM. But approving now since no one else approved it since.

favna (approved):
Woops wrong option just before

PR #38886, opened 2019-10-04
pdfjs-dist: Update some definitions to match implementation in v2.1.266

echo99:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/mozilla/pdf.js/blob/v2.1.266/src/display/api.js>

### DIFF
--- a/types/pdfjs-dist/index.d.ts
+++ b/types/pdfjs-dist/index.d.ts
@@ -241,8 +241,15 @@ interface PDFProgressData {
 }
 
 interface PDFDocumentStats {
-    streamTypes: object; // Used stream types in the document (an item is set to true if specific stream ID was used in the document).
-    fontTypes: object; // Used font types in the document (an item is set to true if specific font ID was used in the document).
+    /**
+     * Used stream types in the document (an item is set to true if specific stream ID was used in the document).
+     */
+    streamTypes: object;
+
+    /**
+     * Used font types in the document (an item is set to true if specific font ID was used in the document).
+     */
+    fontTypes: object;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


types/pdfjs-dist/index.d.ts
```diff
@@ -260,58 +261,128 @@ interface PDFDocumentProxy {
...
 
     /**
      * TODO: return type of Promise<???>
      *  A promise that is resolved with a lookup table for mapping named destinations to reference numbers.
      **/
-    getDestinations(): PDFPromise<any[]>;
+    getDestinations(): Promise<any[]>;
+
+    /**
+     * TODO: return type of Promise<???>
```
armanio123: Are these TODOs still valid?

armanio123 (commented):
I don't have much context on the library, so there's just a few comments. The rest LGTM.

andrewbranch (changes requested):
Requesting changes to address @armanio123’s comments so the bot knows what’s going on.

types/pdfjs-dist/index.d.ts
```diff
@@ -248,6 +240,15 @@ interface PDFProgressData {
     total: number;
 }
 
+interface PDFDocumentStats {
+    streamTypes: object; // Used stream types in the document (an item is set to true if specific stream ID was used in the document).
```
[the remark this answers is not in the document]
echo99: Sure! I was following the existing comment styles in the file so wasn't sure which was preferred, but improving intellisense is definitely a good reason to use jsdoc style comments.

types/pdfjs-dist/index.d.ts
echo99: Yeah. Unfortunately the documentation in PDF.js doesn't specify clearly what these methods return other than a promise, so I was unable to determine what the real type should be

andrewbranch: I guess the TODO should at least be taken out of the JSDoc comment so users don’t see it

andrewbranch (changes requested):
Double check the version that you’re pulling these from—a lot of the methods I spot-checked, like `getViewerOptions` and `getPageLayout`, seem to have been added in 2.2, not 2.1. You just need to update the version string in the header accordingly to 2.2 or 2.3 (if you choose 2.2, please check that there are no 2.3 methods added here—I didn’t check exhaustively). Thanks!

PR #38891, opened 2019-10-04
[tus-js-client] Align definitions with v1.7

LeSuisse:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/tus/tus-js-client/pull/126

### DIFF
--- a/types/tus-js-client/index.d.ts
+++ b/types/tus-js-client/index.d.ts
@@ -1,8 +1,9 @@
-// Type definitions for tus-js-client 1.5
+// Type definitions for tus-js-client 1.7
 // Project: https://github.com/tus/tus-js-client/
 // Definitions by: Kevin Somers-Higgins <https://github.com/kevhiggins>
 //                 Marius Kleidl <https://github.com/Acconut>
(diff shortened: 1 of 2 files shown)


Acconut (commented):
This is mostly looking good, thank you for your efforts! I am only wondering why you chose names such as TusJSClientReadableStream and TusJSClientReadableStreamReadResult instead of ReadableStream etc. Would that conflict with the builtin ReadableStream from the browsers themselves (https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream)?

LeSuisse:
Hi,

Yes, it was to avoid conflicts with the types provided by Typescript. That's being said, after further thinking `TusJSClientReadableStreamReadResult` is a bit useless since we can directly use `ReadableStreamReadResult`. Also, we can build the type for the ReadableStreamReader from `ReadableStreamDefaultReader`, it's closer to what the initial implementation intended.

Acconut:
That's a great idea! The PR looks a lot better now :+1:

Acconut (approved):

PR #38899, opened 2019-10-05
Add types for appboy-web-sdk

jaredrolt:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/appboy-web-sdk/index.d.ts
+++ b/types/appboy-web-sdk/index.d.ts
@@ -1,0 +1,445 @@
+// Type definitions for appboy-web-sdk 2.4
+// Project: https://github.com/Appboy/appboy-web-sdk#readme
+// Definitions by: Jared Rolt <https://github.com/jaredrolt>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.4
+
+/*~ If this module is a UMD module that exposes a global variable 'myLib' when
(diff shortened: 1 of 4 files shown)


types/appboy-web-sdk/index.d.ts
```diff
@@ -0,0 +1,371 @@
...
+    subscribeToFeedUpdates(subscriber: (feed: Feed) => void): string;
+    subscribeToInAppMessage(callback: (inAppMesssage: InAppMessage|ControlMessage) => void): string;
+    subscribeToNewInAppMessages(subscriber: (inAppMessages: Array<InAppMessage|ControlMessage>) => Array<InAppMessage|ControlMessage>): string; // DEPRECATED; use subscribeToInAppMessage instead.
+    toggleAppboyLogging(): void;
+    trackLocation(): void;
+    unregisterAppboyPushMessages(successCallback?: () => void, errorCallback?: () => void): void;
+    wipeData(): void;
+}
+
+export default AppBoy;
```
armanio123: I didn't find any place were a default was exported. You might know more about the library as I don't use it. If you feel like this is the correct implementation I'll trust your criteria.

jaredrolt: Thanks, @armanio123. I'll double check

andrewbranch (changes requested):
Requesting changes via @armanio123’s comment so the bot knows this isn't ready

PR #38917, opened 2019-10-06
[Square-connect] Use const enums

ddimitrioglo:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/square/connect-nodejs-sdk#documentation-for-api-endpoints

### DIFF
--- a/types/square-connect/index.d.ts
+++ b/types/square-connect/index.d.ts
@@ -1,558 +1,686 @@
-// Type definitions for square-connect 2.20190925
+// Type definitions for square-connect 2.20190814
 // Project: https://docs.connect.squareup.com/
 // Definitions by: Dmitri Dimitrioglo <https://github.com/ddimitrioglo>
 //                 Richard Moot <https://github.com/mootrichard>
+//                 Nican <https://github.com/Nican>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.4
 
-export enum CountryEnum {
-    ZZ = 'ZZ',
-    AD = 'AD',
-    AE = 'AE',
-    AF = 'AF',
-    AG = 'AG',
-    AI = 'AI',
-    AL = 'AL',
-    AM = 'AM',
-    AO = 'AO',
-    AQ = 'AQ',
-    AR = 'AR',
-    AS = 'AS',
-    AT = 'AT',
-    AU = 'AU',
-    AW = 'AW',
-    AX = 'AX',
-    AZ = 'AZ',
-    BA = 'BA',
-    BB = 'BB',
-    BD = 'BD',
-    BE = 'BE',
-    BF = 'BF',
-    BG = 'BG',
-    BH = 'BH',
-    BI = 'BI',
-    BJ = 'BJ',
-    BL = 'BL',
-    BM = 'BM',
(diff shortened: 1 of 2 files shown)


mootrichard (approved):

armanio123:
I believe you should not be using `const enums` on `d.ts` files as they break with the flag `--isolatedModules`. Instead the approach is to use a `type`. e.g.:

```js
export type CountryEnum = 'ZZ' | 'AD' | 'AE' | ...
```

More details here: ://github.com/DefinitelyTyped/DefinitelyTyped/issues/23002

armanio123 (changes requested):
See comment above.

mootrichard (changes requested):
I actually agree with the comments regarding not using `const enum`. I wasn't aware of this and it does look like something that should be avoided in `d.ts` files.

ddimitrioglo:
Hi @armanio123 ,
In that case in my code, instead of using `CountryEnum.USD` i need to use `USD`? 
If yes - that's kind of weird to get rid of `enum`s in your typescript code.
Am I missing something?

Thanks.

ddimitrioglo:
@mootrichard what do you think about my comment above?

mootrichard:
@ddimitrioglo forgive my lack of understanding. I'm admittedly not _super_ well versed in TypeScript. What was wrong with the previous way of just having: 

```typescript
export enum CurrencyCode {
  USD = "USD"
}
```

Wouldn't that still allow you to use `CurrencyCode.USD`?

ddimitrioglo:
> What was wrong with the previous way

Nope, I'm getting `undefined` on runtime... actually looks like I'll face the same issue with new approach... 
Don't really know how to declare enums in `*.d.ts` in order to work with them as usual `enum`, not the string.

Hope that @armanio123 will help me here.

mootrichard:
So I went through and adjusted it to this https://gist.github.com/mootrichard/f392bd6f257a05a7b1d5219b71778e2f

Can you try that out and see whether you'd still be getting `undefined` values at runtime?

It just seems like you're not able to `declare` enums in a type definition file. Doing it this other way still defines the types and gives you type inference. 

I'm admittedly a little puzzled, since simply having enums like `CurrencyEnum.USD` is cleaner. I guess the enums are something that should be defined in the SDK rather than the `d.ts` though.

ddimitrioglo:
@mootrichard 
Of course, would be cool to have those enums in SDK, but as it's auto-generated we do not have that luxury.

So, looks like we have to switch to this approach, but with minor changes, `CurrencyEnum` => `CurrencyEnumType` or `CurrencyType` in order to explicitly say that it's not an enum anymore and use strings (ex: `'USD'`) in the code.

Thank you for your time and help!

Nican:
@ddimitrioglo Would you like to just add my ApiClient/constructor definitions into this pull request?
I do not really care for the credits, but I just want it working. 

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38992

ddimitrioglo:
@armanio123 Could you please review my changes.
It blocks me to use new version in my project. Thanks.

armanio123 (approved):
Sorry for the delay and thanks for doing the big change.

mootrichard (approved):

PR #38924, opened 2019-10-07
Fix incorrect Stripe package types

Just-Sieb:
I have two types fixes for in incorrect types.

First the subscription value in the invoice object can be null. You can see that in the docs [here](https://stripe.com/docs/api/invoices/object#invoice_object-subscription).

Second, the customer value in the InvoiceItem object is expandable so I am updating the type to follow the pattern for the other expandable item. Typing for it is in the Stripe docs [here](https://stripe.com/docs/api/invoiceitems/object#invoiceitem_object-customer).

- [ X ] Use a meaningful title for the pull request. Include the name of the package modified.
- [ X ] Test the change in your own code. (Compile and run.)
- [ X ] Add or edit tests to reflect the change. (Run with `npm test`.)
- [ X ] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [ X ] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [ X ] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [ X ] Provide a URL to documentation or source code which provides context for the suggested changes: https://stripe.com/docs/api/invoices/object#invoice_object-subscription and https://stripe.com/docs/api/invoiceitems/object#invoiceitem_object-customer

### DIFF
--- a/types/stripe/index.d.ts
+++ b/types/stripe/index.d.ts
@@ -3741,7 +3741,7 @@ declare namespace Stripe {
             /**
              * The subscription that this invoice was prepared for, if any.
              */
-            subscription: string | subscriptions.ISubscription;
+            subscription: string | subscriptions.ISubscription | null;
 
             /**
              * Only set for upcoming invoices that preview prorations. The time used to calculate prorations.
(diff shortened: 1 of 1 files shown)


pushplay (approved):
Looks good.  Thanks for the links.

PR #38934, opened 2019-10-07
bookshelf: fix return type of Models extend method

pabra:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/bookshelf/bookshelf/blob/master/lib/extend.js

### DIFF
--- a/types/bookshelf/index.d.ts
+++ b/types/bookshelf/index.d.ts
@@ -92,7 +92,7 @@ declare namespace Bookshelf {
 		static collection<T extends Model<any>>(models?: T[], options?: CollectionOptions<T>): Collection<T>;
 		static count(column?: string, options?: SyncOptions): BlueBird<number | string>;
 		/** @deprecated use Typescript classes */
-		static extend<T extends Model<any>>(prototypeProperties?: any, classProperties?: any): Function; // should return a type
+		static extend<T extends Model<any>>(prototypeProperties?: any, classProperties?: any): T;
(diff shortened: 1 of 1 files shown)


types/bookshelf/index.d.ts
```diff
@@ -92,7 +92,7 @@ declare namespace Bookshelf {
 		static collection<T extends Model<any>>(models?: T[], options?: CollectionOptions<T>): Collection<T>;
 		static count(column?: string, options?: SyncOptions): BlueBird<number | string>;
 		/** @deprecated use Typescript classes */
-		static extend<T extends Model<any>>(prototypeProperties?: any, classProperties?: any): Function; // should return a type
+		static extend<T extends Model<any>>(prototypeProperties?: any, classProperties?: any): T;
```
andrewbranch: This doesn't look right based on my reading of extend.js. I'm not sure `extend` is properly typable in TypeScript, but it’s definitely not a `T`; it’s a constructor. More specifically, it looks like it’s `Model`-like. The type parameter is also useless... `Function` isn’t a good type but you’re not going to do much better. It looks to me like

```ts
static extend<T extends object>(prototypeProperties?: any, classProperties?: T): Model<any> & T;
```

would be a slight improvement. But then again, this method is deprecated, so is it worth changing at all?

andrewbranch (changes requested):

PR #38937, opened 2019-10-07
feat: Add types for react-native-phone-input

maael:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/react-native-phone-input/index.d.ts
+++ b/types/react-native-phone-input/index.d.ts
@@ -1,0 +1,154 @@
+// Type definitions for react-native-phone-input 0.2
+// Project: https://github.com/thegamenicorus/react-native-phone-input
+// Definitions by: Matthew Elphick <https://github.com/maael>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.9
+
+import * as React from 'react';
+import { StyleProp, ViewStyle as ViewStyleRaw, TextStyle as TextStyleRaw, TextInput, ImageRequireSource } from 'react-native';
+
+export type ViewStyle = StyleProp<ViewStyleRaw>;
+export type TextStyle = StyleProp<TextStyleRaw>;
+
+export interface CountriesListItem {
+  name: string;
+  iso2: string;
+  dialCode: string;
+  priority: number;
+  areaCodes: ReadonlyArray<string> | null;
+}
+
+export interface PickerData {
+  key: number;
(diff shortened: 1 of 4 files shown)


types/react-native-phone-input/index.d.ts
```diff
@@ -0,0 +1,154 @@
...
+     */
+    onPressFlag?: () => void;
+    /**
+     * Custom countries list
+     */
+    countriesList?: ReadonlyArray<CountriesListItem>;
+}
+
+export default class ReactNativePhoneInput<TextComponentType extends React.ComponentType = typeof TextInput> extends React.Component<ReactNativePhoneInputProps<TextComponentType>> {
+  picker?: React.Ref<ThisType<ReactNativePhoneInput>>;
```
armanio123: Documentation doesn't show `picker` as a property.

types/react-native-phone-input/index.d.ts
```diff
@@ -0,0 +1,154 @@
...
+     * Function to be invoked when press on flag image
+     */
+    onPressFlag?: () => void;
+    /**
+     * Custom countries list
+     */
+    countriesList?: ReadonlyArray<CountriesListItem>;
+}
+
+export default class ReactNativePhoneInput<TextComponentType extends React.ComponentType = typeof TextInput> extends React.Component<ReactNativePhoneInputProps<TextComponentType>> {
```
armanio123: There's some missing properties:
`blur`, `onPressCancel`, `onPressConfirm`

armanio123 (changes requested):
Leaving a couple of comments. Documentation might be outdated. Let me know if you want to merge as is.

types/react-native-phone-input/index.d.ts
maael: It's set [here](https://github.com/thegamenicorus/react-native-phone-input/blob/master/lib/index.js#L216). I'm not too sure whether to include it here like this or not, I don't have much experience typing refs.

types/react-native-phone-input/index.d.ts
maael: Added in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38937/commits/66080914b81ba107d7025918c1507e3b059f28d6

types/react-native-phone-input/index.d.ts
armanio123: Thanks, this is the correct place.

armanio123 (approved):

PR #38942, opened 2019-10-07
Update the types for node-cache 

seungdols:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/node-cache/node-cache/pull/144/files>>

### DIFF
--- a/types/node-cache/index.d.ts
+++ b/types/node-cache/index.d.ts
@@ -14,153 +14,126 @@
 type Key = string | number;
 
 declare namespace NodeCache {
-	interface NodeCache {
-		/** container for cached data */
-		data: Data;
+    interface NodeCache {
+        /** container for cached data */
+        data: Data;
 
-		/** module options */
-		options: Options;
+        /** module options */
+        options: Options;
 
-		/** statistics container */
-		stats: Stats;
+        /** statistics container */
+        stats: Stats;
 
-		/**
-		 * get a cached key and change the stats
-		 *
-		 * @param key cache key or an array of keys
-		 * @param cb Callback function
-		 */
-		get<T>(
-			key: Key,
-			cb?: Callback<T>
-		): T | undefined;
+        /**
+         * get a cached key and change the stats
+         *
+         * @param key cache key or an array of keys
+         * @param cb Callback function
+         */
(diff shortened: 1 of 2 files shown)


dthunell:
I have not been using this library for over a year so I cannot validate the changes made, but it seems it now packages it's own TS definitions so should it not be removed from this repo then to avoid confusion/conflict?

seungdols:
cc. @daluf Excuse me first. I changed the fix regarding the type of node-cache, what is your opinion?

@dthunell Thank you for your good feedback. I think that First of all, you should ask the node-cache maintainer for opinion.

erdii:
Hey @seungdols!
Thanks for your contribution.

I think it would be better to "self-host" the typings inside the project itself because:
* instant "type-script-like" experience for JS developers who use vscode
* no need to do an extraneous "npm install -D @types/node-cache"
* easier to maintain

---

Would you be willing to provide this pull-request against our main repo https://github.com/node-cache/node-cache ?

---

I think we also should add a stub definition here? Something like:

```
> npm i -D @types/moment

npm WARN deprecated @types/moment@2.13.0: This is a stub types definition for Moment (https://github.com/moment/moment). Moment provides its own type definitions, so you don't need @types/moment installed!
```

seungdols:
Thank you for your advice @erdii.

I was thinking that the types of node-cache exist in `@ types` already.

So, I updated @types for node-cache in `@ types` because of updating of node-cache is very important to me.

You said it to me, Why should the type of node cache be maintained? When I was thinking, it was the conclusion that you were right. I agree with you.

It doesn’t need to keep types of node-cache in @types. I suggest you one. **Should can I removed out `types / node-cache` in this repository.** What are you think?

I unaccustomed to open-source contribute So, I hope you understand.

Thank you.

andrewbranch:
It sounds like node-cache should be removed from DefinitelyTyped here. @seungdols, if you’re willing to make the removal, please see [this section of the README](https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package) for instructions. Thanks all!

andrewbranch (changes requested):
_[Tracking the removal of this package]_

PR #38946, opened 2019-10-07
[CI] use NodeJS 10 instead 8 for travis CI

Flarna:
Please fill in this template.

NodeJs 8 is quite old and EOL soon, current LTS is 10 therefore we should use it in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - 8
+  - 10
 
 notifications:
   email: false


MichaelDeBoey:
Maybe use `8`, `10` & `12`? 🤔

Flarna:
What is the benefit of testing several times? The node version used in this is that one running the linter, nothing else. It has nothing to do with the definition itself nor does it increase test coverage for definitons if lister is run with serveral versions.

MichaelDeBoey:
@Flarna Just to be sure everything keeps working on all versions.
Sometimes something breaks on newer/older versions because of breaking changes in API or things like that.

Flarna:
@MichaelDeBoey I think such compat tests for dtslint/types publisher should be done in CI of their repos not here.

MichaelDeBoey:
I totally disagree, I think there could be compatibility issues whatsoever and it doesn’t hurt testing against all supported Node versions.

But let’s see what maintainers decide.

sandersn:
@Flarna This is a good idea, but the important CI system is Azure devops at this point. I'm going to shut off travis as soon as @RyanCavanaugh has time to switch typescript-bot to watching devops' check results instead of travis'.

@andrewbranch I'm fine with merging this if you don't mind watching the PR builds for a while afterward.

@MichaelDeBoey We don't have enough wall clock time or computer time for multiple node runs on libraries with lots of dependents -- React for example. We'll have to hope that Typescript can present a consistent compilation environment on different versions of node, since we *do* test Typescript on multiple versions of node.

sandersn (approved):

PR #38947, opened 2019-10-07
feat: Add `es‑to‑primitive`

ExE-Boss:
Please fill in this template.

If adding a new definition:
- [x] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration`: https://github.com/ljharb/es-to-primitive/pull/44#issuecomment-539032313

---

review?(@ljharb, @andrewbranch, @armanio123, @uniqueiniquity)


types/es-to-primitive/es2015.d.ts
```diff
@@ -0,0 +1,5 @@
+declare function ToPrimitive(
+	input: any,
+	hint?: Function
+): string | number | boolean | symbol | null | undefined;
```
ljharb:
```suggestion
): string | number | bigint | boolean | symbol | null | undefined;
```

types/es-to-primitive/es5.d.ts
```diff
@@ -0,0 +1,5 @@
+declare function ToPrimitive(
+	input: any,
+	hint?: typeof String | typeof Number,
+): string | number | boolean | null | undefined;
```
ljharb:
```suggestion
): string | number | bigint | boolean | null | undefined;
```
i think this can still produce a symbol, whether it’s the es5 algorithm or not.

types/es-to-primitive/es-to-primitive-tests.ts
```diff
@@ -0,0 +1,14 @@
+import ToPrimitive = require('es-to-primitive');
+import ToPrimitiveES5 = require('es-to-primitive/es5');
+import ToPrimitiveES6 = require('es-to-primitive/es2015');
+import ToPrimitiveES2015 = require('es-to-primitive/es2015');
+
+const any: any = void 0;
```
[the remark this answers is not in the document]
ExE-Boss: That will cause the tests to fail on **TypeScript** &lt; 3

types/es-to-primitive/es2015.d.ts
```diff
@@ -0,0 +1,5 @@
+declare function ToPrimitive(
+	input: any,
+	hint?: Function
```
ljharb: not just any function, but only String or Number or Symbol.

ExE-Boss: Well, the implemented algorithm doesn’t actually really care [what function is supplied](https://github.com/ljharb/es-to-primitive/blob/9c0dd264951a7cd5bbd8c5d1951843bf48351113/es2015.js#L47-L54), although [that might be incorrect](https://tc39.es/ecma262/#sec-toprimitive).

types/es-to-primitive/es2015.d.ts
ExE-Boss: I’ve applied this in **TypeScript** 3.2 specific types.

types/es-to-primitive/es-to-primitive-tests.ts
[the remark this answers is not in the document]
ljharb: is there no way to support both, so that it’s any on v2 and unknown on v3?

types/es-to-primitive/es2015.d.ts
ljharb: yeah, there should be an assertion. Either way the types should enforce it.

types/es-to-primitive/ts3.2/es2015.d.ts
```diff
@@ -0,0 +1,5 @@
+declare function ToPrimitive(
+	input: any,
```
ljharb:
```suggestion
	input: unknown,
```

types/es-to-primitive/ts3.2/es5.d.ts
```diff
@@ -0,0 +1,5 @@
+declare function ToPrimitive(
+	input: any,
```
ljharb:
```suggestion
	input: unknown,
```

types/es-to-primitive/ts3.2/es-to-primitive-tests.ts
```diff
@@ -0,0 +1,14 @@
+import ToPrimitive = require('es-to-primitive');
+import ToPrimitiveES5 = require('es-to-primitive/es5');
+import ToPrimitiveES6 = require('es-to-primitive/es2015');
+import ToPrimitiveES2015 = require('es-to-primitive/es2015');
+
+const any: any = void 0;
```
ljharb:
```suggestion
const any: unknown = void 0;
```

types/es-to-primitive/es-to-primitive-tests.ts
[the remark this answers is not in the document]
ExE-Boss: There is, but only from TS 3.1 using the `typesVersions` key in `package.json`.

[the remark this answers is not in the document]
ljharb: ok, so 3.1+ can still benefit from unknown?

armanio123:
@ljharb are all comments addressed and is this ready to be merged?

types/es-to-primitive/.editorconfig
```diff
@@ -0,0 +1,3 @@
+# This package uses tabs
```
armanio123: Could you please remove this file?

armanio123 (changes requested):

armanio123:
@ljharb I noticed that you're the package maintainer for this library. You could also place the `d.ts` file in your repo and ship it with the package as well if you're willing to maintain the file.

This is optional but recommended way of shipping `d.ts` files.

ljharb:
@armanio123 in no way would i ever want to do that unless there was a programatic way to ensure that it stayed in sync with the unannotated JS; without that, I think it's *highly inadvisable* for non-TS code to ship with first-class types.

ljharb (approved):

types/es-to-primitive/.editorconfig
ExE-Boss: If I do this, then it’s very likely that someone would change tabs to spaces under me.

Also, all my DT contributions have this file.

ExE-Boss:
@armanio123

#39055 and #39058 have been approved by @andrewbranch because packages are still free to choose the superior tabs over the inferior spaces for indentation.

armanio123 (approved):
@Exe-Boss I have a discussion with the team about this, and we agreed the PR is ok for now. Nonetheless, we still strongly suggest to go for the defaults defined by the repo. I'm approving the PR.

PR #38956, opened 2019-10-07
[readable-stream] refer to node v10 type definitions

Flarna:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/readable-stream

readable-stream is based on streams of node.js 10 therefore it should refer to these type definitions also to avoid that changes in actual node.js versions require wrong changes in readable-stream.

Not sure if such binding to an older version results in conflicts for projects refering to @types/readable-stream and @types/node@12.x - I did a fast local test and it seems it's fine.

Refs: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38907#issuecomment-538799991

fyi @SimonSchick

### DIFF
--- a/types/readable-stream/index.d.ts
+++ b/types/readable-stream/index.d.ts
@@ -4,12 +4,12 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
-/// <reference types="node" />
+/// <reference types="node/v10" />
 
 import * as events from "events";
 import * as stream from "stream";
 import * as SafeBuffer from "safe-buffer";
-import { StringDecoder } from "string_decoder";
+import { NodeStringDecoder } from "string_decoder";
 
 declare class _Readable extends stream.Readable {
     // static ReadableState: _Readable.ReadableState;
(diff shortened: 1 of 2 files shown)


types/readable-stream/index.d.ts
```diff
@@ -4,12 +4,12 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
-/// <reference types="node" />
+/// <reference types="node/v10" />
```
SimonSchick: Isn't that going to break when installed?

Flarna: To my understanding this results in @types/node v10 installed if @types/readable-stream is the only one asking for node definitions (or all others don't care about the version) which should be fine.

In case someone has additionally @types/node v12.x as dependency the resulting folder structure should be as follows:
`node_modules/@types/node/`  => v12.x definitions
`node_modules/@types/readable-stream/node_modules/@types/node/`=> v10.x definitions
`node_modules/@types/readable-stream/` the readable stream definitions

I put this together in gave this a fast try and it seems to work otherwise I wouldn't create such a PR :grin:. But yes, I still have not the best feeling with this binding.

Any input from typescript/type definiton experts is welcome regarding this!

SimonSchick: Imo it would be better to entirely decouple this package from node streams and basically copy paste stuff in there.

TeamworkGuy2 (approved):
Interesting, guess I never realized readable-stream had stopped mirroring node.js at v10.
I think this change should work fine with node v12.x definitions in the same project, but like @SimonSchick said, it's kind of a pain that we need the dependency. But on the other hand it would be a lot of duplication to copy-paste all the relevant dependencies out of node v10 definitions into this definition.

Flarna:
I retested today and it seems I did something wrong yesterday. This PR would break usercode.
Seems the only option to decouple this will be copy/paste from Node 10.

PR #38960, opened 2019-10-08
Create types for react-native-rss-parser

emilianoLeite:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/react-native-rss-parser/index.d.ts
+++ b/types/react-native-rss-parser/index.d.ts
@@ -1,0 +1,56 @@
+// Type definitions for react-native-rss-parser 1.4
+// Project: https://github.com/jameslawler/react-native-rss-parser
+// Definitions by: Emiliano Leite <https://github.com/emilianoLeite>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.3
+
+export { };
+
+type Maybe<T> = T | undefined;
+
+export interface FeedItem {
+    id: string;
+    title: string;
+    links: Array<{
+        url: string;
+        rel: string;
+    }>;
+    description: string;
+    content: string;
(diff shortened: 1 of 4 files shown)


types/react-native-rss-parser/index.d.ts
```diff
@@ -0,0 +1,83 @@
+// Type definitions for react-native-rss-parser 1.4
+// Project: https://github.com/jameslawler/react-native-rss-parser
+// Definitions by: Emiliano Leite <https://github.com/emilianoLeite>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.3
+
+export { };
```
armanio123: Could you please remove this `export`?

armanio123 (changes requested):

types/react-native-rss-parser/index.d.ts
emilianoLeite: @armanio123 I added this export because I explicitly did not want to export the `Maybe` type, since this type acts only as an helper and is not "core" to the module.

Omitting the `export { }` outputs the following error:
```bash
➜ yarn lint react-native-rss-parser 
yarn run v1.19.0
$ dtslint types react-native-rss-parser
Error: .../DefinitelyTyped/types/react-native-rss-parser/index.d.ts:7:6
ERROR: 7:6  strict-export-declare-modifiers  All declarations in this module are exported automatically. Prefer to explicitly write 'export' for clarity. If you have a good reason not to export this declaration, add 'export {}' to the module to shut off automatic exporting. See: https://github.com/Microsoft/dtslint/blob/master/docs/strict-export-declare-modifiers.md
```

armanio123: Ah makes sense, thanks for clarifying.

armanio123 (approved):

PR #38974, opened 2019-10-08
Add unstable_next() to the React scheduler

mauricedb:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/react/blob/master/packages/scheduler/src/Scheduler.js#L239-L262
- [ ] ~~If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.~~ This API was already present.


eps1lon (approved):
Sorry this fell under the radar. Thanks!

types/scheduler/index.d.ts
```diff
@@ -21,6 +21,7 @@ export const unstable_IdlePriority = 5;
 export const unstable_LowPriority = 4;
 export function unstable_runWithPriority<T>(priorityLevel: number, eventHandler: () => T): T | undefined;
 export function unstable_scheduleCallback(priorityLevel: number, callback: FrameCallbackType, options?: { delay?: number, timeout?: number}): CallbackNode;
+export function unstable_next<T extends () => any>(eventHandler: T): ReturnType<T>;
```
Methuselah96:
```suggestion
export function unstable_next<T>(eventHandler: () => T): T;
```
To bring it into conformity with how we're doing it in the rest of the file (e.g. line 22) and so we don't have to use `ReturnType`.

Methuselah96 (changes requested):

Methuselah96 (approved):

Methuselah96 (approved):

mauricedb:
@Methuselah96 I included the suggested changes. The only things is that this includes `undefined` as a possible return type which I think is incorrect. I believe `unstable_next<T>(eventHandler: () => T): T` would be the correct signature. Please let me know if you want me to update it again.

Methuselah96:
@mauricedb The reason I have the return type as `T | undefined` is because the call to `eventHandler` is wrapped in a `try` block, which means that `unstable_next` will return undefined if the `eventHandler` throws an error (https://github.com/facebook/react/blob/master/packages/scheduler/src/Scheduler.js#L257-L261). At least that's what I think will happen if it catches an error.

`unstable_runWithPriority` and `unstable_wrapCallback` also follow this pattern, so if we decide to drop the `undefined`, then we should drop them for those methods as well.

@eps1lon Do you have any thoughts as to whether we should keep the `| undefined` for these methods?

mauricedb:
@Methuselah96 The call to the callback is in a `try/finally' block without a catch. So the `error` will bubble up to the calling code and not return `undefined` from these functions.

Methuselah96:
@mauricedb Yeah, you're totally right, sorry about that. Can you go ahead and remove `| undefined` from those three functions?

mauricedb:
@Methuselah96 No problem, updated the definition and checks

Methuselah96 (approved):

PR #38978, opened 2019-10-08
selenium-webdriver: added setBrowserName and setBrowserVersion Capabilities methods

Dzenly:
Function descriptions are taken from here:

* https://github.com/SeleniumHQ/selenium/blob/master/javascript/node/selenium-webdriver/lib/capabilities.js

* https://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/index_exports_Capabilities.html


andrewbranch (approved):
Thanks!

PR #34894, opened 2019-04-21
chore: Create Prettier configuration

ExE-Boss:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

PR #34364, opened 2019-03-31
refactor(shelljs): Refactor ShellJS types to be more correct

ExE-Boss:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: http://shelljs.org (specifically: http://shelljs.org#execcommand--options--callback, which states that `exec` returns a `ShellString` for synchronous operations)
- [x] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

---

I’ve moved longer signatures and signatures with `options` to the top, since TypeScript determines the used overload by going from the top to bottom, returning the moment an overload signature matches, which would result in `...files` absorbing the `options`.

I’ve also split the array version of functions from the varargs version, since the varargs version isn't allowed to have mixed strings and arrays.

I’ve also moved all function definitions into callable interfaces to make reuse in `ShellReturnValue` easier.

---

review?(@sandersn)

PR #38979, opened 2019-10-08
feat: Add `es‑abstract` [WIP]

ExE-Boss:
## Depends on:
- [x] [~~#38947~~](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38947) (merged)

---

Please fill in this template.

If adding a new definition:

---

review?(@ljharb, @andrewbranch, @johnnyreilly, @weswigham)


types/es-abstract/es2015.d.ts
```diff
@@ -0,0 +1,160 @@
...
+	OrdinaryDefineOwnProperty(O: object, P: ESPropertyKey, Desc: ES5.PropertyDescriptor): boolean;
+	OrdinaryGetOwnProperty(O: object, P: ESPropertyKey): ES5.PropertyDescriptor | undefined;
+
+	ArrayCreate(length: number, proto?: object | null): unknown[];
+	ArraySetLength(A: unknown[], Desc: ES5.PropertyDescriptor): boolean;
+	CreateHTML(string: unknown, tag: string, attribute: string, value?: unknown): string;
+
+	GetOwnPropertyKeys(O: object, Type: 'String'): string[];
+	GetOwnPropertyKeys(O: object, Type: 'Symbol'): symbol[];
+	GetOwnPropertyKeys(O: object, Type: 'String' | 'Symbol'): string[] | symbol[];
```
ljharb: this isn't a possible return type for the function.

types/es-abstract/es2018.d.ts
```diff
@@ -0,0 +1,47 @@
...
+interface ES2018 extends Omit<ES2017, 'EnumerableOwnProperties' | 'GetSubstitution' | 'IsPropertyDescriptor'> {
+	readonly EnumerableOwnPropertyNames: typeof ES2017.EnumerableOwnProperties;
+
+	// tslint:disable-next-line: ban-types
+	thisSymbolValue(value: symbol | Symbol): symbol;
+
+	IsStringPrefix(p: string, q: string): boolean;
+	NumberToString(m: number): string;
+	CopyDataProperties<T, S, E extends ES2017.PropertyKey>(target: T, source: S, excludedItems: E[]): T & Omit<S, E>;
+	PromiseResolve<T>(C: PromiseConstructor, x: T): Promise<T>;
```
ljharb:
```suggestion
	PromiseResolve: typeof Promise.resolve
```

ExE-Boss: No, because `Promise.resolve` takes one parameter, whereas this takes 2 as it’s essentially:
```js
function PromiseResolve(C, x) {
	return Promise.resolve.call(C, x);
}
```

types/es-abstract/es2015.d.ts
ExE-Boss: This is because of https://github.com/microsoft/TypeScript/issues/14107.

types/es-abstract/es2018.d.ts
ljharb: ahhh fair point. in that case, what about `typeof Promise.resolve.call`?

ExE-Boss: That won’t work, as `typeof Promise.resolve.call` is:
```ts
(this: Function, thisArg: any, ...argArray: any[]) => any
```

ljharb: hmm, i thought the .call overloads were supposed to work here. `Promise.resolve.call` should  have a typeof `(thisArg, value)`.

ExE-Boss: That’s only when you use `strictBindCallApply`, and only for code that actually uses it like so:
```ts
function NewCustomPromise<T>(someValue: T): CustomPromise<T> {
	const p = Promise.resolve.call(CustomPromise, someValue);
	// Do stuff with `p`
	return p;
}
```

ExE-Boss:
needinfo?(@ljharb): Should I also add types for utilities in the `helpers` directory?

ljharb:
everything reachable is part of the public api, so probably yes.

ExE-Boss:
review?(@ljharb)

ExE-Boss:
review?(@ljharb): I&nbsp;have&nbsp;now added&nbsp;type&nbsp;definitions for&nbsp;all&nbsp;reachable&nbsp;files.

types/es-abstract/.editorconfig
```diff
@@ -0,0 +1,3 @@
+# This package uses tabs
```
weswigham: Don't override the root .editorconfig, thanks. We really do want all definitions to conform to that global style guideline. It makes both review and contribution much easier. ❤️

weswigham (dismissed):

types/es-abstract/.editorconfig
ExE-Boss: @sandersn, @PranavSenthilnathan, @andrewbranch and&nbsp;@armanio123 have&nbsp;all&nbsp;allowed me&nbsp;to&nbsp;keep&nbsp;using the&nbsp;superior&nbsp;tabs&nbsp;in:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34364#discussion_r270972375, https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34894#discussion_r279131192, https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39055#issuecomment-542370662, https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38947#pullrequestreview-303558276

weswigham: Yep, and that ends today. We've had a talk, and no - just because someone (mistakenly) allowed a commit with what amounts to a style rule suppression, doesn't mean we're going to keep allowing it.

weswigham: The root `.editorconfig` exists quite literally so I shouldn't have to have this discussion BTW - your insistence is going to force us to update our tooling to forbid nested `.editorconfig`s. Congratulations on being a catalyst for change~

types/es-abstract/helpers/getInferredName.d.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Gets the inferred function name for the `{ [s]() {...} }` construct.
+ *
+ * **Note:** Intended to be used with symbols.
+ *
+ * @param s The symbol to use as the object key.
+ */
+declare function getInferredName(s: unknown): string;
+export = getInferredName;
```
ljharb: the export here could also be `null`, if it's in an older engine.

types/es-abstract/helpers/regexTester.d.ts
```diff
@@ -0,0 +1,2 @@
+declare function regexTester(regex: RegExp): typeof RegExp.prototype.test;
```
ljharb: this type isn't accurate; it returns a call-bound `RegExp.prototype.test`

types/es-abstract/es2016.d.ts
```diff
@@ -0,0 +1,33 @@
+import ES2015 = require('./es2015');
+
+interface ES2016 extends ES2015 {
+	/**
+	 * @param x
+	 * @param y
+	 *
+	 * @throws {TypeError} If `x` or `y` is a `number` or they're different types.
+	 */
+	SameValueNonNumber(x: unknown, y: unknown): boolean;
```
ljharb:
```suggestion
	SameValueNonNumber(x: object | null | undefined | string | boolean | bigint | symbol, y: object | null | undefined | string | boolean | bigint | symbol): boolean;
```

ExE-Boss: That’s what `unknown` is.

types/es-abstract/helpers/regexTester.d.ts
ExE-Boss: That’s not really describable using **TypeScript**’s type system.

types/es-abstract/es2016.d.ts
ljharb: no - `unknown` includes `number`. This does not.

types/es-abstract/helpers/regexTester.d.ts
ljharb: ok, but that doesn't make `typeof RegExp.prototype.test` correct :-) what about:
```ts
(string) => ReturnType<RegExp.prototype.test>
```
?

ExE-Boss: I inlined the return type of `callBind<typeof RegExp.prototype.test>`.

sandersn: you might be able to use a `this: undefined` parameter to represent a call-bound method (or some other type if I misunderstood what call-bound means).

ljharb: @sandersn call-bound means that the first argument becomes the `this` of the bound method, ie, `callBoundFoo(a, b, c)` is like `foo.call(a, b, c)` but without relying on `Function.prototype.call` existing.

ExE-Boss: Call‑bound in this case means:
```ts
regexTester(regex: RegExp) {
	return Function.prototype.call.bind(RegExp.prototype.test, regex);
}
```

ExE-Boss:
These&nbsp;types are&nbsp;still&nbsp;very&nbsp;much a&nbsp;work&nbsp;in&nbsp;progress.

ExE-Boss:
These&nbsp;types are&nbsp;still&nbsp;very&nbsp;much a&nbsp;work&nbsp;in&nbsp;progress.

ExE-Boss:
Also, it’s&nbsp;likely&nbsp;that I’ll&nbsp;soon have&nbsp;to&nbsp;update this&nbsp;to&nbsp;v2.0 because&nbsp;of&nbsp;https://github.com/ljharb/es-abstract/issues/48, https://github.com/ljharb/es-abstract/issues/66, https://github.com/ljharb/es-abstract/pull/67, https://github.com/ljharb/es-abstract/pull/68 and&nbsp;https://github.com/ljharb/es-abstract/pull/69.

ljharb:
I wouldn't wait for v2; it'd be more useful to get v1.17 out.

ljharb:
… why do i need 3 pings

ljharb (approved):

sandersn:
@ljharb the bot is now running asynchrono
@orta you wanted to know about bad behaviour like this I think.
usly!

orta:
Thanks, that is definitely suspicious

PR #38897, opened 2019-10-05
fix: Use ListIterator for the remove predicate.

ajmacd:
Unlike many lodash functions, `remove` does not accept an iteratee shorthand, according to the [documentation](https://lodash.com/docs/4.17.15#remove) and the [source](https://github.com/lodash/lodash/blob/c89637e4db06592b319d3ca53c854fbfb8545349/remove.js#L39).

`ListIterator<T, NotVoid>` is drawn from the existing `ListIteratee<T> type`.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<url here>>
- [ ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

PR #38987, opened 2019-10-08
Revert "fix: Use ListIterator for the remove predicate."

ajmacd:
Reverts DefinitelyTyped/DefinitelyTyped#38897

I manged to delude myself into believing a bug in my production code was due to passing an iteratee shorthand to `remove`. My delusion was validated by the lodash documentation and apparent source here:
https://github.com/lodash/lodash/blob/master/remove.js

But, taking a look at the _actual_ source distributed via npm:
https://github.com/lodash/lodash/blob/4.17.15-npm/remove.js#L41

We can see that iteratee shorthands are absolutely supported.

### DIFF
--- a/types/lodash/common/array.d.ts
+++ b/types/lodash/common/array.d.ts
@@ -1858,7 +1858,7 @@ declare module "../index" {
          */
         remove<T>(
             array: List<T>,
-            predicate?: ListIterator<T, NotVoid>
+            predicate?: ListIteratee<T>
         ): T[];
     }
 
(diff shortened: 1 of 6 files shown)


ajmacd:
@aj-r apologies for wasting your time, but seems my change from the other day should be reverted.

aj-r (approved):
Weird!

crashmstr:
Looks like something stalled out in completing the PR? This is causing issues for me this morning with 47 compiler errors after getting the version of the package this is supposed to fix.

ajmacd:
@armanio123 would you be able to merge this? It's reverting a bad change affecting users.

PR #38994, opened 2019-10-09
fix: sparql-http-client should use built-in fetch and default export

tpluscode:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/zazuko/sparql-http-client/blob/master/index.js#L118

The package does commons default export and I got misled by the tslint error. Following numerous other packages I disabled that lint rule. Also switched to the default typing of fetch as the package does not really import `node-fetch`

### DIFF
--- a/types/sparql-http-client/index.d.ts
+++ b/types/sparql-http-client/index.d.ts
@@ -5,7 +5,6 @@
 // TypeScript Version: 2.3
 
 import { Term } from 'rdf-js';
-import fetch, { Response, RequestInit } from 'node-fetch';
 import { URL } from 'url';
 
 export interface SparqlHttpOptions {
@@ -14,8 +13,8 @@ export interface SparqlHttpOptions {
 }
 
 export interface SparqlClientOptions extends SparqlHttpOptions {
-    fetch: typeof fetch;
-    URL: typeof URL;
+    fetch?: typeof fetch;
+    URL?: typeof URL;
 }
 
 export interface QueryRequestInit extends SparqlHttpOptions, RequestInit {}
@@ -32,8 +31,8 @@ export interface SelectResponse {
     json(): Promise<SelectBindings & AskResult>;
 }
 
-export class SparqlHttp<TResponse extends Response = Response> {
-    constructor(options?: SparqlHttpOptions);
+export default class SparqlHttp<TResponse extends Response = Response> {
+    constructor(options?: SparqlClientOptions);
     updateQuery(query: string, options?: QueryRequestInit): Promise<Response>;
     selectQuery(query: string, options?: QueryRequestInit): Promise<SelectResponse & TResponse>;
     constructQuery(query: string, options?: QueryRequestInit): Promise<TResponse>;
--- a/types/sparql-http-client/sparql-http-client-tests.ts
+++ b/types/sparql-http-client/sparql-http-client-tests.ts
@@ -1,4 +1,4 @@
-import { SparqlHttp } from 'sparql-http-client';
+import SparqlHttp from 'sparql-http-client';
 
 const client = new SparqlHttp();
 
--- a/types/sparql-http-client/tslint.json
+++ b/types/sparql-http-client/tslint.json
@@ -1,1 +1,6 @@
-{ "extends": "dtslint/dt.json" }
+{ 
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "npm-naming": false
+    }
+ }
--- a/types/sparql-http-client/tsconfig.json
+++ b/types/sparql-http-client/tsconfig.json
@@ -2,7 +2,8 @@
     "compilerOptions": {
         "module": "commonjs",
         "lib": [
-            "es6"
+            "es6",
+            "webworker"
         ],
         "noImplicitAny": true,
         "noImplicitThis": true,


types/sparql-http-client/tslint.json
```diff
@@ -1 +1,6 @@
-{ "extends": "dtslint/dt.json" }
+{ 
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "npm-naming": false
```
armanio123: Is there a reason for turning off this rule? I see that the npm package has the same name.

armanio123 (changes requested):

types/sparql-http-client/index.d.ts
```diff
@@ -32,8 +31,8 @@ export interface SelectResponse {
     json(): Promise<SelectBindings & AskResult>;
 }
 
-export class SparqlHttp<TResponse extends Response = Response> {
-    constructor(options?: SparqlHttpOptions);
+export default class SparqlHttp<TResponse extends Response = Response> {
```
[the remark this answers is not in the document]
tpluscode: The package is implemented as follows:

```js
function SparqlHttp (options) {
   // constructor ...
}

module.exports = SparqlHttp
```

How should that be typed if not as a default export?

types/sparql-http-client/tslint.json
tpluscode: Well yes, this particular rule gives the "default export" error which you also pointed out

types/sparql-http-client/index.d.ts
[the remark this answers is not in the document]
armanio123: In that case you should just export the class as it used to be before.

types/sparql-http-client/sparql-http-client-tests.ts
```diff
@@ -1,4 +1,4 @@
-import { SparqlHttp } from 'sparql-http-client';
+import SparqlHttp from 'sparql-http-client';
```
armanio123: Since this package has not default this import is incorrect as this syntax is used for `default`.

types/sparql-http-client/tslint.json
armanio123: This rule is unrelated to the `default`. This is for the package name and the d.ts name which I don't see they should have a problem.

tpluscode:
@armanio123, clearly there is something else wrong still. When I apply the changes as you request, the typings fail me in my ultimate project.

In my tsconfig I have `"esModuleInterop": true`. TypeScript shows an error a shown below

```
import SparqlHttp from 'sparql-http-client'

// TS2709: Cannot use namespace 'SparqlHttp' as a type.
const client: SparqlHttp
```

When I change the import to be `{ SparqlHttp }` then I get a runtime error

`TypeError: sparql_http_client_1.SparqlHttp is not a constructor`

The `default` export was the only way that did not produce any errors

armanio123:
I believe the whole `d.ts` file is incorrect. You might need to refactor it.
First the issues:

> In my tsconfig I have "esModuleInterop": true. TypeScript shows an error a shown below
> ```
> import SparqlHttp from 'sparql-http-client'
>
> // TS2709: Cannot use namespace 'SparqlHttp' as a type.
> const client: SparqlHttp
> ```

This issues happens because you're trying to use the import as a namespace. Thus typescript complains about not being a type. If you type `SparqlHttp.SparqlHttp` it will no longer complain as is able to find the class. Nonetheless this is not the right solution because when you instantiate the object you'll get an error on runtime:
```js
const client: SparqlHttp.SparqlHttp = new SparqlHttp.SparqlHttp(); // This will fail at runtime
```
> When I change the import to be { SparqlHttp } then I get a runtime error
> `TypeError: sparql_http_client_1.SparqlHttp is not a constructor`

This error happens because you're destructuring a variable that doesn't exists on the library. So you're trying to instantiate an `undefined`.

So the correct solution is to refactor the `d.ts` file to use `export =` syntax instead.
- `d.ts` file:
```ts
declare namespace SparqlHttp {
    export interface SparqlHttpOptions {
        endpointUrl?: string;
        updateUrl?: string;
    }

    export interface SparqlClientOptions extends SparqlHttpOptions {
    ...
}

declare class SparqlHttp<TResponse extends Response = Response> {
     constructor(options?: SparqlHttp.SparqlClientOptions);
     updateQuery(query: string, options?: SparqlHttp.QueryRequestInit): Promise<Response>;
    ....
}

export = SparqlHttp;
```

- `client.ts` using `esModuleInterop`:
```ts
import SparqlHttp from 'sparql-http-client';
```
- `client.ts` without `esModuleInterop`:
```ts
import SparqlHttp = 'sparql-http-client';
```

Give it a try on your code and let me know if this worked for you.

tpluscode:
Fabulous, this is it! Thank you @armanio123 for your support.

The only difference is that without `esModuleInterop` the import is `import * as SparqlHttp from 'sprawl-http-client'`. Works like charm though.

Would you like me to update and reopen this PR or submit new?

armanio123:
You can update this PR and reopen, I'll take a look once done.

By the way, be careful with the `import *` syntax as that's used for namespaces which should not be callable. I still recommend using `import sparqlHttp = require('sparql-http-client');`

See this [stack overflow answer](https://stackoverflow.com/questions/39415661/what-does-resolves-to-a-non-module-entity-and-cannot-be-imported-using-this) for details about it.

tpluscode:
Re-submitted as #39966 because GitHub did not pick up changes on the branch

armanio123:
Looks good. I just approve and merge it.

PR #38997, opened 2019-10-09
Fixing confusion with the new and constructor keywords.

erik-krogh:
It sometimes happens that developers confuse the `new` and `constructor` keywords. 
Consider the following example:

```TypeScript
interface Foo {
    constructor(): Bar;
}
```

With the `Foo` interface the developer meant to create an interface with a constructor, but instead created a interface that contains a method named `constructor`, which is not actually a constructor.

The developer might have meant to write:

```TypeScript
interface Foo {
    new(): Bar;
}
```

The same thing can happen when the  `new` keyword is accidentally used instead of the `constructor` keyword in a class definition. 

This pull request attempts to fix most instances of this mistake across DefinitelyTyped.  

The fix is either done by swapping the `new` and `constructor` keywords, or by replacing an `interface` with a `class`. 

The different instances of the mistake was found using a query written in [QL](https://semmle.com/ql). 
The results of which were manually vetted by me (as there [are](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/material-design-lite/index.d.ts#L121) [reasons](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/di-lite/index.d.ts#L57) for using the "wrong" keyword).

Please fill in this template.

- [x] Use a meaningful title for the pull request. ~~Include the name of the package modified.~~

~~If changing an existing definition:~~
- [ ] ~~Provide a URL to documentation or source code which provides context for the suggested changes:~~
- [ ] ~~If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.~~

### DIFF
--- a/types/microsoft-ajax/index.d.ts
+++ b/types/microsoft-ajax/index.d.ts
@@ -627,7 +627,7 @@ declare namespace Sys {
 
         //#region Constructors
 
-        constructor(): void;
+        new(): void;
 
         //#endregion
 
--- a/types/microsoft-ajax/index.d.ts
+++ b/types/microsoft-ajax/index.d.ts
@@ -2505,7 +2505,6 @@ declare namespace Sys {
         * @see {@link http://msdn.microsoft.com/en-us/library/bb383800(v=vs.100).aspx}
         */
         class ProfileService {
-
             constructor();
 
             //#region Fields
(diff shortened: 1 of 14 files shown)


types/react-native/index.d.ts
```diff
@@ -89,8 +89,8 @@ interface EventSubscription {
  * EventSubscriptionVendor stores a set of EventSubscriptions that are
  * subscribed to a particular event type.
  */
-interface EventSubscriptionVendor {
-    constructor(): EventSubscriptionVendor;
+declare class EventSubscriptionVendor {
+    constructor();
```
alloy: 👍

alloy:
Seems like a future opportunity for a linter rule /cc @orta.

jednano (approved):

orta:
Yeah, maybe a tslint warning when you do:

```ts
interface Foo {
    constructor(): Bar;
}
```

Anytime might be enough?

types/microsoft-ajax/index.d.ts
```diff
@@ -2505,8 +2505,7 @@ declare namespace Sys {
         * @see {@link http://msdn.microsoft.com/en-us/library/bb383800(v=vs.100).aspx}
         */
         class ProfileService {
-
-            new(): ProfileService;
+            constructor();
```
andrewbranch: This one can just be removed, I think.

types/microsoft-ajax/index.d.ts
```diff
@@ -627,7 +627,7 @@ declare namespace Sys {
 
         //#region Constructors
 
-        constructor(): void;
+        new(): void;
```
andrewbranch: `void` is probably wrong, but it wasn't constructable before now sooo... 🤷‍♂ @pjmagee can you advise?

andrewbranch (approved):
How, um, did anything ever work before now? 😱

types/microsoft-ajax/index.d.ts
erik-krogh: Well.. yes.. it is useless. 

But the [linked documentation](https://docs.microsoft.com/en-us/previous-versions/bb907039%28v%3dvs.100%29) states that the constructor exists, even though it is kind of useless.  
So I decided to leave it in.

types/microsoft-ajax/index.d.ts
erik-krogh: Yes, `void` is wrong. 

But to have something more precise than `void` requires having some interface/class describing the instance. And just as with the `ProfileService` class, the constructor in the `Application` class is basically useless.    
And changing `Application` from an `interface` to a `class` causes other type errors.  

So keeping the `void` return-type seemed like the least bad option.  
(An alternative is to make it return `{}`).

types/microsoft-ajax/index.d.ts
andrewbranch: Sure, but a parameterless constructor is implied for classes that don’t specify a different construct signature. It’s not hurting anything, it’s just a no-op.

types/microsoft-ajax/index.d.ts
andrewbranch: Right, I mean, if someone tries to use this, they’re _going_ to have problems. You can’t do anything to `void`. But, this interface literally wasn’t constructable before now, so it seems like nobody is using it or nobody cares. It’s very strange.

PR #39001, opened 2019-10-09
[sharp] Update typings to 0.23

fdebijl:

  - Add `smartSubsample` and `reductionEffort` parameters to Webpoptions: https://github.com/lovell/sharp/issues/1545
  - Remove `cutout` parameter from OverlayOptions: https://github.com/lovell/sharp/blob/master/lib/composite.js#L72 (This parameter is not mentioned in the changelogs, but seemingly not present)
  - Add `premultiplied` parameter to OverlayOptions: https://github.com/lovell/sharp/pull/1835
  - Add `skipBlanks` parameter to TileOptions: https://github.com/lovell/sharp/pull/1687
  - Change `layout` parameter in TileOptions to a union type. I believe a union type is more apt for `layout` since only three string values are valid. `Blend` also uses a 16-string union type, so using one for the layout makes more sense than `string` in my opinion.

### DIFF
--- a/types/sharp/index.d.ts
+++ b/types/sharp/index.d.ts
@@ -210,7 +210,7 @@ declare namespace sharp {
        * @throws {Error} Invalid parameters
        * @returns A sharp instance that can be used to chain operations
        */
-      composite(images: Array<OverlayOptions>): Sharp;
+      composite(images: OverlayOptions[]): Sharp;
 
       //#endregion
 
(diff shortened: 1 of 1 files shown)


Fdebijl:
More testing in my current project revealed that the parameters for the composite method are also outdated, `Create` should be a property of `input` when used instead of a `string` or `Buffer`. See the parameter list at https://sharp.pixelplumbing.com/en/stable/api-composite/ and https://github.com/lovell/sharp/blob/master/lib/composite.js#L72

I changed it into `input?: string | Buffer | {create: Create};`, but maybe that should be a separate type - someone more experienced with TS best practices will have to make that call.

A 100x100px black square being created as intended using the new typings.
_A 100x100px black square being created as intended using the new typings._

andrewbranch (approved):
Looks good, thanks!

PR #39003, opened 2019-10-09
[@reach/router v1.2.1] Fixed return type of navigate function

awwit:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/reach/router/blob/master/src/lib/history.js#L48-L70>

### DIFF
--- a/types/reach__router/index.d.ts
+++ b/types/reach__router/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for @reach/router 1.2.1
+// Type definitions for @reach/router 1.2
 // Project: https://github.com/reach/router
 // Definitions by: Kingdaro <https://github.com/kingdaro>,
 //                 A.Mokhtar <https://github.com/xMokAx>,
--- a/types/reach__router/index.d.ts
+++ b/types/reach__router/index.d.ts
@@ -1,6 +1,8 @@
-// Type definitions for @reach/router 1.2
+// Type definitions for @reach/router 1.2.1
 // Project: https://github.com/reach/router
-// Definitions by: Kingdaro <https://github.com/kingdaro>, A.Mokhtar <https://github.com/xMokAx>
+// Definitions by: Kingdaro <https://github.com/kingdaro>,
+//                 A.Mokhtar <https://github.com/xMokAx>,
+//                 Awwit <https://github.com/awwit>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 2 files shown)


awwit:
@kingdaro @xMokAx Guys, please see the corrections.

types/reach__router/reach__router-tests.tsx
```diff
@@ -40,7 +41,7 @@ render(
             {context => (
                 <>
                     <div>hostname is {context.location.hostname}</div>
-                    <button onClick={() => context.navigate('/')}>
+                    <button onClick={(): Promise<void> => context.navigate('/')}>
```
kingdaro: This return annotation shouldn't be necessary

types/reach__router/reach__router-tests.tsx
```diff
@@ -50,7 +51,7 @@ render(
...
```
kingdaro: same here

types/reach__router/index.d.ts
```diff
@@ -89,7 +91,10 @@ export interface MatchRenderProps<TParams> {
 
 export class Match<TParams> extends React.Component<MatchProps<TParams>> {}
 
-export type NavigateFn = (to: string, options?: NavigateOptions<{}>) => void;
+export interface NavigateFn {
+    (to: string, options?: NavigateOptions<{}>): Promise<void>;
+    (to: number): Promise<void>;
```
kingdaro: could you point to the section in the docs showing where navigate can accept a number? does it also accept options in this case?

itsMapleLeaf (changes requested):

types/reach__router/index.d.ts
awwit: Here: https://github.com/reach/router/blob/master/src/lib/history.js#L49-L51

types/reach__router/reach__router-tests.tsx
awwit: A must, as this is a test to check the return value.

types/reach__router/index.d.ts
kingdaro: 🤔 not sure if I like supporting undocumented features, but I guess this is fine

itsMapleLeaf (approved):

types/reach__router/index.d.ts
awwit: By the way, there is a mention of this in the documentation :)
https://github.com/reach/router/blob/master/website/src/markdown/api/navigate.md#to

PR #39008, opened 2019-10-09
feat(koa): Move custom properties into DefaultContext

mrcljx:
This allows having a strict `Koa<{}, {}>` where an arbitrary `ctx.unknownProperty` will raise `"Property 'unknownProperty' does not exist on type"`.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: **no url**
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header. **no change, change to allow stronger typing (might be breaking in some codebases)**

---

```
const app = new Koa<{}, { x: number }>();

app.use((ctx, next) => {
    console.log(ctx.x);
    console.log(ctx.foo); // Error:(l, c) TS2339: Property 'foo' does not exist on type 'ParameterizedContext<{}, { x: number; }>'.
})
```

If no explicit types are used (`const app = new Koa();`) the `DefaultContext` is used (which has an index signature).

### DIFF
--- a/types/koa-bouncer/koa-bouncer-tests.ts
+++ b/types/koa-bouncer/koa-bouncer-tests.ts
@@ -1,43 +1,35 @@
 import Koa = require("koa");
-import Router = require("koa-router");
 import bouncer = require("koa-bouncer");
 
 const app = new Koa();
 
 app.use(bouncer.middleware());
-const router = new Router()
 
-router.post('/users', async (ctx) => {
+app.use(async ctx => {
     ctx.validateBody('uname')
         .required('Username required')
         .isString()
-        .trim()
+        .trim();
 
     ctx.validateBody('email')
         .optional()
         .isString()
         .trim()
-        .isEmail('Invalid email format')
+        .isEmail('Invalid email format');
 
     ctx.validateBody('password1')
         .required('Password required')
         .isString()
-        .isLength(6, 100, 'Password must be 6-100 chars')
+        .isLength(6, 100, 'Password must be 6-100 chars');
 
     ctx.validateBody('password2')
         .required('Password confirmation required')
         .isString()
-        .eq(ctx.vals.password1, 'Passwords must match')
+        .eq(ctx.vals.password1, 'Passwords must match');
 
-    ctx.validateBody('age')
-        .gte(18, 'Must be 18 or older')
+    ctx.validateBody('age').gte(18, 'Must be 18 or older');
 
-    console.log(ctx.vals)
-})
-
-app
-    .use(router.routes())
-    .use(router.allowedMethods());
-    
+    ctx.body = ctx.vals;
+});
 
 app.listen(80);
--- a/types/koa-log/index.d.ts
+++ b/types/koa-log/index.d.ts
@@ -24,7 +24,7 @@ declare namespace koaLog {
 
         function status(ctx: Koa.BaseContext): number;
 
-        function token(name: string, fn: (ctx: Koa.BaseContext) => string): void;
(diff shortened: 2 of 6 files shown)


types/koa-bouncer/koa-bouncer-tests.ts
```diff
@@ -1,43 +1,35 @@
...
 import bouncer = require("koa-bouncer");
 
 const app = new Koa();
 
 app.use(bouncer.middleware());
-const router = new Router()
 
-router.post('/users', async (ctx) => {
+app.use(async ctx => {
     ctx.validateBody('uname')
```
sirlantis: `validateBody` was actually `any`, because this module augments `Koa.Context`, but `Router` passes in `ctx: ParameterizedContext` (`Koa.Context` extends `ParameterizedContext`).

So I took Router out of the equation here (`koa-bouncer` does not depend on it).

types/koa-log/koa-log-tests.ts
```diff
@@ -1,10 +1,8 @@
 import Koa = require('koa');
 import koaLogger = require('koa-log');
 
-koaLogger.morgan.token('request-body', (ctx: Koa.BaseContext) => JSON.stringify(ctx.request.body));
```
sirlantis:
```
interface ExtendableContext extends BaseContext {
        app: Application;
        request: Request;
```

The `request` property is not defined on `BaseContext` (see above). Because of the index signature, `ctx.request` was `any`, and `ctx.request.body` was okay. Without any body parser middleware, `ctx.request.body` isn't okay either!

types/koa-router/koa-router-tests.ts
```diff
@@ -78,7 +78,7 @@ app.use(async (ctx: Koa.ParameterizedContext<MyState, MyContext>, next) => {
 
 const router3 = new Router();
 router3.get('/', (ctx) => {
-    ctx.foo = "bar";
+    ctx.state.foo = 'bar';
```
sirlantis: The rest of the code only operates on `ctx.state.foo` - this seems to be a bug found by the more strict typing.

types/koa/index.d.ts
```diff
@@ -680,10 +685,6 @@ declare namespace Application {
          * Default error handling.
          */
         onerror(err: Error): void;
-        /**
-         * Custom properties.
-         */
-        [key: string]: any;
```
martijnthe: 👍

types/koa-log/index.d.ts
```diff
@@ -24,7 +24,7 @@ declare namespace koaLog {
 
         function status(ctx: Koa.BaseContext): number;
 
-        function token(name: string, fn: (ctx: Koa.BaseContext) => string): void;
+        function token(name: string, fn: (ctx: Koa.ParameterizedContext) => string): void;
```
havenchyk: LGTM

andrewbranch:
Interpreting the 👍 and “LGTM” from owners as an approval. (If you use “Approve changes” in GitHub’s UI, the PR merging process will get streamlined!)

havenchyk:
@andrewbranch this change touches parts that I'm not aware of, I could only  check code usage in 2 lines  of change, so it's not a criteria for a final approval imho.

Do you think it's better to use GitHub approve button? 😄

andrewbranch:
@havenchyk I guess that’s fair! Since it didn’t have an official approval, I took a closer look than I would have otherwise, but I’m not really familiar with Koa, so it was mostly a sanity check. But given that it’s been a week and you were the only owner who showed up, I was skeptical that we were going to get a “final approval.” 😬

PR #39011, opened 2019-10-09
Fix for aria-label ToggleSmall of carbon-components-react 

Brenne:
Change aria-label of ToggleSmall of carbon-components-react to Kebab Case to be consistent with implementation https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/ToggleSmall/ToggleSmall.js#L121

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/ToggleSmall/ToggleSmall.js#L121

If removing a declaration:

### DIFF
--- a/types/carbon-components-react/lib/components/ToggleSmall/ToggleSmall.d.ts
+++ b/types/carbon-components-react/lib/components/ToggleSmall/ToggleSmall.d.ts
@@ -6,7 +6,7 @@ interface InheritedProps extends
     Omit<ReactInputAttr, ExcludedAttributes>,
     RequiresIdProps
 {
-    ["aria-Label"]: NonNullable<React.AriaAttributes["aria-label"]>,
+    ["aria-label"]: NonNullable<React.AriaAttributes["aria-label"]>,
 }
 
 export interface ToggleSmallProps extends InheritedProps {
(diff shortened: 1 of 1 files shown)


types/carbon-components-react/lib/components/ToggleSmall/ToggleSmall.d.ts
```diff
@@ -6,7 +6,7 @@ interface InheritedProps extends
     Omit<ReactInputAttr, ExcludedAttributes>,
     RequiresIdProps
 {
-    ariaLabel: NonNullable<React.AriaAttributes["aria-label"]>,
+    ["aria-Label"]: NonNullable<React.AriaAttributes["aria-label"]>,
```
kalbert312: Should this be `aria-label`(lowercase L) or does it matter?

Brenne: yes you're correct. I'll push my changes

kalbert312 (approved):

types/carbon-components-react/lib/components/ToggleSmall/ToggleSmall.d.ts
```diff
@@ -6,7 +6,7 @@ interface InheritedProps extends
     Omit<ReactInputAttr, ExcludedAttributes>,
     RequiresIdProps
 {
-    ariaLabel: NonNullable<React.AriaAttributes["aria-label"]>,
+    ["aria-label"]: NonNullable<React.AriaAttributes["aria-label"]>,
```
andrewbranch: FYI, the square brackets aren’t needed.

andrewbranch (approved):

PR #39012, opened 2019-10-09
multer-s3 - extended req and file definition to work with @koa/multer types

abelkius:
I extended types in this definition to also work with `koa-multer` package. The only difference is the type for `req` that is passed through and `S3Storage` type that also differs just because of the `req` that is passed through. 

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/koa-multer/index.d.ts

### DIFF
--- a/types/multer-s3/index.d.ts
+++ b/types/multer-s3/index.d.ts
@@ -8,11 +8,12 @@
 // TypeScript Version: 2.3
 
 import * as AWS from "aws-sdk";
-import { IncomingMessage } from 'http';
+import { IncomingMessage } from "http";
 import { StorageEngine } from "multer";
+import { StorageEngine as KoaStorageEngine } from "koa-multer";
 
 type MulterRequest = Express.Request | IncomingMessage;
-type MulterFile = Express.Multer.File | File;
+type MulterFile = Express.Multer.File;
 
 interface Options {
     s3: AWS.S3;
@@ -45,7 +46,7 @@ declare global {
 }
 
 interface S3Storage {
-    (options?: Options): StorageEngine;
+    (options?: Options): StorageEngine & KoaStorageEngine;
 
     AUTO_CONTENT_TYPE(
         req: MulterRequest,
--- a/types/multer-s3/index.d.ts
+++ b/types/multer-s3/index.d.ts
@@ -9,7 +9,6 @@
 
 import * as AWS from "aws-sdk";
 
-
 declare namespace MulterS3 {
     interface File {
         /** Field name specified in the form */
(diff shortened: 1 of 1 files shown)


andrewbranch (changes requested):
Hm, I think I see the problem you’re trying to solve, but this isn’t the right approach. Really, the type of `req` should be generic. While the status quo is wrong, and this PR is arguably a step in the right direction, it would cause breakages for people who _are_ using Express and relying on the types to be express-like.

Unfortunately I'm not familiar enough with these packages to be able to tell you exactly what this would look like. If you can post an example of how this is used with Koa, that would be helpful.

abelkius:
Hi @andrewbranch, thank you a ton for your review. I updated the PR and now I use generics in the exported interface. The only problem that appeared with this approach is that I needed to disable the `no-unnecessary-generics` in few places. I hope it is not a problem, because I don't think I should actually use `unknown` type in there. What do you think?

types/multer-s3/index.d.ts
```diff
@@ -3,54 +3,60 @@
...
-        req: Express.Request,
-        file: Express.Multer.File,
-        callback: (error: any, mime?: string, stream?: NodeJS.ReadableStream) => void): void;
-    DEFAULT_CONTENT_TYPE(
-        req: Express.Request,
-        file: Express.Multer.File,
-        callback: (error: any, mime?: string) => void): void;
+    interface S3Storage {
+        // tslint:disable-next-line:no-unnecessary-generics
+        <E, R>(options?: Options<R>): E;
```
weswigham, line 47: Why is the return type not still `StorageEngine`? I don't think the caller can change the fact that a `S3Storage` instance is returned.

types/multer-s3/index.d.ts
```diff
@@ -3,54 +3,60 @@
...
+        /** Location of the uploaded file (DiskStorage) */
+        path: string;
+        /** A Buffer of the entire file (MemoryStorage) */
+        buffer: Buffer;
     }
-}
 
-interface S3Storage {
-    (options?: Options): StorageEngine;
+    interface Options<R> {
```
weswigham, line 34: Consider adding `R = Express.Request` for backwards compatibility.

types/multer-s3/index.d.ts
```diff
@@ -3,54 +3,60 @@
...
-        file: Express.Multer.File,
-        callback: (error: any, mime?: string, stream?: NodeJS.ReadableStream) => void): void;
-    DEFAULT_CONTENT_TYPE(
-        req: Express.Request,
-        file: Express.Multer.File,
-        callback: (error: any, mime?: string) => void): void;
+    interface S3Storage {
+        // tslint:disable-next-line:no-unnecessary-generics
+        <E, R>(options?: Options<R>): E;
+        AUTO_CONTENT_TYPE<R>(
```
weswigham, line 48: In these cases, you can actually say that `req` is `any`. The implementation quite literally does not inspect the parameter, and does not care a whit about its value. That should make it so you don't need the type parameters here.

weswigham (changes requested):

PR #39018, opened 2019-10-09
Remove husky and precise-commits

orta:
Removes the husky hooks, and turns off the commit generator.


orta:
Please leave feedback here if you see problems around the remove githook, I'll do what I can to try address them

PR #39022, opened 2019-10-10
Rebass: Add tx prop to BoxKnownProps

tpict:
The `tx` prop is used to modify the lookup on the theme object when a `variant` is specified. Handy if you want to write an extension of a Rebass component with its own variants.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/rebassjs/rebass/blob/c8aaa05e7207f2304897e4f4a78071827dd085fe/packages/reflexbox/src/index.js#L16-L25

### DIFF
--- a/types/rebass/index.d.ts
+++ b/types/rebass/index.d.ts
@@ -57,7 +57,7 @@ interface BoxKnownProps
         StyledSystem.AlignSelfProps,
         SxProps {
     variant?: StyledSystem.ResponsiveValue<string>;
-    tx?: number | string;
+    tx?: string;
 }
 export interface BoxProps
     extends BoxKnownProps,
--- a/types/rebass/index.d.ts
+++ b/types/rebass/index.d.ts
@@ -57,6 +57,7 @@ interface BoxKnownProps
         StyledSystem.AlignSelfProps,
         SxProps {
     variant?: StyledSystem.ResponsiveValue<string>;
+    tx?: number | string;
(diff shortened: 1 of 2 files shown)


tpict:
Nor I 😢what's the DefinitelyTyped policy on undocumented props?

Ilaiwi:
I've been using this feature as well. Not sure why it is not documented though

rhysd (approved):
Looks good. Though I'm not understanding when this value becomes number.

tpict:
I was considering the possibility that a given set of variants might be keyed by number, but since the `tx` prop corresponds to the _name_ of a variant, it should probably be string only, so I'll change that. Thanks!

andrewbranch:
> what's the DefinitelyTyped policy on undocumented props?

There’s no policy; it’s up to the owners of the package. The thing you should consider is if the target package follows semver, but doesn’t consider internal/undocumented API changes as breaking, they could decide to change or remove it in a patch version. If that happens, it would be fairly easy for users to get their typings and JS out of sync even though they appear to match in major/minor version.

In this scenario, you’d make another update here, and it would be published as a patch. Users would have to update to that patch in order to have safe typings again. So, it’s not the end of the world, but could be a minor risk or nuisance if things go poorly. Your job here as package owners is simply to weigh the potential costs and benefits.

rhysd (approved):

PR #39025, opened 2019-10-10
[express-serve-static-core] Small improvements

vonagam:
List of changes:

1) In `CookieOptions` (which are options for [`cookie`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f58986ca20210f8e6612942db077b9228740a315/types/cookie/index.d.ts#L6)) for [`Response.cookie`](https://github.com/expressjs/express/blob/e1b45ebd050b6f06aa38cda5aaf0c21708b0c71e/lib/response.js#L831): `expires` can't be a boolean, only a date, `secure` can't be `'auto'` string, only a boolean, `encode` function should return a string.

2) In `Request`: there is no `clearCookie` method, it is present only in `Response`.

3) In `Response`: [`header`](https://github.com/expressjs/express/blob/e1b45ebd050b6f06aa38cda5aaf0c21708b0c71e/lib/response.js#L754) method just like `set` method (because it is an alias) accepts an array as a second argument.

4) In `Request`, `Response` and `Application`: in methods which return `this` set return value to `this` instead of a class.

Last one needed because if you want to extend for example `Application` with a compatible interface like this:
```ts
interface A { fun(): A }
interface B { fun(): this }
interface A extends B {}
```
You will get an error saying `'A' is assignable to the constraint of type 'this', but 'this' could be instantiated with a different subtype of constraint 'A'` on first line. If instead interface A returns `this` in its method there is no error.

* Also in `Response`: [`get`](https://github.com/expressjs/express/blob/e1b45ebd050b6f06aa38cda5aaf0c21708b0c71e/lib/response.js#L788) method should have `number | string | string[] | undefined` as a return type (just like `http.OutgoingMessage.getHeader` as it is an alias) instead of [simply `string`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f58986ca20210f8e6612942db077b9228740a315/types/express-serve-static-core/index.d.ts#L747), but as it would be a breaking change decided not to include it.

### DIFF
--- a/types/express-serve-static-core/index.d.ts
+++ b/types/express-serve-static-core/index.d.ts
@@ -174,12 +174,12 @@ export interface Router extends IRouter { }
 export interface CookieOptions {
     maxAge?: number;
     signed?: boolean;
-    expires?: Date | boolean;
+    expires?: Date;
     httpOnly?: boolean;
     path?: string;
     domain?: string;
-    secure?: boolean | 'auto';
-    encode?: (val: string) => void;
+    secure?: boolean;
+    encode?: (val: string) => string;
     sameSite?: boolean | string;
 }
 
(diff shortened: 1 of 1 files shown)


andrewbranch (approved):
Looks right to me. Thanks!

PR #39026, opened 2019-10-10
feat(@types/emoji-js) create types

jasondibenedetto:

### DIFF
--- a/types/emoji-js/emoji-js-tests.ts
+++ b/types/emoji-js/emoji-js-tests.ts
@@ -1,0 +1,24 @@
+import emojijs = require('emoji-js');
+
+const emoji = new emojijs.EmojiConvertor();
+
+emoji.replace_colons(':smile:');
+
+emoji.replace_unified('\u{1F604}');
+
+emoji.text_mode = true;
+
+emoji.include_title = true;
+
+emoji.img_sets.apple.path = 'http://my-cdn.com/emoji-apple-64/';
+
+emoji.img_sets.apple.sheet = 'http://my-cdn.com/emoji-apple-sheet-64.png';
+
+emoji.use_sheet = true;
+
+emoji.addAliases({
+    doge: '1f415',
+    cat: '1f346',
+});
+
+emoji.removeAliases(['doge', 'cat']);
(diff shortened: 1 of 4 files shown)


types/emoji-js/index.d.ts
```diff
@@ -0,0 +1,47 @@
+// Type definitions for emoji-js 3.4
+// Project: https://github.com/iamcal/js-emoji
+// Definitions by: Jason Di Benedetto <https://github.com/me>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
+export default class EmojiConvertor {
```
armanio123: I don't think the library exports a `default`. Please remove this.

types/emoji-js/index.d.ts
```diff
@@ -0,0 +1,47 @@
...
+    constructor();
```
armanio123: There's no point on defining an empty constructor. You can remove this line.

types/emoji-js/index.d.ts
```diff
@@ -0,0 +1,47 @@
+// Type definitions for emoji-js 3.4
+// Project: https://github.com/iamcal/js-emoji
+// Definitions by: Jason Di Benedetto <https://github.com/me>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
```
armanio123: Library appears to be an UMD module. You also need to define `export as namespace EmojiConvertor;`

types/emoji-js/emoji-js-tests.ts
```diff
@@ -0,0 +1,24 @@
+import EmojiConvertor from 'emoji-js';
```
armanio123: Since there's no `default`, you need to change this import to `import EmojiConvertor = require('emoji-js');`

armanio123 (changes requested):

jasondibenedetto:
Thanks for the feedback @armanio123. I have amended my commit with your suggestions.

armanio123 (approved):

PR #39037, opened 2019-10-10
[react-router] add string[] to matchPath and useRouteMatch arg types

1000hz:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://reacttraining.com/react-router/web/api/matchPath/props>

### DIFF
--- a/types/react-router/index.d.ts
+++ b/types/react-router/index.d.ts
@@ -127,7 +127,7 @@ export interface match<Params extends { [K in keyof Params]?: string } = {}> {
 // Omit taken from https://github.com/Microsoft/TypeScript/issues/28339#issuecomment-467220238
 export type Omit<T, K extends keyof T> = T extends any ? Pick<T, Exclude<keyof T, K>> : never;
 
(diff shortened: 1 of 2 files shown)


types/react-router/index.d.ts
```diff
@@ -156,5 +156,5 @@ export function useLocation<S = H.LocationState>(): H.Location<S>;
 export function useParams<Params extends { [K in keyof Params]?: string } = {}>(): { [p in keyof Params]: string };
 
 export function useRouteMatch<Params extends { [K in keyof Params]?: string } = {}>(
-    path?: string | RouteProps,
+    path?: string | string[] | RouteProps,
```
1000hz: alternatively, we could change this to
```
path?: RouteProps | RouteProps["path"],
```
(and same for the `props` parameter of `matchPath`)

johnnyreilly: Would you prefer to make that change? Would it make a difference?

johnnyreilly (approved):

types/react-router/index.d.ts
1000hz: Nah, just something that crossed my mind as I was doing this. I’ll leave it as is if no one has any strong opinions.

PR #39055, opened 2019-10-11
feat: Add `array.from`

ExE-Boss:
Please fill in this template.

If adding a new definition:

---

review?(@ljharb, @andrewbranch)

### DIFF
--- a/types/array.from/.editorconfig
+++ b/types/array.from/.editorconfig
@@ -1,0 +1,3 @@
+# This package uses tabs
+[*]
+indent_style = tab
--- a/types/array.from/array.from-tests.ts
+++ b/types/array.from/array.from-tests.ts
@@ -1,0 +1,12 @@
+import ArrayFrom = require('array.from');
+
+/**
+ * The `expectType` function from https://www.npmjs.com/package/tsd,
+ * except instead of returning `void`, it returns `T`.
+ */
+declare function expectType<T>(t: T): T;
+
+expectType<typeof Array.from>(ArrayFrom);
+expectType<typeof Array.from>(ArrayFrom.implementation);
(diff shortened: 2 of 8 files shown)


types/array.from/array.from-tests.ts
```diff
@@ -0,0 +1,12 @@
+import ArrayFrom = require('array.from');
+
+/**
+ * The `expectType` function from https://www.npmjs.com/package/tsd,
+ * except instead of returning `void`, it returns `T`.
+ */
+declare function expectType<T>(t: T): T;
```
ExE-Boss: I have to use this because the type needed in the `// $ExpectType` would otherwise be very long and complex and error‑prone.

types/array.from/.editorconfig
```diff
@@ -0,0 +1,3 @@
+# This package uses tabs
+[*]
+indent_style = tab
```
andrewbranch: Is this necessary?

ExE-Boss: Yes.

Otherwise Prettier will convert tabs to spaces, which I don’t want.

andrewbranch (changes requested):
Thanks for the contribution, but please don’t customize the formatter to your preferences.

ExE-Boss:
@ PranavSenthilnathan said&nbsp;in&nbsp;<https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34894#discussion_r279131192>:
> It's fine for specific libraries to use tabs if they feel like it

andrewbranch:
Hm, I missed the fact that we disabled enforcement of prettier after initially introducing it at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/35672. I guess until it’s built into the build pipeline, people can continue to do whatever they want.

PR #39058, opened 2019-10-11
feat: Add `define‑properties`

ExE-Boss:
Please fill in this template.

If adding a new definition:

---

review?(@ljharb, @andrewbranch)

### DIFF
--- a/types/define-properties/.editorconfig
+++ b/types/define-properties/.editorconfig
@@ -1,0 +1,3 @@
+# This package uses tabs
+[*]
+indent_style = tab
--- a/types/define-properties/index.d.ts
+++ b/types/define-properties/index.d.ts
@@ -1,6 +1,6 @@
 // Type definitions for define-properties 1.1
 // Project: https://github.com/ljharb/define-properties#readme
-// Definitions by: ExE Boss <https://github.com/me>
+// Definitions by: ExE Boss <https://github.com/ExE-Boss>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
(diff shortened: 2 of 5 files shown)


types/define-properties/index.d.ts
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Defines new properties in `map` as non-enumerable if they don't already
+ * exist on `object`.
+ *
+ * @param object The object to define non-enumerable properties on.
+ * @param map The map of newly defined properties.
+ * @param predicates The optional predicates map, return `true` to override existing properties on `object`.
+ */
+declare function defineProperties<K extends keyof any>(
```
ljharb: why is this any? The K must be a Property Key, ie a string/number/symbol/bigint

ljharb (approved):

types/define-properties/index.d.ts
ExE-Boss: It extends `keyof any`, which resolves to `string | number | symbol`.

Also, last I checked, `bigint` couldn’t be used to index properties directly in TypeScript (unless they changed that).

ljharb: true, that hasn’t been fixed yet.

types/define-properties/.editorconfig
```diff
@@ -0,0 +1,3 @@
+# This package uses tabs
+[*]
+indent_style = tab
```
andrewbranch: Is this necessary?

ExE-Boss: Yes.

Otherwise, Prettier will&nbsp;convert tabs&nbsp;to&nbsp;spaces, which&nbsp;I&nbsp;don’t&nbsp;want.

PR #39069, opened 2019-10-11
Fixed Fabric Object animation function to include correct parameters

KyleKaniecki:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: http://fabricjs.com/fabric-intro-part-2

### DIFF
--- a/types/fabric/fabric-impl.d.ts
+++ b/types/fabric/fabric-impl.d.ts
@@ -368,10 +368,10 @@ interface IObjectAnimation<T> {
 	/**
 	 * Animates object's properties
 	 * object.animate({ left: ..., top: ... }, { duration: ... });
-	 * @param properties Properties to animate
-	 * @param value Options object
+	 * @param properties Properties to animate with values to animate to
+	 * @param options The animation options
 	 */
-	animate(properties: any, options?: IAnimationOptions): Object;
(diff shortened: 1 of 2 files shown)


types/fabric/fabric-impl.d.ts
```diff
@@ -3506,8 +3506,19 @@ export class Object {
...
+	 */
+	animate(property: string, value: number | string, options?: IAnimationOptions): Object;
+	/**
+	 * Animates object's properties
+	 * object.animate({ left: ..., top: ... }, { duration: ... });
+	 * @param properties Properties to animate
+	 * @param value Options object
 	 */
-	animate(): Object;
+	animate(properties: any, options?: IAnimationOptions): Object;
```
NotWoods: Ideally it would be nice to have an interface specifying the valid type. Otherwise, `object` would be better than `any` here.

KyleKaniecki: Ok, I believe I got the right interface for the parameter. Had to do a little digging through source to make sure, but I believe that [only numbers or strings are _really_ allowed.](https://github.com/fabricjs/fabric.js/blob/master/src/mixins/animation.mixin.js#L161)

NotWoods (approved):

PR #39070, opened 2019-10-11
fix passport-google-oauth20 additional options.

Acerbic:
Using TS module augmentation, allow google-specific authorization
options to be used with "authorize" and "authenticate" passport
calls, when "google" strategy is used.

1. Removed "accessType" from strategy initialization options, as it is not being used by [actual constructor code](https://github.com/jaredhanson/passport-google-oauth2/blob/723e8f3e8e711275f89e0163e2c77cfebae33f25/lib/strategy.js#L47).
2. Added interface `AuthenticateOptionsGoogle` which extends basic authentication options interface from "passport" package with Google-specific fields that are being picked up by [strategy (code)](https://github.com/jaredhanson/passport-google-oauth2/blob/723e8f3e8e711275f89e0163e2c77cfebae33f25/lib/strategy.js#L138).
3. With TypeScript (module augmentation)[http://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation] added an override for functions `authenticate` and `authorize` when they are being used with this strategy directly.

Resulting change:
- additional options, like "accessType" could be used without explicit types assertions, when `authenticate` or `authorize` are called with "google" strategy.
- for other use cases (i.e. if `authenticate` is called with array of strategies names, instead of a single name), `AuthenticateOptionsGoogle` interface assertion could be used to provide additional options.

### DIFF
--- a/types/passport-google-oauth20/index.d.ts
+++ b/types/passport-google-oauth20/index.d.ts
@@ -3,6 +3,7 @@
 // Definitions by: Yasunori Ohoka <https://github.com/yasupeke>
 //                 Eduard Zintz <https://github.com/ezintz>
 //                 Tan Nguyen <https://github.com/ngtan>
+//                 Gleb Varenov <https://github.com/acerbic>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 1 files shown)


andrewbranch:
Looks reasonable to me, but it would be good to see a test demonstrating the module augmentation. In the meantime, I’ll ping the maintainers again, along with the most recent contributor to @types/passport itself. 👋  @yasupeke @ezintz @ngtan @vaskevich Thanks!

Acerbic:
Not sure how to put it in your testing framework, but with this pull request the following code no longer gives an error and moreso, actually works - provides "refreshToken" from Google, in addition to "accessToken".
```typescript
app.use(
   passport.authenticate("google", {scope: ["profile"], session: false, accessType: "offline"}),
   (req, res) => { /*...*/ }
);
```

weswigham:
Seems fine to me, it's been another week of nobody looking at it - I'm merging this as-is 🐱

PR #39097, opened 2019-10-13
it-pushable: initial definition

yoursunny:

### DIFF
--- a/types/it-pushable/index.d.ts
+++ b/types/it-pushable/index.d.ts
@@ -1,0 +1,21 @@
+// Type definitions for it-pushable 1.3
+// Project: https://github.com/alanshaw/it-pushable
+// Definitions by: Junxiao Shi <https://github.com/yoursunny>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.2
+
+interface Options {
+    onEnd?: (err?: Error) => any;
+    writev?: boolean;
+}
+
+interface Pushable<T, R> extends AsyncIterable<R> {
+    push(value: T): this;
+    end(err?: Error): this;
+}
+
+declare function pushable<T>(options: Options & { writev: true }): Pushable<T, T[]>;
+
+declare function pushable<T>(options?: Options): Pushable<T, T>;
+
+export = pushable;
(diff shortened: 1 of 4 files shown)


types/it-pushable/index.d.ts
```diff
@@ -0,0 +1,23 @@
+// Type definitions for it-pushable 1.3
+// Project: https://github.com/alanshaw/it-pushable
+// Definitions by: Junxiao Shi <https://github.com/yoursunny>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.2
+
+/// <reference types="node" />
```
andrewbranch: I’m probably overlooking something obvious, but why is this needed?

yoursunny: It's needed for `AsyncIterable`. I'm going to change it to a "lib" reference.

types/it-pushable/index.d.ts
```diff
@@ -0,0 +1,23 @@
+// Type definitions for it-pushable 1.3
+// Project: https://github.com/alanshaw/it-pushable
+// Definitions by: Junxiao Shi <https://github.com/yoursunny>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.2
+
+/// <reference lib="esnext.asynciterable" />
```
andrewbranch: Hm, so, I talked to the team about this, and I think we can’t have this here. It will pull those types in for anyone who installs these typings, even if their own lib settings don’t include it. (Granted, it seems unlikely that this is a real problem in practice, since anyone installing these typings _probably_ has a runtime that supports async iterable.) The right thing to do is probably to put this in the tsconfig `lib` section (for some reason I thought that DefinitelyTyped _only_ let you use `es6` and `dom`, but I might be misremembering), or, if that doesn’t work, go back to the `@types/node` reference since Node ships with an AsyncIterable implementation.

andrewbranch (changes requested):

types/it-pushable/index.d.ts
yoursunny: Moved to `tsconfig.json`. `npm test` passes.

yoursunny:
@typescript-bot Can you tag this as "waiting for review"?
I already addressed the comments and need @andrewbranch 's review.

andrewbranch (approved):

PR #39098, opened 2019-10-14
Adds two private methods used when building custom adapters

mfeckie:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/emberjs/data/blob/master/packages/adapter/addon/rest.js#L1142-L1179

This change adds two private methods that may be overridden by folks building custom adapters.

Feels a bit off to have their return types be `any` and `any[]`, but this seems accurate based on the implementation.

### DIFF
--- a/types/ember-data/index.d.ts
+++ b/types/ember-data/index.d.ts
@@ -1482,6 +1482,10 @@ export namespace DS {
          * Determines the pathname for a given type.
          */
         pathForType<K extends keyof ModelRegistry>(modelName: K): string;
+
+        private parseErrorResponse(responseTest: string): any;
+
+        private normalizeErrorResponse(status: number, headers: any, payload: any): any[];
     }
     /**
      * ## Using Embedded Records


chriskrycho:
Hey @mfeckie, thanks for the PR. Two notes:

1. We very assiduously avoid ever shipping private APIs in these types, so I’m going to have to reject this one. (There is a workaround for your needs, which I’ll share in a follow-up comment this evening. Have a wedding to go to this morning, or I’d do it now.)
2. These won’t do what you want anyway because of how privacy works in TS. You can see the problem in [this playground]; subclasses cannot access private members of their parent class.

If you want to add the private APIs to the Ember Data types, you should be able to use [interface merging] to define those extensions in your own projects. We’ve designed our type declarations to support that, so if you find one that doesn’t; that’s a bug, please let us know!

[this playground]: https://www.typescriptlang.org/play/#code/CYUwxgNghgTiAEkoGdnwAqxAOwC7wG8AoeeABxgEsA3KXBZAewFsQBZEXAC0eAApYAcwBc8ZLirZBASlEBXbAGtsjAO7YA3EQC+RIklTwAwl0oRg8EAA962YGkxw8hEolPmO3Xn2kvSpZDkyEBgAOiZWTx5+AHJcOhDkGOktUl1tIA
[interface merging]: https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-interfaces

mike-north (changes requested):
see @chriskrycho's feedback above

mfeckie:
Thanks folks, I'll close this PR

PR #39102, opened 2019-10-14
mongodb: make `query` parameter to `distinct` optional

LinusU:
Please fill in this template.

  - https://github.com/mongodb/node-mongodb-native/blob/0463f7838dbfc0740f6fa52174b96241eb0c3980/lib/collection.js#L1516
  - https://github.com/mongodb/node-mongodb-native/pull/2162

### DIFF
--- a/types/mongodb/index.d.ts
+++ b/types/mongodb/index.d.ts
@@ -928,8 +928,9 @@ export interface Collection<TSchema = Default> {
     deleteOne(filter: FilterQuery<TSchema>, options?: CommonOptions & { bypassDocumentValidation?: boolean }): Promise<DeleteWriteOpResultObject>;
     deleteOne(filter: FilterQuery<TSchema>, options: CommonOptions & { bypassDocumentValidation?: boolean }, callback: MongoCallback<DeleteWriteOpResultObject>): void;
     /** http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#distinct */
+    distinct(key: string, callback: MongoCallback<any>): void;
     distinct(key: string, query: FilterQuery<TSchema>, callback: MongoCallback<any>): void;
-    distinct(key: string, query: FilterQuery<TSchema>, options?: { readPreference?: ReadPreference | string, maxTimeMS?: number, session?: ClientSession }): Promise<any>;
+    distinct(key: string, query?: FilterQuery<TSchema>, options?: { readPreference?: ReadPreference | string, maxTimeMS?: number, session?: ClientSession }): Promise<any>;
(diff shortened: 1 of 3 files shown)


CaselIT (approved):

CaselIT:
thanks for also making the pr to update the documentation

types/mongodb/index.d.ts
```diff
@@ -928,8 +928,9 @@ export interface Collection<TSchema = Default> {
     deleteOne(filter: FilterQuery<TSchema>, options?: CommonOptions & { bypassDocumentValidation?: boolean }): Promise<DeleteWriteOpResultObject>;
     deleteOne(filter: FilterQuery<TSchema>, options: CommonOptions & { bypassDocumentValidation?: boolean }, callback: MongoCallback<DeleteWriteOpResultObject>): void;
     /** http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#distinct */
+    distinct(key: string, callback: MongoCallback<any>): void;
```
albertossilva: Would you please specify the type of the callback at least? or use generic from distinct or the Collection?

types/mongodb/index.d.ts
```diff
@@ -928,8 +928,9 @@ export interface Collection<TSchema = Default> {
...
     distinct(key: string, query: FilterQuery<TSchema>, callback: MongoCallback<any>): void;
-    distinct(key: string, query: FilterQuery<TSchema>, options?: { readPreference?: ReadPreference | string, maxTimeMS?: number, session?: ClientSession }): Promise<any>;
+    distinct(key: string, query?: FilterQuery<TSchema>, options?: { readPreference?: ReadPreference | string, maxTimeMS?: number, session?: ClientSession }): Promise<any>;
```
albertossilva: Last... may you please write some tests?

albertossilva (changes requested):

types/mongodb/index.d.ts
LinusU: The callback can return `any` as you can specify any key and it can any data-type in the database.

We could potentially make it check if it can extract `TSchema[key]`, but as this PR doesn't touch the callback (this line is just duplicated from the ones below) I think that that belongs in a separate PR.

types/mongodb/index.d.ts
LinusU: ✔️

types/mongodb/index.d.ts
LinusU: Actually, more correct is probably `any[]`. But I'd rather submit that as a second PR as it isn't related to the changes I did now...

LinusU:
ping @albertossilva, could we get this merged? ☺️

albertossilva (approved):

PR #39107, opened 2019-10-14
[dojo] Use gender-neutral pronouns

bensalilijames:
Continuation of #39053 since Dojo types were large enough to make CI run out of memory. This commit fixes up that single remaining package.

### DIFF
--- a/types/dojo/dojox.css3.d.ts
+++ b/types/dojo/dojox.css3.d.ts
@@ -187,8 +187,8 @@ declare namespace dojox {
              */
             expand(args: Object): any;
             /**
-             * Returns an animation that flips an element around his y axis.
-             * Flips an element around his y axis. The default is a 360deg flip
+             * Returns an animation that flips an element around its y axis.
+             * Flips an element around its y axis. The default is a 360deg flip
              * but it is possible to run a partial flip using args.whichAnims.
              *
              * @param args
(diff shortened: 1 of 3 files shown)


antoinerousseau (approved):

antoinerousseau:
Looks like the issue remain, even when ran alone...

Haroenv (approved):

benhjames:
Having a quick experiment, I found that Node 8.16.2 (what CI uses) runs out of memory, but Node 12.12.0 runs ok. When I get some more time I can dig deeper into why this might be.

michael-vansickle (approved):

RyanCavanaugh:
I'm going to bypass CI on this one since it's only 3 files.

Also 👀

> the node is flipped around his center

dojo authors may want to start looking into the OOM.

sandersn:
@benhjames 12.12.0 fails for me locally as well. Where did you get a passing run with it?

@types/dojo is rarely updated and little used these days so I'm not sure it's worth anybody's time to track down the problem. On the other hand, this means that having an updatable types/dojo isn't worth that much either.

benhjames:
I managed to get a passing run locally with 12.2.0 (via `nvm use 12 && node node_modules/types-publisher/bin/tester/test.js --run-from-definitely-typed`).

I tested with five runs on each of Node 8 and Node 12 switching between them between each run - all five on Node 8 failed with OOM, and all five on Node 12 passed. Perhaps Node 12 makes it less likely to fail (maybe there's some optimisation present?), but maybe there's something else either transient or unrelated.

Anyway, agreed that it sounds like it's not worth the time to track down. Will leave this sitting unless I get a very rainy day soon. 🙂

PR #39116, opened 2019-10-14
feat(node): v12.11

SimonSchick:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#12.11.1

Repeat of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38907 without stream changes for now.


SimonSchick:
@flarna for early eyes

types/node/net.d.ts
```diff
@@ -18,7 +18,26 @@ declare module "net" {
...
 
-    interface TcpSocketConnectOpts {
+    interface OnReadOpts {
+        buffer?: Uint8Array | (() => Uint8Array);
+        /**
+         * This function is called for every chunk of incoming data.
+         * Two arguments are passed to it: the number of bytes written to buffer and a reference to buffer.
+         * Return false from this function to implicitly pause() the socket.
+         */
+        callback?(bytesWritten: number, buf: Uint8Array): boolean;
```
Flarna: I think `callback` is mandatory here.

types/node/index.d.ts
```diff
@@ -87,6 +87,16 @@ interface SharedArrayBuffer {
...
 }
 
+interface PromiseSettlement<T> {
+    status: 'rejected' | 'fulfilled';
+    reason?: any;
+    value?: T;
+}
+
+interface PromiseConstructor {
+    allSettled<R>(iterable: Iterable<R | PromiseLike<R>>): Promise<Array<PromiseSettlement<R>>>;
```
ExE-Boss: I’d rather we do https://github.com/microsoft/TypeScript/pull/34065.

types/readable-stream/tsconfig.json
```diff
@@ -12,6 +12,11 @@
         "typeRoots": [
             "../"
         ],
+        "paths": {
```
Flarna: Does this work? How does this effect the final @types/readable-stream NPM package? Are there any issues to expect if someone uses readable-stream togehter with node 12 definitions?
By the way, maybe it's better to skip stream changes in this PR to get the other changes merged and then add one more PR for stream.

types/node/index.d.ts
SimonSchick: Agree, I actually posted a comment about this in another issue (which I can't find anymore), I will remove this for now and wait for the lib to avoid future conflicts.

types/readable-stream/tsconfig.json
SimonSchick: Forgot about this bit :)

types/node/net.d.ts
SimonSchick: I looked at the source and they don't seem mandatory.

SimonSchick: https://github.com/nodejs/node/blob/1447a79dc435578bffba32ead9b2ffa0425fc30a/lib/net.js#L295
Not 100% certain though, can you perhaps verify?

types/node/index.d.ts
```diff
@@ -87,6 +87,12 @@ interface SharedArrayBuffer {
     slice(begin?: number, end?: number): SharedArrayBuffer;
 }
 
+interface PromiseSettlement<T> {
```
ExE-Boss: This is also in https://github.com/microsoft/TypeScript/pull/34065.

And it doesn’t appear to be used in this PR anywhere else.

types/node/events.d.ts
```diff
@@ -1,8 +1,16 @@
...
+        once(event: string | symbol, listener: (...args: any[]) => void): this;
+    }
+
+    interface DomEventTarget {
+        addEventListener(event: string, listener: (...args: any[]) => void, opts: { once: boolean }): any;
+    }
+
     namespace internal {
-        function once(emitter: EventEmitter, event: string | symbol): Promise<any[]>;
+        function once(emitter: NodeEventTarget | DomEventTarget, event: string | symbol): Promise<any[]>;
```
galkin: Correct me if I'm wrong, but it seems to me that these changes do not meet the documentation.

Could you double check this file change with https://nodejs.org/dist/latest-v12.x/docs/api/events.html#events_events_once_emitter_name ?

galkin (changes requested):

types/node/events.d.ts
SimonSchick:
```
➜  node git:(feat/node-v12.11) node
Welcome to Node.js v12.11.1.
Type ".help" for more information.
> class MyEventTarget { addEventListener(name, fn, opts) { console.log('listening to', name, fn, opts); setTimeout(() =
> fn(123), 100); } } 
undefined
> events.once(new MyEventTarget(), 'test').then(data => console.log('received', data))
listening to test [Function] { once: true }
Promise { <pending> }
> received [ 123 ]
```

The documentation does point you to the DOM standard which I think I captured correctly here, you could argue that `opts: { once: boolean }` may not be required (although desired to prevent listener leak).

I intentionally did __not__ reference the DOM typings as this has caused name conflicts and issues in the past and would require DOM typings to be loaded for each node project.

SimonSchick: One could also argue that `symbol`s are technically not compatible with `addEventListener` so I might separate these into distinct functions

types/node/net.d.ts
Flarna: Well, node allows to omit `buffer` and/or `callback` and falls back to normal operation like if `onRead` has not been given at all. But I don't think that there is any real use case for passing `onRead` but skip `callback` or `buffer`.

SimonSchick: Gotta love node code.. I will make them required :)

SimonSchick:
Picking random DT person:
Hey @andrewbranch the current build failure appears to be unrelated to my changes, can you merge and/or re-run the CI please? :)

Flarna (approved):

PR #39128, opened 2019-10-14
Add types for camunda-external-task-client-js

MacRusher:
Typings for https://github.com/camunda/camunda-external-task-client-js/
Maintainers decided that they will not maintain type definitions themselves: https://github.com/camunda/camunda-external-task-client-js/issues/20#issuecomment-513653355

Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/camunda-external-task-client-js/index.d.ts
+++ b/types/camunda-external-task-client-js/index.d.ts
@@ -4,8 +4,16 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
-export class Client {
-    constructor(config: {
+declare namespace CamundaExternalTaskClient {
+    class Client {
+        constructor(config: ClientConfig);
+        start(): void;
+        stop(): void;
+        subscribe(topic: string, options: SubscribeOptions, handler: Handler): TopicSubscription;
+        subscribe(topic: string, handler: Handler): TopicSubscription;
+    }
+
+    interface ClientConfig {
         baseUrl: string;
         workerId?: string;
         maxTasks?: number;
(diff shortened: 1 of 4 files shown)


types/camunda-external-task-client-js/index.d.ts
```diff
@@ -0,0 +1,149 @@
...
+
+type Logger = Middleware & {
+    success(text: string): void;
+
+    error(text: string): void;
+};
+
+export const logger: Logger;
+
+export {};
```
andrewbranch: The module kind is slightly wrong here. You’re representing named ES2015 exports, but the target package is a CommonJS module exporting an object. What you want to do is declare a namespace, put your exports in there, and export the namespace as a CommonJS module:

```ts
declare namespace CamundaExternalTaskClient { // or whatever, name is flexible here
  export class Client {
    // ...
  }
  export {}; // Although really it’s nice to export your support types
}

export = CamundaExternalTaskClient;

andrewbranch (changes requested):

types/camunda-external-task-client-js/index.d.ts
MacRusher: I've applied your remarks, thanks for the review!
Unfortunately, I needed to silence the `export-just-namespace` rule, and I feel a little bad about this. Is there a better way how this could be handled?

andrewbranch: Oh no, I think I told you the wrong thing! I forgot TypeScript actually treats these equivalently as long as there’s no default export. Modules are hard 🙈. You had it right the first time, I’m sorry!

MacRusher: No problem :smile: I've moved it back to top-level, and also exported all support types as you suggested.
Let me know if there is something more I can do.

andrewbranch (approved):
Thanks!

PR #39130, opened 2019-10-15
Add typings for @onfleet/node-onfleet package

marcobeltempo:
Please fill in this template.

If adding a new definition:


types/onfleet__node-onfleet/index.d.ts
```diff
@@ -0,0 +1,10 @@
+// Type definitions for @onfleet/node-onfleet 1.0
+// Project: https://github.com/onfleet/node-onfleet
+// Definitions by: Marco Beltempo <https://github.com/marcobeltempo>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.5
+import { Onfleet } from './onfleet';
+export * from './error';
```
andrewbranch: I don’t see that these are actually exported from here, but I could be missing something? https://unpkg.com/browse/@onfleet/node-onfleet@1.0.4/lib/onfleet.js

andrewbranch (changes requested):
Wow, this looks like very thorough work! Thanks for the contribution!

The bad news is that the module kind for everything in here is wrong—it’s describing ES2015 modules when the actual package is using CommonJS modules. The good news is that the fix should be fairly straightforward.

Basically, anywhere the JavaScript says `module.exports = `, you have to have a `export = ` in your type definitions. CommonJS modules can only really export one thing (whatever `module.exports` points to), so to export multiple things (e.g., in most of these files, one class and a bunch of supporting types), you can take advantage of declaration merging—certain constructs sharing the same name which get merged together. It’s much easier to explain by example, so I’ve gone ahead and converted `Organization.d.ts` for you:

```ts
declare class Organization {
  get(): Promise<Organization.OnfleetOrganization[]>;
  get(id: string): Promise<Organization.OnfleetOrganization | Organization.Delegatee>;
  insertTask(id: string, obj: { tasks: string[] }): Promise<any>;
}

declare namespace Organization {
  interface Delegatee {
    country: string;
    email: string;
    id: string;
    name: string;
    timezone: string;
  }

  interface OnfleetOrganization {
    id: string;
    timeCreated: number;
    timeLastModified: number;
    name: string;
    email: string;
    image: string;
    timezone: string;
    country: string;
    delegatees: string[];
  }
}

export = Organization;
```

So the export refers to the class, but it also refers to a container for `OnfleetOrganization` and `Delegatee`. A consumer (e.g., your tests—they’ll need to be updated) can access them like this:

```ts
import Organization = require("./organization");

new Organization(); // `Organization` refers to the class
type Del = Organization.Delegatee; // `Organization` refers to the namespace
```

(However, if the consumer enables `esModuleInterop` in their tsconfig, then they can do simply `import Organization, { Delegatee } from "./organization";`.)

Make sense?

types/onfleet__node-onfleet/index.d.ts
marcobeltempo: Thanks @andrewbranch for the in depth review/guidance! I've gone ahead and updated the modules to CommonJS. 

In regards to the consumer example for the tests. Using relative paths throws the following lint error:
`Test file should not use a relative import. Use a global import as if this were a user of the package.`

types/onfleet__node-onfleet/index.d.ts
```diff
@@ -0,0 +1,9 @@
+// Type definitions for @onfleet/node-onfleet 1.0
+// Project: https://github.com/onfleet/node-onfleet
+// Definitions by: Marco Beltempo <https://github.com/marcobeltempo>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.5
+
+import Onfleet = require('./onfleet');
+
```
marcobeltempo: Is there a way I can easily import types/interfaces from the `Resources` directory using module interop?

For example:
`import Onfleet, { CreateTaskProps } from 'onfleet__node-onfleet';`

andrewbranch: You should be able to do:

without esModuleInterop:
```ts
import Onfleet = require('@onfleet/node-onfleet');
import Task = require('@onfleet/node-onfleet/Resources/Task');
// and access Task.CreateTaskProps
```

with esModuleInterop:
```ts
import Onfleet from '@onfleet/node-onfleet';
import { CreateTaskProps } from '@onfleet/node-onfleet/Resources/Task';
```

You could conceivably also nest namespaces together so that they’re all accessible from the index:

```ts
import Onfleet = require('@onfleet/node-onfleet');
type T = Onfleet.Task.CreateTaskProps;
```

but you’d have to be careful that those namespaces _only_ contain types, because the actual `Task` class isn’t going to exist on `Onfleet.Task` at runtime. Not sure those hoops are worth jumping through.

types/onfleet__node-onfleet/index.d.ts
andrewbranch: You should be able to import from `@onfleet/node-onfleet`.

marcobeltempo:
@andrewbranch I'm not sure what's causing the pipeline to fail.
[Azure Build 18864](https://dev.azure.com/definitelytyped/DefinitelyTyped/_build/results?buildId=18864&view=logs&jobId=275f1d19-1bd8-5591-b06b-07d489ea915a&taskId=40b1ee41-44d6-5bba-aa04-4b76a5c732e5&lineStart=15&lineEnd=20&colStart=1&colEnd=91)

```
{ [Error: ENOENT: no such file or directory, open '/home/vsts/work/1/s/types/onfleet__node-onfleet/resources/Administrators/index.d.ts']
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path:
   '/home/vsts/work/1/s/types/onfleet__node-onfleet/resources/Administrators/index.d.ts' }
```

marcobeltempo:
> @marcobeltempo Linux file system is case-sensitive :) [DefinitelyTyped/DefinitelyTyped/pull/39130/files#diff-9c9fc4a4ae11871ac9d757ba7387b3b3R1](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39130/files#diff-9c9fc4a4ae11871ac9d757ba7387b3b3R1)

The paths seem to match. Unless I'm missing something

andrewbranch:
Those import paths have a lowercase `r` in `resources`, whereas the folder is actually named with an uppercase `R`.

marcobeltempo:
Ahhh I see the issue. I had the directory renamed locally but git wasn't picking up change.

Solution was to set `git config core.ignorecase false`

andrewbranch:
Looks like you need to `git rm --cached` the unused ones. Also, I’d keep the uppercase `R` since that’s what the JavaScript uses—that could be important for consumers who are on case-sensitive file systems (or have `forceConsistentCasing` turned on).

andrewbranch (approved):
Nice work! 🙌

marcobeltempo:
@andrewbranch Thanks again for your help 👏

PR #39133, opened 2019-10-15
[knockout] knockout no longer needed

### LINKED ISSUE #37490: ../node_modules/@types/knockout/index.d.ts:1060:13 - error TS2403
../node_modules/@types/knockout/index.d.ts:1060:13 - error TS2403: Subsequent variable declarations must have the same type.  Variable 'ko' must be of type 'typeof import(".../node_modules/knockout/build/types/knockout")', but
here has type 'KnockoutStatic'.

1060 declare var ko: KnockoutStatic;

knockout.d.ts(5, 1): 'ko' was also declared here.

**The above fails to compile the index.d.ts type definition file.** 
As read in a stackoverflow comment, this is a **Typescript augmentation issue.** Is there a fix or patch available for this? 

I'm new to Typescript and this is a frustrating compile headache. 

My work around was to rename var ko to something else e.g. kot and the corresponding export value in the index.d.ts file @1060. Every fiber in my body knows this cant be right but compilation works for my js application.

kindly advice/ fix

viceice:
Please fill in this template.

If adding a new definition:

If removing a declaration:

depends on: https://github.com/microsoft/types-publisher/pull/665
fix https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37490


conficient (commented):
Can I ask why you have removed Knockout? The PR has no details on what you are doing and why, and for such a large change some justification is required

edit: OK I've seen that Knockout now has its own types, but I am not sure removing it from DT is necessary - a lot of sites will depend on the defs here. Are the @types conflicting?

ViceIce:
@conficient Yes, the types conflicting. The new types are renamed and namespaced.

ltlombardi:
Correct if I'm wrong:
What is the normal behavior when something like this happens? 
If this were a "normal" software repository, a new version would be released with all the breaking changes and new features explained, and consumer of the library could just keep using the previous released.

Here, things are a little different. We don't have releases...
The releases are provided by NPM and are automatic. Can be seen [here](https://www.npmjs.com/package/@types/knockout).

What will happen to all those previous versions in NPM compatible with Knockout version 3.4 ? 
If they will be removed, than I don't thing it's a good idea to remove these typings.

jessetrinity:
@ViceIce I'm thinking it is unlikely that you will get sign off from all of the owners of the dependent packages on this one PR. I would suggest opening PRs separately for the dependent packages to use the built in types for knockout before trying to remove them from DT.

@sandersn FYI

ViceIce:
@jessetrinity How get I a successful ci build, if I only change depending package? It'll find duplicate knockout package.

jessetrinity:
@ltlombardi the previous versions of the @types package will still be available up to the `asOfVersion` entry in `notNeededPackages.json`.

sandersn:
From questions raised by @ltlombardi and @Retsam, it sounds like the community may want to continue updating the 3.4 typings even though 3.5 is out. This makes sense for a stable, widely used library.

If that's the case, we should keep these typings so the maintainers can continue to improve/fix the types for KO 3.4. @ltlombardi @Retsam is it the case that you (or others) want to keep updating the 3.4 types?

sandersn (changes requested):
Requesting changes so typescript-bot knows the current state of the PR.

Retsam:
@sandersn Yeah, I'd like to be able to keep updating the community types.

ViceIce:
Sorry for late answer, I'm on holiday. I'll split this or the next weeks

PR #39134, opened 2019-10-15
[@types/react] Change .relatedTarget type to include null

Andarist:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://developer.mozilla.org/en-US/docs/Web/API/FocusEvent/relatedTarget ("relatedTarget may also be set to null for security reasons, like when tabbing in or out of a page.")

There were no tests relating `relatedTarget` and I'm not sure if I should one (or how it would look like).

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -1085,7 +1085,7 @@ declare namespace React {
     }
 
     interface FocusEvent<T = Element> extends SyntheticEvent<T, NativeFocusEvent> {
-        relatedTarget: EventTarget;
+        relatedTarget: EventTarget | null;
         target: EventTarget & T;
     }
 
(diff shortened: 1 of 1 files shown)


eps1lon:
I remember that there was some discussion about how react polyfills `relatedTarget`. So while (not verified) it is correct that this can be `null` the official spec for FocusEvent is not applicable. I'll see if we can add a test that also adds some documentation that `relatedTarget` behaves slightly different in some browsers.

Andarist:
@eps1lon I've stumbled upon this while working with React's synthetic events - they might maybe try to polyfill this (or maybe they might try to do that in their react-events experiments), but it certainly isn't handled in all cases and quite frankly it can't be. The most notable example would be blurring a field by focusing dev tools (or something else outside of a browser tab) - `relatedTarget` just can't be set to anything valuable in such a case.

eps1lon (approved):
I guess this is fine as is. Documenting react issues in the types is probably not a good idea.

Just want to emphasize that reacts relatedTarget does not have the same browser compatibility as native relatedTarget: https://github.com/facebook/react/issues/3751

PR #31610, opened 2018-12-22
Add ioredis client support for connect-redis

renarsvilnis:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/tj/connect-redis#custom-redis-clients
- [ ] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

-----

Not sure how DefinitelyTyped handles versions, but the existing type definitions didn't have any version as this is a patch I didn't increment anything

PR #39137, opened 2019-10-15
[connect-redis] add `ioredis` as client type

iamolegga:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/tj/connect-redis#client

### DIFF
--- a/types/connect-redis/index.d.ts
+++ b/types/connect-redis/index.d.ts
@@ -2,18 +2,20 @@
 // Project: https://npmjs.com/package/connect-redis
 // Definitions by: Xavier Stouder <https://github.com/xstoudi>
 //				   Seth Butler <https://github.com/sbutler2901>
+//				   Oleg Repin <https://github.com/iamolegga>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
 /// <reference types="express" />
 /// <reference types="express-session" />
 /// <reference types="redis" />
+/// <reference types="ioredis" />
 
 declare module "connect-redis" {
     import * as express from "express";
     import * as session from "express-session";
     import * as redis from "redis";
-
+    import * as ioredis from "ioredis";
 
     function s(options: (options?: session.SessionOptions) => express.RequestHandler): s.RedisStore;
 
@@ -23,7 +25,7 @@ declare module "connect-redis" {
             client: redis.RedisClient;
         }
         interface RedisStoreOptions {
-            client?: redis.RedisClient;
+            client?: redis.RedisClient | ioredis.Redis;
             host?: string;
             port?: number;
             socket?: string;
--- a/types/connect-redis/index.d.ts
+++ b/types/connect-redis/index.d.ts
@@ -4,7 +4,7 @@
 //				   Seth Butler <https://github.com/sbutler2901>
 //				   Oleg Repin <https://github.com/iamolegga>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+// TypeScript Version: 2.8
 
 /// <reference types="express" />
 /// <reference types="express-session" />
@@ -21,7 +21,7 @@ declare module "connect-redis" {
 
     namespace s {
         interface RedisStore extends session.Store {
-            new (options: RedisStoreOptions): RedisStore;
+            new(options: RedisStoreOptions): RedisStore;
             client: redis.RedisClient;
         }
         interface RedisStoreOptions {
--- a/types/connect-redis/connect-redis-tests.ts
+++ b/types/connect-redis/connect-redis-tests.ts
@@ -1,5 +1,7 @@
 import connectRedis = require("connect-redis");
 import session = require("express-session");
+import redis = require("redis");
+import IORedis = require("ioredis");
 
 let RedisStore = connectRedis(session);
 const store = new RedisStore({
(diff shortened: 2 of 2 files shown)


types/connect-redis/index.d.ts
```diff
@@ -2,28 +2,30 @@
...
 
     namespace s {
         interface RedisStore extends session.Store {
-            new (options: RedisStoreOptions): RedisStore;
+            new(options: RedisStoreOptions): RedisStore;
             client: redis.RedisClient;
         }
         interface RedisStoreOptions {
-            client?: redis.RedisClient;
+            client?: redis.RedisClient | ioredis.Redis;
```
weswigham: To quote [the me of a year ago](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/31610#discussion_r244602290):
> I would say that the correct thing to do here is create an interface for the common parts of `redis.RedisClient` and `ioredis.Redis` within the `redis` library, us that interface here (rather than the exact class type) and ensure that `ioredis` conforms to it, rather than using a union. It doesn't make sense for all `connect-redis` types to depend on `ioredis` when it's likely not in use.

weswigham (changes requested):

types/connect-redis/index.d.ts
iamolegga: I beg to differ. For now it's easier to keep it like that, then look at internal implementation of library and look what exactly is required, especially when it says in docs that it support only [three libraries](https://github.com/tj/connect-redis#client) (and [one of each has types alias to another](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/redis-mock/index.d.ts)). So, I feel like you are requiring something that no one need, I think this can be shipped now and let end-users get rid of `client: new IORedis() as any`. I would agree with you if there will be more clients, but I'm not sure for now

trymoto: I would agree with @iamolegga because to me, maintaining common interface within this def seems impractical.

weswigham: The docs specifically say "Any clients compatible with the redis API will work". If the `ioredis` client isn't compatible with the types for the redis client, then it's referring to just a subset of that client API. Enumerating the important part of that subset here makes sense. I'm also going to stress that the connect-redis readme lists three _known_ client APIs - there very well could be more which would also benefit from that change. I really dislike this change because it's super one-off and introduces a dependency in the types which doesn't exist in the JS. If all these packages shipped their own types, we'd be introducing a phantom, unused dependency with this.

iamolegga: > The docs specifically say "Any clients compatible with the redis API will work". If the ioredis client isn't compatible with the types for the redis client, then it's referring to just a subset of that client API. Enumerating the important part of that subset here makes sense.

Ok, just look at the usage example https://github.com/tj/connect-redis/blob/master/lib/connect-redis.js#L46-L58
it uses client's method `set` as `(args: [type1, type2, type3], cb: Function) => any`, for now it's neither  supported by typings of even _original_ client https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/redis/index.d.ts#L846-L849
nor by typings of _alternative_ client https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ioredis/index.d.ts#L79-L84
and the same with all the other methods, that are used there.

so, you suggest me to add more overloads in two different client libs in the very first, and then come back again and create common interface, right? Isn't it too much for such a little improvement?

> I'm also going to stress that the connect-redis readme lists three known client APIs - there very well could be more which would also benefit from that change.

sure,  there are a lot of redis clients, every day someone creates one more, and adds it to list of supported clients.

> I really dislike this change because it's super one-off and introduces a dependency in the types which doesn't exist in the JS. If all these packages shipped their own types, we'd be introducing a phantom, unused dependency with this.

but somehow there is `redis` dependency already. How come? Why there was no such rule firstly? Is this rule works for all the other libs, or only for this one?

weswigham: No, I'm suggesting you define an
```ts
interface AbstractRedisClient {
  // ...
}
```
here in `connect-redis` which defines the common subset of the type that `connect-redis` actually needs, then using that here, thereby removing the dependency on `redis.RedisClient` as well.

iamolegga: But I'm saying, they are not compatible now. If you want to create an abstract interface as it is used by library, then both clients need changes. Does it worth it? For me it's easier to use `as any`. And you will get another one same PR, where you will quote yourself again . It's up to you

weswigham: I mean, as a bare minimum you can define something like
```ts
interface AbstractRedisClient {
  set: (...args: any) => any;
}
```
and refine it later. Simply requiring properties of the right name of the right domain is pretty good when they client types themselves are already incomplete.

weswigham: It looks like you only need to refer to `get`, `set`, `mget`, `scan`, `expire`, and `del`, as that's all `connect-redis` relies on.

iamolegga: I'm sorry, but for me your suggestion doesn't seem like really valuable for end user, and any way better than ([not](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36247) [only](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/31610)) my proposal. Looks like other opinions (which belong to different people) are not accepted here 👎

PR #39143, opened 2019-10-15
Lowercase js-cookie same site cookie values

theodorejb:
This matches the documented usage at https://github.com/js-cookie/js-cookie/#samesite, and also is more consistent with other attribute values typically entered in lowercase (e.g. domain names and boolean values).


theodorejb:
This shouldn't be a big backwards compatibility break since the `sameSite` type definition was just added last week in PR #38855.

nreynis (approved):

PR #39144, opened 2019-10-15
[dom-mediacapture-record] Nullable event listeners for MediaRecorder

laszlopandy:
Allow setting event listeners to null when using strictNullChecks.
In lib.dom.d.ts (provided by TypeScript), all event handlers are marked as nullable so you can unset your event handler.

Please fill in this template.

### DIFF
--- a/types/dom-mediacapture-record/index.d.ts
+++ b/types/dom-mediacapture-record/index.d.ts
@@ -39,12 +39,12 @@ declare class MediaRecorder extends EventTarget {
     readonly videoBitsPerSecond: number;
     readonly audioBitsPerSecond: number;
 
-    ondataavailable: (event: BlobEvent) => void;
-    onerror: (event: MediaRecorderErrorEvent) => void;
-    onpause: EventListener;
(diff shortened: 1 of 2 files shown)


types/dom-mediacapture-record/index.d.ts
```diff
@@ -39,12 +39,12 @@ declare class MediaRecorder extends EventTarget {
...
-    onpause: EventListener;
-    onresume: EventListener;
-    onstart: EventListener;
-    onstop: EventListener;
+    ondataavailable: ((event: BlobEvent) => void) | null;
+    onerror: ((event: MediaRecorderErrorEvent) => void) | null;
+    onpause: EventListener | null;
+    onresume: EventListener | null;
+    onstart: EventListener | null;
+    onstop: EventListener | null;
```
eliasmeire: I would make these optional (like `onstop?: EventListener`) instead of `| null`. That way you don't need to assign something and `recorder.onstop = null;` will still pass.

elsmr (approved):

PR #39151, opened 2019-10-15
add types for babel-plugin-macros

billykwok:
## Overview
Add types for babel-plugin-macros 2.6+
[GitHub Repo (https://github.com/kentcdodds/babel-plugin-macros)](https://github.com/kentcdodds/babel-plugin-macros)

## Template

If adding a new definition:


types/babel-plugin-macros/index.d.ts
```diff
@@ -0,0 +1,36 @@
+// Type definitions for babel-plugin-macros 2.6
+// Project: https://github.com/kentcdodds/babel-plugin-macros
+// Definitions by: Billy Kwok <https://github.com/billykwok>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.6
+import * as Babel from 'babel__core';
```
andrewbranch: I think this needs to be `@babel/core` (you already have it setup in your path mapping 👍)

types/babel-plugin-macros/index.d.ts
```diff
@@ -0,0 +1,36 @@
+// Type definitions for babel-plugin-macros 2.6
+// Project: https://github.com/kentcdodds/babel-plugin-macros
+// Definitions by: Billy Kwok <https://github.com/billykwok>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.6
```
andrewbranch: This is the _minimum_ TypeScript version required. You should set it to the lowest value possible.

andrewbranch (changes requested):

types/babel-plugin-macros/index.d.ts
billykwok: Thanks for your comment. Fixed.

types/babel-plugin-macros/index.d.ts
billykwok: Thanks for your comment. Fixed.

andrewbranch (approved):

PR #39153, opened 2019-10-15
feat(react): make existing inputMode type narrower

zieka:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://html.spec.whatwg.org/multipage/interaction.html#input-modalities:-the-inputmode-attribute


eps1lon (approved):
Thanks. Could you change the PR title so that it is obvious that we're narrowing an existing type? 

Since typescript widens return types by default this can cause issues and I would prefer it if this is obvious from the title. The type was always there. Just too wide.

PR #39160, opened 2019-10-16
Fix type definitions for forge-apis

bcmarinacci:
All of the function signatures for `@types/forge-apis` were incorrect, which resulted in TypeScript not being able to compile with correct function calls. I updated the existing types and added the correct types for the rest of the package, all of which now match the SDK docs: https://github.com/Autodesk-Forge/forge-api-nodejs-client#api-documentation. For the couple of cases where the SDK docs weren't complete, I referenced the source API docs: https://forge.autodesk.com/en/docs/data/v2/reference/http to fill in the missing pieces. I also updated existing tests and added new ones for every function.

I am currently only using a subset of the SDK (mostly the Buckets and Derivatives APIs) but everything I have tested so far works.

There is definitely more work that can be done but this pull request covers nearly everything other than things like explicit response types, which would require a significantly more time. Overall, this seemed like a good starting point to incrementally build upon as time goes on.

For the `If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.` item, I wasn't sure if I should bump the type version from `0.4` as the existing types aren't actually valid and the type definitions from this pull request are valid for version `0.4.5` (the most recent version as of Tue, 15 Oct 2019) of `forge-apis`.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/Autodesk-Forge/forge-api-nodejs-client

### DIFF
--- a/types/forge-apis/README.md
+++ b/types/forge-apis/README.md
@@ -1,6 +1,10 @@
 # Installation
 
-> `npm install --save @types/forge-apis`
+```bash
+yarn add --dev @types/forge-apis
+# or
+npm install --save-dev @types/forge-apis
+```
 
 # Summary
 
--- a/types/forge-apis/README.md
+++ b/types/forge-apis/README.md
@@ -1,16 +1,22 @@
 # Installation
+
 > `npm install --save @types/forge-apis`
 
 # Summary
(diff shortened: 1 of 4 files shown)


weswigham:
@autodesk-forge any input? You've got maybe a few more days before I just merge this, since these changes look to be well thought out.

PR #39173, opened 2019-10-16
Node: `util.promisify`: widen callback error parameter to `unknown`

OliverJAsh:
An error can be of any type, not just `Error | null`—similar to how you can `throw` any type. For example:

```ts
const fetch = (callback: (error: string | null, value: string) => void) => callback(null, 'yay');
```

Under the current types, when this function (or any function that has callback error parameter whose type is wider than `Error | null`) is ran through `promisify`, it does not return the correct type:

```ts
// Type '(arg1: (error: string | null, value: string) => void) => Promise<string>' is not assignable to type '() => Promise<string>'.ts(2322)
const result: () => Promise<string> = util.promisify(fetch);
```

This PR fixes that that.

---

Please fill in this template.


types/node/util.d.ts
```diff
@@ -83,24 +83,27 @@ declare module "util" {
         fn: (arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, arg6: T6) => Promise<TResult>
     ): (arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, arg6: T6, callback: (err: NodeJS.ErrnoException | null, result: TResult) => void) => void;
 
+    // The value type here is a "poor man's `unknown`". When these types support TypeScript
+    // 3.0+, we can replace this with `unknown`.
+    type PoorMansUnknown = {} | null | undefined;
```
OliverJAsh: I thought about avoiding the need for this by making this change to the 3.1+ types, but I think we ought to provide this bug fix to all users.

OliverJAsh:
I don't understand why the build is failing:

```
Error: Error: Command failed: npm install --ignore-scripts --no-shrinkwrap --no-package-lock --no-bin-links --no-save
npm ERR! code ENOENT
npm ERR! syscall lchown
npm ERR! path /home/vsts/.npm/_locks/staging-737dbd926de1c3da.lock
npm ERR! errno -2
npm ERR! enoent ENOENT: no such file or directory, lchown '/home/vsts/.npm/_locks/staging-737dbd926de1c3da.lock'
npm ERR! enoent This is related to npm not being able to find a file.
npm ERR! enoent 
```

https://dev.azure.com/definitelytyped/DefinitelyTyped/_build/results?buildId=18816

SimonSchick:
Didn't a similar change cause issues before? cc @Flarna
I recall some issues raised after the last introduction of `PoorMansUnknown` in another place.
Also going to lean out of the window here and ask what library it is that is calling back with non-errors.

OliverJAsh:
> ask what library it is that is calling back with non-errors

I don't have one. I was writing types for a module, and when it came to defining the error type, I decided to use `unknown` because it's not safe to make assumptions. Then, when I came to promisify this function, I ran into this.

Flarna:
The `PoorMansUnknown` was added in querystring via #33997 - but not as named type.
It's also used in `UnhandledRejectionListener` (again not as named type).
I think there were complains after changing `UnhandledRejectionListener` from using `any` to `PoorMansUnknown` as it was breaking code from some people - `any` fits all wheras `PoorMansUnknown` fits nothing.

`Promise.catch` accepts a `(reason: any) => TResult` according to lib.es5.d.ts so somehow `any` would match better than `PoorMansUnknown`.
I think for the proposed changes in this PR it's actually don't care what's used as the error type in the callbacks is not used anywhere therefore a type allowing anything should be fine - why not `any`?

I think this PR has also no potential to break anything. Or do I overlook something?

OliverJAsh:
I tend to prefer `unknown` over `any` because the latter disables type checking and in some situations it can leak into your code, e.g. via inference:

```ts
type Callback = (error: any, result: string) => void;
const myCallback: Callback = (error, result) => {
  error.property.that.does.not.exist // runtime error!
}
```

Flarna:
I agree that `unknown` is in general a better choice than `any` but I don't see the scenario here where an `any` could leak.

OliverJAsh:
Maybe not in this case, as the callback will never be inlined as in my example. 🤔

If we were to use `unknown`, how might it break people's code in an undesirable way?

SimonSchick:
Also want to point out there are linting rules for `no-any` which are very good at catching these things, it's not _perfect_ but I'm willing to say it catches 95% of all unwanted `any`s.

Flarna:
If we would have the real `unknown` here than I would clearly favor it.
We have seen confusions regarding `PoorMansUnknown` in the past therefore I slightly tend towards `any` - but I will for sure not block this.

types/node/util.d.ts
Flarna: Maybe move the `PoorMansUnknown` to NodeJS namespace in global.d.ts and use it also at other locations.

Flarna (approved):

PR #39186, opened 2019-10-16
react-autosuggest: restrict input props

OliverJAsh:
This brings `InputProps` inline with React's types for `input`. That is, the following errors in React:

```tsx
<input anything={false} />
```

This also means that `keyof InputProps<T>` returns a union of literals instead of `string`. This is helpful because it means that keys will be validated in operations such as `Pick<InputProps<T>, 'autoFocus'>`.

---

Please fill in this template.

### DIFF
--- a/types/react-autosuggest/index.d.ts
+++ b/types/react-autosuggest/index.d.ts
@@ -68,7 +68,6 @@ declare namespace Autosuggest {
         onChange(event: React.FormEvent<any>, params: ChangeEvent): void;
         onBlur?(event: React.FocusEvent<any>, params?: BlurEvent<TSuggestion>): void;
         value: string;
-        [key: string]: any;
     }
 
     interface SuggestionSelectedEventData<TSuggestion> {
--- a/types/react-autosuggest/react-autosuggest-tests.tsx
+++ b/types/react-autosuggest/react-autosuggest-tests.tsx
@@ -558,3 +558,9 @@ export class ReactAutosuggestCustomTest extends React.Component<any, any> {
 
 ReactDOM.render(
    <ReactAutosuggestCustomTest/>, document.getElementById('app'));
+
+const test: Autosuggest.InputProps<{ foo: string }> = {
+    onChange: () => {},
+    value: 'foo',
+    anything: false // $ExpectError
+};


ulrichb (approved):

andrewbranch (approved):

thenovelnomad:
This may be a case of your bug is my feature, but this broke our build process, and while I won't argue that it should not have been changed, it maybe should not have just been a patch version change.

thenovelnomad:
Ahh, thanks for the details, @andrewbranch. That makes sense. I guess we should lock down our types packages to the patch version.

types/react-autosuggest/index.d.ts
```diff
@@ -68,7 +68,6 @@ declare namespace Autosuggest {
         onChange(event: React.FormEvent<any>, params: ChangeEvent): void;
         onBlur?(event: React.FocusEvent<any>, params?: BlurEvent<TSuggestion>): void;
         value: string;
-        [key: string]: any;
```
dolezel: This is definitely not correct as everything you pass as `inputProps` into the component you get as param of `renderInputComponent`

dolezel: https://github.com/mui-org/material-ui/blob/master/docs/src/pages/components/autocomplete/IntegrationAutosuggest.js
this stopped working...

OliverJAsh: @dolezel Please can you provide a reduced test case demonstrating the error you're seeing?

dolezel: I've posted link above. E.g. here https://github.com/mui-org/material-ui/blob/master/docs/src/pages/components/autocomplete/IntegrationAutosuggest.js#L199 it complains that `'classes' does not exist in type 'InputProps'`, but it should be possible to pass anything through and access it (e.g. https://github.com/mui-org/material-ui/blob/master/docs/src/pages/components/autocomplete/IntegrationAutosuggest.js#L50)

OliverJAsh: I see. In that case, `InputProps` should probably be a generic that is shared with `renderInputComponent`. Until someone is able to make that change, I would suggest using a cast on the properties which are causing this problem (e.g. `classes`):

```tsx
inputProps={{
  // We want to pass these extra props through to `renderInputComponent`. A cast is needed because they are not included in the `InputProps` type.
  ...{ classes } as InputProps<T>,
  id: 'whatever',
  // other properties
}}
```

cheeZery: react-autosuggest itself, or react-autowhatever by the same creator, which is heavily used by react-autosuggest, sets `ref` as an attribute on every input. After your change, this can't be accessed anymore (without workarounds), since it isn't a default HTML/DOM attribute. 

https://github.com/moroshko/react-autowhatever/blob/a74c882d4f63fabe7856e5d90b93c2af81c89a5b/src/Autowhatever.js#L339

OliverJAsh: Good catch, want to raise a PR to whitelist that prop?

cheeZery: Gladly, which type do you suggest? `React.RefAttributes<HTMLInputElement>`?

```typescript
    interface InputProps<TSuggestion>
        extends Omit<React.InputHTMLAttributes<any>, 'onChange' | 'onBlur'> {
        onChange(event: React.FormEvent<any>, params: ChangeEvent): void;
        onBlur?(event: React.FocusEvent<any>, params?: BlurEvent<TSuggestion>): void;
        value: string;
        ref: React.RefAttributes<HTMLInputElement>;
    }
```

`react-autowhatever` always seems to add a ref method, never an `RefObject` (nor `null`) 🤔

ulrichb: @cheeZery Could you explain you use case?

I don't really get why to expose the `ref` prop because AutoSuggest overrides it anyways in the code line you posted above (and _doesn't_ "re-wire" the input ref as it happens for `onChange` in line 239 or `onBlur` in line 249).

cheeZery: @ulrichb sure, we use `renderInputComponent` to create our own input component. In that component we need a reference to the input element. And since moroshko/react-autosuggest also needs the reference, we have to manually call the `ref` method.

Here's a stub which tries its best to showcase our use case:

```typescript
import { InputProps } from "react-autosuggest";

class MyInputComponent extends React.Component<{ inputProps: InputProps<object>}> {
    private inputRef: HTMLInputElement;

    public render() {
        return (
            <input
                {...this.props.inputProps}
                ref={this.setRef.bind(this)}
            />
        );
    }

    private setRef(inputRef: HTMLInputElement) {
        this.ref = inputRef;
        this.inputProps.ref(inputRef); // << Property 'ref' does not exist on type 'InputProps<object>'.
    }
}

```

ulrichb: Ahhh `renderInputComponent`,  I see.

In this case I would recommend to add the `ref` member it to the `RenderInputComponent` callback type ([line 123](https://github.com/OliverJAsh/DefinitelyTyped/blob/d25d8cce76e3522b99476423171d25c64ae11a86/types/react-autosuggest/index.d.ts#L123)).
Something like `(inputProps: InputProps<TSuggestion> & { readonly ref: (instance: HTMLInputElement) => void }) => ...`.

This would avoid adding it to `InputProps` thus as _input value for the `AutoSuggest` component_ (which would be misleading/wrong because it will get overwritten by AutoSuggest).

cdeutsch: Since `renderInputComponent` is a thing and you can render components from other libraries in it,  I don't think `[key: string]: any;` should have been removed.

This broke the code I was using with Ant Design.

d-szymanski: It also broke our implementation with `data-` prefixed properties.
It's valid in React so why should it be removed?
Maybe we should extend from a different interface to maintain backward compatibility as this change was "patch" but was breaking change for us.

OliverJAsh: > `InputProps` should probably be a generic that is shared with `renderInputComponent`

This is the correct fix that doesn't remove type safety. Would someone like to send a PR? I'm happy to review it.

d-szymanski: [dataset](https://github.com/Microsoft/TypeScript/blob/master/src/lib/dom.generated.d.ts#L7970) is valid thing in HTML. Yet, we still can't define regex on properties names so we can't type safety `date-name` properties.
If we could do that (there is a proposal: https://github.com/Microsoft/TypeScript/issues/6579), it would be so easy to implement:
```typescript
interface DataAttributes {
    [key: /data-\w+/]: number | string | boolean;
}
```

yet as for now, I am proposing: 
```typescript
// DOM data attributes later used 
interface DataAttributes {
    [key: string]: string;
}
```

PR #39188, opened 2019-10-16
stampit 4.3 - better typing through generics

PopGoesTheWza:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/stampit-org/stampit/pull/348>


andrewbranch:
I’m not sure why typescript-bot didn’t ping the existing owners: @koresar @lummish, can you please review? Thanks!

koresar:
Hi all. I've just got my first baby. Going back to development next week. Sorry for the delay.

PopGoesTheWza:
@koresar no need to rush.
Before approving for publish, you should also look at the [stampit PR](https://github.com/stampit-org/stampit/pull/348) as an alternative for we have two options:

1. update `@types/stampit`
2. include types into the `stampit` package and deprecate/remove `@types/stampit`

Also note that I only did only minimal testing since my knowledge of `stampit` is rather limited.

types/stampit/index.d.ts
```diff
@@ -1,518 +1,560 @@
...
+interface StampSignature {
+    (options?: PropertyMap, ...args: unknown[]): any;
+    compose: ComposeMethod & stampit.Descriptor<any, any>;
+}
+
+/**
+ * @internal Extracts the `Stamp` type.
+ * @template Original The type to extract the `Stamp` type from.
+ */
+type StampType<Original> = Original extends /* disallowed types */ [] | bigint
```
koresar: This whole `type` is a black magic to me. :) I'll pretend it's all correct. Although, the `[]`, `bigint`, `true` look weird.

types/stampit/index.d.ts
```diff
@@ -1,518 +1,560 @@
...
 declare namespace stampit {
+    /** A composable object (either a `Stamp` or a `Descriptor`.) */
+    type Composable = StampSignature | Descriptor<any, StampSignature>;
+
+    /**
+     * A `Stamp`'s metadata.
+     * @template Obj The type of the object instance being produced by the `Stamp`.
+     * @template S̤t̤a̤m̤p̤ The type of the `Stamp` (when extending a `Stamp`.)
+     */
+    interface Descriptor<Obj, S̤t̤a̤m̤p̤ extends StampSignature = Stamp<Obj>> {
```
koresar: This should be renamed to "ExtendedDescriptor". The actual Descriptor do not have shortcuts like "props", "init".

types/stampit/index.d.ts
```diff
@@ -1,518 +1,560 @@
...
+type Pojo = object; // { [s: string]: any; }
+
+/** @internal Base type for all `properties`-like metadata. */
+// TODO: discriminate Array
+type PropertyMap = object; // { [s: string]: any; }
+
+/** @internal Signature common to every `Stamp`s. */
+interface StampSignature {
+    (options?: PropertyMap, ...args: unknown[]): any;
+    compose: ComposeMethod & stampit.Descriptor<any, any>;
```
koresar: After renaming `stampit.Descriptor` to `stampit.ExtendedDescriptor` we should add (somewhere above) the `Descriptor`. It should have only these exactly 11 properties: https://github.com/stampit-org/stamp-specification#the-stamp-descriptor

koresar: In other words, at runtime the `MyStamp.compose.*` properties are limited to those 11. There will never be `props`, or `init`, or other shortened properties.

types/stampit/index.d.ts
```diff
@@ -1,518 +1,560 @@
...
-         * @param descriptors
-         * @returns A new Stamp
-         */
-        staticPropertyDescriptors(...descriptors: Array<{}>): Stamp;
+    /**
+     * A function used as `.composers` argument.
+     * @template S̤t̤a̤m̤p̤ The type of the `Stamp` produced by the `.compose()` method.
+     */
+    interface Composer<S̤t̤a̤m̤p̤ extends StampSignature> {
+        // ??? should the type of `this` be the object or the stamp
```
koresar: Answering the question. The `compoers` functions do not have any `this`.  Although, the specification does not mention it. https://github.com/stampit-org/stamp-specification/blob/master/README.md#composer-parameters

koresar: Also, the return type of a composer should be `Stamp | undefined`

types/stampit/index.d.ts
PopGoesTheWza: It is indeed rather black magic for it will test all cases I could think of and try to detect if and how a stamp type can be inferred: it works out of Descriptor and regular Stamp types. It worked with “extended” Stamp types as far as ì could test to (cf. the test .ts file)
The initial type test is to properly return the never type for types which would otherwise produce (wrongfully) a valid return types.
This use of TypeScript types is not typical. It attempts to cover as many uses of stamps as possible without requiring too formal type definition upfront.

types/stampit/index.d.ts
```diff
@@ -1,518 +1,560 @@
...
-         * @returns A new Stamp
-         */
-        deepConfiguration(...deepConfs: Array<{}>): Stamp;
+    /**
+     * A function used as `.initializers` argument.
+     * @template Obj The type of the object instance being produced by the `Stamp`.
+     * @template S̤t̤a̤m̤p̤ The type of the `Stamp` producing the instance.
+     */
+    interface Initializer<Obj, S̤t̤a̤m̤p̤ extends StampSignature> {
+        // ??? should the type of `this` be the object or the stamp
```
koresar: Answering the question. The `this` context of initializers is the object instance created from the stamp.

types/stampit/index.d.ts
```diff
@@ -1,518 +1,560 @@
...
-         * @returns A new Stamp
-         */
-        staticPropertyDescriptors(...descriptors: Array<{}>): Stamp;
+    /**
+     * A function used as `.composers` argument.
+     * @template S̤t̤a̤m̤p̤ The type of the `Stamp` produced by the `.compose()` method.
+     */
+    interface Composer<S̤t̤a̤m̤p̤ extends StampSignature> {
+        // ??? should the type of `this` be the object or the stamp
+        (/* this: S, */ parameters: ComposerParameters<S̤t̤a̤m̤p̤>): S̤t̤a̤m̤p̤;
```
koresar: I believe this line should be something like
```
(stamp: S̤t̤a̤m̤p̤, composables: Composable[]): void | S̤t̤a̤m̤p̤;
```

koresar: Wrong! :)
This is better:
```
(parameters: ComposerParameters<S̤t̤a̤m̤p̤>): void | S̤t̤a̤m̤p̤;
```

types/stampit/stampit-tests.ts
```diff
@@ -24,22 +51,19 @@ foo.getB(); // "b"
 // Here's a mixin with public methods, and some refs:
 const membership = stampit({
     methods: {
-        members: {},
+        // members: {},
```
koresar: Right! I remember now why this test exists. :)
Even though it's called `methods` you can put anything you want here. Including plain objects like the above `members`. So I believe the MethodsMap should be extended to allow any types:

```ts
interface MethodMap<This> {
    [s: string|Symbol]: (this: This, ...args: any[]) => any;
    [s: string|Symbol]: any;
}
```

types/stampit/stampit-tests.ts
```diff
@@ -24,22 +51,19 @@ foo.getB(); // "b"
...
         }
-    },
-    refs: {
-        members: {}
     }
 });
 
 // Let's set some defaults:
-const defaults = stampit().refs({
+const defaults = stampit().conf({
```
koresar: The v3.0 `refs` was renamed to `props` in v4.0. Sorry for the confusion. In these tests please rename `refs` to `props`. And `props` to `deepProps`. Here is the breaking changes release notes: https://github.com/stampit-org/stampit/releases/tag/v4.0.0

types/stampit/stampit-tests.ts
```diff
@@ -149,15 +173,126 @@ interface SomeStampInstance {
...
     .init(function(params: { a: number; b: boolean}) {
         this.a = '' + a;
         this.b = '' + b;
-    }) as SomeStamp;
+    });
 
 SomeStamp({ a: 1, b: false }); // $ExpectType SomeStampInstance
 SomeStamp({ a: 1, b: false }).a; // $ExpectType string
+
+const d: stampit.Descriptor<{}> = {
```
koresar: This should be `stampit.ExtendedDescriptor`

koresar (changes requested):
Some changes are necessary. See comments.

types/stampit/index.d.ts
koresar:
```
compose: ComposeMethod & Descriptor;
```

types/stampit/stampit-tests.ts
PopGoesTheWza: In the context of `stampit` types, what would be the benefit of distinguishing Stamp's `Descriptor` from `stampit`'s `Descriptor`?

types/stampit/index.d.ts
PopGoesTheWza: Ok

types/stampit/index.d.ts
PopGoesTheWza: `undefined`? Because it can be anything?
Or should it be `void | S̤t̤a̤m̤p̤` similar to Initializer?

PopGoesTheWza:
@koresar only open issue left is regarding `MethodMap` for which if we loosen typing restriction we lose the benefit of in-context typed `this`

types/stampit/stampit-tests.ts
koresar: Can we maybe inherit `MethodMap` from `PropertyMap`?

koresar:
Yep. All is good. Just the `MethodMap` left.

I'll copy my question here: can we inherit `MethodMap` from the `PropertyMap`?

Something like:
```ts
interface MethodMap<This> extends PropertyMap {
    [s: string]: ((this: This, ...args: any[]) => any) | any;
}
```

PopGoesTheWza:
@koresar is the following acceptable?

```ts
interface MethodMap<This> {
    [s: string]: ((this: This, ...args: any[]) => any) | {};
}
```

Which allows this use of `methods`

```ts
const membership = stampit({
    methods: {
        members: {
            itCanBe: "anything",
        },
        add(member: any) {
            this.members[member.name] = member;
            return this;
        },
        getMember(name: any) {
            return this.members[name];
        }
    },
    props: {
        members: {}
    }
});
```

types/stampit/stampit-tests.ts
PopGoesTheWza: @koresar see my latest comment below.

types/stampit/index.d.ts
PopGoesTheWza: Done

koresar:
Sorry @PopGoesTheWza 
I was not clear explaining what type of data the `methods` can contain. The [spec](https://github.com/stampit-org/stamp-specification#the-stamp-descriptor) says that the `methods` can contain any data, not just functions. In other words, the following is pretty widespread:

```js
const Stamp = stampit({
  methods: {
    myMembers: {},
    dbConnString: "pgs://data.local",
    MY_LIMIT: 100,
    myRegExp: /foobar/g,
   // literally any other type
  }
});
```

This whole confusion is coming from the fact, that the `Stamp.compose.methods` becomes the prototype of object instances created from the stamp:

```js
Stamp.compose.methods === Stamp().__proto__
```

Can we please relax the `methods` so that the above works for people?

koresar:
Maybe this? (I replaced the `{}` with `any`.

```ts
interface MethodMap<This> {
  [s: string]: ((this: This, ...args: any[]) => any) | any;
}
```

PopGoesTheWza:
@koresar current type definition for `MethodMap` allows the *intersection* of

- a set of functions/methods
- and a pojo.

Thus the following `methods` is valid and gets **correct** typing:

```ts
const descriptor0: stampit.ExtendedDescriptor<Object0> = {
methods: {
  logLocalThis() {
      this; // $ExpectType Object0
  },
  myMembers: {},
  dbConnString: "pgs://data.local",
  MY_LIMIT: 100,
  myRegExp: /foobar/g,
},
[...]
```

As per the specification, I could be mistaken but it reads: **`methods` - A set of methods that will be added to the object's delegate prototype.** ;)

So I guess we're all good in the fact that we cover both the specification and `stampit` extended definition for `methods`

koresar:
Okay. Let's merge it then! :)

koresar (approved):

PR #31020, opened 2018-12-03
iframe-resizer: make iframeResizer default export

polar-lukas:
"iframeResizer" is the default export of the typed library. Hence it should also be declared such.

If you try to require iframeResizer as AMD module, the function "iframeResizer" is not provided within an object but as function. Using the provided types results in an undefined propery "iframeResizer" on the "define"d object.

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: 
https://github.com/davidjbradshaw/iframe-resizer/blob/d0e8addce5ff9c657292c5732cceb6cfe90b23ae/src/iframeResizer.js#L1296
https://github.com/davidjbradshaw/iframe-resizer/blob/d0e8addce5ff9c657292c5732cceb6cfe90b23ae/src/iframeResizer.js#L1251
- [x] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #39193, opened 2019-10-16
iframe-resizer - Update to fix function export

chuskeyLT:
      https://github.com/DefinitelyTyped/DefinitelyTyped/pull/31020#issuecomment-509014842
      https://github.com/davidjbradshaw/iframe-resizer/issues/478

### DIFF
--- a/types/iframe-resizer/iframe-resizer-tests.ts
+++ b/types/iframe-resizer/iframe-resizer-tests.ts
@@ -3,8 +3,8 @@ import ifr = require("iframe-resizer");
 
 function testOne(): void {
   const iframe: HTMLIFrameElement = document.createElement('iframe');
-  const options: iframeResizer.IFrameOptions = {log: true};
-  const components: iframeResizer.IFrameComponent[] = iframeResizer.iframeResizer(options, iframe);
+  const options: iframeResizer.IFrameOptions = { log: true };
+  const components: iframeResizer.IFrameComponent[] = iframeResizer(options, iframe);
   if (components) {
     components.forEach(component => console.log(component.iFrameResizer));
   } else {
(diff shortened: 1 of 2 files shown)


weswigham:
cc @sandersn who went over #31020 - the library, at least nowadays, definitely has some [distinct named exports](https://runkit.com/embed/2yyvte23tedq). Thing is - that totally mismatches with what the readme says, and what the expected browser usage is. `require`ing the package vs using the minified bundle seems to result in completely different APIs.

sandersn (changes requested):
I think you can still `import iframeResizer = require("iframeResizer")`, right? (Or `import iframeResizer from 'iframeResizer'` with `--esModuleInterop`.) I sounds like the library supports both.

In that case I think the better fix is to just add a `iframeResizer` function that's exported from the namespace:

```ts
export namespace iframeResizer {
  // ...
  function iframeResizer(options: IFrameOptions, target: string | HTMLElement): IFrameComponent[];
}
```

And then the tests can make sure that both work:

```ts
import ifr = require('iframeResizer')
import { iframeResizer } from 'iframeResizer'
```

types/iframe-resizer/iframe-resizer-tests.ts
```diff
@@ -1,9 +1,10 @@
-import iframeResizer = require("iframe-resizer");
+import { iframeResizer } from 'iframe-resizer';
+import ifr = require("iframe-resizer");
```
sandersn: `ifr` should still be callable, so (1) there should be a test of that (2) you'll still need all the previous code in index.d.ts.

chuskeyLT: @sandersn - I think this is what you were referring to. This is the combination I could find that would allow `ifr` to be callable and still allow the `import { iframeResizer } from 'iframe-resizer';` version to work as expected.

sandersn (approved):

PR #39196, opened 2019-10-17
fix(newrelic): make noticeError customAttributes signature wider

zieka:
Please fill in this template.

     - https://docs.newrelic.com/docs/browser/new-relic-browser/browser-agent-spa-api/notice-error
        ```js
        try {
          // code that throws an error
        } catch (err) {
          newrelic.noticeError(err, { attribute1: 'value1', attribute2: 2 });
        }
        ```
     - https://docs.newrelic.com/docs/browser/new-relic-browser/browser-agent-spa-api/set-custom-attribute
        > Custom attribute values cannot be complex objects, only simple types such as Strings and Integers.

### DIFF
--- a/types/newrelic/index.d.ts
+++ b/types/newrelic/index.d.ts
@@ -67,14 +67,14 @@ export function setControllerName(name: string, action: string): void;
  *
  * Most recently set value wins.
  */
-export function addCustomAttribute(key: string, value: string | number): void;
+export function addCustomAttribute(key: string, value: string | number | boolean): void;
 
 /**
  * Adds all custom attributes in an object to the current transaction.
  *
  * See documentation for `addCustomAttribute` for more information on setting custom attributes.
  */
-export function addCustomAttributes(atts: { [key: string]: string | number }): void;
+export function addCustomAttributes(atts: { [key: string]: string | number | boolean }): void;
 
 /**
  * Tell the tracer whether to ignore the current transaction.
--- a/types/newrelic/index.d.ts
+++ b/types/newrelic/index.d.ts
@@ -3,6 +3,7 @@
 // Definitions by: Matt R. Wilson <https://github.com/mastermatt>
 //                 Brooks Patton <https://github.com/brookspatton>
 //                 Michael Bond <https://github.com/MichaelRBond>
+//                 Kyle Scully <https://github.com/zieka>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 // https://docs.newrelic.com/docs/agents/nodejs-agent/api-guides/nodejs-agent-api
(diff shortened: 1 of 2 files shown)


types/newrelic/index.d.ts
```diff
@@ -93,7 +94,7 @@ export function setIgnoreTransaction(ignored: boolean): void;
  *
  *  Optional. Any custom attributes to be displayed in the New Relic UI.
  */
-export function noticeError(error: Error, customAttributes?: { [key: string]: string }): void;
+export function noticeError(error: Error, customAttributes?: { [key: string]: string | number }): void;
```
zieka: this is the main change; the rest is a result of prettier

mastermatt:
So interesting enough, the docs you linked to were for the browser API. But as far as I know, these types are only intended for the Node agent, whose docs do not specify what the custom attributes can be https://docs.newrelic.com/docs/agents/nodejs-agent/api-guides/nodejs-agent-api#noticeError

I dug around the tests and examples in the https://github.com/newrelic/node-newrelic to no avail before digging into the meat of the code. The values are passed around, never validated, and eventually end up in being added to an `ErrorTraceAggregator` in an error collector before I lose the trail. 

I'd actually be curious to know if @astormnewrelic could clarify what type of data is valid for custom attributes. I presume `addCustomAttribute`, `addCustomAttribute`, and `noticeError` all digest the attributes the same. Are they limited to strings and numbers?

zieka:
Hey @mastermatt
so i did some code spelunking and it looks like it checks to make sure they are string | number | boolean:
https://github.com/newrelic/node-newrelic/blob/master/lib/util/attribute-types.js#L3-L18

I'll widen the type on my current PR to include boolean

types/newrelic/index.d.ts
```diff
@@ -66,14 +67,14 @@ export function setControllerName(name: string, action: string): void;
...
-export function addCustomAttribute(key: string, value: string|number): void;
+export function addCustomAttribute(key: string, value: string | number): void;
 
 /**
  * Adds all custom attributes in an object to the current transaction.
  *
  * See documentation for `addCustomAttribute` for more information on setting custom attributes.
  */
-export function addCustomAttributes(atts: { [key: string]: string|number }): void;
+export function addCustomAttributes(atts: { [key: string]: string | number }): void;
```
mastermatt:
```suggestion
export function addCustomAttributes(atts: { [key: string]: string | number | boolean }): void;
```

types/newrelic/index.d.ts
```diff
@@ -66,14 +67,14 @@ export function setControllerName(name: string, action: string): void;
  *
  * Most recently set value wins.
  */
-export function addCustomAttribute(key: string, value: string|number): void;
+export function addCustomAttribute(key: string, value: string | number): void;
```
mastermatt:
```suggestion
export function addCustomAttribute(key: string, value: string | number | boolean): void;
```

mastermatt (commented):
I skimmed right pass that call, thanks for finding that @zieka.
Can we add the bool update to the other attr signatures too.

mastermatt (approved):

astormnewrelic:
@mastermatt Woah, neat, types! Just to repeat what you'all already found by spelunking, attributes values are generally limited to string/number/boolean values.

PR #39199, opened 2019-10-17
[use-global-hook] Updates def for simpler usage

ojameso:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: N/A, no changes, just better type def.

### DIFF
--- a/types/use-global-hook/index.d.ts
+++ b/types/use-global-hook/index.d.ts
@@ -4,23 +4,36 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.6
 
-import React = require('react');
-type ReactType = typeof React;
+// Use an interface so that different versions of React can be used
+interface ReactInterface {
+  // tslint:disable-next-line ban-types
+  useEffect: Function;
+  // tslint:disable-next-line ban-types
+  useState: Function;
+  // tslint:disable-next-line ban-types
+  useMemo: Function;
+}
 // to ignore strict-export-declare-modifiers error
 export {};
 
 // Where S is typeof state and A is typeof associated actions
 export interface Store<S, A> {
-    state: S;
-    actions: A;
-    setState(state: S, afterUpdateCallback?: () => void): void;
+  state: S;
+  actions: A;
+  setState(state: S, afterUpdateCallback?: () => void): void;
 }
 
 export type InitializerFunction<S, A> = (store: Store<S, A>) => void;
 
+type UseGlobal<S, A> = (() => [S, A]) &
+  (<NS>(stateFunc: (state: S) => NS) => [NS, A]) &
+  (<NS, NA>(stateFunc: (state: S) => NS, actionsFunc: (state: A) => NA) => [NS, NA]) &
+  // tslint:disable-next-line no-unnecessary-generics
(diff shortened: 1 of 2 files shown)


types/use-global-hook/index.d.ts
```diff
@@ -4,23 +4,36 @@
...
+  setState(state: S, afterUpdateCallback?: () => void): void;
 }
 
 export type InitializerFunction<S, A> = (store: Store<S, A>) => void;
 
+type UseGlobal<S, A> = (() => [S, A]) &
+  (<NS>(stateFunc: (state: S) => NS) => [NS, A]) &
+  (<NS, NA>(stateFunc: (state: S) => NS, actionsFunc: (state: A) => NA) => [NS, NA]) &
+  // tslint:disable-next-line no-unnecessary-generics
+  (<NS, NA>(stateFunc: undefined, actionsFunc: (state: A) => NA) => [NS, NA]);
```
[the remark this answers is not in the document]
ojameso: Great catch!

types/use-global-hook/index.d.ts
```diff
@@ -4,23 +4,36 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.6
 
-import React = require('react');
-type ReactType = typeof React;
+// Use an interface so that different versions of React can be used
+interface ReactInterface {
+  // tslint:disable-next-line ban-types
+  useEffect: Function;
```
weswigham: Just use `(...args: any[]) => any` rather than the `Function` global and a lint suppression.

types/use-global-hook/index.d.ts
```diff
@@ -4,23 +4,36 @@
...
+  // tslint:disable-next-line ban-types
+  useState: Function;
```
weswigham: Same here.

types/use-global-hook/index.d.ts
```diff
@@ -4,23 +4,36 @@
...
+  // tslint:disable-next-line ban-types
+  useMemo: Function;
```
weswigham: And here.

types/use-global-hook/index.d.ts
```diff
@@ -4,23 +4,36 @@
...
+  actions: A;
+  setState(state: S, afterUpdateCallback?: () => void): void;
 }
 
 export type InitializerFunction<S, A> = (store: Store<S, A>) => void;
 
+type UseGlobal<S, A> = (() => [S, A]) &
+  (<NS>(stateFunc: (state: S) => NS) => [NS, A]) &
+  (<NS, NA>(stateFunc: (state: S) => NS, actionsFunc: (state: A) => NA) => [NS, NA]) &
+  // tslint:disable-next-line no-unnecessary-generics
```
weswigham: Is this suppression needed? `NA` certainly seems to be used in both the `actionsFunc` argument and the return position.

weswigham (changes requested):

types/use-global-hook/index.d.ts
ojameso: Makes sense. 👍

types/use-global-hook/index.d.ts
ojameso: Was from a previous def and forgot to remove. Thanks for the catch!

weswigham (approved):

PR #39202, opened 2019-10-17
[slonik] update types to support sql.join

bram-l:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/gajus/slonik/commit/910177db4e8bdb8f2d2b4a9e82bd017044a6801d#diff-196906dac4d96fa06e8bde4437b30d65

As mentioned in the [#38524](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38524#issuecomment-537039019) support for `sql.join` was still missing from the Slonik v19 types.

### DIFF
--- a/types/slonik/index.d.ts
+++ b/types/slonik/index.d.ts
@@ -26,6 +26,17 @@ export type LogicalBooleanOperatorType = 'AND' | 'OR';
 // EXPRESSIONS AND TOKENS
 // ----------------------------------------------------------------------
 
+export type TypeNameIdentifierType =
+  'bool' |
+  'bytea' |
+  'float4' |
+  'float8' |
+  'int2' |
+  'int4' |
+  'json' |
+  'text' |
+  'timestamptz';
+
 export type SerializableValueType =
     | string
     | number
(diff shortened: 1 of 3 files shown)


types/slonik/index.d.ts
```diff
@@ -281,20 +304,24 @@ export interface SqlTaggedTemplateType {
...
+    ) => SqlTokenType;
     unnest: (
-        // Value might be PrimitiveValueExpressionType[],
+        // Value might be ReadonlyArray<ReadonlyArray<PrimitiveValueExpressionType>>,
         // or it can be infinitely nested array, e.g.
         // https://github.com/gajus/slonik/issues/44
-        tuples: any[][],
-        columnTypes: string[]
+        tuples: ReadonlyArray<ReadonlyArray<any>>,
+        columnTypes: ReadonlyArray<string>
```
mmkal: What's the benefit of requiring `ReadonlyArray<string>`? It's fine if the function is always used inline, but slightly more inconvenient to work with otherwise, since now I can't do:

```typescript
const tuples = [['a', 'b']]
const columnTypes = ['foo', 'bar']
...
slonik.unnest(tuples, columnTypes)
```

mmkal (approved):

types/slonik/index.d.ts
bram-l: I added the `ReadonlyArray` type in line with the [original Flow types](https://github.com/gajus/slonik/blob/master/src/types.js#L349). To my understanding there are no restrictions to the array passed in as parameter, but the `ReadonlyArray` would simply imply that the function itself is not performing any mutations on the provided array.

PR #39204, opened 2019-10-17
[@types/react-modal] Allow `null` in bodyOpenClassName, htmlOpenClassName & role

pleunv:
As per the [docs](https://github.com/reactjs/react-modal/blob/2ae092aecb376170b72ec3d7e069670cbaff275c/docs/styles/classes.md#for-the-documentbody-and-html-tag) and [tests](https://github.com/reactjs/react-modal/blob/c747c247ae22bf09dd4188719a3bfa7ce1c644c0/specs/Modal.spec.js) the following props can be `null`:

* `bodyOpenClassName`
* `htmlOpenClassName`
* `role`

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [docs](https://github.com/reactjs/react-modal/blob/2ae092aecb376170b72ec3d7e069670cbaff275c/docs/styles/classes.md#for-the-documentbody-and-html-tag) and [tests](https://github.com/reactjs/react-modal/blob/c747c247ae22bf09dd4188719a3bfa7ce1c644c0/specs/Modal.spec.js)

### DIFF
--- a/types/react-modal/index.d.ts
+++ b/types/react-modal/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for react-modal 3.8
+// Type definitions for react-modal 3.10
 // Project: https://github.com/reactjs/react-modal
 // Definitions by: Rajab Shakirov <https://github.com/radziksh>,
 //                 Drew Noakes <https://github.com/drewnoakes>,
@@ -50,11 +50,11 @@ declare namespace ReactModal {
         /* String className to be applied to the portal. Defaults to "ReactModalPortal". */
         portalClassName?: string;
 
-        /* String className to be applied to the document.body. */
-        bodyOpenClassName?: string;
+        /* String className to be applied to the document.body (must be a constant string). When set to null it doesn't add any class to document.body. */
+        bodyOpenClassName?: string | null;
 
-        /* String className to be applied to the document.html. */
-        htmlOpenClassName?: string;
+        /* String className to be applied to the document.html (must be a constant string). Defaults to null. */
+        htmlOpenClassName?: string | null;
 
         /* String or object className to be applied to the modal content. */
         className?: string | Classes;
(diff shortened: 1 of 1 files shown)


types/react-modal/index.d.ts
```diff
@@ -50,11 +50,11 @@ declare namespace ReactModal {
         /* String className to be applied to the portal. Defaults to "ReactModalPortal". */
         portalClassName?: string;
 
-        /* String className to be applied to the document.body. */
-        bodyOpenClassName?: string;
+        /* String className to be applied to the document.body (must be a constant string). When set to null it doesn't add any class to document.body. */
+        bodyOpenClassName?: string | null;
```
drewnoakes: Would this be clearer as:

```suggestion
        bodyOpenClassName: string | null | undefined;
```

Can this (and the other changed properties) actually be both `null` and `undefined`?

ttamminen: I think the `bodyOpenClassName` is optional, so optional parameter (`bodyOpenClassName?`) is semantically correct.

The end result is the same, `?` adds the `undefined` to the type, but at least for me, it is easier to read the type if I can quickly glance what is mandatory and what is optional instead of searching `undefined` from the list of types separated by pipe operators.

pleunv: All of them are indeed optional but nullable. In which case I tend to prefer the optional notation as well.

ttamminen (approved):

types/react-modal/index.d.ts
andrewbranch: Removing the `?` is definitely not what you want. It would force users to supply the prop on JSX elements 100% of the time, even if to set to `undefined`: https://www.typescriptlang.org/play/?jsx=2#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4BYAKCpgE8wk4AFHMAZzgF44BvONFADsAQkgCqggCZJMwQUkkAuOGxhQ5AczgAfOAFcpMuQrgBfCpWloANiiJxMBjMAiC4AYVyR5gmAAowVjZlFgh2AEplACkAZQANADoAUWskECRfCzRXVX4uOAAeT3BXDPgAegA+LJz4SXyir1LffiFRCWlZeUlOHgNO40lTOCrSIA

PR #39206, opened 2019-10-17
vuelidate: Add function call signature to ValidationRule

bbugh:
`ValidationRule` is a function object that, when called, takes in parameters and then applies them to the inner validator function. The call signature for `ValidationRule` to behave as a function is currently missing. This PR simply adds that call signature to `ValidationRule`. 

We ran into this while writing tests for built-in validator-composed functions, like so:

```typescript
export const greaterThanZero: ValidationRule = and(
  minValue(0),
  not((value: UserInput) => value == 0)
);
export const greaterThanOrEqualToZero: ValidationRule = minValue(0);
```

For example, `greaterThanOrEqualToZero` could written as a `CustomRule` like so:

```typescript
const greaterThanOrEqualToZero = (value: UserInput) => !helpers.req(value) || value! >= 0
```

But it can also just be composed from the existing `ValidationRule`, `minValue`:

```typescript
const greaterThanOrEqualToZero: ValidationRule = minValue(0);
```

These behave identically in a validation form context, but we were unable to correctly unit test them without this type fix. Simply adding this call signature allowed our tests to pass. We also tested thoroughly through the `validations` option and various form setups before submitting this PR.

Vuelidate `withParams` reference that demonstrates `ValidationRule` behaves as a function: https://github.com/vuelidate/vuelidate/blob/bb23f01ab4f14a813a25cff8defb7788cd8859ac/src/params.js#L36-L55

---

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [Vuelidate function return reference for a ValidationRule](https://github.com/vuelidate/vuelidate/blob/bb23f01ab4f14a813a25cff8defb7788cd8859ac/src/params.js#L36-L55)

### DIFF
--- a/types/vuelidate/vuelidate-tests.ts
+++ b/types/vuelidate/vuelidate-tests.ts
@@ -19,7 +19,7 @@ const contains = (param: string): CustomRule =>
 
 const mustBeCool3 = helpers.withParams(
     { type: 'mustBeCool3' },
-    (value) => !helpers.req(value) || value.indexOf('cool') >= 0
+    (value: any) => !helpers.req(value) || value.indexOf('cool') >= 0
 )
 
 const mustBeCool3Result: boolean = mustBeCool3(50)
--- a/types/vuelidate/vuelidate-tests.ts
+++ b/types/vuelidate/vuelidate-tests.ts
@@ -22,6 +22,8 @@ const mustBeCool3 = helpers.withParams(
     (value) => !helpers.req(value) || value.indexOf('cool') >= 0
 )
 
+const mustBeCool3Result: boolean = mustBeCool3(50)
+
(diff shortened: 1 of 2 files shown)


types/vuelidate/vuelidate-tests.ts
```diff
@@ -19,9 +19,11 @@ const contains = (param: string): CustomRule =>
 
 const mustBeCool3 = helpers.withParams(
     { type: 'mustBeCool3' },
-    (value) => !helpers.req(value) || value.indexOf('cool') >= 0
+    (value: any) => !helpers.req(value) || value.indexOf('cool') >= 0
```
bbugh: The `TypeScript@next` test appears to have `noImplicitAny` enabled, which broke on previously accepted code that I didn't change. I updated with an explicit `any` to allow the `TypeScript@next` test to pass.

PR #39209, opened 2019-10-17
Add support for node 12.2 module.createRequire(filename) method

anstarovoyt:
1. Add support for node 12.2 module.createRequire(filename) method
2. Deprecated module. createRequireFromPath(filename)
3. Update return type for createRequireFromPath (use NodeRequireFunction)

https://nodejs.org/api/modules.html#modules_module_createrequire_filename

- [ x ] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/api/modules.html#modules_module_createrequire_filename
- [ x ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.

### DIFF
--- a/types/node/globals.d.ts
+++ b/types/node/globals.d.ts
@@ -1132,7 +1132,12 @@ declare namespace NodeJS {
     class Module {
         static runMain(): void;
         static wrap(code: string): string;
-        static createRequireFromPath(path: string): (path: string) => any;
+
+        /**
+         * @deprecated Deprecated since: v12.2.0. Please use createRequire() instead.
+         */
+        static createRequireFromPath(path: string): NodeRequireFunction;
+        static createRequire(path: string): NodeRequireFunction;
         static builtinModules: string[];
 
         static Module: typeof Module;
--- a/types/node/globals.d.ts
+++ b/types/node/globals.d.ts
@@ -1137,7 +1137,7 @@ declare namespace NodeJS {
          * @deprecated Deprecated since: v12.2.0. Please use createRequire() instead.
          */
         static createRequireFromPath(path: string): NodeRequireFunction;
-        static createRequire(path: string): NodeRequireFunction;
+        static createRequire(path: string | URL): NodeRequireFunction;
         static builtinModules: string[];
 
         static Module: typeof Module;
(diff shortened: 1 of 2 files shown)


types/node/globals.d.ts
```diff
@@ -1132,7 +1132,12 @@ declare namespace NodeJS {
     class Module {
         static runMain(): void;
         static wrap(code: string): string;
-        static createRequireFromPath(path: string): (path: string) => any;
+
+        /**
+         * @deprecated Deprecated since: v12.2.0. Please use createRequire() instead.
+         */
+        static createRequireFromPath(path: string): NodeRequireFunction;
+        static createRequire(path: string): NodeRequireFunction;
```
Flarna: `createRequire` allows `string | URL`

anstarovoyt: updated

anstarovoyt: removed the url parameter. Actually I think we should import url from "url.d.ts" but I am not sure how it must be done in right way.

SimonSchick: iirc referencing modules in a global context doesn't work.

Flarna (approved):

types/node/globals.d.ts
andrewbranch: `string | import('url').URL`?

andrewbranch: That was added in TypeScript 2.9, though. To support your TS 2.1 definition, I believe it would work to `import { URL } from "url";` at the top and then wrap the rest of the file in `declare global { ... }`.

Flarna: I tried the `declare global {}` variant but failed. it complains in process.d.ts that node typings are missing.

Off topic but I have to comment regarding "To support **_your_** TS 2.1 definition": we asked several times to raise the min typescript version for definitions here but this was rejected by the maintainers of this repo.
Doing it by hand means updating >1000 dependents and this ends up in a PR that unicorns because of a too big changeset and as a result gets never merged as it can't be reviewed.

anstarovoyt:
any updates on this?

types/node/globals.d.ts
andrewbranch: Ugh, it looks like the thing that’s preventing `declare global` is a bug that ran all the way up until 2.9: https://github.com/microsoft/TypeScript/issues/23796

anstarovoyt:
awesome! Thank you

PR #39211, opened 2019-10-17
fix: enums for googlemaps

simonhaenisch:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: see below

I've had some troubles with enum types, because a lot of them are actually strings but weren't typed as such.

To get all the values of the enums, I wrote a script that prints them to the console from the actual api code. See this fiddle: https://jsfiddle.net/9u8t4w01/.

```html
<script src="https://maps.googleapis.com/maps/api/js?libraries=places,drawing,geometry,visualization,adsense"></script>
```

```js
const getByPath = (object, path) => path.split('.').reduce((nestedObject, key) => nestedObject && nestedObject[key], object);

const enums = ['MapTypeControlStyle', 'ScaleControlStyle', 'ZoomControlStyle', 'ControlPosition', 'StrokePosition', 'GeocoderStatus', 'GeocoderLocationType', 'TravelMode', 'UnitSystem', 'TransitMode', 'TransitRoutePreference', 'TrafficModel', 'DirectionsStatus', 'ElevationStatus', 'MaxZoomStatus', 'DistanceMatrixStatus', 'DistanceMatrixElementStatus', 'KmlLayerStatus', 'StreetViewPreference', 'StreetViewSource', 'StreetViewStatus', 'adsense.AdFormat', 'places.PlacesServiceStatus', 'places.RankBy', 'drawing.OverlayType', 'visualization.MapsEngineStatus'];

enums.map(key => console.log(JSON.stringify({ [key]: getByPath(google.maps, key) }, null, 2)));
```

The `PlacesServiceStatus` enum had a field `ERROR` which doesn't actually exist as a type (also see https://developers.google.com/maps/documentation/javascript/reference/places-service#PlacesServiceStatus). Therefore I removed it, but I'm not sure: would this be considered a breaking change for the types?

### DIFF
--- a/types/googlemaps/index.d.ts
+++ b/types/googlemaps/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for Google Maps JavaScript API 3.37
+// Type definitions for Google Maps JavaScript API 3.38
 // Project: https://developers.google.com/maps/
 // Definitions by: Chris Wrench <https://github.com/cgwrench>,
 //                 Kiarash Ghiaseddin <https://github.com/Silver-Connection>,
@@ -12,6 +12,7 @@
 //                 Colin Doig <https://github.com/captain-igloo>
 //                 Dmitry Demensky <https://github.com/demensky>
 //                 Vladimir Dashukevich <https://github.com/life777>
+//                 Simon Haenisch <https://github.com/simonhaenisch>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 // TypeScript Version: 2.7
@@ -286,9 +287,11 @@ declare namespace google.maps {
     }
 
     enum MapTypeControlStyle {
-        DEFAULT,
-        DROPDOWN_MENU,
-        HORIZONTAL_BAR,
+        DEFAULT = 0,
+        DROPDOWN_MENU = 1,
+        HORIZONTAL_BAR = 2,
+        INSET = 3,
+        INSET_LARGE = 4,
     }
 
     type GestureHandlingOptions = 'cooperative' | 'greedy' | 'none' | 'auto';
@@ -343,7 +346,7 @@ declare namespace google.maps {
     }
 
     enum ScaleControlStyle {
-        DEFAULT,
+        DEFAULT = 0,
     }
 
     /** Options for the rendering of the Street View pegman control on the map. */
@@ -370,9 +373,9 @@ declare namespace google.maps {
     }
 
     enum ZoomControlStyle {
-        DEFAULT,
-        LARGE,
-        SMALL,
+        DEFAULT = 0,
+        LARGE = 1,
+        SMALL = 2,
     }
 
     /**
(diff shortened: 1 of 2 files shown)


types/googlemaps/index.d.ts
```diff
@@ -381,45 +384,55 @@ declare namespace google.maps {
...
+        RIGHT_TOP = 7,
+        /** Same as `TOP_CENTER`. */
+        TOP = 2,
+        /** Elements are positioned in the center of the top row. */
+        TOP_CENTER = 2,
         /** Elements are positioned in the top right and flow towards the middle. */
-        TOP_RIGHT,
+        TOP_LEFT = 1,
+        /** Elements are positioned in the top right and flow towards the middle. */
+        TOP_RIGHT = 3,
```
demensky: Please mark options that are not in the [documentation] as `@deprecated`.

image


[documentation]: https://developers.google.com/maps/documentation/javascript/reference/control#ControlPosition

types/googlemaps/index.d.ts
```diff
@@ -381,45 +384,55 @@ declare namespace google.maps {
...
-        RIGHT_CENTER,
+        RIGHT_CENTER = 8,
         /** Elements are positioned on the right, below top-right elements, and flow downwards. */
-        RIGHT_TOP,
-        /**    Elements are positioned in the center of the top row. */
-        TOP_CENTER,
-        /** Elements are positioned in the top left and flow towards the middle. */
-        TOP_LEFT,
+        RIGHT_TOP = 7,
+        /** Same as `TOP_CENTER`. */
```
demensky: Use `@see` for better experience.

```suggestion
        /** @see {@link TOP_CENTER} */
```

image

demensky (changes requested):

types/googlemaps/index.d.ts
simonhaenisch: Thanks for the comments, I wasn't aware that these were deprecated, so I decided to remove them instead.

types/googlemaps/index.d.ts
```diff
@@ -286,9 +287,11 @@ declare namespace google.maps {
     }
 
     enum MapTypeControlStyle {
-        DEFAULT,
-        DROPDOWN_MENU,
-        HORIZONTAL_BAR,
+        DEFAULT = 0,
+        DROPDOWN_MENU = 1,
+        HORIZONTAL_BAR = 2,
```
demensky: Wrong `DROPDOWN_MENU` & `HORIZONTAL_BAR` value.

image

types/googlemaps/index.d.ts
```diff
@@ -370,9 +373,9 @@ declare namespace google.maps {
     }
 
     enum ZoomControlStyle {
-        DEFAULT,
-        LARGE,
-        SMALL,
+        DEFAULT = 0,
+        LARGE = 1,
+        SMALL = 2,
```
demensky: Wrong `LARGE` & `SMALL` value.

image

types/googlemaps/index.d.ts
```diff
@@ -3373,8 +3376,8 @@ declare namespace google.maps {
         }
 
         enum RankBy {
-            DISTANCE,
-            PROMINENCE,
+            DISTANCE = 1,
+            PROMINENCE = 0,
```
demensky: Wrong `DISTANCE` & `PROMINENCE` value.

image

demensky (changes requested):
I suggest reordering to avoid confusion in the future.

types/googlemaps/index.d.ts
simonhaenisch: The values are correct, but yeah the order was wrong.

types/googlemaps/index.d.ts
simonhaenisch: Thanks, fixed those and sorted them according to the enums. I was trying to keep the previous ordering (by name) in place, but yeah it obviously confused me a bit 🤓

demensky (approved):

PR #39215, opened 2019-10-17
Add types for JSX extensions to ESTree AST

antross:
Declare JSX types and export a modified `Node` type containing
the union of `Node` from `@types/estree` plus JSX node types.

Like `@types/estree` this is a types-only package to aid consuming
ASTs created by various JavaScript parsers used in tooling.

A couple questions:
1. This imports types from `@types/estree`. Do I need to do anything special to ensure this is referenced in the final NPM package or is that handled automatically based on the `import`?
2. Since this is a types-only package I had to disable the `dt-header` and `npm-naming` rules in `tslint.json`. I was not able to figure out how to do this using a disable comment. Is there a better approach for headers in types-only packages to pass linting?

Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/estree-jsx/estree-jsx-tests.ts
+++ b/types/estree-jsx/estree-jsx-tests.ts
@@ -1,0 +1,122 @@
+import * as ESTree from 'estree';
+
+import {
+    Node,
+    JSXIdentifier,
+    JSXNamespacedName,
+    JSXMemberExpression,
+    JSXEmptyExpression,
+    JSXExpressionContainer,
+    JSXSpreadChild,
+    JSXSpreadAttribute,
+    JSXAttribute,
+    JSXOpeningElement,
+    JSXOpeningFragment,
+    JSXClosingElement,
+    JSXClosingFragment,
+    JSXElement,
+    JSXFragment,
+    JSXText
+} from 'estree-jsx';
+
+declare let node: Node;
+declare let identifier: JSXIdentifier;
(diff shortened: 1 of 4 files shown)


andrewbranch:
I think this makes sense, but pinging @RReverser (owner of @types/estree) for general thoughts.

> 1. This imports types from `@types/estree`. Do I need to do anything special to ensure this is referenced in the final NPM package or is that handled automatically based on the `import`?

It’s handled automatically.

> 2. Since this is a types-only package I had to disable the `dt-header` and `npm-naming` rules in `tslint.json`. I was not able to figure out how to do this using a disable comment. Is there a better approach for headers in types-only packages to pass linting?

I think this is the correct thing to do.

RReverser:
This sounds like a great addition, but I'm afraid I'm rather swamped at the moment to do a proper review. Feel free to either proceed without me or ping me back in a ~week. Thank you!

andrewbranch (approved):
I just wanted to ensure this didn’t belong in estree proper. Thanks @RReverser and @antross!

antross:
Awesome, thanks!

@andrewbranch By the way, I see it currently published as `v0.0.0`. Is the version number normally pulled from the header comments (which may be why it ended up weird here)?

andrewbranch:
Yes, the major and minor version is pulled from the header, then the patch increments every time we publish an update. You can see @types/estree is also on the 0.0.x track due to not having a header.

PR #39220, opened 2019-10-17
[@types/enzyme] Add component display name selector generic overloads

ndresx:
This adds generic overloads for the selector functions when using a **component display name** instead of the component constructor (or any other valid selector type).

**Before:** 
```ts
shallowWrapper.find('AnotherComponent').props().anotherStringProp; // Error: Property 'anotherStringProp' does not exist on type 'HTMLAttributes'.ts(2339)
(shallowWrapper.find('AnotherComponent').props() as AnotherComponentProps).anotherStringProp; // OK
```

**~After:~**
```ts
shallowWrapper.find<AnotherComponentProps>('AnotherComponent').props().anotherStringProp; // OK
```

I have added the `// tslint:disable-line unified-signatures` flags to the newly added overloads because this way it's more explicit if you want to get typed props with a `displayName` selector. Otherwise, the linter suggests merging 3 of them together which ends up in not so clear signatures anymore.

----

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://airbnb.io/enzyme/docs/api/selector.html#3-a-react-components-displayname

### DIFF
--- a/types/enzyme/enzyme-tests.tsx
+++ b/types/enzyme/enzyme-tests.tsx
@@ -128,6 +128,7 @@ function ShallowWrapperTest() {
     let elementWrapper: ShallowWrapper<HTMLAttributes<{}>>;
     let anotherStatelessWrapper: ShallowWrapper<AnotherStatelessProps, never>;
     let anotherComponentWrapper: ShallowWrapper<AnotherComponentProps, any>;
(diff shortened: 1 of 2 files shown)


ndresx:
I'm self-closing this due to the fact that this is just a disguised type assertion. After reading through some more resources, it makes sense to actively cast it rather than using a generic for it.

PR #39229, opened 2019-10-18
[filesize-parser] Add types for filesize-parser

loilo:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/filesize-parser/index.d.ts
+++ b/types/filesize-parser/index.d.ts
@@ -4,7 +4,7 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 interface FilesizeParserOptions {
-    base?: number;
+    base?: 2 | 10;
 }
 
 interface StringLike {
--- a/types/filesize-parser/index.d.ts
+++ b/types/filesize-parser/index.d.ts
@@ -7,6 +7,8 @@ interface FilesizeParserOptions {
     base?: number;
 }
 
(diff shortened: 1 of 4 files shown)


types/filesize-parser/index.d.ts
```diff
@@ -0,0 +1,16 @@
+// Type definitions for filesize-parser 1.5
+// Project: https://github.com/patrickkettner/filesize-parser
+// Definitions by: Florian Reuschel <https://github.com/loilo>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+interface FilesizeParserOptions {
+    base?: 2 | 10;
+}
+
+interface StringLike {
```
weswigham: I wouldn't do this. This allow you to do innane things like `filesizeParser(42)`, or `filesizeParser(null)`. Just use `string`.

weswigham (changes requested):

types/filesize-parser/index.d.ts
loilo: I actually had `string` in the first commit, but since the library author explicitely calls `toString()` on the argument I figured that this would better represent original intent.

However, I'd definitely find `string` cleaner as well, so I'm happily going to count on your experience with this. :)

loilo: Again, now that I've reverted the commits I'm not sure about this anymore. Just noticed that the `filesizeParser(someNumber)` even is [in the library's tests](https://github.com/patrickkettner/filesize-parser/blob/master/test.js#L15).

PR #39230, opened 2019-10-18
[recharts] adds allowEscapeViewBox type to Tooltip

dimitropoulos:
now that https://github.com/recharts/recharts/pull/1908 is merged

### DIFF
--- a/types/recharts/index.d.ts
+++ b/types/recharts/index.d.ts
@@ -836,6 +836,8 @@ export interface Coordinate {
     y: number;
 }
 
+export type AllowEscapeViewBox = { x: boolean } | { y: boolean } | { x: boolean, y: boolean };
+
 export interface TooltipPayload {
     name: string;
     value: string | number | Array<string | number>;
@@ -850,6 +852,7 @@ export interface TooltipPayload {
 export interface TooltipProps extends Animatable {
     content?: React.ReactElement | React.StatelessComponent<any> | ContentRenderer<TooltipProps>;
     viewBox?: ViewBox;
+    allowEscapeViewBox?: AllowEscapeViewBox;
     active?: boolean;
     separator?: string;
     formatter?: TooltipFormatter;
--- a/types/recharts/index.d.ts
+++ b/types/recharts/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for Recharts 1.7
+// Type definitions for Recharts 1.8.0
 // Project: http://recharts.org/, https://github.com/recharts/recharts
 // Definitions by: Raphael Mueller <https://github.com/rapmue>
 //                 Roy Xue <https://github.com/royxue>
--- a/types/recharts/index.d.ts
+++ b/types/recharts/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for Recharts 1.8.0
+// Type definitions for Recharts 1.8
 // Project: http://recharts.org/, https://github.com/recharts/recharts
 // Definitions by: Raphael Mueller <https://github.com/rapmue>
 //                 Roy Xue <https://github.com/royxue>
--- a/types/recharts/index.d.ts
+++ b/types/recharts/index.d.ts
@@ -15,6 +15,7 @@
 //                 Dave Vedder <https://github.com/veddermatic>
 //                 Konstantin Azizov <https://github.com/g07cha>
 //                 Gonzalo Nicolas D'Elia <https://github.com/gndelia>
+//                 Dimitri Mitropoulos <https://github.com/dimitropoulos>
 //                 Eliot Ball <https://github.com/eliotball>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8


types/recharts/index.d.ts
```diff
@@ -834,6 +834,8 @@ export interface Coordinate {
     y: number;
 }
 
+export type AllowEscapeViewBox = { x: boolean } | { y: boolean } | { x: boolean, y: boolean };
```
G07cha: Why not mark them as optional?

```suggestion
export type AllowEscapeViewBox = { x?: boolean, y?: boolean };
```

dimitropoulos: because then a `allowEscapeViewBox={}` prop would be valid input, and, if you look at the code, doing such a thing will cause a typeerror in runtime

G07cha: Could you please share an example where it causes typeerror? Looking at merged PR in Recharts it seems like it's going to be totally fine since `allowEscapeViewBox` is still an object.

dimitropoulos: You know what.  You're right.  I stand corrected that it would cause a runtime error.

That said, the semantics of the code are in line with https://github.com/sindresorhus/type-fest/blob/master/source/require-at-least-one.d.ts.  I now, with your comment, realize that it wouldn't _fail_... but the semantics are still not quite right.

here's the code as it exists now:
```tsx
if (allowEscapeViewBox[key]) {
  return unrestricted;
}
```

If an empty object is passed in, this case will appropriately not trigger... but the intent is that there's always something there, that is to say, that the type of `allowEscapeViewBox.x` and `allowEscapeViewBox.y` are always `boolean`s, as opposed to `undefined`.

Still, if you feel passionately about it, I'm happy to make the change you suggest.

:)

G07cha: No, it's totally fine, I was just wondering if there is some hidden knowledge I was missing on 🙂

By the way, could you please increment typings versions since your addition in recharts would likely land in `1.8.4`?

dimitropoulos: sure! :)

dimitropoulos: just to be clear, you would increment to `1.7.1`, right?

G07cha: I would increment it to `1.8` to be aligned with version of the library

dimitropoulos: like this? https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39230/commits/996fbd33f4810bb8d537906f19cd625bfe8ed166

I know some projects go back and forth on whether to align the type version with the project because there's always that situation where the upstream releases `1.1.1` and the types are updated, but then someone discovers a bug in the types and suddenly the types are `1.1.2` and the project is still `1.1.1`.

Sounds good though :)

dimitropoulos: > I would also hold off merging it until a new version of recharts will be released containing the feature.

Sounds fine.  Though, I will note, that it's a backwards compatible change in that if someone does specify a value here it will just do nothing.  I'd prefer if we could merge it sooner than later so I can use it, but I can always copy/past into my project for now so no biggie :)

Thanks for your review!

types/recharts/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for Recharts 1.7
+// Type definitions for Recharts 1.8.0
```
G07cha:
```suggestion
// Type definitions for Recharts 1.8
```

types/recharts/index.d.ts
G07cha: >like this? 996fbd3

>I know some projects go back and forth on whether to align the type version with the project because there's always that situation where the upstream releases 1.1.1 and the types are updated, but then someone discovers a bug in the types and suddenly the types are 1.1.2 and the project is still 1.1.1.

Indeed, I, personally, don't have a preference for that but since the comment says `Type definitions for Recharts x.x` I assume it means that type definitions are aligned with specified version.

gndelia (approved):

gndelia:
LGTM

@dimitropoulos  consider adding yourself to the list of users, so you appear credited as well 😄

dimitropoulos:
> @dimitropoulos consider adding yourself to the list of users, so you appear credited as well

to which list do you refer?

gndelia:
in `index.d.ts`

```
// Type definitions for Recharts 1.8
// Project: http://recharts.org/, https://github.com/recharts/recharts
// Definitions by: Raphael Mueller <https://github.com/rapmue>
//                 Roy Xue <https://github.com/royxue>
//                 Zheyang Song <https://github.com/ZheyangSong>
//                 Rich Baird <https://github.com/richbai90>
//                 Dan Torberg <https://github.com/caspeco-dan>
//                 Peter Keuter <https://github.com/pkeuter>
//                 Jamie Saunders <https://github.com/jrsaunde>
//                 Harry Cruse <https://github.com/crusectrl>
//                 Andrew Palugniok <https://github.com/apalugniok>
//                 Robert Stigsson <https://github.com/RobertStigsson>
//                 Kosaku Kurino <https://github.com/kousaku-maron>
//                 Leon Ng <https://github.com/iflp>
//                 Dave Vedder <https://github.com/veddermatic>
//                 Konstantin Azizov <https://github.com/g07cha>
//                 Gonzalo Nicolas D'Elia <https://github.com/gndelia>

// You can add here yourself :)

// TypeScript Version: 2.8
```

you can check the readme 😄

PR #39231, opened 2019-10-18
[HOLD] [relay-runtime] Update fragment reference definitions

alloy:
⚠️ This is a breaking change, but alas there’s no other way around it.

As of TS 3.6 an intersection of multiple unique symbols results in
`never`, which means our fragment reference checking is now unsafe.

Depends on https://github.com/relay-tools/relay-compiler-language-typescript/pull/139

### DIFF
--- a/types/react-relay/test/react-relay-tests.tsx
+++ b/types/react-relay/test/react-relay-tests.tsx
@@ -2,7 +2,15 @@
 // tslint:disable:use-default-type-parameter
 
 import * as React from 'react';
-import { Environment, Network, RecordSource, Store, ConnectionHandler } from 'relay-runtime';
+import {
+    Environment,
+    Network,
+    RecordSource,
+    Store,
+    ConnectionHandler,
+    _FragmentRefs,
+    FragmentRefs,
+} from 'relay-runtime';
 
 import {
     commitMutation,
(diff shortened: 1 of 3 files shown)


alloy:
cc @djsheldrick

alloy:
So we tried to add `package.json` file that listed the latest version of the `relay-compiler-language-typescript` plugin as a peer dep, so people updating just the DT typings would get a warning about unmatched versions. Alas that doesn’t appear to be possible,

```
Error: /home/travis/build/DefinitelyTyped/DefinitelyTyped/types/relay-runtime/package.json should not include field peerDependencies
```

kastermester:
I'm a bit pressured for time right now, but having taken a quick glance at this I think it looks fine. :)

alloy:
@kastermester Thanks! I’ll go forward and self-merge now then, as we need to sync the release with the plugin.

renanmav (approved):

ds300:
Here's the relevant update to the plugin https://github.com/relay-tools/relay-compiler-language-typescript/pull/139

alloy:
We’re holding-off merging until we know the https://github.com/relay-tools/relay-compiler-language-typescript release process is in a good place. /cc @zephraph

alloy:
Alright, the new plugin version is released, so going to merge this now too.

renanmav:
Which plugin @alloy?

PR #39252, opened 2019-10-19
[@types/screeps] Add factories, strongholds, deposits and ruins

0l1v:
- Add factories and invader cores (structures, constants)
- Add all commodities (constants)
- Add deposits and ruins (interface and constants)
- Add some missing constants
- Add unboost method on StructureLab

- Better resource constants definition: no duplication anymore, and RESOURCE_CONSTANT is split in smaller sets which allows better typing of other interfaces (Mineral, StructureLab, ...)
- Define EventItem so as to leverage typescript discriminated union

- [OK] Use a meaningful title for the pull request. Include the name of the package modified.
- [OK] Test the change in your own code. (Compile and run.)
- [OK] Add or edit tests to reflect the change. (Run with `npm test`.)
- [OK] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [OK] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [OK] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [OK] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/screeps/screeps/commits/master and https://docs.screeps.com/api/
- [OK] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

### DIFF
--- a/types/screeps/index.d.ts
+++ b/types/screeps/index.d.ts
@@ -89,21 +89,26 @@ declare const CREEP_CLAIM_LIFE_TIME: 600;
 declare const CREEP_CORPSE_RATE: 0.2;
 
 declare const OBSTACLE_OBJECT_TYPES: [
-    "spawn",
-    "creep",
-    "wall",
-    "source",
-    "constructedWall",
-    "extension",
-    "link",
-    "storage",
-    "tower",
-    "observer",
-    "powerSpawn",
-    "powerBank",
-    "lab",
-    "terminal",
-    "nuker"
+    "spawn", 
+    "creep", 
+    "powerCreep", 
+    "source", 
+    "mineral", 
+    "deposit", 
+    "controller", 
+    "constructedWall", 
+    "extension", 
+    "link", 
+    "storage", 
+    "tower", 
+    "observer", 
+    "powerSpawn", 
+    "powerBank", 
+    "lab", 
+    "terminal", 
+    "nuker", 
+    "factory", 
+    "invaderCore"
 ];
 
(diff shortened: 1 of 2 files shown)


kotarou (approved):
Seems fine.

0l1v:
Thanks @Kotarou.
Actually I realised I forgot to extend the input parameters of Creep.withdraw and Creep.harvest to accept respectively Ruin and Deposit, it is now fixed (I also added some more constants which where not defined).

kotarou:
@0l1v For future changes, it would be good to go through screepers/typed-screeps instead of directly to DefinitleyTyped. 

I'm still okay with this being approved, since it seems good, but splitting the process is likely to cause issues later.

0l1v:
@kotarou Ok sorry, I was not aware of the process, I didn't know screepers/typed-screeps was the source of this package.
If screepers/typed-screeps is regularly ported here, then we should probably close this PR, right ?
Then I will submit a new one in screepers/typed-screeps.

0l1v:
The process is actually to deliver in screepers/typed-screeps first, sorry, this is my first contribution and I was not aware of this.

PR #39253, opened 2019-10-19
[braintree] fix search api to be a stream instead of a promise

sharno:
Reference to stream changes of braintree search api:
https://developers.braintreepayments.com/reference/request/subscription/search/node

### DIFF
--- a/types/braintree/braintree-tests.ts
+++ b/types/braintree/braintree-tests.ts
@@ -63,14 +63,14 @@ const gateway: BraintreeGateway = new braintree.BraintreeGateway({
     const response = await gateway.paymentMethod.create(paymentMethodRequest).catch(console.error);
     if (!response) return;
     const { token }: PaymentMethod = response.paymentMethod;
-    const applePayCard = <ApplePayCard>response.paymentMethod;
-    const paypalAccount = <PayPalAccount>response.paymentMethod;
-    const androidPayCard = <AndroidPayCard>response.paymentMethod;
-    const creditCard = <CreditCard>response.paymentMethod;
-    const venmoAccount = <braintree.VenmoAccount>response.paymentMethod;
-    const visaCheckoutCard = <VisaCheckoutCard>response.paymentMethod;
-    const samsungPayCard = <SamsungPayCard>response.paymentMethod;
-    const masterpassCard = <MasterpassCard>response.paymentMethod;
+    const applePayCard = <ApplePayCard> response.paymentMethod;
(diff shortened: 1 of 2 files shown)


types/braintree/index.d.ts
```diff
@@ -16,11 +20,11 @@ declare namespace braintree {
         Development = 'Development',
         Production = 'Production',
         Qa = 'Qa',
-        Sandbox = 'Sandbox'
+        Sandbox = 'Sandbox',
     }
 
     export interface GatewayConfig {
-        environment: any;
+        environment: Environment;
```
smrubin: 👍

types/braintree/braintree-tests.ts
```diff
@@ -23,7 +34,8 @@ const gateway: BraintreeGateway = new braintree.BraintreeGateway({
         customerId: '123456',
         streetAddress: '222 Oak Street',
     };
-    const response = await gateway.address.create(addressRequest);
+    const response = await gateway.address.create(addressRequest).catch(console.error);
+    if (!response) return;
```
smrubin: Yeah, I suppose this is a good idea for other contributors' envs where they might not have a gateway setup.

types/braintree/index.d.ts
```diff
@@ -94,14 +102,14 @@ declare namespace braintree {
     }
 
     interface CreditCardVerificationGateway {
-        search(searchFn: any): Promise<CreditCardVerification[]>;
+        search(searchFn: any): stream.Readable;
```
smrubin: @sharno Thanks for fixing search and adding in streams in. Do you know if there is there a way to add typings to the stream objects (i.e. `CreditCardVerification` here)? That could be very nice to have.

smrubin (approved):
Looks good @sharno. Thanks for your work on `search` and the linting as well.

Had one comment about adding types to the stream objects, but that could be a follow-up to this PR, if it's possible and you're interested in taking on.

types/braintree/index.d.ts
sharno: I tried to look for something like this but I couldn't find any. It's much helpful of course if we could type the stream but it doesn't look the be the case in `@types/node`.

A solution might be looking into if it's possible to make `@types/node`'s `stream.Readable` take a type param.

A custom stream type for braintree is possible but wouldn't work well with the other stream types from `@types/node`

types/braintree/index.d.ts
sharno: Thanks a lot for reviewing and approving this PR

PR #39254, opened 2019-10-19
[bent] Add definitions

Ovyerus:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/bent/index.d.ts
+++ b/types/bent/index.d.ts
@@ -19,10 +19,10 @@ declare function bent(type: 'buffer', ...args: bent.Options[]): bent.RequestFunc
 declare function bent(type: 'json', ...args: bent.Options[]): bent.RequestFunction<bent.Json>;
 
 // Method or url first
-declare function bent(baseUrl: HttpMethod | string, type: 'string', ...args: bent.Options[]): bent.RequestFunction<string>;
-declare function bent(baseUrl: HttpMethod | string, type: 'buffer', ...args: bent.Options[]): bent.RequestFunction<Buffer | ArrayBuffer>;
-declare function bent(baseUrl: HttpMethod | string, type: 'json', ...args: bent.Options[]): bent.RequestFunction<bent.Json>;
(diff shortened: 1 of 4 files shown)


types/bent/index.d.ts
```diff
@@ -0,0 +1,41 @@
...
+interface Headers { [x: string]: any; }
+
+type Json = {[x: string]: any} | any[];
+type ValidResponse = bent.BentResponse | string | Buffer | ArrayBuffer | Json;
+
+declare function bent(...args: bent.Options[]): bent.RequestFunction;
+
+// Can't use these until we get some sort of pattern checking for strings, ala https://github.com/microsoft/TypeScript/issues/6579
+// Otherwise, `string` from BaseUrl just eats up the extra string literals here, meaning it ends up defaulting to the first type.
+// declare function bent(...args: bent.Options[]): bent.RequestFunction<bent.BentResponse>;
```
weswigham: These should work fine. Write this instead:
```ts
declare function bent(type: 'string', ...args: bent.Options[]): bent.RequestFunction<string>;
declare function bent(type: 'buffer', ...args: bent.Options[]): bent.RequestFunction<Buffer | ArrayBuffer>;
declare function bent(type: 'json', ...args: bent.Options[]): bent.RequestFunction<Json>;
declare function bent(...args: bent.Options[]): bent.RequestFunction<bent.BentResponse>;
```

Order matters, and being more specific helps.

weswigham (changes requested):

Ovyerus:
The problem is bent allows the type - or any other argument - at any
position in the parameters list, so it would break for some people's usage.

types/bent/index.d.ts
Ovyerus: The problem is bent allows the type - or any other argument - at any
position in the parameters list, so it would break for some people's usage.

weswigham: I think it's OK to only give the more specific types for certain forms - if anything, it's a little better than what's there right now.

weswigham (approved):

weswigham (approved):

PR #39259, opened 2019-10-19
[stripe] Add support for the Person concept on Stripe Connect Accounts

bruun:
Documentation: https://stripe.com/docs/api/persons?lang=node

The interface overlap a lot with the IIndividual interface, but I have chosen to keep them completely separate to avoid issues going forward if any of them change without the other.

Please fill in this template.

### DIFF
--- a/types/stripe/index.d.ts
+++ b/types/stripe/index.d.ts
@@ -1045,6 +1045,337 @@ declare namespace Stripe {
              */
             ssn_last_4?: string;
         }
+
+        interface IPersonDocumentCreateUpdateOptions {
+            /**
+             * The back of an ID returned by a file upload with a purpose value of identity_document.
+             * The uploaded file needs to be a color image (smaller than 8,000px by 8,000px),
+             * in JPG or PNG format, and less than 10 MB in size.
+             */
+            back?: string;
+
+            /**
+             * The front of an ID returned by a file upload with a purpose value of identity_document.
+             * The uploaded file needs to be a color image (smaller than 8,000px by 8,000px),
+             * in JPG or PNG format, and less than 10 MB in size.
+             */
+            front?: string;
+        }
+
+        interface IPersonDocument extends IPersonDocumentCreateUpdateOptions {
+            /**
+             * A user-displayable string describing the verification state of this document.
+             * For example, if a document is uploaded and the picture is too fuzzy, this may
+             * say “Identity document is too unclear to read”.
+             */
+            details?: string;
+
+            /**
+             * One of document_corrupt, document_country_not_supported, document_expired, document_failed_copy,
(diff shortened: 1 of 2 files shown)


types/stripe/index.d.ts
```diff
@@ -1045,6 +1045,331 @@ declare namespace Stripe {
...
+             */
+            first_name_kana?: string;
+
+            /**
+             * The Kanji variation of the person’s first name (Japan only).
+             */
+            first_name_kanji?: string;
+
+            /**
+             * The person’s gender (International regulations require either “male” or “female”).
```
pushplay: 🤦

types/stripe/index.d.ts
```diff
@@ -1045,6 +1045,331 @@ declare namespace Stripe {
...
+             * The person’s maiden name.
+             */
+            maiden_name?: string;
+
+            /**
+             * Set of key-value pairs that you can attach to an object.
+             * This can be useful for storing additional information about
+             * the object in a structured format.
+             */
+            metadata?: {
```
pushplay: You should pull this out and use `IOptionsMetadata` in create and update and use `IMetadata` in the object.

types/stripe/index.d.ts
```diff
@@ -9421,6 +9746,41 @@ declare namespace Stripe {
                 redirectUrl?: string,
                 response?: IResponseFn<accounts.ILoginLink>,
             ): Promise<accounts.ILoginLink>;
+
```
pushplay: I believe all of these CRUDL functions should have an overload that accepts a `HeaderOptions` param before the `IResponseFn`.

types/stripe/stripe-tests.ts
```diff
@@ -1256,6 +1256,38 @@ stripe.accounts.createLoginLink('acct_17wV8KBoqMA9o2xk', 'http://localhost:3000'
     const created: number = loginLink.created;
     const url: string = loginLink.url;
 });
+
```
pushplay: The account persons tests should probably be another section to make it clear to future contributors.

pushplay (changes requested):
The details look pretty good.  Thanks for the tests.  Just a few small things to take care of.

types/stripe/index.d.ts
bruun: Yeah 🤦‍♂️

types/stripe/index.d.ts
bruun: Good catch, I didn't think to look at the other CRUDL types and only looked at the documentation where this isn't explicitly mentioned on each function. I _think_ it should be correct now

pushplay (approved):

pushplay:
Looks good.  Thanks for the changes.

PR #39275, opened 2019-10-20
Highland - added split() and splitBy()

jhasselgren:
Please fill in this template.

- [ x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [ x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).

- [x ] Provide a URL to documentation or source code which provides context for the suggested changes: https://highlandjs.org/#splitBy
- [ x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

### DIFF
--- a/types/highland/highland-tests.ts
+++ b/types/highland/highland-tests.ts
@@ -273,6 +273,10 @@ barStream = fooStream.scan(bar, (memo: Bar, x: Foo) => {
 
 fooStream = fooStream.stopOnError((e: Error) => {});
 
+fooStream = fooStream.split();
+
+fooStream = fooStream.splitBy(str);
+
 fooStream = fooStream.take(num);
 
 fooStream.tap((x: Foo) => {});
--- a/types/highland/highland-tests.ts
+++ b/types/highland/highland-tests.ts
@@ -19,6 +19,7 @@ var anyArr: any[];
 var strArr: string[];
 var numArr: string[];
 var funcArr: Function[];
+var regExp = /a/
 
 var readable: NodeJS.ReadableStream;
 var readwritable: NodeJS.ReadWriteStream;
@@ -276,6 +277,8 @@ fooStream = fooStream.stopOnError((e: Error) => {});
 fooStream = fooStream.split();
 
 fooStream = fooStream.splitBy(str);
+
+fooStream = fooStream.splitBy(regExp)
 
 fooStream = fooStream.take(num);
 
--- a/types/highland/index.d.ts
+++ b/types/highland/index.d.ts
@@ -911,7 +911,7 @@ declare namespace Highland {
 		 * @param {String|RegExp} sep - String | RegExp - the separator to split on
 		 * @api public
 		 */
-        splitBy(sep: string): Stream<R>;
+		splitBy(sep: string | RegExp): Stream<R>;
 		/**
 		 * Creates a new Stream with the first `n` values from the source.
 		 *
--- a/types/highland/index.d.ts
+++ b/types/highland/index.d.ts
@@ -889,9 +889,9 @@ declare namespace Highland {
 		 * @param {Function} f - the function to handle an error
 		 * @api public
 		 */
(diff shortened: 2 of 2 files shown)


types/highland/index.d.ts
```diff
@@ -888,8 +889,28 @@ declare namespace Highland {
 		 * @param {Function} f - the function to handle an error
 		 * @api public
 		 */
-		stopOnError(f: (err: Error) => void): Stream<R>;
-
```
NotBobTheBuilder: Could you undo this de-dent and have the new methods line up with existing indentation?

types/highland/index.d.ts
```diff
@@ -888,8 +889,28 @@ declare namespace Highland {
...
+    /**
+		 * Splits the source Stream by a separator and emits the pieces in between, much like splitting a string.
+		 *
+		 * @id splitBy
+		 * @section Streams
+		 * @name Stream.splitBy(sep)
+		 * @param {String|RegExp} sep - String | RegExp - the separator to split on
+		 * @api public
+		 */
+    splitBy(sep: string): Stream<R>;
```
NotBobTheBuilder: Tiny nitpick - could you add a blank line between this definition and the next one?

NotBobTheBuilder: Should the type here be string | RegExp?

types/highland/highland-tests.ts
```diff
@@ -273,6 +273,10 @@ barStream = fooStream.scan(bar, (memo: Bar, x: Foo) => {
 
 fooStream = fooStream.stopOnError((e: Error) => {});
 
+fooStream = fooStream.split();
+
+fooStream = fooStream.splitBy(str);
+
```
NotBobTheBuilder: Could we add a test using .splitBy with a regex?

NotBobTheBuilder (changes requested):
Thanks for filing this PR!

types/highland/index.d.ts
jhasselgren: will fix

types/highland/index.d.ts
jhasselgren: fix, looks like the indent are using spaces and sometime tabs...

types/highland/index.d.ts
jhasselgren: fix

types/highland/highland-tests.ts
```diff
@@ -19,6 +19,7 @@ var anyArr: any[];
 var strArr: string[];
 var numArr: string[];
 var funcArr: Function[];
+var regExp = /a/
```
NotBobTheBuilder: It looks like line 17 already defines`exp: RegExp`, maybe that variable can be reused?

types/highland/index.d.ts
NotBobTheBuilder: Oh that’s not nice, well spotted, I’ll keep fixing that in mind for a future PR

types/highland/index.d.ts
```diff
@@ -899,6 +921,7 @@ declare namespace Highland {
 		 * @param {Number} n - integer representing number of values to read from source
 		 * @api public
 		 */
+
```
NotBobTheBuilder: Could you move this new line so as to be on 915? (IE, between the end of the splitBy definition and the beginning of the comment describing the take definition)

NotBobTheBuilder (changes requested):
Looking good, minor pair of nitpicks to keep things tidy, then should be good to merge

PR #39291, opened 2019-10-21
[jest] fix test.each typing when array is readonly

ycmjason:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://jestjs.io/docs/en/api#testeachtable-name-fn-timeout

-----------------------
**Note**

This PR fix issues when trying to do 

```ts
jest.each([['a', 'b', 'c'], ['d', 'e', 'f']] as const)((c1, c2, c3) => {
  // ...
})
```

`c1`, `c2` and `c3` will be typed to `'a' | 'd'`, `'b' | 'e'` and `'c' | 'f'` respectively.

Because of the typescript constraint, I couldn't make the `ReadonlyArray` to a writable array required by `...args`. A compromise is put in place which includes 10 arguments in the callback. (Please advice if I should upgrade the typescript version instead, which might potentially affect many users and other packages depending on jest.)

```ts
...args: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7], T[8], T[9]]
```

If we upgrade the typescript version to `3.4`, we can potentially do

```ts
...args: { -readonly [k in keyof T]: T[k] }
```

### DIFF
--- a/types/jest/index.d.ts
+++ b/types/jest/index.d.ts
@@ -45,29 +45,29 @@ declare var xtest: jest.It;
 declare const expect: jest.Expect;
 
 type ExtractEachCallbackArgs<T extends ReadonlyArray<any>> = {
-  1: [T[0]],
-  2: [T[0], T[1]],
-  3: [T[0], T[1], T[2]],
-  4: [T[0], T[1], T[2], T[3]],
-  5: [T[0], T[1], T[2], T[3], T[4]],
-  6: [T[0], T[1], T[2], T[3], T[4], T[5]],
-  7: [T[0], T[1], T[2], T[3], T[4], T[5], T[6]],
-  8: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7]],
-  9: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7], T[8]],
-  10: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7], T[8], T[9]],
-  'fallback': Array<(T extends ReadonlyArray<infer U>? U: any)>
+    1: [T[0]],
+    2: [T[0], T[1]],
+    3: [T[0], T[1], T[2]],
+    4: [T[0], T[1], T[2], T[3]],
+    5: [T[0], T[1], T[2], T[3], T[4]],
+    6: [T[0], T[1], T[2], T[3], T[4], T[5]],
+    7: [T[0], T[1], T[2], T[3], T[4], T[5], T[6]],
+    8: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7]],
+    9: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7], T[8]],
+    10: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7], T[8], T[9]],
+    'fallback': Array<(T extends ReadonlyArray<infer U>? U: any)>
 }[
-  T extends Readonly<[any]> ? 1
-    : T extends Readonly<[any, any]> ? 2
-    : T extends Readonly<[any, any, any]> ? 3
-    : T extends Readonly<[any, any, any, any]> ? 4
-    : T extends Readonly<[any, any, any, any, any]> ? 5
-    : T extends Readonly<[any, any, any, any, any, any]> ? 6
-    : T extends Readonly<[any, any, any, any, any, any, any]> ? 7
-    : T extends Readonly<[any, any, any, any, any, any, any, any]> ? 8
-    : T extends Readonly<[any, any, any, any, any, any, any, any, any]> ? 9
-    : T extends Readonly<[any, any, any, any, any, any, any, any, any, any]> ? 10
-    : 'fallback'
+    T extends Readonly<[any]> ? 1
+        : T extends Readonly<[any, any]> ? 2
+        : T extends Readonly<[any, any, any]> ? 3
+        : T extends Readonly<[any, any, any, any]> ? 4
+        : T extends Readonly<[any, any, any, any, any]> ? 5
+        : T extends Readonly<[any, any, any, any, any, any]> ? 6
+        : T extends Readonly<[any, any, any, any, any, any, any]> ? 7
+        : T extends Readonly<[any, any, any, any, any, any, any, any]> ? 8
+        : T extends Readonly<[any, any, any, any, any, any, any, any, any]> ? 9
+        : T extends Readonly<[any, any, any, any, any, any, any, any, any, any]> ? 10
+        : 'fallback'
 ];
 
 interface NodeRequire {
--- a/types/jest/index.d.ts
+++ b/types/jest/index.d.ts
@@ -44,6 +44,32 @@ declare var xtest: jest.It;
 
 declare const expect: jest.Expect;
 
+type ExtractEachCallbackArgs<T extends ReadonlyArray<any>> = {
+  1: [T[0]],
+  2: [T[0], T[1]],
+  3: [T[0], T[1], T[2]],
+  4: [T[0], T[1], T[2], T[3]],
+  5: [T[0], T[1], T[2], T[3], T[4]],
+  6: [T[0], T[1], T[2], T[3], T[4], T[5]],
+  7: [T[0], T[1], T[2], T[3], T[4], T[5], T[6]],
+  8: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7]],
+  9: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7], T[8]],
+  10: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7], T[8], T[9]],
+  'fallback': Array<(T extends ReadonlyArray<infer U>? U: any)>
+}[
+  T extends Readonly<[any]> ? 1
+    : T extends Readonly<[any, any]> ? 2
+    : T extends Readonly<[any, any, any]> ? 3
+    : T extends Readonly<[any, any, any, any]> ? 4
+    : T extends Readonly<[any, any, any, any, any]> ? 5
+    : T extends Readonly<[any, any, any, any, any, any]> ? 6
+    : T extends Readonly<[any, any, any, any, any, any, any]> ? 7
+    : T extends Readonly<[any, any, any, any, any, any, any, any]> ? 8
+    : T extends Readonly<[any, any, any, any, any, any, any, any, any]> ? 9
+    : T extends Readonly<[any, any, any, any, any, any, any, any, any, any]> ? 10
+    : 'fallback'
+];
+
 interface NodeRequire {
     /**
      * Returns the actual module instead of a mock, bypassing all checks on
@@ -300,7 +326,7 @@ declare namespace jest {
 
     interface Each {
         // Exclusively arrays.
-        <T extends any[]>(cases: ReadonlyArray<T>): (name: string, fn: (...args: T) => any, timeout?: number) => void;
+        <T extends ReadonlyArray<any>>(cases: ReadonlyArray<T>): (name: string, fn: (...args: ExtractEachCallbackArgs<T>) => any, timeout?: number) => void;
         // Not arrays.
         <T>(cases: ReadonlyArray<T>): (name: string, fn: (...args: T[]) => any, timeout?: number) => void;
         (cases: ReadonlyArray<ReadonlyArray<any>>): (
--- a/types/jest/index.d.ts
+++ b/types/jest/index.d.ts
@@ -326,6 +326,7 @@ declare namespace jest {
 
     interface Each {
         // Exclusively arrays.
+        <T extends any[]>(cases: ReadonlyArray<T>): (name: string, fn: (...args: T) => any, timeout?: number) => void;
         <T extends ReadonlyArray<any>>(cases: ReadonlyArray<T>): (name: string, fn: (...args: ExtractEachCallbackArgs<T>) => any, timeout?: number) => void;
         // Not arrays.
         <T>(cases: ReadonlyArray<T>): (name: string, fn: (...args: T[]) => any, timeout?: number) => void;
--- a/types/jest/index.d.ts
+++ b/types/jest/index.d.ts
@@ -24,6 +24,7 @@
 //                 Alex Bolenok <https://github.com/quassnoi>
 //                 Mario Beltrán Alarcón <https://github.com/Belco90>
 //                 Tony Hallett <https://github.com/tonyhallett>
+//                 Jason Yu <https://github.com/ycmjason>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
--- a/types/jest/jest-tests.ts
+++ b/types/jest/jest-tests.ts
@@ -1499,13 +1499,9 @@ test.each([[1, 1, 2], [1, 2, 3], [2, 1, 3]])(
 );
 
 declare const constCases: [['a', 'b', 'ab'], ['d', 2, 'd2']];
-test.each(constCases)('%s + %s', (s1, s2, expected) => {
-    // $ExpectType "a" | "d"
-    s1;
-    // $ExpectType 2 | "b"
-    s2;
-    // $ExpectType "ab" | "d2"
-    expected;
+test.each(constCases)('%s + %s', (...args) => {
+    // $ExpectType ["a", "b", "ab"] | ["d", 2, "d2"]
+    args;
 });
 
 declare const constCasesWithMoreThanTen: [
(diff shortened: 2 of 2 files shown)


types/jest/index.d.ts
```diff
@@ -299,6 +299,7 @@ declare namespace jest {
 
     interface Each {
         // Exclusively arrays.
+        <T extends ReadonlyArray<any>>(cases: ReadonlyArray<T>): (name: string, fn: (...args: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7], T[8], T[9]]) => any, timeout?: number) => void;
```
ExE-Boss: This should probably be:
```suggestion
        <T extends ReadonlyArray<any>>(cases: ReadonlyArray<T>): (name: string, fn: (...args: T) => any, timeout?: number) => void;
```

ycmjason: I tried. But TS was complaining about spread args can't be read-only array.

ExE-Boss: This will convert it to a read‑write array.

Also, you might want to move this after `T extends any[]`, since `any[]` extends `readonly any[]`.

```suggestion
        <T extends ReadonlyArray<any>>(cases: ReadonlyArray<T>): (name: string, fn: (...args: T extends ReadonlyArray<infer U> ? U[] : any[]) => any, timeout?: number) => void;
```

wsmd: @ExE-Boss I think that will infer the union of all possible types used in the inner arrays (the cases), but not at the actual argument index (which is more accurate).

wsmd: @ExE-Boss I think that will infer the union of all possible types used in the inner arrays (the cases), but not at the actual argument index when declared as a tuple (which is more accurate).

ycmjason: @wsmd I think you are right. 

Any ideas on how should we type this better?

ycmjason: @ExE-Boss I have removed the `<T extends any[]>`.

ExE-Boss: @ycmjason

> I have removed the `<T extends any[]>`.

That&nbsp;only&nbsp;makes the&nbsp;situation&nbsp;worse for&nbsp;read&#x2011;write&nbsp;unions of&nbsp;11 or&nbsp;more&nbsp;elements.

ExE-Boss (changes requested):

ycmjason:
@ExE-Boss @wsmd 

Thank you so much for giving me feedback.

I have updated the implementation so that it fallback to less accurate types when there are more than 10 elements.

Appropriate tests are added to address this.

types/jest/index.d.ts
```diff
@@ -299,7 +325,7 @@ declare namespace jest {
 
     interface Each {
         // Exclusively arrays.
-        <T extends any[]>(cases: ReadonlyArray<T>): (name: string, fn: (...args: T) => any, timeout?: number) => void;
```
ExE-Boss: Please keep this line before the newly inserted line.

ExE-Boss (changes requested):

types/jest/index.d.ts
ycmjason: @ExE-Boss

The new line now covers this case tho?

ExE-Boss: It&nbsp;will cause&nbsp;regressions to&nbsp;read‑write&nbsp;tuples with&nbsp;11&nbsp;or&nbsp;more distinct&nbsp;elements.

ycmjason: I don't understand can you elaborate?

ExE-Boss: This will break:

```ts
const cases: [
	/*  0: */ typeof String,
	/*  1: */ typeof Boolean,
	/*  2: */ typeof Symbol,
	/*  3: */ typeof BigInt,
	/*  4: */ typeof Object,
	/*  5: */ typeof Reflect,
	/*  6: */ typeof Math,
	/*  7: */ typeof Map,
	/*  8: */ typeof Set,
	/*  9: */ typeof WeakMap,
	/* 10: */ typeof WeakSet,
] = [
	/*  0: */ String,
	/*  1: */ Boolean,
	/*  2: */ Symbol,
	/*  3: */ BigInt,
	/*  4: */ Object,
	/*  5: */ Reflect,
	/*  6: */ Math,
	/*  7: */ Map,
	/*  8: */ Set,
	/*  9: */ WeakMap,
	/* 10: */ WeakSet,
];

it.each([cases])('doStuff', (...args) => {
	args; // $ExpectType [StringConstructor, BooleanConstructor, SymbolConstructor, BigIntConstructor, ObjectConstructor, typeof Reflect, Math, MapConstructor, SetConstructor, WeakMapConstructor, WeakSetConstructor]
});
```

ycmjason: @ExE-Boss 
that is how it is currently behaving anyways? my implementation make the `args` to be exact for <=10 arguments.

ycmjason:
```
'fallback': Array<(T extends ReadonlyArray<infer U>? U: any)>
```

already cover what the following does.
```
 <T extends any[]>(cases: ReadonlyArray<T>): (name: string, fn: (...args: T) => any, timeout?: number) => void;
```

ExE-Boss: No, that’s&nbsp;not&nbsp;what `Array<(T extends ReadonlyArray<infer U>? U: any)>` does.

The&nbsp;current type&nbsp;of&nbsp;`args` for&nbsp;very&nbsp;long&nbsp;tuples is&nbsp;the&nbsp;tuple, but&nbsp;for&nbsp;your&nbsp;code, it’s&nbsp;an&nbsp;array&nbsp;whose&nbsp;types is&nbsp;the&nbsp;union of&nbsp;all&nbsp;types present&nbsp;in&nbsp;the&nbsp;tuple.

types/jest/index.d.ts
```diff
@@ -300,7 +326,7 @@ declare namespace jest {
...
+        <T extends ReadonlyArray<any>>(cases: ReadonlyArray<T>): (name: string, fn: (...args: ExtractEachCallbackArgs<T>) => any, timeout?: number) => void;
```
ExE-Boss:
```suggestion
        <T extends any[]>(cases: ReadonlyArray<T>): (name: string, fn: (...args: T) => any, timeout?: number) => void;
        <T extends ReadonlyArray<any>>(cases: ReadonlyArray<T>): (name: string, fn: (...args: ExtractEachCallbackArgs<T>) => any, timeout?: number) => void;
```

types/jest/index.d.ts
```diff
@@ -44,6 +44,32 @@ declare var xtest: jest.It;
...
+    : T extends Readonly<[any, any, any]> ? 3
+    : T extends Readonly<[any, any, any, any]> ? 4
+    : T extends Readonly<[any, any, any, any, any]> ? 5
+    : T extends Readonly<[any, any, any, any, any, any]> ? 6
+    : T extends Readonly<[any, any, any, any, any, any, any]> ? 7
+    : T extends Readonly<[any, any, any, any, any, any, any, any]> ? 8
+    : T extends Readonly<[any, any, any, any, any, any, any, any, any]> ? 9
+    : T extends Readonly<[any, any, any, any, any, any, any, any, any, any]> ? 10
+    : 'fallback'
+];
```
ExE-Boss: This has the wrong indentation, it should be:
```ts
type ExtractEachCallbackArgs<T extends ReadonlyArray<any>> = {
    1: [T[0]],
    2: [T[0], T[1]],
    3: [T[0], T[1], T[2]],
    4: [T[0], T[1], T[2], T[3]],
    5: [T[0], T[1], T[2], T[3], T[4]],
    6: [T[0], T[1], T[2], T[3], T[4], T[5]],
    7: [T[0], T[1], T[2], T[3], T[4], T[5], T[6]],
    8: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7]],
    9: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7], T[8]],
    10: [T[0], T[1], T[2], T[3], T[4], T[5], T[6], T[7], T[8], T[9]],
    'fallback': Array<(T extends ReadonlyArray<infer U>? U: any)>
}[
    T extends Readonly<[any]> ? 1
        : T extends Readonly<[any, any]> ? 2
        : T extends Readonly<[any, any, any]> ? 3
        : T extends Readonly<[any, any, any, any]> ? 4
        : T extends Readonly<[any, any, any, any, any]> ? 5
        : T extends Readonly<[any, any, any, any, any, any]> ? 6
        : T extends Readonly<[any, any, any, any, any, any, any]> ? 7
        : T extends Readonly<[any, any, any, any, any, any, any, any]> ? 8
        : T extends Readonly<[any, any, any, any, any, any, any, any, any]> ? 9
        : T extends Readonly<[any, any, any, any, any, any, any, any, any, any]> ? 10
        : 'fallback'
];
```

types/jest/jest-tests.ts
```diff
@@ -1498,6 +1498,32 @@ test.each([[1, 1, 2], [1, 2, 3], [2, 1, 3]])(
...
+    expect(s1 + s2).toBe(expected);
+});
+
+const constCasesWithMoreThanTen: [
+  [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11],
+  [91, 92, 93, 94, 95, 96, 97, 98, 99, 910, 911]
+] = [
+  [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11],
+  [91, 92, 93, 94, 95, 96, 97, 98, 99, 910, 911]
+];
```
ExE-Boss: Same here:
```ts
declare const constCasesWithMoreThanTen: [
    [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11],
    [91, 92, 93, 94, 95, 96, 97, 98, 99, 910, 911]
];
```

types/jest/jest-tests.ts
```diff
@@ -1498,6 +1498,32 @@ test.each([[1, 1, 2], [1, 2, 3], [2, 1, 3]])(
...
+test.each(constCasesWithMoreThanTen)('should fall back with more than 10 args', (n1, n2, n3, n4, n5, n6, n7, n8, n9, n10, n11) => {
```
ExE-Boss: This&nbsp;should test&nbsp;that the&nbsp;`args`&nbsp;array is&nbsp;still&nbsp;correctly a&nbsp;tuple, even&nbsp;for&nbsp;overly long&nbsp;tuples.

```suggestion
test.each(constCasesWithMoreThanTen)('should fall back with more than 10 args', (...args) => {
    args; // $ExpectType [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11] | [91, 92, 93, 94, 95, 96, 97, 98, 99, 910, 911]
    const [n1, n2, n3, n4, n5, n6, n7, n8, n9, n10, n11] = args;
```

types/jest/jest-tests.ts
```diff
@@ -1498,6 +1498,32 @@ test.each([[1, 1, 2], [1, 2, 3], [2, 1, 3]])(
     5000
 );
 
+const constCases: [['a', 'b', 'ab'], ['d', 2, 'd2']] = [['a', 'b', 'ab'], ['d', 2, 'd2']];
```
ExE-Boss: You can just do `declare const`, since these are compilation tests, so there’s no need for code duplication:

```suggestion
declare const constCases: [['a', 'b', 'ab'], ['d', 2, 'd2']];
```

types/jest/jest-tests.ts
```diff
@@ -1498,6 +1498,32 @@ test.each([[1, 1, 2], [1, 2, 3], [2, 1, 3]])(
...
+test.each(constCases)('%s + %s', (s1, s2, expected) => {
```
ExE-Boss: This&nbsp;should test&nbsp;that the&nbsp;`args`&nbsp;array is&nbsp;correctly a&nbsp;tuple.

```suggestion
test.each(constCases)('%s + %s', (...args) => {
    args; // $ExpectType ["a", "b", "ab"] | ["d", 2, "d2"]
    const [s1, s2, expected] = args;
```

ExE-Boss (changes requested):

types/jest/index.d.ts
ycmjason: @ExE-Boss

Thank you for the thorough review. But I think this line will actually infer `args` to be array of union of types.

See:
http://www.typescriptlang.org/play/#code/GYVwdgxgLglg9mABAUwIYQBYB4AqKAeUyYAJgM6KpgCeA2gLoB8AFBKmcmQFyIBKaJBABtqAQQBO41NVyMAlD2bAwigHTrU4gObdEOOYgC8jSjQPHEANzgwSiAN4AoRC8TjkUEOKRKw5k07OrgC+jqGOaJjMtLQAjAA0iABMiQDkqKn09HLMzPiJ1IkAXv4OQS5hckA

ExE-Boss: In&nbsp;your&nbsp;example, you’re&nbsp;actually&nbsp;supplying `(string | number)[]`, but&nbsp;in&nbsp;my&nbsp;case, it&nbsp;will&nbsp;be&nbsp;inferred&nbsp;correctly:&nbsp;[🔗](https://www.typescriptlang.org/play/?ssl=10&ssc=16&pln=10&pc=17#code/GYVwdgxgLglg9mABAUwIYQBYB4AqKAeUyYAJgM6KpgCeA2gLoB8AFBKmcmQFyIBKaJBABtqAQQBO41NVyMAlD2bAwigHTrU4gObdEOOYgC8jSjQPHEANzgwSiAN4AoRC8TjkUEOKRKw5k07OrgC+jqGOaJjMtLQAjAA0iABMiQDkqKn0lBRxiSmI6Zn0cszM+InUiQBe-g5BLhAIZFCI+Ek8sUatANz1iI1gzYjU7cld1L2u-U0tVaOFXVW9wXJAA)

types/jest/jest-tests.ts
```diff
@@ -1498,30 +1498,45 @@ test.each([[1, 1, 2], [1, 2, 3], [2, 1, 3]])(
     5000
 );
 
-const constCases: [['a', 'b', 'ab'], ['d', 2, 'd2']] = [['a', 'b', 'ab'], ['d', 2, 'd2']];
+declare const constCases: [['a', 'b', 'ab'], ['d', 2, 'd2']];
 test.each(constCases)('%s + %s', (s1, s2, expected) => {
-    expect(s1 + s2).toBe(expected);
+    // $ExpectType 'a' | 'd'
```
ExE-Boss:
```suggestion
    // $ExpectType "a" | "d"
```

types/jest/jest-tests.ts
```diff
@@ -1498,30 +1498,45 @@ test.each([[1, 1, 2], [1, 2, 3], [2, 1, 3]])(
...
+    s1;
+    // $ExpectType 'b' | 2
```
ExE-Boss:
```suggestion
    // $ExpectType "b" | 2
```

types/jest/jest-tests.ts
```diff
@@ -1498,30 +1498,45 @@ test.each([[1, 1, 2], [1, 2, 3], [2, 1, 3]])(
...
+    s2;
+    // $ExpectType 'ab' | 'd2'
```
ExE-Boss:
```suggestion
    // $ExpectType "ab" | "d2"
```

types/jest/jest-tests.ts
ycmjason: tried putting in this but seems like union types are really flaky.
https://github.com/microsoft/dtslint/issues/191

types/jest/jest-tests.ts
```diff
@@ -1498,30 +1498,45 @@ test.each([[1, 1, 2], [1, 2, 3], [2, 1, 3]])(
...
-  expect(n4).toBeDefined();
-  expect(n5).toBeDefined();
-  expect(n6).toBeDefined();
-  expect(n7).toBeDefined();
-  expect(n8).toBeDefined();
-  expect(n9).toBeDefined();
-  expect(n10).toBeDefined();
-  expect(n11).toBeDefined();
+
+test.each(constCasesWithMoreThanTen)('should fall back with more than 10 args', (...args) => {
```
ExE-Boss:
```suggestion
test.each(constCasesWithMoreThanTen)('should fall back with more than 10 args', (...args) => {
    args; // $ExpectType [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11] | [91, 92, 93, 94, 95, 96, 97, 98, 99, 910, 911]
```

ExE-Boss (changes requested):
Yeah, you have to use double quotes in `$ExpectType`, because that’s what types are stringified to.

Also, please apply https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39291#discussion_r339665011

types/jest/index.d.ts
ycmjason: ahhhh. make sense now.

Sorry for being stubborn on this and will put a fix in now.

ycmjason: @ExE-Boss 

Can you think of a better way to do the fallback?

types/jest/jest-tests.ts
ycmjason: Something really happen when I try to do that.
image
No matter what i put, it fails.

ycmjason:
@ExE-Boss

I tried making the test pass, but the type it evaluate is different everytime i change the assertion.

I cannot make it pass, maybe you will have better idea?

ycmjason:
@ExE-Boss

shall we just skip the test XD

ycmjason:
@ExE-Boss @wsmd 

I have pushed a commit to skip the assertion. I left a comment on why I skip it in case dtslint ever fixes it.

ExE-Boss (approved):

PR #39303, opened 2019-10-21
Add logger function to video.js Log interface

ckybonist:

If adding a new definition:

### Description
Add `logger` function in `interface Log {}`.


joeflateau (changes requested):
Please update this so that it only has changes related adding the logger function and not the formatting changes. You may have to turn off your code formatter

ckybonist:
> Please update this so that it only has changes related adding the logger function and not the formatting changes. You may have to turn off your code formatter

@joeflateau Ok, I'll fix it. Just follow the contribution guideline in README though

joeflateau (changes requested):
Thank you, I do understand the guidelines, but it'd be better to reformat the code in a PR for that specifically rather than muddying this PR with unrelated changes.

createLogger was added in videojs 7.3 so please also bump the version up from 7.2 at the top of this file. and also add a test for createLogger, also remove the "Definitions by: [name]" from the jsdoc comment near createLogger, you should instead add your name to the top of the file w/ the rest of us.

ckybonist:
Added a test function, but I don't know how to verify the log outputs, only pass `npm run test`...

joeflateau:
@ckybonist you're doing that right. the test file doesn't actually get run, it's just used as syntax that should work w/ the typings (so there's no actual phase of verifying the log output). But, please revert the whitespace/formatting changes in the test file, the linter doesn't like the spaces. See the output here: https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/601907814

ckybonist:
> @ckybonist you're doing that right. the test file doesn't actually get run, it's just used as syntax that should work w/ the typings (so there's no actual phase of verifying the log output). But, please revert the whitespace/formatting changes in the test file, the linter doesn't like the spaces. See the output here: https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/601907814

Fixed weird whitespace, please check

joeflateau (approved):

PR #39164, opened 2019-10-16
remove dotenv

maxbeatty:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If removing a declaration:
- [x] Delete the package's directory.
- [x] Add it to `notNeededPackages.json`.

PR #39308, opened 2019-10-21
[Google.Visualization] Add Missing Properties to LineChartOptions

dckorben:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://developers.google.com/chart/interactive/docs/gallery/linechart#overview

General Notes:
Added missing "data" class which carries manipulations grouping and joining DataTable/DataView.
Fixed several inconsistencies with the documentation around DataTable.
Relaxed several return types which are inaccurately restrictive per documentation.

### DIFF
--- a/types/google.visualization/index.d.ts
+++ b/types/google.visualization/index.d.ts
@@ -645,6 +645,19 @@ declare namespace google {
             draw(data: DataTable | DataView, options: LineChartOptions): void;
         }
 
+        // https://developers.google.com/chart/interactive/docs/gallery/trendlines
+        export interface LineChartTrendlineOptions {
+            type?: 'linear' | 'exponential' | 'polynomial';
+            degree?: number;
+            color?: string;
+            lineWidth?: number;
+            opacity?: number;
+            pointsVisible?: boolean;
+            labelInLegend?: string;
+            visibleInLegend?: boolean;
+            showR2?: boolean
+        }
+        
         // https://developers.google.com/chart/interactive/docs/gallery/linechart#Configuration_Options
         export interface LineChartOptions {
             aggregationTarget?: string;
@@ -674,7 +687,7 @@ declare namespace google {
             selectionMode?: string // single / multiple
             series?: any;
             domainAxis?: { type: string };
-            trendlines?: { [key: number]: any; };
+            trendlines?: { [key: number]: LineChartTrendlineOptions; };
             pointShape?: string | 'circle' | 'triangle' | 'square' | 'diamond' | 'star' | 'polygon';
             intervals?: { style: string };
             interval?: any;
--- a/types/google.visualization/index.d.ts
+++ b/types/google.visualization/index.d.ts
@@ -661,6 +661,7 @@ declare namespace google {
             color?: string;
             lineWidth?: number;
             opacity?: number;
+            pointSize?: number;
             pointsVisible?: boolean;
             labelInLegend?: string;
             visibleInLegend?: boolean;
--- a/types/google.visualization/index.d.ts
+++ b/types/google.visualization/index.d.ts
@@ -674,7 +674,7 @@ declare namespace google {
             series?: any;
             domainAxis?: { type: string };
             trendlines?: any;
-            pointShape?: any;
+            pointShape?: string;
             intervals?: { style: string };
             interval?: any;
             theme?: string;
--- a/types/google.visualization/index.d.ts
+++ b/types/google.visualization/index.d.ts
@@ -84,7 +84,7 @@ declare namespace google {
             static count(values: number[]): number;
             
             // https://developers.google.com/chart/interactive/docs/reference#join
-            static join(dataA: DataTable | DataView, dataB: DataTable | DataView, joinMethod: string, keys: number[][], columnsA: number[], columnsB: number[]): DataTable;
+            static join(dataA: DataTable | DataView, dataB: DataTable | DataView, joinMethod: 'full' | 'inner' | 'left' | 'right', keys: number[][], columnsA: number[], columnsB: number[]): DataTable;
         }
         //#endregion
         
--- a/types/google.visualization/index.d.ts
+++ b/types/google.visualization/index.d.ts
@@ -707,7 +707,7 @@ declare namespace google {
             selectionMode?: string // single / multiple
             series?: any;
             domainAxis?: { type: string };
-            trendlines?: { [key: number]: LineChartTrendlineOptions; };
+            trendlines?: Record<number, LineChartTrendlineOptions>;
             pointShape?: string | 'circle' | 'triangle' | 'square' | 'diamond' | 'star' | 'polygon';
             intervals?: { style: string };
             interval?: any;
--- a/types/google.visualization/index.d.ts
+++ b/types/google.visualization/index.d.ts
@@ -73,6 +73,14 @@ declare namespace google {
         
         //#region data
         // https://developers.google.com/chart/interactive/docs/reference#google_visualization_data_group
+        export interface GroupKeyOptions {
+            column: number;
+            type: string;
+            modifier?: (value: any) => any;
+            label?: string;
+            id?: any;
+        }
+
         export interface GroupColumnOptions {
             column: number;
             aggregation: (values: any[]) => any;
@@ -83,8 +91,7 @@ declare namespace google {
         
         export class data {
             // https://developers.google.com/chart/interactive/docs/reference#group
-            static group(data: DataTable, keys: any[], columns?: GroupColumnOptions[]): DataTable;
-            static group(data: DataView, keys: any[], columns?: GroupColumnOptions[]): DataTable;
+            static group(data: DataTable | DataView, keys: (number | GroupKeyOptions)[], columns?: GroupColumnOptions[]): DataTable;
             
             static sum(values: number[] | string[] | Date[]): number;
             static avg(values: number[] | string[] | Date[]): number;
--- a/types/google.visualization/index.d.ts
+++ b/types/google.visualization/index.d.ts
@@ -72,10 +72,19 @@ declare namespace google {
         //#endregion
         
         //#region data
+        // https://developers.google.com/chart/interactive/docs/reference#google_visualization_data_group
+        export interface GroupColumnOptions {
+            column: number;
+            aggregation: (values: any[]) => any;
+            type: string;
+            label?: string;
+            id?: any;
+        }
+        
         export class data {
             // https://developers.google.com/chart/interactive/docs/reference#group
-            static group(data: DataTable, keys: any[], columns: any[]): DataTable;
-            static group(data: DataView, keys: any[], columns: any[]): DataTable;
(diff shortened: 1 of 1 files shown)


types/google.visualization/index.d.ts
```diff
@@ -654,6 +672,11 @@ declare namespace google {
             reverseCategories?: boolean;
             selectionMode?: string // single / multiple
             series?: any;
+            domainAxis?: { type: string };
+            trendlines?: any;
+            pointShape?: any;
```
shrujalshah28: According to [docs](https://developers.google.com/chart/interactive/docs/points), `pointShape` accepts `string` value like `circle` and `triangle` etc.

types/google.visualization/index.d.ts
```diff
@@ -654,6 +672,11 @@ declare namespace google {
             reverseCategories?: boolean;
             selectionMode?: string // single / multiple
             series?: any;
+            domainAxis?: { type: string };
+            trendlines?: any;
```
shrujalshah28: `trendlines` is type of `object` not `any`.

shrujalshah28 (changes requested):

types/google.visualization/index.d.ts
```diff
@@ -261,7 +279,7 @@ declare namespace google {
             getTableColumnIndex(viewColumnIndex: number): number;
             getTableRowIndex(viewRowIndex: number): number;
             getViewColumnIndex(tableColumnIndex: number): number;
-            getViewColumns(): number[];
+            getViewColumns(): any[];
```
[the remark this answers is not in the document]
dckorben: This is actually incorrect in the documentation. If you create a derived/calculated column in the DataView, it will return it, which has an object structure. Since the union of number[] or a new object doesn't improve typechecking, this seemed the appropriate correction.

types/google.visualization/index.d.ts
dckorben: Agreed, corrected

types/google.visualization/index.d.ts
dckorben: Changed to number keyed list of anys.

dckorben:
I don't see any way to move this back to needing review.

types/google.visualization/index.d.ts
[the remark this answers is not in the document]
IvanBisultanov: Seems better to specify types, IDE will show developers notes.
Could you please add types that specified in documentation?
https://developers.google.com/chart/interactive/docs/reference#DataView
`columnIndexes - An array of numbers and/or objects (can be mixed)` section

IvanBisultanov (changes requested):

types/google.visualization/index.d.ts
```diff
@@ -67,10 +69,26 @@ declare namespace google {
...
             setView(view_spec: string): void;
         }
-
         //#endregion
+        
+        //#region data
+        export class data {
+            // https://developers.google.com/chart/interactive/docs/reference#group
+            static group(data: DataTable, keys: any[], columns: any[]): DataTable;
+            static group(data: DataView, keys: any[], columns: any[]): DataTable;
```
shrujalshah28: `columns` is an optional argument.

types/google.visualization/index.d.ts
```diff
@@ -67,10 +69,26 @@ declare namespace google {
...
+            
+            static sum(values: number[]): number;
+            static avg(values: number[]): number;
+            static min(values: number[]): number;
+            static max(values: number[]): number;
```
shrujalshah28: `values` can have multiple types.

types/google.visualization/index.d.ts
```diff
@@ -67,10 +69,26 @@ declare namespace google {
...
+            static count(values: number[]): number;
+            
+            // https://developers.google.com/chart/interactive/docs/reference#join
+            static join(dataA: DataTable | DataView, dataB: DataTable | DataView, joinMethod: string, keys: number[][], columnsA: number[], columnsB: number[]): DataTable;
```
shrujalshah28:
```js
joinMethod: 'full' | 'inner' | 'left' | 'right'
```

shrujalshah28 (changes requested):

types/google.visualization/index.d.ts
```diff
@@ -654,6 +672,11 @@ declare namespace google {
             reverseCategories?: boolean;
             selectionMode?: string // single / multiple
             series?: any;
+            domainAxis?: { type: string };
+            trendlines?: { [key: number]: any; };
```
shrujalshah28: Please create an interface for all propertied that `trendlines` can have.

shrujalshah28 (changes requested):

types/google.visualization/index.d.ts
dckorben: amended

types/google.visualization/index.d.ts
dckorben: These would be called directly by the group() function. An end user invoking them directly would be unlikely. However, being able to know they exist on this class is beneficial. Values generically are allowed to be of any[] because you can build your own custom aggregators, but given these built in aggregators don't make sense taking anything but numbers I think the choice is reasonable.

shrujalshah28: Yes, but from this [table](https://developers.google.com/chart/interactive/docs/reference#provided-aggregation-functions), We can define that `values: number[] | string[] | date[]`

shrujalshah28: And also return type for `min` and `max` can be ` number | string | date`

types/google.visualization/index.d.ts
```diff
@@ -67,10 +69,26 @@ declare namespace google {
...
+        export class data {
+            // https://developers.google.com/chart/interactive/docs/reference#group
+            static group(data: DataTable, keys: any[], columns: any[]): DataTable;
+            static group(data: DataView, keys: any[], columns: any[]): DataTable;
+            
+            static sum(values: number[]): number;
+            static avg(values: number[]): number;
+            static min(values: number[]): number;
+            static max(values: number[]): number;
+            static count(values: number[]): number;
```
shrujalshah28: `values` here is type of `any`

shrujalshah28 (changes requested):

types/google.visualization/index.d.ts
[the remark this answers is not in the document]
dckorben: By providing both possibilities, it made sense to switch any to ColumnSpec

types/google.visualization/index.d.ts
dckorben: Ok, based on usage in the rest of the class it seemed like unions were being actively avoided. So, I'll correct.

types/google.visualization/index.d.ts
dckorben: changed in the same commit as the other aggregates

types/google.visualization/index.d.ts
dckorben: Added the interface here as well.

types/google.visualization/index.d.ts
dckorben: updated

types/google.visualization/index.d.ts
```diff
@@ -626,6 +645,19 @@ declare namespace google {
             draw(data: DataTable | DataView, options: LineChartOptions): void;
         }
 
+        // https://developers.google.com/chart/interactive/docs/gallery/trendlines
+        export interface LineChartTrendlineOptions {
```
shrujalshah28: `pointSize` is missing

shrujalshah28 (changes requested):

types/google.visualization/index.d.ts
```diff
@@ -72,10 +72,19 @@ declare namespace google {
...
+            type: string;
+            label?: string;
+            id?: any;
+        }
+        
         export class data {
             // https://developers.google.com/chart/interactive/docs/reference#group
-            static group(data: DataTable, keys: any[], columns: any[]): DataTable;
-            static group(data: DataView, keys: any[], columns: any[]): DataTable;
+            static group(data: DataTable, keys: any[], columns?: GroupColumnOptions[]): DataTable;
```
shrujalshah28: Please create a separate interface for `keys` also.
And use that like this
`(number|newTyle)[]`

shrujalshah28 (changes requested):

types/google.visualization/index.d.ts
dckorben: Missed that one. Thanks.

types/google.visualization/index.d.ts
dckorben: Done, added modifier func and re-ordered the class to make more sense with the document citations

types/google.visualization/index.d.ts
```diff
@@ -654,6 +703,11 @@ declare namespace google {
             reverseCategories?: boolean;
             selectionMode?: string // single / multiple
             series?: any;
+            domainAxis?: { type: string };
+            trendlines?: { [key: number]: LineChartTrendlineOptions; };
```
shrujalshah28: Try if it works:
`trendlines?: Record<number, LineChartTrendlineOptions>;

shrujalshah28 (approved):
LGTM

IvanBisultanov (approved):

types/google.visualization/index.d.ts
dckorben: Yea, that works, one moment on the commit.

shrujalshah28 (approved):

shrujalshah28:
Hi please refer this [commit](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39164/commits/5597f9f8ad24d9dd9d5430857e247fe069f6be24) if you need help to add header.

dckorben:
Looks like it might want ts 2.9. One moment

dckorben:
Header is conflicting with dependent package 'dygraphs'

shrujalshah28:
Try 2.9 and check. If it still fill that. We need to remove `Record`.

dckorben:
With package 'dygraphs' in there I don't think we can spec a ts version at all anyway.

shrujalshah28:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5344bfc80508c53a23dae37b860fb0c905ff7b24/types/dygraphs/index.d.ts#L6

dckorben:
As this is my first PR ever, I'm gonna need some advice.

shrujalshah28:
As per my knowledge you have 2 way.
1. You raise another PR in that package that also force TS 2.9. But that PR can take time to merge.
2. You remove `Record` from `trendlines` and go back to old syntax.

dckorben:
Ok, so keyed number list is appropriate then?

types/google.visualization/index.d.ts
dckorben: Reverted

dckorben:
I don't see any other revisions requested or needed. I do have a small list for a follow-up PR for more in-general fixes to the @type.

PR #39311, opened 2019-10-21
[@types/stoppable] Generic Server

qwelias:
There's more that just `http.Server`, `net.Server` is the base **Server**, these changes allow you to use the lib with any sort of server, which extends `net.Server`.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/api/http.html#http_class_http_server and https://nodejs.org/api/net.html#net_class_net_server

### DIFF
--- a/types/stoppable/index.d.ts
+++ b/types/stoppable/index.d.ts
@@ -2,17 +2,19 @@
 // Project: https://github.com/hunterloftis/stoppable
 // Definitions by: Eric Byers <https://github.com/EricByers>
 //                 John Plusjé <https://github.com/jplusje>
+//                 Ilia Baryshnikov <https://github.com/qwelias>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 /// <reference types="node" />
-import { Server } from 'http';
+import { Server as NetServer } from 'net';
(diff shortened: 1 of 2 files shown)


qwelias:
@EricByers 
@jplusje

types/stoppable/index.d.ts
```diff
@@ -2,17 +2,20 @@
...
+import { Server as NetServer } from 'net';
+import { Server as HttpServer } from 'http';
+
+declare function stoppable <S extends NetServer>(server: S, grace?: number): stoppable.StoppableServer<S>;
 
 declare namespace stoppable {
-  interface StoppableServer extends Server {
-    stop(callback?: (e: Error, gracefully: boolean) => void): void;
-  }
+    type StoppableServer<S = HttpServer> = S & {
```
jessetrinity: Does this need to be a type alias rather than an interface? Interfaces generally give better error messages.

jessetrinity (changes requested):

types/stoppable/index.d.ts
qwelias: @jessetrinity How can I make it a generic interface that extends `S` and also keep it backward compatible?

PR #39320, opened 2019-10-22
Updated passport-google-oauth2 to include missing function signature

mikefrancis:
There is a missing type for a function signature which is not documented in the current typings:

https://github.com/jaredhanson/passport-oauth2/blob/master/lib/strategy.js#L193

This PR adds in the missing type so that the `params` is available in the callback function as mentioned in https://github.com/jaredhanson/passport-google-oauth2/issues/28

Please fill in this template.

### DIFF
--- a/types/passport-google-oauth2/index.d.ts
+++ b/types/passport-google-oauth2/index.d.ts
@@ -1,6 +1,7 @@
 // Type definitions for passport-google-oauth2 0.1
 // Project: https://github.com/mstade/passport-google-oauth2
 // Definitions by: Elliot Blackburn <https://github.com/bluehatbrit>
+//                 Mike Francis <https://github.com/mikefrancis>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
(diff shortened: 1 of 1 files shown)


types/passport-google-oauth2/index.d.ts
```diff
@@ -43,7 +59,7 @@ export class Strategy implements Strategy {
   name: string;
   authenticate: (req: Request, options?: object) => void;
 
-  constructor(options: StrategyOptionsWithRequest, verify: VerifyFunctionWithRequest);
+  constructor(options: StrategyOptionsWithRequest, verify: VerifyFunctionWithRequest | VerifyFunctionWithRequestAndParams);
```
BlueHatbRit: I think it might be a better idea to introduce a new constructor for this so users don't need to add type guards using these signatures. What do you think?

elliotblackburn (commented):
Ooh this looks good, nice spot! I just have one question about the constructor update.

types/passport-google-oauth2/index.d.ts
mikefrancis: I did do this originally but the linter failed told me to do it like this

BlueHatbRit: Oh so I see. I've gone back and had a look at the linter rule, I'm not entirely sure it's a great rule but that's not up to me! Well with that in mind I'll approve this 😄

elliotblackburn (approved):

PR #39321, opened 2019-10-22
[react-input-mask] exporting interface which was was not exported before

shytikov:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://powerapps.microsoft.com/en-us/blog/use-of-react-and-office-ui-fabric-react-in-the-powerapps-component-framework-is-now-available-for-public-preview/ to use `react-input-mask` component I it's required to create component manually, like following:
```
ReactDOM.render(
  React.createElement(
    ReactInputMask,
    this.props
  ),
  this.theContainer
);
``` 
Thus `this.props` needs to be typed object. But since respective interface was not exported, correctly create such object was impossible. Only option would be create a copy of `Props` interface. Exporting original `Props` interface fixed that issue.

### DIFF
--- a/types/react-input-mask/index.d.ts
+++ b/types/react-input-mask/index.d.ts
@@ -25,7 +25,7 @@ export interface MaskOptions {
   permanents: number[];
 }
 
-interface Props extends React.InputHTMLAttributes<HTMLInputElement> {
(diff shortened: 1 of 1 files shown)


shytikov:
Ping @apare, @dima7a14.

dima7a14 (approved):

PR #39334, opened 2019-10-22
Type declarations for eccrypt package

randomnerd:
Please fill in this template.

If adding a new definition:

If removing a declaration:
- [X] If a package was never on DefinitelyTyped, you don't need to do anything. (If you wrote a package and provided types, you don't need to register it with us.)
- [X] Delete the package's directory.
- [X] Add it to `notNeededPackages.json`.

### DIFF
--- a/types/eccrypto/eccrypto-tests.ts
+++ b/types/eccrypto/eccrypto-tests.ts
@@ -1,0 +1,10 @@
+import eccrypto = require("eccrypto");
+import expect = require("expect");
+const privKey = eccrypto.generatePrivate();
+const pubKey = eccrypto.getPublicCompressed(privKey);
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -22,6 +22,7 @@
         "lint": "dtslint types"
     },
     "devDependencies": {
+        "@types/node": "^12.11.2",
```
weswigham: Revert this change to the root `package.json` and just rely on the `reference types` tooling to do the right thing, thanks~ ❤️

weswigham (changes requested):

package.json
randomnerd: Done!

weswigham (approved):

PR #39336, opened 2019-10-22
Fix electron dependency in electron-clipboard-extended to 6

sandersn:
Electron 7's Clipboard doesn't have the `on` method anymore.

### DIFF
--- a/types/electron-clipboard-extended/package.json
+++ b/types/electron-clipboard-extended/package.json
@@ -1,6 +1,6 @@
 {
     "private": true,
     "dependencies": {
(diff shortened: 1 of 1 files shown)


sandersn:
@Jungwoo-An I'm not sure if upgrading to Electron 7 is difficult for these types, so I just fixed the dependency to 6 for now. Let me know if you'd rather upgrade it.

Jungwoo-An:
@sandersn Thanks! i have no plan to upgrade it 👍

Jungwoo-An (approved):

PR #39339, opened 2019-10-23
Add Before and After for cypress-cucumber-preprocessor

FalconerTC:
cypress-cucumber-preprocessor added support for Cypress hooks Before and After in 1.14

### DIFF
--- a/types/cypress-cucumber-preprocessor/steps/index.d.ts
+++ b/types/cypress-cucumber-preprocessor/steps/index.d.ts
@@ -12,3 +12,5 @@ export function When(expression: RegExp | string, implementation: (...args: any[
 export function Then(expression: RegExp | string, implementation: (...args: any[]) => void): void;
 export function And(expression: RegExp | string, implementation: (...args: any[]) => void): void;
 export function But(expression: RegExp | string, implementation: (...args: any[]) => void): void;
+export function Before(args_0: object | ((...args: any[]) => void), args_1?: (...args: any[]) => void): void;
+export function After(args_0: object | ((...args: any[]) => void), args_1?: (...args: any[]) => void): void;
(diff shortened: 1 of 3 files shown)


types/cypress-cucumber-preprocessor/steps/index.d.ts
```diff
@@ -12,3 +12,5 @@ export function When(expression: RegExp | string, implementation: (...args: any[
 export function Then(expression: RegExp | string, implementation: (...args: any[]) => void): void;
 export function And(expression: RegExp | string, implementation: (...args: any[]) => void): void;
 export function But(expression: RegExp | string, implementation: (...args: any[]) => void): void;
+export function Before(args_0: object | ((...args: any[]) => void), args_1?: (...args: any[]) => void): void;
```
aleccool213: Can you link to where this code is defined in the original repo with a reply to this comment? I have an idea why `args_0` is defined this way but don't understand `args_1`.

FalconerTC: It's use is documented [here](https://github.com/cucumber/cucumber-js/blob/master/docs/support_files/hooks.md). I'm new to Typescript so let me know if this is not the correct bindings interpretation.

FalconerTC: The function that parses this in cypress-cucumber-preprocessor is [here](https://github.com/TheBrainFamily/cypress-cucumber-preprocessor/blob/master/lib/resolveStepDefinition.js#L153)

aleccool213: Thanks. This looks correct!

I dislike the Cucumber api for `Before/After` due to is optional first argument of `options`. This made it difficult to know if your code was correct or not, not your fault. Let's name the arguments to something more specific.

```typescript
export function Before(optionsOrImplementation: object | ((...args: any[]) => void), implementation?: (...args: any[]) => void): void;
```

aleccool213 (changes requested):

aleccool213 (approved):

PR #39354, opened 2019-10-23
mui-datatables@2.12 Missing fields…

KrzysztofMadejski:
Changes:
- add `filterType` as column option using FilterType enum [documented in README](https://github.com/gregnb/mui-datatables)
- add missing field `columns` on `TableState` (I'm using that to detect which sorting was enabled instead of parsing announceText which seemed to me like a hack)
- added `"none"` as a possible value of `sortDirection`. Warning: I don't believe this should be a valid value as `undefined` is possible as well, but I was receiving `none` so wanted to account that. Is that a bug that should be fixed?

CC @gabrielliwerant @MasterKale could you review it? It might be also good to pack definitions along mui-datatables to keep it up to date.

Checks:

Check If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/gregnb/mui-datatables (README)
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header. (it updates the given version)

### DIFF
--- a/types/mui-datatables/index.d.ts
+++ b/types/mui-datatables/index.d.ts
@@ -28,9 +28,15 @@ interface MUIDataTableStateRows {
 export interface MUIDataTableState {
     activeColumn: string | null;
     announceText: string | null;
+    columns: MUIDataTableColumnState[];
+    // TODO count is also returned
+    // TODO data
+    // TODO displayData
     expandedRows: MUIDataTableStateRows;
+    // TODO filterData
     filterList: string[][];
     page: number;
+    // TODO previousSelectedRow
     rowsPerPage: number;
     rowsPerPageOptions: number[];
     searchText: string | null;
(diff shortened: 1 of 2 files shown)


favna (changes requested):
When adding TODO's it would be preferred if you also actually add them.

Favna:
> It might be also good to pack definitions along mui-datatables to keep it up to date.

That's a good option. At the time I initiated typings for this package they were only a gist and DefinitetlyTyped is the fastest way to get typings for a library when unsure if the author of the library will merge them in and update the package accordingly, ergo I published them here.

Regardless, if @gregnb's is open to include the typings directly in the library and also maintain them then by all means he can do so by taking this file as base and when he adds them they can be removed from DefinitetlyTyped. Maintain is the important part here - making releases of the library without also updating the typings accordingly is extremely bad practice and ill advised with typings inside the library. However, as library author you have no commitment to keep typings in DefinitetlyTyped up-to-date.

KrzysztofMadejski:
I haven't added all TODOs as I was not sure if I'm not exposing internals too much.

Favna:
> I haven't added all TODOs as I was not sure if I'm not exposing internals too much.

As long as the value is exported (included in `export` or `module.export`) then it can be in the type declarations. If they are not then they should not be. I'd have to dig into the code myself to give a better answer right now.

KrzysztofMadejski:
Sure, I won't have time for a while to fix it, so please close it for now.
Thanks for the review everyone

wt., 29 paź 2019, 17:09 użytkownik TypeScript Bot <notifications@github.com>
napisał:

PR #39356, opened 2019-10-23
[html-validator] improve response message typings

ddgenome:
Enhance response message typings to accurately reflect all of the
types and subtypes of the messages.  Peruse
https://github.com/validator/validator/tree/master/src/nu/validator/messages/types
to confirm the representations here are correct.

Reflect that a message either has or does not have location
information.  Note that this implementation results in the location
properties not being available on the returned messages without
asserting them to the ValidationMessageLocationObject type (see the
updated test).  This approach is more correct and easier when using
strict null checks, but makes it more cumbersome to access the
location properties when your TypeScript compiler configuration has
strict null checks set to false.  Another option would be to do away
with the concept of a separate location sub-object and just make all
the location properties optional.

There was recently a new major version release of the upstream
project, so this seems like a good time to introduce this breaking
change.

Fix indentation.

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/validator/validator/tree/master/src/nu/validator/messages/types

### DIFF
--- a/types/html-validator/index.d.ts
+++ b/types/html-validator/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for html-validator 4.1
+// Type definitions for html-validator 5.0
 // Project: https://github.com/zrrrzzt/html-validator
 // Definitions by: Takesi Tokugawa <https://github.com/TokugawaTakesi>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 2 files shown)


types/html-validator/index.d.ts
```diff
@@ -42,51 +42,77 @@ declare namespace HtmlValidator {
...
-      extract: string;
+    interface ValidationMessageBasicObject {
+        message: string;
     }
+
+    interface ValidationMessageBasicLocationObject {
+        lastLine: number;
+        firstColumn: number;
+        lastColumn: number;
+        hiliteStart: number;
```
TokugawaTakesi: Please make sure that no typo mistake here.

TokugawaTakeshi (approved):
Nice work! You has quite investigated the Validator API.

PR #39372, opened 2019-10-24
Add types for systemjs@6

internettrans:
Hi there 👋. I am a current, active maintainer of SystemJS. We rewrote the whole project late last year and the typescript definitions are now out of date. I am hoping to maintain the 0.20 typescript definitions for those using systemjs@<1, but add definitions for systemjs@6 as the `latest` types on npm.

I attempted to do so by creating a `v0` directory and moving the old definitions there. I don't know if that is the correct thing to do, but saw other DefinitelyTyped packages doing that. If I missed anything there, let me know.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/systemjs/systemjs/blob/master/docs/api.md

### DIFF
--- a/types/systemjs/index.d.ts
+++ b/types/systemjs/index.d.ts
@@ -68,7 +68,7 @@ declare const System: {
 // tslint:disable-next-line no-unnecessary-generics
 type ImportFn = <T extends Module>(moduleId: string, parentUrl?: string) => Promise<T>;
 
-type DeclareFn = (_export: typeof ExportFn, _context: Context) => Declare;
+type DeclareFn = (_export: ExportFn, _context: Context) => Declare;
 interface Declare {
   setters?: SetterFn[];
   execute?(): any;
@@ -76,8 +76,10 @@ interface Declare {
 type SetterFn = (moduleValue: Module) => any;
 type ExecuteFn = () => any;
 
-declare function ExportFn(exportName: string, value: any): void;
-declare function ExportFn(exports: object): void;
+interface ExportFn {
+  (exportName: string, value: any): void;
+  (exports: object): void;
+}
 
 type UpdateModuleFn = () => void;
 
--- a/types/systemjs/index.d.ts
+++ b/types/systemjs/index.d.ts
@@ -9,7 +9,6 @@ declare const System: {
    * Loads a javascript module from either a url or bare specifier that is in an import map.
    * You may optionally provide a parentUrl that will be used for resolving relative urls.
    */
-  // tslint:disable-next-line no-unnecessary-generics
   import: System.ImportFn;
 
   /**
--- a/types/systemjs/index.d.ts
+++ b/types/systemjs/index.d.ts
@@ -10,7 +10,7 @@ declare const System: {
    * You may optionally provide a parentUrl that will be used for resolving relative urls.
    */
   // tslint:disable-next-line no-unnecessary-generics
-  import: ImportFn;
+  import: System.ImportFn;
 
   /**
    * Inserts a new module into the SystemJS module registry. The System.register format is
@@ -19,8 +19,8 @@ declare const System: {
    * Register may be called with a name argument if you are using the named-register extra. (See
    * https://github.com/systemjs/systemjs#extras).
    */
-  register(dependencies: string[], declare: DeclareFn): void;
-  register(name: string, dependencies: string[], declare: DeclareFn): void;
+  register(dependencies: string[], declare: System.DeclareFn): void;
+  register(name: string, dependencies: string[], declare: System.DeclareFn): void;
 
   /**
    * Resolve any moduleId to its full URL. For a moduleId that is in the import map, this will resolve
@@ -36,14 +36,14 @@ declare const System: {
    * The returned function is intended for use after re-importing the module. Calling the function
    * will re-bind all the exports of the re-imported module to every module that depends on the module.
    */
-  delete(moduleId: string): false | UpdateModuleFn;
+  delete(moduleId: string): false | System.UpdateModuleFn;
 
   /**
    * Get a module from the SystemJS module registry. Note that the moduleId almost always must be a full url
    * and that you might need to call System.resolve() to obtain the moduleId. If the module does not exist in
    * the registry, null is returned.
    */
-  get(moduleId: string): Module | null;
+  get(moduleId: string): System.Module | null;
   // tslint:disable-next-line no-unnecessary-generics
   get<T>(moduleId: string): T | null;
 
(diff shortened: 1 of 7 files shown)


types/systemjs/index.d.ts
```diff
@@ -1,388 +1,103 @@
...
+    /**
+     * Use for (let entry of System.entries()) to access all of the modules in the SystemJS registry.
+     */
+    entries(): Iterable<typeof ModuleEntry>;
+  }
+
+  // tslint:disable-next-line no-unnecessary-generics
+  function ImportFn<T extends Module>(moduleId: string, parentUrl?: string): Promise<T>;
+
+  function DeclareFn(_export: typeof ExportFn, _context: Context): Declare;
```
weswigham: What are all these uppercase-named functions for? Near as I can tell, there's no `System.DeclareFn` in the public API?

joeldenning: They are for reuse within this file and are not meant to be exported. I noticed that they were being exported but didn't know how to implement types, interfaces, and functions in a global typescript declaration without all of them being exported. I tried to follow [this pattern](https://www.typescriptlang.org/docs/handbook/namespaces.html#working-with-other-javascript-libraries) from the typescript handbook.

Note that my types here have a `SystemJS` namespace and then a `System` const. Newly with systemjs@6, there is actually no longer a `SystemJS` global variable at all. The only global variable is `System`.

weswigham: Add `export {}` into the namespace declaration - then only other members with an `export` modifier will be exported. But if you's just using these to `typeof` them later, it's probably better if you just used type aliases, too, eg `type ImportFn = <T extends Module>(moduleId: string, parentUrl?: string) => Promise<T>;`

joeldenning: 👍 thanks I will update this.

types/systemjs/index.d.ts
```diff
@@ -1,388 +1,104 @@
...
+
+declare namespace System {
+  export {};
+
+  /**
+   * Loads a javascript module from either a url or bare specifier that is in an import map.
+   * You may optionally provide a parentUrl that will be used for resolving relative urls.
+   */
+  // tslint:disable-next-line no-unnecessary-generics
+  export const import: ImportFn;
```
joeldenning: Typescript fails to compile this because it thinks it's invalid to export a thing called `import`. Any ideas on how to fix?

types/systemjs/index.d.ts
```diff
@@ -1,388 +1,104 @@
...
+   */
+  export function resolve(moduleId: string, parentUrl?: string): string;
+
+  /**
+   * Delete a module from the module registry. Note that the moduleId almost always must be a full url and that
+   * you might need to call System.resolve() to obtain the moduleId for modules in an import map.
+   * The returned function is intended for use after re-importing the module. Calling the function
+   * will re-bind all the exports of the re-imported module to every module that depends on the module.
+   */
+  export function delete(moduleId: string): false | UpdateModuleFn;
```
joeldenning: Typescript fails to compile this because it thinks it's invalid to export a thing called delete. Any ideas on how to fix?

types/systemjs/systemjs-tests.ts
```diff
@@ -1,68 +1,81 @@
...
+        setters: [
+            module => {
+                foo = module;
+            },
+            module => {
+                bar = module;
+            }
+        ],
+        execute() {
+            _export('a', 'thing');
```
joeldenning: The tests fail on these `_export()` calls, even though they used to be working and I have no idea why they're failing now.

The error message I'm getting:

```
Error: /Users/joeldennning/code/DefinitelyTyped/types/systemjs/systemjs-tests.ts:25:21
ERROR: 25:21  expect  TypeScript@next compile error: 
Argument of type '"a"' is not assignable to parameter of type 'never'.
ERROR: 26:21  expect  TypeScript@next compile error: 
Argument of type '"b"' is not assignable to parameter of type 'never'.
ERROR: 27:21  expect  TypeScript@next compile error: 
Argument of type '"c"' is not assignable to parameter of type 'never'.
ERROR: 28:13  expect  TypeScript@next compile error: 
Expected 2 arguments, but got 1.
```

I don't know where the `never` is coming from in that error.

types/systemjs/index.d.ts
weswigham: Heh, we interestingly you actually can't have a const named import - go try it in your js repl of choice. Object members can be keywords, though. You may need to replace the namespace with a const to handle this. :S

joeldenning: 👍 I can switch it over to being a const instead of a namespace. How will I be able to create reusable types that don't get exported, though, if it's a const?

types/systemjs/systemjs-tests.ts
joeldenning: Figured this one out -- went back to polymorphic function signature and `typeof`, which solved the problem here.

types/systemjs/index.d.ts
weswigham: You can merge a namespace that only contains types with the const.

types/systemjs/systemjs-tests.ts
weswigham: Oh, it's because you don't want a union type, you want an overload list. Looks like this:
```ts
type ExportFn = {
    (exportName: string, value: any): void;
    (exports: object): void;
}; 
```

weswigham (changes requested):
@joeldenning 
> Oh, it's because you don't want a union type, you want an overload list. Looks like this:
```ts
type ExportFn = {
    (exportName: string, value: any): void;
    (exports: object): void;
}; 
```

?

joeldenning:
Thanks for the bump - I have updated this with your feedback. The linter didn't like that it was a type instead of an interface, so i changed it to an interface.

joeldenning:
> You can merge a namespace that only contains types with the const.

How can I do this? Will it result in the namespace I create becoming a global variable itself?

weswigham:
```
declare var System: System.Instance;
declare namespace System {
    export interface Instance {}
    export interface Blah {}
}
```

joeldenning:
👍 thanks for your help! I just pushed a commit that does that.

types/systemjs/index.d.ts
```diff
@@ -1,388 +1,103 @@
...
-        warnings: boolean;
-    }
+// TypeScript Version: 3.5
+
+declare const System: {
+  /**
+   * Loads a javascript module from either a url or bare specifier that is in an import map.
+   * You may optionally provide a parentUrl that will be used for resolving relative urls.
+   */
+  // tslint:disable-next-line no-unnecessary-generics
```
weswigham: This suppression seems unused here? right?

joeldenning: Yep good catch. Just pushed a commit that removes that

weswigham (approved):
cc @sandersn @jessetrinity - you'll want to double check the build here and then merge, I guess?

joeldenning:
🎉

PR #39381, opened 2019-10-24
Node: `util.promisify`: widen callback error parameter to `any`

OliverJAsh:
In https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39173 I had to use `PoorMansUnknown` instead of `unknown` because the types had to support TS versions < 3 (i.e. when `unknown` support was added).

However, I didn't realise that this wouldn't allow `unknown` to be used by TS 3 users. E.g. this test would fail:

```ts
// Type '(arg1: string, arg2: (err: unknown, result: number) => void) => Promise<number>' is not assignable to type '(arg: string) => Promise<number>'.
const arg1UnknownErrorA: (arg: string) => Promise<number> = promisify((arg: string, cb: (err: unknown, result: number) => void): void => { });
```

This is because `unknown` is not assignable to `PoorMansUnknown`.

```ts
let unknown: unknown;
let poorMansUnknown: {} | null | undefined;

// error
poorMansUnknown = unknown;
```

~This PR adds types specifically for TS 3 users so they may use `unknown`, whilst < TS 3 users may continue to use `PoorMansUnknown`.~

This PR widens the error type to `any`.

---

Please fill in this template.

### DIFF
--- a/types/node/ts3.2/node-tests.ts
+++ b/types/node/ts3.2/node-tests.ts
@@ -24,7 +24,7 @@ import '../util';
 import '../worker_threads';
 import '../zlib';
 
-import { types } from 'util';
+import { types, promisify } from 'util';
 import { BigIntStats, statSync, Stats } from 'fs';
 
 //////////////////////////////////////////////////////////
@@ -49,6 +49,9 @@ import { BigIntStats, statSync, Stats } from 'fs';
         // $ExpectType number
         const b = value;
     }
+
+    const arg1UnknownError: (arg: string) => Promise<number> = promisify((arg: string, cb: (err: unknown, result: number) => void): void => { });
+    const arg1AnyError: (arg: string) => Promise<number> = promisify((arg: string, cb: (err: any, result: number) => void): void => { });
 }
 
 // FS Tests
--- a/types/node/ts3.2/node-tests.ts
+++ b/types/node/ts3.2/node-tests.ts
@@ -24,7 +24,7 @@ import '../util';
 import '../worker_threads';
 import '../zlib';
 
-import { types } from 'util';
+import { types, promisify } from 'util';
 import { BigIntStats, statSync, Stats } from 'fs';
 
 //////////////////////////////////////////////////////////
@@ -49,6 +49,8 @@ import { BigIntStats, statSync, Stats } from 'fs';
         // $ExpectType number
         const b = value;
     }
+
+    const arg1UnknownErrorA: (arg: string) => Promise<number> = promisify((arg: string, cb: (err: unknown, result: number) => void): void => { });
 }
 
 // FS Tests
--- a/types/node/ts3.2/node-tests.ts
+++ b/types/node/ts3.2/node-tests.ts
@@ -24,7 +24,7 @@ import '../util';
 import '../worker_threads';
 import '../zlib';
 
-import { types, promisify } from 'util';
+import { types } from 'util';
 import { BigIntStats, statSync, Stats } from 'fs';
 
 //////////////////////////////////////////////////////////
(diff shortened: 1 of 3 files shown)


types/node/ts3.2/util.d.ts
```diff
@@ -5,9 +5,33 @@ declare module "util" {
...
+    ): (arg1: T1, arg2: T2, arg3: T3, arg4: T4) => Promise<TResult>;
+    function promisify<T1, T2, T3, T4>(fn: (arg1: T1, arg2: T2, arg3: T3, arg4: T4, callback: (err?: unknown) => void) => void):
+        (arg1: T1, arg2: T2, arg3: T3, arg4: T4) => Promise<void>;
+    function promisify<T1, T2, T3, T4, T5, TResult>(
+        fn: (arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, callback: (err: unknown, result: TResult) => void) => void,
+    ): (arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5) => Promise<TResult>;
+    function promisify<T1, T2, T3, T4, T5>(
+        fn: (arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5, callback: (err?: unknown) => void) => void,
+    ): (arg1: T1, arg2: T2, arg3: T3, arg4: T4, arg5: T5) => Promise<void>;
+    function promisify(fn: Function): Function;
```
OliverJAsh: Since `unknown` was added in TS 3, should I add these types for >= 3 instead of >= 3.2? I added them here since this file already existed.

types/node/ts3.2/node-tests.ts
```diff
@@ -49,6 +49,8 @@ import { BigIntStats, statSync, Stats } from 'fs';
         // $ExpectType number
         const b = value;
     }
+
+    const arg1UnknownErrorA: (arg: string) => Promise<number> = promisify((arg: string, cb: (err: unknown, result: number) => void): void => { });
```
OliverJAsh: We should also fix the same problem for other usages of `PoorMansUnknown`. For example, this currently fails:

```ts
/*
Argument of type '{ [key: string]: unknown; }' is not assignable to parameter of type 'ParsedUrlQueryInput'.
  Index signatures are incompatible.
    Type 'unknown' is not assignable to type '{} | null | undefined'.
      Type 'unknown' is not assignable to type '{}'.
*/
const unknownObj: { [key: string]: unknown; } = {};
querystring.stringify(unknownObj);
```

It would be better if we could override `PoorMansUnknown` to make it actually `unknown`, but I don't think that's possible.

types/node/ts3.2/util.d.ts
Flarna: Please do not create one more version of the type definitions except they are really, really needed. Someone hast to maintain this...

Flarna:
Based on the discussions in #39173 I would recommend to use `any` here instead `PoorMansUnknown`/`unknown` to avoid duplication.
In the actual case `any` has no disadvantages but duplication adds maintenance burden.

OliverJAsh:
You're right, it's fine to use `any` in this case.: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39173#issuecomment-543148918. I'll send a PR for that.

Although the same isn't true for the other place we use `PoorMansUnknown`, which we also need to fix: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39381#discussion_r338532250. Do you have any ideas how we should fix that case?

Flarna:
> Do you have any ideas how we should fix that case?

Sorry, no idea at the moment. But I haven't look into the other places in detail. Maybe keep it as it is for now as no one complained since this was added for some month now.

OliverJAsh:
Updated this PR to use `any`

Flarna (approved):

PR #29073, opened 2018-09-21
fix(react): React.createContext parameter is optional

bradleyayers:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [ ] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [ ] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: https://reactjs.org/docs/context.html#reactcreatecontext
- [ ] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

> Note: passing undefined as a Provider value does not cause Consumers to use defaultValue.

PR #39383, opened 2019-10-24
React: make defaultValue argument of createContext optional

capaj:
reasoning here: https://twitter.com/capajj/status/1187344742465576960

also from the docs it is obvious the value is not required: https://reactjs.org/docs/context.html#reactcreatecontext

image

Please fill in this template.

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -338,7 +338,7 @@ declare namespace React {
         displayName?: string;
     }
     function createContext<T>(
-        defaultValue: T,
+        defaultValue?: T,
         calculateChangedBits?: (prev: T, next: T) => number
     ): Context<T>;
 


eps1lon (changes requested):
Prior rejected work: 
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37481
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37024
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/34854
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/29073

And most importantly the statement from a member of React's core team:

> If you really want to make it undefined you can pass it explicitly as an argument but it doesn’t make sense as a “default” to me. In my opinion calling it without an argument shouldn’t type check at all.

-- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/24509#issuecomment-382213106

eps1lon:
> So how am I supposed to use a react context without hacking around TS type errors?

`createContext<T | undefined>(undefined)` should work or if you want to skip type safety `createContext<T>(undefined as T)`.

capaj:
`createContext<T | undefined>(undefined)` is a pain to use-the type of context then becomes `T | undefined`. I don't care about undefined. I only use the context in places where the value is provided.

cast with `as` does indeed work, but it's really not very nice, because you have to cast into `unknown` first.

image

ferdaber:
use `createContext<T>(undefined!)` instead.

PR #39184, opened 2019-10-16
[@types/jqueryui] Change event type for widget function callbacks to JQuery.Event

MasterOdin:
Updates the typing so that all callbacks for widget functions are of type `JQuery.Event` and not just standard `Event`, which is consistent with that the API documentation for jquery-ui has the `Event` type link to https://api.jquery.com/Types/#Event.

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes:
https://api.jqueryui.com/category/widgets/

PR #39384, opened 2019-10-24
[@types/jqueryui] Replace JQuery.Event with legacy JQueryEventObject

MasterOdin:
Follow-up to #39184 fixing the break for `@types/jquery@1` and `@types/jquery@2`, which defined the `JQuery` export differently, as well as imports from it, meaning that `JQuery.Event` is not available. Instead, use `JQueryEventObject` which is deprecated on `@types/jquery@3` (available through `legacy.d.ts`), but fully available in `@1` and `@2` (available in their `index.d.ts` files).

Please fill in this template.

### DIFF
--- a/types/jqueryui/index.d.ts
+++ b/types/jqueryui/index.d.ts
@@ -28,7 +28,7 @@ declare namespace JQueryUI {
     }
 
     interface AccordionEvent {
-        (event: JQuery.Event, ui: AccordionUIParams): void;
+        (event: JQueryEventObject, ui: AccordionUIParams): void;
     }
 
     interface AccordionEvents {
@@ -68,7 +68,7 @@ declare namespace JQueryUI {
     }
 
     interface AutocompleteEvent {
-        (event: JQuery.Event, ui: AutocompleteUIParams): void;
+        (event: JQueryEventObject, ui: AutocompleteUIParams): void;
     }
 
     interface AutocompleteEvents {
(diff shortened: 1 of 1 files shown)


jessetrinity:
@MasterOdin I'm not familiar with these types at all so this might be a stupid question - why can't you use `JQueryEventObject | JQuery.Event`?

MasterOdin:
I do not believe so? The types in question could be summed up as that `JQueryEventObject === JQuery.Event` in `@types/jquery@3`, but that in `@types/jquery@2`, `JQuery.Event` does not exist at all (which is causing compile errors with the existing solo `JQuery.Event`). I'm not sure the history of why v2 and v3 have such differing type signatures, but I'm guessing the `legacy.d.ts` file for v3 is for instances like this where a plugin (jquery-ui here) supports multiple major versions of jquery concurrently, which is where the original type definitions of v2 reside. If jquery-ui was a v3 only library, then one would want to just use the `JQuery` namespace exports (like `JQuery.Event`).

Hopefully that makes sense? Explaining typing using historical git context is definitely not an area I've much practice in!

sandersn:
I don't entirely understand the break either, but I'm going to merge this since it a lot of people seem to be broken. @MasterOdin can you watch to see whether this fixes it for others?

MasterOdin:
For a test, the following snippet passes on jQuery 1, 2, and 3 (would have failed on jqueryui < 1.12.8 for all versions) with `@types/jqueryui@1.12.9`:

```typescript
import $ = require('jquery');

let d = $('test').draggable({
  drag: (event) => {
    console.log(event.screenX);
  }
});
```

PR #39394, opened 2019-10-24
[@testing-library/cypress] Fix feedback from #37692 

simjes:
Cypress finally released a new version with updated jQuery typings, so trying this again.
Fixing feedback from #37692

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: #37692

### DIFF
--- a/types/testing-library__cypress/index.d.ts
+++ b/types/testing-library__cypress/index.d.ts
@@ -17,11 +17,9 @@ import {
     SelectorMatcherOptions as DTLSelectorMatcherOptions,
 } from '@testing-library/dom';
 
-import * as JQuery from 'jquery';
-
 export interface CTLMatcherOptions {
     timeout?: number;
-    container?: JQuery;
+    container?: HTMLElement | JQuery;
 }
 
 export type MatcherOptions = DTLMatcherOptions | CTLMatcherOptions;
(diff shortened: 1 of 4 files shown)


aaronmcadam (approved):
So much cleaner! 👏

bastibuck (commented):
Looks good and clean to me :) 

Didn't perform any tests in a real project though

silviuavram:
Hi. I updated Downshift project to @testing-library/cypress 5.0.2 and I think it picked these changes as well. Build is failing with multiple errors like `queryByRole(id: Matcher, options?: MatcherOptions): Chainable<JQuery>` and I saw in this PR changes that JQuery has been removed.

Do you think this caused and, if so, can I have the JQuery import back in that file? I can create a PR myself if you think it's the way to go. Thanks! @simjes @aaronmcadam

simjes:
@silviuavram this now depends on `cypress 3.5.0` and should get its jQuery typings from there. The original jQuery import in the file shouldn't have been there, but my last PR where this was added was merged before having proper QA. I will have to investigate.

simjes:
So I'm seeing the same issue as you are, and I think it is because of npm flattening the dependencies. I tried to explicitly add `/// <refrence types="Cypress" />` in `index.d.ts`, but now I am seeing some type collision with Jest, since both Cypress and Jest declare types in the global scope.

I will create an issue for this as I am not sure how to solve it.

PR #39398, opened 2019-10-24
[aws-lambda] Support for 2.1 S3 event schema, fix versionId which can be omitted.

alex-bolenok-centralreach:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://docs.aws.amazon.com/en_us/lambda/latest/dg/lambda-services.html


simonbuchan (changes requested):
This seems mostly well done (thanks especially for the separate prettier commit!), but I'm confused by why you need a new event type - S3 seems to send the latest schema unconditionally and it's compatible, of course. You're not even breaking test data, since the new property is optional!

Please simply add the `glacierEventData` field to the existing record type, unless I'm missing something (In which case there should also be a new `Handler` type, and there are some name issues).

alex-bolenok-centralreach:
> This seems mostly well done (thanks especially for the separate prettier commit!), but I'm confused by why you need a new event type - S3 seems to send the latest schema unconditionally and it's compatible, of course. You're not even breaking test data, since the new property is optional!

I've also added static typings to `eventVersion` and `s3SchemaVersion` in the new type.

Ideally, we should have added the `eventVersion: '2.0'` to the old event type and make the `S3Event` a union type of the two, however, that's quite a bold move which could potentially break some stuff.

S3 does send the latest schema unconditionally, but even if the lambda is set up as event handler, it can still consume events generated outside AWS (it's a common pattern to emulate events by sending them directly to the lambdas). These events can have all kinds of shapes. Also, there's local AWS emulators, test suites and millions of other scenarios where the `2.0` schema should be treated as a valid input.

The new type something you opt-in to, if you want the typed event version and a schema version.

> Please simply add the `glacierEventData` field to the existing record type, unless I'm missing something (In which case there should also be a new `Handler` type, and there are some name issues).

Good catch about the `Handler` type, will fix it!

What name issues did you mean and how do you think we should solve them?

alex-bolenok-centralreach:
> This seems mostly well done (thanks especially for the separate prettier commit!), but I'm confused by why you need a new event type - S3 seems to send the latest schema unconditionally and it's compatible, of course. You're not even breaking test data, since the new property is optional!

I've also added static typings to `eventVersion` and `s3SchemaVersion` in the new type.

Ideally, we should have added the `eventVersion: '2.0'` to the old event type and make the `S3Event` a union type of the two, however, that's quite a bold move which could potentially break some stuff.

S3 does send the latest schema unconditionally, but even if the lambda is set up as event handler, it can still consume events generated outside AWS (it's a common pattern to emulate events by sending them directly to the lambdas). These events can have all kinds of shapes. Also, there's local AWS emulators, test suites and millions of other scenarios where the `2.0` schema should be treated as a valid input.

The new type something you opt-in to, if you want the typed event version and a schema version.

> Please simply add the `glacierEventData` field to the existing record type, unless I'm missing something (In which case there should also be a new `Handler` type, and there are some name issues).

Good catch about the `Handler` type, will fix it!

What name issues did you mean and how do you think we should solve them?

simonbuchan:
> I've also added static typings to `eventVersion` and `s3SchemaVersion` in the new type.

Yes, but typing the version doesn't help much when it's not something like a union descriminant. You might be able to check `eventVersion == "2.0"` and know that `glacierEventData` will be undefined, but you could just check `glacierEventData`! If there's some *semantic* difference that isn't captured in the type system between the versions, then clearly typing the version values doesn't help there either.

> Ideally, we should have added `eventVersion: '2.0'` to the old event type and make the `S3Event` a union type of the two, however, that's quite a bold move which could potentially break some stuff.

That would be the preferred option if you needed to discriminate between the old and new event version (which doesn't seem to be a realistic use case to me, see below). In general, avoid it, since typescript is still pretty bad at ergonomics for unions (largely due to the lack of exact/closed types).

> Also, there's local AWS emulators, test suites and millions of other scenarios where the `2.0` schema should be treated as a valid input.

And still would be: the addition of an optional property isn't incompatible in either input or output positions. Even if the change was output incompatible then my suggestion would be for the simulator case, update the simulator. For the test cases, either update the tests to include the info that your handlers will receive (it will receive them in production, after all!) or declare the handler as taking your own event type declared with what you actually need, then assign to the actual type (e.g. `export const handler: S3Handler = actualHandler;`) to check compatibility, or use a proxy or just cast. This has nothing to do with AWS, it's just a fact about writing tests for an API that interacts with an external system in typescript. See my comments on #38342.

> The new type is something you opt-in to, if you want typed event version and schema version.

Consider: any version N+1 event sent must be input compatible with the previous version N (e.g. N+1 is assignable to N) or the lambda must opt into receiving the new version: otherwise AWS would be breaking you! Since you don't opt into the new values, why do you have to opt into the new types? Narrowing input types is not normally a breaking change, so narrowing the versions to literals also wouldn't be enough, even if I did agree that that was a good idea.

> What name issues did you mean and how do you think we should solve them?

`Glacier*` interface types should be `S3` prefixed or inlined. `S3Data1` should be something like `S3EventRecordS31` (as ugly as it is) to match the naming pattern (otherwise where is this `Data` name coming from?) or inlined for consistency with the existing `S3EventRecord`.

alex-bolenok-centralreach:
@simonbuchan yeah it seems I was overthinking it

simonbuchan (approved):
Looks good! Thanks for your work and patience.

PR #39399, opened 2019-10-24
chore(node): remove non current/lts versions

SimonSchick:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/en/about/releases/

Changes keep being made to non LTS/Current versions of the node typings, imo we should not maintain these for the same reason the node team doesn't support these versions.

Since the maintenance effort is significantly lower however, I think adding an extra 6 months as a grace period for LTS versions is reasonable.


SimonSchick:
Looks like removing version `0` breaks the build pipeline lol.

Flarna (approved):

SimonSchick:
@jessetrinity can I get some 👀 ?

jessetrinity:
Unrelated to the failure - why do you want to remove all of the old versions?

nguymin4:
I guess Simon's idea was to remove all end-of-life version https://github.com/nodejs/Release
But my opinion is that we only remove versions < 8  which is maintainance LTS until the end of December 2019.

SimonSchick:
@jessetrinity rationale is in OP
@nguymin4 That is correct, I want to retain all *active* LTS versions but at the same time remove all non unmaintained non-LTS versions after v8 (9, 11) as they have the lowest value for consumers.

SimonSchick:
@jessetrinity do I have to re-open this?

PR #39416, opened 2019-10-25
[react, react-dom] declare experimental APIs

Jessidhia:
Add a optionally includable `react/experimental` and `react-dom/experimental` module augmentations that declare the types and functions exported by the experimental builds of React.

See https://github.com/facebook/react/blob/master/packages/shared/ReactFeatureFlags.js for the features currently flagged as `__EXPERIMENTAL__`; this currently includes [`useTransition`, `useDeferredValue`, `SuspenseList` and `unstable_withSuspenseConfig`](https://github.com/facebook/react/blob/f6b8d31a76cbbcbbeb2f1d59074dfe72e0c82806/packages/react/src/React.js#L110-L115) in `react`, and [`createRoot`, `createBlockingRoot`, `unstable_discreteUpdates`, `unstable_flushDiscreteUpdates`, `unstable_flushControlled` and `unstable_scheduleHydration`](https://github.com/facebook/react/blob/f6b8d31a76cbbcbbeb2f1d59074dfe72e0c82806/packages/react-dom/src/client/ReactDOM.js#L662-L675) in `react-dom`.


Jessidhia:
I'm hitting these CI failures locally too -- they are not actually CI errors with React, and these changes are for types that don't affect existing libraries anyway. Something in `dtslint` is triggering a race condition with `npm`.

johnnyreilly (approved):

types/react/experimental.d.ts
```diff
@@ -0,0 +1,205 @@
...
+         *
+         * **If some state update causes a component to suspend, that state update should be wrapped in a transition.**
+         *
+         * @param callback A _synchronous_ function which causes state updates that can be deferred.
+         */
+        (callback: TransitionFunction): void;
+    }
+
+    /**
+     * **NOTE: passing a function value causes it to be treated like a `useState` initializer**
```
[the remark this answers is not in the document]
eps1lon: I would not add undocumented behavior to the typings. It's not our job to document every react bug. Especially for experimental features this puts more maintenance burden on us than necessary.

types/react-dom/experimental.d.ts
```diff
@@ -0,0 +1,102 @@
...
+        render(children: React.ReactChild | React.ReactNodeArray): Work;
+        then(onComplete?: () => void): void;
+    }
+
+    interface RootOptions {
+        hydrate?: boolean;
+    }
+
+    interface BlockingRoot {
+        render(children: React.ReactChild | React.ReactNodeArray, callback?: () => void): Work;
```
eps1lon: These no longer return `Work`: https://github.com/facebook/react/blob/f6b8d31a76cbbcbbeb2f1d59074dfe72e0c82806/packages/react-dom/src/client/ReactDOM.js#L185

types/react-dom/experimental.d.ts
```diff
@@ -0,0 +1,102 @@
...
+        commit(): void;
+        render(children: React.ReactChild | React.ReactNodeArray): Work;
+        then(onComplete?: () => void): void;
+    }
+
+    interface RootOptions {
+        hydrate?: boolean;
+    }
+
+    interface BlockingRoot {
```
eps1lon: As far as I know the interface for the roots returned from createRoot and createBlockingRoot are identical. We should remove the createRoot typings from the index.d.ts anyway since they got [removed in 16.10](https://github.com/facebook/react/blob/master/CHANGELOG.md#react-dom) and then  we can reuse them here.

types/react-dom/experimental.d.ts
```diff
@@ -0,0 +1,102 @@
...
+     * @see https://reactjs.org/docs/concurrent-mode-adoption.html#migration-step-blocking-mode
+     * @see https://reactjs.org/docs/concurrent-mode-reference.html#createblockingroot
+     */
+    function createBlockingRoot(
+        container: Element | Document | DocumentFragment | Comment,
+        options?: RootOptions,
+    ): BlockingRoot;
+
+    interface Root extends BlockingRoot {
+        createBatch(): Batch;
```
eps1lon: Batches no longer exist for rendering.

eps1lon (changes requested):
I'm not so happy with the solutions to enable these types. Without an existing  `types` entry we would need to add these references/imports to every file, right?

I would rather see a `paths` entry where we could write
```json
"paths": { "react": "react/experimental" } 
```

and be done. The experimental.d.ts could then re-export everything from index.d.ts.

This should work as well, no?

types/react-dom/experimental.d.ts
Jessidhia: Forgot to update these 😓

types/react/experimental.d.ts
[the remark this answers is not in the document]
Jessidhia: TBH the one thing I wanted to do is `extends not Function`, but this is the best I can do for now. I'll be looking into fixing the React bug myself so I'll update it here when it's done.

types/react-dom/experimental.d.ts
eps1lon: I would've done the same :smile: That's what reviews are for.

Jessidhia:
@eps1lon Using `"paths"` would actually turn this into a _hazard_ because when respected by, say, `tsconfig-paths-webpack-plugin `, it would rewrite _correct_ imports into imports to a module that does not exist.

The three methods I've outlined (`compilerOptions.types`, triple-slash or explicitly empty import) only have to be done once, in _any_ file that is directly or indirectly by files within the project (i.e. belonging to the same `tsconfig.json`). It could be done in the index file, for example:

```ts
/// <reference types='react/experimental' />
/// <reference types='react-dom/experimental' />
```

and this would be enough to make the hooks / SuspenseList / createRoot / etc available to any other file, even over UMD references to React or ReactDOM.

types/react-dom/experimental.d.ts
```diff
@@ -0,0 +1,102 @@
...
+
+import React = require('react');
+import ReactDOM = require('.');
+
+export {};
+
+declare module '.' {
+    // enableSuspenseServerRenderer feature
+    interface HydrationOptions {
+        onHydrated?(suspenseInstance: Comment): void;
```
Jessidhia: The server renderer writes boundary comments that look like `<!--$-->` or `<!--/$-->` (`ReactPartialRenderer.js`) around `Suspense`s that, uh, suspended, and the `suspenseInstance` itself is the first boundary comment.

Nothing much useful to do with this outside of tests, though, it just exists; it's only partially implemented.

Jessidhia:
I am impressed, CI passed

eps1lon:
> tsconfig-paths-webpack-plugin , it would rewrite correct imports into imports to a module that does not exist.

Why would it do that? That's not how `paths` is supposed to work and directly conflicts with recommendations of the DT repo:

https://github.com/DefinitelyTyped/DefinitelyTyped#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package

eps1lon (approved):

Jessidhia:
> Why would it do that? That's not how paths is supposed to work and directly conflicts with recommendations of the DT repo

This is because those guidelines are for DefinitelyTyped's own tests, which are type-only tests -- in that case it doesn't matter if the aliased imports are entirely fictional, the code is never emitted or executed.

eps1lon:
> > Why would it do that? That's not how paths is supposed to work and directly conflicts with recommendations of the DT repo
> 
> This is because those guidelines are for DefinitelyTyped's own tests, which are type-only tests -- in that case it doesn't matter if the aliased imports are entirely fictional, the code is never emitted or executed.

What makes you think that? They apply just as well to application code where I import history v1 and want the correct typings for it. 

And it still does not address how `paths` is specified to work: https://github.com/microsoft/TypeScript/issues/9910#issuecomment-234729007

`paths` can reflect your module mapping not the other way around.

Jessidhia:
If you _also_ want to make `paths` work, assuming it would indeed have no effects (i.e. no webpack plugin taking the aliases "seriously"), then the `export {}` line on the modules could be changed to `export = React` / `export = ReactDOM` instead -- but then it would make it possible for one to mistakenly actually import things from `react/experimental` which doesn't exist.

It's not possible to "hide" the imports to only be available in the `experimental` imports (i.e. only with the `paths` mapping) either because `react` / `react-dom` are not ES modules; they either have to reexport the original object or do some truly annoying `namespace` unions, and it would require `tsconfig.json` modification, unlike the current solution which works without config changes if you write a triple-slash reference.

FWIW, the pattern I went with here is the same as was used in `styled-components/cssprop.d.ts`.

PR #39419, opened 2019-10-25
analytics-node - export whole AnalyticsNode namespace

dgobaud:
Sometimes need access to the other items in the namespace - current export doesn't allow it

- [x] - this is a minor change


dgobaud:
this doesn't work i'm not actually sure if it can be done will make an issue instead

PR #39421, opened 2019-10-25
[express-session] Add 'auto' value to cookie.secure

CatsMiaow:
https://github.com/expressjs/session/issues/703

express-session supports the `auto` value for the `cookie.secure` option.
https://github.com/expressjs/session#cookiesecure
https://github.com/expressjs/session/blob/master/index.js#L163
> The cookie.secure option can also be set to the special value 'auto' to have this setting automatically match the determined security of the connection.

Updated TypeScript Version with dependencies and passed the npm test.

### DIFF
--- a/types/express-session/index.d.ts
+++ b/types/express-session/index.d.ts
@@ -5,7 +5,7 @@
 //                 Naoto Yokoyama <https://github.com/builtinnya>
 //                 Ryan Cannon <https://github.com/ry7n>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+// TypeScript Version: 2.8
 
 /// <reference types="node" />
 
@@ -53,12 +53,14 @@ declare global {
 
 declare function session(options?: session.SessionOptions): express.RequestHandler;
 
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+
 declare namespace session {
   interface SessionOptions {
     secret: string | string[];
     name?: string;
     store?: Store | MemoryStore;
-    cookie?: express.CookieOptions;
+    cookie?: Omit<express.CookieOptions, 'secure'> & { secure?: boolean | 'auto' };
     genid?(req: express.Request): string;
     rolling?: boolean;
     resave?: boolean;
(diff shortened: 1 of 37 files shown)


types/express-session/index.d.ts
```diff
@@ -53,12 +53,14 @@ declare global {
...
 
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+
 declare namespace session {
   interface SessionOptions {
     secret: string | string[];
     name?: string;
     store?: Store | MemoryStore;
-    cookie?: express.CookieOptions;
+    cookie?: Omit<express.CookieOptions, 'secure'> & { secure?: boolean | 'auto' };
```
Rokt33r: I think we can deal with the problem without introducing `Omit`. Express session is one of most popular packages. If we upgrade minimum typescript version, millions of servers should be affected and might cause a huge disaster.

My suggestion is like the below. Could you try this?

```ts
cookie?: express.CookieOptions & { secure?: boolean | 'auto' }
```

Rokt33r (changes requested):

types/express-session/index.d.ts
CatsMiaow: Yes, the first thing I tried.
image

cannot add incompatible types to the type of the parent attribute.
See https://stackoverflow.com/q/41285211

Rokt33r: Okay, if so, I think we should write a new interface rather than using `express.CookieOptions`.

cjbarth: If we upgrade the minimum TypeScript version, I don't see how we'll cause a huge disaster. The servers should be package-locked to a version of the types, and if they are on an older version of TypeScript, they shouldn't get these newer definitions. I actually think that it would be OK to use `Omit` as that appears the proper idiomatic way of doing this.

Rokt33r: @cjbarth @CatsMiaow But, in the document of express-session, it doesn't say anything explicitly about that you can use same options of `express.CookieOptions` for `cookie` prop.

Rokt33r: @cjbarth @CatsMiaow But, in the document of express-session, it doesn't say anything explicitly about that you can use same options of express.CookieOptions for cookie prop.

Rokt33r: So using `express.CookieOptions` might be wrong in the first place.

CatsMiaow: `express` and `express-session` have dependencies on the same `cookie` module.

https://github.com/expressjs/express/blob/master/lib/response.js#L30
https://github.com/expressjs/session/blob/master/session/cookie.js#L14

Rokt33r: It doesn't matter. Express-session is wraping cookie. It doesn't let you touch options for cookie module directly.

Rokt33r:
I'll do it by myself.

PR #39426, opened 2019-10-25
feat: Add `internal‑slot`

ExE-Boss:
Please fill in this template.

If adding a new definition:

---

review?(@ljharb)

### DIFF
--- a/types/internal-slot/index.d.ts
+++ b/types/internal-slot/index.d.ts
@@ -1,0 +1,26 @@
+// Type definitions for internal-slot 1.0
+// Project: https://github.com/ljharb/internal-slot#readme
+// Definitions by: ExE Boss <https://github.com/ExE-Boss>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
+/**
+ * Asserts that `O` is an `object`, `slot` is a `string`
+ * and `O` has the internal slot `slot`.
+ */
+export function assert(O: object, slot: string): void;
+
+/**
+ * Gets the value of `O`'s internal slot `slot`.
+ */
+export function get(O: object, slot: string): any;
+
+/**
+ * Checks that `O` has the internal slot `slot`.
+ */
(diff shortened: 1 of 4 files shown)


types/internal-slot/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+// Project: https://github.com/ljharb/internal-slot#readme
+// Definitions by: ExE Boss <https://github.com/ExE-Boss>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
+
+/**
+ * Asserts that `O` is an `object`, `slot` is a `string`
+ * and `O` has the internal slot `slot`.
+ */
+export function assert(O: object, slot: string): void;
```
ljharb:
```suggestion
export function assert(O: object, slot: string): void | never;
```
perhaps? also this should use that new syntax that indicates that it can throw, if possible

types/internal-slot/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Gets the value of `O`'s internal slot `slot`.
+ */
+export function get(O: object, slot: string): any;
```
ljharb:
```suggestion
export function get(O: object, slot: string): unknown;
```

types/internal-slot/index.d.ts
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Checks that `O` has the internal slot `slot`.
+ */
+export function has(O: object, slot: string): boolean;
+
+/**
+ * Sets the object `O`'s internal slot `slot`'s value to `V`.
+ */
+export function set(O: object, slot: string, V: any): void;
```
ljharb:
```suggestion
export function set(O: object, slot: string, V: unknown): void;
```

ljharb (approved):

types/internal-slot/index.d.ts
ExE-Boss: That just collapses into `void`.

PR #39435, opened 2019-10-25
Add node-beanstalkd-worker library types

SomaticIT:
Please fill in this template.

If adding a new definition:

**/!\ Note and Warning: Depends on PR #39434**

### DIFF
--- a/types/beanstalkd-worker/index.d.ts
+++ b/types/beanstalkd-worker/index.d.ts
@@ -1,0 +1,248 @@
+// Type definitions for beanstalkd-worker 1.2
+// Project: https://github.com/burstable/node-beanstalkd-worker
+// Definitions by: Maxime LUCE <https://github.com/SomaticIT>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+import BeanstalkdClient, { BeanstalkdJobState, BeanstalkdJobStats } from "beanstalkd";
+
+export default class BeanstalkdWorker {
+                   host?: string;
+                   port?: number;
+                   options?: {};
+                   running: boolean;
+
+                   constructor(host?: string, port?: number, options?: {});
+
+                   /**
+                    * Create or reuse a connection to Beanstalkd with given id.
+                    *
+                    * @param id Id for the connection.
+                    */
+                   connection(id: string): Promise<BeanstalkdClient>;
+
+                   /**
+                    * Get or create a Tube.
+                    *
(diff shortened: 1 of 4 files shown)


types/beanstalkd-worker/index.d.ts
```diff
@@ -0,0 +1,248 @@
+// Type definitions for beanstalkd-worker 1.2
+// Project: https://github.com/burstable/node-beanstalkd-worker
+// Definitions by: Maxime LUCE <https://github.com/SomaticIT>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+import BeanstalkdClient, { BeanstalkdJobState, BeanstalkdJobStats } from "beanstalkd";
+
+export default class BeanstalkdWorker {
```
weswigham: OK. So. I hate to say this, but [the package at runtime does not export a default to commonjs](https://runkit.com/embed/yij9h0iju8g3). The babel plugin (old version of babel?) they use to build hoists their `default` export into the entire module. So, I suggest:
```suggestion
export = BeanstalkdWorker;
class BeanstalkdWorker {
```
You'll need to lump all the other exported typed into a `namespace BeanstalkdWorker` to get them exported as well.

weswigham (changes requested):

types/beanstalkd-worker/index.d.ts
SomaticIT: Hello @weswigham,

I checked on the built version of the plugin and it's exported on `default` export.

```javascript
exports['default'] = BeanstalkdWorker;
module.exports = exports['default'];
```

So I think that the default export is the best way to type it.

SomaticIT: No, after a runtime check, you're right, it seems to be exported badly...

SomaticIT: I fixed it in commit https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39435/commits/07c7d5f73cfc901df332ca496b6cf1d745a3f082

SomaticIT:
Any news on the new version?
Thanks

types/beanstalkd/index.d.ts
```diff
@@ -0,0 +1,536 @@
+// Type definitions for beanstalkd 2.2
```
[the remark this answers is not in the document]
SomaticIT: Rebased

weswigham (approved):

PR #39456, opened 2019-10-25
logform: Provides its own types

a-tarasyuk:
This will deprecate `@types/logform` in favor of just `logform`. CC @DABH


DABH (approved):
This seems fine to me as long as any other downstream dependencies in DT have been taken care of (I am not aware of any but you should double check).  Thanks!

PR #39464, opened 2019-10-25
mobx-devtools-mst: Provides its own types

a-tarasyuk:
This will deprecate `@types/mobx-devtools-mst` in favor of just `mobx-devtools-mst`. CC @pluma


pluma (approved):
lgtm

PR #39472, opened 2019-10-25
natsort: Provides its own types

a-tarasyuk:
This will deprecate `@types/natsort` in favor of just `natsort`. CC @mgroenhoff


mgroenhoff (approved):
Thanks!

PR #39476, opened 2019-10-25
neffos.js: Provides its own types

a-tarasyuk:
This will deprecate `@types/neffos.js` in favor of just `neffos.js`. CC @kataras


kataras:
Why?

a-tarasyuk:
@kataras [neffos.js](https://github.com/kataras/neffos.js/blob/master/package.json#L21) provides types definition.

cc @sandersn

PR #39481, opened 2019-10-25
node-sql-parser: Provides its own types

a-tarasyuk:
This will deprecate `@types/node-sql-parser` in favor of just `node-sql-parser`. CC @taozhi8833998


taozhi8833998 (approved):
thanks for that

PR #39486, opened 2019-10-25
oembed-parser: Provides its own types

a-tarasyuk:
This will deprecate `@types/oembed-parser` in favor of just `oembed-parser`. CC @BendingBender, @CodeBast4rd


CodeBast4rd (approved):
Changes to the typings are now in a pull-request of the packages repository and can be removed from here.

PR #39493, opened 2019-10-25
phin: Provides its own types

a-tarasyuk:
This will deprecate `@types/phin` in favor of just `phin`. CC @twooster


twooster (approved):
Bye bye phin

PR #39494, opened 2019-10-25
pixi.js: Provides its own types

a-tarasyuk:
This will deprecate `@types/pixi.js` in favor of just `pixi.js`. CC @clark-stevenson


clark-stevenson:
@a-tarasyuk 

You are a Scholar 🙇

clark-stevenson (approved):
Awesome

clark-stevenson:
@a-tarasyuk 

Just a second!  I felt I knew what you said when I first read it. 

I caught `not-needed-packages` . 

What does this mean for people who are still using V4?  Those people require @types/pixi.js,  will that still be available to them?

sandersn:
Those stay on npm. Only version 5 is tagged as deprecated. But you can’t update 4.x anymore after deleting it from DT.

sandersn (changes requested):
Sounds like the owner still has questions.

clark-stevenson:
Thanks @sandersn in that case this looks good to me!

clark-stevenson (approved):
Happy to proceed

sandersn (approved):

PR #39496, opened 2019-10-25
printf: Provides its own types

a-tarasyuk:
This will deprecate `@types/printf` in favor of just `printf`. CC @AluisioASG


AluisioASG (approved):
LGTM

PR #39521, opened 2019-10-25
react-navigation: Provides its own types

a-tarasyuk:
This will deprecate `@types/react-navigation` in favor of just `react-navigation`. CC @huhuanming, @mhcgrq, @fangpenlin, @petejkim, @iRoachie, @phanalpha, @charlesfamu, @timwangdev, @bang88, @svbutko, @levito, @YourGamesBeOver, @ArmandoAssuncao, @cliedeman, @magrinj, @TizioFittizio, @stigi, @LinusU, @jshosomichi, @jakebooyah, @brunoro, @DenisFrezzato, @mickaelw, @maxdavidson, @alechill, @builtbyproxy, @jkillian, @jeroenvervaeke, @chagasaway, @denissb, @skovhus, @azrosen92, @hmajid2301
WARNING: other packages depend on this!

PR #39522, opened 2019-10-25
react-native-tab-view: Provides its own types

a-tarasyuk:
This will deprecate `@types/react-native-tab-view` in favor of just `react-native-tab-view`. CC @kaoDev, @iRoachie, @timwangdev, @geriux, @kazyk


kaoDev (approved):
@a-tarasyuk thanks for the update

iRoachie:
This build fails because `react-router-navigation-core` depends on this package. Will need to update `react-router-navigation-core` first.

iRoachie (changes requested):
This build fails because `react-router-navigation-core` depends on this package. Will need to update `react-router-navigation-core` first.

Also this seems to be same PR as #39521. Which is one is the real PR

a-tarasyuk:
cc @sandersn

iRoachie (approved):

PR #39527, opened 2019-10-25
react-sortable-pane: Provides its own types

a-tarasyuk:
This will deprecate `@types/react-sortable-pane` in favor of just `react-sortable-pane`. CC @rhysd


rhysd (approved):
Thank you for this patch. I did not notice the package had been rewritten with TypeScript. Looks good.

PR #39529, opened 2019-10-25
react-swipeable: Provides its own types

a-tarasyuk:
This will deprecate `@types/react-swipeable` in favor of just `react-swipeable`. CC @GiedriusGrabauskas, @mctep, @horiuchi, @adambowles


adambowles:
@hartzis and I decided to leave this up as it covers the v4 types (react-swipeable added its own types in v5)

I am happy to take advice on this though

sandersn:
@adambowles it looks like you made 1 change to the types this year, after v5 came out. If you don't expect to make any more, then it's fine to remove the types -- the existing definitions stay on npm of course.

If you might make changes in the future, it also fine to close this PR.

My guess is that removing the types is the right thing, but you know react-swipeable and I don't.

sandersn (changes requested):
Requesting changes so we don't merge by mistake.

adambowles:
@sandersn I helped write the types that react-swipeable now provides for itself. I don’t know how DefinitelyTyped deals with that situation. If deprecating this will leave it available to be installed for those still using react-swipeable@4.x the great, let’s do that.

sandersn:
Yep, they will still be there! `npm install @types/react-swipeable@4` will continue to work and with the source gone from DT, we'll never publish a new 4.x version.

PR #39530, opened 2019-10-25
react-tether: Provides its own types

a-tarasyuk:
This will deprecate `@types/react-tether` in favor of just `react-tether`. CC @ryprice


ryprice:
Looks good to me. I think I stopped using this definition in favor of react-tether in 2018. Thanks for cleaning up!

ryprice (approved):
Looks good to me. I think I stopped using this definition in favor of react-tether in 2018. Thanks for cleaning up!

PR #39534, opened 2019-10-25
redom: Provides its own types

a-tarasyuk:
This will deprecate `@types/redom` in favor of just `redom`. CC @RauliL, @nemoinho


RauliL:
Looks OK to me.

RauliL (approved):

PR #39536, opened 2019-10-25
rfc4648: Provides its own types

a-tarasyuk:
This will deprecate `@types/rfc4648` in favor of just `rfc4648`. CC @thibmeu


notNeededPackages.json
```diff
@@ -3186,6 +3186,12 @@
             "sourceRepoURL": "https://github.com/sindresorhus/rev-hash",
             "asOfVersion": "3.0.0"
         },
+        {
+            "libraryName": "rfc4648",
+            "typingsPackageName": "rfc4648",
+            "sourceRepoURL": "https://github.com/swansontec/rfc4648",
```
thibmeu: URL is not valid. I believe it should be `https://github.com/swansontec/rfc4648.js`

a-tarasyuk:
cc @sandersn URL is exactly as in [yarn not-needed rfc4648 1.3.0 https://github.com/swansontec/rfc4648](https://gist.github.com/sandersn/d9ca982ff4d66be453d2c6f8f9be7fb9#file-not-needed-packages-txt-L1302). Is it correct?

sandersn:
The URL is pulled from the header, which can be wrong. In this case, @thibmeu is right and the header is wrong.

This is only used in the deprecation README, so it's not paramount to have it right anyway. Most people will assume that if `@types/rfc4648` is deprecated, they should just install `rfc4648`.

PR #39537, opened 2019-10-25
riot: Provides its own types

a-tarasyuk:
This will deprecate `@types/riot` in favor of just `riot`. CC @Stubb0rn


Stubb0rn (approved):
Looks good. `riot` 4 was basically a complete rewrite, so these typings aren't useful for anyone using `riot` v4+

PR #39546, opened 2019-10-25
solidity-parser-antlr: Provides its own types

a-tarasyuk:
This will deprecate `@types/solidity-parser-antlr` in favor of just `solidity-parser-antlr`. CC @LogvinovLeon, @albrow, @yxliang01


albrow (approved):
Looks good and we confirmed that this change shouldn't break anything on our end.

Thanks @jalextowle for helping review!

PR #39547, opened 2019-10-25
stacktrace-js: Provides its own types

a-tarasyuk:
This will deprecate `@types/stacktrace-js` in favor of just `stacktrace-js`. CC @exceptionless, @pilagod


pilagod:
@felixfbecker thanks ! 

It seems that you have a misspelling for "libraryName", which should be "stacktrace-js" but not "stacktrace.js".

And could you please make sure if there are any packages in DefinitelyTyped currently dependent on `@types/stacktrace-js` ? You might also need to fix them ([Removing a package](https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package)) by presenting dependency in `package.json`. Thanks : )

pilagod (approved):

PR #39557, opened 2019-10-25
ts-nameof: Provides its own types

a-tarasyuk:
This will deprecate `@types/ts-nameof` in favor of just `ts-nameof`. CC @dsherret


dsherret:
I know why this was done, but please do not merge this! I will comment again when I have time tomorrow.

dsherret:
ts-nameof has its own types, but it distributes a global `nameof` function via the @types/ts-nameof package. Having the @types/ts-nameof package makes it so someone only has to install that and no other configuration is required to get the typings.

I haven't found a better way... the alternative seems to be getting everyone to add a `/// <reference path="node_modules/ts-nameof/ts-nameof.d.ts" />` into a file, but people didn't like doing that.

So if someone could tell me a better way to make this work without a @types/ts-nameof that would be great, but until then I would like to keep this package here.

dsherret (changes requested):
Please close the PR or this needs resolution: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39557#issuecomment-546624899

a-tarasyuk:
cc @sandersn

dsherret:
@a-tarasyuk I think we can close: https://twitter.com/sanders_n/status/1188877386902695945

PR #39558, opened 2019-10-25
use-dark-mode: Provides its own types

a-tarasyuk:
This will deprecate `@types/use-dark-mode` in favor of just `use-dark-mode`. CC @ChaosinaCan


ChaosinaCan:
This looks correct to me, but the tests are complaining because the version in @types is 2.3. I'm guessing I had an older version of use-dark-mode checked out when I wrote the types, so I missed that types already existed in a newer version of it, and then DefinitelyTyped's template generator automatically added the latest version to the index.d.ts.

I don't know how safe it is to ignore the tests. Maybe we can trick it by changing the version in the header of index.d.ts back to 2.1.2 and then removing the file in two separate commits?

PR #39559, opened 2019-10-25
vanilla-tilt: Provides its own types

a-tarasyuk:
This will deprecate `@types/vanilla-tilt` in favor of just `vanilla-tilt`. CC @BrunnerLivio


BrunnerLivio:
@a-tarasyuk LGTM. Though is this the correct way to deprecate the package on DefinitelyTyped?

BrunnerLivio (approved):

PR #39560, opened 2019-10-25
vec2: Provides its own types

a-tarasyuk:
This will deprecate `@types/vec2` in favor of just `vec2`. CC @Nielio


Nielio (changes requested):
`@types/polygon` depends on this package.
You should add a package.json to it, which has `vec2` as a dependency.
See [#removing-a-package](https://github.com/DefinitelyTyped/DefinitelyTyped#removing-a-package)
```json
{
  "private": true,
  "dependencies": {
    "vec2": "^1.6.1"
  }
}
```

a-tarasyuk:
[vec2 1.6.1](https://www.npmjs.com/package/vec2) === [@types/vec2 1.6.1](https://www.npmjs.com/package/@types/vec2)

cc @sandersn

PR #39561, opened 2019-10-25
vexdb: Provides its own types

a-tarasyuk:
This will deprecate `@types/vexdb` in favor of just `vexdb`. CC @MayorMonty


MayorMonty:
Thanks for this, I swore I removed the types already from the registry

PR #39565, opened 2019-10-25
uuidv4: Provides its own types

a-tarasyuk:
This will deprecate `@types/uuidv4` in favor of just `uuidv4`. CC @Hinaser


Hinaser (approved):
Seems to be Okay.

PR #39571, opened 2019-10-25
webmidi: Provides its own types

a-tarasyuk:
This will deprecate `@types/webmidi` in favor of just `webmidi`. CC @lostfictions


lostfictions:
Hey there. I can't really remember when I contributed these typings so I might be wrong, but it looks like they're for the [Web MIDI W3C Draft API](https://www.w3.org/TR/webmidi/) that lives in the browser, not webmidi the NPM package. Looking at the typings provided by the webmidi-the-npm-package, they don't seem to define the browser APIs, so there may not be overlap with this @types/ package and the having the latter on DefinitelyTyped may continue to be of interest to people who aren't using the NPM package anyway.

That said, there's a chance that these APIs have been added to Typescript's lib.dom.d.ts (or one of its siblings) since these typings were contributed, so this package may no longer be necessary anyway? But the draft is not currently [widely supported](https://developer.mozilla.org/en-US/docs/Web/API/MIDIAccess#Browser_compatibility) so I'm not sure that's the case. Could you verify whether they are?

sandersn (changes requested):
I checked https://github.com/microsoft/TSJS-lib-generator/tree/master/inputfiles/idSources.json and https://www.w3.org/TR/webmidi/ isn't in there. So I think these types are still needed.

sandersn:
Thanks @lostfictions for the heads up.

lostfictions:
@sandersn Thanks for checking! (And noted that that's where .d.ts files for DOM APIs are generated from, for next time I need to check.)

Should this PR simply be closed, then? One actionable might be to rename these typings so they're published as something like `@types/webmidi-api` rather than `@types/webmidi` -- that might alleviate the confusion causing folks to associate this with the `webmidi` npm package.

Not sure how DT works for this kind of thing, but could we publish another version of `@types/webmidi` that adds an install-time deprecation message like the following?

> The @types/webmidi package has been renamed @types/webmidi-api. If you're looking for typings for the webmidi NPM package, they're now included with that library!

a-tarasyuk:
@sandersn Is it acceptable to add WebMidi to TSJS-lib-generator and then merge this PR?

sandersn:
I think we should only put webmidi in lib.dom.d.ts if more than one browser supports it. On MDN it looks like just Chrome. @saschanaz you might have a different opinion?

I think it's OK to live with the confusion for now and keep the types here.

saschanaz:
>I think we should only put webmidi in lib.dom.d.ts if more than one browser supports it.

I agree, as it has long been Chrome-only API and Firefox is deferring implementation based on security concerns https://github.com/mozilla/standards-positions/issues/58

PR #39574, opened 2019-10-25
word-wrap: Provides its own types

a-tarasyuk:
This will deprecate `@types/word-wrap` in favor of just `word-wrap`. CC @claasahl


claasahl (approved):
Thanks for highlighting this, @a-tarasyuk

PR #39582, opened 2019-10-25
jss: Provides its own types

a-tarasyuk:
This will deprecate `@types/jss` in favor of just `jss`. CC @appsforartists, @kof, @pelotom, @eps1lon


eps1lon:
Can we combine this with the theming PR so that we avoid publishing inconstistent versions at different times?

Then the travis build should work as well.

PR #39587, opened 2019-10-25
jest-validate: Provides its own types

a-tarasyuk:
This will deprecate `@types/jest-validate` in favor of just `jest-validate`. CC @ikatyang


ikatyang (approved):

ikatyang (changes requested):
Ref: #39666

It seems it's not complete enough to replace the current types with the ones provided by jest itself, `@types/jest-validate` should follow the same status of `@types/jest` I think.

PR #39596, opened 2019-10-25
http-status: Provides its own types

a-tarasyuk:
This will deprecate `@types/http-status` in favor of just `http-status`. CC @misak113


sandersn:
@misak113 the last change to these types was 2017 so I think it's safe to say that the old versions won't need any future fixes.

PR #39604, opened 2019-10-25
getopts: Provides its own types

a-tarasyuk:
This will deprecate `@types/getopts` in favor of just `getopts`. CC @azasypkin


azasypkin (approved):
LGTM, thanks!

PR #39608, opened 2019-10-25
forge-di: Provides its own types

a-tarasyuk:
This will deprecate `@types/forge-di` in favor of just `forge-di`. CC @adamcarr


adamcarr (approved):
lgtm

PR #39622, opened 2019-10-25
emoji-regex: Provides its own types

a-tarasyuk:
This will deprecate `@types/emoji-regex` in favor of just `emoji-regex`. CC @iKBAHT, @ifiokjr


a-tarasyuk:
`emoji-regex` and `@types/emoji-regex` has the same versions `8.0.0`

cc @sandersn

PR #39624, opened 2019-10-25
dom-helpers: Provides its own types

a-tarasyuk:
This will deprecate `@types/dom-helpers` in favor of just `dom-helpers`. CC @mctep


mctep (approved):
👍

PR #39627, opened 2019-10-25
dialogflow: Provides its own types

a-tarasyuk:
This will deprecate `@types/dialogflow` in favor of just `dialogflow`. CC @dyladan, @tcarrio, @khalilchoudhry, @jrthib


tcarrio:
@a-tarasyuk on mobile. Could you provide a reference to the update to nodejs-dialogflow that includes the types? I don't see any types defined in the master branch for that repo.

a-tarasyuk:
cc @sandersn

MagsMagnoli:
@tcarrio the types were introduced here https://github.com/googleapis/nodejs-dialogflow/pull/443/files

dyladan (approved):

elizabeth-dev:
Hi, it seems that Typescript compiler does not recognize proto/proto.d.ts as a definition file while compiling, is it sure that this package is not needed anymore?

tcarrio:
I hadn't transitioned our project over to the protobuf generated types yet so I didn't know whether those worked. I'll take a look at it today as well.

@elizabeth-dev would you be able to provide a sample of your project as well?

elizabeth-dev:
I can't provide full source code nor usage context, but updating @types/dialogflow from 0.11.0 (which worked fine) to 4.0.4 causes the following error at compile time:


> src/fakepath/fakefile.ts:1:29 - error TS7016: Could not find a declaration file for module 'dialogflow'. '/projectpath/node_modules/dialogflow/src/index.js' implicitly has an 'any' type.
> Try `npm install @types/dialogflow` if it exists or add a new declaration (.d.ts) file containing `declare module 'dialogflow';`
> ¹ import * as dialogflow from "dialogflow";


I don't know if we're missing any required TS configuration, this is our tsconfig.json:

```
{
  "compilerOptions": {
    "target": "ES2015",
    "module": "commonjs",
    "declaration": true,
    "outDir": "lib",
    "strict": true,
    "removeComments": true,
    "lib": ["es2016"]
  },
  "compileOnSave": true,
  "include": ["src"],
  "exclude": ["node_modules"]
}
```

sandersn:
dialogflow hasn't added a `"types"` entry in package.json, so the compiler can't find the types. protos/protos.d.ts looks like the only .d.ts file so I think it's the right one to use. The fix should be as simple as a PR to add that line to package.json.

In the meantime, you can go back to `@types/dialogflow@0.11.0`. The types for 0.14  are more namespaced so you'll probably need some work to upgrade anyway.
I think, although I haven't tested this, that you can get started by manually pointing to protos/protos.d.ts:

```ts
import * as dialogflow from "dialogflow/protos/protos"
```

That might satisfy the compiler although it's unlikely to work when you run it.

elizabeth-dev:
Hm, I understand. Idk, I would have no problem opening a PR adding that to the package.json but... I don't know if I have enough knowledge about the library to guarantee that it would work (and it would feel awkward to send a PR modifying one line).

dyladan:
One line PRs are not a problem if it is a necessary line :)

The biggest issue in my mind is that their types won't have the same naming conventions, so there will be some minor implementation burden when you switch over. They may also have omitted the `types` field intentionally to signal to users that the types are not authoritative.

sandersn:
> I don't know if I have enough knowledge about the library to guarantee that it would work

Agreed, their types looked pretty different. That's why I'd recommend testing them with `import * as dialogflow from "dialogflow/protos/protos"` to see if you can pre-upgrade your code to match them. If you can, that should give you confidence that making them widely available is OK.

nightillusions:
Tested it with a new project. The native typing makes no sense. Would be great to revert this pr.

Xzya:
> Tested it with a new project. The native typing makes no sense. Would be great to revert this pr.

Any update on this? The native typings are unusable.

peternann:
@tcarrio - Minimal, from-scratch demonstration of this not working is outlined here:
https://stackoverflow.com/questions/60024886/how-to-use-dialogflow-client-library-with-typescript/61072027#61072027

tcarrio:
It looks as though the `dialogflow` package is undergoing a heavy conversion to TypeScript. If it's not already available, it should be soon.

Reference: https://github.com/googleapis/nodejs-dialogflow

> Version 2.0.0 renames dialogflow to @google-cloud/dialogflow on npm, along with introducing TypeScript types.

While the protobuf types for which this @types package was removed was not a reasonable change for projects using `@types/dialogflow`, it seems that the full TypeScript conversion should allow similar usage as had been done previously. Just taking a look at the `index.ts`, the same classes and types are exported there as is used in our project. [Take a look here](https://github.com/googleapis/nodejs-dialogflow/blob/master/src/index.ts).

Hang tight @peternann, the frustration may be over soon. In the meantime, you can install `types/dialogflow@0.11.0` if you need them. We have not noticed any issue with these types and the upcoming conversion to TypeScript in the official `dialogflow` validates the deprecated status of this one, enough so that I would disadvise reverting this PR until there are known issues with the mainline package post-release.

Furthermore, when that is released, the package will be renamed to `@google-cloud/dialogflow` from the note in the README, as a heads up to anyone checking on this.

dyladan:
@tcarrio If the typescript rewrite will have a new name, that may be a good reason to revert this for people who don't upgrade

tcarrio:
@dyladan that's a good point. They may deprecate the `dialogflow` package and hint towards `@google-cloud/dialogflow` themselves, so it would be a manual process to install the new package then and remove the old `dialogflow` and `@types/dialogflow` packages. Until then, this package would provide the proper typings. :+1:

dyladan:
i'll make a quick PR to revert this

dyladan:
Agh it looks like the current version of the dialogflow package is `1.2.0` but somehow the version of the `@types/dialogflow` package is `4.0.4`. not sure how that slipped through the cracks, but releasing a `1.2.0` version of the types won't help for most people because `npm install @types/dialogflow` will grab the 4.x stub version, not the 1.x working version.

@tcarrio looks like the `@google-cloud/dialogflow` package isn't on npm yet. Any idea what the timeline is on that?

tcarrio:
Oof. I don't think npm allows you to delete packages either, and who knows if the pipeline around this knows how to manage this scenario. Good catch though. We might be able to bump the version to `5.0.0`, to bypass this issue? It's ugly but it would work.

tcarrio:
I have a branch open for this now, we'll see if this works. Worst case, people continue installing the `0.11.0` version :man_shrugging:

tcarrio:
Yeah, the issue is the version referenced in the @types deprecation was [version 4.0.3](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39627/files#diff-b50619b03cddf2414287731414a13d35R739) which isn't at all what the version of dialogflow was.

tcarrio:
@google-cloud/dialogflow is already published, so I would suggest teams looking for types to use the official library.

dcsan:
can someone give examples how to use this?

I just get `any` for all types from visualcode when using the official library.

image

frederickjjoubert:
Any update on this? I'm trying to figure out how to use google dialog flow in an angular project using typescript

tcarrio:
@frederickjjoubert are you using the [new Dialogflow client](https://www.npmjs.com/package/@google-cloud/dialogflow)?

frederickjjoubert:
@tcarrio yes I performed `npm install @google-cloud/dialogflow` into my angular project but all the example code on the github page is written for js, I cannot find any examples of using dialogflow with typescript.

dinhanhthi:
@frederickjjoubert  the same problem from my side.

tcarrio:
@dinhanhthi what was the issue you encountered?

I just pulled down the latest quickstart sample and that works out of the box as a TypeScript module for me, with the CommonJS `require`s converted to ES6 `import`s.

It seems like types are inferred as expected, you can see my sample conversion [here](https://github.com/tcarrio/nodejs-dialogflow-typescript-samples).

dinhanhthi:
@tcarrio Thanks for your reply. In fact, I need a return type for my function (which is `Agent`), the old `@types/dialogflow` has it out of the box but I cannot find the same in `@google-cloud/dialogflow`.

tcarrio:
@dinhanhthi If what you're referring to is the old `Agent`, that's generated from their gRPC definitions now, which is exported under `protos`. What you're looking for is likely `protos.google.cloud.dialogflow.v2.Agent`.

You should be able to access this with

```typescript
import * as dialogflow from "@google-cloud/dialogflow";

type Agent = dialogflow.protos.google.cloud.dialogflow.v2.Agent;
```

This `@types/dialogflow` project was started by my team at Dynatrace and developed entirely outside of Google, so it's not much of a surprise that they didn't maintain backwards compatibility with the community typings.

I hope the code snippet helps :wave:

dinhanhthi:
@tcarrio Thank you so much, it works!

peternann:
Thanks @tcarrio! Now if only Google would document this craziness and preferably supply a Typescript-based example for the API.

This is a very abnormal/hidden way to use typings from a module, which most people will not find:

> type Agent = dialogflow.protos.google.cloud.dialogflow.v2.Agent;

PR #39629, opened 2019-10-25
deep-freeze-es6: Provides its own types

a-tarasyuk:
This will deprecate `@types/deep-freeze-es6` in favor of just `deep-freeze-es6`. CC @mattbishop


mattbishop (approved):
yay!

PR #39637, opened 2019-10-25
consola: Provides its own types

a-tarasyuk:
This will deprecate `@types/consola` in favor of just `consola`. CC @Jungwoo-An


Jungwoo-An (approved):
Thanks! 👍

PR #39648, opened 2019-10-25
bip39: Provides its own types

a-tarasyuk:
This will deprecate `@types/bip39` in favor of just `bip39`. CC @micksatana


micksatana (approved):
Acknowledged.

PR #39650, opened 2019-10-25
bitcoinjs-lib: Provides its own types

a-tarasyuk:
This will deprecate `@types/bitcoinjs-lib` in favor of just `bitcoinjs-lib`. CC @mhegazy, @dlebrecht, @rbuckton, @micksatana, @youssefgh, @kento1218


youssefgh (approved):
LGTM. Thank you

PR #39652, opened 2019-10-25
bezier-easing: Provides its own types

a-tarasyuk:
This will deprecate `@types/bezier-easing` in favor of just `bezier-easing`. CC @ptlis


ptlis (approved):

ptlis:
Awesome, thanks @a-tarasyuk, happy to see this is no longer required.

PR #39655, opened 2019-10-25
baconjs: Provides its own types

a-tarasyuk:
This will deprecate `@types/baconjs` in favor of just `baconjs`. CC @alexander-matsievsky, @gekkio


alexander-matsievsky:
@a-tarasyuk Hi, Alexander! Sorry for missing the notification... Great to see that `baconjs` provides the official typings now!

PR #39659, opened 2019-10-25
arcgis-js-api: Provides its own types

a-tarasyuk:
This will deprecate `@types/arcgis-js-api` in favor of just `arcgis-js-api`. CC @Esri, @bsvensson


bsvensson (changes requested):
While it's true that we (Esri) now provide the types within the npm package, most our users are using our API via CDN (or local download) with the DefinitelyTyped types.

Also, we have currently two main versions (3 and 4), and there is no npm package for version "3". This PR would remove types for those users.

Please close this PR without merging.  Thank you :)

a-tarasyuk:
cc @sandersn

PR #39663, opened 2019-10-25
akumina-core: Provides its own types

a-tarasyuk:
This will deprecate `@types/akumina-core` in favor of just `akumina-core`. CC @akumina, @jasonarden, @MarshHawk


MarshHawk:
akumina-core package will be deprecated... It’s at 4.1 and won’t be updated for 4.5

MarshHawk (changes requested):
Hi a-tarasyuk,

So, we are a small team owning/maintaining akumina-core, and the decision was made to only release the 4.5 types package for now. It was a deliberate decision, knowing that the recommended practice is to include the types with the source package. We can't really deprecate akumina-core because it's still valid for 4.1 users. Perhaps, there are some significant changes down the road for this library; however, for now we are grateful for the opportunity to host our types in DefinetlyTyped. We hope to continue sharing them here as we iteratively iron out our wrinkles, to use an idiom. In short, this pull request is not needed. Thanks.

a-tarasyuk:
@MarshHawk The PR was created based on this list [not-needed-packages](https://gist.github.com/sandersn/d9ca982ff4d66be453d2c6f8f9be7fb9#file-not-needed-packages-txt-L626). Also, the previous versions will be exist 

cc @sandersn

MarshHawk:
@a-tarasyuk looks like that file was generated automatically. In the case, "our version is greater" is intentional

PR #39666, opened 2019-10-26
jest: Provides its own types

a-tarasyuk:
And comment PR: This will deprecate `@types/jest` in favor of just `jest`. CC @NoHomey, @jwbay, @asvetliakov, @alexjoverm, @epicallan, @ikatyang, @wsmd, @JamieMason, @douglasduteil, @ahnpnl, @joshuakgoldberg, @UselessPickles, @r3nya, @hotell, @sebald, @andys8, @antoinebrault, @favna, @gstamac, @ExE-Boss, @quassnoi, @Belco90


favna (changes requested):
Jests' own typings are extremely lacking. in the Jest package it only has:
```
import * as cli from 'jest-cli';
export = cli;
```

And in jest-cli then: https://cdn.jsdelivr.net/npm/jest-cli@24.9.0/build/index.d.ts

As you can see these are only the cli typings and are far, far, faar from enough. Therefore @types/jest cannot be deprecated unless Facebook developers fix up their typings to include all the required data.

ExE-Boss (changes requested):
I agree with @Favna.

a-tarasyuk:
cc @sandersn [Jest](https://gist.github.com/sandersn/d9ca982ff4d66be453d2c6f8f9be7fb9#file-not-needed-packages-txt-L572)

wsmd:
Jest codebase was indeed migrated to typescript, but that doesn't mean that consumers of Jest will no longer need to use `@types/jest` (at least for the time being).

Per https://github.com/facebook/jest/issues/7807

> It is currently a non-goal to distribute TS types for using Jest - that is excellently handled in `@types/jest` already. At some point we might want to distribute that with jest as well - but that's a separate issue.

wsmd (changes requested):
Jest codebase was indeed migrated to typescript, but this migration doesn't mean that it's ready to  replace `@types/jest`. Consumers of Jest still need to use `@types/jest` at least for the time being.

Per https://github.com/facebook/jest/issues/7807

> It is currently a non-goal to distribute TS types for using Jest - that is excellently handled in `@types/jest` already. At some point we might want to distribute that with jest as well - but that's a separate issue.

Favna:
This should be closed as per comments above.

PR #39668, opened 2019-10-26
ladda: Provides its own types

a-tarasyuk:
And comment PR: This will deprecate `@types/ladda` in favor of just `ladda`. CC @dflor003, @leemicw


leemicw:
@a-tarasyuk this seems reasonable to me.  I am not sure I am qualified to make that call.

PR #39670, opened 2019-10-26
rot-js: Provides its own types

a-tarasyuk:
This will deprecate `@types/rot-js` in favor of just `rot-js`. CC @atiaxi


atiaxi (approved):
Last I looked there were some things missing from the official types, but those seem to have been internal so there's no real reason not to switch over.

PR #39671, opened 2019-10-26
fixed memory type, Cannot read anything from any memory type

categoricalcat:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://docs.screeps.com/api/#Creep.memory

### DIFF
--- a/types/screeps/index.d.ts
+++ b/types/screeps/index.d.ts
@@ -2593,7 +2593,7 @@ interface Memory {
     spawns: {[name: string]: SpawnMemory};
 }
 
-interface AnyMemory { [key: string]: any }
+interface AnyMemory { [key: string]: any; }
 interface CreepMemory extends AnyMemory {}
 interface FlagMemory extends AnyMemory {}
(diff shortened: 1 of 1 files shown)


pyrodogg (changes requested):
Hi, 
This has previously been [discussed](https://github.com/screepers/typed-screeps/pull/90) and it's considered better to keep Memory strictly typed by default. [Click here](https://github.com/screepers/typed-screeps#breaking-changes) for more. It should be left to the individual user if they want to un-type memory since it's not possible to override or remove the index signature if defined here.

walze:
> Hi,
> This has previously been [discussed](https://github.com/screepers/typed-screeps/pull/90) and it's considered better to keep Memory strictly typed by default. [Click here](https://github.com/screepers/typed-screeps#breaking-changes) for more. It should be left to the individual user if they want to un-type memory since it's not possible to override or remove the index signature if defined here.

Ah, ok then. Thanks for letting me know, ill just close the PR

PR #39672, opened 2019-10-26
Bugfix: Change the Export Method of UIDGenerator Class from Named-Export to Default-Export - @types/uid-generator

kylechine:
#### Motivation

In the original package (uid-generator), the class is exported by Default-Export:

  `module.exports = UIDGenerator;`

But in the current version of the type definition, the class is exported by Named-Export:

  `export const UIDGenerator: UIDGeneratorConstructor;`

Which leads to a piece of wrong JavaScript code after compilation:

```
  const uid_generator_1 = require("uid-generator");
  const uidgen = new uid_generator_1.UIDGenerator(bitLength);
```

The correct piece should be:

```
  const uid_generator_1 = require("uid-generator");
  const uidgen = new uid_generator_1(bitLength);
```

So I feel it may be necessary to fix this.

#### Fixed List

1. Interface Renaming

```
UIDGenerator            -> UIDGeneratorInstance
UIDGeneratorConstructor -> UIDGeneratorClass
```

In the current version, the name UIDGenerator used as both the name of the interface and the class, which may leads to confusion. So I give them a little bit more clear names.

2. Default-Export UIDGenerator

3. UIDGeneratorInstance.generate(call-back): Error Parameter Added

In the orginal (uid-generator) package, the call-back function also accepts an error parameter. And when error did triggered, uid parameter may be optional.

4. Corresponding updates in test script

### DIFF
--- a/types/uid-generator/index.d.ts
+++ b/types/uid-generator/index.d.ts
@@ -11,7 +11,7 @@ interface UIDGeneratorInstance {
   readonly base: number;
   generateSync(): string;
   generate(): Promise<string>;
-  generate(cb: (error: Error|null, uid?: string) => any): void;
+  generate(cb: (error: Error|null, uid: string) => any): void;
 }
 
 interface UIDGeneratorClass {
--- a/types/uid-generator/index.d.ts
+++ b/types/uid-generator/index.d.ts
@@ -11,7 +11,7 @@ interface UIDGeneratorInstance {
   readonly base: number;
   generateSync(): string;
   generate(): Promise<string>;
-  generate(cb: (error:Error|null, uid: string) => any): void;
+  generate(cb: (error:Error|null, uid?: string) => any): void;
 }
 
 interface UIDGeneratorClass {
--- a/types/uid-generator/index.d.ts
+++ b/types/uid-generator/index.d.ts
@@ -4,7 +4,7 @@
 //                 Kyle Chine <https://github.com/kylechine>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-declare interface UIDGenerator {
+interface UIDGenerator {
   readonly bitSize: number;
   readonly uidLength: number;
   readonly baseEncoding: string;
@@ -14,7 +14,7 @@ declare interface UIDGenerator {
   generate(cb: (uid: string) => any): void;
 }
 
-declare interface UIDGeneratorConstructor {
+interface UIDGeneratorConstructor {
   new (bitSize?: number, baseEncoding?: string): UIDGenerator;
   new (baseEncoding?: string): UIDGenerator;
   readonly BASE16: '0123456789abcdef';
--- a/types/uid-generator/index.d.ts
+++ b/types/uid-generator/index.d.ts
@@ -11,7 +11,7 @@ interface UIDGeneratorInstance {
   readonly base: number;
   generateSync(): string;
   generate(): Promise<string>;
-  generate(cb: (error:Error|null, uid?: string) => any): void;
+  generate(cb: (error: Error|null, uid?: string) => any): void;
 }
 
 interface UIDGeneratorClass {
@@ -28,4 +28,4 @@ interface UIDGeneratorClass {
 
 declare const UIDGenerator: UIDGeneratorClass;
 
-export {UIDGenerator as default, UIDGeneratorClass, UIDGeneratorInstance};
+export { UIDGenerator as default, UIDGeneratorClass, UIDGeneratorInstance };
--- a/types/uid-generator/index.d.ts
+++ b/types/uid-generator/index.d.ts
@@ -1,7 +1,7 @@
 // Type definitions for uid-generator 2.0
 // Project: https://github.com/nwoltman/node-uid-generator
 // Definitions by: TheEmrio <https://github.com/TheEmrio>
-//                 Kyle Chine <https://github.com/kylechine>
+//                 KyleChine <https://github.com/kylechine>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 interface UIDGeneratorInstance {
@@ -28,4 +28,4 @@ interface UIDGeneratorClass {
 
 declare const UIDGenerator: UIDGeneratorClass;
 
-export { UIDGenerator as default, UIDGeneratorClass, UIDGeneratorInstance };
+export {UIDGenerator as default, UIDGeneratorClass, UIDGeneratorInstance};
--- a/types/uid-generator/index.d.ts
+++ b/types/uid-generator/index.d.ts
@@ -4,19 +4,19 @@
 //                 Kyle Chine <https://github.com/kylechine>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-interface UIDGenerator {
+interface UIDGeneratorInstance {
   readonly bitSize: number;
   readonly uidLength: number;
   readonly baseEncoding: string;
   readonly base: number;
   generateSync(): string;
   generate(): Promise<string>;
-  generate(cb: (uid: string) => any): void;
+  generate(cb: (error:Error|null, uid: string) => any): void;
 }
 
-interface UIDGeneratorConstructor {
-  new (bitSize?: number, baseEncoding?: string): UIDGenerator;
-  new (baseEncoding?: string): UIDGenerator;
+interface UIDGeneratorClass {
+  new (bitSize?: number, baseEncoding?: string): UIDGeneratorInstance;
+  new (baseEncoding?: string): UIDGeneratorInstance;
   readonly BASE16: '0123456789abcdef';
   readonly BASE36: '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ';
   readonly BASE58: '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz';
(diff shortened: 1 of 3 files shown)


kylechine:
@typescript-bot I guess this version is working now, am I right?

kylechine:
Sorry for too many commits. This is my first time to use this system. The latest version should stable for review. Thank you!

types/uid-generator/index.d.ts
```diff
@@ -1,21 +1,22 @@
...
-export interface UIDGenerator {
+interface UIDGeneratorInstance {
   readonly bitSize: number;
   readonly uidLength: number;
   readonly baseEncoding: string;
   readonly base: number;
   generateSync(): string;
   generate(): Promise<string>;
-  generate(cb: (uid: string) => any): void;
+  generate(cb: (error: Error|null, uid?: string) => any): void;
```
TheEmrio:
```suggestion
  generate(cb: (error: Error|null, uid: string) => any): void;
```

Why did you put a `?` in front of the `uid` parameter?
When I look at the type definitions for `fs.readFile` for instance, I see
```ts
function readFile(path: PathLike | number, callback: (err: NodeJS.ErrnoException | null, data: Buffer) => void): void;
```
`data` here does not have a `?`. Maybe you should remove it here?

types/uid-generator/tsconfig.json
```diff
@@ -4,6 +4,7 @@
         "lib": [
             "es6"
         ],
+        "target":            "ES2015",
```
TheEmrio:
```suggestion
        "target": "ES2015",
```
Why so many whitespaces here? xD

types/uid-generator/uid-generator-tests.ts
```diff
@@ -1,14 +1,17 @@
...
-const generator = new UIDGenerator(128, 'abc'); // $ExpectType UIDGenerator
+new UIDGenerator('abc'); // $ExpectType UIDGeneratorInstance
+const generator = new UIDGenerator(128, 'abc'); // $ExpectType UIDGeneratorInstance
 
 generator.generateSync(); // $ExpectType string
-generator.generate(uid => {
-  uid; // $ExpectType string
+generator.generate((err, uid?) => {
+  err; // $ExpectType Error | null
+  uid; // $ExpectType string | undefined
```
TheEmrio:
```suggestion
  uid; // $ExpectType string
```
`uid` should, therefore, remain as a `string`

Emrio (changes requested):

types/uid-generator/index.d.ts
kylechine: From <https://github.com/nwoltman/node-uid-generator/blob/master/index.js>, Line 43 - 50:

```javascript
    crypto.randomBytes(this._byteSize, (err, buffer) => {
      if (err) {
        cb(err);
      } else {
        cb(null, bufferToString(buffer, this.baseEncoding, this.uidLength));
      }
    });
  }
```

When some error did triggered, the call-back function will be called as:

`cb(err);` which means the second parameter should be left as `undefined`

types/uid-generator/uid-generator-tests.ts
kylechine: This is the same as <https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39672#discussion_r340930766>, if some error triggered, the parameter uid could be `undefined`.

types/uid-generator/tsconfig.json
kylechine: This can't be true.
I have only committed this Remove-Space modification, then DefinitelyTyped.BenchmarkPR fail?

kylechine:
DefinitelyTyped.BenchmarkPR failure, because in <https://github.com/microsoft/dtslint/blob/master/package.json>, the dependencies section says:

`"typescript": "next"`

But the 'next typescript version' (for now is 3.7.0) is not exists.

This is out of our control.

kylechine:
> DefinitelyTyped.BenchmarkPR failure, because in https://github.com/microsoft/dtslint/blob/master/package.json, the dependencies section says:
> 
> `"typescript": "next"`
> 
> But the 'next typescript version' (for now is 3.7.0) is not exists.
> 
> This is out of our control.

The TypeScript NPM page says <https://www.npmjs.com/package/typescript>: `npm install -g typescript@next` means their nightly builds.

So `npm install -g typescript@next` probably means currently there may something wrong with their nightly builds version -- still out of our control.

types/uid-generator/index.d.ts
TheEmrio: I understand what you try to do, but I think that the convention in typescript is not to put a `?` in callback responses.
This is because the tsc server can't know that `err === null` means `uid` is a string.
For example, in `fs.readFile`, even tho `data` is null if there is an error, the type still doesn't have a `?` 
Additionally, it would be a hassle always to write `uid!` instead of `uid` x)
Typescript types assume that developers know that they should check the `err` parameter before using callback values.

TheEmrio:
> > DefinitelyTyped.BenchmarkPR failure, because in https://github.com/microsoft/dtslint/blob/master/package.json, the dependencies section says:
> > `"typescript": "next"`
> > But the 'next typescript version' (for now is 3.7.0) is not exists.
> > This is out of our control.
> 
> The TypeScript NPM page says https://www.npmjs.com/package/typescript: `npm install -g typescript@next` means their nightly builds.
> 
> So `npm install -g typescript@next` error probably means currently there may something wrong with their nightly builds version -- still out of our control.

Why don't you leave `ES2015`?

kylechine:
> > > DefinitelyTyped.BenchmarkPR failure, because in https://github.com/microsoft/dtslint/blob/master/package.json, the dependencies section says:
> > > `"typescript": "next"`
> > > But the 'next typescript version' (for now is 3.7.0) is not exists.
> > > This is out of our control.
> > 
> > 
> > The TypeScript NPM page says https://www.npmjs.com/package/typescript: `npm install -g typescript@next` means their nightly builds.
> > So `npm install -g typescript@next` error probably means currently there may something wrong with their nightly builds version -- still out of our control.
> 
> Why don't you leave `ES2015`?
> I saw that the build failed, maybe try to put it back, it might be a random error.

The DefinitelyTyped.BenchmarkPR failed after I removed a few spaces between `"target": <spaces> "es6"`, although that version passed all test last time.

This is odd. So I have tried to:

1. Change it to 'es6'
2. Add more spaces
3. Add more spaces with 'es2015' or 'ES2015'
4. And even restore exactly the previous version.

All above did not work.

Then according to the error log:

```
npm ERR! code ETARGET

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/vsts/.npm/_logs/2019-10-31T02_49_15_565Z-debug.log
```

It was triggered because the test script want to install every version of TypeScript from 2.0 to the Next version (means nightly built version).

When it ran to 3.7, because 3.7 is completely deleted (now the nightly built version is 3.8), so error was triggered. It was succeeded because last time 3.7 was there and can be installed.

types/uid-generator/uid-generator-tests.ts
```diff
@@ -1,14 +1,17 @@
...
+import { default as UIDGenerator } from 'uid-generator';
 
-new UIDGenerator('abc'); // $ExpectType UIDGenerator
-const generator = new UIDGenerator(128, 'abc'); // $ExpectType UIDGenerator
+new UIDGenerator('abc'); // $ExpectType UIDGeneratorInstance
+const generator = new UIDGenerator(128, 'abc'); // $ExpectType UIDGeneratorInstance
 
 generator.generateSync(); // $ExpectType string
-generator.generate(uid => {
+generator.generate((err, uid?) => {
```
TheEmrio: Don't want to be picky here but you might have forgotten to remove a `?` from your previous commits

types/uid-generator/tsconfig.json
```diff
@@ -4,6 +4,7 @@
         "lib": [
             "es6"
         ],
+        "target": "es6",
```
TheEmrio: Maybe the tests would pass if we don't specify the `target` parameter at all and let the test script choose x)

Emrio (changes requested):

types/uid-generator/uid-generator-tests.ts
kylechine: Thanks for the reminder. Now it removed.

types/uid-generator/tsconfig.json
kylechine: Tried, unfortunately it doesn't work. I add the target option because of the conflict between CommonJS and ES6 module style. It seems only with this option syntax such as Export as Default could work properly.

kylechine: The bug (test script insists to install TypeScript 3.7 witch does not exist) could be fixed very soon. Because it may cause failures for all PRs with then entire @Types packages.

It most likely be caused by someone in TypeScript team who deleted 3.7 by accident.

kylechine: Someone already created a issue <https://github.com/DefinitelyTyped/DefinitelyTyped/issues/40054>

kylechine: Same bug at: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39973#issuecomment-548502541

TheEmrio: Alright, let's wait then

Emrio (approved):

jessetrinity:
We _did_ cause the error you see with the version, but AFAIK it _should_ be fixed for new runs. Is this one still good to go (it looks like you might have tried to make changes on your end to get it to work)?

TheEmrio:
Yes, I think this PR can be merged. Can you force tests to run again?

jessetrinity:
BenchmarkPR passing isn't a blocker for merges currently and we are having some other issues with it. As long as the other two checks pass you're good.

Issue #37077, opened 2019-07-23
[@types/react-native] SectionList can't specify item/section type


Aure77:
If you know how to fix the issue, make a pull request instead.

- [x] I tried using the `@types/react-native` package and had problems.
- [x] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @alloy, @huhuanming, @iRoachie, @skn0tt, @timwangdev, @kamal, @alexdunne, @swissmanu, @a-tarasyuk, @mvdam, @esemesek, @mrnickel, @souvik-ghosh, @nossbigg, @saranshkataria, @franzmoro, @tykus160, @jakebloom, @ceyhun, @mcmar

I can't specify SectionList item/section type. If I use generic way, i got the following message:

> Expected 0 type arguments, but got 1.ts(2558)

```
<SectionList<MyItem> // not allowed
  renderItem={({ item, index, section }) => ( // item have type "any"
    <Text key={index}>{item}</Text>
  )}
  renderSectionHeader={({ section: { title } }) => ( // section is typed with SectionListData<any> so title as type "any"
    <Text style={{ fontWeight: "bold" }}>{title}</Text>
  )}
  sections={[
    { title: "Title1", data: ["item1", "item2"] },
    { title: "Title2", data: ["item3", "item4"] },
    { title: "Title3", data: ["item5", "item6"] }
  ]}
  keyExtractor={(item, index) => item + index}
/>
```

Generic typing is working on FlatList:
`<FlatList<MyItem> ...`

Skn0tt:
This could be because of this line: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/518db0303d72b139099b7abeca8c1ce09a008b4c/types/react-native/index.d.ts#L9228

Please try changing it locally to see if it helps.

Aure77:
If I remove this line, typescript doesn't recognize SectionList as a component :
> 'SectionList' only refers to a type, but is being used as a value here.ts(2693)

Skn0tt:
Removing this line is like removing it from the exports, this behavior is expected.
Try to change it to the following:

```tsx
interface MyItem {
[...]

PR #39674, opened 2019-10-26
[react-native] Fix SectionList<T> generics (#37077), add new methods (Re-open of #37525)

Skn0tt:
This is a re-open of #37525 (which was a re-open of #37178).

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -8662,15 +8662,6 @@ export interface KeyboardStatic extends NativeEventEmitter {
 export const ART: ARTStatic;
 export type ART = ARTStatic;
 
-export const ImagePickerIOS: ImagePickerIOSStatic;
-export type ImagePickerIOS = ImagePickerIOSStatic;
-
-export const LayoutAnimation: LayoutAnimationStatic;
-export type LayoutAnimation = LayoutAnimationStatic;
-
-export const Systrace: SystraceStatic;
-export type Systrace = SystraceStatic;
-
 //////////// APIS //////////////
 export const ActionSheetIOS: ActionSheetIOSStatic;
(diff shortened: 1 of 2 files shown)


Skn0tt:
Well, here we are again with a failing build. I‘ll look into it ...

types/react-native/index.d.ts
```diff
@@ -8651,8 +8662,14 @@ export interface KeyboardStatic extends NativeEventEmitter {
...
-export const SectionList: SectionListStatic<any>;
-export type SectionList<ItemT> = SectionListStatic<ItemT>;
+export const ImagePickerIOS: ImagePickerIOSStatic;
+export type ImagePickerIOS = ImagePickerIOSStatic;
+
+export const LayoutAnimation: LayoutAnimationStatic;
+export type LayoutAnimation = LayoutAnimationStatic;
+
+export const Systrace: SystraceStatic;
+export type Systrace = SystraceStatic;
```
iRoachie: Why's this adding new constants? This doesn't have anything to do with the PR description. This is why the tests are failing because these are already declared and you're declaring them again here.

https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/603212892#L3867-L3878

iRoachie (changes requested):

types/react-native/index.d.ts
Skn0tt: I have no clue where they're coming from - maybe from some rebase I did? Anyway, I removed them.

alloy:
CI is currently failing like last time, saying that these typings only pass using TS 3.2. I am unsure what change specifically leads to that issue; it would be great if we’d be able to understand that to determine if we really need that and warrant the TS requirement bump.

PR #39677, opened 2019-10-26
promise-fs: Add type definitions.

djcsdy:

### DIFF
--- a/types/promise-fs/index.d.ts
+++ b/types/promise-fs/index.d.ts
@@ -7,15 +7,7 @@
 
 import { PathLike, WriteFileOptions, RmDirAsyncOptions, MakeDirectoryOptions, Dirent, Stats } from "fs";
 
-export {
-    PathLike, StatsBase, Stats, Dirent, FSWatcher, ReadStream, WriteStream, renameSync, truncateSync,
-    ftruncateSync, chownSync, fchownSync, lchownSync, chmodSync, fchmodSync, lchmodSync, statSync,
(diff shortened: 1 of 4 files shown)


types/promise-fs/index.d.ts
```diff
@@ -0,0 +1,333 @@
...
+
+export {
+    PathLike, StatsBase, Stats, Dirent, FSWatcher, ReadStream, WriteStream, renameSync, truncateSync,
+    ftruncateSync, chownSync, fchownSync, lchownSync, chmodSync, fchmodSync, lchmodSync, statSync,
+    fstatSync, lstatSync, linkSync, symlinkSync, readlinkSync, realpathSync, unlinkSync, RmDirOptions,
+    RmDirAsyncOptions, rmdirSync, MakeDirectoryOptions, mkdirSync, mkdtempSync, readdirSync, closeSync,
+    openSync, utimesSync, futimesSync, fsyncSync, writeSync, readSync, readFileSync, WriteFileOptions,
+    writeFileSync, appendFileSync, watchFile, unwatchFile, watch, existsSync, constants, accessSync,
+    createReadStream, createWriteStream, fdatasyncSync, copyFileSync, writevSync
+} from "fs";
```
ExE-Boss: This should just be:
```suggestion
export * from "fs";
```

djcsdy: No it shouldn’t. promise-fs doesn’t re-export the asynchronous functions in fs.

ExE-Boss: Except that `export * from` doesn’t export what is overridden in the current module, and it correctly describes what `promise‑fs` does, which your code doesn’t.

ExE-Boss (approved):

PR #39682, opened 2019-10-26
angular-dynamic-locale: Provides its own types

felixfbecker:
This will deprecate `@types/angular-dynamic-locale` in favor of just `angular-dynamic-locale`. CC @stephenlautier


felixfbecker:
Closing in favor of #39662

PR #39684, opened 2019-10-26
aphrodite: Provides its own types

felixfbecker:
This will deprecate `@types/aphrodite` in favor of just `aphrodite`. CC @asvetliakov


felixfbecker:
Closing in favor of #39660

PR #39686, opened 2019-10-26
autolinker: Provides its own types

felixfbecker:
This will deprecate `@types/autolinker` in favor of just `autolinker`. CC @leonyu


felixfbecker:
Closing in favor of #39658

PR #39688, opened 2019-10-26
badgen: Provides its own types

felixfbecker:
This will deprecate `@types/badgen` in favor of just `badgen`. CC @m5d215


felixfbecker:
Closing in favor of #39656

PR #39689, opened 2019-10-26
baconjs: Provides its own types

felixfbecker:
This will deprecate `@types/baconjs` in favor of just `baconjs`. CC @alexander-matsievsky, @gekkio


felixfbecker:
Closing in favor of #39655

PR #39690, opened 2019-10-26
base-x: Provides its own types

felixfbecker:
This will deprecate `@types/base-x` in favor of just `base-x`. CC @chrootsu


felixfbecker:
Closing in favor of #39654

PR #39691, opened 2019-10-26
bem-cn: Provides its own types

felixfbecker:
This will deprecate `@types/bem-cn` in favor of just `bem-cn`. CC @selkinvitaly


felixfbecker:
Closing in favor of #39653

PR #39692, opened 2019-10-26
bezier-easing: Provides its own types

felixfbecker:
This will deprecate `@types/bezier-easing` in favor of just `bezier-easing`. CC @ptlis


felixfbecker:
Closing in favor of #39652

PR #39693, opened 2019-10-26
bip32: Provides its own types

felixfbecker:
This will deprecate `@types/bip32` in favor of just `bip32`. CC @eduhenke


felixfbecker:
Closing in favor of #39651

PR #39694, opened 2019-10-26
bitcoinjs-lib: Provides its own types

felixfbecker:
This will deprecate `@types/bitcoinjs-lib` in favor of just `bitcoinjs-lib`. CC @mhegazy, @dlebrecht, @rbuckton, @micksatana, @youssefgh, @kento1218


felixfbecker:
Closing in favor of #39650

PR #39695, opened 2019-10-26
blob-util: Provides its own types

felixfbecker:
This will deprecate `@types/blob-util` in favor of just `blob-util`. CC @WorldMaker


WorldMaker (approved):

felixfbecker:
Closing in favor of #39649

PR #39696, opened 2019-10-26
bip39: Provides its own types

felixfbecker:
This will deprecate `@types/bip39` in favor of just `bip39`. CC @micksatana


felixfbecker:
Closing in favor of #39648

PR #39699, opened 2019-10-26
broccoli-plugin: Provides its own types

felixfbecker:
This will deprecate `@types/broccoli-plugin` in favor of just `broccoli-plugin`. CC @dfreeman


felixfbecker:
Closing in favor of #39645

PR #39701, opened 2019-10-26
camaro: Provides its own types

felixfbecker:
This will deprecate `@types/camaro` in favor of just `camaro`. CC @tuananh


felixfbecker:
Closing in favor of #39643

PR #39702, opened 2019-10-26
camljs: Provides its own types

felixfbecker:
This will deprecate `@types/camljs` in favor of just `camljs`. CC @andrei-markeev


felixfbecker:
Closing in favor of #39642

PR #39705, opened 2019-10-26
compare-versions: Provides its own types

felixfbecker:
This will deprecate `@types/compare-versions` in favor of just `compare-versions`. CC @LogvinovLeon


felixfbecker:
Closing in favor of #39639

PR #39706, opened 2019-10-26
commons-validator-js: Provides its own types

felixfbecker:
This will deprecate `@types/commons-validator-js` in favor of just `commons-validator-js`. CC @rtmruczek


felixfbecker:
Closing in favor of #39638

PR #39707, opened 2019-10-26
consola: Provides its own types

felixfbecker:
This will deprecate `@types/consola` in favor of just `consola`. CC @Jungwoo-An


felixfbecker:
Closing in favor of #39637

PR #39708, opened 2019-10-26
composer-concerto: Provides its own types

felixfbecker:
This will deprecate `@types/composer-concerto` in favor of just `composer-concerto`. CC @j4m3sb0mb


felixfbecker:
Closing in favor of #39636

PR #39710, opened 2019-10-26
countup.js: Provides its own types

felixfbecker:
This will deprecate `@types/countup.js` in favor of just `countup.js`. CC @shermendev


felixfbecker:
Closing in favor of #39634

PR #39633, opened 2019-10-25
cropperjs: Provides its own types

a-tarasyuk:
This will deprecate `@types/cropperjs` in favor of just `cropperjs`. CC @stepancar
WARNING: other packages depend on this!

PR #39711, opened 2019-10-26
cropperjs: Provides its own types

felixfbecker:
This will deprecate `@types/cropperjs` in favor of just `cropperjs`. CC @stepancar


felixfbecker:
Closing in favor of #39633

PR #39712, opened 2019-10-26
cson-parser: Provides its own types

felixfbecker:
This will deprecate `@types/cson-parser` in favor of just `cson-parser`. CC @lenovouser


felixfbecker:
Closing in favor of #39632

PR #39714, opened 2019-10-26
cucumber-html-reporter: Provides its own types

felixfbecker:
This will deprecate `@types/cucumber-html-reporter` in favor of just `cucumber-html-reporter`. CC @malavshah2110


felixfbecker:
Closing in favor of #39630

PR #39715, opened 2019-10-26
deep-freeze-es6: Provides its own types

felixfbecker:
This will deprecate `@types/deep-freeze-es6` in favor of just `deep-freeze-es6`. CC @mattbishop


felixfbecker:
Closing in favor of #39629

PR #39716, opened 2019-10-26
delegated-events: Provides its own types

felixfbecker:
This will deprecate `@types/delegated-events` in favor of just `delegated-events`. CC @cwparsons


felixfbecker:
Closing in favor of #39628

PR #39718, opened 2019-10-26
document-promises: Provides its own types

felixfbecker:
This will deprecate `@types/document-promises` in favor of just `document-promises`. CC @NotWoods


felixfbecker:
Closing in favor of #39626

PR #39719, opened 2019-10-26
domelementtype: Provides its own types

felixfbecker:
This will deprecate `@types/domelementtype` in favor of just `domelementtype`. CC @johandavidson


felixfbecker:
Closing in favor of #39625

PR #39720, opened 2019-10-26
dom-helpers: Provides its own types

felixfbecker:
This will deprecate `@types/dom-helpers` in favor of just `dom-helpers`. CC @mctep


felixfbecker:
Closing in favor of #39624

PR #39722, opened 2019-10-26
emoji-regex: Provides its own types

felixfbecker:
This will deprecate `@types/emoji-regex` in favor of just `emoji-regex`. CC @iKBAHT, @ifiokjr


felixfbecker:
Closing in favor of #39622

PR #39723, opened 2019-10-26
ethereumjs-tx: Provides its own types

felixfbecker:
This will deprecate `@types/ethereumjs-tx` in favor of just `ethereumjs-tx`. CC @LogvinovLeon, @dmihal


felixfbecker:
Closing in favor of #39621

PR #39618, opened 2019-10-25
expect: Provides its own types

a-tarasyuk:
This will deprecate `@types/expect` in favor of just `expect`. CC @jmreidy, @merrywhether
WARNING: other packages depend on this!

PR #39726, opened 2019-10-26
expect: Provides its own types

felixfbecker:
This will deprecate `@types/expect` in favor of just `expect`. CC @jmreidy, @merrywhether


felixfbecker:
Closing in favor of #39618

PR #39727, opened 2019-10-26
ex-react-native-i18n: Provides its own types

felixfbecker:
This will deprecate `@types/ex-react-native-i18n` in favor of just `ex-react-native-i18n`. CC @maplerichie


felixfbecker:
Closing in favor of #39616

PR #39728, opened 2019-10-26
expo-localization: Provides its own types

felixfbecker:
This will deprecate `@types/expo-localization` in favor of just `expo-localization`. CC @burtek


felixfbecker:
Closing in favor of #39615

PR #39729, opened 2019-10-26
express-openapi: Provides its own types

felixfbecker:
This will deprecate `@types/express-openapi` in favor of just `express-openapi`. CC @mugeso


felixfbecker:
Closing in favor of #39614

PR #39730, opened 2019-10-26
express-promise-router: Provides its own types

felixfbecker:
This will deprecate `@types/express-promise-router` in favor of just `express-promise-router`. CC @wanganjun


felixfbecker:
Closing in favor of #39613

PR #39731, opened 2019-10-26
express-winston: Provides its own types

felixfbecker:
This will deprecate `@types/express-winston` in favor of just `express-winston`. CC @bricka


felixfbecker:
Closing in favor of #39612

PR #39733, opened 2019-10-26
find-java-home: Provides its own types

felixfbecker:
This will deprecate `@types/find-java-home` in favor of just `find-java-home`. CC @sjx233


felixfbecker:
Closing in favor of #39610

PR #39734, opened 2019-10-26
find-process: Provides its own types

felixfbecker:
This will deprecate `@types/find-process` in favor of just `find-process`. CC @buaban


felixfbecker:
Closing in favor of #39609

PR #39735, opened 2019-10-26
forge-di: Provides its own types

felixfbecker:
This will deprecate `@types/forge-di` in favor of just `forge-di`. CC @adamcarr


felixfbecker:
Closing in favor of #39608

PR #39736, opened 2019-10-26
format-number: Provides its own types

felixfbecker:
This will deprecate `@types/format-number` in favor of just `format-number`. CC @codelovesme


felixfbecker:
Closing in favor of #39607

PR #39737, opened 2019-10-26
fromnow: Provides its own types

felixfbecker:
This will deprecate `@types/fromnow` in favor of just `fromnow`. CC @marinewater


felixfbecker:
Closing in favor of #39606

PR #39738, opened 2019-10-26
ganache-core: Provides its own types

felixfbecker:
This will deprecate `@types/ganache-core` in favor of just `ganache-core`. CC @LogvinovLeon


felixfbecker:
Closing in favor of #39605

PR #39739, opened 2019-10-26
getopts: Provides its own types

felixfbecker:
This will deprecate `@types/getopts` in favor of just `getopts`. CC @azasypkin


felixfbecker:
Closing in favor of #39604

PR #39740, opened 2019-10-26
graphql-query-complexity: Provides its own types

felixfbecker:
This will deprecate `@types/graphql-query-complexity` in favor of just `graphql-query-complexity`. CC @abhikmitra


felixfbecker:
Closing in favor of #39603

PR #39741, opened 2019-10-26
gulp-pug: Provides its own types

felixfbecker:
This will deprecate `@types/gulp-pug` in favor of just `gulp-pug`. CC @remisery


felixfbecker:
Closing in favor of #39602

PR #39742, opened 2019-10-26
gulp-shell: Provides its own types

felixfbecker:
This will deprecate `@types/gulp-shell` in favor of just `gulp-shell`. CC @tkqubo


felixfbecker:
Closing in favor of #39601

PR #39600, opened 2019-10-25
headroom: Provides its own types

a-tarasyuk:
This will deprecate `@types/headroom` in favor of just `headroom`. CC @hakubo
WARNING: our version is greater!

PR #39743, opened 2019-10-26
headroom: Provides its own types

felixfbecker:
This will deprecate `@types/headroom` in favor of just `headroom`. CC @hakubo


felixfbecker:
Closing in favor of #39600

PR #39745, opened 2019-10-26
html-webpack-tags-plugin: Provides its own types

felixfbecker:
This will deprecate `@types/html-webpack-tags-plugin` in favor of just `html-webpack-tags-plugin`. CC @VladimirGrenaderov, @maxbogus


felixfbecker:
Closing in favor of #39598

PR #39746, opened 2019-10-26
http-graceful-shutdown: Provides its own types

felixfbecker:
This will deprecate `@types/http-graceful-shutdown` in favor of just `http-graceful-shutdown`. CC @dlee-nvisia


felixfbecker:
Closing in favor of #39597

PR #39595, opened 2019-10-25
ink: Provides its own types

a-tarasyuk:
This will deprecate `@types/ink` in favor of just `ink`. CC @cprecioso, @jamesadarich
WARNING: other packages depend on this!

PR #39748, opened 2019-10-26
ink: Provides its own types

felixfbecker:
This will deprecate `@types/ink` in favor of just `ink`. CC @cprecioso, @jamesadarich


felixfbecker:
Closing in favor of #39595

PR #39750, opened 2019-10-26
is-mobile: Provides its own types

felixfbecker:
This will deprecate `@types/is-mobile` in favor of just `is-mobile`. CC @LogvinovLeon


felixfbecker:
Closing in favor of #39593

PR #39590, opened 2019-10-25
jest-cli: Provides its own types

a-tarasyuk:
This will deprecate `@types/jest-cli` in favor of just `jest-cli`. CC @lifeiscontent

PR #39754, opened 2019-10-26
jest-cli: Provides its own types

felixfbecker:
This will deprecate `@types/jest-cli` in favor of just `jest-cli`. CC @lifeiscontent


felixfbecker:
Closing in favor of #39590

PR #39589, opened 2019-10-25
jest-diff: Provides its own types

a-tarasyuk:
This will deprecate `@types/jest-diff` in favor of just `jest-diff`. CC @myabc
WARNING: other packages depend on this!

PR #39755, opened 2019-10-26
jest-diff: Provides its own types

felixfbecker:
This will deprecate `@types/jest-diff` in favor of just `jest-diff`. CC @myabc


felixfbecker:
Closing in favor of #39589

PR #39756, opened 2019-10-26
jest-each: Provides its own types

felixfbecker:
This will deprecate `@types/jest-each` in favor of just `jest-each`. CC @theutz, @nickmccurdy


felixfbecker:
Closing in favor of #39588

PR #39757, opened 2019-10-26
jest-validate: Provides its own types

felixfbecker:
This will deprecate `@types/jest-validate` in favor of just `jest-validate`. CC @ikatyang


felixfbecker:
Closing in favor of #39587

PR #39758, opened 2019-10-26
jest-get-type: Provides its own types

felixfbecker:
This will deprecate `@types/jest-get-type` in favor of just `jest-get-type`. CC @myabc


felixfbecker:
Closing in favor of #39586

PR #39760, opened 2019-10-26
json-rules-engine: Provides its own types

felixfbecker:
This will deprecate `@types/json-rules-engine` in favor of just `json-rules-engine`. CC @scottdj92


felixfbecker:
Closing in favor of #39584

PR #39766, opened 2019-10-26
kafkajs: Provides its own types

felixfbecker:
This will deprecate `@types/kafkajs` in favor of just `kafkajs`. CC @michal-b-kaminski


felixfbecker:
Closing in favor of #39446

PR #39451, opened 2019-10-25
koa-better-body: Provides its own types

a-tarasyuk:
This will deprecate `@types/koa-better-body` in favor of just `koa-better-body`. CC @danwbyrne

PR #39770, opened 2019-10-26
koa-better-body: Provides its own types

felixfbecker:
This will deprecate `@types/koa-better-body` in favor of just `koa-better-body`. CC @danwbyrne


felixfbecker:
Closing in favor of #39451

PR #39771, opened 2019-10-26
koa-pug: Provides its own types

felixfbecker:
This will deprecate `@types/koa-pug` in favor of just `koa-pug`. CC @Xstoudi


felixfbecker:
Closing in favor of #39452

PR #39454, opened 2019-10-25
koa-useragent: Provides its own types

a-tarasyuk:
This will deprecate `@types/koa-useragent` in favor of just `koa-useragent`. CC @rvboris

PR #39773, opened 2019-10-26
koa-useragent: Provides its own types

felixfbecker:
This will deprecate `@types/koa-useragent` in favor of just `koa-useragent`. CC @rvboris


felixfbecker:
Closing in favor of #39454

PR #39774, opened 2019-10-26
ladda: Provides its own types

felixfbecker:
This will deprecate `@types/ladda` in favor of just `ladda`. CC @dflor003, @leemicw


felixfbecker:
Closing in favor of #39668

PR #39775, opened 2019-10-26
libra-core: Provides its own types

felixfbecker:
This will deprecate `@types/libra-core` in favor of just `libra-core`. CC @mmsqe, @morgansliman


felixfbecker:
Closing in favor of #39455

PR #39776, opened 2019-10-26
logform: Provides its own types

felixfbecker:
This will deprecate `@types/logform` in favor of just `logform`. CC @DABH


felixfbecker:
Closing in favor of #39456

PR #39778, opened 2019-10-26
looks-same: Provides its own types

felixfbecker:
This will deprecate `@types/looks-same` in favor of just `looks-same`. CC @xcatliu


felixfbecker:
Closing in favor of #39669

PR #39779, opened 2019-10-26
mali-compose: Provides its own types

felixfbecker:
This will deprecate `@types/mali-compose` in favor of just `mali-compose`. CC @danwbyrne


felixfbecker:
Closing in favor of #39458

PR #39780, opened 2019-10-26
mali-onerror: Provides its own types

felixfbecker:
This will deprecate `@types/mali-onerror` in favor of just `mali-onerror`. CC @danwbyrne


felixfbecker:
Closing in favor of #39459

PR #39460, opened 2019-10-25
material-components-web: Provides its own types

a-tarasyuk:
This will deprecate `@types/material-components-web` in favor of just `material-components-web`. CC @BrentDouglas, @ckosti, @agroupp

PR #39781, opened 2019-10-26
material-components-web: Provides its own types

felixfbecker:
This will deprecate `@types/material-components-web` in favor of just `material-components-web`. CC @BrentDouglas, @ckosti, @agroupp


felixfbecker:
Closing in favor of #39460

PR #39782, opened 2019-10-26
memoize-one: Provides its own types

felixfbecker:
This will deprecate `@types/memoize-one` in favor of just `memoize-one`. CC @karol-majewski, @franklixuefei


felixfbecker:
Closing in favor of #39461

PR #39785, opened 2019-10-26
mobx-devtools-mst: Provides its own types

felixfbecker:
This will deprecate `@types/mobx-devtools-mst` in favor of just `mobx-devtools-mst`. CC @pluma


felixfbecker:
Closing in favor of #39464

PR #39786, opened 2019-10-26
mobx-task: Provides its own types

felixfbecker:
This will deprecate `@types/mobx-task` in favor of just `mobx-task`. CC @JulianWielga


felixfbecker:
Closing in favor of #39465

PR #39788, opened 2019-10-26
moment-range: Provides its own types

felixfbecker:
This will deprecate `@types/moment-range` in favor of just `moment-range`. CC @Burgov, @wilgert, @franjuan, @MartynasZilinskas, @chemass


felixfbecker:
Closing in favor of #39468

PR #39790, opened 2019-10-26
msportalfx-test: Provides its own types

felixfbecker:
This will deprecate `@types/msportalfx-test` in favor of just `msportalfx-test`. CC @julioct


felixfbecker:
Closing in favor of #39470

PR #39791, opened 2019-10-26
morphdom: Provides its own types

felixfbecker:
This will deprecate `@types/morphdom` in favor of just `morphdom`. CC @arvitaly


felixfbecker:
Closing in favor of #39471

PR #39792, opened 2019-10-26
natsort: Provides its own types

felixfbecker:
This will deprecate `@types/natsort` in favor of just `natsort`. CC @mgroenhoff


felixfbecker:
Closing in favor of #39472

PR #39473, opened 2019-10-25
navigation: Provides its own types

a-tarasyuk:
This will deprecate `@types/navigation` in favor of just `navigation`. CC @grahammendick
WARNING: other packages depend on this!

PR #39793, opened 2019-10-26
navigation: Provides its own types

felixfbecker:
This will deprecate `@types/navigation` in favor of just `navigation`. CC @grahammendick


felixfbecker:
Closing in favor of #39473

PR #39794, opened 2019-10-26
nblas: Provides its own types

felixfbecker:
This will deprecate `@types/nblas` in favor of just `nblas`. CC @erikgerrits


felixfbecker:
Closing in favor of #39474

PR #39795, opened 2019-10-26
navigation-react: Provides its own types

felixfbecker:
This will deprecate `@types/navigation-react` in favor of just `navigation-react`. CC @grahammendick


felixfbecker:
Closing in favor of #39475

PR #39796, opened 2019-10-26
neffos.js: Provides its own types

felixfbecker:
This will deprecate `@types/neffos.js` in favor of just `neffos.js`. CC @kataras


felixfbecker:
Closing in favor of #39476

PR #39797, opened 2019-10-26
nats-hemera: Provides its own types

felixfbecker:
This will deprecate `@types/nats-hemera` in favor of just `nats-hemera`. CC @vforv


felixfbecker:
Closing in favor of #39477

PR #39478, opened 2019-10-25
next-redux-wrapper: Provides its own types

a-tarasyuk:
This will deprecate `@types/next-redux-wrapper` in favor of just `next-redux-wrapper`. CC @stevegeek, @Jungwoo-An
WARNING: other packages depend on this!

PR #39798, opened 2019-10-26
next-redux-wrapper: Provides its own types

felixfbecker:
This will deprecate `@types/next-redux-wrapper` in favor of just `next-redux-wrapper`. CC @stevegeek, @Jungwoo-An


felixfbecker:
Closing in favor of #39478

PR #39799, opened 2019-10-26
node-cache: Provides its own types

felixfbecker:
This will deprecate `@types/node-cache` in favor of just `node-cache`. CC @chrootsu, @dthunell, @useltmann


felixfbecker:
Closing in favor of #39479

PR #39480, opened 2019-10-25
next: Provides its own types

a-tarasyuk:
This will deprecate `@types/next` in favor of just `next`. CC @dru89, @brikou, @jthegedus, @resir014, @scottdj92, @joaovieira, @ajliv
WARNING: other packages depend on this!

PR #39800, opened 2019-10-26
next: Provides its own types

felixfbecker:
This will deprecate `@types/next` in favor of just `next`. CC @dru89, @brikou, @jthegedus, @resir014, @scottdj92, @joaovieira, @ajliv


felixfbecker:
Closing in favor of #39480

PR #39802, opened 2019-10-26
node-json-db: Provides its own types

felixfbecker:
This will deprecate `@types/node-json-db` in favor of just `node-json-db`. CC @kuzn-ilya


felixfbecker:
Closing in favor of #39482

PR #39803, opened 2019-10-26
nookies: Provides its own types

felixfbecker:
This will deprecate `@types/nookies` in favor of just `nookies`. CC @andreasbergqvist


felixfbecker:
Closing in favor of #39483

PR #39804, opened 2019-10-26
notyf: Provides its own types

felixfbecker:
This will deprecate `@types/notyf` in favor of just `notyf`. CC @gurov


felixfbecker:
Closing in favor of #39484

PR #39806, opened 2019-10-26
onoff: Provides its own types

felixfbecker:
This will deprecate `@types/onoff` in favor of just `onoff`. CC @marcel-ernst, @Kallu609


felixfbecker:
Closing in favor of #39487

PR #39807, opened 2019-10-26
odata: Provides its own types

felixfbecker:
This will deprecate `@types/odata` in favor of just `odata`. CC @janhommes, @jcchalte


felixfbecker:
Closing in favor of #39488

PR #39809, opened 2019-10-26
pad: Provides its own types

felixfbecker:
This will deprecate `@types/pad` in favor of just `pad`. CC @mhegazy


felixfbecker:
Closing in favor of #39490

PR #39491, opened 2019-10-25
passport-unique-token: Provides its own types

a-tarasyuk:
This will deprecate `@types/passport-unique-token` in favor of just `passport-unique-token`. CC @briman0094, @SomaticIT

PR #39810, opened 2019-10-26
passport-unique-token: Provides its own types

felixfbecker:
This will deprecate `@types/passport-unique-token` in favor of just `passport-unique-token`. CC @briman0094, @SomaticIT


felixfbecker:
Closing in favor of #39491

PR #39811, opened 2019-10-26
peerjs: Provides its own types

felixfbecker:
This will deprecate `@types/peerjs` in favor of just `peerjs`. CC @nakakura


felixfbecker:
Closing in favor of #39492

PR #39812, opened 2019-10-26
phin: Provides its own types

felixfbecker:
This will deprecate `@types/phin` in favor of just `phin`. CC @twooster


felixfbecker:
Closing in favor of #39493

PR #39813, opened 2019-10-26
pixi.js: Provides its own types

felixfbecker:
This will deprecate `@types/pixi.js` in favor of just `pixi.js`. CC @clark-stevenson


felixfbecker:
Closing in favor of #39494

PR #39814, opened 2019-10-26
postmark: Provides its own types

felixfbecker:
This will deprecate `@types/postmark` in favor of just `postmark`. CC @benbayard, @jineshshah36


felixfbecker:
Closing in favor of #39495

PR #39815, opened 2019-10-26
printf: Provides its own types

felixfbecker:
This will deprecate `@types/printf` in favor of just `printf`. CC @AluisioASG


felixfbecker:
Closing in favor of #39496

PR #39817, opened 2019-10-26
qunit-dom: Provides its own types

felixfbecker:
This will deprecate `@types/qunit-dom` in favor of just `qunit-dom`. CC @happycollision


felixfbecker:
Closing in favor of #39498

PR #39818, opened 2019-10-26
random-js: Provides its own types

felixfbecker:
This will deprecate `@types/random-js` in favor of just `random-js`. CC @pistacchio


felixfbecker:
Closing in favor of #39499

PR #39819, opened 2019-10-26
ractive: Provides its own types

felixfbecker:
This will deprecate `@types/ractive` in favor of just `ractive`. CC Han Lin Yap (http://yap.nu)


felixfbecker:
Closing in favor of #39500

PR #39821, opened 2019-10-26
rdflib: Provides its own types

felixfbecker:
This will deprecate `@types/rdflib` in favor of just `rdflib`. CC @cenotelie


felixfbecker:
Closing in favor of #39502

PR #39823, opened 2019-10-26
react-circular-progressbar: Provides its own types

felixfbecker:
This will deprecate `@types/react-circular-progressbar` in favor of just `react-circular-progressbar`. CC @lstanden


felixfbecker:
Closing in favor of #39504

PR #39824, opened 2019-10-26
react-collapsible: Provides its own types

felixfbecker:
This will deprecate `@types/react-collapsible` in favor of just `react-collapsible`. CC @knegusen


felixfbecker:
Closing in favor of #39505

PR #39825, opened 2019-10-26
react-content-loader: Provides its own types

felixfbecker:
This will deprecate `@types/react-content-loader` in favor of just `react-content-loader`. CC @alaatm, @samwalshnz


felixfbecker:
Closing in favor of #39506

PR #39826, opened 2019-10-26
react-daum-postcode: Provides its own types

felixfbecker:
This will deprecate `@types/react-daum-postcode` in favor of just `react-daum-postcode`. CC @Sa-ryong


felixfbecker:
Closing in favor of #39507

PR #39829, opened 2019-10-26
react-hot-loader: Provides its own types

felixfbecker:
This will deprecate `@types/react-hot-loader` in favor of just `react-hot-loader`. CC @jacekjagiello, @MartynasZilinskas, @DovydasNavickas


felixfbecker:
Closing in favor of #39510

PR #39831, opened 2019-10-26
react-json-pretty: Provides its own types

felixfbecker:
This will deprecate `@types/react-json-pretty` in favor of just `react-json-pretty`. CC @LKay


felixfbecker:
Closing in favor of #39512

PR #39833, opened 2019-10-26
react-jss: Provides its own types

felixfbecker:
This will deprecate `@types/react-jss` in favor of just `react-jss`. CC @eps1lon, @jlaw90


felixfbecker:
Closing in favor of #39514

PR #39836, opened 2019-10-26
react-native-keychain: Provides its own types

felixfbecker:
This will deprecate `@types/react-native-keychain` in favor of just `react-native-keychain`. CC @DavidFarinha


felixfbecker:
Closing in favor of #39517

PR #39837, opened 2019-10-26
react-native-mauron85-background-geolocation: Provides its own types

felixfbecker:
This will deprecate `@types/react-native-mauron85-background-geolocation` in favor of just `react-native-mauron85-background-geolocation`. CC @djereg


felixfbecker:
Closing in favor of #39518

PR #39839, opened 2019-10-26
react-native-safe-area: Provides its own types

felixfbecker:
This will deprecate `@types/react-native-safe-area` in favor of just `react-native-safe-area`. CC @pvinis


felixfbecker:
Closing in favor of #39520

PR #39843, opened 2019-10-26
react-native-swipe-gestures: Provides its own types

felixfbecker:
This will deprecate `@types/react-native-swipe-gestures` in favor of just `react-native-swipe-gestures`. CC @ian-rudge


felixfbecker:
Closing in favor of #39524

PR #39845, opened 2019-10-26
react-sortable-hoc: Provides its own types

felixfbecker:
This will deprecate `@types/react-sortable-hoc` in favor of just `react-sortable-hoc`. CC @NoHomey, @charlesrey


felixfbecker:
Closing in favor of #39526

PR #39846, opened 2019-10-26
react-sortable-pane: Provides its own types

felixfbecker:
This will deprecate `@types/react-sortable-pane` in favor of just `react-sortable-pane`. CC @rhysd


felixfbecker:
Closing in favor of #39527

PR #39847, opened 2019-10-26
react-sticky-box: Provides its own types

felixfbecker:
This will deprecate `@types/react-sticky-box` in favor of just `react-sticky-box`. CC @koss-lebedev


felixfbecker:
Closing in favor of #39528

PR #39849, opened 2019-10-26
react-tether: Provides its own types

felixfbecker:
This will deprecate `@types/react-tether` in favor of just `react-tether`. CC @ryprice


felixfbecker:
Closing in favor of #39530

PR #39850, opened 2019-10-26
react-toastify: Provides its own types

felixfbecker:
This will deprecate `@types/react-toastify` in favor of just `react-toastify`. CC @icopp


felixfbecker:
Closing in favor of #39531

PR #39852, opened 2019-10-26
reduce-reducers: Provides its own types

felixfbecker:
This will deprecate `@types/reduce-reducers` in favor of just `reduce-reducers`. CC @huy-nguyen, @daliusd


felixfbecker:
Closing in favor of #39533

PR #39853, opened 2019-10-26
redom: Provides its own types

felixfbecker:
This will deprecate `@types/redom` in favor of just `redom`. CC @RauliL, @nemoinho


felixfbecker:
Closing in favor of #39534

PR #39854, opened 2019-10-26
redux-promise-middleware: Provides its own types

felixfbecker:
This will deprecate `@types/redux-promise-middleware` in favor of just `redux-promise-middleware`. CC @ianks


felixfbecker:
Closing in favor of #39535

PR #39855, opened 2019-10-26
rfc4648: Provides its own types

felixfbecker:
This will deprecate `@types/rfc4648` in favor of just `rfc4648`. CC @thibmeu


felixfbecker:
Closing in favor of #39536

PR #39857, opened 2019-10-26
rot-js: Provides its own types

felixfbecker:
This will deprecate `@types/rot-js` in favor of just `rot-js`. CC @atiaxi


felixfbecker:
Closing in favor of #39670

PR #39858, opened 2019-10-26
rsmq: Provides its own types

felixfbecker:
This will deprecate `@types/rsmq` in favor of just `rsmq`. CC @MugeSo


felixfbecker:
Closing in favor of #39538

PR #39859, opened 2019-10-26
sass-webpack-plugin: Provides its own types

felixfbecker:
This will deprecate `@types/sass-webpack-plugin` in favor of just `sass-webpack-plugin`. CC @AepKill


felixfbecker:
Closing in favor of #39540

PR #39860, opened 2019-10-26
sauronjs: Provides its own types

felixfbecker:
This will deprecate `@types/sauronjs` in favor of just `sauronjs`. CC @EricPMulligan


felixfbecker:
Closing in favor of #39541

PR #39861, opened 2019-10-26
schema-utils: Provides its own types

felixfbecker:
This will deprecate `@types/schema-utils` in favor of just `schema-utils`. CC @yavanosta


felixfbecker:
Closing in favor of #39542

PR #39544, opened 2019-10-25
skatejs: Provides its own types

a-tarasyuk:
This will deprecate `@types/skatejs` in favor of just `skatejs`. CC @Hotell
WARNING: our version is greater!

PR #39863, opened 2019-10-26
skatejs: Provides its own types

felixfbecker:
This will deprecate `@types/skatejs` in favor of just `skatejs`. CC @Hotell


felixfbecker:
Closing in favor of #39544

PR #39545, opened 2019-10-25
snowboy: Provides its own types

a-tarasyuk:
This will deprecate `@types/snowboy` in favor of just `snowboy`. CC @dolanmiu

PR #39864, opened 2019-10-26
snowboy: Provides its own types

felixfbecker:
This will deprecate `@types/snowboy` in favor of just `snowboy`. CC @dolanmiu


felixfbecker:
Closing in favor of #39545

PR #39866, opened 2019-10-26
stacktrace-js: Provides its own types

felixfbecker:
This will deprecate `@types/stacktrace-js` in favor of just `stacktrace-js`. CC @exceptionless, @pilagod


felixfbecker:
Closing in favor of #39547

PR #39870, opened 2019-10-26
string-argv: Provides its own types

felixfbecker:
This will deprecate `@types/string-argv` in favor of just `string-argv`. CC @vladimir-tikhonov


felixfbecker:
Closing in favor of #39551

PR #39872, opened 2019-10-26
systeminformation: Provides its own types

felixfbecker:
This will deprecate `@types/systeminformation` in favor of just `systeminformation`. CC @PixelcrabAT


felixfbecker:
Closing in favor of #39553

PR #39873, opened 2019-10-26
text-clipper: Provides its own types

felixfbecker:
This will deprecate `@types/text-clipper` in favor of just `text-clipper`. CC @krzysztof-grzybek


felixfbecker:
Closing in favor of #39554

PR #39874, opened 2019-10-26
theming: Provides its own types

felixfbecker:
This will deprecate `@types/theming` in favor of just `theming`. CC @eps1lon


felixfbecker:
Closing in favor of #39555

PR #39877, opened 2019-10-26
use-dark-mode: Provides its own types

felixfbecker:
This will deprecate `@types/use-dark-mode` in favor of just `use-dark-mode`. CC @ChaosinaCan


felixfbecker:
Closing in favor of #39558

PR #39879, opened 2019-10-26
vec2: Provides its own types

felixfbecker:
This will deprecate `@types/vec2` in favor of just `vec2`. CC @Nielio


felixfbecker:
Closing in favor of #39560

PR #39880, opened 2019-10-26
vexdb: Provides its own types

felixfbecker:
This will deprecate `@types/vexdb` in favor of just `vexdb`. CC @MayorMonty


felixfbecker:
Closing in favor of #39561

PR #39881, opened 2019-10-26
vfile: Provides its own types

felixfbecker:
This will deprecate `@types/vfile` in favor of just `vfile`. CC @bizen241, @rokt33r


felixfbecker:
Closing in favor of #39562

PR #39667, opened 2019-10-26
vfile-message: Provides its own types

a-tarasyuk:
And comment PR: This will deprecate `@types/vfile-message` in favor of just `vfile-message`. CC @rokt33r
WARNING: other packages depend on this!

PR #39882, opened 2019-10-26
vfile-message: Provides its own types

felixfbecker:
This will deprecate `@types/vfile-message` in favor of just `vfile-message`. CC @rokt33r


felixfbecker:
Closing in favor of #39667

PR #39883, opened 2019-10-26
velocity-animate: Provides its own types

felixfbecker:
This will deprecate `@types/velocity-animate` in favor of just `velocity-animate`. CC @smrq


felixfbecker:
Closing in favor of #39563

PR #39884, opened 2019-10-26
viewerjs: Provides its own types

felixfbecker:
This will deprecate `@types/viewerjs` in favor of just `viewerjs`. CC @lrh3321


felixfbecker:
Closing in favor of #39564

PR #39885, opened 2019-10-26
uuidv4: Provides its own types

felixfbecker:
This will deprecate `@types/uuidv4` in favor of just `uuidv4`. CC @Hinaser


felixfbecker:
Closing in favor of #39565

PR #39887, opened 2019-10-26
vue-resource: Provides its own types

felixfbecker:
This will deprecate `@types/vue-resource` in favor of just `vue-resource`. CC @kaorun343


felixfbecker:
Closing in favor of #39567

PR #39889, opened 2019-10-26
web3-eth-abi: Provides its own types

felixfbecker:
This will deprecate `@types/web3-eth-abi` in favor of just `web3-eth-abi`. CC @LogvinovLeon


felixfbecker:
Closing in favor of #39569

PR #39891, opened 2019-10-26
webmidi: Provides its own types

felixfbecker:
This will deprecate `@types/webmidi` in favor of just `webmidi`. CC @lostfictions


felixfbecker:
Closing in favor of #39571

PR #39892, opened 2019-10-26
which-pm: Provides its own types

felixfbecker:
This will deprecate `@types/which-pm` in favor of just `which-pm`. CC @jorgegonzalez


felixfbecker:
Closing in favor of #39572

PR #39894, opened 2019-10-26
word-wrap: Provides its own types

felixfbecker:
This will deprecate `@types/word-wrap` in favor of just `word-wrap`. CC @claasahl


felixfbecker:
Closing in favor of #39574

PR #39895, opened 2019-10-26
wouter: Provides its own types

felixfbecker:
This will deprecate `@types/wouter` in favor of just `wouter`. CC @StrayFromThePath


felixfbecker:
Closing in favor of #39575

PR #39897, opened 2019-10-26
zipkin-transport-http: Provides its own types

felixfbecker:
This will deprecate `@types/zipkin-transport-http` in favor of just `zipkin-transport-http`. CC @plantain-00


felixfbecker:
Closing in favor of #39577

PR #39898, opened 2019-10-26
zipkin-instrumentation-express: Provides its own types

felixfbecker:
This will deprecate `@types/zipkin-instrumentation-express` in favor of just `zipkin-instrumentation-express`. CC @plantain-00


felixfbecker:
Closing in favor of #39445

PR #39250, opened 2019-10-18
feat(node): v12.12

SimonSchick:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#12.12.0
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

Didn't go heavy on the docs.

PR #39914, opened 2019-10-27
feat(node): v12.12

SimonSchick:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#12.12.0

Repeat of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39250
Also cleaned up some incorrectly declared exports.

### DIFF
--- a/types/node/index.d.ts
+++ b/types/node/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for non-npm package Node.js 12.11
+// Type definitions for non-npm package Node.js 12.12
 // Project: http://nodejs.org/
 // Definitions by: Microsoft TypeScript <https://github.com/Microsoft>
 //                 DefinitelyTyped <https://github.com/DefinitelyTyped>
--- a/types/node/crypto.d.ts
+++ b/types/node/crypto.d.ts
@@ -141,7 +141,7 @@ declare module "crypto" {
         digest(encoding: HexBase64Latin1Encoding): string;
     }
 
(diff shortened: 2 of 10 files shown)


types/node/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for non-npm package Node.js 12.11
+// Type definitions for non-npm package Node.js 12.12
```
galkin: should we use 12.13 here?

SimonSchick: 12.13 didn't really introduce anything new.

galkin: So, let's use 12.13 here. Could we?

Flarna (approved):

types/node/index.d.ts
SimonSchick: There is no advantage in doing so and would just trigger another build and delay things further, I will leave as is *unless* another change will be made.

trivikr (approved):

ddehghan:
Hi guys - we just started getting this error in our CI:

 ../node_modules/@grpc/grpc-js/build/src/metadata.d.ts(71,29): error TS2694: Namespace '"http2"' has no exported member 'OutgoingHttpHeaders'.

seems that it is a regression.

Flarna:
@ddehghan PR to fix this: #39992

PR #39923, opened 2019-10-28
Prepare for TS #34742

ahejlsberg:
This PR prepares for microsoft/TypeScript#34742 by making a couple of changes to `mergerino`. The `FunctionPatch<T>` type now excludes `undefined` from `T` when it is passed as the first parameter, and a test with seven levels of array nesting is reduced to four such that it is fully checked with older versions of the TypeScript compiler.


ahejlsberg:
Ugh, started on an old branch. Abandoning this PR.

PR #39924, opened 2019-10-28
Fix mergerino types and tests in preparation for TS #34742

ahejlsberg:
This PR prepares for microsoft/TypeScript#34742 by making a couple of changes to `mergerino`. The `FunctionPatch<T>` type now excludes `undefined` from `T` when it is passed as the first parameter, and a test with seven levels of array nesting is reduced to four such that it is fully checked with older versions of the TypeScript compiler.

### DIFF
--- a/types/mergerino/index.d.ts
+++ b/types/mergerino/index.d.ts
@@ -26,7 +26,7 @@ export type DeletePatch = undefined;
  * and the merge function as the second. The return value will be the replacement.
  * The value you return will bypass merging logic and simply overwrite the property.
  */
-export type FunctionPatch<T> = (val: Exclude<T, undefined>, merge: Merge<T extends object ? T : never>) => T;
+export type FunctionPatch<T> = (val: T, merge: Merge<T extends object ? T : never>) => T;
 
 /**
  * If you want to replace a array specify new array as the value.
--- a/types/mergerino/index.d.ts
+++ b/types/mergerino/index.d.ts
@@ -26,7 +26,7 @@ export type DeletePatch = undefined;
  * and the merge function as the second. The return value will be the replacement.
  * The value you return will bypass merging logic and simply overwrite the property.
  */
-export type FunctionPatch<T> = (val: T, merge: Merge<T extends object ? T : never>) => T;
+export type FunctionPatch<T> = (val: Exclude<T, undefined>, merge: Merge<T extends object ? T : never>) => T;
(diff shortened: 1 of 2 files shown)


fivitti:
> The ```FunctionPatch<T>``` type now excludes undefined from T when it is passed as the first parameter

I think that it is wrong, because you can patch an optional property using function patch and in this case the value will ```undefined```.

For example:
```typescript
interface Foo {
    bar: number,
    baz?: number
}

const item: Foo = {
    bar: 1
};

const out = merge(item, { baz: (bazValue) => 42 });
```

It is correct mergerino code and in this example ```bazValue``` is ```undefined```.
We need to keep it as is, without exclude ```undefined```.

types/mergerino/index.d.ts
```diff
@@ -26,7 +26,7 @@ export type DeletePatch = undefined;
  * and the merge function as the second. The return value will be the replacement.
  * The value you return will bypass merging logic and simply overwrite the property.
  */
-export type FunctionPatch<T> = (val: T, merge: Merge<T extends object ? T : never>) => T;
+export type FunctionPatch<T> = (val: Exclude<T, undefined>, merge: Merge<T extends object ? T : never>) => T;
```
fivitti: I think that it is wrong, because you can patch an optional property using function patch and in this case the value will ```undefined```.

For example:
```typescript
interface Foo {
    bar: number,
    baz?: number
}

const item: Foo = {
    bar: 1
};

const out = merge(item, { baz: (bazValue) => 42 });
```

It is correct mergerino code and in this example ```bazValue``` is ```undefined```.
We need to keep it as is, without exclude ```undefined```.

fivitti (changes requested):

fivitti (changes requested):
Adding ```Exclude<T, undefined>``` is incompatible with ```mergerino```.
It can cause silence null checks.

For example:
```typescript
interface Foo {
    bar: number,
    baz?: number
}

const item: Foo = {
    bar: 1
};

const out = merge(item, { baz: (bazValue) => bazValue + 42 }); // baz is set to NaN
```
If we exclude ```undefined``` user with enable **strictNullChecks** don't get any information about addition number to ```undefined```.

ahejlsberg:
Ok, the test didn't include undefined so I assumed it would be filtered out. I have fixed the test instead.

PR #39938, opened 2019-10-28
added "alloy" typings.

ffMathy:


jessetrinity:
@ffMathy you are modifying existing typings so you should seek sign off from the owners
@appcelerator
@janvennemann

types/alloy/index.d.ts
```diff
@@ -0,0 +1,195 @@
+// Type definitions for npm package Alloy 1.14
+// Project: https://github.com/appcelerator/alloy
+// Definitions by: Axway Appcelerator <https://github.com/appcelerator>
+//                 Mathias Lorenzen <https://github.com/ffMathy>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.6
```
jessetrinity: Be aware that this means that TS3.6 will be the _min_ required version to use this package

types/alloy/tslint.json
```diff
@@ -0,0 +1,13 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
```
jessetrinity: Please do not disable lint rules in new definitions.

types/alloy/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "dom",
+            "es6"
+        ],
+        "noImplicitAny": true,
+        "noImplicitThis": false,
```
jessetrinity: [These should all be set to true](https://github.com/definitelytyped/definitelytyped#some-packages-have-no-tslintjson-and-some-tsconfigjson-are-missing-noimplicitany-true-noimplicitthis-true-or-strictnullchecks-true)

jessetrinity (changes requested):

types/alloy/index.d.ts
ffMathy: What version do you recommend?

types/alloy/tsconfig.json
ffMathy: Got it - copy-pasted from an existing definition. That was probably a mistake, as I can see now from the docs. Sorry!

types/alloy/tslint.json
ffMathy: Got it :+1:

types/alloy/index.d.ts
ffMathy: I settled for `3.0`. I assume that's better?

ffMathy:
@jessetrinity I made the corrections you requested.

ffMathy:
Hmmm the current build failure did not occur before, despite that part of the code never changing. Is it new?

How am I supposed to extend the interface of another type now?

types/alloy/index.d.ts
jessetrinity: It does mean more people might use it. I can't say how likely that is so I hope I'm not coming off as nit picky.

I think `3.1` had some big breaking changes so that seems like a pretty good general older supported version.

janvennemann:
@ffMathy these typings won't work in the Titanium environment due to the `jquery` and `backbone` references. Titanium creates its own JavaScript environment which is not compatible with a browser environment. You should see a lot of errors like this:

```
node_modules/@types/backbone/index.d.ts:502:26 - error TS2304: Cannot find name 'Window'.

502         getHash(window?: Window): string;
```

These could be resolved by adding the "dom" lib, but this is not recommended since it undermines type safety.

ffMathy:
But if I had to make a choice between any casts everywhere or add that lib, I would choose the latter.

Is there another way to get around it?

janvennemann:
As already mentioned in https://github.com/appcelerator/titanium-boilerplates/pull/5 the issues start with the `backbone` types. They include a reference to `jquery` which automatically registers the `$` global. However, since Titanium behaves like a Node environment jQuery never actually registers its `$` global, thus falsely populating it with jQuery types. We use `$` as a short hand for `AlloyController`, so it shouldn't contain any jQuery types.

BTW, you extend `AlloyController` from `Backbone.Model`. However, Alloy controllers extend from `Backbone.Events`. Any specific reason you'd choose `Backbone.Model` over `Backbone.Events`? See here:

https://github.com/appcelerator/alloy/blob/14803afc36b63bacbed4d900748c3024efca3463/Alloy/lib/alloy/controllers/BaseController.js#L28

The only solution i currently see is that we manually copy/include the Backbone typings that we actually use in Alloy and don't rely on the official typings package.

ffMathy:
Good point! I'll get back to all these concerns soon.

ffMathy:
@janvennemann how do you suggest we fix the Backbone typings? Any creative ideas? I have time to look at the concerns now.

ffMathy:
I started doing some work over at #40295.

ffMathy:
@janvennemann fixing the issues with jQuery and Backbone typings is a huge amount of work it would seem. I just keep going deeper down the rabbit hole.

Can at least agree that adding "dom" to "libs" is better, and then have full strong typing?

PR #39940, opened 2019-10-28
[simple-oauth2] Allow http options for getToken

bertyhell:
According to the docs of simple-oauth2 the getToken function optionally accepts an httpOptions object and a scope property on the token config object
https://www.npmjs.com/package/simple-oauth2#authorization-code-flow

```javascript
// Get the access token object (the authorization code is given from the previous step).
const tokenConfig = {
  code: '<code>',
  redirect_uri: 'http://localhost:3000/callback',


  //-------------------- this -------------------------
  scope: '<scope>', // also can be an array of multiple scopes, ex. ['<scope1>, '<scope2>', '...']


};
 
// Optional per-call http options
const httpOptions = {};
 
// Save the access token
try {


  //-------------------- this -------------------------
  const result = await oauth2.authorizationCode.getToken(tokenConfig, httpOptions);


  const accessToken = oauth2.accessToken.create(result);
} catch (error) {
  console.log('Access Token Error', error.message);
}
```

Please fill in this template.

### DIFF
--- a/types/simple-oauth2/index.d.ts
+++ b/types/simple-oauth2/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for simple-oauth2 2.2
+// Type definitions for simple-oauth2 2.5.2
 // Project: https://github.com/lelylan/simple-oauth2
 // Definitions by: Michael Müller <https://github.com/mad-mike>,
 //                 Troy Lamerton <https://github.com/troy-lamerton>
--- a/types/simple-oauth2/index.d.ts
+++ b/types/simple-oauth2/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for simple-oauth2 2.5.2
+// Type definitions for simple-oauth2 2.5
 // Project: https://github.com/lelylan/simple-oauth2
 // Definitions by: Michael Müller <https://github.com/mad-mike>,
 //                 Troy Lamerton <https://github.com/troy-lamerton>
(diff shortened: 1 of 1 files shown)


types/simple-oauth2/index.d.ts
```diff
@@ -113,7 +115,7 @@ export interface OAuthClient<ClientIdName extends string = 'client_id'> {
         ): string,
 
         /** Returns the Access Token object */
-        getToken(params: AuthorizationTokenConfig): Promise<Token>;
+        getToken(params: AuthorizationTokenConfig, httpOptions?: any): Promise<Token>;
```
netux: According to [this commit on simple-oauth2](https://github.com/lelylan/simple-oauth2/commit/beb0c316b4fc8feab835071d77646daaf95bc641), `ownerPassword`, `clientCredentials` and `accessToken` also got this parameter added.

netux: Also, what types can `httpOptions` be?

netux (changes requested):
Pretty good, thanks! But some things have to be changed.

types/simple-oauth2/index.d.ts
bertyhell: Ok, added the param to the other functions

The type seems to be an http options object from the Wreck library:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/wreck/index.d.ts#L17-L41

I've copied the interface over

netux (approved):

PR #39942, opened 2019-10-28
Add typings for adhan

ikr4-m:

If adding a new definition:

### DIFF
--- a/types/adhan/index.d.ts
+++ b/types/adhan/index.d.ts
@@ -38,7 +38,7 @@ export const Date: AdhanDate;
 
 export const Math: AdhanMath;
 
-// Declare Interface privately
+// Declare Interface
 export interface AdhanCalculationMethod {
   Dubai(): Paramater;
   Egyptian(): Paramater;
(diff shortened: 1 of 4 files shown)


types/adhan/index.d.ts
```diff
@@ -0,0 +1,93 @@
+// Type definitions for adhan-js 3.0
+// Project: https://github.com/batoulapps/adhan-js
+// Definitions by: Ikramullah <https://github.com/skymunn>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+// Start declare module
+// declare module 'adhan' { export = Adhan }
```
jessetrinity: Did you mean to export this here?

[adhan](https://github.com/batoulapps/adhan-js#node) looks like a UMD library

https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#umd

jessetrinity (changes requested):

types/adhan/index.d.ts
skymunn: Yes, Adhan is UMD library. I can call it globally when using <script> in HTML.

jessetrinity (approved):

PR #39951, opened 2019-10-28
Add types for `@ember/ordered-set`.

### LINKED ISSUE #884: Types for @ember/ordered-set
### Which package(s) does this enhancement pertain to?

  - [x] Other: [@ember/ordered-set](https://github.com/emberjs/ember-ordered-set)

### Please write a user story for this feature

Follow the form

> *As a **user or developer of ember-data**, I want **types for @ember/ordered-set** so that **I don't have to vendor them**.*

chriskrycho:
Defines the public API for [@ember/ordered-set](https://github.com/emberjs/ember-ordered-set). Note that the API is only documented thus:

> The `OrderedSet` class has mostly the same API as the native [`Set`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)
class with a few differences:
>
> - The constructor does not take any arguments
> - A static `create()` method exists for symmetry with `Ember.Object`
> - A static `length` property does not exist on `OrderedSet`
> - `OrderedSet` has an `isEmpty()` method
> - There are no `entries()`, `keys()` and `values()` methods, but there is a
>   `toArray()` method instead
> - The `@@iterator` symbol is not defined
> - `OrderedSet` has a `copy()` method

As such I used the test suite as a guide to the API, and used the same naming conventions as those used in the TS `Set` definition.

---

If adding a new definition:

    ***Note:** I've added it for the npm-naming rule, but also added a comment to clarify why!*

cc. @rwjblue @dfreeman @jamescdavis @mike-north 

---

Resolves typed-ember/ember-cli-typescript#884

### DIFF
--- a/types/ember__ordered-set/index.d.ts
+++ b/types/ember__ordered-set/index.d.ts
@@ -23,8 +23,8 @@ export default class OrderedSet<T = unknown> {
     add(value: T): this;
     clear(): void;
     delete(value: T): boolean;
-    forEach(callbackfn: (this: undefined, value: T, value2: T, set: Set<T>) => void): void;
-    forEach<Ctx>(callbackfn: (this: Ctx, value: T, value2: T, set: Set<T>) => void, context: Ctx): void;
+    forEach(callbackfn: (this: undefined, value: T, value2: T, set: OrderedSet<T>) => void): void;
+    forEach<Ctx>(callbackfn: (this: Ctx, value: T, value2: T, set: OrderedSet<T>) => void, context: Ctx): void;
     has(value: T): boolean;
     isEmpty(): boolean;
     toArray(): T[];
--- a/types/ember__ordered-set/index.d.ts
+++ b/types/ember__ordered-set/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for @ember/object 2.0
+// Type definitions for @ember/ordered-set 2.0
 // Project: https://github.com/emberjs/ember-ordered-set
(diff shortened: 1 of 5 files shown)


types/ember__ordered-set/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "npm-naming": false // The ember addon package exports *build-time*, not *run-time* code.
```
jessetrinity: I don't see this rule disabled in the other ember packages. What is different about this one?

jessetrinity (changes requested):

types/ember__ordered-set/tslint.json
chriskrycho: In this case the Ember package is actually being published separately to npm, whereas most of the `@ember/*` packages do *not* have distinct npm packages and are resolved via Ber’s build process (and are actually part of `ember-source`), but in this case the package also has no export matching what dtslint expects because of how Ember addons work.

types/ember__ordered-set/index.d.ts
```diff
@@ -0,0 +1,32 @@
+// Type definitions for @ember/object 2.0
```
dfreeman: Should this be `@ember/ordered-set`?

chriskrycho: Goood catch! I’ll push a fix shortly!

jessetrinity (approved):

PR #39957, opened 2019-10-29
[mongodb] Fix and add watch to db definition

rauno56:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [ChangeStreamOptions](https://mongodb.github.io/node-mongodb-native/3.3/api/global.html#ChangeStreamOptions) & [watch](https://mongodb.github.io/node-mongodb-native/3.3/api/Db.html#watch)

### DIFF
--- a/types/mongodb/index.d.ts
+++ b/types/mongodb/index.d.ts
@@ -72,7 +72,7 @@ export class MongoClient extends EventEmitter {
     logout(options: { dbName?: string }, callback: MongoCallback<any>): void;
     /** http://mongodb.github.io/node-mongodb-native/3.1/api/MongoClient.html#startSession */
     startSession(options?: SessionOptions): ClientSession;
-    /** http://mongodb.github.io/node-mongodb-native/3.3/api/MongoClient.html#watch */
+    /** http://mongodb.github.io/node-mongodb-native/3.1/api/MongoClient.html#watch */
     watch(pipeline?: object[], options?: ChangeStreamOptions & { session?: ClientSession }): ChangeStream;
     /** http://mongodb.github.io/node-mongodb-native/3.1/api/MongoClient.html#withSession */
     withSession(operation: (session: ClientSession) => Promise<any>): Promise<void>;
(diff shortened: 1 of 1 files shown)


types/mongodb/index.d.ts
```diff
@@ -2257,6 +2260,8 @@ export interface GridFSBucketWriteStreamOptions extends WriteConcern {
 
 /** http://mongodb.github.io/node-mongodb-native/3.1/api/ChangeStream.html */
 export class ChangeStream extends Readable {
+    resumeToken: ResumeToken;
```
CaselIT: Can you also update the documentation link of ChangeStream to 3.3? (resumeToken was not documented in the 3.1)

CaselIT (changes requested):
Just a documentation link change.

Ok otherwise

types/mongodb/index.d.ts
```diff
@@ -2278,14 +2283,20 @@ export class ChangeStream extends Readable {
     stream(options?: { transform?: Function }): Cursor;
 }
 
+export interface ResumeToken {
+    _data: string;
```
daprahamian: This is documented [here](https://docs.mongodb.com/master/changeStreams/#change-stream-resume-token) as being either a `string` or `BinData` (being of type `BSON.Binary`).

More importantly, this is meant to be a relatively opaque type. The intent of [the `ResumeToken` type](https://mongodb.github.io/node-mongodb-native/3.3/api/global.html#ResumeToken) is not meant to document the internals of a resumeToken. Rather it is to provide more clarity than `object` into what should be passed in to `startAfter` and `resumeAfter`.

I'm not sure what the policy is for DefinitelyTyped typings, but I would discourage describing the internals of ResumeToken here.

daprahamian (changes requested):

types/mongodb/index.d.ts
Rauno56: Will do. Thanks.

types/mongodb/index.d.ts
Rauno56: Fair point. I wanted to document that because what you'd usually do is pass the same object you got from the driver back to it. I lack enough experience in TS to know how to express that.

Any ideas?

CaselIT: I think an empty class could work

```ts
export class ResumeToken {}
```

Issue #38009, opened 2019-08-29
[@types/react-select] StateManager props definition


jbojcic1:
If you know how to fix the issue, make a pull request instead.

- [x] I tried using the `@types/xxxx` package and had problems.th
- [x] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @claasahl @jonfreedman @Methuselah96 @thchia 

I wanted to create wrapper component around react-select component and since I am using typescript I wanted to reuse props interface and just extend it but I wasn't able to find one interface that has all the props. Am I missing something?

Also when I look into the ts definitions, it seems that state manager props are built from couple of interfaces/types like this:

```
StateProps<SelectProps<OptionType>> & Props<OptionType> & SelectProps<OptionType>
```

where state props is:

```
type StateProps<T extends SelectProps<any>> = Pick<T, Exclude<keyof T,
  | 'inputValue'
  | 'value'
  | 'menuIsOpen'
  | 'onChange'
  | 'onInputChange'
  | 'onMenuClose'
  | 'onMenuOpen'
>>;
```

so I am wondering what is the point of taking subset of `SelectProps` with `StateProps<SelectProps<OptionType>>` and then doing & with `SelectProps` again. Am I missing something or result will again have all properties from `SelectProps`?

jonfreedman:
There's an example of a wrapper @ https://github.com/mui-org/material-ui/blob/master/docs/src/pages/components/autocomplete/IntegrationReactSelect.js

Unfortunately adding types to this is a bit of work that I can't share however it is possible to create a type safe wrapper with styles [using a hook](https://stackoverflow.com/questions/55512097/how-do-i-export-a-typescript-react-component-which-uses-material-ui-withstyles).

I'm not sure what you're trying to do - if you want to allow all of the use cases (e.g. asyc, creatable, multi) you'll either need to union the types yourself, or have separate wrappers.  Specifically with the StateManager type, we only use that to maintain a reference e.g.

```typescript
private readonly selectRef: React.RefObject<StateManager<TI>>;
```

gricey432:
That link is gone

JKirchartz:
[...]

PR #39960, opened 2019-10-29
[react-select] export types from main entry-point

danieldelcore:
## What? 
Re-exports internal types from the main entry point (`/index.d.ts`)

## Why? 
The current approach requires importing internal types like so:

`import { OptionsType } from 'react-select/src/types';`

Which can be quite counterintuitive based on my previous experiences with other libraries. You have to search the source code of the `node_module` to find the correct path then import that path explicitly. Internal changes to the types package would be a breaking change even if types are not altered.

It would be nice to make types accessible via the main entry-point, similar to other common libraries. For example: `import React, { ReactNode } from 'react';`

It looks like I'm not the only one, there are a few issues in DefinitelyTyped and ReactSelect stemming from the same confusion:

> It is weird to have a typing library for react-select and not be able to do anything with it. How do you guys customize components for a select if you're using typescript? What type hosts innerProps?

- https://github.com/JedWatson/react-select/issues/3509
- https://github.com/DefinitelyTyped/DefinitelyTyped/issues/38009
- https://github.com/DefinitelyTyped/DefinitelyTyped/issues/37444

### DIFF
--- a/types/react-select/index.d.ts
+++ b/types/react-select/index.d.ts
@@ -11,6 +11,20 @@ import { StateManager } from './src/stateManager';
 
 export default StateManager;
 
+export * from './src/types';
 export { createFilter } from './src/filters';
-export { components } from './src/components/index';
-export { mergeStyles } from './src/styles';
+export { mergeStyles, Styles, StylesConfig } from './src/styles';
+
+export { Props, FormatOptionLabelMeta } from './src/Select';
+
+export { components, SelectComponentsConfig, IndicatorComponentType } from './src/components';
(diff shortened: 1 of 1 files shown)


danieldelcore:
Hey @thchia, would really appreciate another review, if you have the time 😊
What's the process of becoming a maintainer of this package? I wouldn't mind helping out with maintaining these typings. Do i just need to add myself to the header?

thchia:
@danieldelcore yes just add your username to the header!

thchia (approved):

danieldelcore:
Thank you very much @thchia <3

PR #39973, opened 2019-10-29
Update node-resque types to match v5.5.7.

petetnt:
Signed-off-by: petetnt <pete.a.nykanen@gmail.com>

This PR updates the `node-resque` types to match the current version 5.5.7. that seems to have quite a lot of new methods but not much breaking changes.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/taskrabbit/node-resque

### DIFF
--- a/types/node-resque/index.d.ts
+++ b/types/node-resque/index.d.ts
@@ -2,6 +2,7 @@
 // Project: http://github.com/taskrabbit/node-resque
 // Definitions by: Gordey Doronin <https://github.com/gordey4doronin>, Pete Nykänen <https://github.com/petetnt>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.1
 
 /// <reference types="node" />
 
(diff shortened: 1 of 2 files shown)


types/node-resque/index.d.ts
```diff
@@ -36,14 +36,46 @@ export interface QueueOptions {
     connection?: ConnectionOptions;
 }
 
+type WorkerStatus = {
```
gordey4doronin: I think using interfaces over types is still preferred in that repository?
BTW That's one of the reasons why the build fails.

types/node-resque/index.d.ts
```diff
@@ -1,6 +1,6 @@
-// Type definitions for node-resque 5.4
+// Type definitions for node-resque 5.5.7
```
gordey4doronin: The build says patch version is not allowed. Honestly I'm not sure why, would be interested to know. But if you want build to pass you probably need to change it. 😬

petetnt:
Thanks for the early review @gordey4doronin ! The bot was a bit hasty, published this one as a draft yesterday as I didn't have the time to finish it yesterday but the bot still went ahead and did all its tasks 😅. I'll do some additions + tests today and ping you after that.

types/node-resque/index.d.ts
petetnt: Yeah makes sense, patches shouldn't contain API changes (not that semver is set in stone). I'll change those!

gfoltz:
The `ConnectionOptions` interface also needs an optional `redis` attribute.

See https://github.com/taskrabbit/node-resque/blob/master/lib/connection.js#L41
And the readme at https://github.com/taskrabbit/node-resque/blame/master/README.md#L164-L170

petetnt:
Fixed some of the types, added tests and cleaned up linting errors.

I added the `redis?` parameter as `any`, it should be `redis?: IORedis` but referencing it with `/// references types="ioredis"` has some external errors that should be fixed...

petetnt:
The BenchmarkPR test failed as `typescript@3.7` is not out yet, right? ¯\_(ツ)_/¯

gordey4doronin (approved):

types/node-resque/node-resque-tests.ts
```diff
@@ -64,9 +64,41 @@ worker.connect().then(() => worker.start());
 // start scheduler
 scheduler.connect().then(() => scheduler.start());
 
-queue.connect().then(() => {
+queue.connect().then(async () => {
```
gordey4doronin: Can we also add the following snippet after line 24, similar to #40027 to make it fully consistent to @gfoltz changes.

```ts
 const connectionWithCustomRedis: ConnectionOptions = { redis: "fake_redis_client" };
```

gordey4doronin:
Here is an issue for failing build https://github.com/DefinitelyTyped/DefinitelyTyped/issues/40054 🤔

types/node-resque/node-resque-tests.ts
petetnt: Updated in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39973/commits/45a3df7a5e9f1801a0722d4857f842d4dced6b99

gordey4doronin: 👍 Hopefully failing CI would be figured out soon and you can merge then. 🙏

PR #39983, opened 2019-10-29
[react-helmet] relax htmlAttributes and bodyAttributes types

### LINKED ISSUE #39357: [@types/react-helmet] htmlAttributes type no longer allows data attributes
  - Authors: @ykzts, @evanbb, @isman-usoh, @sammkj, @yuit

---

Looks like `htmlAttributes` was changed from `any` to `JSX.IntrinsicElements['html']` in #39161. This does not allow for `data-*` attributes, which should be valid (AFAIK).

#### Example

```jsx
<Helmet htmlAttributes={{ 'data-foo': 'bar' }} />
```
```
Object literal may only specify known properties, and ''data-foo'' does not exist in
type 'DetailedHTMLProps<HtmlHTMLAttributes<HTMLHtmlElement>, HTMLHtmlElement>'.
```

Not sure of the appropriate fix. Thoughts?

### LINKED ISSUE #39430: [@types/react-helmet] bodyAttributes now requires 'className' property not being mapped in react-helmet ^5.2.0
- [x] I tried using the `@types/react-helmet^5.0.13` package and had problems.
  - Authors: @ykzts, @evanbb, @isman-usoh, @sammkj, @yuit 

Using `react-helmet^5.2.0`.
Since `bodyAttributes` is now a `JSX.IntrinsicElements['body']` instead of an `Object`, it now requires a `className` field. However, the `REACT_TAG_MAP` constant in the react-helmet library maps `class` to `className` and is thus ignoring the classes.

Maybe create a temporary type which extends `JSX.IntrinsicElements['body']` and adds a `class` field to it? While waiting to see if `react-helmet` would change the mapping.

wKovacs64:
This augments (and partially reverts) #39161 to preserve type-checking of known attributes but also allow for arbitrary attributes like `data-*`.

Closes #39357
Closes #39430

**P.S.** Sorry about the formatting commit, but it was wrong so I corrected it. If you want this PR to be more pure, I can remove it.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/39357#issuecomment-547630291

### DIFF
--- a/types/react-helmet/index.d.ts
+++ b/types/react-helmet/index.d.ts
@@ -9,11 +9,11 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
-import * as React from "react";
+import * as React from 'react';
 
 type HtmlProps = JSX.IntrinsicElements['html'];
 
-type BodyProps = JSX.IntrinsicElements['body']
+type BodyProps = JSX.IntrinsicElements['body'];
 
 type LinkProps = JSX.IntrinsicElements['link'];
 
@@ -36,11 +36,7 @@ export interface HelmetProps {
     defer?: boolean;
     encodeSpecialCharacters?: boolean;
     htmlAttributes?: HtmlProps;
-    onChangeClientState?: (
-        newState: any,
-        addedTags: HelmetTags,
-        removedTags: HelmetTags,
-    ) => void;
+    onChangeClientState?: (newState: any, addedTags: HelmetTags, removedTags: HelmetTags) => void;
     link?: LinkProps[];
     meta?: MetaProps[];
     noscript?: Array<any>;
--- a/types/react-helmet/index.d.ts
+++ b/types/react-helmet/index.d.ts
@@ -6,14 +6,19 @@
 //                 Kok Sam <https://github.com/sammkj>
 //                 Yui T. <https://github.com/yuit>
 //                 Yamagishi Kazutoshi <https://github.com/ykzts>
+//                 Justin Hall <https://github.com/wKovacs64>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
 import * as React from 'react';
 
-type HtmlProps = JSX.IntrinsicElements['html'];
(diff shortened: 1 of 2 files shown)


evanbb (approved):
LGTM!

fluidsonic:
@wKovacs64 this PR has actually made the types for most attributes stricter than before, and prevents some valid values from being used.

In the original definition from React - depending on the attribute - values could also be `undefined`, `boolean`, `number` or otherwise. That is no longer possible because `[key: string]: string;` used in this PR forces **all** properties to be `string` and nothing else.
Basically all type info from the original definition is lost.

Please refer to that unit test for an example:
https://github.com/nfl/react-helmet/blob/420810c644f94c3743f7b088321dbd62a8037b7b/test/HelmetTest.js#L467-L475

`undefined` is legit but doesn't work anymore.

wKovacs64:
@fluidsonic `htmlAttributes` was originally typed as `any`, until @ykzts changed it to `JSX.IntrinsicElements['html']` in #39161. That broke `data-` attributes, so I added the intersection with `[key: string]: string;`, but you're right - that isn't necessarily a string. We could change it to `[key: string]: any;`?

evanbb:
good catch @fluidsonic !

as @wKovacs64 mentioned, the original intention was to provide better type definitions than just `any`. maybe instead of `[key:string]: any` we could do `[key:string]: React.ReactNode` which would include `undefined`, `boolean`, `number`, etc.

wKovacs64:
I'm down with `React.ReactNode`, if that's sufficient (certainly better than `any`). I'll update the PR.

evanbb:
i had a bit of a brain fart yesterday - i think it would be misleading to include `ReactElement`, `ReactFragment` and `ReactPortal` in the union of valid types for these attributes - it would probably be best to spell it out explicitly:

`[key: string]: string | number | boolean | null | undefined`

@wKovacs64 thoughts on updating #41352 ?

wKovacs64:
I was wondering about that, but I figured I just didn't know a good reason to allow the other stuff. My bad, I should've spoken up. I'll update the PR again.

evanbb:
Awesome, thanks @wKovacs64 ! I'll share the blame for the rework :rofl:

PR #39991, opened 2019-10-30
Expand create-hash API support

jordaaash:
The `create-hash` project [README](https://github.com/crypto-browserify/createHash/blob/master/README.md) states

> API is the same as hashes in node

Referring to the Node API: https://nodejs.org/api/crypto.html#crypto_crypto_createhash_algorithm_options

The `update` and `write` methods should allow for input and output of `Buffer` and other binary data objects besides strings, and return a chainable interface.

The `digest` method should return a string or a `Buffer` depending on whether an argument is provided.

This can be further verified by reading the source implementation.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/crypto-browserify/createHash/blob/master/README.md

### DIFF
--- a/types/create-hash/index.d.ts
+++ b/types/create-hash/index.d.ts
@@ -1,16 +1,24 @@
 // Type definitions for create-hash 1.2
 // Project: https://github.com/crypto-browserify/createHash
 // Definitions by: BendingBender <https://github.com/BendingBender>,
-//                 Konstantin Yuriev <https://github.com/gallowsmaker>
-//                 Max Boguslavskiy <https://github.com/maxbogus>
+//                 Konstantin Yuriev <https://github.com/gallowsmaker>,
+//                 Max Boguslavskiy <https://github.com/maxbogus>,
+//                 Jordan Sexton <https://github.com/jordansexton>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.1
 
+/// <reference types="node" />
+
 declare namespace createHash {
+    type TypedArray = Uint8Array | Uint8ClampedArray | Uint16Array | Uint32Array | Int8Array | Int16Array | Int32Array | Float32Array | Float64Array;
+
     interface HashAlgorithm {
-        update(data: string): void;
-        digest(target?: encoding): string;
-        write(data: string): void;
+        digest(target: encoding): string;
+        digest(): Buffer;
+
+        update(data: string | Buffer | TypedArray | DataView, encoding?: string): this;
+        write(data: string | Buffer | TypedArray | DataView, encoding?: string): this;
+
         end(): void;
         read(): void;
     }


jordansexton:
Tests are failing but the failure doesn't appear to be related to this package. Is this a known or common issue?

types/create-hash/index.d.ts
```diff
@@ -1,16 +1,22 @@
 // Type definitions for create-hash 1.2
 // Project: https://github.com/crypto-browserify/createHash
 // Definitions by: BendingBender <https://github.com/BendingBender>,
-//                 Konstantin Yuriev <https://github.com/gallowsmaker>
-//                 Max Boguslavskiy <https://github.com/maxbogus>
+//                 Konstantin Yuriev <https://github.com/gallowsmaker>,
+//                 Max Boguslavskiy <https://github.com/maxbogus>,
+//                 Jordan Sexton <https://github.com/jordansexton>,
```
maxbogus: I think you should remove comma at the end of the sentence.
Commit changes and run "npm test" again.

types/create-hash/index.d.ts
```diff
@@ -1,16 +1,22 @@
...
 // TypeScript Version: 2.1
 
 declare namespace createHash {
+    type TypedArray = Uint8Array | Uint8ClampedArray | Uint16Array | Uint32Array | Int8Array | Int16Array | Int32Array | Float32Array | Float64Array;
+
     interface HashAlgorithm {
-        update(data: string): void;
-        digest(target?: encoding): string;
-        write(data: string): void;
+        digest(target: encoding): string;
```
maxbogus: You should return back optional parameter to digest. Check this guide. https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#use-optional-parameters

types/create-hash/index.d.ts
jordansexton: Got it, I'll try that. Thanks!

types/create-hash/index.d.ts
jordansexton: The parameter to `digest` is optional, but it changes the return type of the function if used. This is why I instead used two declarations.to overload the function:

```typescript
digest(target: encoding): string;
digest(): Buffer;
```

This way the calling code receives back the correct type based on its use of the argument.

jordansexton:
Tests on Travis CI passing now. Some other testing I'm not familiar with on Azure is failing, but for a reason unrelated to this change: https://dev.azure.com/definitelytyped/DefinitelyTyped/_build/results?buildId=21814

```
Testing...
npm ERR! code ETARGET
```

types/create-hash/index.d.ts
maxbogus: Didn't notice that. In that case your variant is correct.

types/create-hash/index.d.ts
```diff
@@ -1,16 +1,24 @@
...
 // Definitions by: BendingBender <https://github.com/BendingBender>,
-//                 Konstantin Yuriev <https://github.com/gallowsmaker>
-//                 Max Boguslavskiy <https://github.com/maxbogus>
+//                 Konstantin Yuriev <https://github.com/gallowsmaker>,
+//                 Max Boguslavskiy <https://github.com/maxbogus>,
+//                 Jordan Sexton <https://github.com/jordansexton>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.1
 
+/// <reference types="node" />
```
maxbogus: Is it really required to use node types? This typings are used for browser version too.

jordansexton: It's for `Buffer`, tsc throws otherwise. `Buffer` is a special case in the browser because Webpack usually automagically shims it. I noticed this is how the `secp256k1` package handles this issue, so I did the same: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4b6756548680a242aaf29d7a8ee3d20b017a89ca/types/secp256k1/index.d.ts#L6

maxbogus: I think that we should use https://github.com/feross/buffer for this case.
@VladimirGrenaderov what do you think?

jordansexton: I don't see how this would declare that as a dependency, but also, `create-hash` does not do this and just relies on `Buffer` being available.

FWIW, this is also how it's done in several other packages, notably even https://github.com/feross/is-buffer

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/2206d001afbac3592a459dd68a99f742d46eff68/types/is-buffer/index.d.ts#L7

jordansexton: (if I'm reading the code correctly?)

jordansexton: Ah, it looks like internally it does this:

https://github.com/crypto-browserify/cipher-base/blob/bc18dc847050c34f11f7a0472ff13a76b53fedc6/index.js#L1

jordansexton: However, looking at typings for `cipher-base` it still seems like this is the idiomatic way to do it:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/4b6756548680a242aaf29d7a8ee3d20b017a89ca/types/cipher-base/index.d.ts#L6

maxbogus: Ok, let's do this.

maxbogus: I think you should push some small change to force Travis to rerun its build.

jordansexton: I'll try that.

jordansexton: Done!

jordansexton:
Anything else that I should do on this?

maxbogus:
> Anything else that I should do on this?

Have you tried running npm test locally?

maxbogus:
I found this bug : https://github.com/microsoft/dtslint/issues/235
Somehow dts-header-parser fails to read header.

jordansexton:
> Have you tried running npm test locally?

Yep! I did, tests pass locally.

maxbogus:
I think it's a bug in the build system. You could make another "small" commit and try to get "lucky" build. Or search for this problem on google.

jordansexton:
I've tried a few times. Looks like there's a new issue about this: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/40054

Maybe related: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/40050

jordansexton:
Per https://github.com/DefinitelyTyped/DefinitelyTyped/issues/40054#issuecomment-548879184 it looks like maybe this can be merged as is?

maxbogus (approved):

maxbogus:
@jordansexton i approved pull req. i hope that we can merge it with failing build. if not, we had to wait till this issue will be fixed.

sandersn:
Yep, sorry about the confusion. The main build is fixed but the benchmark is still broken.

PR #39995, opened 2019-10-30
[relay-compiler] Add missing types for upcoming relay-compiler-language-typescript

sorenhoyer:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/relay-tools/relay-compiler-language-typescript/pull/155

### DIFF
--- a/types/relay-compiler/lib/core/RelayCompilerError.d.ts
+++ b/types/relay-compiler/lib/core/RelayCompilerError.d.ts
@@ -5,6 +5,6 @@ export type UserError = GraphQLError;
 
 export function createUserError(
   message: string,
-  locations?: readonly Location[],
-  nodes?: readonly ASTNode[],
+  locations?: ReadonlyArray<Location>,
+  nodes?: ReadonlyArray<ASTNode>,
 ): UserError;
--- a/types/relay-compiler/lib/core/RelayCompilerError.d.ts
+++ b/types/relay-compiler/lib/core/RelayCompilerError.d.ts
@@ -1,0 +1,10 @@
+import { ASTNode, GraphQLError } from 'graphql';
+import { Location } from './GraphQLIR';
+
+export type UserError = GraphQLError;
+
+export const createUserError: (
+  message: string,
+  locations?: readonly Location[],
+  nodes?: readonly ASTNode[],
+) => UserError;
--- a/types/relay-compiler/lib/core/RelayCompilerError.d.ts
+++ b/types/relay-compiler/lib/core/RelayCompilerError.d.ts
@@ -3,8 +3,8 @@ import { Location } from './GraphQLIR';
 
 export type UserError = GraphQLError;
 
-export const createUserError: (
+export function createUserError(
   message: string,
   locations?: readonly Location[],
   nodes?: readonly ASTNode[],
-) => UserError;
+): UserError;
--- a/types/relay-compiler/index.d.ts
+++ b/types/relay-compiler/index.d.ts
@@ -65,6 +65,8 @@ export {
     Variable,
 } from './lib/core/GraphQLIR';
 
+export { createUserError } from './lib/core/RelayCompilerError';
+
 export {
     EnumTypeID,
     FieldID,
--- a/types/relay-compiler/index.d.ts
+++ b/types/relay-compiler/index.d.ts
@@ -3,7 +3,7 @@
 // Definitions by: n1ru4l <https://github.com/n1ru4l>
 //                 Eloy Durán <https://github.com/alloy>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 3.0
+// TypeScript Version: 3.6
 
 import * as ASTConvert from './lib/core/ASTConvert';
 import * as Parser from './lib/core/RelayParser';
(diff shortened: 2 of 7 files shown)


sorenhoyer:
In GraphQLIR.d line 59, ideally I wanted to do:

```ts
import { TypeReferenceNode } from 'typescript';

export interface Connection {
    key: string;
    conditional: boolean;
    value: TypeReferenceNode;
}
```

But that would result in the following error:

```
Error: C:/code/DefinitelyTyped/node_modules/typescript/lib/typescript.d.ts:1:1
ERROR: 1:1  no-outside-dependencies  File C:/code/DefinitelyTyped/node_modules/typescript/lib/typescript.d.ts comes from a `node_modules` but is not declared in this type's `package.json`.  See: https://github.com/Microsoft/dtslint/blob/master/docs/no-outside-dependencies.md

    at C:\code\DefinitelyTyped\node_modules\dtslint\bin\index.js:190:19
    at fulfilled (C:\code\DefinitelyTyped\node_modules\dtslint\bin\index.js:5:58)
```

We might get away with simply turning off the rule for that line, but I'm not sure if the linter has a point, so I settled with `any` to begin with.

What do you think? Should I insert a tslint ignore rule and keep it strongly typed or use `any`?

sorenhoyer:
Another question:

I had to also update TS version in the index.d.ts header of relay-config since it depends on relay-compiler.

I had to bump the version to TS 3.4 due to the readonly modifier I used in createUserError in RelayCompilerError.d.ts.

By bumping the TS version in relay-config (dependant on relay-compiler), do we need to also bump the major version. And the same question goes for relay-compiler. Do we bump to new major version everytime we change TypeScript version, or how do you usually do?

sorenhoyer:
TypeScript version was updated from 3.0 to 3.4 in relay-compiler and relay-config (has dependency to relay-compiler). This is necessary since I used readonly modifier for array types in createUserError in RelayCompilerError.d.ts. It's not a big change but since there is a change in relay-config, should we bump the version to 7.0.0 now?

types/relay-compiler/lib/core/RelayCompilerError.d.ts
```diff
@@ -0,0 +1,10 @@
+import { ASTNode, GraphQLError } from 'graphql';
+import { Location } from './GraphQLIR';
+
+export type UserError = GraphQLError;
+
+export function createUserError(
+  message: string,
+  locations?: readonly Location[],
+  nodes?: readonly ASTNode[],
```
alloy: Seeing as these are equivalent to `ReadonlyArray<…>`, I would suggest using that instead so we don’t need to bump the TS version requirement yet.

alloy (changes requested):
Regarding the `TypeReferenceNode`, why not just add `typescript` to the list of dependencies?

> By bumping the TS version 4 minor versions in relay-config (dependant on relay-compiler), do we need to also bump the major version of relay-config? And the same question goes for relay-compiler. Do we bump to new major version everytime we change TypeScript version, or how do you usually do?

The major and minor version of DT packages should reflect the _minimum_ version of the package it describes that it will support. The patch version gets increased automatically for each change to the typings. So alas there’s no way for us to convey the semantic meaning you’re referring to.

types/relay-compiler/lib/core/RelayCompilerError.d.ts
sorenhoyer: Agree, misread the documentation so thought ReadonlyArray was also new. Changed accordingly.

alloy: Great 👍

sorenhoyer: About the other thing about adding typescript  to the dependencies, isn't it the same for DT as with all other packages, that dependencies are installed when you install the package eg. @types/relay-compiler? In that then you'd typically just have typescript (but also graphql as I can see is already a dependency) as dev- and peerDependencies, right? Is DT not installing these dependencies the same way?

sorenhoyer:
> Regarding the `TypeReferenceNode`, why not just add `typescript` to the list of dependencies?

Isn't it the same for DT as with all other packages, that dependencies are installed when you install the package eg. @types/relay-compiler? In that then you'd typically want to have typescript (but also graphql as I can see is already a dependency) as dev- and peerDependencies, right? Is DT not installing these dependencies the same way? If yes, they I would go a long way to avoid the need of having dependencies in types.

alloy:
There’s many considerations around type package dependencies, see https://github.com/microsoft/types-publisher/pull/655 for instance, but right now this is the only thing we can do.

I feel like the dependency managers should be able to resolve these, but in reality it sometimes requires some manual massaging of the lockfile to ensure you only have a single resolved version of the dependency, in this case the `typescript` package. Having said that, I also feel like only people working on tooling will ever install the types for `relay-compiler` and those people are probably very able to deal with those quirks of the current dependency managers, so I’m ok with at least trying it out.

sorenhoyer:
> There’s many considerations around type package dependencies, see [microsoft/types-publisher#655](https://github.com/microsoft/types-publisher/pull/655) for instance, but right now this is the only thing we can do.
> 
> I feel like the dependency managers should be able to resolve these, but in reality it sometimes requires some manual massaging of the lockfile to ensure you only have a single resolved version of the dependency, in this case the `typescript` package. Having said that, I also feel like only people working on tooling will ever install the types for `relay-compiler` and those people are probably very able to deal with those quirks of the current dependency managers, so I’m ok with at least trying it out.

Good point, I've added it back.

alloy (approved):

alloy:
Great, thanks!

PR #40013, opened 2019-10-30
[jasmine] Add definition for jasmine.Env.configure and mark methods as deprecated

djungowski:
Add definition for jasmine.Env.configure and mark Env methods throwOnExpectationFailure, stopOnSpecFailure, seed and randomizeTests as deprecated

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://jasmine.github.io/api/3.4/Env.html
- [N/A] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [N/A] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

### DIFF
--- a/types/jasmine/index.d.ts
+++ b/types/jasmine/index.d.ts
@@ -15,6 +15,7 @@
 //                 Stephen Farrar <https://github.com/stephenfarrar>
 //                 Mochamad Arfin <https://github.com/ndunks>
 //                 Alex Povar <https://github.com/zvirja>
+//                 Dominik Ehrenberg <https://github.com/djungowski>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 // For ddescribe / iit use : https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/karma-jasmine/karma-jasmine.d.ts
@@ -157,6 +158,18 @@ interface DoneFn extends Function {
 
     /** fails the spec and indicates that it has completed. If the message is an Error, Error.message is used */
     fail: (message?: Error | string) => void;
+}
+
+/**
+ * Configuration that can be used when configuring Jasmine via {@link jasmine.Env.configure}
+ */
+interface EnvConfiguration {
+    random?: boolean;
+    seed?: number;
+    failFast?: boolean;
+    oneFailurePerSpec?: boolean;
+    hideDisabled?: boolean;
+    specFilter?: Function;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


ndunks (approved):

types/jasmine/ts3.1/index.d.ts
```diff
@@ -155,6 +156,18 @@ interface DoneFn extends Function {
     fail: (message?: Error | string) => void;
 }
 
+/**
+ * Configuration that can be used when configuring Jasmine via {@link jasmine.Env.configure}
+ */
+interface Configuration {
```
zvirja: Could you please rename it to `EnvConfiguration` to not occupy generic words? If later we have any other configuration, it might be nasty to have this one.

types/jasmine/ts3.1/index.d.ts
```diff
@@ -155,6 +156,18 @@ interface DoneFn extends Function {
...
+    random?: boolean;
+    seed?: number;
+    failFast?: boolean;
+    oneFailurePerSpec?: boolean;
+    hideDisabled?: boolean;
+    specFilter?: Function;
```
zvirja: Does it make sense to use `Fn` instead?

types/jasmine/ts3.1/index.d.ts
```diff
@@ -13,6 +13,7 @@
 //                 Stephen Farrar <https://github.com/stephenfarrar>
 //                 Mochamad Arfin <https://github.com/ndunks>
 //                 Alex Povar <https://github.com/zvirja>
+//                 Dominik Ehrenberg <https://github.com/djungowski>
```
zvirja: Notice, you should put your name here only if you would like to become a maintainer of this project and stay here for a long period of time. You will be notified each time there is a change related to this project.

So consider yourself whether that's exactly what you would like to sign in :wink:

zvirja (changes requested):
I've noticed that you updated typings for TS 3.1+ only without touching prior typings (under the `jasmine` directory immediately). Your changes doesn't contain anything which is specific to TS version, so could you please update that file as well?

Please also ensure that test files for "pre TS 3.1" and "TS 3.1+" are synchronized. It means that each line is the same for both files and the only difference might be a comment saying that errors is expected/not expected here. It will help to maintain the project going forward.

Thanks for doing that and making type types better!

types/jasmine/ts3.1/index.d.ts
djungowski: Fixed in a148df5612

types/jasmine/ts3.1/index.d.ts
djungowski: Not sure, tbh

types/jasmine/ts3.1/index.d.ts
djungowski: Since I use jasmine a lot, that would be fine for me - I guess 😅

djungowski:
@zvirja Synchronized the files in 40ef06fdc7

types/jasmine/index.d.ts
```diff
@@ -159,6 +160,18 @@ interface DoneFn extends Function {
     fail: (message?: Error | string) => void;
 }
 
+/**
+ * Configuration that can be used when configuring Jasmine via {@link jasmine.Env.configure}
+ */
+interface EnvConfiguration {
```
zvirja: Sorry for the late notice, overlooked that during previous iteration. Could you please move it under `jasmine` namespace? That helps to not pollute the global scope with jasmine specific types.

types/jasmine/ts3.1/index.d.ts
```diff
@@ -155,6 +156,18 @@ interface DoneFn extends Function {
     fail: (message?: Error | string) => void;
 }
 
+/**
+ * Configuration that can be used when configuring Jasmine via {@link jasmine.Env.configure}
+ */
+interface EnvConfiguration {
```
zvirja: Please move it under `jasmine` namespace.

types/jasmine/ts3.1/index.d.ts
zvirja: We use this type each time we expect function, so let's use it here as well. Same for previous file.

zvirja:
@djungowski Thanks for following up. Added few more things to change and then we are fine to go. Sorry that I overlooked them during previous review 😟

zvirja (approved):
LGTM, thanks! 👍

PR #40014, opened 2019-10-30
docs: improve readme.cn on external dependencies

Kerminate:


sandersn (approved):
Round-tripping through google translate seems reasonable, since I don't read Chinese (sorry).

PR #40017, opened 2019-10-30
[node] Improve os.constants.errno

galkin:
**This pull is additional part of adding deprecation warning to `constants` module.**

Check list:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/api/deprecations.html#deprecations_dep0008_require_constants


types/node/os.d.ts
```diff
@@ -172,6 +172,64 @@ declare module "os" {
             ETXTBSY: number;
             EWOULDBLOCK: number;
             EXDEV: number;
+            WSAEINTR: number;
```
SimonSchick: Nit: Can we sort those please?

SimonSchick (commented):
I was considering moving all of enums to interfaces and just make `constants` a type union to get rid of all of this duplication, thoughts?

types/node/os.d.ts
galkin: I sorted this list as this sorted at [source code](https://github.com/nodejs/node/blob/902c9fac194d06aec5fd5307a6712c39a6da18a7/src/node_constants.cc#L379).

galkin:
@SimonSchick, I thought about that too. During #39912 I found that @types/node is locked to TypeScript v2.0, not v2.1. There are some packages related on @types/node with v2.0 at description.

As result I don't have required typescript construction to implement my vision. [For example](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/os.d.ts#L55):
```typescript
// os module
const constants: {
...
signals: { [key in NodeJS.Signals]: number; };
...
```
What do you think? Maybe, I miss some simpler methods to achieve that? If yes, could you give me example?

galkin:
@jessetrinity, sorry it this questions are duplications. I want to ask:
1) Did you see about typescript@3.7 errors in pull requests today? 
2) What do you think about typescript@2.0 requirement for `@types/node`, because of requirements in related packages?

jessetrinity:
>Did you see about typescript@3.7 errors in pull requests today?

Yes, we broke it and we're working on unbreaking it :)

>What do you think about typescript@2.0 requirement for @types/node, because of requirements in related packages?

I'm not sure what you mean.

galkin:
@jessetrinity, if you will add line `// TypeScript Version: 2.1` to https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/index.d.ts, then you will have similar error https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/603594841#L248

jessetrinity:
@galkin are you referring to [this guidance](https://github.com/definitelytyped/definitelytyped#i-want-to-use-features-from-typescript-21-or-above)?

yes, that is what you would do, but you would need to add `// TypeScript Version: 2.1` everywhere you use node.

You could use `typesVersions` if you wanted to maintain two separate versions of node types, one with fancy new TS features and an old one that just gets bugfixes or whatever. This only multiplies the work you have to do and I believe node types contributors are trying to maintain less?

Has there been any conversation with the contributors about this?

galkin:
1) Yes, we are trying to maintain less.
2) I don't know any conversation about this issue.

I will ask more specific question. There [is comment at index.d.ts](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/index.d.ts#L48)
```
// NOTE: TypeScript version-specific augmentations can be found in the following paths:
//          - ~/base.d.ts         - Shared definitions common to all TypeScript versions
//          - ~/index.d.ts        - Definitions specific to TypeScript 2.1
//          - ~/ts3.2/index.d.ts  - Definitions specific to TypeScript 3.2
```
Should we change this `Definitions specific to TypeScript 2.1` to `Definitions specific to TypeScript 2.0`?

galkin:
@trivikr @Flarna @SimonSchick, could you provide feedbacks?

SimonSchick:
I'd prefer we export all enums as interfaces from their actual module sources and then export a type union instead, this is in line with what nodejs actually does.
See: https://github.com/nodejs/node/blob/master/lib/constants.js#L31

SimonSchick:
That said, this prevents us from declaring each individual member as deprecated but I don't feel this is actually necessary, I believe deprecating the top level `constants` module is good enough.

sandersn:
1. @SimonSchick have you signed off? I can't tell.
2. I'm going to start on removing 2.0 - 2.7 from the publish pipeline so you will soon be able to use 2.8+ features.
3. Azure's build failed and I can't tell why. I'm trying to re-run it.

SimonSchick (approved):
Looks good but I will likely create a followup PR that gets rid of duplication.

sandersn:
OK, and I got the Azure build to pass but it's not reporting correctly here. So I'll merge this.

PR #40021, opened 2019-10-30
Jasmine: Add Matchers.toBeInstanceOf

yelvert:
See: https://jasmine.github.io/api/edge/matchers.html#toBeInstanceOf

### DIFF
--- a/types/jasmine/index.d.ts
+++ b/types/jasmine/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for Jasmine 3.4
+// Type definitions for Jasmine 3.5
 // Project: http://jasmine.github.io
 // Definitions by: Boris Yankov <https://github.com/borisyankov>
 //                 Theodore Brown <https://github.com/theodorejb>
@@ -554,6 +554,7 @@ declare namespace jasmine {
         toThrowMatching(predicate: (thrown: any) => boolean): boolean;
         toBeNegativeInfinity(expectationFailOutput?: any): boolean;
         toBePositiveInfinity(expectationFailOutput?: any): boolean;
+        toBeInstanceOf(expected: Constructor): boolean;
 
         /**
          * Expect the actual value to be a DOM element that has the expected class.
--- a/types/jasmine/jasmine-tests.ts
+++ b/types/jasmine/jasmine-tests.ts
@@ -179,6 +179,11 @@ describe("Included matchers:", () => {
         expect(foo).toThrowError(TypeError, "foo bar baz");
     });
 
+    it("The 'toBeInstanceOf' matcher is for testing an object is an instance of the supplied class/constructor", () => {
+        class Foo {}
+        expect(new Foo()).toBeInstanceOf(Foo);
+    });
+
     it("async matchers", async () => {
         const badness = new Error("badness");
         await expectAsync(Promise.resolve()).toBeResolved();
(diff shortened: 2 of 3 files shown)


devoto13 (changes requested):

types/jasmine/index.d.ts
```diff
@@ -554,6 +554,7 @@ declare namespace jasmine {
         toThrowMatching(predicate: (thrown: any) => boolean): boolean;
         toBeNegativeInfinity(expectationFailOutput?: any): boolean;
         toBePositiveInfinity(expectationFailOutput?: any): boolean;
+        toBeInstanceOf(expected: any): boolean;
```
zvirja: Could you please verify whether we couldn't use more specific type here? As far as I can see, only prototype must be allowed. Probably, we could use `jasmine.Constructor` instead.

Also please add the tests to verify that it compiles for prototypes and fails for arbitrary values. Ensure to keep both "pre 3.1" and "3.1+" types in sync, so that they match line by line.

zvirja (changes requested):

yelvert:
Requested changes made

types/jasmine/ts3.1/index.d.ts
```diff
@@ -556,6 +556,7 @@ declare namespace jasmine {
         toThrowMatching(predicate: (thrown: any) => boolean): boolean;
         toBeNegativeInfinity(expectationFailOutput?: any): boolean;
         toBePositiveInfinity(expectationFailOutput?: any): boolean;
+        toBeInstanceOf(expected: any): boolean;
```
zvirja: Please use `Constructor` type here as well.

types/jasmine/jasmine-tests.ts
```diff
@@ -179,6 +179,11 @@ describe("Included matchers:", () => {
         expect(foo).toThrowError(TypeError, "foo bar baz");
     });
 
+    it("The 'toBeInstanceOf' matcher is for testing an object is an instance of the supplied class/constructor", () => {
+        class Foo {}
+        expect(new Foo()).toBeInstanceOf(Foo);
```
zvirja: Please add a few cases:
- for `Number` and `Boolean` built-in types
- negative case when you use not a prototype and expect compiler warning on that line

types/jasmine/jasmine-tests.ts
```diff
@@ -179,6 +179,11 @@ describe("Included matchers:", () => {
         expect(foo).toThrowError(TypeError, "foo bar baz");
     });
 
+    it("The 'toBeInstanceOf' matcher is for testing an object is an instance of the supplied class/constructor", () => {
```
zvirja: Please update also `jasmine-tests.ts` under the `ts3.1` folder and apply _exactly_ same change to keep files synchronized.

types/jasmine/index.d.ts
```diff
@@ -554,6 +554,7 @@ declare namespace jasmine {
         toThrowMatching(predicate: (thrown: any) => boolean): boolean;
         toBeNegativeInfinity(expectationFailOutput?: any): boolean;
         toBePositiveInfinity(expectationFailOutput?: any): boolean;
+        toBeInstanceOf(expected: Constructor): boolean;
```
zvirja: Please add documentation:

```
    /**
     * Expect the actual value to be an instance of the expected class.
     * @since 3.5.0
     */
```

Same for the definition under `ts3.1` folder.

zvirja (changes requested):
Please find out a couple of things to follow up on.
Thanks for working on this!

PR #40031, opened 2019-10-31
update   type of command zadd args

tong3jie:
update   type of command zadd args

Please fill in this template.

### DIFF
--- a/types/ioredis/index.d.ts
+++ b/types/ioredis/index.d.ts
@@ -229,7 +229,7 @@ declare namespace IORedis {
         smembers(key: KeyType, callback: (err: Error, res: any) => void): void;
         smembers(key: KeyType): Promise<any>;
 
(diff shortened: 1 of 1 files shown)


types/ioredis/index.d.ts
```diff
@@ -229,7 +229,7 @@ declare namespace IORedis {
         smembers(key: KeyType, callback: (err: Error, res: any) => void): void;
         smembers(key: KeyType): Promise<any>;
 
-        zadd(key: KeyType, ...args: string[]): Promise<number | string>;
+        zadd(key: KeyType, ...args: string | number[]): Promise<number | string>;
```
shaharmor: This needs to be `Array<string | number>` to have both `string` and `number` as possible values in the array

shaharmor (changes requested):
Update the type to make tests pass, and then we can approve

PR #40048, opened 2019-10-31
Add missing shelljs.exec option

pbelbin:
The shelljs `exec` function is missing the `fatal` option, which I want to use.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [https://documentup.com/shelljs/shelljs#execcommand--options--callback](https://documentup.com/shelljs/shelljs#execcommand--options--callback)

### DIFF
--- a/types/shelljs/index.d.ts
+++ b/types/shelljs/index.d.ts
@@ -841,7 +841,7 @@ export interface ExecOptions extends child.ExecOptions {
 	 * @default false
 	 */
 	fatal?: boolean;
-	
+
 	/**
 	 * Asynchronous execution.
 	 *
--- a/types/shelljs/index.d.ts
+++ b/types/shelljs/index.d.ts
@@ -836,7 +836,7 @@ export interface ExecOptions extends child.ExecOptions {
 	silent?: boolean;
 
 	/**
-	 * Exit when command return code is non-zero
+	 * Exit when command return code is non-zero.
 	 *
 	 * @default false
 	 */
--- a/types/shelljs/index.d.ts
+++ b/types/shelljs/index.d.ts
@@ -837,7 +837,7 @@ export interface ExecOptions extends child.ExecOptions {
 
 	/**
 	 * Exit when command return code is non-zero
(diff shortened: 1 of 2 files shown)


types/shelljs/index.d.ts
```diff
@@ -835,6 +835,13 @@ export interface ExecOptions extends child.ExecOptions {
 	 */
 	silent?: boolean;
 
+	/**
+	 * Exit when command return code is non-zero
```
gkalpak: Missing trailing period (`.`).

types/shelljs/index.d.ts
```diff
@@ -835,6 +835,13 @@ export interface ExecOptions extends child.ExecOptions {
...
+	 *
+	 * @default false
```
gkalpak: Actually, this (as well as `ailent` above) defaults to the value set in the ["global" config][1].

[1]: https://documentup.com/shelljs/shelljs#configfatal

gkalpak (changes requested):
Good catch :+1:

types/shelljs/index.d.ts
pbelbin: Hey there George (@gkalpak), 

I agree with your observation, *but*, the documentation at the shelljs website says it defaults to `false`.

I *guess* the documentation for these options at the shelljs website presumes that the global settings are not altered from their defaults.  That they are altered by the global config may well be mentioned somewhere else.

So: are you just making an observation, or are you going to *require* this be changed before it will be accepted?

pbelbin:
Hey there folks.  The benchmark check is failing because it seems to be trying to install TypeScript 3.7 which it can not find.

```
npm ERR! code ETARGET
```

types/shelljs/index.d.ts
gkalpak: Fair enough. Let's keep it as is. If anything, the docs should be fixed as well.

gkalpak (approved):
LGTM

PR #40049, opened 2019-10-31
relay-runtime: Add overloads to proxies

mattkrick:
Goals:
- no breaking changes
- reduce the number of null checks & nonnull coercions 
- extra type info when you want it. just add a generic, e.g. `commitMutation<AddCommentMutation>` and type info will trickle down to your updater

Example:

  ```gql
graphql`
mutation AddComment($newComment: NewCommentInput!) {
  addComment(newComment: $newComment) {
    error {
      message
    }
    title
    body
    topicId
  }
}
`
```

```ts
updater: (store) => {
  store.getRootField('addCommentz') // returns `RecordProxy<{}> | null` instead of full shape, 
  const payload = store.getRootField('addComment') // OK
  payload.getLinkedPayload('random') // doesn't return full shape
  payload.getValue('tittle') // returns union of all possibilities
  const topicId = payload.getValue('topicId') // string  
  const topic = store.get<ITopic>(topicId) // it can take a hint! these types can be generated using gql2ts or similar
}
```

PS: I'd _love_ to have typos in prop names turn into red squiggles, but I'm not smart enough to do that and make the `passToHelper` test pass ☹️  

NOTE: 

This code generates the following error when running `yarn lint relay-runtime`:
```
Error: /Users/mattkrick/code/DefinitelyTyped/types/relay-runtime/lib/store/RelayStoreTypes.d.ts:368:46
ERROR: 368:46  no-unnecessary-generics  Type parameter T is used only once. See: https://github.com/Microsoft/dtslint/blob/master/docs/no-unnecessary-generics.md
```

This is an error with the linting rule. See https://github.com/Microsoft/dtslint/issues/212

### DIFF
--- a/types/relay-runtime/lib/store/RelayStoreTypes.d.ts
+++ b/types/relay-runtime/lib/store/RelayStoreTypes.d.ts
@@ -294,10 +294,10 @@ export interface Store {
  */
 export type Scheduler = (callback: () => void) => void;
 
-type MaybeNull<T> = null extends T ? null : never;
-type Unarray<T> = T extends Array<infer U> ? U : T;
-type NullableRecord<T> = RecordProxy<NonNullable<T>> | MaybeNull<T>;
-type DeepNullable<T> = {
+export type MaybeNull<T> = null extends T ? null : never;
+export type Unarray<T> = T extends Array<infer U> ? U : T;
+export type NullableRecord<T> = RecordProxy<NonNullable<T>> | MaybeNull<T>;
+export type DeepNullable<T> = {
     [P in keyof T]: T[P] extends Array<infer U>
         ? Array<DeepNullable<U>> | null
         : T[P] extends ReadonlyArray<infer U>
@@ -331,7 +331,7 @@ export interface RecordProxy<T = { [key: string]: any }> {
       name: K,
       args?: Variables | null,
     ): RecordProxy;
-    setValue<K extends keyof T>(value: any, name: K, args?: Variables | null): RecordProxy;
+    setValue(value: any, name: keyof T, args?: Variables | null): RecordProxy;
 }
 
 export interface ReadOnlyRecordProxy {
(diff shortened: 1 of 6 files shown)


alloy (changes requested):
Nice one 👌

Could you add your example as a test case please?

alloy:
Btw, can you also add a comment to disable the failing linter rule?

mattkrick:
@alloy not sure how to make the other tests pass, but this should be good to go.

this was tested on a pretty large codebase & the changes didn't cause additional errors, but YMMV.

alloy (approved):

alloy:
Oh btw, I just now see your comment about wanting to assert type failures in the tests. There’s examples of the magic comment that’s used for that in the react-relay typings’ tests, iirc.

types/relay-runtime/lib/store/RelayStoreTypes.d.ts
```diff
@@ -347,9 +379,11 @@ export interface ReadOnlyRecordSourceProxy {
  * Extends the RecordSourceProxy interface with methods for accessing the root
  * fields of a Selector.
  */
-export interface RecordSourceSelectorProxy extends RecordSourceProxy {
-    getRootField(fieldName: string): RecordProxy | null | undefined;
-    getPluralRootField(fieldName: string): Array<RecordProxy | null | undefined> | null | undefined;
+
+export interface RecordSourceSelectorProxy<T = {}> extends RecordSourceProxy {
+    getRootField<K extends keyof T>(fieldName: K): RecordProxy<NonNullable<T[K]>>;
+    getRootField(fieldName: string): RecordProxy | null;
```
taion: @mattkrick @alloy Why did we remove `| undefined` here? It's what the upstream Flow types declare:

https://github.com/facebook/relay/blob/7fd936ac9eddb44b6f463affa03679a53a6d0561/packages/relay-runtime/store/RelayStoreTypes.js#L378

alloy: Hmm, yep indeed–my bad for forgetting about flow’s `?Foo` details.

mattkrick: ah, that's my bad, must've gotten a little too excited.
just outta curiosity, how do you get that to return undefined?? i thought the server spec only allowed null?

taion: Oh, I don't know really whether it ever does. It's just marked as such in Flow. I only noticed because the removal of `undefined` here caused me to get type checking failures from different versions of this typing lib being installed.

PR #40057, opened 2019-10-31
[mapbox-gl] Add missing public properties and method for Mapbox-gl

life777:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/mapbox/mapbox-gl-js/releases/tag/v1.5.0>>

I found a few properties and methods that were added before but missed here:
1) Property `workerCount`. [docs](https://docs.mapbox.com/mapbox-gl-js/api/#workercount)
2) Property `maxParallelImageRequests`. [docs](https://docs.mapbox.com/mapbox-gl-js/api/#maxparallelimagerequests)
3) Method `clearStorage`. [docs](https://docs.mapbox.com/mapbox-gl-js/api/#clearstorage)


life777:
This failure isn't caused by my commit. It tries to load typescript 3.7 and failed. Could we restart this step?:
Screenshot from 2019-10-31 18-05-20

dmytro-gokun:
Hi @life777! Thank you for your PR. There are two notes i'd like to make here before accepting this.
1) This partially overlaps with https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39961, which is still in process. Let's wait for that to be merged first, so there is no merge conflict.
2) Please add some tests for signatures you've added (except those already covered by https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39961).

jessetrinity:
The build error should be resolved. We're still having trouble with BenchmarkPR but its success is not required.

dmytro-gokun:
@life777 Thanks for the update! This looks really good. But we need some tests added. I know, the changes look trivial. Still, it's better to add some minimal test cases.

dmytro-gokun (approved):

PR #40068, opened 2019-11-01
[@types/jest]: improve definition of MatcherUtils

G-Rath:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/jest/blob/master/packages/jest-matcher-utils/src/index.ts

### DIFF
--- a/types/jest/index.d.ts
+++ b/types/jest/index.d.ts
@@ -453,6 +453,8 @@ declare namespace jest {
 
     type MatcherColorFn = ChalkFunction & { supportsColor: ChalkColorSupport };
 
+    type EqualityTester = (a: any, b: any) => boolean | undefined;
+
     interface MatcherUtils {
         readonly isNot: boolean;
         readonly dontThrow: () => void;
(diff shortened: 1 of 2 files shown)


types/jest/index.d.ts
```diff
@@ -424,33 +424,79 @@ declare namespace jest {
...
+            readonly INVERTED_COLOR: import("chalk").Chalk & {
+              supportsColor: import("chalk").ColorSupport;
+            };
+            readonly BOLD_WEIGHT: import("chalk").Chalk & {
+              supportsColor: import("chalk").ColorSupport;
+            };
+            readonly DIM_COLOR: import("chalk").Chalk & {
+              supportsColor: import("chalk").ColorSupport;
+            };
+            readonly SUGGEST_TO_CONTAIN_EQUAL: string;
```
Favna: The chalk import should go the top of the file and only be imported once

favna (changes requested):
Just one change

types/jest/index.d.ts
G-Rath: Awesome, cheers.

G-Rath: @Favna That doesn't seem to be working.

Is there a disadvantage to doing how it was previously aside from being a bit repetitive, and if not, is this something we could fix/improve straight after this PR instead of blocking here?

I don't mind making the changes, but I'd like to get these published for usage ideally asap & am aware of how busy people can be :)

favna (approved):
LGTM. I tried out my suggestion of before as well and it was a pretty big mess - but the solution you applied since also works.

types/jest/index.d.ts
Favna: Confirmed it didn't work. Didn't *expect* it to break that badly.

G-Rath:
Cheers - I don't know if it's that chalk is a special case, but I figured for now that would be the cleanest solution, while making it easy to tackle on it's own afterwards 🤷‍♀

favna (approved):

PR #40076, opened 2019-11-01
Add type definitions for blip-sdk

henriquetorquato:


sandersn (changes requested):
Can you please add tests? It's a little odd that a package on npm uses a namespace and not a module.

henriquetorquato:
The definition exports only interface for this package, the only way i could think of testing an interface definition is by testing the implementation itself, so i decided to remove the test.

As for the namespace export, i used a type definition for another package that's common for this library as a reference ([this one](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/lime-js)), i just kept the same definition style, to maintain code consistency.

sandersn (approved):

PR #40091, opened 2019-11-02
Bugfix: use "export =" instead of "export default" - @types/uid-generator

kylechine:
### Motivation

With ES6 module syntax in Node.js:

```typescript
export { UIDGenerator as default, UIDGeneratorClass, UIDGeneratorInstance };
```

Below code will not be compiled correctly:

```typescript
import { default as UIDGenerator } from 'uid-generator';

new UIDGenerator(128, UIDGenerator.BASE58);

// will be compiled to JavaScript as:
const uid_generator1 = require("uid-generator");
new uid_generator1.default(128, uid_generator1.default.BASE58);

// Correct version should be:
const uid_generator1 = require("uid-generator");
new uid_generator1(128, uid_generator1.BASE58);
```

### Solution

Use "export =" syntax. Refer to <https://www.typescriptlang.org/docs/handbook/modules.html>

### DIFF
--- a/types/uid-generator/index.d.ts
+++ b/types/uid-generator/index.d.ts
@@ -28,4 +28,4 @@ interface UIDGeneratorClass {
 
 declare const UIDGenerator: UIDGeneratorClass;
 
-export { UIDGenerator as default, UIDGeneratorClass, UIDGeneratorInstance };
+export = UIDGenerator;
--- a/types/uid-generator/uid-generator-tests.ts
+++ b/types/uid-generator/uid-generator-tests.ts
@@ -1,4 +1,4 @@
-import { default as UIDGenerator } from 'uid-generator';
+import UIDGenerator = require('uid-generator');
 
 new UIDGenerator('abc'); // $ExpectType UIDGeneratorInstance
 const generator = new UIDGenerator(128, 'abc'); // $ExpectType UIDGeneratorInstance


types/uid-generator/index.d.ts
```diff
@@ -28,4 +28,4 @@ interface UIDGeneratorClass {
 
 declare const UIDGenerator: UIDGeneratorClass;
 
-export { UIDGenerator as default, UIDGeneratorClass, UIDGeneratorInstance };
+export = UIDGenerator;
```
TheEmrio: Maybe we could do something like this:
```ts
declare namespace UIDGenerator {
  interface UIDGeneratorInstance {
    readonly bitSize: number;
    readonly uidLength: number;
    readonly baseEncoding: string;
    readonly base: number;
    generateSync(): string;
    generate(): Promise<string>;
    generate(cb: (error: Error|null, uid: string) => any): void;
  }

  interface UIDGeneratorClass {
    new (bitSize?: number, baseEncoding?: string): UIDGeneratorInstance;
    new (baseEncoding?: string): UIDGeneratorInstance;
    readonly BASE16: '0123456789abcdef';
    readonly BASE36: '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZ';
    readonly BASE58: '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz';
    readonly BASE62: '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz';
    readonly BASE66: '0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-._~';
    readonly BASE71: "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz!'()*-._~";
    readonly BASE94: "!\"#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]^_`abcdefghijklmnopqrstuvwxyz{|}~";
  }
}

declare const UIDGenerator: UIDGenerator.UIDGeneratorClass;

export = UIDGenerator;
``` 
With this, the package would use the correct `export =` while still having its types

Emrio (changes requested):

types/uid-generator/index.d.ts
kylechine: Default-export should be forbidden indeed. :-D

I think maybe it does not necessary to export Interfaces in practice:

Typically, user will write something like this:

```typescript
import UIDGenerator =  require('uid-generator');
const uidgen    = new UIDGenerator(bitLength, UIDGenerator.BASE58);
let uid = uidgen.generateSync();
uidgen.generate(uid=>{
  // uid = string
});
```

The situation when explicit type is needed could be very rare:

```typescript
import UIDGenerator =  require('uid-generator');
const uidgen:UIDGenerator.UIDGeneratorInstance    = new UIDGenerator(bitLength, UIDGenerator.BASE58);
let uid = uidgen.generateSync();
uidgen.generate(uid=>{
  // uid = string
});
```

Plus, because of we already defined returning type at: 

```typescript
new (bitSize?: number, baseEncoding?: string): UIDGeneratorInstance;
new (baseEncoding?: string): UIDGeneratorInstance;
```

TypeScript already knows the returning type structure. (I have tested, my atom editor understands both the class instance constants and methods)

So basically, I tend to believe explicit export types could be unnecessary.

The drawback of using a namespace is that it imports a global namespace. There is no guarantee that another package would never use the same namespace.

TheEmrio: I see what you mean, and I think you're right. I'll approve this version then :)

Emrio (approved):

PR #40094, opened 2019-11-02
[react-slider] Creates generic and updates types

waynevanson:
There were two render functions missing alongside a missing generic, so there is no error handling for `typeof newState === 'number'` required when using the `onChange` handler.

Please fill in this template.

+ <https://zillow.github.io/react-slider/>
+ <https://github.com/zillow/react-slider/blob/master/src/components/ReactSlider/ReactSlider.jsx#L74-L260>


types/react-slider/index.d.ts
```diff
@@ -1,36 +1,140 @@
-// Type definitions for react-slider 0.8
+// Type definitions for react-slider 1.0
 // Project: https://github.com/mpowaga/react-slider
 // Definitions by: Jason Unger <https://github.com/jsonunger>
-// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+//                 Wayne Van Son <https://github.com/waynevanson>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped/types/react-slider
+// TypeScript Version: 3.6
```
sandersn: What features of typescript 3.6 does this require? Moving from 2.8 to 3.6 means that this update will only be available by default to people on 3.6 or higher. Everybody else will be stuck with the old version.

types/react-slider/react-slider-tests.tsx
```diff
@@ -1,15 +1,18 @@
 import * as React from 'react';
 import ReactSlider from 'react-slider';
```
sandersn: Use the commonjs syntax `import ReactSlider = require('react-slider')`, and change `"esModuleInterop": false` in tsconfig.

Consumers of the package who want to use `import ReactSlider from 'react-slider'` will need to add `"esModuleInterop": true`, but Definitely Typed tests need to be backward compatible to show that the types will work with old code bases.

sandersn (changes requested):

waynevanson:
Sorry about this! I'll fix it up now. Thank you for your feedback.

waynevanson:
I'm having an error with the react imports.

I used this import syntax
```js
import React = require('react');
```
And this is what I got when running `npm run lint react-slider`:

``` terminal
../react/index.d.ts(33,22): error TS2307: Cannot find module 'csstype'.
```

If @bjorgvin is still working then we may as well just use his one.

PR #40096, opened 2019-11-02
feat: Add `globalThis`

ExE-Boss:
[The&nbsp;`globalThis`&nbsp;polyfill](https://github.com/es-shims/globalThis) implements a&nbsp;cross‑platform compatible implementation of&nbsp;the&nbsp;[ES2020&nbsp;`globalThis`&nbsp;property](https://tc39.es/ecma262/#sec-globalthis).

---

Please fill in this template.

If adding a new definition:

---

review?(@ljharb)

### DIFF
--- a/types/globalthis/index.d.ts
+++ b/types/globalthis/index.d.ts
@@ -1,0 +1,29 @@
+// Type definitions for globalthis 1.0
+// Project: https://github.com/ljharb/System.global#readme
+// Definitions by: ExE Boss <https://github.com/ExE-Boss>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.4
+
+/**
+ * Gets the `globalThis` object.
+ */
+declare function getGlobal(): typeof globalThis;
+
+declare namespace getGlobal {
+	/**
+	 * Gets the `globalThis` object.
+	 */
+	function getPolyfill(): ReturnType<typeof import('./polyfill')>;
+
+	/**
+	 * The `globalThis` object.
+	 */
+	const implementation: typeof import('./implementation');
+
+	/**
+	 * Installs the `globalThis` property onto the global object.
+	 */
+	function shim(): ReturnType<typeof import('./shim')>;
+}
+
+export = getGlobal;
--- a/types/globalthis/globalthis-tests.ts
+++ b/types/globalthis/globalthis-tests.ts
@@ -1,0 +1,6 @@
+import getGlobal = require('globalthis');
+
+getGlobal(); // $ExpectType typeof globalThis
+getGlobal.implementation; // $ExpectType typeof globalThis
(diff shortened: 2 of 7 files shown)


types/globalthis/index.d.ts
```diff
@@ -0,0 +1,29 @@
+// Type definitions for globalthis 1.0
+// Project: https://github.com/ljharb/System.global#readme
+// Definitions by: ExE Boss <https://github.com/ExE-Boss>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.4
```
ExE-Boss: Support for&nbsp;`globalThis` was added in&nbsp;[**TypeScript&nbsp;3.4**](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#type-checking-for-globalthis).

ljharb: it’s the same type as `windows or `global`; were these types not in earlier versions?

ExE-Boss: There is&nbsp;`declare var window`, but&nbsp;that&nbsp;requires `lib: ["dom"]`, and&nbsp;`declare var global`, which&nbsp;requires&nbsp;`@types/node`, but&nbsp;only the&nbsp;former one&nbsp;has, and&nbsp;this&nbsp;only exists in&nbsp;**TypeScript&nbsp;3.4+**:
```ts
declare var window: Window & typeof globalThis;
declare var self: Window & typeof globalThis;
```

Node’s `global` only has:
```ts
declare var global: NodeJS.Global;
```

---

Also, even if you do:
```ts
var something;
```

Even though it&nbsp;will be&nbsp;available as&nbsp;`window.something` during&nbsp;runtime, without&nbsp;`globalThis`, the&nbsp;compiler wouldn’t&nbsp;know&nbsp;that.

---

Also, in&nbsp;`lib.dom.d.ts`, `Window`&nbsp;functions have&nbsp;to&nbsp;duplicated in&nbsp;the&nbsp;global&nbsp;scope, so&nbsp;that&nbsp;you&nbsp;can&nbsp;do:
```ts
addEventListener("foo", e => {
	// Process the `foo` event
})
```

Same&nbsp;with&nbsp;Node&nbsp;globals.

ljharb (approved):
seems like there should be a better thing than `typeof globalThis` for older TS versions; perhaps `typeof (function () { return this; }())`?

ExE-Boss:
@ljharb That&nbsp;won't&nbsp;work for&nbsp;several&nbsp;reasons:

1. `.d.ts`&nbsp;files can’t&nbsp;have&nbsp;function&nbsp;bodies, since&nbsp;they&nbsp;just&nbsp;define the&nbsp;type&nbsp;signatures.
2. The&nbsp;resulting&nbsp;type of&nbsp;`(function () { return this; })();` is&nbsp;`any`.

PR #40097, opened 2019-11-02
[@types/bull] Add getter signature to job.progress function

douglascayers:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/OptimalBits/bull/blob/develop/lib/job.js#L122

### DIFF
--- a/types/bull/index.d.ts
+++ b/types/bull/index.d.ts
@@ -171,6 +171,11 @@ declare namespace Bull {
     returnvalue: any;
 
     /**
+     * Get progress on a job
+     */
+    progress(): any;
+
+    /**
      * Report progress on a job
      */
     progress(value: any): Promise<void>;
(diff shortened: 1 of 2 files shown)


types/bull/index.d.ts
```diff
@@ -170,6 +170,11 @@ declare namespace Bull {
 
     returnvalue: any;
 
+    /**
+     * Get progress on a job
+     */
+    progress(): any;
```
douglascayers: This method signature is specified separately from `progress(value: any): Promise<void>` specifically so that the return type can be expressed as the synchronous result `any` and not `Promise<any>`, because the Job object doesn't return a resolved promise, it returns the actual value.

bondz (approved):

PR #40105, opened 2019-11-03
libxmljs: get on element with namespaces

ghost:
I added [this](https://github.com/libxmljs/libxmljs/wiki/Element#elementgetxpath-ns_uri-1) function.

From the help:
> Return the first node that matches the xpath This is syntactic sugar for find(xpath, namespaces)[0]

This was not possible with the current definitions.

### DIFF
--- a/types/libxmljs/index.d.ts
+++ b/types/libxmljs/index.d.ts
@@ -2,6 +2,7 @@
 // Project: https://github.com/libxmljs/libxmljs
 // Definitions by: François de Campredon <https://github.com/fdecampredon>
 //                 ComFreek <https://github.com/ComFreek>
+//                 Roeland <https://github.com/r03>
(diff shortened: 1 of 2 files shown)


ComFreek:
@r03 Thanks for the PR and sorry I missed reviewing this. I'd like to add that I don't see how your added `get(xpath: string, namespaces: StringMap): Element|null` signature is mentioned in the [docs you linked](https://github.com/libxmljs/libxmljs/wiki/Element#elementgetxpath-ns_uri-1). In case the docs are just outdated, would you be willing to just add it to libxmljs' docs? Just to keep everything consistent.

r03:
Yes, my link to the doc is correct, but there is an error in the doc title. It should be namespaces instead of ns_uri.

PR #40106, opened 2019-11-03
[webpack] Updates definitions for Chunk and Module

ofhouse:
Hi,

I made some some contributions to two compilation classes to improve the types for them.

In this PR:
* Updates definitions for [`webpack.compilation.Chunk`](https://github.com/webpack/webpack/blob/v4.41.2/lib/Chunk.js)
* Adds definitions for [`webpack.compilation.Module`](https://github.com/webpack/webpack/blob/v4.41.2/lib/Module.js)

The types are mostly copy-pasted from the JS-doc from the linked source.

  - https://github.com/webpack/webpack/blob/v4.41.2/lib/Chunk.js
  - https://github.com/webpack/webpack/blob/v4.41.2/lib/Module.js

### DIFF
--- a/types/webpack/index.d.ts
+++ b/types/webpack/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for webpack 4.39
+// Type definitions for webpack 4.41
 // Project: https://github.com/webpack/webpack
 // Definitions by: Qubo <https://github.com/tkqubo>
 //                 Benjamin Lim <https://github.com/bumbleblym>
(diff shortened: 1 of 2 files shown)


ofhouse:
Seems like travis failed internally, trying another force-push to retrigger the build...
https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/606893464?

types/webpack/index.d.ts
```diff
@@ -18,6 +18,7 @@
 //                 Grgur Grisogono <https://github.com/grgur>
```
PranavSenthilnathan: I know this PR isn't updating the types for a newer version of webpack, but I think it would still be a good idea to bump up the version at the top of this file to 4.41. This PR has a few breaking changes and if someone does get broken they can go back to the 4.39 type definitions.

PranavSenthilnathan (changes requested):

types/webpack/index.d.ts
ofhouse: Thanks for the feedback 👍
Agree with that and increased the version to `4.41` accordingly.

I've also checked if there were new changes to introduced in this release to the affected types but nothing changed there.

PR #40116, opened 2019-11-04
fix(object-types): add quoteStyle option

Akuukis:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [changelog](https://github.com/inspect-js/object-inspect/releases/tag/v1.5.0) and [readme](https://www.npmjs.com/package/object-inspect)


demurgos (approved):
Thanks for the PR

PR #40118, opened 2019-11-04
Node: URL: correct format/parse parameter types

OliverJAsh:
Please fill in this template.

- - https://nodejs.org/api/url.html#url_url_format_url_options
- - https://nodejs.org/api/url.html#url_url_parse_urlstring_parsequerystring_slashesdenotehost

### DIFF
--- a/types/node/url.d.ts
+++ b/types/node/url.d.ts
@@ -23,7 +23,7 @@ declare module "url" {
         hash: string | null;
         host: string | null;
         hostname: string | null;
-        href: string | null;
+        href: string;
         path: string | null;
         pathname: string | null;
         protocol: string | null;
--- a/types/node/url.d.ts
+++ b/types/node/url.d.ts
@@ -27,7 +27,7 @@ declare module "url" {
         path: string | null;
         pathname: string | null;
         protocol: string | null;
-        search: string | null | null;
+        search: string | null;
         slashes: boolean | null;
         port: string | null;
         query: string | null | ParsedUrlQuery;
--- a/types/node/url.d.ts
+++ b/types/node/url.d.ts
@@ -1,29 +1,36 @@
 declare module "url" {
     import { ParsedUrlQuery, ParsedUrlQueryInput } from 'querystring';
 
-    interface UrlObjectCommon {
-        auth?: string;
-        hash?: string;
-        host?: string;
-        hostname?: string;
-        href?: string;
-        path?: string;
-        pathname?: string;
-        protocol?: string;
-        search?: string;
-        slashes?: boolean;
-    }
-
     // Input to `url.format`
-    interface UrlObject extends UrlObjectCommon {
-        port?: string | number;
+    interface UrlObject {
+        auth?: string | null;
+        hash?: string | null;
+        host?: string | null;
+        hostname?: string | null;
+        href?: string | null;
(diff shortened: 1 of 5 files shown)


galkin (commented):
@OliverJAsh, thank you for your work.
I highlight author of affected packages. I will approve your pull after feedback from them.

types/node/url.d.ts
```diff
@@ -1,29 +1,36 @@
...
+        slashes?: boolean | null;
+        port?: string | number | null;
         query?: string | null | ParsedUrlQueryInput;
     }
 
     // Output of `url.parse`
-    interface Url extends UrlObjectCommon {
-        port?: string;
-        query?: string | null | ParsedUrlQuery;
+    interface Url {
```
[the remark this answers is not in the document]
OliverJAsh: I thought about doing this, but the types are slightly different (e.g. `port` cannot be a `number` here)—so I don't think we could use that here.

types/node/url.d.ts
```diff
@@ -1,29 +1,36 @@
...
+        auth: string | null;
+        hash: string | null;
+        host: string | null;
+        hostname: string | null;
+        href: string | null;
```
ikokostya: Could you provide case when `href` can be `null`? In the following tests `href` is always `string`:

```sh
$ node --version
v10.16.0
$ node -p 'url.parse("")'
Url {
  protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: null,
  pathname: null,
  path: null,
  href: '' }
$ nvm use 8
Now using node v8.16.2 (npm v6.4.1)
$ node -p 'url.parse("")'
Url {
  protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: null,
  pathname: null,
  path: null,
  href: '' }
```

OliverJAsh: Good catch, I've reverted that particular change.

galkin: Could you squash your commits into one?

OliverJAsh: Is that really necessary? I've never had to do that before when contributing to DefinitelyTyped.

sandersn: Squash merge is the default for DT maintainers so I don't think it's necessary for contributors to squash them ahead of time.

types/node/url.d.ts
```diff
@@ -1,29 +1,36 @@
...
+    interface Url {
+        auth: string | null;
+        hash: string | null;
+        host: string | null;
+        hostname: string | null;
+        href: string;
+        path: string | null;
+        pathname: string | null;
+        protocol: string | null;
+        search: string | null | null;
```
Huachao: string | null

ExE-Boss:
```suggestion
	        search: string | null;
```

galkin:
@OliverJAsh, could you make squash for your commits?

OliverJAsh:
@galkin That's not necessary: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40118#discussion_r342116396

galkin:
Thank you i missed the comment above.

PR #40120, opened 2019-11-04
[spotify-web-playback-sdk] Extend Image interface with width and height properties

hanna-becker:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: repo is still private

### DIFF
--- a/types/spotify-web-playback-sdk/index.d.ts
+++ b/types/spotify-web-playback-sdk/index.d.ts
@@ -4,6 +4,7 @@
 //                 Marcus Weiner <https://github.com/mraerino>
 //                 Moritz Gunz <https://github.com/NeoLegends>
 //                 Daniel Almaguer <https://github.com/deini>
+//                 Hanna Becker <https://github.com/hanna-becker>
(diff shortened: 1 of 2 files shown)


types/spotify-web-playback-sdk/index.d.ts
```diff
@@ -33,7 +34,9 @@ declare namespace Spotify {
     type ErrorTypes = 'account_error' | 'authentication_error' | 'initialization_error' | 'playback_error';
 
     interface Image {
+        height: number;
         url: string;
+        width: number;
```
NeoLegends: As per https://developer.spotify.com/documentation/web-api/reference/object-model/#image-object, `height` and `width` may be `null` or `undefined`, so this should be `height?: number | null` and `width?: number | null`.

NeoLegends (changes requested):

types/spotify-web-playback-sdk/index.d.ts
hanna-becker: good point!

mraerino (approved):
Looks good! 🎉

NeoLegends (approved):

PR #40132, opened 2019-11-04
[@types/google-maps-react] Make Marker and InfoWindow components extandable.

MityaSavchenko:
Add missing event handlers for marker props.
Add missing methods and variables for Marker and InfoWindow classes.


types/google-maps-react/index.d.ts
```diff
@@ -94,15 +103,21 @@ export class HeatMap extends React.Component<any, any> {
 }
 
 export interface InfoWindowProps extends Partial<google.maps.InfoWindowOptions> {
-  google: typeof google;
-  map: google.maps.Map;
-  marker: google.maps.Marker;
+  google?: typeof google;
```
gburgett: I believe these properties are intended to be required.  Would you please look up the documentation and determine if they are optional?

types/google-maps-react/index.d.ts
```diff
@@ -71,14 +71,23 @@ export interface MarkerProps extends Partial<google.maps.MarkerOptions> {
...
+  onMouseup?: markerEventHandler;
+  onRecenter?: markerEventHandler;
 }
 
 export class Map extends React.Component<MapProps, any> {
 }
 
-export class Marker extends React.Component<MarkerProps, any> {
+export class Marker<P extends MarkerProps = MarkerProps, S = any> extends React.Component<P, S> {
+  marker?: google.maps.Marker;
```
gburgett: When you create a `Marker` class, is it possible for this to not exist?

gburgett (changes requested):
Please double check on those props, whether they are optional or required.

types/google-maps-react/index.d.ts
MityaSavchenko: And should i fix it?

MityaSavchenko: > I believe these properties are intended to be required. Would you please look up the documentation and determine if they are optional?
And should i fix it?

types/google-maps-react/index.d.ts
gburgett: Ok this one's fine then.

types/google-maps-react/index.d.ts
gburgett: Right, you're supposed to pass them like this:

```tsx
public render() {
  return <div>
    <InfoWindow google={google} map={map} marker={marker} />
  </div>
}
```

And Typescript ought to tell you that you're missing one of them.

MityaSavchenko: Okey. Thanks. Fixed that

gburgett (approved):

gburgett:
@MityaSavchenko Thanks for addressing my concerns!

PR #40146, opened 2019-11-05
Fix typings for tsc esModuleInterop=false

zorji:
Fix typings when the compilation options don't have "esModuleInterop" enabled.

The existing typings only works when "esModuleInterop" is enabled in the compilation options. i.e.

```
import promiseRetry from 'promise-retry'
```

However, some repositories don't have "esModuleInterop" enabled for compatibility reasons. i.e.

```
import * as promiseRetry from 'promise-retry'
```

And the error `error TS2497: This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.` is raised on a compilation.

This change fixes the error and makes it also work in the environment doesn't' have "esModuleInterop" enabled.

### DIFF
--- a/types/promise-retry/index.d.ts
+++ b/types/promise-retry/index.d.ts
@@ -24,4 +24,5 @@ type RetryableFn<ResolutionType> = ((retry: (error: any) => never, attempt: numb
  */
 declare function promiseRetry<ResolutionType>(retryableFn: RetryableFn<ResolutionType>, options?: OperationOptions): Promise<ResolutionType>;
 declare function promiseRetry<ResolutionType>(options: OperationOptions, retryableFn: RetryableFn<ResolutionType>): Promise<ResolutionType>;
+declare namespace promiseRetry {};
 export = promiseRetry;
--- a/types/promise-retry/index.d.ts
+++ b/types/promise-retry/index.d.ts
@@ -24,5 +24,5 @@ type RetryableFn<ResolutionType> = ((retry: (error: any) => never, attempt: numb
  */
 declare function promiseRetry<ResolutionType>(retryableFn: RetryableFn<ResolutionType>, options?: OperationOptions): Promise<ResolutionType>;
 declare function promiseRetry<ResolutionType>(options: OperationOptions, retryableFn: RetryableFn<ResolutionType>): Promise<ResolutionType>;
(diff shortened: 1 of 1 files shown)


PranavSenthilnathan:
I think the current definition is correct and the usage in those libraries may be wrong. Without the esModuleInterop flag they should be doing importing the library like:
```
import promiseRetry = require('promise-retry');

promiseRetry(...)
```

PranavSenthilnathan (changes requested):
As mentioned in the comment, I think it was fine before. Could you share an example (import as well as use-site) of when this change is needed?

zorji:
> I think the current definition is correct and the usage in those libraries may be wrong. Without the esModuleInterop flag they should be doing importing the library like:
> 
> ```
> import promiseRetry = require('promise-retry');
> 
> promiseRetry(...)
> ```

@PranavSenthilnathan thanks for reviewing

Here is my usage and I am getting error `error TS2497: This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export.`

```
import * as promiseRetry from 'promise-retry'
```

Using `import promiseRetry = require('promise-retry');` doesn't throw the error but using `require` instead of `import` lose all the typing.

Do you mind give it a try?

zorji:
Here is the minimal step to reproduce the issue

```
mkdir example
cd example
npm init -y
npm i -S typescript promise-retry @types/promise-retry ts-node @types/node
npx tsc --init
echo "import * as promiseRetry from 'promise-retry'" > index.ts
npx tsc index.ts
```

PranavSenthilnathan:
> Using import promiseRetry = require('promise-retry'); doesn't throw the error but using require instead of import lose all the typing.

Where do you lose the typing? If you have `@types/promise-retry` then this will be well typed.

The library is not using the ES6 style export so the import shouldn't be an ES6 import either (it should be using the `require` statement above).

zorji:
Hi @PranavSenthilnathan 
I just tested it using `require()` and it works. Sorry I used to thought `import` was the only correct to import modules in TypeScript.

I also found the documentation in case anyone interested: https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

PR #40148, opened 2019-11-05
[cosmiconfig] Remove package

ffflorian:
Please fill in this template.

If removing a declaration:

See https://github.com/davidtheclark/cosmiconfig/blob/6.0.0/package.json#L6.


ffflorian:
Blocked by https://github.com/microsoft/types-publisher/pull/695.

ffflorian:
@sandersn unfortunately Travis still thinks that `cosmiconfig` is not among the whitelisted dependencies although https://github.com/microsoft/types-publisher/pull/695 was merged.

sandersn:
Weird, I thought I cleared the cache. I will try again.

PranavSenthilnathan:
This PR does remove the v4 typings, which people may be depending on. @ozum @szeck87 @saadq @jinwoo @bmatcuk could you take a look?

bmatcuk (commented):
Well, removing this package doesn't delete the existing `@types/cosmiconfig` packages from npm, right? So, if someone is depending on the v4 or v5 types, they should be fine.

My only concern is that postcss-load-config currently depends on cosmiconfig `^5.0.0`, not v6, so the changes to the postcss-load-config types is not _technically_ correct.

sandersn:
@bmatcuk they remain but you can't ship fixes to them since the source is gone.

The types haven't changed since June 2018, though, so I doubt 4.x or 5.x will need any more changes.

If postcss-load-config compiles with the version of 6 it gets as of today, I'd recommend pinning an exact version in its package.json. It only depends on the ExplorerOptions types, so I'd guess it's pretty safe anyway; people tend not to delete things from options types, only add them.

bmatcuk:
Oh ya, that's true... but, as you mention, haven't been any changes in a long time, so probably ok.

Re: postcss-load-config: As long as the v6 options are identical to the v5 ExplorerOptions, and the version is pinned in package.json, I'm ok with that.

PR #40150, opened 2019-11-05
Fix: rox-react-native `fetch` type

glenna:
Updated the `fetch` function type -- it's not an async function. Tested the change in our repo.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: It's not an async method: https://support.rollout.io/docs/javascript-browser-api#section-fetch


g-guirado (approved):
We would like fetch() to return a Promise in the future; but this is indeed not the case yet. Therefore, 👍

glenna:
@g-guirado oh sweet, you're from rollout! thanks! 👍

PR #40155, opened 2019-11-05
react-intl-redux: Add formats prop to IntlState

steos:
The updateIntl opts parameter can have a formats prop which is missing in the interface.
see https://github.com/ratson/react-intl-redux#formatting-data


PranavSenthilnathan:

PR #40158, opened 2019-11-05
Update split.js destroy and drag methods

pgherveou:
-  Updated destroy method (added boolean parameters)
see:  https://www.npmjs.com/package/split.js#destroypreservestyles--false-preservegutters--false
- Updated onDrag methods (added sizes parameter)
see: https://github.com/nathancahill/split/tree/master/packages/splitjs#ondrag-ondragstart-ondragend
- Updated header version to 1.4 since destroy options have been introduced in 1.4.x

### DIFF
--- a/types/split.js/index.d.ts
+++ b/types/split.js/index.d.ts
@@ -83,6 +83,6 @@ declare namespace Split {
     collapse(index: number): void;
 
     // Destroy the instance. It removes the gutter elements, and the size CSS styles Split.js set.
-    destroy(): void;
+    destroy(preserveStyles?: boolean, preserveGutters?: boolean): void;
   }
 }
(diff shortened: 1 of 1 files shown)


icholy:
Looks like the types have been merged upstream. Would be better to update them there first. https://github.com/nathancahill/split/blob/master/packages/splitjs/index.d.ts

pgherveou:
@icholy good point will send a PR over there

pgherveou:
see: https://github.com/nathancahill/split/pull/214

icholy:
@sandersn how do we go about deprecating a package's types?

sandersn:
`npm run not-needed`. I think you'll need:

`npm run not-needed -- split.js 1.4 https://github.com/nathancahill/split splitjs`

But I haven't tested it, you'll have to `npm test` (or let CI run on the PR).

PR #40162, opened 2019-11-05
Update Roarr types to allow function argument to `child'.

Aankhen:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/gajus/roarr#function-parameter

I noticed that [the library has its own Flow type definitions](https://github.com/gajus/roarr/blob/master/src/types.js), and they don’t quite line up with the types we have here. I’ve stuck to the style of the types here for now, but is there a way to align them without breaking backwards compatibility?

### DIFF
--- a/types/roarr/index.d.ts
+++ b/types/roarr/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for roarr 2.13
+// Type definitions for roarr 2.14
 // Project: https://github.com/gajus/roarr#readme
 // Definitions by: Philip Saxton <https://github.com/psaxton>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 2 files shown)


psaxton:
Thank you for taking the time to submit a pull request. This looks good.

This was my first contribution to this project and could certainly be improved. So long as moving to the form type definitions works with the actual ROARR implementation, I would be happy to review such changes.

psaxton (approved):

Aankhen:
Thanks for the approval, @psaxton! It’s not that the types don’t work, it’s just that there’s a bit of mismatch. I’m not sure whether aligning them would be possible, because it would technically break backwards-compatibility. I do think that the types that are inside `namespace Roarr` should be moved out of them, because right now there isn’t any way to refer to them in clients.

PR #40167, opened 2019-11-05
[react-relay] Handle all variations of nullability and arrays for `useFragment`

janicduplessis:
Please fill in this template.

The previous implementation did not detect nullability properly and also missed handling 2 cases for arrays (`[X]!` and `[X!]`). 

To do this we always use the `TKey` generic parameter which should be the exact type of `XFragment_foo$key`. Nullability is determined by the type of the `fragmentRef` parameter (the fragment type doesn't actually encode nullability anyway). For arrays the key type will already be an array so we have to handle then slightly differently. We have one overload for each possible combinaison of nullability (2 for regular fields, 4 for arrays).

- Type `ReturnType` is the non-nullable type for a regular field key. Nullability is added on usage.
- Type `NonNullableArrayReturnType` is the non-nullable array type for an array field key. Nullability of the **field** is added on usage.
- Type `NullableArrayReturnType` is the nullable array type for an array field key. Nullability of the **field** is added on usage.

Tested all 6 possible combinaisons of array / nullability with `useFragment<XFragment_foo$key>(...)` and validated that the ts type returned matched the nullability / and whether it's an array of the schema.

### DIFF
--- a/types/react-relay/lib/relay-experimental/useFragment.d.ts
+++ b/types/react-relay/lib/relay-experimental/useFragment.d.ts
@@ -9,13 +9,12 @@ import { GraphQLTaggedNode } from 'relay-runtime';
 
 type $Call<Fn extends (...args: any[]) => any> = Fn extends (arg: any) => infer RT ? RT : never;
 
-type KeyType = { readonly ' $data'?: unknown  }
+type KeyType = { readonly ' $data'?: unknown };
+type ArrayKeyType = ReadonlyArray<{ readonly ' $data'?: ReadonlyArray<unknown> }>;
 
 type ReturnType<T extends KeyType> = (arg: T) => NonNullable<T[' $data']>;
-type NonNullableArrayReturnType<T extends ReadonlyArray<KeyType>> = (arg: T) => ReadonlyArray<NonNullable<T[0][' $data']>>;
-type NullableArrayReturnType<T extends ReadonlyArray<KeyType>> = (
-    arg: T,
-) => ReadonlyArray<NonNullable<T[0][' $data']> | null>;
+type NonNullableArrayReturnType<T extends ArrayKeyType> = (arg: T) => ReadonlyArray<NonNullable<T[0][' $data']>[0]>;
+type NullableArrayReturnType<T extends ArrayKeyType> = (arg: T) => ReadonlyArray<NonNullable<T[0][' $data']>[0] | null>;
 
 export function useFragment<TKey extends KeyType>(
     fragmentInput: GraphQLTaggedNode,
@@ -27,22 +26,22 @@ export function useFragment<TKey extends KeyType>(
     fragmentRef: TKey | null,
 ): $Call<ReturnType<TKey>> | null;
 
-export function useFragment<TKey extends ReadonlyArray<KeyType>>(
+export function useFragment<TKey extends ArrayKeyType>(
     fragmentInput: GraphQLTaggedNode,
     fragmentRef: TKey,
 ): $Call<NonNullableArrayReturnType<TKey>>;
 
-export function useFragment<TKey extends ReadonlyArray<KeyType>>(
+export function useFragment<TKey extends ArrayKeyType>(
     fragmentInput: GraphQLTaggedNode,
     fragmentRef: TKey | null,
 ): $Call<NonNullableArrayReturnType<TKey>> | null;
 
-export function useFragment<TKey extends ReadonlyArray<KeyType>>(
+export function useFragment<TKey extends ArrayKeyType>(
     fragmentInput: GraphQLTaggedNode,
     fragmentRef: ReadonlyArray<TKey[0] | null>,
 ): $Call<NullableArrayReturnType<TKey>>;
 
-export function useFragment<TKey extends ReadonlyArray<KeyType>>(
+export function useFragment<TKey extends ArrayKeyType>(
     fragmentInput: GraphQLTaggedNode,
     fragmentRef: ReadonlyArray<TKey[0] | null> | null,
 ): $Call<NullableArrayReturnType<TKey>> | null;
--- a/types/react-relay/lib/relay-experimental/useFragment.d.ts
+++ b/types/react-relay/lib/relay-experimental/useFragment.d.ts
@@ -12,14 +12,10 @@ type $Call<Fn extends (...args: any[]) => any> = Fn extends (arg: any) => infer 
 interface KeyType {
     readonly ' $data'?: unknown;
 }
-type ArrayKeyType = ReadonlyArray<{ readonly ' $data'?: ReadonlyArray<unknown> }>;
-type NullableArrayKeyType = ReadonlyArray<{ readonly ' $data'?: ReadonlyArray<unknown> } | null>;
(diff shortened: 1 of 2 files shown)


renanmav:
Related with https://github.com/relay-tools/relay-compiler-language-typescript/issues/157

alloy:
Nice! It would be great to have some test coverage for these changes.

janicduplessis:
@renanmav Actually it is not, not sure how we can get that type of inference working without changing the compiled artifacts.

@alloy Was trying to figure out how to do that :) Fixed the one for nullability and added one for arrays. I'll add a few other ones for all missing nullability / array combinations in a min.

alloy:
@janicduplessis You can actually also test failures, there’s a magic comment for that (on my phone and don’t know it off-hand, but check other React-Relay tests).

janicduplessis:
@alloy I was able to test useFragment with non array using type assertions but for array types it returned different types in different ts versions and wasn't able to get lint passing so I didn't add it there.

Ended up being harder than expected to get it working for array of nullable items, for some reason typescript had issues inferring those types and got `unknown[]`. I had to add 2 extra overloads so it works for both passing an explicit annotation (`useFragment<Foo_bar$key>()`) and the infered version.

janicduplessis:
@jstejada How do you handle arrays of nullable types for fragment components prop types with Flow?

For a regular fragment component we'd use

```ts
type Props = {
  bar: Foo_bar$key
}
// type of fragment will be `Foo_bar`

// or

type Props = {
  bar: Foo_bar$key | null
}
// type of fragment will be `Foo_bar | null`
```

but for a nullable array the key type does not encode the nullability

```ts
type Props = {
  bars: Foo_bars$key
}
// type of fragment will be `ReadonlyArray<{...}>`

// ???
type Props = {
  bar: ReadonlyArray<Foo_bar$key[0] | null>
}
// type of fragment will be `ReadonlyArray<{...} | null>`
```

alloy:
@janicduplessis I wish I could be of assistance here, but as I haven’t been using these APIs yet and currently don’t have the time to do so I’m alas of little help 😞

Let me know when this is mergeable, though.

janicduplessis:
@alloy Not sure if it is possible to make the typedef simpler but it does work with all the cases I tested so I think this is good to merge.

jstejada:
@janicduplessis 

sorry if I'm misunderstanding:

```javascript
// ???
type Props = {
  bar: ReadonlyArray<Foo_bar$key[0] | null>
}
// type of fragment will be `ReadonlyArray<{...} | null>`
```

When would this case ever occur? or when would you declare your props as such? at least in Flow, if the fragment is plural, we generate the key type as a `$ReadOnlyArray<{...}>`, so the elements are never nullable; when the parent reads a plural fragment ref, we filter out any null refs, so that's why we always guarantee an array of non nullable values

janicduplessis:
@jstejada Ok I see. So if I have the following:

```graphql
type Query {
  nodes(ids: [ID!]!): [Node]! 
}
```

```ts
const query = useLazyLoadQuery<AppQuery>(graphql`
  query AppQuery($ids: [ID!]!) {
    nodes(ids: $ids) {
       ...App_nodes
    }
  }
`);
// type of query is { nodes: ReadonlyArray<{$fragmentRefs: ... } | null> }
const nodes = useFragment(graphql`
  fragment App_nodes on Node @relay(plural: true) {
    id
  }
`, query.nodes);
// type of nodes is ReadonlyArray<{id: string}>
```
So in this case when useFragment reads the query it will filter out nulls automatically if I understand correctly? If so I can simplify the typedef quite a bit :)

jstejada:
@janicduplessis 

that is correct. specifically, in Flow, we will generate this:

```javascript
// Parent data type
export type AppQueryResponse = {|
  +nodes: $ReadOnlyArray<{|
    +$fragmentRefs: AppTest_nodes$ref
  |}>
|};
```

```javascript
// Fragment $key type

export type AppTest_nodes$key = $ReadOnlyArray<{
  +$data?: AppTest_nodes$data,
  +$fragmentRefs: AppTest_nodes$ref,
}>;

```

Note that the type of `nodes` can flow into `AppTest_nodes$key`, so e.g. the following shouldn't produce an error:

```javascript
declare var pluralRef: $ReadOnlyArray<{|
    +$fragmentRefs: AppTest_nodes$ref
  |};

(pluralRef: AppTest_nodes$key) // no error
```

janicduplessis:
@jstejada Wouldn't flow generate the following type? I checked and this what typescript does for an array of nullable types.

```js
// Parent data type
export type AppQueryResponse = {|
  +nodes: $ReadOnlyArray<{|
    +$fragmentRefs: AppTest_nodes$ref
  |} | null> // <==
|};
```

jstejada:
@janicduplessis, no, we don't generate nullable plural refs as I mentioned earlier. 

that's an actual generated type that I copy-pasted

janicduplessis:
@jstejada Interesting, we should probably do the same with the typescript compiler. For now I can make it swallow the nulls at the `useFragment` level.

Will update this PR so plural fragment values can never be null.

alloy:
> Interesting, we should probably do the same with the typescript compiler.

:+1:

I also wonder if this applies to the HOCs, which should be largely the same as the hooks, I’d expect.

jstejada:
@alloy @janicduplessis 

just a small clarification after chatting with @kassens, I misspoke in my earlier comments, and this is what we actually do:

- We filter out null refs when a plugarl ref (array of refs) is passed to useFragment or a Fragment Container
-The Flow types we generate for plural results will match the schema definition of the plural. So if it's defined in the schema e.g. as `[Int]`, we would generate the type of the result as an array of nullable ints. *However*, our schema at Facebook does not have any nullable plurals, the server will guarantee that null values are filtered out from the plural. Given this setup, it so happens that all of the types for plural fields we generate internally will be arrays of non-nullable elements.

We think the typescript type generator should do the same. apologies for the confusion!

alloy:
> Given this setup, it so happens that all of the types for plural fields we generate internally will be arrays of non-nullable elements.

So if I understand this correctly; FB just happens to define all their lists as `[Int!]`, but we should keep emitting arrays with nullable members when a schema defines `[Int]`?

jstejada:
> So if I understand this correctly; FB just happens to define all their lists as [Int!], but we should keep emitting arrays with nullable members when a schema defines [Int]?

that's exactly correct (in fact fb defines all lists as `[Int!]!`, so technically the server will always return an array).

but yes, the correct thing to do is to generate arrays with nullable members if the schema defines it as such

alloy:
Great, thanks–so @janicduplessis that means this needs amending, right?

jstejada:
not sure if it needs amending, but technically the types should reflect the following:

- useFragment technically /can/ receive a plural ref (array of fragment refs) that contains null fragment refs
- when reading data, useFragment will skip the null refs, so the return type of useFragment will still technically be an array of non-nullable members.

janicduplessis:
@alloy @jstejada Alright, this is what I understood in the last update I made to this PR so this is fine, no amending needed. The typedef for `useFragment` can accept a plural ref but will always return an array of non-nullable members.

alloy:
Ahh gotcha, cool 👍

PR #40183, opened 2019-11-06
Types stripe invoice and tax fixes

Stilborg:
Please fill in this template.

- [] Add or edit tests to reflect the change. (Run with `npm test`.)

https://stripe.com/docs/api/tax_rates/create 
https://stripe.com/docs/api/invoices/object


types/stripe/index.d.ts
```diff
@@ -6449,6 +6449,91 @@ declare namespace Stripe {
              */
             percentage: number | null;
         }
+
+        interface ITaxRateCreationOptions {
+            /**
+             * The display name of the tax rate, which will be shown to users.
+             */
+            display_name: string;
+            /**
```
pushplay: Please include a blank link above comments.

types/stripe/index.d.ts
```diff
@@ -6449,6 +6449,91 @@ declare namespace Stripe {
...
+            description?: string;
+            /**
+             * The jurisdiction for the tax rate.
+             */
+            jurisdiction?: string;
+            /**
+             * Set of key-value pairs that you can attach to an object. This can be useful for storing additional information about the object in a structured format.
+             * Individual keys can be unset by posting an empty value to them. All keys can be unset by posting an empty value to metadata.
+             */
+            metadata?: IMetadata;
```
pushplay: `metadata` in Options interfaces should be `IOptionsMetadata` so they can accept numbers.

types/stripe/index.d.ts
```diff
@@ -6449,6 +6449,91 @@ declare namespace Stripe {
...
+        }
+
+        interface ItaxRateSearchOptions {
```
pushplay: Extend `IListOptions` to remove duplication.

types/stripe/index.d.ts
```diff
@@ -10280,6 +10374,51 @@ declare namespace Stripe {
             ): Promise<checkouts.sessions.ICheckoutSession>;
         }
 
+        class TaxRates extends StripeResource {
+            /** Create TaxRate: https://stripe.com/docs/api/tax_rates/create */
```
pushplay: Please copy the documentation for the action.

pushplay (changes requested):
Thanks for the improvements.  There are a few formatting issues.  If you can add some tests that would also be appreciated.

Stilborg:
Thanks @pushplay, good solid feedback as always 👍 
I have implemented your requested changes

pushplay:
That's a great question.  I really don't know and I find that aspect of DefinitelyTyped quite confusing.  It would be nice if it was in the contribution docs.  I just look at it in IntelliJ and rely on that to tell me what's up.

types/stripe/index.d.ts
```diff
@@ -6449,6 +6449,97 @@ declare namespace Stripe {
...
+            display_name?: string;
+            /**
+             * The jurisdiction for the tax rate.
+             */
+            jurisdiction?: string;
+            /**
+             * Set of key-value pairs that you can attach to an object. This can be useful for storing additional information about the object in a structured format.
+             * Individual keys can be unset by posting an empty value to them. All keys can be unset by posting an empty value to metadata.
+             */
+            metadata?: IMetadata;
```
pushplay: This should also be `IOptionsMetadata` and the whitespacing should be fixed on this interface as well.

types/stripe/index.d.ts
```diff
@@ -8499,15 +8595,19 @@ declare namespace Stripe {
...
+            default_payment_method: string | paymentMethods.IPaymentMethod | null;
 
             /**
              * ID of the default payment source for the subscription.
              * It must belong to the customer associated with the subscription and be in a chargeable state.
              * If not set, defaults to the customer’s default source. [Expandable]
              */
             default_source: string | null;
-
+            /**
```
pushplay: And white space here.

pushplay (changes requested):
Thanks for the fixes.  Just needs to be taken the rest of the way and it's good to go.

vaskevich:
@Stilborg I think you want to run `npm run lint stripe`. Those tests in there aren't actually executable tests, they're just there for types.

types/stripe/index.d.ts
Stilborg: Done

types/stripe/index.d.ts
Stilborg: @pushplay Ok done, I do not know if I agree with this whitespace rule though. 
It is not enforced by the linter and the default prettier behaviour is: Collapse multiple blank lines to one blank line and otherwise leave the line breaks alone.

Having this rule on top of that, just seems to slow down the process.

what do you think @vaskevich ?

Stilborg:
> @Stilborg I think you want to run `npm run lint stripe`. Those tests in there aren't actually executable tests, they're just there for types.

@vaskevich Ah yes, that is obviously right, I see the usage of the tests now. I validate them using tsc.
I do run the linter and find that combining it with prettier, drastically reduce the problems I have had in the past with white spaces.

pushplay:
I would be be more than happy (prefer) to adjust the tslint rules to maintain this.  The style should be kept consistent and the more tooling can do that for us automatically the better.

I've skimmed the tslint core rules and I'm not sure what the rule would be.  I would expect there to be one.  I know FxCop enforces this.

Stilborg:
> I would be be more than happy (prefer) to adjust the tslint rules to maintain this. The style should be kept consistent and the more tooling can do that for us automatically the better.
> 
> I've skimmed the tslint core rules and I'm not sure what the rule would be. I would expect there to be one. I know FxCop enforces this.

I agree, I looked at the prettier docs, and they do not have this. I would prefer that running prettier results in something that will always pass the linter, if possible. I this case In would vote for not having this rules.

But we can talk about that 'in another episode'.

Stilborg:
@pushplay & @vaskevich Anything else for me to do here before we can release? I kinda need this out to be able to finish a task.

Stilborg:
@pushplay What is the next step here? This is ready to go, right?

vaskevich (approved):

Stilborg:
Thank you @vaskevich, I fixed the merge conflict and it is ready to go.
But this thill has the _revision needed_ tag, so I'm assuming that it will not be published?
Does @pushplay have to do something?

pushplay (approved):
Once vaskevich approved it I assumed it was good.  🤷‍♀

Stilborg:
It is all good @pushplay , so did I :)

PR #40186, opened 2019-11-06
Added type definitions for n-readlines (node-readlines)

DiefBell:
Please fill in this template.

### DIFF
--- a/types/n-readlines/index.d.ts
+++ b/types/n-readlines/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for socket.io-file 1.0
+// Type definitions for n-readlines 1.0
 // Project: https://github.com/nacholibre/node-readlines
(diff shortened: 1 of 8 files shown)


types/n-readlines/index.d.ts
```diff
@@ -0,0 +1,23 @@
+// Type definitions for socket.io-file 1.0
```
sandersn: This should be n-readlines.

Side note: How did this pass the CI tests?

sandersn (changes requested):
Header is incorrect

types/n-readlines/index.d.ts
MerrickKing: Ah good spot, I copy pasted this from my previous one, thought I'd changed it all but I guess it was just the version number I did

sandersn (approved):

PR #40189, opened 2019-11-06
[chai] Fixed OwnProperty interface. Added optional `value` attribute.

bparan:
Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.chaijs.com/api/bdd/#method_property. At the very end of the section it mentions that assertions .ownProperty and .haveOwnProperty can be used interchangeably with .own.property

If removing a declaration:

### DIFF
--- a/types/chai/index.d.ts
+++ b/types/chai/index.d.ts
@@ -11,6 +11,7 @@
 //                 Satana Charuwichitratana <https://github.com/micksatana>
 //                 Erik Schierboom <https://github.com/ErikSchierboom>
 //                 Rebecca Turner <https://github.com/9999years>
+//                 Bogdan Paranytsia <https://github.com/bparan>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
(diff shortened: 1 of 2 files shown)


types/chai/index.d.ts
```diff
@@ -330,7 +331,7 @@ declare namespace Chai {
     }
 
     interface OwnProperty {
-        (name: string, message?: string): Assertion;
+        (name: string, value?: any, message?: string): Assertion;
```
jedmao: Shouldn't this be an overload instead of replacing the existing signature?

jednano (changes requested):

types/chai/index.d.ts
bparan: I basically made it the same as Property interface. I'm not sure what reads better 2 optional parameters or 2 overloaded methods with optional parameters.
I'm good with adding overloaded method. What do you think of adding the method to Property interface and delete OwnProperty interface?

jedmao: Technically, what you've done here will still work, but the overload is more expressive as to what the middle argument represents in the case that you only provide 2 arguments. That's why I would suggest the overload. If the `Property` interface works the same, perhaps it should have an overload too.

bparan: I made adjustments to the `Property` interface and deleted `OwnProperty` interface.

jednano (approved):

PranavSenthilnathan:

PR #33127, opened 2019-02-17
feat(node): v11.10

SimonSchick:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: 
  - https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V11.md#11.10.0 
  - https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V11.md#2019-01-17-version-1170-current-bridgear
  - https://nodejs.org/docs/latest-v11.x/api/child_process.html#child_process_options_stdio
- [x] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

Added child process `connected` property.
Made all `childProcess` properties readonly (aside from std*).
Made `childProcess.stdio` an array rather than a tuple.
Made `create(De)Cipher` IV nullable.
Added `process.features`.
Added `http(s|2)` `writeHead` return value.
Added `perf_hooks.monitorEventLoopDelay`.
Added `replServer.setupHistory`.
Added `tls` `session` event for socket.
Added `brotli` to crypto.
Generally moved affected tests to `/test` folder

Fixes #33125
Fixes #32965

PR #40198, opened 2019-11-06
node: Add forked child process pipe methods

frangio:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: Unfortunately this isn't documented very clearly. Here is a link where the function is mentioned: https://github.com/nodejs/node/issues/9313. This is the code where the function is implemented: https://github.com/nodejs/node/blob/6072e01c938e22498cd46c3d8a2c21a16a632456/src/handle_wrap.cc#L151

### DIFF
--- a/types/node/stream.d.ts
+++ b/types/node/stream.d.ts
@@ -307,7 +307,9 @@ declare module "stream" {
             ): Promise<void>;
         }
 
-        interface Pipe { }
+        interface Pipe {
+            unref(): void;
+        }
     }
 
     export = internal;
--- a/types/node/stream.d.ts
+++ b/types/node/stream.d.ts
@@ -308,6 +308,9 @@ declare module "stream" {
         }
 
         interface Pipe {
+            close(): void;
+            hasRef(): boolean;
+            ref(): void;
             unref(): void;
         }
     }
--- a/types/node/child_process.d.ts
+++ b/types/node/child_process.d.ts
@@ -7,7 +7,7 @@ declare module "child_process" {
         stdin: Writable | null;
         stdout: Readable | null;
         stderr: Readable | null;
-        readonly channel?: Pipe;
+        readonly channel?: Pipe | null;
         readonly stdio: [
             Writable | null, // stdin
             Readable | null, // stdout
--- a/types/node/child_process.d.ts
+++ b/types/node/child_process.d.ts
@@ -7,7 +7,7 @@ declare module "child_process" {
         stdin: Writable | null;
         stdout: Readable | null;
         stderr: Readable | null;
-        readonly channel?: Pipe | null;
+        readonly channel?: Pipe;
         readonly stdio: [
             Writable | null, // stdin
             Readable | null, // stdout
--- a/types/node/test/child_process.ts
+++ b/types/node/test/child_process.ts
@@ -46,6 +46,7 @@ import { Writable, Readable, Pipe } from 'stream';
         execArgv: ['asda']
     });
     const ipc: Pipe = forked.channel!;
+    ipc.unref();
 }
 
 async function testPromisify() {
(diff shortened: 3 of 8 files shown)


types/node/stream.d.ts
```diff
@@ -307,7 +307,9 @@ declare module "stream" {
             ): Promise<void>;
         }
 
-        interface Pipe { }
+        interface Pipe {
+            unref(): void;
```
galkin:
```suggest
hasRef(): boolean;
close(): void
ref(): void;
unref(): void;
```

galkin (changes requested):
Thank you for your pull! I made code review

frangio:
Thank you @galkin for the review! As a sidenote, I don't understand why `Pipe` is in the `stream` module. If you run `require('stream').Pipe` you get `undefined`. It seems that `Pipe` is only used in `child_process`, wouldn't it be better to define it there? Or is it that since this is a type whose constructor isn't exposed, we just have to decide where to place it and `stream` sounded reasonable?

frangio:
I noticed that `ChildProcess.channel` was marked as optionally `null`. I couldn't find this documented anywhere. According to the [docs](https://nodejs.org/api/child_process.html#child_process_subprocess_channel) it can only be `undefined`. So I've changed it accordingly.

frangio:
I don't understand the linter errors here. The `channel` property is optional, the non-null assertion should be necessary.

Flarna:
@frangio please see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33127#discussion_r257522417 regarding `channel` and `pipe` - in special why it was typed as empty interface only.
I agree that `stream` is a strange place.

I can't remember what I did that time to get `null` in `channel` but there is at least following line in code: https://github.com/nodejs/node/blob/v12.13.0/lib/internal/child_process.js#L585

frangio:
Oh I see! Thanks for explaining @Flarna. I have reverted that change.

galkin (approved):

frangio:
Can someone help me understand the linter error?

```typescript
        const ipc: stream.Pipe = forked.channel!;
```

```
ERROR: 2119:34  no-unnecessary-type-assertion  This assertion is unnecessary
since it does not change the type of the expression.
```

However, the expression is nullable.

```typescript
        readonly channel?: stream.Pipe | null;
```

I don't understand how it could type without the non null assertion.

Flarna:
v10/tsconfig.json for has `strictNullChecks: false` therefore `Pipe | null` is the same as `Pipe`.

frangio:
Thanks!

types/node/stream.d.ts
```diff
@@ -307,7 +307,12 @@ declare module "stream" {
             ): Promise<void>;
         }
 
-        interface Pipe { }
+        interface Pipe {
+            close(): void;
+            hasRef(): boolean;
+            ref(): void;
+            unref(): void;
```
Flarna: I think `ref()` and `unref()` usually return `this`.
Not sure if all variants of Pipes/Handles used in child process use case actually have a `close()` API and I'm even not sure if it is a good idea to add them here as NodeJs doc tells `Accessing the IPC channel fd in any way other than process.send() or using the IPC channel with a child process that is not a Node.js instance is not supported.` (see [doc](https://nodejs.org/dist/latest-v13.x/docs/api/child_process.html#child_process_options_stdio) point 2).
In spawned process there is `process.disconnect()` to close `process.channel` (see [docs](https://nodejs.org/dist/latest-v13.x/docs/api/process.html#process_process_disconnect)) but it seems there is no `disconnect` on parent side.

frangio: > I think `ref()` and `unref()` usually return `this`.

In the case of this `Pipe` type they don't seem to.

> Accessing the IPC channel fd in any way other than ...

I don't think this applies because none of these method return the `fd`.

Flarna: well, they don't return the `fd` but I'm quite sure `close()` interacts with it.

Anyhow, I assume you use this in your app and it works there so it should be fine.

Flarna (approved):

PR #40200, opened 2019-11-07
[chart.js] Add weight property to ChartDataSets interface for Doughnut/Pie charts.

jordanweiler:
The Chart.js Doughnut/Pie chart can accept a `weight` property that will affect the ring weight but it wasn't yet supported by @types/chart.js so I created this PR. I also ran prettier on the index.d.ts file and added those changes to this PR. I put them on a separate commit to separate the actual change from the cosmetic changes.

Please let me know if you would prefer I remove the prettier fixes from this PR and put them on a different PR or just not do them at all.

  * https://www.chartjs.org/docs/latest/charts/doughnut.html
  * https://github.com/chartjs/Chart.js/blob/master/src/controllers/controller.doughnut.js

### DIFF
--- a/types/chart.js/index.d.ts
+++ b/types/chart.js/index.d.ts
@@ -26,18 +26,14 @@
 declare class Chart {
     static readonly Chart: typeof Chart;
     constructor(
-        context:
-            | string
-            | CanvasRenderingContext2D
-            | HTMLCanvasElement
-            | ArrayLike<CanvasRenderingContext2D | HTMLCanvasElement>,
-        options: Chart.ChartConfiguration,
+        context: string | CanvasRenderingContext2D | HTMLCanvasElement | ArrayLike<CanvasRenderingContext2D | HTMLCanvasElement>,
+        options: Chart.ChartConfiguration
     );
     config: Chart.ChartConfiguration;
     data: Chart.ChartData;
     destroy: () => {};
-    update: ({ duration, lazy, easing }?: Chart.ChartUpdateProps) => {};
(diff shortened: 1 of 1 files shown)


types/chart.js/index.d.ts
```diff
@@ -26,14 +26,18 @@
 declare class Chart {
     static readonly Chart: typeof Chart;
     constructor(
-        context: string | CanvasRenderingContext2D | HTMLCanvasElement | ArrayLike<CanvasRenderingContext2D | HTMLCanvasElement>,
-        options: Chart.ChartConfiguration
+        context:
```
ricardo-mello: @ElianCordoba should these formatting be reverted?

ElianCordoba: Yes please, all of them. Beside that the change looks good

ElianCordoba (changes requested):

types/chart.js/index.d.ts
jordanweiler: I've reverted all of those prettier changes on this PR so that only the weight property is on this PR.

Would you like me to add those prettier formatting changes on a separate PR or just ignore them altogether?

ElianCordoba (approved):
LGTM, thanks!

types/chart.js/index.d.ts
sandersn: As DT maintainer I'd like to see the prettier changes eventually, but I'll defer to @ElianCordoba and @ricardo-mello.

PR #40202, opened 2019-11-07
[jira-client]: Updated export to match version 6.13

rkesters:
Please fill in this template.

```
The babel output does not create an ES6 default export, instead it creates a commonjs default export. 
````
[Stack overflow](https://stackoverflow.com/questions/33505992/babel-6-changes-how-it-exports-default)

### DIFF
--- a/types/jira-client/index.d.ts
+++ b/types/jira-client/index.d.ts
@@ -1,7 +1,8 @@
-// Type definitions for jira-client 6.4
+// Type definitions for jira-client 6.13
 // Project: http://github.com/jira-node/node-jira-client
 // Definitions by: Anatoliy Ostapenko <https://github.com/KOPTE3>
 //                 Orta Therox <https://github.com/orta>
+//                 Robert Kesterson <https://github.com/rkesters>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
@@ -363,5 +364,4 @@ declare namespace JiraApi {
         intermediatePath?: string;
     }
 }
-
-export default JiraApi;
+ export = JiraApi;
--- a/types/jira-client/jira-client-tests.ts
+++ b/types/jira-client/jira-client-tests.ts
@@ -1,4 +1,4 @@
-import JiraApi from 'jira-client';
+import * as JiraApi from 'jira-client';
 
 // Initialize
 const jira = new JiraApi({

### CODE REVIEW

**Comment by rkesters** [comment]
@kopte3
@orta 

Please review

**Comment by orta** [comment]
This looks like a breaking change for everyone using jira-client's types right now, did they make a breaking change without a major server bump?

**Comment by rkesters** [comment]
> This looks like a breaking change for everyone using jira-client's types right now, did they make a breaking change without a major server bump?

The current types don't work for me on version 6.13.0, also when I revert to 6.4.3 the current types still do not work. 

This is how the export is done for both 6.13.0 and 6.4.3
```
exports.default = JiraApi;
module.exports = exports['default'];
```
This does not result in a ES6/TS default export,  for that it would be 

```
module.exports['default'] = exports['default'];
```

With the way the code is exported we need to use `export = `.

While the JS code does use `export default` the babel config is setup to make this a commonjs default. That is, the author of node-jira-client wants to do 

```
const JiraApi = require('jira-client') 
```
and not
```
const JiraApi = require('jira-client').default 
```

Therefore converting to `export = ` will allow, in TS

```
import * as JiraApi from 'jira-client'
```

**Comment by PranavSenthilnathan** [comment]
The package is exported so that both CommonJS and ES6 style import will work in JavaScript. However, in TypeScript we don't support *both* of them at the same time, so the current definitions chose to expose the package as an ES6 export (people can use default imports with the package). If users want to import this using `require` then they would need to do `require('jira-client').default` as you mentioned.

Changing the definition file to use `export =` will have the opposite problem. You will be able to import the package using just `require('jira-client')` but people that want to use ES6 style imports can't use default imports anymore.

Since it's just a matter of style, I'd suggest we don't change the definition because people might already depend on the current behavior. I would be interested in knowing why the current types don't work for you. I think both of these should work:

```
import JiraApi = require('jira-client').default;
import JiraApi from 'jira-client';
```

**Comment by rkesters** [comment]
@PranavSenthilnathan 

Either I'm miss-understanding something or not explaining my self very well.

```
import JiraApi = require('jira-client').default;
import JiraApi from 'jira-client';
```

The above does not work because the, actual JS in npm does not export a `default`. This is caused by how the babel is configured. This can be seen in [example](https://github.com/jira-node/node-jira-client#create-the-jira-client), 

For the above to work the babel'd transpilled code would need to be 

```
module.exports['default'] = exports['default'];
```

but instead,  transpilled code is

```
exports.default = JiraApi;
module.exports = exports['default'];
```

I generated the following using the current type via ts-node
Screenshot 2019-11-14 20 57 27

I generated the following using my PR
Screenshot 2019-11-14 20 48 44

PranavSenthilnathan:
Ah okay, I see, I'm getting the same behavior. Took a look at the package and the export in this PR is correct.

PranavSenthilnathan:

PR #40208, opened 2019-11-07
Updated definition to match gridstack v0.5

adumesny:
* MOST changes are for stricter lint issue - spacing, semicolon (file was compiled in my project source).
* added missing methods based on all methods in the code (fine comb trough each).
* added more code samples in doc

https://github.com/gridstack/gridstack.js/tree/develop/doc

this will become the base if/when I switch gridstack code base to TS now that I'm co-owner

If adding a new definition:

### DIFF
--- a/types/gridstack/gridstack-tests.ts
+++ b/types/gridstack/gridstack-tests.ts
@@ -1,6 +1,6 @@
 /// <reference types="jquery"/>
 
-var options: IGridstackOptions = {
+var options: GridstackOptions = {
     float: true
 };
 var element: JQuery = $(document).gridstack(options);
--- a/types/gridstack/index.d.ts
+++ b/types/gridstack/index.d.ts
@@ -90,6 +90,7 @@ interface GridStack {
 
     /**
     * Disables widgets moving/resizing. This is a shortcut for:
+    * @example
     *  grid.movable('.grid-stack-item', false);
     *  grid.resizable('.grid-stack-item', false);
     */
@@ -97,6 +98,7 @@ interface GridStack {
 
     /**
     * Enables widgets moving/resizing. This is a shortcut for:
+    * @example
     *  grid.movable('.grid-stack-item', true);
     *  grid.resizable('.grid-stack-item', true);
     */
@@ -105,7 +107,9 @@ interface GridStack {
     /**
     * Enables/disables widget moving.
     * This is a shortcut for:
+    * @example
     * grid.movable(this.container.children('.' + this.opts.itemClass), doEnable);
+    *
     * @param {boolean} doEnable
     * @param {boolean} includeNewWidgets will force new widgets to be draggable as per
     * doEnable`s value by changing the disableDrag grid option.
@@ -119,6 +123,7 @@ interface GridStack {
     * doEnable`s value by changing the disableResize grid option.
     *
     * This is a shortcut for:
+    * @example
     *  grid.resizable(this.container.children('.' + this.opts.itemClass), doEnable);
     */
     enableResize(doEnable: boolean, includeNewWidgets: boolean): void;
--- a/types/gridstack/index.d.ts
+++ b/types/gridstack/index.d.ts
@@ -292,19 +292,19 @@ interface JQuery {
   }
 
   /**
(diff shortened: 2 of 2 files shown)


adumesny:
@PascalSenn @ZoolWay @Sl1MBoy @JohnArcher

adumesny:
@radiolips

JohnArcher (commented):
Is the used/indicated TypeScript version 2.3 correct?

adumesny:
> Is the used/indicated TypeScript version 2.3 correct?
@JohnArcher you know I didn't change that or know how it's used... doing a `yarn list typescript` in that depot I get **typescript@3.8.0-dev.2019110** and I use this in my code with  typescript@3.1.6

How is that used and what should it be ? I'll be happy to make the change.

JohnArcher:
In order to improve the types further please add the following type:

```typescript
type GridStackWidgetType = string | HTMLElement | JQuery;
```

Then update the `el` parameter/type for the following methods:

```typescript
addWidget(el: GridStackWidgetType, ...)
locked(el: GridStackWidgetType, ...)
makeWidget(el: GridStackWidgetType)
maxWidth(el: GridStackWidgetType, ...)
minWidth(el: GridStackWidgetType, ...)
maxHeight(el: GridStackWidgetType, ...)
minHeight(el: GridStackWidgetType, ...)
movable(el: GridStackWidgetType, ...)
move(el: GridStackWidgetType, ...)
removeWidget(el: GridStackWidgetType, ...)
resize(el: GridStackWidgetType, ...)
resizable(el: GridStackWidgetType, ...)
update(el: GridStackWidgetType, ...)
```

JohnArcher:
> > Is the used/indicated TypeScript version 2.3 correct?
> 
> @JohnArcher you know I didn't change that or know how it's used... 
> [...]
> How is that used and what should it be ? I'll be happy to make the change.

Hm, I am not entirely sure. Have a look at this from the DT readme: https://github.com/DefinitelyTyped/DefinitelyTyped#i-want-to-use-features-from-typescript-29-or-above

I guess it depends what TypeScript features gridstack (or the definition? or both ...) uses. I know right now gridstack itself was not transformed to TypeScript yet. I am not sure whether we break something if we set the version number to a more current version (at least 3.0)!?

JohnArcher (commented):
This looks very good and works like a charm, no more warnings!

I just discovered a little styling problem. I use Visual Studio Code. If I hover over the `addWidget()` method in my code, the part where you print the code example looks messed up:

image

But if you add an `@example` tag right above the example code it really looks nice then:
image

Of course this is a minor issue, but still, this is quite nice with less effort.

I haven't looked for other examples in the definition, though.

adumesny:
> type GridStackWidgetType = string | HTMLElement | JQuery;

don't think we want word Type (things are explicitly Class | Type based on casing) and maybe have `WidgetStackElement` since it really is a dom element handle... the jquery one might go away in the future too.

adumesny:
> https://github.com/DefinitelyTyped/DefinitelyTyped#i-want-to-use-features-from-typescript-29-or-above

@JohnArcher looks like starting this month they only test (whatever that means) from 2.8 on (I still have an old  depot and they install and test from TS 2.0->3.7+next which passes with 
`1> gridstack START
1> gridstack OK`
whereas current build only does START but no return/test. Looks like we are not using anything in the definition that is >2.3 or it would fail locally. So I'm relunctant to change that without knowing what else/who it may break.

My plan is actually to include that d.ts in the next gridstack NPM build so users will not have to use this (much easier for me to rev it up as API changes in the lib), with hopefully longer term re-write lib to use TS anyway...

JohnArcher (approved):
Looks good now!

JohnArcher:
> should we have a type for `@param {number | string}` we have quite a few of those (which I need to verify as the code handles either in some cases...)

Not sure about this one ... In this case I tend to not introduce a new type, but as I said I am unsure. If it helps/simplifies development of gridstack, do it.

Regarding "TypeScript 2.3": I also think it is best to leave it that way, then. And it would indeed by nice to provide the definition with the gridstack release.

PranavSenthilnathan:

PR #40209, opened 2019-11-07
[styled-components] Update for TypeScript 3.7

Jessidhia:
Converts self-referential interfaces to recursive types (works around Microsoft/TypeScript#34796). Recursive types only exist in 3.7 so this uses typesVersions.

Finally enables code paths that were commented out because they'd break on TypeScript versions older than 3.2.

This replaces https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40064.

As this creates a `typesVersions` fork, I tried to separate the commits into one that just creates the fork, and did the changes in commits following it.

---

NOTE: testing locally is currently broken because of https://github.com/sandersn/dts-critic/issues/10

NOTE2: oops pushed to a branch in DT instead of in my own fork, CI is going to be annoying

### DIFF
--- a/types/styled-components/package.json
+++ b/types/styled-components/package.json
@@ -2,5 +2,9 @@
     "private": true,
     "dependencies": {
         "csstype": "^2.2.0"
+    },
+    "types": "index",
+    "typesVersions": {
+        ">=3.7.0-0": { "*": ["ts3.7/*"] }
     }
 }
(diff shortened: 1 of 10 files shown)


clentfort:
Thanks for all your amazing work on so many typings! :)

Jessidhia:
Interesting, I expected performance to get worse, not better, which is why I waited this long for the benchmark. It barely changed anyway.

PR #40212, opened 2019-11-07
[mongoose]  Modified methods "lean"

Iosif380:
The type of the returned parameter is determined from Any to T

Please fill in this template.

### DIFF
--- a/types/mongoose/index.d.ts
+++ b/types/mongoose/index.d.ts
@@ -2000,7 +2000,7 @@ declare module "mongoose" {
      * getters/setters or other Mongoose magic applied.
      * @param {Boolean|Object} bool defaults to true
      */
-    lean(bool?: boolean | object): Query<any> & QueryHelpers;
+    lean<T>(bool?: boolean | object): Query<T> & QueryHelpers;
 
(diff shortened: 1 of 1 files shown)


mingchen (approved):

types/mongoose/index.d.ts
```diff
@@ -2000,7 +2000,7 @@ declare module "mongoose" {
      * getters/setters or other Mongoose magic applied.
      * @param {Boolean|Object} bool defaults to true
      */
-    lean(bool?: boolean | object): Query<any> & QueryHelpers;
+    lean<T>(bool?: boolean | object): Query<T> & QueryHelpers;
```
Marsup: Shouldn't it be `lean(bool?: boolean | object): Query<T> & QueryHelpers;` ? (without the 1st <T>)
This seems to make all results `unknown` right now.

Iosif380: Yes of course. It is semantically more correct than type `any`.

Marsup: Indeed but what I meant is we do have the type, lean could be correctly (or as correctly as can be) typed instead of returning unknown.

PR #40214, opened 2019-11-07
Provide types for convert-units package

vladkampov:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/convert-units/convert-units-tests.ts
+++ b/types/convert-units/convert-units-tests.ts
@@ -1,0 +1,6 @@
+import convert = require('convert-units');
+
+const convertedMass = convert(25).from('mcg').to('t');
+const convertedMassBack =  convert(convertedMass).from('t').to('mcg');
+
+const unit = convert(66).getUnit<'mcg'>('mcg');
--- a/types/convert-units/tsconfig.json
+++ b/types/convert-units/tsconfig.json
@@ -1,0 +1,23 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6"
+        ],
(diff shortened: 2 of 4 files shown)


vladkampov:
fyi `convert-units` lib has types but it's seems like it's not maintained by it owner: latest changes with types was in April 2018 and hasn't been released to npm.

types/convert-units/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+        "strictFunctionTypes": true,
+        "strictNullChecks": true,
+        "baseUrl": "../",
+        "typeRoots": [
+            "../"
+        ],
+        "types": [],
+        "noEmit": true,
+        "forceConsistentCasingInFileNames": true,
+        "esModuleInterop": true
```
sandersn: this needs to be false for backward compatibility

types/convert-units/convert-units-tests.ts
```diff
@@ -0,0 +1,6 @@
+import convert from 'convert-units';
```
sandersn: Use the commonjs syntax `import convert = require('convert-units')`.

(You can still use `import convert from 'convert-units'` with `"esModuleInterop": true`.)

sandersn (changes requested):

types/convert-units/tsconfig.json
vladkampov: Fixed

types/convert-units/convert-units-tests.ts
vladkampov: Fixed

sandersn (approved):

PR #40218, opened 2019-11-07
[invariant] Assertion functions

josh:
Make use of TypeScript 3.7 assertion function feature to implement `invariant`.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions>

### DIFF
--- a/types/invariant/index.d.ts
+++ b/types/invariant/index.d.ts
@@ -2,7 +2,9 @@
 // Project: https://github.com/zertosh/invariant
 // Definitions by: MichaelBennett <https://github.com/bennett000>
 //                 dtinth <https://github.com/dtinth>
+//                 josh <https://github.com/josh>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.7
 
 declare let invariant:invariant.InvariantStatic;
 
@@ -12,7 +14,7 @@ export as namespace invariant;
 declare namespace invariant {
   interface InvariantStatic {
     (testValue:false, format?:string, ...extra:any[]):never;
-    (testValue:any, format?:string, ...extra:any[]):void;
+    (testValue:any, format?:string, ...extra:any[]):asserts testValue;
   }
 }
 


bennett000:
Looks good to me.

That said I'm not a reviewer (on the right sidebar) and I cannot seem to approve as @typescript-bot suggests

types/invariant/invariant-tests.ts
```diff
@@ -21,5 +21,5 @@ invariant('', 'Some other error');
 // handles extra variables
 invariant(true, 'Error, error, read all about it', 37, {}, 'hello');
 
-// $ExpectType {}
+// $ExpectType void | {}
```
dtinth: I believe this should stay as `{}` to facilitate this pattern:

```ts
const thing = getThing() || invariant(false, 'Expected getThing to return something')
```

as well as

```ts
const thing = getThing() ?? invariant(false, 'Expected getThing to return something')
```

types/invariant/index.d.ts
```diff
@@ -11,8 +13,7 @@ export as namespace invariant;
 
 declare namespace invariant {
   interface InvariantStatic {
-    (testValue:false, format?:string, ...extra:any[]):never;
```
dtinth: IMO we should keep this overload to not break existing test (see comment below).

types/invariant/index.d.ts
```diff
@@ -11,8 +13,7 @@ export as namespace invariant;
...
-    (testValue:any, format?:string, ...extra:any[]):void;
+    (testValue:any, format?:string, ...extra:any[]):asserts testValue;
```
dtinth: Would be great if some test can be added for this.

dtinth (changes requested):
Thanks for the PR!

I'm excited to use assertion functions in TypeScript as well. However I think we should retain support for `invariant (false)` pattern.

josh:
> I believe this should stay as `{}` to facilitate this pattern:

Is it possible for an `asserts` function to have a return value?

> However I think we should retain support for `invariant (false)` pattern.

👍 that makes sense.

dtinth (commented):
@josh Thanks! Mind adding a corresponding test for this change? Maybe a test could look like this:

```ts
declare const x: any
invariant(typeof x === 'string')
const s = x
// $ExpectType string
x
```

josh:
> Mind adding a corresponding test for this change? Maybe a test could look like this:

Tried that, but got:

```
expect  TypeScript@next expected type to be:
  string
got:
  any
```

PR #40223, opened 2019-11-07
styled-components: allow ThemeProvider to have multiple nodes as children

surgeboris:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/styled-components/styled-components/issues/1325#issuecomment-519400201

### DIFF
--- a/types/styled-components/index.d.ts
+++ b/types/styled-components/index.d.ts
@@ -414,7 +414,7 @@ export const withTheme: WithThemeFnInterface<DefaultTheme>;
 export interface DefaultTheme {}
 
 export interface ThemeProviderProps<T extends object, U extends object = T> {
-    children?: React.ReactChild; // only one child is allowed, goes through React.Children.only
+    children?: React.ReactNode;
     theme: T | ((theme: U) => T);
 }
 export type BaseThemeProviderComponent<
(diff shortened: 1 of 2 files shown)


eps1lon (changes requested):
It looks like this is only available [in 4.4 onwards](https://github.com/styled-components/styled-components/commit/560213155847ca3a546a1fbe1151b6aae15ef6a8). This means we need to bump the version in the header in index.d.ts as well.

surgeboris:
@eps1lon I wasnt't sure about this because this change does NOT break any old projects (they'll still continue to work with `ThemeProvider` having the only child). And for new projects it's reasonable to expect that they will start with the latest version, isn't it?

If you anyway think that we need a version bump - I'll do this, just let me know.

eps1lon:
>  I wasnt't sure about this because this change does NOT break any old projects

This change allows multiple children in 4.3 which will throw at runtime because of `React.children.only`.

It's not about breakage but only listing features that are available in the version that is specified in the header. If you use `styled-components@4.2.0` but `@types/styled-components@4.3` then features from 4.3 will not cause a type error but definitely should since they aren't available at runtime.

eps1lon (approved):

PR #40227, opened 2019-11-07
react-native-dialog: Make container's modal props optional

sandersn:
The tests use them in a way that indicates they are not required, so I added Partial around reactNativeModal.ModalProps.

This probably resulted from a recent update to react-native-modal.

I think react-native-dialog also needs [a second fix to react-native-modal's import syntax](https://github.com/react-native-community/react-native-modal/pull/372) before the build will pass.

Edit: That fix has now shipped.

### DIFF
--- a/types/react-native-dialog/index.d.ts
+++ b/types/react-native-dialog/index.d.ts
@@ -61,7 +61,7 @@ interface DescriptionProps {
     children: string;
 }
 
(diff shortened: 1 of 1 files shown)


sandersn:
I'm going to merge this to unblock react typings changes, but @MrLuje @stackbuilders @ibarrae @yaron1m if Partial is not the correct fix, please follow up with a better fix.

Issue #25758, opened 2018-05-14
[@types/lodash] `_.find` - type `number` is not assignable to type 'X'


th0r:
Here is a code snippet:
```ts
import _ from 'lodash';

interface Item {
  value: any;
}

export class Foo {
  items: Item[];
  selectedItems: Item[];

  updateSelected(selectedStr: string) {
    const selected = selectedStr.split(',');
    this.selectedItems = _.transform(selected, (items, value) => {
      const found = _.find(this.items, {value});
      if (found) {
        // Error (see below)
        items.push(found);
      }
    }, [] as Item[]);
  }
}

```

Line `items.push(found)` produces the following error:
```
Argument of type 'number | (<U>(callbackfn: (value: Item, index: number, array: Item[]) => U, thisArg?: any) => U[]...' is not assignable to parameter of type 'Item'.
  Type 'number' is not assignable to type 'Item'.
```

The fun thing is if I change `Item['value']` type to `string` the error goes away.

Versions:
- Tested with `typescript@2.7.2 / @2.8.3`
- `@types/lodash@4.14.108`

Mentions: @bczengel @chrootsu @stepancar @aj-r @Ailrun @e-cloud @thorn0 @jtmthf @DomiR

aj-r:
Looks like it's matching the wrong overload. It's supposed to match this:

```ts
        find<T>(
            collection: List<T> | null | undefined,
            predicate?: ListIterateeCustom<T, boolean>,
            fromIndex?: number
        ): T|undefined;
```

However it doesn't due to this error:

```
Argument of type '{ value: string; }' is not assignable to parameter of type 'string | [string, any] | ListIterator<Item, boolean> | PartialDeep<Item> | undefined'.
  Type '{ value: string; }' is not assignable to type 'PartialDeep<Item>'.
    Types of property 'value' are incompatible.
      Type 'string' is not assignable to type 'PartialDeep<any> | undefined'.
```

There are a few workarounds - adding `value as any` seems to fix the issue:

```ts
const found = _.find(this.items, {value: value as any});
```

or this:

```ts
    this.selectedItems = _.transform(selected, (items, value: any) => {
      const found = _.find(this.items, {value});
      if (found) {
        // Error (see below)
        items.push(found);
      }
    }, [] as Item[]);
```

It should be possible to fix this using conditional types in TS 2.8, but that would require us to drop support for all earlier versions of typescript, which we're not ready to do.

[...]

PR #40233, opened 2019-11-08
fix(lodash): commons partialDeep to support values (#25758)

Harshit369:
Please fill in this template.

### DIFF
--- a/types/lodash/common/common.d.ts
+++ b/types/lodash/common/common.d.ts
@@ -243,7 +243,7 @@ declare module "../index" {
     }
 
     type PartialDeep<T> = {
-        [P in keyof T]?: PartialDeep<T[P]>;
+        [P in keyof T]?: PartialDeep<T[P]> | T[P];
     };
 
     // For backwards compatibility


aj-r:
What is this trying to fix? Can you add a test that fails when this change is not applied?

aj-r (changes requested):
What is this trying to fix? Can you add a test that fails when this change is not applied?

Harshit369:
> What is this trying to fix? Can you add a test that fails when this change is not applied?

@aj-r for some reason my _.findIndex was picking definitions from version <3.0 and this iteratorShorthand used to throw an error: 
```ts
 _.findIndex({first: 'a', second: 'b', third: 'c'}, { first: 'a' })
```

because of ts path below
```ts
type PartialDeep<T> = {
    [P in keyof T]?: PartialDeep<T[P]>;
};

type IterateeShorthand<T> = PropertyName | [PropertyName, any] | PartialDeep<T>;

findIndex<T>(
   array: List<T> | null | undefined,
   predicate?: ListIterateeCustom<T, boolean>,
   fromIndex?: number
): number;
```

PartialDeep recursively keeps on checking its child value. but in tests, I found this and it doesn't break. **I wonder, how?**. 
```ts
_.findIndex(list, {a: 42}); // $ExpectType number
```

types for >3.1 drives similar behavior through this
```ts
type PartialShallow<T> = {
   [P in keyof T]?: T[P] extends object ? object : T[P]
};
```

aj-r:
Weird, I'm not sure what's going on then. Without a test to reproduce it, though, I'm not sure this is the right fix.

Harshit369:
@aj-r I think we can close this for now. something is weird here as I also found a few issues related to this. but somehow it works fine in tests. I'll raise another PR with supporting test cases once I catch the rat in the kitchen. Thanks for your time.

PR #40238, opened 2019-11-08
[indefinite] Fix export and add new option

thewilkybarkid:
The library uses a default export rather than a named one, so wasn't working locally. Also brought the options up to date.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/tandrewnichols/indefinite
- ~[ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.~

### DIFF
--- a/types/indefinite/index.d.ts
+++ b/types/indefinite/index.d.ts
@@ -4,8 +4,9 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 export interface Options {
+    articleOnly?: boolean,
     capitalize?: boolean;
     caseInsensitive?: boolean;
     numbers?: "colloquial";
 }
-export function indefinite(word: string | number, opts?: Options): string;
+export default function(word: string | number, opts?: Options): string;
--- a/types/indefinite/index.d.ts
+++ b/types/indefinite/index.d.ts
@@ -4,7 +4,7 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 export interface Options {
-    articleOnly?: boolean,
+    articleOnly?: boolean;
     capitalize?: boolean;
     caseInsensitive?: boolean;
     numbers?: "colloquial";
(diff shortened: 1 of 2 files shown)


types/indefinite/index.d.ts
```diff
@@ -1,11 +1,12 @@
-// Type definitions for indefinite 2.2
+// Type definitions for indefinite 2.3
 // Project: https://github.com/tandrewnichols/indefinite
-// Definitions by: omaishar <https://github.com/omaishr>
+// Definitions by: omaishar <https://github.com/omaishar>
```
thewilkybarkid: Hope this is right @omaishar!

types/indefinite/index.d.ts
```diff
@@ -1,11 +1,12 @@
...
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 export interface Options {
+    articleOnly?: boolean;
     capitalize?: boolean;
     caseInsensitive?: boolean;
     numbers?: "colloquial";
 }
-export function indefinite(word: string | number, opts?: Options): string;
+export default function(word: string | number, opts?: Options): string;
```
PranavSenthilnathan: You actually want to use `export =` instead since the underlying library doesn't use default exports. See [this](https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-function-d-ts.html) for more details.

PranavSenthilnathan (changes requested):

types/indefinite/index.d.ts
thewilkybarkid: Hopefully https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40238/commits/e20feba31ebc4126246767b95dd03db65b9850a2 is right!

PranavSenthilnathan (approved):

PR #40242, opened 2019-11-08
fix(sinon): sinon.spy property accessor types

Zirak:
This can properly be done better.

Please fill in this template.

Note that running `npm test` on sinon without my changes currently fails with some errors beyond my current comprehension. No errors arose from dependencies.
Same note as with `test`, there are currently linting issues with sinon in master.

In sinon, one can spy on getters and setters like so:

```js
const sinon = require('sinon');
const foo = { get bar() {}, set bar() {} };
const s = sinon.spy(foo, 'bar', ['get']);
```

`s` now contains the property descriptor for `foo.bar`, where `s.get` is a spy for the getter, so:

```js
s.get.called; // false
foo.bar;
s.get.called; // true
```

The typings for `sinon.spy` did not support this overload. Linting the above code before would tell you how `s.get` is undefined, as it thinks `s` is a spy itself.
This patch attempts to type this overload. The returned value is an object which looks like a property descriptor, with the `get` and `set` as spies.

This however isn't foolproof. In the above example, while `s.get` is a spy, `s.set` is the original function. My typescript-fu isn't strong enough to solve this, but it's a start?

### DIFF
--- a/types/sinon/ts3.1/index.d.ts
+++ b/types/sinon/ts3.1/index.d.ts
@@ -335,6 +335,8 @@ declare namespace Sinon {
         restore(): void;
     }
 
+    type PropertyAccessorType = 'get' | 'set';
+
     interface SinonSpyStatic {
         /**
          * Creates an anonymous function that records arguments, this value, exceptions and return values for all calls.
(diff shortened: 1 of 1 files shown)


types/sinon/ts3.1/index.d.ts
```diff
@@ -351,11 +353,23 @@ declare namespace Sinon {
...
             ? SinonSpy<TArgs, TReturnValue>
             : SinonSpy;
+
+        <T, K extends keyof T>(obj: T, method: K, types: PropertyAccessorType[]): {
+            [P in PropertyAccessorType]?: T[K] extends (
+                ...args: infer TArgs
+            ) => infer TReturnValue
+                    ? SinonSpy<TArgs, TReturnValue>
+                    : SinonSpy;
+        } & {
```
43081j: if possible the object should be an extension of a `PropertyDescriptor` in some way. we don't want to be maintaining our own copy of what a descriptor looks like

PR #40249, opened 2019-11-08
[passport] Added request in serializeUser/deserializeUser

boenrobot:
Please fill in this template.

- [x] Test the change in your own code. (Compile)
- [ ] Test the change in your own code. (run.)

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: jaredhanson/passport#743


itayganor:
@boenrobot
Assuming `TR extends IncomingMessage` equals to `express.Request` by default is not correct, because `serializeUser` and `deserializeUser` are also used for socket connections (for example, using Socket.io).

I understand that we can *assume* that in most basic applications `TR = express.Request` is correct, but as soon as you connect a socket solution, this becomes wrong.

In Socket.io, `socket.request` is of type `IncomingMessage`.

PR #40257, opened 2019-11-08
bunyan-format: Add merged function declaration to support calling default export as a function

Ashpabb:
Signed-off-by: Ashley Abbott <ashpabb@gmail.com>

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/thlorenz/bunyan-format/blob/v0.2.0/index.js>>


dex4er (approved):

types/bunyan-format/index.d.ts
```diff
@@ -1,8 +1,9 @@
 // Type definitions for bunyan-format 0.2
 // Project: https://github.com/thlorenz/bunyan-format
 // Definitions by: Piotr Roszatycki <https://github.com/dex4er>
+//                 Ashley Abbott <https://github.com/ashpabb>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.1
+// TypeScript Version: 3.6
```
PranavSenthilnathan: Is this needed? I don't think these changes require 3.6, so it would be nice to keep the TypeScript version low to keep supporting people using those versions.

PranavSenthilnathan (changes requested):

types/bunyan-format/index.d.ts
dex4er: It throws `'new' expression, whose target lacks a construct signature, implicitly has an 'any' type.` on older Typescript version.

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-6.html#ambient-classes-and-functions-can-merge

Ashpabb: The ambient merging feature is highly readable to me but also, I'm not sure how else I would define the type of a callable class like this without manually adding a bunch of prototype methods - is poly-filling the merge feature of 3.6 worth it?

I'd be inclined to say no but if this type def ought to continue supporting v2, I can make a new commit (suggestions would be welcomed).

dex4er: I don't see any problem here with newer syntax and it is still possible to use older version of @types/bunyan-format if anyone is locked with older version of Typescript.

dex4er (approved):

dex4er (approved):

types/bunyan-format/index.d.ts
Ashpabb: @PranavSenthilnathan can you weigh in here? I'd like to get this wrapped up

PranavSenthilnathan: I think this is reasonable.

PranavSenthilnathan (approved):

dex4er:
@Ashpabb this change broke `bunyan-format`. See #40444. I think we should try to switch to older syntax or drop this change completely as far as I don't see what is broken with this "ambient merge". Maybe this is some problem with Typescript?

PR #40278, opened 2019-11-09
[@types/stripe] Added "transfer_group" as property on stripe.ITransfer

axelfran:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://stripe.com/docs/api/transfers/object#transfer_object-transfer_group


types/stripe/index.d.ts
```diff
@@ -7667,7 +7667,12 @@ declare namespace Stripe {
             status: Statuses;
 
             /**
-             * Can be card, bank_account, or stripe_account.
+             * A string that identifies this transaction as part of a group.
+             */
+            transfer_group: string;
```
pushplay: Is this always set or can it be `null`?

axelfran: @pushplay Good catch! You are right and I've updated the PR.

pushplay (approved):

PR #40287, opened 2019-11-11
[bluebird] Fix typings for .all() and .some()

thislooksfun:
When calling `.all()` or `.some(<count>)` before with complex types, it would default to the `.all(): Promise<never>` definition, and likewise for `.some()`. This change fixes those issues, bringing them in line with the implementations of `.each()`, `.map()`, and the like.

Example of previous (incorrect) behavior:
```typescript
import Promise from "bluebird"
const p1: Promise<Array<Foo | null>> = Promise.resolve([]);
const p2 = p1.all();  // Here p2 has type Promise<never>
```

Example of fixed behavior:
```typescript
import Promise from "bluebird"
const p1: Promise<Array<Foo | null>> = Promise.resolve([]);
const p2 = p1.all();  // Here p2 has type Promise<(Foo | null)[]>
```

---

Please fill in this template.

### DIFF
--- a/types/bluebird/index.d.ts
+++ b/types/bluebird/index.d.ts
@@ -1,6 +1,6 @@
 // Type definitions for bluebird 3.5
 // Project: https://github.com/petkaantonov/bluebird
-// Definitions by: Leonard Hecker <https://github.com/lhecker>
+// Definitions by: Leonard Hecker <https://github.com/lhecker>, thislooksfun <https://github.com/thislooksfun>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.2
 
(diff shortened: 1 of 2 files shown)


lhecker (approved):

types/bluebird/index.d.ts
```diff
@@ -566,7 +566,7 @@ declare class Bluebird<R> implements PromiseLike<R>, Bluebird.Inspection<R> {
   /**
    * Same as calling `Promise.all(thisPromise)`. With the exception that if this promise is bound to a value, the returned promise is bound to that value too.
    */
-  all(this: Bluebird<Iterable<{}>>): Bluebird<R>;
+  all<Q>(this: Bluebird<R & Iterable<Q>>): Bluebird<R>;
```
PranavSenthilnathan: I think having no generic is probably a little better since the generic is only used in one place in the signature. However, as you mentioned in your description, this is consistent with the other functions in this class.

PranavSenthilnathan (approved):

types/bluebird/index.d.ts
thislooksfun: If you have some other way to make the new tests pass without needing generics that would be fine by me, but I don't know of one off the top of my head.

PR #40301, opened 2019-11-11
Update API from figma.d.ts v1 update 5

darknoon:
The source of truth for these types is on the [figma developer site](https://www.figma.com/plugin-docs/api/typings/) currently.

It's unclear what semver should apply here since there were breaking changes but they are calling it still v1.

Changes I made:
- needed to add `: void` return types for some functions based on the declared settings
- ran prettier with the settings for this project
- changed `any | undefined` to just `any` to pass lint check

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.figma.com/plugin-docs/api/typings/

I'm assuming this means the header of the file. It's a little confusing what "version" means for figma plugins, so hopefully someone at figma can weigh in.

### DIFF
--- a/types/figma/figma-tests.ts
+++ b/types/figma/figma-tests.ts
@@ -1,3 +1,15 @@
 const frame = figma.createFrame();
 frame.appendChild(figma.createRectangle());
+
+figma.on('selectionchange', () => {
+    figma.currentPage.selection;
+});
+
+figma.clientStorage.setAsync('happy', true);
+figma.clientStorage.setAsync('happy', { a: 123 });
+
+async () => {
+    const { a } = await figma.clientStorage.getAsync('happy');
+};
+
 figma.closePlugin();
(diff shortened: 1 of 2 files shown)


rudi-c (approved):
Thanks for taking the time to submit this PR! This all looks good to me.

Aside regarding semver: the thing with semver is that it implies you can pin the existing API and behavior of a dependency. But the Figma API is a host, not a package. Hence why the API itself doesn't use semver. See https://www.figma.com/plugin-docs/stability-updates/

The version of _typings_ is a separate thing entirely from the version of the API. There, the plugin can actually pin a specific typings version (even if it becomes a bit out of date). The critical thing with versioning the API is to keep existing plugins working to the extent that is possible. `v1.5` is probably fine for typings, since it reflects the file that shipped with "Version 1, Update 5".

PranavSenthilnathan:

PR #40311, opened 2019-11-11
redux-async-queue > Added Action type

deanvanniekerk:
Adds a type for the redux async queue action.

Example usage

```
import { AsyncQueueAction } from "redux-async-queue";

export const doWorkAction = (): AsyncQueueAction => {
    return {
        queue: "MY_QUEUE",
        callback: (next, dispatch) => {
            //do work...
        },
    };
};
```

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/zackargyle/redux-async-queue/blob/master/src/index.js>>

### DIFF
--- a/types/redux-async-queue/index.d.ts
+++ b/types/redux-async-queue/index.d.ts
@@ -1,10 +1,16 @@
 // Type definitions for redux-async-queue 1.0
 // Project: https://github.com/zackargyle/redux-async-queue
 // Definitions by: Andrei Horodinca <https://github.com/andreiho>
+//                 Dean van Niekerk <https://github.com/deanvanniekerk>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
-import { AnyAction } from 'redux';
+import { Action, AnyAction, Dispatch } from "redux";
+
+export interface AsyncQueueAction<T extends Action = AnyAction> {
+    queue: string;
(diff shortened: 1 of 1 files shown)


andreiho (approved):

types/redux-async-queue/index.d.ts
```diff
@@ -1,10 +1,16 @@
...
 // Project: https://github.com/zackargyle/redux-async-queue
 // Definitions by: Andrei Horodinca <https://github.com/andreiho>
+//                 Dean van Niekerk <https://github.com/deanvanniekerk>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
-import { AnyAction } from 'redux';
+import { Action, AnyAction, Dispatch } from "redux";
+
+export interface AsyncQueueAction<T extends Action = AnyAction> {
```
PranavSenthilnathan: I'm kind of confused about where this type is used. `queueMiddleware` seems to be the only other thing exported from this library and `AsyncQueueAction` doesn't appear in its signature.

DeanVanNiekerk: Its the type of Action that the `redux-async-queue` middleware is looking for. 
See example usage in the description.

Additionally have a look at the handling of the Action in the middleware [here](https://github.com/zackargyle/redux-async-queue/blob/e6d9a46941603b25c7c88d276e29579306714e2a/src/index.js#L34)

PranavSenthilnathan:

PR #40313, opened 2019-11-11
Add types for compute-argmax

EricCrosson:
Please fill in this template.

If adding a new definition:

I hope I got this one right, it just exports a single function

Thanks for reviewing this PR!

### DIFF
--- a/types/compute-argmax/compute-argmax-tests.ts
+++ b/types/compute-argmax/compute-argmax-tests.ts
@@ -1,0 +1,4 @@
+import compute = require('compute-argmax');
+
+const data = [3, 2, 5, 2, 5];
+const idx: number[] = compute(data);
--- a/types/compute-argmax/index.d.ts
+++ b/types/compute-argmax/index.d.ts
@@ -1,0 +1,8 @@
+// Type definitions for compute-argmax 1.0
+// Project: https://github.com/compute-io/argmax
+// Definitions by: Eric Crosson <https://github.com/EricCrosson>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare function argmax(values: number[]): number[];
+
+export = argmax;
--- a/types/compute-argmax/index.d.ts
+++ b/types/compute-argmax/index.d.ts
@@ -3,6 +3,6 @@
 // Definitions by: Eric Crosson <https://github.com/EricCrosson>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-declare function argmax(values: number[]): number[];
+declare function argmax(values: ArrayLike<number>): number[];
 
 export = argmax;
--- a/types/compute-argmax/tsconfig.json
+++ b/types/compute-argmax/tsconfig.json
@@ -1,0 +1,23 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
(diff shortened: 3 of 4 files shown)


types/compute-argmax/compute-argmax-tests.ts
```diff
@@ -0,0 +1,4 @@
+import compute = require('compute-argmax');
+
+const data = [3, 2, 5, 2, 5];
+const idx: number[] = compute.argmax(data);
```
PranavSenthilnathan: According to the usage and the underlying library [export](https://github.com/compute-io/argmax/blob/master/lib/index.js#L64) it looks like the function is exported directly. Changing the export in your `index.d.ts` to `export = ...` should fix it.

PranavSenthilnathan (changes requested):

types/compute-argmax/compute-argmax-tests.ts
EricCrosson: Hi Pranav, thanks for the feedback

I am a little confused on the syntax, as

```
export = function argmax(values: number[]): number[];
```

tells me I am missing some curly braces {}

EricCrosson: oo I think I got it

types/compute-argmax/index.d.ts
```diff
@@ -0,0 +1,8 @@
+// Type definitions for compute-argmax 1.0
+// Project: https://github.com/compute-io/argmax
+// Definitions by: Eric Crosson <https://github.com/EricCrosson>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare function argmax(values: number[]): number[];
```
demensky: Let's simplify `values` type to `ArrayLike<number>`?

```suggestion
declare function argmax(values: ArrayLike<number>): number[];
```

EricCrosson: Seems like a good idea, thank you.

I'm a little hazy on what may be considered arraylike but is not an array, is there any concrete example you can provide to help my mental model?

demensky: You're right! Sorry for my carelessness. The [source code checks] to see if it is an array. If you remove the check, you can use it, for example, with `arguments`.

```js
function argmax(arr) {
  var len = arr.length, max = arr[0], idx = [0], val;

  for (var i = 1; i < len; i++) {
    val = arr[i];
    if (val > max) {
      max = val;
      idx.length = 0;
      idx.push(i);
    } else if (val === max) {
      idx.push(i);
    }
  }
  return idx;
}

function test() {
  return argmax(arguments)
}

test(3, 2, 5, 2, 5); // [2, 4]
```

Therefore, I suggest changing this to:

```ts
declare function argmax(values: ReadonlyArray<number>): number[];
```

[source code checks]: https://github.com/compute-io/argmax/blob/master/lib/index.js#L39-L41

EricCrosson: Oh neato! I always thought `arguments` *was* an array. The [docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments) say otherwise, thanks for teaching me something today

PranavSenthilnathan (approved):

PR #40314, opened 2019-11-11
Updated express-session types

fiznool:
Follow up to #39421, this PR adds support for the following:

- Allow `secure: 'auto'` cookie setting
- Restrict `sameSite` cookie setting to boolean or valid string value
  according to (draft) spec

Both options are outlined in the express-session documentation:

- [SameSite option](https://github.com/expressjs/session#cookiesamesite)
- [Secure option](https://github.com/expressjs/session#cookiesecure)

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39421

### DIFF
--- a/types/express-session/index.d.ts
+++ b/types/express-session/index.d.ts
@@ -4,6 +4,7 @@
 //                 Jacob Bogers <https://github.com/jacobbogers>
 //                 Naoto Yokoyama <https://github.com/builtinnya>
 //                 Ryan Cannon <https://github.com/ry7n>
+//                 Tom Spencer <https://github.com/fiznool>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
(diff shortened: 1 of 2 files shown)


types/express-session/index.d.ts
```diff
@@ -58,7 +59,17 @@ declare namespace session {
     secret: string | string[];
     name?: string;
     store?: Store | MemoryStore;
-    cookie?: express.CookieOptions;
+    cookie?: {
```
BorntraegerMarc: @fiznool @sheetalkamat can we move this type to an own interface? Because I had a method before which was returning a shared config (for our microservices) the type `express.CookieOptions`.

Now after this PR I have to create manually an interface to keep the code. These types should provide all necessary types (including the cookie structure).

sheetalkamat: Please submit the PR with the changes for these changes.

fiznool: @BorntraegerMarc you should be able to access this using TypeScript's [indexed access types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types):

``` ts
import { SessionOptions } from 'express-session';
type CookieOptions = SessionOptions['cookie'];
```

BorntraegerMarc: I didn't know one can do this. Thanks a lot @fiznool !

JasonKleban:
@sheetalkamat Are these sameSite values broadly treated as case insensitive? I usually find them documented (informally) with capital initial letters, `"Lax" | "Strict" | "None"`. Do you have a more definitive reference? Thanks!

PR #40320, opened 2019-11-12
typing corrections for Titanium and Alloy

ffMathy:
made some more adjustments to typings of Titanium and Alloy.

### DIFF
--- a/types/titanium/index.d.ts
+++ b/types/titanium/index.d.ts
@@ -9584,7 +9584,7 @@ declare namespace Titanium {
 			/**
 			 * Replaces an existing section.
 			 */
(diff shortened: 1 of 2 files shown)


types/titanium/index.d.ts
```diff
@@ -9584,7 +9584,7 @@ declare namespace Titanium {
 			/**
 			 * Replaces an existing section.
 			 */
-			replaceSectionAt(sectionIndex: number, section: Titanium.UI.ListSection, animation: ListViewAnimationProperties): void;
+			replaceSectionAt(sectionIndex: number, section: Titanium.UI.ListSection, animation?: ListViewAnimationProperties): void;
```
janvennemann: @ffMathy Please make sure that is reflected in our docs as well. Otherwise this change will be overwritten the next we generate the typings from docs.

PR #40325, opened 2019-11-12
Allowing group of columns

simusr2:
A column can be a "group" of columns (Example: group header column -> Measurements, grouped column -> Length, Width, Height)


simusr2:
http://tabulator.info/docs/4.4/columns#groups

Jojoshua (approved):
Thanks for the test

PR #40327, opened 2019-11-12
[react-dom/experimental]: add null type to createRoot

juicyarts:
Please fill in this template.

An element selector can potentially return null. But this case is handled by validity checks like isValidContainer. So one should be able to pass null which will then be rejected by the checks further down. Same goes for the hydration context.

Before:
Screenshot 2019-11-12 at 14 15 16

After: 
Screenshot 2019-11-12 at 14 14 52

// Validity checks (since these are in expermeintal builds i don't know where to reference it properly so i added screens:
Screenshot 2019-11-12 at 14 25 25

Screenshot 2019-11-12 at 14 25 42

### DIFF
--- a/types/react-dom/experimental.d.ts
+++ b/types/react-dom/experimental.d.ts
@@ -74,7 +74,7 @@ declare module '.' {
      *
      * @see https://reactjs.org/docs/concurrent-mode-reference.html#createroot
      */
-    function createRoot(container: Element | Document | DocumentFragment | Comment, options?: RootOptions): Root;
+    function createRoot(container: Element | Document | DocumentFragment | Comment | null, options?: RootOptions): Root;
 
     function unstable_discreteUpdates<R>(callback: () => R): R;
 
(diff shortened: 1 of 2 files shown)


Jessidhia (changes requested):
The current restriction on the types is correct; `null` is not a valid input, it will throw a runtime error saying your input is invalid. Just because there is validation to guard against passing `null` instead of just crashing with a `TypeError` doesn't mean it is valid input as, well, it is rejected by validation.

 If you want to ignore the fact that `getElementById` can return `null`, then you should just assert it with the postfix `!` operator.

```ts
React.createRoot(
  document.getElementById('root')!
)
```

juicyarts:
Ignoring this case is not the issue. I raised the point mainly because of the inconsistency in handling the arguments in this case compared to the current implementation of the render method.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b638a27a775eea0eda437ea38ab77c7928927b79/types/react-dom/index.d.ts#L53-L61
I'm fine with it if there is a reason for not being consistent here ;)

PR #40330, opened 2019-11-12
Grant remove grant from gas

grant:
Removes old owner.
There is no new owner. Someone can create a PR to add themselves if they wish.


grant:
I'm sorry @mtgto, I am working on other things rather than Apps Script, which is still older JavaScript ES3 from 2009. You can be the owner if you'd like.

PR #40335, opened 2019-11-12
[mongoose] fixes lean return type

simllll:
commit https://github.com/DefinitelyTyped/DefinitelyTyped/commit/c74f860f8b80bb5542fbdafa0a81e260f2fec147#diff-ec6e6a45ee09ed93f5d81cb304c5a0b3 introduced typings for lean (instead of any). 
I like this approach, but it breaks old code, because type script assume T is undefined in this case. Also it didn't respect find or findOne and always used T directly.

This approach fixes therefore two issues:
- by default (if no type is provided), it is any again (backswards compatiblity)
- Depending on the used functino it either returns the given type or an array of the given type.

Please fill in this template.

- [] Add or edit tests to reflect the change. (Run with `npm test`.)

- [X] Improves an already existing defintion

### DIFF
--- a/types/mongoose/index.d.ts
+++ b/types/mongoose/index.d.ts
@@ -2000,7 +2000,7 @@ declare module "mongoose" {
      * getters/setters or other Mongoose magic applied.
      * @param {Boolean|Object} bool defaults to true
      */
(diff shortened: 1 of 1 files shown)


simllll:
@Iosif380 @sandersn please verify :) thanks

Alorel (approved):

PR #40340, opened 2019-11-12
[jasmine] Support Jasmine 3.5

simondel:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/jasmine/jasmine/blob/master/release_notes/3.5.0.md

### DIFF
--- a/types/jasmine/index.d.ts
+++ b/types/jasmine/index.d.ts
@@ -273,6 +273,7 @@ declare namespace jasmine {
     function addCustomEqualityTester(equalityTester: CustomEqualityTester): void;
 
     function addMatchers(matchers: CustomMatcherFactories): void;
+    function addAsyncMatchers(matchers: CustomMatcherFactories): void;
 
     function stringMatching(str: string | RegExp): AsymmetricMatcher<string>;
 
--- a/types/jasmine/index.d.ts
+++ b/types/jasmine/index.d.ts
@@ -258,6 +258,7 @@ declare namespace jasmine {
     function arrayContaining<T>(sample: ArrayLike<T>): ArrayContaining<T>;
     function arrayWithExactContents<T>(sample: ArrayLike<T>): ArrayContaining<T>;
     function objectContaining<T>(sample: Partial<T>): ObjectContaining<T>;
+    function setDefaultSpyStrategy(and: SpyAnd): void;
     function createSpy(name?: string, originalFn?: Function): Spy;
 
     function createSpyObj(baseName: string, methodNames: SpyObjMethodNames): any;
--- a/types/jasmine/index.d.ts
+++ b/types/jasmine/index.d.ts
@@ -204,6 +204,11 @@ declare namespace jasmine {
             (ReadonlyArray<string> | { [methodName: string]: any }) :
             (ReadonlyArray<keyof T> | { [P in keyof T]?: T[P] extends Func ? ReturnType<T[P]> : any });
 
+    type SpyObjPropertyNames<T = undefined> =
+        T extends undefined ?
+                (ReadonlyArray<string> | { [propertyName: string]: any }) :
+                (ReadonlyArray<keyof T>);
+
     /**
      * Configuration that can be used when configuring Jasmine via {@link jasmine.Env.configure}
      */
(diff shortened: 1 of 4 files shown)


types/jasmine/index.d.ts
```diff
@@ -204,13 +204,19 @@ declare namespace jasmine {
...
+                (ReadonlyArray<keyof T>);
+
     /**
      * Configuration that can be used when configuring Jasmine via {@link jasmine.Env.configure}
      */
     interface EnvConfiguration {
         random?: boolean;
         seed?: number;
         failFast?: boolean;
+        failSpecWithNoExpectations?: boolean;
```
djungowski: According to the [Configuration docs](https://jasmine.github.io/api/3.5/Configuration.html) there is also a `Promise` option since 3.5.0, please add this one as well

types/jasmine/jasmine-tests.ts
```diff
@@ -1479,6 +1500,39 @@ describe("createSpyObj", function() {
...
     });
+
+    it("creates an object with spy properties if a second list is passed", function() {
+        const spyObj = jasmine.createSpyObj('base', ['method1'], ['prop1']);
+
+        expect(spyObj).toEqual({
+          method1: jasmine.any(Function)
+        });
+
+        expect(spyObj.prop1).toBeUndefined();
```
djungowski: Not sure this test adds any value, as it would also pass if you simply left out the second list

types/jasmine/index.d.ts
```diff
@@ -272,6 +278,7 @@ declare namespace jasmine {
     function addCustomEqualityTester(equalityTester: CustomEqualityTester): void;
 
     function addMatchers(matchers: CustomMatcherFactories): void;
+    function addAsyncMatchers(matchers: CustomMatcherFactories): void;
```
djungowski: There is also the async matcher `toBeRejectedWithError` that has been added in 3.5.0 (see [async-matchers](https://jasmine.github.io/api/3.5/async-matchers.html). Please add it to the AsyncMatchers

djungowski (changes requested):
Basically this looks good to me, there are however some more things that have been introduced in 3.5.0 - please add them

types/jasmine/index.d.ts
simondel: Hi @djungowski do you have any tips for the matchers? I'm having some issues with the tests. They don't seem to accept the Error parameter.

The commit with the change is here: https://github.com/simondel/DefinitelyTyped/commit/ed90db146791e6bc72db73ef14975c82dd194739

After It's fixed, I'll create a new PR.

djungowski: @simondel From what I see, a test like
```
await expectAsync(Promise.reject(badness)).toBeRejectedWithError(Error, "badness");
```
would have to look like
```
await expectAsync(Promise.reject(new Error('badness'))).toBeRejectedWithError(new Error('badness'));
```

djungowski: @simondel If `toBeRejectedWithError` works like `toThrowError`, it might also work with

```
await expectAsync(Promise.reject(new Error('badness')).toBeRejectedWithError('badness');
```

simondel: Yeah, I was thinking about that as well.. But that's not what the tests inside of Jasmine use, so I'm not sure if that's the test that I should have in the types.

If you look at lines 5 through 8, or 19 through 22 [of the PR](https://github.com/jasmine/jasmine/pull/1686/files#diff-ab83340b986e6941a2b629c15a1ae79f) you can see code like this:
```js
var matcher = jasmineUnderTest.asyncMatchers.toBeRejectedWithError(jasmineUnderTest.matchersUtil),
actual = Promise.reject(new TypeError('foo'));

matcher.compare(actual, TypeError) // TypeError is used instead of a new instance
```
and
```js
matcher = jasmineUnderTest.asyncMatchers.toBeRejectedWithError(jasmineUnderTest.matchersUtil),
actual = Promise.reject(new TypeError('foo'));

matcher.compare(actual, TypeError, 'foo') // TypeError is used instead of a new instance
```

djungowski: So I've looked at the typing definition of `toThrow` and the tests work if you change your new typings for `toBeRejectedWithError` as follows:

index.d.ts
```javascript
toBeRejectedWithError(message?: string | RegExp): PromiseLike<void>;
toBeRejectedWithError(expected?: new (...args: any[]) => Error, message?: string | RegExp): PromiseLike<void>;
```

ts3.1/index.d.ts
```javascript
toBeRejectedWithError(message?: string | RegExp): Promise<void>;
toBeRejectedWithError(expected?: new (...args: any[]) => Error, message?: string | RegExp): Promise<void>;
```

(The only difference between index.d.ts and ts3.1/index.d.ts is PromiseLike vs. Promise)

simondel: Thanks for the help! ♥ I've created a new PR: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40568

PR #40344, opened 2019-11-13
reference @types/node in vfile-location

sandersn:
vfile references @types/node but types-publisher doesn't know that. Instead, reference @types/node in vfile-location, since it now depends on vfile instead of @types/vfile.


sandersn:
Thanks @typescript-bot, but you couldn't know that node was *previously* a (transitive) dependency of vfile-location, so these numbers are just back to their previous values.

PR #40353, opened 2019-11-13
[use-deep-compare-effect] Add package

JPeer264:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/use-deep-compare-effect/index.d.ts
+++ b/types/use-deep-compare-effect/index.d.ts
@@ -1,0 +1,10 @@
+// Type definitions for use-deep-compare-effect 1.2
+// Project: https://github.com/kentcdodds/use-deep-compare-effect#readme
+// Definitions by: JPeer264 <https://github.com/JPeer264>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import { useEffect } from 'react';
+
+export const useDeepCompareNoEffect: typeof useEffect;
+
+export default useEffect;
--- a/types/use-deep-compare-effect/index.d.ts
+++ b/types/use-deep-compare-effect/index.d.ts
@@ -5,6 +5,8 @@
 
 import { useEffect } from 'react';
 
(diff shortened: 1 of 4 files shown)


types/use-deep-compare-effect/index.d.ts
```diff
@@ -0,0 +1,10 @@
+// Type definitions for use-deep-compare-effect 1.2
+// Project: https://github.com/kentcdodds/use-deep-compare-effect#readme
+// Definitions by: JPeer264 <https://github.com/JPeer264>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import { useEffect } from 'react';
+
+export const useDeepCompareNoEffect: typeof useEffect;
```
PranavSenthilnathan: I don't see this in the [package](https://unpkg.com/use-deep-compare-effect@1.2.0/dist/use-deep-compare-effect.cjs.js).

types/use-deep-compare-effect/index.d.ts
```diff
@@ -0,0 +1,10 @@
...
+
+export default useEffect;
```
PranavSenthilnathan: This will work, but for clarity you might want to declare a const of type `typeof useEffect` and export that. Currently it looks like you're exporting a value that's the same as `useEffect` rather than having the same type as `useEffect`.

PranavSenthilnathan (changes requested):

types/use-deep-compare-effect/index.d.ts
JPeer264: True, I misread the function and updated it to `useDeepCompareEffectNoCheck`

types/use-deep-compare-effect/index.d.ts
JPeer264: Alright, makes sense indeed. Updated

PranavSenthilnathan (approved):

PranavSenthilnathan:

PR #40355, opened 2019-11-13
[aws-lambda] Improved Lambda@Edge types

apepper:
Fine tuned types for Lambda@Edge events.

Sources:
* https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-event-structure.html
* ttps://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/lambda-include-body-access.html

### DIFF
--- a/types/apex.js/index.d.ts
+++ b/types/apex.js/index.d.ts
@@ -2,7 +2,7 @@
 // Project: https://github.com/apex/node-apex
 // Definitions by: Yoriki Yamaguchi <https://github.com/y13i>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+// TypeScript Version: 3.5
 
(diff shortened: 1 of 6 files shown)


simonbuchan (approved):
Looks great, thanks for this!

I'm going to approve this, but will note that `Omit<T, K>` is fairly easily implemented as `Pick<T, Exclude<keyof T, K>>`, or simply extracting the non-omitted fields out to a base interface; typescript 3.5 hasn't been out that long, and forcing updates on users for a single use is a bit much.

apepper:
@simonbuchan Thank you for your feedback. I replaced Omit with Exclude and downgraded the needed TypeScript version to 2.8.

simonbuchan:
Thanks! It's already approved, so it should be merged by the DT team some time soon (generally within a day)

PR #40361, opened 2019-11-13
[relay-runtime] Correct typing for default missing field handlers

ethanve:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/relay/blob/master/packages/relay-runtime/handlers/RelayDefaultMissingFieldHandlers.js#L31

### DIFF
--- a/types/relay-runtime/index.d.ts
+++ b/types/relay-runtime/index.d.ts
@@ -170,9 +170,7 @@ export { readInlineData } from './lib/store/readInlineData';
 
 // Extensions
 export { RelayDefaultHandlerProvider as DefaultHandlerProvider } from './lib/handlers/RelayDefaultHandlerProvider';
-export {
-    missingViewerFieldHandler as DefaultMissingFieldHandlers,
-} from './lib/handlers/RelayDefaultMissingFieldHandlers';
+export { DefaultMissingFieldHandlers } from './lib/handlers/RelayDefaultMissingFieldHandlers';
 import * as ConnectionHandler from './lib/handlers/connection/RelayConnectionHandler';
 export { ConnectionHandler };
 
--- a/types/relay-runtime/index.d.ts
+++ b/types/relay-runtime/index.d.ts
@@ -170,7 +170,9 @@ export { readInlineData } from './lib/store/readInlineData';
 
 // Extensions
 export { RelayDefaultHandlerProvider as DefaultHandlerProvider } from './lib/handlers/RelayDefaultHandlerProvider';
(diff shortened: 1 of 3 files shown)


types/relay-runtime/lib/handlers/RelayDefaultMissingFieldHandlers.d.ts
```diff
@@ -1,3 +1,3 @@
 import { MissingFieldHandler } from '../store/RelayStoreTypes';
 
-export const missingViewerFieldHandler: MissingFieldHandler;
```
ethanve: `missingViewerFieldHandler` isn't being exported by this filer rather, tt's exporting an array with `missingViewerFieldHandler` as the only element. Not sure if this new typing is accurately representing that

alloy: Good catch!

1. However, the upstream version doesn’t actually export a named type, rather a default…

types/relay-runtime/lib/handlers/RelayDefaultMissingFieldHandlers.d.ts
```diff
@@ -1,3 +1,3 @@
...
+export const DefaultMissingFieldHandlers: MissingFieldHandler[];
```
alloy: 2. …so this should really be something like:

```suggestion
declare const DefaultMissingFieldHandlers: MissingFieldHandler[];
export default DefaultMissingFieldHandlers
```

types/relay-runtime/index.d.ts
```diff
@@ -170,9 +170,7 @@ export { readInlineData } from './lib/store/readInlineData';
 
 // Extensions
 export { RelayDefaultHandlerProvider as DefaultHandlerProvider } from './lib/handlers/RelayDefaultHandlerProvider';
-export {
-    missingViewerFieldHandler as DefaultMissingFieldHandlers,
-} from './lib/handlers/RelayDefaultMissingFieldHandlers';
+export { DefaultMissingFieldHandlers } from './lib/handlers/RelayDefaultMissingFieldHandlers';
```
alloy: 3. After which [the below suggested change] this should be:

```suggestion
import DefaultMissingFieldHandlers from './lib/handlers/RelayDefaultMissingFieldHandlers';
export { DefaultMissingFieldHandlers };
```

alloy (changes requested):
Conceptually good, just some minor suggestions to make it follow upstream exactly 👍

alloy:
Great, thanks!

PR #40371, opened 2019-11-14
[react-linkify] Export namespace to expose props type defs

akhilstanis:
It would be nice to expose the props type definition which would allow users to extend the original component.

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://gist.github.com/akhilstanis/5eba81c6721b349cc1da919d2f0548f6

### DIFF
--- a/types/react-linkify/index.d.ts
+++ b/types/react-linkify/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for react-linkify 0.2
+// Type definitions for react-linkify 1.0.0-alpha
 // Project: http://tasti.github.io/react-linkify/
 // Definitions by: Michael James <https://github.com/majames>
 //                 Jacky Wang <https://github.com/jackywang529>
--- a/types/react-linkify/index.d.ts
+++ b/types/react-linkify/index.d.ts
@@ -7,7 +7,7 @@
 
 import * as React from "react";
 
-declare namespace ReactLinkify {
+export declare namespace ReactLinkify {
     /**
      * Matching information
      */
--- a/types/react-linkify/index.d.ts
+++ b/types/react-linkify/index.d.ts
@@ -2,64 +2,61 @@
 // Project: http://tasti.github.io/react-linkify/
 // Definitions by: Michael James <https://github.com/majames>
 //                 Jacky Wang <https://github.com/jackywang529>
+//                 Akhil Stanislavose <https://github.com/akhilstanis>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.4
 
 import * as React from "react";
 
-export namespace ReactLinkify {
-    /**
-     * Matching information
-     */
-    interface MatchInfo {
-        /**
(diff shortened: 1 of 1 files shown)


akhilstanis:
cc @jackywang529 @majames

Upgraded the version and exported the namespace to use prop types. Seeking review...

types/react-linkify/index.d.ts
```diff
@@ -7,7 +7,7 @@
 
 import * as React from "react";
 
-declare namespace ReactLinkify {
+export namespace ReactLinkify {
```
jackywang529: thanks for bringing this up, this currently works for `import Linkify from "react-linkify";` and 
```
<Linkify componentDecorator={decoratorOverride}>
              // some text
</Linkify>
```
wondering why you need such namespace being exported?

types/react-linkify/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for react-linkify 0.2
+// Type definitions for react-linkify 1.0
```
jackywang529: LGTM

types/react-linkify/index.d.ts
jackywang529: would you mind sharing how you plan on extending such, can it be achieved with current decorator overrides?

types/react-linkify/index.d.ts
majames: Looks like the stable version of react linkify is still in alpha... should we update the version here to reflect that? https://www.npmjs.com/package/react-linkify

majames (approved):
LGTM after question addressed

types/react-linkify/index.d.ts
sheetalkamat: This is incorrect change.. The export of this namespace also exports value space (because of class `ReactLinkify` class declaration in it) but `react-linkify` doesnt export any value names `ReactLinkify`
Ideally you would want to move all the definitions in the namespace outside the namespace (removing the namespace declaration) and export only interface definitions

sheetalkamat (changes requested):

types/react-linkify/index.d.ts
akhilstanis: There is a linter rule which doesn't let you set a patch version. https://github.com/Microsoft/dtslint/blob/master/docs/dt-header.md

types/react-linkify/index.d.ts
akhilstanis: I basically need access to the props so that I can build components that extend the original functionality. Something like...
```tsx
import ReactLinkify, { ReactLinkify as RL } from 'react-linkify';

export const Linkify: React.FC<RL.Props> = (props) => (
  <ReactLinkify componentDecorator={componentDecorator} {...props}>
    {props.children}
  </ReactLinkify>
);

const componentDecorator = (href: string, text: string) => (
  <a href={href} target="_blank">
    {text}
  </a>
);
```

akhilstanis: That sounds right. 👍

PR #40373, opened 2019-11-14
Added typing for string-placeholder

fbritoferreira:
Please fill in this template.

-  [x] Test the change in your own code. (Compile and run.)

If adding a new definition:

### DIFF
--- a/types/string-placeholder/index.d.ts
+++ b/types/string-placeholder/index.d.ts
@@ -7,11 +7,11 @@
 export = template;
 export as namespace template;
 
-declare function template(str: string, data: [] | {}, options?: Options): string;
+declare function template(str: string, data: Readonly<[]> | Readonly<{}>, options?: Options): string;
 
 interface Options {
     before?: string;
     after?: string;
     escape?: string;
-    clean?: boolean | [];
+    clean?: boolean | Readonly<[]>;
 }
--- a/types/string-placeholder/index.d.ts
+++ b/types/string-placeholder/index.d.ts
@@ -4,7 +4,7 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.1
 
-export function template(str: string, data: Readonly<[]> | Readonly<{}>, options?: Options): string;
+export function template(str: string, data: Readonly<unknown>, options?: Readonly<Options>): string;
 
 interface Options {
     before?: string;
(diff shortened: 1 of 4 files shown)


types/string-placeholder/index.d.ts
```diff
@@ -0,0 +1,17 @@
+// Type definitions for string-placeholder 1.0
+// Project: https://github.com/crysalead-js/string-placeholder#readme
+// Definitions by: Filipe Ferreira <https://github.com/iTsFILIPOficial>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.1
+
+export = template;
+export as namespace template;
```
PranavSenthilnathan: I don't this module exports a global so this line isn't needed.

PranavSenthilnathan (changes requested):

types/string-placeholder/string-placeholder-tests.ts
```diff
@@ -0,0 +1,19 @@
+import template from 'string-placeholder';
+
+// $ExpectType string
+template('{0} test', ['here']);
+
```
demensky: Please add a case with an object.

```suggestion

template('${foo} test ${bar}', {foo: 'Bob', bar: '65'});

```

demensky (changes requested):

types/string-placeholder/index.d.ts
```diff
@@ -0,0 +1,16 @@
...
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.1
+
+export function template(str: string, data: Readonly<[]> | Readonly<{}>, options?: Options): string;
+
+interface Options {
+    before?: string;
+    after?: string;
+    escape?: string;
+    clean?: boolean | Readonly<[]>;
```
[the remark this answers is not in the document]
iTsFILIPOficial: @demesky https://github.com/crysalead-js/string-placeholder/blob/master/index.js#L25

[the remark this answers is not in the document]
demensky: @iTsFILIPOficial. I agree with the comments mentioned array. But in the source code, it is used only as a boolean value. I can’t even find a spec where an array is passed to `clean`.

demensky (approved):

types/string-placeholder/index.d.ts
```diff
@@ -0,0 +1,16 @@
...
+export function template(str: string, data: Readonly<unknown>, options?: Readonly<Options>): string;
+
+interface Options {
+    before?: string;
+    after?: string;
+    escape?: string;
+    clean?: boolean;
+}
+
+export default template;
```
PranavSenthilnathan:
```suggestion
export = template;
```

PranavSenthilnathan (changes requested):

PranavSenthilnathan (approved):

PranavSenthilnathan:

PR #40388, opened 2019-11-14
Stripe: Update ITier

durchanek:
Please fill in this template.

https://stripe.com/docs/api/plans/object?lang=node#plan_object-tiers


types/stripe/index.d.ts
```diff
@@ -6965,18 +6965,29 @@ declare namespace Stripe {
     namespace plans {
         interface ITier {
             /**
-             * A flat fee for the tier to be added into the total amount
+             * Up to and including to this quantity will be contained in the tier.
              */
-            flat_amount: number | null;
+            up_to: number | null;
```
pushplay: This one is optional but the others required?  I don't see anything indicating that in the docs.

durchanek: Should be the other way around, `up_to` is required but others are optional: https://stripe.com/docs/api/plans/create#create_plan-tiers I copied this from the previous version, but there definitively should be the option `inf` for a fallback tier. Not sure if it can ever be null - will test tomorrow and update the PR.

pushplay: 🤦‍♂ Yeah I got optional and required backwards in my question.

durchanek: I added `inf` for the fallback tier and it also can be null when retrieving a plan. I will separate these into ITierCreateOptions when possible.

pushplay:
I think this is still waiting on `ITierCreateOptions`.

sheetalkamat:
> I think this is still waiting on ITierCreateOptions.

That can follow in new PR

PR #40409, opened 2019-11-15
connect: add req.originalUrl; improve API; API docs from README

geekley:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.npmjs.com/package/connect#appuseroute-fn


types/connect/index.d.ts
```diff
@@ -1,94 +1,163 @@
...
+declare function createServer(): connect.Server;
 
 declare namespace createServer {
+    export interface Request extends http.IncomingMessage {
+        /**
+         * The request URL, altered to remove the route part. For example, if the handle function is used
+         * at the route `/foo`, the request for `/foo/bar` will invoke it with `req.url === '/bar'` and
+         * `req.originalUrl === '/foo/bar'`.
+         */
+        url: string;
```
EvanHahn: This is [already a property of `http.IncomingMessage`](https://nodejs.org/api/http.html#http_message_url). Have you added this for documentation purposes?

EvanHahn (changes requested):
IMO, too many things are happening in this pull request. It seems like there are a few distinct pieces:

1. Add documentation
1. Add `connect.Request`, which extends `http.IncomingMessage` with an `originalUrl`
1. Removes a lot of TSLint stuff

Could we split these into multiple changes to make it easier to review?

types/connect/index.d.ts
geekley: Yes, the documentation override is important because this url is conceptually different from the superclass (since it doesn't contain the route part). Also the not-null guarantee. It helps if you have strict null checking, you can do `req.url` without an assertion.

types/connect/index.d.ts
```diff
@@ -1,94 +1,163 @@
...
+    }
+
     export type ServerHandle = HandleFunction | http.Server;
 
-    type NextFunction = (err?: any) => void;
+    type NextFunction = (
+        /** An optional error, to make the app skip to error-handling middleware. */ err?: any
+    ) => void;
 
-    export type SimpleHandleFunction = (req: http.IncomingMessage, res: http.ServerResponse) => void;
```
geekley: Optional param in callback was removed according to https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#optional-parameters-in-callbacks

types/connect/index.d.ts
```diff
@@ -1,94 +1,163 @@
...
+         * Start the app listening for requests. This method will internally create a Node.js HTTP server and
+         * call `.listen()` on it.
+         *
+         * This is an alias to the `server.listen()` method in the version of Node.js running, so consult
+         * the Node.js documentation for all the different variations. The most common signature is `app.listen(port)`.
+         */
+        listen(port: number, hostname?: string, backlog?: number, listeningListener?: () => void): http.Server;
+        listen(port: number, hostname?: string, listeningListener?: () => void): http.Server;
+        listen(path: string, listeningListener?: () => void): http.Server;
+        listen(handle: any, listeningListener?: () => void): http.Server;
```
geekley: using `listeningListener?: () => void` instead of Function, to match signature of Server.listen

types/connect/index.d.ts
```diff
@@ -1,94 +1,163 @@
...
+            /** The error that occurred. */ err: any,
+            /** The Node.js http request object. */ req: Request,
+            /** The Node.js http response object. */ res: http.ServerResponse,
+            /** A function to call to invoke the next middleware. */ next: NextFunction
+        ) => void): Server;
+
+        /** Register a middleware. */
         use(fn: HandleFunction): Server;
+        /** Mount a middleware on the specified route. */
         use(route: string, fn: HandleFunction): Server;
```
geekley: I originally intended to replace them with the overloads above, with the direct explicit parameters of `fn` callbacks visible, because that's more practical to handle in the editor (VSCode) after autocomplete, etc. Also, that allows more accurate documentation on each specific overload.
However, I had to keep the overloads with `HandleFunction` to avoid breaking compatibility in the tests of other packages, which referenced `HandleFunction` interface directly by name. Maybe I should have added `@deprecated` here as well.

types/connect/index.d.ts
```diff
@@ -1,94 +1,163 @@
...
+ * proceed to look for the error middleware that was declared after that middleware and invoke it,
+ * skipping any error middleware above that middleware and any non-error middleware below.
+ *
+ * The last step is to actually use the Connect app in a server. The `.listen()` method is a
+ * convenience to start a HTTP server (and is identical to the `http.Server`'s `listen` method).
+ * The app itself is really just a function with three arguments, so it can also be handed to
+ * `.createServer()` in Node.js.
  */
-declare function createServer(): createServer.Server;
+declare function createServer(): connect.Server;
```
geekley: When hovering `const connect = require("connect");` the signature here shows the namespace as "connect" instead of "createServer" beacuse of the import trick above (line 11).

types/connect/index.d.ts
```diff
@@ -1,94 +1,163 @@
...
-    export type SimpleHandleFunction = (req: http.IncomingMessage, res: http.ServerResponse) => void;
-    export type NextHandleFunction = (req: http.IncomingMessage, res: http.ServerResponse, next: NextFunction) => void;
-    export type ErrorHandleFunction = (err: any, req: http.IncomingMessage, res: http.ServerResponse, next: NextFunction) => void;
-    export type HandleFunction = SimpleHandleFunction | NextHandleFunction | ErrorHandleFunction;
+    /** @deprecated `next` is always available to the handler */ export type SimpleHandleFunction = NextHandleFunction;
+    export type NextHandleFunction =
+        (req: Request | http.IncomingMessage, res: http.ServerResponse, next: NextFunction) => void;
+    export type ErrorHandleFunction =
+        (err: any, req: Request | http.IncomingMessage, res: http.ServerResponse, next: NextFunction) => void;
+    export type HandleFunction = NextHandleFunction | ErrorHandleFunction;
```
geekley: Param `req` was originally changed to be `connect.Request` (with originalUrl) but I had to keep `http.IncomingMessage` too to avoid breaking a test in another package which referenced the `HandleFunction` type directly by name (even though its a subtype).
I should probably have `@deprecated` all of those `*HandleFunction` too anyways.

types/connect/index.d.ts
```diff
@@ -1,94 +1,163 @@
...
         route: string;
         handle: ServerHandle;
     }
 
     export interface Server extends NodeJS.EventEmitter {
-        (req: http.IncomingMessage, res: http.ServerResponse, next?: Function): void;
+        /**
+         * The `app` itself is a function. This is just an alias to `app.handle`.
+         */
+        (req: http.IncomingMessage, res: http.ServerResponse, next?: NextFunction): void;
```
geekley: This is the exact same as `handle(req, res, out)` according to code and API.

types/connect/index.d.ts
```diff
@@ -1,94 +1,163 @@
...
-        * the middleware stack.
-        *
-        * @private
-        */
-        handle(req: http.IncomingMessage, res: http.ServerResponse, next: Function): void;
+         * Calling the function will run the middleware stack against the given Node.js http request (`req`)
+         * and response (`res`) objects. An optional function `out` can be provided that will be called if the
+         * request (or error) was not handled by the middleware stack.
+         */
+        handle(req: http.IncomingMessage, res: http.ServerResponse, out?: NextFunction): void;
```
geekley: > An optional function out can be provided that will be called if the request (or error) was not handled by the middleware stack.
[Code](https://github.com/senchalabs/connect/blob/master/index.js#L127) (lines 127, 151) confirms that the out param is optional and called with err as parameter.

types/connect/index.d.ts
geekley: I could have also added the other overloads of Server.listen (since it's the exact same method) but I wasn't sure if I should do that (I only looked at the latest version of Node) so I left only the ones already present.

geekley:
> IMO, too many things are happening in this pull request. It seems like there are a few distinct pieces
@EvanHahn I have added comments to clarify the most significant code changes. The rest is pretty much documentation (not included in the first commit). All code changes are already basically concentrated in the first commit, so it might be easier to review that one first.

types/connect/connect-tests.ts
```diff
@@ -1,45 +1,65 @@
+// npm i --no-save connect
```
EvanHahn: What's the intent behind these two comments? Are they needed?

types/connect/connect-tests.ts
```diff
@@ -1,45 +1,65 @@
...
+app.use(logReq);
+
+// "Throw" a number
+app.use("/403", (req: connect.Request, res: http.ServerResponse, next: connect.NextFunction) => {
+    next(403);
 });
 
 // Use legacy `Function` for `next` parameter.
-app.use((req: http.IncomingMessage, res: http.ServerResponse, next: Function) => {
+// tslint:disable-next-line ban-types
```
EvanHahn: Why does this line need linting disabled?

types/connect/connect-tests.ts
```diff
@@ -1,45 +1,65 @@
...
+});
+
+// Catch only on specific route
+app2.use("/err/catch", stopOnErr);
 
-//create node.js http server and listen on port using connect shortcut
-app.listen(3000);
+// create node.js http server and listen on port using connect shortcut
+app2.listen(3001);
+console.log("http://localhost:3001/");
```
EvanHahn: Why is this lot needed?

types/connect/connect-tests.ts
```diff
@@ -1,45 +1,65 @@
...
-app.use((req: http.IncomingMessage, res: http.ServerResponse, next: connect.NextFunction) => {
-    next(404);
-});
+}
 
 // Stop on errors
-app.use((err: any, req: http.IncomingMessage, res: http.ServerResponse, next: connect.NextFunction) => {
+function stopOnErr(err: any, req: connect.Request, res: http.ServerResponse, next: connect.NextFunction) {
     if (err) {
-        return res.end(`Error: ${err}`);
```
EvanHahn: Why did this need to change?

types/connect/connect-tests.ts
```diff
@@ -1,45 +1,65 @@
...
+        res.statusCode = +err || 500;
```
EvanHahn: This `+err` thing feels weird. If we decide to keep these changes, could we only set the status code if `typeof err === 'number'`?

types/connect/connect-tests.ts
```diff
@@ -1,45 +1,65 @@
...
-//create node.js http server and listen on port
+app.use(stopOnErr);
+
+// create node.js http server and listen on port
 http.createServer(app).listen(3000);
+console.log("http://localhost:3000/");
+
+// ---
+
+const app2 = connect();
```
EvanHahn: Why do we need to create a second Connect app to test the types?

types/connect/connect-tests.ts
```diff
@@ -1,45 +1,65 @@
+// npm i --no-save connect
+// tsc --noEmit false && node connect-tests.js
+
 import * as http from "http";
 import connect = require("connect");
 
 const app = connect();
 
 // log all requests
-app.use((req: http.IncomingMessage, res: http.ServerResponse, next: connect.NextFunction) => {
```
EvanHahn: We're ripping out a lot of tests that ensure that `req` can be an `IncomingMessage`. I still think that should be valid, so could you make sure all of the cases are tested with both types?

types/connect/connect-tests.ts
```diff
@@ -1,45 +1,65 @@
...
 // respond to all requests
-app.use((req: http.IncomingMessage, res: http.ServerResponse) => {
+app.use((req: connect.Request, res: http.ServerResponse) => {
     res.end("Hello from Connect!\n");
 });
 
-//create node.js http server and listen on port
+app.use(stopOnErr);
+
+// create node.js http server and listen on port
```
EvanHahn: Fine to clean up this comment, but I think that should be done separately to shrink the size of these changes.

types/connect/index.d.ts
```diff
@@ -1,94 +1,163 @@
-// Type definitions for connect v3.4.0
+// Type definitions for connect 3.7
```
EvanHahn: Should we leave an exact version number?

types/connect/index.d.ts
```diff
@@ -1,94 +1,163 @@
...
 import * as http from "http";
+import connect = createServer;
 
 /**
- * Create a new connect server.
- * @public
+ * When the `connect` module is required, a function is returned that will construct a new app when called.
+ * This app will store all the middleware added and is, itself, a function.
+ *
+ * The core of Connect is "using" middleware. Middleware are added as a "stack" where incoming
```
EvanHahn: The first two paragraphs seem fine but why is there so much here? Is this a typical amount of documentation?

types/connect/index.d.ts
```diff
@@ -1,94 +1,163 @@
...
+        /** An optional error, to make the app skip to error-handling middleware. */ err?: any
+    ) => void;
 
-    export type SimpleHandleFunction = (req: http.IncomingMessage, res: http.ServerResponse) => void;
-    export type NextHandleFunction = (req: http.IncomingMessage, res: http.ServerResponse, next: NextFunction) => void;
-    export type ErrorHandleFunction = (err: any, req: http.IncomingMessage, res: http.ServerResponse, next: NextFunction) => void;
-    export type HandleFunction = SimpleHandleFunction | NextHandleFunction | ErrorHandleFunction;
+    /** @deprecated `next` is always available to the handler */ export type SimpleHandleFunction = NextHandleFunction;
+    export type NextHandleFunction =
+        (req: Request | http.IncomingMessage, res: http.ServerResponse, next: NextFunction) => void;
```
EvanHahn: This may be a stupid question, but if `Request` is a subtype of `IncomingMessage`, can we just use `Request` here?

EvanHahn (changes requested):
Thanks for doing this.

I'd still prefer this to be multiple PRs for a few reasons:

1. The commit message title includes some, but not all, of the things you're doing in this change. This is really a full rewrite of the package. For example, you've changed TSLint, cleaned up some of the test files, and changed the exported interface from `createServer` to `connect`. Too many things are happening to capture in the title of a single commit or PR.
1. Even with your helpful comments, it's still hard to review. There are lots of things I don't understand the intention behind, even if they're a good idea. For example, you added some comments to the top of the test file, but I don't know why.
1. Someone going through the commit history in the future will also have similar issues breaking down the change. For example, someone trying to see what happened a year from now might not be able to figure out the intention behind some change.
1. Large changes have a bigger chance of causing disruption to users of `@types/connect`. If you made a mistake and I fail to catch it in review, we're going to cause a bunch of people a lot of trouble. If the changes are smaller, that could still happen, but the issues are localized.

geekley:
OK, I understand.
I'm going to respond to your comments later and after that I'll redo the commits with a more step-by-step approach as you said.

Just one question though: should I break it into multiple pull requests or can I have a single pull request with all the commits of each individual change?

I'm new to this (pull requests), so I don't know how it would work if you were going to, for example, approve some change and reject another.
What I mean is: are you able to approve only up to a certain commit in the pull request, or are you only able to approve/reject all of it? Can you approve specific blocks of code only (like staging)?
If you need to edit code, does that create more commits in the pull request, or is it "just changed"?
Also, should I use the same branch for everything?

EvanHahn:
I think there are various pieces that could be done completely in parallel. For example, you could probably add support `req.originalUrl` all on its own; you could change `createServer` to `connect` all on its own; you could update the TSLint file all on its own. Some of the changes might depend on others, but I think it'd be useful to try to break them down into the smallest possible pieces.

PR #40410, opened 2019-11-15
add yog-ral

ssddi456:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/yog-ral/index.d.ts
+++ b/types/yog-ral/index.d.ts
@@ -1,4 +1,5 @@
 // Type definitions for yog-ral 0.20.0
+// Project: https://gitlab.baidu.com/fex/node-ral
 // Definitions by: ssddi456 <https://github.com/ssddi456>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.7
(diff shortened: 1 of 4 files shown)


types/yog-ral/tslint.json
```diff
@@ -0,0 +1,8 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "no-object-literal-type-assertion": false,
+        "only-arrow-functions": false,
+        "no-unnecessary-generics": false
```
PranavSenthilnathan: Please remove these. In general it's good practice to not have these, and in this case these aren't for the definition file itself.

types/yog-ral/index.d.ts
```diff
@@ -0,0 +1,7 @@
+// Type definitions for yog-ral 0.20
+// Project: https://gitlab.baidu.com/fex/node-ral
+// Definitions by: ssddi456 <https://github.com/ssddi456>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.7
```
PranavSenthilnathan: This restricts the library to only be used with TypeScript 3.7+. This file doesn't use any 3.7 specific features, so to allow more users to use this package, this can be bumped down.

PranavSenthilnathan (changes requested):

types/yog-ral/tslint.json
ssddi456: ok， i'll remove these

types/yog-ral/index.d.ts
ssddi456: ok

PranavSenthilnathan (approved):

PR #40417, opened 2019-11-15
[underscore] Fixes sortBy on chained array wrapper

ojhall:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://underscorejs.org/#sortBy

Calling `_([1, 2, 3]).chain()` correctly returns a `_Chain<number, number[]>` type, due to:

```ts
interface UnderscoreStatic {
    chain<T>(obj: T[]): _Chain<T, T[]>;
}
```

However, a subsequent `sortBy` call then returns a `_Chain<number, number>` type, due to:

```ts
interface _Chain<T, V = T> {
    sortBy(iterator?: _.ListIterator<T, any>, context?: any): _Chain<T>;
}
```

As a result, `_([1, 2, 3]).chain().sortBy(x => -x).value()` returns a `number` type instead of `number[]`, which doesn't match the implementation.

This PR sorts that out by trying to more closely match the other `sortBy` type specifications, specifically returning a `_Chain<T, T[]>`.

### DIFF
--- a/types/underscore/index.d.ts
+++ b/types/underscore/index.d.ts
@@ -5306,13 +5306,13 @@ declare module _ {
         * Wrapped type `any[]`.
         * @see _.sortBy
         **/
-        sortBy(iterator?: _.ListIterator<T, any>, context?: any): _Chain<T>;
+        sortBy(iterator?: _.ListIterator<T, any>, context?: any): _Chain<T, V>;
 
         /**
         * Wrapped type `any[]`.
         * @see _.sortBy
         **/
-        sortBy(iterator: string, context?: any): _Chain<T>;
+        sortBy(iterator: string, context?: any): _Chain<T, V>;
 
         /**
         * Wrapped type `any[]`.
--- a/types/underscore/index.d.ts
+++ b/types/underscore/index.d.ts
@@ -5306,13 +5306,13 @@ declare module _ {
         * Wrapped type `any[]`.
         * @see _.sortBy
         **/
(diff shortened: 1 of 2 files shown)


types/underscore/index.d.ts
```diff
@@ -5306,13 +5306,13 @@ declare module _ {
         * Wrapped type `any[]`.
         * @see _.sortBy
         **/
-        sortBy(iterator?: _.ListIterator<T, any>, context?: any): _Chain<T>;
+        sortBy<TSort>(iterator?: _.ListIterator<T, TSort>, context?: any): _Chain<T, T[]>;
```
jgonggrijp: Is the new `TSort` type parameter really necessary, and if so, what for? I wonder whether the `_Chain<T, T[]>` return type wouldn't already be a sufficient solution by itself.

jgonggrijp (changes requested):
I'm just asking for clarification below, but I chose "Request changes" to inform the bot that there should be another round of feedback.

types/underscore/index.d.ts
ojhall: You're right as far as I'm able to see - I cargo-culted it from the other `sortBy` implementation further up. I'll change that back. Thanks!

ojhall: My codebase is happy with it. :+1:

jgonggrijp (approved):
Looks completely sensible to me now. 👍

PR #40423, opened 2019-11-15
Fix relative paths in tests, part 2

sandersn:
Follow up to  #40420

### DIFF
--- a/types/setimmediate/setimmediate-tests.ts
+++ b/types/setimmediate/setimmediate-tests.ts
@@ -1,4 +1,4 @@
-import './index';
+import 'setimmediate';
 
 // $ExpectType number
 const i = setImmediate((...args) => {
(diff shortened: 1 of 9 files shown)


types/setimmediate/setimmediate-tests.ts
```diff
@@ -1,4 +1,4 @@
-import './index';
+import 'setimmediate';
```
ExE-Boss: I&nbsp;totally&nbsp;forgot to&nbsp;do&nbsp;this because&nbsp;I&nbsp;was using&nbsp;`'./index'` to&nbsp;work&nbsp;around an&nbsp;**IntelliSense**&nbsp;bug.

Somehow, it&nbsp;wasn't&nbsp;caught by&nbsp;**TSLint** in&nbsp;#40012, like&nbsp;it&nbsp;was in&nbsp;other&nbsp;packages, where&nbsp;I&nbsp;had to&nbsp;use said&nbsp;workaround during&nbsp;development.

sandersn: Yep there is a bug in the tslint rule that I need to address. This pair of PRs just fixes up Definitely Typed in preparation.

PR #40429, opened 2019-11-15
Generics and typed attributes for Parse.Object and sub classes

pocketcolin:
Updated type definitions to allow generics to be passed to Parse.Object class and all classes extending Parse.Object. Parse.Object.attributes is now typed using the generic type param.

These generics are optional and default to `any`. Update does not include breaking changes.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: http://parseplatform.org/Parse-SDK-JS/api/2.9.1/Parse.Object.html

### DIFF
--- a/types/parse/index.d.ts
+++ b/types/parse/index.d.ts
@@ -1,6 +1,6 @@
 // Type definitions for parse 2.9
 // Project: https://parseplatform.org/
-// Definitions by:  Ullisen Media Group <http://ullisenmedia.com>
+// Definitions by:  Ullisen Media Group <https://github.com/ullisenmedia>
 //                  David Poetzsch-Heffter <https://github.com/dpoetzsch>
 //                  Cedric Kemp <https://github.com/jaeggerr>
 //                  Flavio Negrão <https://github.com/flavionegrao>
--- a/types/parse/index.d.ts
+++ b/types/parse/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for parse 2.2.9
+// Type definitions for parse 2.9
 // Project: https://parseplatform.org/
 // Definitions by:  Ullisen Media Group <http://ullisenmedia.com>
 //                  David Poetzsch-Heffter <https://github.com/dpoetzsch>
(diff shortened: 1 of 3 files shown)


pocketcolin:
@rdhelms

rdhelms (approved):

yagotome (approved):
LGTM! Have you seen similar PR #36172?

rdhelms:
@yagotome We didn’t see that PR! It looks like you guys ran into a challenge that we also noticed. That led to us creating this TS issue:
https://github.com/microsoft/TypeScript/issues/35110
Our workaround was to use `any` as the type parameter’s constraint to avoid breaking changes.

It also looks like that PR was aiming for a wider impact than this PR. We were hoping that this PR could be the first step towards building out even more robust types throughout the namespace.

types/parse/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for parse 2.2.9
+// Type definitions for parse 2.9
```
PranavSenthilnathan: Since your change fixes the lint error, could you remove the "dt-header" rule from the lint file? If you want to you can remove the others too and see what breaks - ideally all of them should be removed.

PranavSenthilnathan (changes requested):

types/parse/index.d.ts
rdhelms: Removed the `dt-header` rule 👍 
Removing the others still results in many errors, but we're thinking of doing another PR soon where we could do a sweep of those.

rdhelms:
@PranavSenthilnathan Any chance of being able to update your review here?

PranavSenthilnathan (approved):

RaschidJFR:
Hey guys @pocketcolin @rdhelms I'm working on a follow-up PR for this in RaschidJFR/DefinitelyTyped@a63fa5e. I've just noticed that dot notation for nested attributes won't work in type check:

```ts
// This throws an error as `Order` object has prop `customer` 
// but it won't accept `customer.user`
Parse.Object.createWithoutData<Order>(id).fetchWithInclude(['customer.user']);
```

It's not a big deal as you can use just do `Parse.Object.createWithoutData(id).fetchWithInclude(['customer.user']);` (without the generic) to avoid the error. But if there's something in your mind to make it better I can gladly include it in the next PR.

rdhelms:
@RaschidJFR - Sorry for the late replies, just now catching up after the holidays. We didn't realize that `fetchWithInclude` accepted dot notation, so that's a good catch.

I'm ok with the `any` workaround when users want to use dot notation, but I would also understand if we wanted to set the params that accept dot notation to simply be `string`.

RaschidJFR:
Tnx @rdhelms I think the `any` will do, as in #41051 most of the functions use dot notation. I think it's pretty good for now.

PR #40284, opened 2019-11-10
feat(puppeteer): 2.0

SimonSchick:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/GoogleChrome/puppeteer/releases/tag/v2.0.0
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

Rinse and repeat (one last time) https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40061

Changes:
 - Removed 0.x
 - Updated 1.x to 2.x in place

Rationale:
 - 0.x (unstable) is 2 years old and no longer maintained, the package will remain
 - 1.x Moved to own folder, package currently has ~250k downloads per week and will probably need minor maintenance in the future.
 - 2.x VERY few changes major release was due to dropping node v6. Used opportunity to bump to TS 3.0 and change some instances of `any` to `unknown` for improved type safety.

PR #40432, opened 2019-11-16
update puppeteer to use the its generated cdp protocol types

connorjclark:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/GoogleChrome/puppeteer/pull/5174

### DIFF
--- a/types/puppeteer/index.d.ts
+++ b/types/puppeteer/index.d.ts
@@ -15,6 +15,9 @@
 
 import { EventEmitter } from "events";
 import { ChildProcess } from "child_process";
+// @ts-ignore: suppress error if using an older version of puppeteer, which don't
+// publish types for the protocol.
+import Protocol from 'puppeteer/lib/protocol';
 
 import * as errors from "./Errors";
 import * as devices from "./DeviceDescriptors";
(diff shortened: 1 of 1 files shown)


types/puppeteer/index.d.ts
```diff
@@ -15,6 +15,9 @@
 
 import { EventEmitter } from "events";
 import { ChildProcess } from "child_process";
+// @ts-ignore: suppress error if using an older version of puppeteer, which don't
```
connorjclark: The goal here is to not force users to use a new version of puppeteer - these types are for power users, so shouldn't force everyone to update. If these are missing, the CDPSession types use a sensible fallback of method: string | number | symbol and params: any, which isn't much different from the previous.

Unfortunately there doesn't seem to be a way to disable this rule. I tried:

```
// tslint:disable-next-line ban-ts-ignore
// tslint:disable-next-line ts-ignore
```

connorjclark:
This shouldn't land until https://github.com/GoogleChrome/puppeteer/pull/5174 is published - I'll comment on this PR when it's ready.

types/puppeteer/index.d.ts
SimonSchick: I'd kindly ask you to way until my puppeteer 2.0 PR lands https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40284

I'm currently working on it.

OT: This feels pretty hacky, if you really want to support TS as first class, why not move the project to TS all together?

connorjclark: Porting puppeteer to TS would be a huge effort.

Yeah it's pretty hacky - I think the "don't force to update for no benefit" constraint can be avoided if we 1) wait for next puppeteer version to be released (@mathiasbynens, maybe we could do a patch release?), which will have this protocol file published and 2) tack this change onto your PR, setting the minimum version to the next (unreleased 2.0.1) version

SimonSchick: @connorjclark iirc the `// Type definitions for PACKAGE MAJOR.MINOR` line does not support specifying patch versions (which makes sense since the public API shouldn't change).
I think it would be reasonable to do a minor release in this case.

Issue #32162, opened 2019-01-15
[@types/d3-array] Please update to latest version v2.0.3


piyushgupta1:
Hello,

D3-array has been updated to major version release v2.0.0. I am using this in typescript based lib 

Typescript Version - v3.2.2
Node Version - v8.11.2

If you know how to fix the issue, make a pull request instead.

- [x] I tried using the `@types/d3-array` package and had problems.
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @gustavderdrache , @borisyankov , @tomwanzek , @denisname

If you do not mention the authors the issue will be ignored.

Ledragon:
Thanks for reporting this, I will try tackling it in the upcoming days (not a promise, but will try my best)

maneetgoyal:
I am using `"d3-array": "^2.2.0",`, can the types be updated please? Getting errors, for instance, the function `groups` is missing from the types.

cc: @gustavderdrache, @borislavjivkov, @tomwanzek, @denisname, @ledragon

Ledragon:
Hi,
Thanks for reporting this. There is actually a lot of work to update all d3 modules to their latest versions... And I am not sure as to how much time any of us can devote to this currently (you might have noticed that not much has been done recently). Feel free to submit a PR to update the types. I can try fitting this one in my schedule, but cannot promise anything...

maneetgoyal:
@Ledragon Thanks for the prompt response. Can you please share the steps to generate type definition from the [source files](https://github.com/d3/d3-array)? Will be useful for creating PRs for other devs, if you're not able to update the definitions by yourself.

Ledragon:
Actually slower on the second reply :-)
Not sure I have "steps" to provide. The way I tend to do these is by reading the documentation, adding the definitions in the types, and adding associated tests. I never personally used the original source to extract definitions automatically. I go read the original code when I am unsure about the types of the parameters or the like.

chriszrc:
Any updates here? Are we really the only 4 people in the world using d3-array in typescript?

Ledragon:
Sorry, time is tight, and as I don't use v5 much these days, this has become a lesser priority... But this stays on my mind still

Schizzomarino:
That would be real good @Ledragon and all other contributors.
[...]

PR #40433, opened 2019-11-16
fix d3-array bin case

caleb15:
bin is lowercase, not uppercase.

When you import bin like so: `import { bin as d3Bin } from 'd3-array';` you get a type error:

> Module '"../../../../../../node_modules/@types/d3-array"' has no exported member 'bin'. Did you mean 'Bin'?ts(2724)

I tried changing it to Bin, uppercase b, but I got an error `'d3Bin' only refers to a type, but is being used as a value here.ts(2693)` when referencing it. I looked at https://github.com/d3/d3-array#bin and it appears that bin is lowercase, so when I tried importing the uppercase Bin it imported the type instead of the actual function I wanted.

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/d3/d3-array#bin

### DIFF
--- a/types/d3-array/d3-array-tests.ts
+++ b/types/d3-array/d3-array-tests.ts
@@ -869,10 +869,10 @@ undef = d3Array.histogram()([])[0].x0 as undefined;
 undef = d3Array.histogram<number | undefined, number | undefined>()([undefined])[0].x0 as undefined;
 
 // number - number
-let binsNumber_Number: Array<d3Array.Bin<number, number>>;
+let binsNumber_Number: Array<d3Array.bin<number, number>>;
 binsNumber_Number = histoNumber_Number([-1, 0, 1, 1, 3, 20, 234]);
 
-let binNumber_Number: d3Array.Bin<number, number>;
+let binNumber_Number: d3Array.bin<number, number>;
 binNumber_Number = binsNumber_Number[0];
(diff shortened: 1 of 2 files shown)


denisname:
This is a backward incompatible change.

gustavderdrache (changes requested):
I disagree with this change. Utlimately, I believe that you are applying the wrong fix to the issue you've identified. Here is your description of the problem:

> When you import `bin` like so: `import { bin as d3Bin } from 'd3-array'`; you get a type error:

The documentation you link to refers to a `bin` _function_, not a `bin` _type_. The change you're making here renames the `Bin` _type_ to `bin`. It doesn't add the `bin` function, which is what I think your goal really is.

I think that the issue here is the same as in #32162 - namely, that `@types/d3-array` is behind the `d3-array` package, and you're trying to use a function that doesn't have type definitions added.

I think there are two possible fixes:

1. If the implementations are the same, then you can downgrade `d3-array` to a version that uses `histogram` instead.
2. Otherwise, you will need to contribute type definitions (and tests) for the new `bin` function.

I can't speak to your project's needs or timeline, so I don't have a recommendation. Unfortunately, this change isn't going to address the need that you seem to have - unless there's something I've missed.

PR #28900, opened 2018-09-14
`node`: [fix] node buffers have a `constructor` property

ljharb:
- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [ ] Test the change in your own code. (Compile and run.)
- [ ] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: `new Buffer('').constructor === Buffer`, in every node version ever
- [x] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #40438, opened 2019-11-16
WIP: feat(node): v13

SimonSchick:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V13.md#13.1.0

This ended up being quite a large PR and I want some feedback on if we can proceed with this or if this needs to be broken down further.

General list of changes:
- Changed constants to be re-exported in `constants` module to avoid duplication.
- Removed various es2015, 16, 17 shims (requires es2018 lib now).
  - ~Have to potentially revert since `///` references don't support `lib` until TS 3.x~ async iteration ones retained
- Various type restrictions in API that could be considered breaking changes.
- Alias ~Moved~ various global interface into `NodeJS` namespace.
  - ~Have to potentially retain some globals to not break other packages.~
- Scheduling in `cluster`
- Use of `unique symbol`
- Removal of redundant interfaces
- Replaced global `EventEmitter` class with interface as it's purely virtual and only there for global declarations.
  - This required various updates in other package since they sometimes used `extends EventEmitter` or `extends NodeJS.EventEmitter`.

This is somewhat WIP and I expect lots of changes.


SimonSchick:
Hey @Flarna mind giving some early input?
This is pretty WIP and I will likely have to revert a thing or 2 :)

Flarna:
Had a few minutes only therefore not seen most of the changes.
2 findings:
- I think it's possible to create IPC channels also via `spawn()` by specifying e.g `[0, 1, 2, 'ipc']`, therefore `ChildProcessIPC` is not a special case for `fork()` only
- it seems you missed the last commit in url (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40118)

types/asynciterator/index.d.ts
```diff
@@ -8,7 +8,7 @@
 
 import { EventEmitter } from "events";
 
-export abstract class AsyncIterator<T> extends NodeJS.EventEmitter {
+export abstract class AsyncIterator<T> extends EventEmitter {
```
Flarna: I think `/// <reference types="node" />` is not needed anymore after this change.

types/node/cluster.d.ts
```diff
@@ -170,11 +170,14 @@ declare module "cluster" {
         prependOnceListener(event: "setup", listener: (settings: ClusterSettings) => void): this;
     }
 
+    const SCHED_NONE: number;
```
Flarna: should we add these two also to `interface cluster`?

types/asynciterator/index.d.ts
SimonSchick: Please sure it's still needed, how else would it know where to get `events` imported from?

types/node/test/events.ts
```diff
@@ -71,10 +71,6 @@ const any: any = 1;
     }
 }
 
-{
-    new events();
```
Flarna: actually this should work, see https://github.com/nodejs/node/blob/64eacd8741f23cb91569faf7ec793c83badb04e3/lib/events.js#L46

types/node/events.d.ts
```diff
@@ -9,31 +7,27 @@ declare module "events" {
...
+        setMaxListeners(n: number): this;
+        getMaxListeners(): number;
+        listeners(event: string | symbol): Function[];
+        rawListeners(event: string | symbol): Function[];
+        emit(event: string | symbol, ...args: any[]): boolean;
+        eventNames(): Array<string | symbol>;
+        listenerCount(type: string | symbol): number;
     }
-
-    export = internal;
```
Flarna: `EventEmitter` constructor is exported like this (and as property `EventEmitter`, see https://github.com/nodejs/node/blob/64eacd8741f23cb91569faf7ec793c83badb04e3/lib/events.js#L46

types/node/test/events.ts
SimonSchick: Interesting, have you seen this used anywhere in the ecosystem?
Things like these are super annoying to model and usually require hacks such as fake inheritance.
Any ideas?

types/asynciterator/index.d.ts
Flarna: There is a dedicated `events` package (folder types\events\), see https://www.npmjs.com/package/@types/events and https://www.npmjs.com/package/events

types/node/globals.d.ts
```diff
@@ -85,13 +85,6 @@ interface Console {
     warn(message?: any, ...optionalParams: any[]): void;
 
     // --- Inspector mode only ---
-    /**
-     * This method does not display anything unless used in the inspector.
-     *  The console.markTimeline() method is the deprecated form of console.timeStamp().
-     *
-     * @deprecated Use console.timeStamp() instead.
-     */
-    markTimeline(label?: string): void;
```
Flarna: I'm not sure about `markTimeline`, `timeline` and `timelineEnd`. Doc says there are only available if inspector is enabled and I don't see them in repl. doc tells the same for `profile` and `profileEnd` but they are present.

types/node/test/events.ts
Flarna: Can't remember a specific place; I expect this is a relict from node 0.x so if this is used somewhere then most likely in old modules.
It's the same/similar for `module` (see https://github.com/nodejs/node/blob/v13.1.0/lib/internal/modules/cjs/loader.js#L1261) and `stream` as also similar.
What was the problem with the exisiting variant `export = internal`?

types/node/globals.d.ts
SimonSchick: They don't exist in either case, see https://github.com/nodejs/node/pull/30346

types/node/test/events.ts
SimonSchick: I figured v13 was a good opportunity to clean things up a bit and remove this `virtual` inheritance nonsense, but it sounds like this could break a few people 🤔

types/node/globals.d.ts
```diff
@@ -187,59 +165,15 @@ declare function clearImmediate(immediateId: NodeJS.Immediate): void;
...
+declare var require: NodeJS.Require;
+declare var module: NodeJS.Module;
 
 // Same as module.exports
 declare var exports: any;
 
 // Buffer class
 type BufferEncoding = "ascii" | "utf8" | "utf-8" | "utf16le" | "ucs2" | "ucs-2" | "base64" | "latin1" | "binary" | "hex";
 
-interface Buffer {
```
Flarna: Why remove this?

SimonSchick: It's not really needed and was a special case that was added just because someone wanted to have access to the type safe constructor ¯\\_(ツ)_/¯

Flarna: I think I remember... was related to webpack to avoid that too much is packed.
So you plan is to break the build for this user? 😕

SimonSchick: That sounds odd, usually types don't affect webpack builds.

Flarna: see https://github.com/DefinitelyTyped/DefinitelyTyped/pull/28900#issuecomment-421829307

SimonSchick: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/28900 there is no mention of `Webpack` in this issue. Seems like it was pure _convenience_, since buffer was moved to a class it most likely doesn't matter anymore.

Flarna: The comment was `Because if i do that, bundlers include an entire Buffer polyfill`, I assumed the bundler was webpack but could be anything.

types/node/module.d.ts
```diff
@@ -1,3 +1,29 @@
...
-    export = NodeJS.Module;
+    class Module {
+        static runMain(): void;
+        static wrap(code: string): string;
+
+        /**
+         * @deprecated Deprecated since: v12.2.0. Please use createRequire() instead.
+         */
+        static createRequireFromPath(path: string): NodeJS.RequireFunction;
+        static createRequire(path: string): NodeJS.RequireFunction;
```
Flarna: A while ago I tried also to move Module from gobal to here but failed for some reason and never created a PR. Maybe because of some typescript version no longer supported...
Anyhow, could you please add `URL` support here: `static createRequire(path: string | URL)` and `import { URL } from "url";`.

That time I had also some you could reuse now: 
```
import m = require('module');
import { createRequireFromPath, createRequire } from 'module';
import { URL } from "url";

let rf: (m: string) => any;

rf = createRequire('mod');
rf = createRequireFromPath('mod');
rf = createRequire(new URL('file:///C:/path/'));

const aModule: NodeModule = new m.Module("s");
const bModule: NodeModule = new m("b", aModule);

const builtIn: string[] = m.builtinModules;
```

SimonSchick: Sure.

types/node/globals.d.ts
SimonSchick: webpack doesn't implicitly emit node polyfills anymore.
And in any case the person can just cast `Buffer.constructor` to `typeof Buffer` 😛

types/node/test/events.ts
SimonSchick: I'm actually pretty curious to see how many people would be affected by this.
Do you think it would be an acceptable risk to do this on v13 or does it have too much potential for breaking changes?

types/node/cluster.d.ts
SimonSchick: Trying to dedupe cluster, will take care of this tho!

types/node/events.d.ts
SimonSchick: Closing in favor of other discussion.

types/node/test/events.ts
Flarna: As far as I know ES6 modules don't allow such exports anyway so seems to be a step forward. Fixing an app is easy so I'm ok with it. Reverting in case of a lot complains should be also not that hard.

SimonSchick: Same applies to `module`, `stream`. Although I won't be touching `stream` for now.

Flarna: Honestly speaking I would be also careful with `module` as this is usually just required to get the static methods exposed there. So at least these should be kept.

SimonSchick:
@Flarna I was able to move a bunch of things from `global.d.ts` including `console` and `process` into their own files and remove some redundant declarations. 🎉

SimonSchick:
@RyanCavanaugh Can I maybe get some input from you and the larger DT team?
Node typings are probably *the* most popular ones around I want to make sure to have plenty 👀 and thoughts on those changes.

SimonSchick:
Also, going to add changes for v13.2 soon-ish.

SimonSchick:
@sandersn I'm not sure who's in an authority position but can we potentially halt all merged to the node package until v13 is done?
I have to constantly rebase and manually update the v12 folder.
Also the build has been broken for days (passes locally).

Is there any way we can move forward in getting this in?

SimonSchick:
Great, so the requirement for file inclusion changed...
How do I fix this correctly?
Can we skip this for now?

sandersn:
You haven't touched tsconfig, so you should be able to merge from master to get rid of the now-excess d.ts references in the files list. See https://github.com/microsoft/types-publisher/issues/708 for details.

sandersn:
As for reviewers from the TS team, @rbuckton is the local node expert, although he's on vacation for Thanksgiving until next Monday.

SimonSchick:
I missed updating the moved tsconfig file yes, I already fixed it.

sandersn:
Now there's a failure with an unused test file:

Unused file /home/vsts/work/1/s/types/node/v12/test/async_hooks.ts (used files: ["assert.ts","buffer.ts","child_process.ts","cluster.ts","crypto.ts","dgram.ts","events.ts","fs.ts","global.ts","http.ts","http2.ts","net.ts","os.ts","path.ts","perf_hooks.ts","process.ts","querystring.ts","readline.ts","repl.ts","stream.ts","string_decoder.ts","tls.ts","tty.ts","util.ts","v8.ts","vm.ts","worker_threads.ts","zlib.ts"])

This looks to me like it's just missing from v12/tsconfig.json.

SimonSchick:
Yea I'm on it, I tried to be too smart with git :)

SimonSchick:
Seems like there are more packages that violate the new tsconfig rules:

```
Found 6232 packages.
Parsing in parallel...
Error: Unused file /Users/simon/projects/DefinitelyTyped/types/flux/lib/FluxContainer.d.ts
```

Btw it would be great if it would emit a list of *all* files, not just the first it can find and bail.

sandersn:
Is that from a local run? I don't see it on my local run. You merged from master already, right?

SimonSchick:
Well no, but apparently npm doesn't like updating sometimes when using non npm packages eg. `github:Microsoft/types-publisher#production` and it also doesn't show up as outdated 😛 
Updating that package seems to have fixed it.

SimonSchick:
@rbuckton Can we move forward on this before the bot it's thing again (mildly frustrating btw).

SimonSchick:
I'm growing progressively dissatisfied with this bot...

SimonSchick:
@rbuckton (sorry for the many mentions), but what do?
If I re-open the PR the build will probably fail again 😅

SimonSchick:
ping

Flarna:
@SimonSchick To keep a PR alive it's needed to keep the commit date young. a merge from master, rebase or just `git commit --amend` without a change followed by a push should be enough as keep alive. On the other hand creating a new PR based on old branch result in a fast close by bot.

SimonSchick:
I'm hesitant to create a new PR after seeing how this one went.
Can we *please* have some sort of exception for the `node` package for the bot not to close PRs for it, or at least keep them open longer?
The build/ci process has been notoriously unstable (randomly breaking) and lengthy for anything that touches the node typings.

PR #40454, opened 2019-11-18
updates to titanium

ffMathy:
Losened some types that were too strong.


ffMathy:
This is not "unowned". It clearly has listed owners (me included). How do we fix that?

types/alloy/index.d.ts
```diff
@@ -148,7 +145,7 @@ interface AlloyInterface {
    * @param name Name of controller to instantiate.
    * @param args Arguments to pass to the controller
    */
-  createController(name: string, args?: any): AlloyController;
+  createController(name: string, args?: any): AlloyController & any;
```
sheetalkamat: `AlloyController & any` = `any` .. Either use specific type or `any` if its suppose to be anything.

sheetalkamat: Refer to https://github.com/DefinitelyTyped/DefinitelyTyped#common-mistakes

sheetalkamat (changes requested):

types/alloy/index.d.ts
ffMathy: Fixed! Sorry about that. That was a brainfart.

janvennemann:
@ffMathy please make sure to reflect the changes to the Titanium typings you made here back into our definition generator. Otherwise they will be overwritten the next time we auto generate them from our docs.

ffMathy:
@ffMathy can you link to that definition generator?

janvennemann:
Sure, you can find it here: [typescript_generator.js](https://github.com/appcelerator/docs-devkit/blob/develop/packages/titanium-docgen/generators/typescript_generator.js)

PR #40467, opened 2019-11-18
[@types/mongoose] improve mongoose create type

sakulstra:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/typegoose/typegoose/issues/129>>

This is the same pr, but for an older version of mongoose: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/33781

---

I've taken this solution from the referenced pr, but am wondering if a more strict `new(doc?: Omit<T, keyof Document>): T;` would be a better fit.

The current solution will allow you to omit required fields:
```
export interface ITest {
  foo: string;
  bar: string;
}

const testSchema: Schema = new mongoose.Schema({
  foo: String,
  bar: String,
});

export interface ITestModel extends ITest, Document {}

const Test: Model<ITestModel> = mongoose.model<ITestModel>('Test', testSchema);

// okay
new Test({ foo: 'foo', bar: 'bar' });
// missing key, but okay
new Test({ foo: 'foo' });
// error: key wrong type
new Test({ foo: 3 });
// error: key doesn't exist
new Test({ zar: 'bar' });
```

When using `new(doc?: Omit<T, keyof Document>): T;` instead `new Test({ foo: 'foo' });` would fail due to the missing key.

### DIFF
--- a/types/mongoose/index.d.ts
+++ b/types/mongoose/index.d.ts
@@ -2892,7 +2892,7 @@ declare module "mongoose" {
      *   Model#ensureIndexes. If an error occurred it is passed with the event.
      *   The fields, options, and index name are also passed.
      */
-    new(doc?: any): T;
+    new(doc?: Partial<T>): T;
 
     /**
      * Requires a replica set running MongoDB >= 3.6.0. Watches the underlying collection for changes using MongoDB change streams.
--- a/types/mongoose/index.d.ts
+++ b/types/mongoose/index.d.ts
@@ -33,6 +33,7 @@
 //                 Chathu Vishwajith <https://github.com/iamchathu>
 //                 Tom Yam <https://github.com/tomyam1>
 //                 Thomas Pischulski <https://github.com/nephix>
+//                 Lukas Strassel <https://github.com/sakulstra>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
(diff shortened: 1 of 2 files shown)


sakulstra:
after writing this, i wonder if we should do the same for `create` and `insertMany`?

Ostro:
@sakulstra, just running into the same problem yesterday, 
You need to be a little smoother because this gonna break in case you don't type your model. 

`mongoose.model<MyType>(...)` => this gonna work, 
but `mongoose.model(...)` not

sakulstra:
Ah... didn't think about that at all, thanks for pointing out.
Added a failing test-case, but have no clue on how to correctly approach a fix to this (still very new to ts) :thinking:

The problem is that - as you pointed out - `mongoose.model(...)` will default to `<Document>` when no generic is set.

One fix to this could be to default to `any`, but I guess this will have some drawbacks throughout the types as it will return `Model<T = any>` in case T is not set.
```git
diff --git a/types/mongoose/index.d.ts b/types/mongoose/index.d.ts
index c64dbbfb45..0aad88554c 100644
--- a/types/mongoose/index.d.ts
+++ b/types/mongoose/index.d.ts
@@ -156,7 +156,7 @@ declare module "mongoose" {
    * @param collection (optional, induced from model name)
    * @param skipInit whether to skip initialization (defaults to false)
    */
-  export function model<T extends Document>(name: string, schema?: Schema, collection?: string,
+  export function model<T extends Document = any>(name: string, schema?: Schema, collection?: string,
```

Any ideas on how to properly fix this?

PR #40482, opened 2019-11-19
[node/globals] Proposal: Overload JSON.parse with a common nodejs use case (Buffer argument)

GeorgeTaveras1231:
I'm proposing this change because the standard ES types that come from typescript are too restrictive. My original intention was actually to contribute a similar change to [Typescript](https://github.com/microsoft/TypeScript/blob/master/lib/lib.es5.d.ts#L1052-L1074) to allow any object that implements the `toString` method. However, this same request was rejected in [this issue](https://github.com/microsoft/TypeScript/issues/11842) with the argument that it allows too many potentially erronous usages of `JSON.parse`, such as `JSON.parse({})`

Please fill in this template.

    - [JSON.parse ES 5.1 spec](https://www.ecma-international.org/ecma-262/5.1/#sec-15.12.2)
    - [ToString abstract operation in ES 5.1 spec - Which Buffer is compatible with](https://www.ecma-international.org/ecma-262/5.1/#sec-9.8)


SimonSchick (changes requested):
Let's not special-case for Buffer.

```
> JSON.parse({ toString() { console.log('e'); return '{}'; } })
e
{}
```

JSON.parse will invoke `toString()` on anything that is passed, this should be fixed in the base libs instead (if at all).

Just my 2c: I'd suggest not adding this kind of handling because *very* often this implicit conversion is now what the developer intends which can lead to undetected bugs.

SimonSchick:
To explain the issue a little further:
If we allowed all objects with a `toString()` method to be passed it would compromise type safety as pretty much every object and primitive type has this method.

GeorgeTaveras1231:
@SimonSchick 
> JSON.parse will invoke toString() on anything that is passed, this should be fixed in the base libs instead (if at all)

As I mentioned, there was a request to address this on the base libs but your point:
> If we allowed all objects with a toString() method to be passed it would compromise type safety as pretty much every object and primitive type has this method.

Was exactly the reason why it was rejected.

I'd understand if you are against treating Buffer as a special case. 

My idea is that Buffers are many times used interchangeably with Strings, and I think many developers prefer to avoid converting it to a string unnecessarily when passing them around from one module to another. And since this API is completely safe, I thought it would only help to have it officially supported.

galkin (commented):
@GeorgeTaveras1231, I think typescript community should reject this pull.
Your proposed changes look like decreasing type system safety for the core module.

galkin (changes requested):
```typescript
Buffer.from("test") === "test" // false
Buffer.from("test") == "test" // true
```
so, let's not merge this pull.

nguymin4:
Totally agree with Simon: Explicit is better than Implicit.

GeorgeTaveras1231:
Thanks for al your feedback 😅

I will admit though; I dont see how this change reduces type safety, since using Buffers as the first argument to JSON.parse is safe (AFAIK). And it is explicitly only supporting that, not every object that has a toString method (which would certainly reduce type safety). Maybe im missing something. I also think that forcing the `toString` conversion could also hurt type safety.

For example:

```ts
const mySerializer = (jsonString) => {
  //call toString to support Buffer
  return JSON.parse(jsonString.toString())
}

MySerializer({whoops: ‘this is not type safe’})

```

PR #40491, opened 2019-11-19
Remove named exports from validator index file

### LINKED ISSUE #40477: [@types/validator] Circular definition of import alias 'Validator'.

  - Authors: @MunifTanjim @samuelneff @CodeAnimal @drinchev @babolivier @kukoo1 @oktapodia @MorpheusXAUT @TitaneBoy @zjy01 @nidzov @Raigen @todd @nrschultz @Thomas-B @Antoine38660 @smff @truongkhanhduy95 @emmanuelgautier @dancrumb @lalayueh @rohitsud @Torhal @tgfjt @chrootsu @IOAyman @louy @kacepe @deptno @builtinnya @qqilihq @keatz55

Example:
```ts
import validator from 'validator';
```

```
5:11:26 PM: ERROR in /opt/build/repo/node_modules/@types/validator/index.d.ts
5:11:26 PM: ERROR in /opt/build/repo/node_modules/@types/validator/index.d.ts(1247,8):
5:11:26 PM: TS2303: Circular definition of import alias 'Validator'.
```

It is caused by the release of @types/validator@12.0.0 (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40192)

louy:
Fixes #40477 

There are no named exports in the validator package as you can see from the source code here
https://github.com/validatorjs/validator.js/blob/b5a1d1f/src/index.js

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/validatorjs/validator.js/blob/b5a1d1f/src/index.js


louy:
Btw, validator@12 was only published yesterday and it was broken since it was published, so I appreciate that this is a breaking change from the previous version but since it never worked anyway I thought it should be fine

types/convict/convict-tests.ts
```diff
@@ -1,6 +1,6 @@
 import * as convict from 'convict';
-import * as validator from 'validator';
```
louy: these imports are irrelevant so I removed them to remove the dependency between the convict package and the validator/js-yaml type definitions

types/validator/index.d.ts
```diff
@@ -13,1237 +13,1237 @@
...
-    max?: number;
-}
-
-/**
- * Check if the string's length (in UTF-8 bytes) falls in a range.
- *
- * @param [options] - Options
- */
-export function isByteLength(str: string, options?: IsByteLengthOptions): boolean;
+declare namespace validator {
```
louy: I just moved everything in this file inside a namespace and changed the export statement at the bottom

types/validator/index.d.ts
```diff
@@ -13,1237 +13,1237 @@
...
- * @param chars - The characters are used in a `RegExp` and so you will need to escape some chars, e.g. `whitelist(input, '\\[\\]')`.
- */
-export function whitelist(input: string, chars: string): string;
-
-/**
- * Converts to string.
- */
-export function toString(input: any): string;
-
-import * as Validator from './';
```
louy: this is the line that was causing the bug

types/convict/convict-tests.ts
oktapodia: Are you sure that this usage won't break this test https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0a4230ce80fdd64586301e88bce0666660f2b37e/types/convict/convict-tests.ts#L69 ?

louy:
I think the benchmark job is stuck?

types/convict/convict-tests.ts
louy: yes since these tests aren't actually being run but only type-checked

esetnik:
> I think the benchmark job is stuck?
The benchmark job takes several hours and doesn't seem to show any output for a long time. I thought the same when I pushed a PR to DefinitelyTyped but it did eventually go.

louy:
Huh that's strange. Fair enough good to know. Thanks :)

types/convict/convict-tests.ts
oktapodia: If it is not ran, I think this is better to totally remove the usage instead of making it `void 0`, that would be less confusing for the next contributor

types/convict/convict-tests.ts
```diff
@@ -1,6 +1,6 @@
...
-import { safeLoad } from 'js-yaml';
```
oktapodia: Same as above here as it must be a function

zoxon (approved):

oktapodia (changes requested):
Couple of comments

esetnik:
FYI @louy validator@12.1.0 was released today. Maybe we should update the type version as well as part of this pull request as `@types/validator@12.0.0` is totally broken.

types/convict/convict-tests.ts
louy: I thought about doing that, but it would be a significant change to this file, and it looks like the file itself was copy-pasted from the original package’s documentation, so I wanted to keep this file intact as much as possible. Wdyt?

louy:
@esetnik I can’t see any changes that affects the types in v12.1. They introduced tree shaking but the signature remains the same. Am I missing something?

types/convict/convict-tests.ts
oktapodia: If a new maintainer is joining the crew, it may be very difficult to understand it... I personnaly prefer to keep the stuff easy to understand, maybe the advice from @esetnik or @zoxon can be useful?

pscanf:
Maybe @esetnik just meant that types that might have changed due to the new version of validator should also be fixed in this PR. Not that I've done anything to deserve a saying, but I'd prefer this to be merged as it is, indeed because the types are now totally broken, so it's better to have a usable version as soon as possible.

types/convict/convict-tests.ts
louy: This is taking longer than I hoped, so instead, I’ll just revert those changes. Wdyt?

bondz (approved):

types/convict/convict-tests.ts
oktapodia: ok for me

oktapodia:
any updates?

louy:
yep updated

esetnik:
> @esetnik I can’t see any changes that affects the types in v12.1. They introduced tree shaking but the signature remains the same. Am I missing something?

Sorry I was just mentioning it as I thought we might want to update this line https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7edeb1789f4587c3620d8da87c1cea570e07ce6d/types/validator/index.d.ts#L1 to reference the fact that it's also covering the latest version of validator. But I'm happy to see this merged as is.

oktapodia (approved):

types/validator/index.d.ts
weagle08: Was just coming to fix this. Thanks!

PR #40495, opened 2019-11-19
fix: [react] Deleted unused type ReactComponentElement

keroxp:
Please fill in this template.

If adding a new definition:

N/A

If removing a declaration:

N/A

## Description

Deleted `ReactComponentElement` as it is not referred from anywhere and broken. In the latest TypeScript, `keyof X` cannot be applied to `string` (is resolved as `string | number`)

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -111,7 +111,7 @@ declare namespace React {
     }
 
     interface ReactComponentElement<
-        T extends keyof JSX.IntrinsicElements | JSXElementConstructor<any>,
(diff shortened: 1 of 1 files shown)


ferdaber:
People may be using this inside their application code, we should instead mark it as `@deprecated`.

keroxp:
スクリーンショット 2019-11-20 15 00 04

@ferdaber I found that this error only occurred when compiling patched type definition files on [Deno](https://github.com/denoland/deno) v0.24.0. I don't know why but it may not be problem in Node.js world. So I close the PR. Thank you.

PR #40506, opened 2019-11-19
[@types/multimap] Add types for multimap NPM package

tallenstudios:


types/multimap/index.d.ts
```diff
@@ -0,0 +1,72 @@
...
+     * @return all the values in the map
+     */
+    values(): { next: () => { value: any, done: boolean } };
+}
+
+interface multimapConstructor {
+    new(iterable?: any): Multimap;
+}
+
+declare const Multimap: multimapConstructor;
```
sheetalkamat: Instead of constant this need to be a class.. Refer https://github.com/DefinitelyTyped/DefinitelyTyped#common-mistakes

sheetalkamat (changes requested):

PR #40511, opened 2019-11-19
[promise.allsettled] Added option to provide a type for PromiseRejection when using the allSettled function

nikmash:

### DIFF
--- a/types/promise.allsettled/implementation.d.ts
+++ b/types/promise.allsettled/implementation.d.ts
@@ -5,6 +5,7 @@ type PromiseResultTuple<T extends [unknown, ...unknown[]]> = {[P in keyof T]: Pr
 
 declare function allSettled(): Promise<[]>;
 declare function allSettled<T extends [unknown, ...unknown[]]>(iterable: PromiseTuple<T>): Promise<PromiseResultTuple<T>>;
+// tslint:disable-next-line no-unnecessary-generics
 declare function allSettled<T, E>(iterable: Iterable<Promise<T> | T>): Promise<Array<PromiseResult<T, E>>>;
 
 export = allSettled;
--- a/types/promise.allsettled/implementation.d.ts
+++ b/types/promise.allsettled/implementation.d.ts
@@ -5,6 +5,6 @@ type PromiseResultTuple<T extends [unknown, ...unknown[]]> = {[P in keyof T]: Pr
 
 declare function allSettled(): Promise<[]>;
 declare function allSettled<T extends [unknown, ...unknown[]]>(iterable: PromiseTuple<T>): Promise<PromiseResultTuple<T>>;
-declare function allSettled<T>(iterable: Iterable<Promise<T> | T>): Promise<Array<PromiseResult<T>>>;
+declare function allSettled<T, E>(iterable: Iterable<Promise<T> | T>): Promise<Array<PromiseResult<T, E>>>;
 
 export = allSettled;
--- a/types/promise.allsettled/index.d.ts
+++ b/types/promise.allsettled/index.d.ts
@@ -2,7 +2,7 @@
 // Project: https://github.com/ljharb/promise.allsettled#readme
 // Definitions by: Martin Jurča <https://github.com/jurca>
 //                 Elizabeth Lorenz <https://github.com/kisaraofpern>
-//		   Nikhil Mashettiwar <https://github.com/nikmash>
+//                 Nikhil Mashettiwar <https://github.com/nikmash>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.1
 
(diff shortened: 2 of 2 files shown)


nikmash:
Was seeing compiler errors when trying to use the `reason` property returned by the `allSettled` function because the type is `unknown` in an error scenario. The underlying type takes in a generic so if we are able to use a generic in the `allSettled` function then the compiler will be aware of the type that the `reason` property has in the `PromiseRejection` scenario.

nikmash:
@ljharb @jurca if y'all don't mind please review! 🙏

types/promise.allsettled/implementation.d.ts
```diff
@@ -5,6 +5,7 @@ type PromiseResultTuple<T extends [unknown, ...unknown[]]> = {[P in keyof T]: Pr
 
 declare function allSettled(): Promise<[]>;
 declare function allSettled<T extends [unknown, ...unknown[]]>(iterable: PromiseTuple<T>): Promise<PromiseResultTuple<T>>;
-declare function allSettled<T>(iterable: Iterable<Promise<T> | T>): Promise<Array<PromiseResult<T>>>;
+// tslint:disable-next-line no-unnecessary-generics
+declare function allSettled<T, E>(iterable: Iterable<Promise<T> | T>): Promise<Array<PromiseResult<T, E>>>;
```
ljharb: should E default to unknown, so it’s optional?

jurca:
@nikmash Hello, and you for submitting this PR.

While I have originally toyed with allowing this (that's why it is there for the underlying type), in the end I decided against it.

The reason for that is that you cannot have a reasonable guarantee what value will be used to reject the promise, since any value, including `undefined` can be used to reject a promise or passed to the `throw` statement.

I'm pretty sure that's also the reason why TypeScript does not allow you to track the type of the value rejecting a Promise - typescript/lib/lib.es5.d.ts:1399-1414:

```typescript
interface Promise<T> {
    /**
     * Attaches callbacks for the resolution and/or rejection of the Promise.
     * @param onfulfilled The callback to execute when the Promise is resolved.
     * @param onrejected The callback to execute when the Promise is rejected.
     * @returns A Promise for the completion of which ever callback is executed.
     */
    then<TResult1 = T, TResult2 = never>(onfulfilled?: ((value: T) => TResult1 | PromiseLike<TResult1>) | undefined | null, onrejected?: ((reason: any) => TResult2 | PromiseLike<TResult2>) | undefined | null): Promise<TResult1 | TResult2>;

    /**
     * Attaches a callback for only the rejection of the Promise.
     * @param onrejected The callback to execute when the Promise is rejected.
     * @returns A Promise for the completion of the callback.
     */
    catch<TResult = never>(onrejected?: ((reason: any) => TResult | PromiseLike<TResult>) | undefined | null): Promise<T | TResult>;
}
```

Notice that the `reason` argument passed to the `onrejected` callback is of the `any` type.

Because of this, adding this type parameter would result in a silent type cast in a disguise. That's also the reason why I opted for `unknown` instead of `any` - sacrificing some convenience for better type safety.

I hope you understand.

I am inclined to reject this PR for these reasons. Should you prefer to change the type from `unknown` to `any` to match the declarations that are included with TypeScript, I would be willing to accept that.

Do you agree, @ljharb ?

ljharb:
I don’t agree; any value can be in almost any position, but the point of TS is to narrow that down, and TS *should* allow the type of a rejected promise to be similarly restricted. (you make a good case that until TS fixes this, that these types probably shouldn’t attempt to)

unknown is certainly better than any, always - but what’s the point of a type system if it’s not going to be used to narrow types at every opportunity?

jurca:
Well, I would prefer keeping the `unknown` type as well - that's why I used it in the first place.

nikmash:
@jurca in this case I'm using the `Axios` module which would actually have a type of `AxiosError` in the case that the promise is rejected. That's why I need to define a type for `E`.

jurca:
@nikmash I understand, however, you have no guarantee that the error will always be an `AxiosError` instance.

I would recommend to use the following code in your case:

```typescript
const results = await allSettled([...])
for (const result of results) {
  if (result.status === 'rejected') {
    if (result.reason instanceof AxiosError) {
     const reason = result.reason as AxiosError
     // do something
    } else {
      throw result.reason
    }
  }
}
```

nikmash:
@jurca ah I had to dig through the typescript code, but now I understand what you mean by the fact that there's no guarantee that the error is an `AxiosError`. Although the code above wouldn't work because `AxiosError` is an interface not a value so you cannot call `instanceof` on an interface. 

Instead of `result.reason instanceof AxiosError` I believe a User Defined type guard would need to be used like this:
```
  isAxiosError(arg: any): arg is AxiosError {
    return arg && arg.isAxiosError && typeof(arg.isAxiosError) === "boolean";
  }
```

But thanks for your help @jurca I think this makes sense based on the typescript spec. Until typescript narrows down the possible types of a rejected promise I guess we can't fix this.

PR #40521, opened 2019-11-20
fix(final-form-focus): Pass FormValues to Decorator

danieldelcore:
Decorator expects a generic of `FormValues` to be passed in, otherwise it defaults to `object`

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [https://github.com/final-form/final-form/blob/master/src/index.d.ts#L286](source)

### DIFF
--- a/types/final-form-focus/tslint.json
+++ b/types/final-form-focus/tslint.json
@@ -1,1 +1,3 @@
-{ "extends": "dtslint/dt.json" }
+{
+    "extends": "dtslint/dt.json"
+}
--- a/types/final-form-focus/index.d.ts
+++ b/types/final-form-focus/index.d.ts
@@ -15,9 +15,11 @@ export type GetInputs = () => FocusableInput[];
 
 export type FindInput = (inputs: FocusableInput[], errors: object) => FocusableInput | undefined;
 
-export default function createDecorator(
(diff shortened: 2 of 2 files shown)


danieldelcore:
CC: @huan086

huan086:
I added the generics initially, but DefinitelyTyped CI failed. Any idea how to solve this?

huan086 (approved):

danieldelcore:
Thanks! looks like most packages turn that rule off. If you're happy with that, i just updated my pr.

Let me know what you think :)

huan086 (approved):

types/final-form-focus/tslint.json
```diff
@@ -1 +1,6 @@
-{ "extends": "dtslint/dt.json" }
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
```
sheetalkamat: Do not add rules to tslint and disable the rule for package.. Fix errors instead or disable error at the location it fails.

sheetalkamat (changes requested):

types/final-form-focus/tslint.json
danieldelcore: Ok sure thing 😊
Just pushed the changes, let me know what you think, @sheetalkamat

PR #40525, opened 2019-11-20
Adding HeaderLines type for mailparser

Amit-A:
Typing recently introduced `headerLines`; see this commit: nodemailer/mailparser@56ea813

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodemailer/mailparser/commit/56ea81366ba045aa7dd12d5847794f75fb54d71a

### DIFF
--- a/types/mailparser/index.d.ts
+++ b/types/mailparser/index.d.ts
@@ -42,7 +42,13 @@ export type HeaderValue = string | string[] | AddressObject | Date | StructuredH
  * A Map object with lowercase header keys.
  */
 export type Headers = Map<string, HeaderValue>;
-
+/**
+ * An array of raw header lines
+ */
+export type HeaderLines = Array<{
+    key: string;
+    line: string
+}>;
 /**
  * Address details.
  */
@@ -108,6 +114,10 @@ interface AttachmentCommon {
 	 * A Map value that holds MIME headers for the attachment node.
 	 */
 	headers: Headers;
+    /**
+	 * An array of raw header lines for the attachment node.
+	 */
+	headerLines: HeaderLines;
 	/**
 	 * A MD5 hash of the message content.
 	 */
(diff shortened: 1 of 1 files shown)


types/mailparser/index.d.ts
```diff
@@ -177,6 +187,10 @@ interface ParsedMail {
 	 * - `date` value is a Date object.
 	 */
 	headers: Headers;
+    /**
+	 * An array of raw header lines
+	 */
+	headerLines: HeaderLines;
```
amcasey: I know nothing about the underlying library, but it looks like `false` might also be a valid value?
https://github.com/nodemailer/mailparser/blob/9636083b806e60a953a640a4f833a700931b5618/lib/mail-parser.js#L137

types/mailparser/index.d.ts
```diff
@@ -42,7 +42,13 @@ export type HeaderValue = string | string[] | AddressObject | Date | StructuredH
  * A Map object with lowercase header keys.
  */
 export type Headers = Map<string, HeaderValue>;
-
+/**
+ * An array of raw header lines
+ */
+export type HeaderLines = Array<{
```
amcasey: Read-only array?

amcasey (changes requested):
Quick questions.

types/mailparser/index.d.ts
Amit-A: You're right. `Headers` doesn't allow for a `false` value either despite being initialized as false. Both should be fine though - this is an email parsing library - receiving an email with no headers is highly unlikely (from, to, subject, etc. are headers).

types/mailparser/index.d.ts
Amit-A: Yup, fixed.

Amit-A (commented):
HeaderLines: Array -> ReadonlyArray

PR #40533, opened 2019-11-20
Added definitions for protoo-server

antoniom:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/protoo-server/index.d.ts
+++ b/types/protoo-server/index.d.ts
@@ -1,0 +1,92 @@
+// Type definitions for protoo-server 4.0
+// Project: https://protoojs.org
+// Definitions by: Antonis Balasas <https://github.com/antoniom>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/// <reference types="node" />
+
+import { Server as HttpServer, IncomingMessage } from 'http';
+import { Server as HttpsServer } from 'https';
+import { Socket } from 'net';
+import { IServerConfig } from 'websocket';
+
+export interface ConnectionRequestInfo {
+    request: IncomingMessage;
+    origin: string;
+    socket: Socket;
+}
+
+export interface ProtooRequest {
+    request: true;
+    id: number;
+    method: string;
+    data: any;
+}
+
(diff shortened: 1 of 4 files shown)


types/protoo-server/index.d.ts
```diff
@@ -0,0 +1,92 @@
...
+
+export type NotificationCb = (notification: ProtooNotification) => void;
+
+export class WebSocketServer {
+    constructor(server: HttpServer | HttpsServer, options?: IServerConfig);
+    stop(): void;
+    on(eventType: 'connectionrequest', callback: ConnectionRequestCb): void;
+}
+
+export class WebSocketTransport {}
```
sheetalkamat: This definition cannot be present here since this library does not export anything like this.
```js
const m = require("protoo-server");
console.log(m.WebSocketTransport );
```
this reports `undefined`

types/protoo-server/index.d.ts
```diff
@@ -0,0 +1,92 @@
...
+    closed: boolean;
+    constructor();
+    createPeer(peerId: string, transport: WebSocketTransport): Promise<Peer>;
+    hasPeer(peerId: string): boolean;
+    getPeer(peerId: string): Peer;
+    close(): void;
+    on(eventType: 'close', callback: EmptyCb): void;
+}
+
+export class Peer {
```
sheetalkamat: This definition cannot be present here since this library does not export anything like this.
```js
const m = require("protoo-server");
console.log(m.Peer);
```
reports `undefined`

sheetalkamat (changes requested):

types/protoo-server/index.d.ts
antoniom: @sheetalkamat What would be the correct way to solve this?
Can I replace my Peer class code
```ts
export class Peer {/**/}
```
With this?
````ts
class PeerCls {}
export type Peer = PeerCls
```

sheetalkamat: There is no Peer class in the module so you would need to define it as interface if it is suppose to define type.

PR #40537, opened 2019-11-20
[chart.js] Add DoughnutModel interface

jordanweiler:
The existing `Model` interface didn't have most of the Doughnut chart model properties defined. I'm creating a new `DoughnutModel` interface that is specific to the Doughnut chart. I don't have time to create a new model interface for each chart type but hopefully that can be incrementally updated in future PRs.

Please fill in this template.

  * https://github.com/chartjs/Chart.js/blob/master/src/controllers/controller.doughnut.js#L242-L254
  * https://github.com/chartjs/Chart.js/blob/master/test/specs/controller.doughnut.tests.js

### DIFF
--- a/types/chart.js/index.d.ts
+++ b/types/chart.js/index.d.ts
@@ -97,26 +97,35 @@ interface MetaData {
     _chart: Chart;
     _datasetIndex: number;
     _index: number;
-    _model: Chart.ModelType;
+    _model: Model;
     _start?: any;
-    _view: Chart.ModelType;
+    _view: Model;
     _xScale: Chart.ChartScales;
     _yScale: Chart.ChartScales;
     hidden?: boolean;
 }
 
+// NOTE: This model is generic with a bunch of optional properties to represent all types of chart models.
+// Each chart type defines their own unique model structure so some of these optional properties
+// might always have values depending on the chart type.
 interface Model {
     backgroundColor: string;
+    borderAlign?: Chart.BorderAlignment;
     borderColor: string;
     borderWidth?: number;
+    circumference?: number;
     controlPointNextX: number;
     controlPointNextY: number;
     controlPointPreviousX: number;
     controlPointPreviousY: number;
+    endAngle?: number;
     hitRadius: number;
+    innerRadius?: number;
+    outerRadius?: number;
     pointStyle: string;
     radius: string;
     skip?: boolean;
+    startAngle?: number;
     steppedLine?: undefined;
     tension: number;
     x: number;
@@ -148,8 +157,6 @@ declare namespace Chart {
     type BorderAlignment = 'center' | 'inner';
 
     type BorderWidth = number | { [key in PositionType]?: number };
-
-    type ModelType = Model | DoughnutModel;
 
     interface ChartArea {
         top: number;
@@ -804,6 +811,7 @@ declare namespace Chart {
         easing?: Easing;
     }
 
+    // Model used with the doughnut chart
     interface DoughnutModel {
         backgroundColor: ChartColor;
         borderAlign: BorderAlignment;
--- a/types/chart.js/index.d.ts
+++ b/types/chart.js/index.d.ts
@@ -97,9 +97,9 @@ interface MetaData {
     _chart: Chart;
     _datasetIndex: number;
     _index: number;
-    _model: Model;
+    _model: Chart.ModelType;
     _start?: any;
-    _view: Model;
+    _view: Chart.ModelType;
     _xScale: Chart.ChartScales;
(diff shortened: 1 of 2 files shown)


types/chart.js/index.d.ts
```diff
@@ -801,6 +803,20 @@ declare namespace Chart {
         lazy?: boolean;
         easing?: Easing;
     }
+
+    interface DoughnutModel {
```
amcasey: Was this intended to be a subtype of Model?  If not, it seems like the properties that are on `Model` but not `DoughnutModel` will be unavailable on `ModelType`.

amcasey (changes requested):
A quick question - not blocking.

types/chart.js/index.d.ts
jordanweiler: Good catch! These would be major breaking changes as they are currently written, which I didn't realize at first.

I wasn't intending for `DoughnutModel` to be a subtype of `Model`. Each chart model seemed very different looking at the chart.js code. I was hoping these models would be very explicit on what properties are going to be defined to avoid a bunch of optional properties or base properties defined that would never have values. For example, I didn't want to just add `circumference?: number;` to the `Model` definition because when I get the data for the Doughnut chart I know it is always defined.

Any suggestions on how to proceed?

I was thinking that the model and view definitions would eventually turn into something like `_model: BarModel | BubbleModel | DoughnutModel | LineModel | PolarAreaModel | RadarModel;` long-term. That was my mindset when creating this change. I don't have the time to do that right now so I just focused on getting the DoughnutModel updated on this PR.

amcasey: How would a consumer know which of the XModel types to use?  Without that information, they'd only have access to the properties common to all of them.

jordanweiler: Consumers would need to use `as` to use assertion to one of the specific types of models. Otherwise, yes they would just get the common properties across all union types. That might not be a good way to do this because consumers might not know they need to do that even if the names of the models match the chart types.

Is there a better way to add these missing properties from the Doughnut chart without making everything optional? When I create a Doughnut chart, I know the chart's data will be in a specific format so I was hoping I could reflect that in these definitions somehow.

amcasey: One common way to handle this would be to have a `kind` property on the model that tells you which set of properties are available.  I take it the underlying library doesn't have anything like that.  If you were consuming `_model` in pure JS, how would you know which properties to look at?

jordanweiler: The chart.js library seems to define a completely different `_model` structure for each type of chart it can render. It seems like inherit knowledge of the type of chart being rendered is needed for the `_model` property to be of any use to consumers.

* [Bubble _model](https://github.com/chartjs/Chart.js/blob/master/src/controllers/controller.bubble.js#L140-L151)
* [Doughnut _model](https://github.com/chartjs/Chart.js/blob/master/src/controllers/controller.doughnut.js#L242-L257)
* [Line _model](https://github.com/chartjs/Chart.js/blob/master/src/controllers/controller.line.js#L122-L135)

No common `kind` (or the equivalent) property seems to exist in the JS form. If that did exist, doing the discriminated unions would be possible.

amcasey: Personally, I haven't seen a good solution to this problem.  Options including loading `Model` with an optional declaration of every possible property, or cobbling together something like 

```ts
interface Model1 {
    P1: string;
    P2: number;
    P3?: never;
}
interface Model2 {
    P1?:never;
    P2: number;
    P3: boolean;
}

declare const _model: Model1 | Model2;
```

`_model` will have `P1` and `P3` optional and `P2` required and the declaration accurately conveys that the caller should probably assert one of the individual model types.  The downside is that all the declarations have to stay in sync (there might be something clever you can do with index signatures, but I didn't find it while experimenting).

jordanweiler: I think I'll pursue the optional declaration for every possible property on `Model` to not break existing users and then possibly still export `DoughnutModel` for users that want to assert to that model type. I don't like the other option of adding a bunch of `never` properties for the `DoughnutModel`. That seems to unnecessarily complicate that model and keeping everything in sync seems difficult.

amcasey: Understandable.  Sorry I didn't have a better answer for you.

amcasey (approved):

types/chart.js/index.d.ts
jordanweiler: That's okay. Hopefully as these definitions progress, we can come up with a better solution for this scenario.

Thanks for all the help with this PR. I really appreciate the discussion.

PR #40556, opened 2019-11-21
Node: Missing features in TLS module

dex4er:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: see below

Added all missing features in TLS module, like:

https://nodejs.org/api/tls.html#tls_tls_default_min_version
https://nodejs.org/api/tls.html#tls_tls_default_max_version
https://nodejs.org/api/tls.html#tls_server_getticketkeys
https://nodejs.org/api/tls.html#tls_server_setticketkeys_keys
https://nodejs.org/api/tls.html#tls_server_setsecurecontext_options
https://nodejs.org/api/tls.html#tls_tlssocket_disablerenegotiation
https://nodejs.org/api/tls.html#tls_tlssocket_getcertificate
https://nodejs.org/api/tls.html#tls_tlssocket_getephemeralkeyinfo
https://nodejs.org/api/tls.html#tls_tlssocket_getfinished
https://nodejs.org/api/tls.html#tls_tlssocket_getpeerfinished
https://nodejs.org/api/tls.html#tls_tlssocket_getsharedsigalgs
https://nodejs.org/api/tls.html#tls_tlssocket_issessionreused

### DIFF
--- a/types/node/tls.d.ts
+++ b/types/node/tls.d.ts
@@ -225,6 +225,10 @@ declare module "tls" {
          */
         getTLSTicket(): Buffer | undefined;
         /**
+         * Returns true if the session was reused, false otherwise.
+         */
+        isSessionReused(): boolean;
+        /**
          * Initiate TLS renegotiation process.
          *
          * NOTE: Can be used to request peer's certificate after the secure connection has been established.
--- a/types/node/tls.d.ts
+++ b/types/node/tls.d.ts
@@ -79,6 +79,28 @@ declare module "tls" {
         size: number;
     }
 
+    interface KeyObject {
+        /**
+         * Private keys in PEM format.
+         */
+        pem: string | Buffer;
+        /**
+         * Optional passphrase.
+         */
+        passphrase?: string;
+    }
+
+    interface PxfObject {
+        /**
+         * PFX or PKCS12 encoded private key and certificate chain.
+         */
+        buf: string | Buffer;
+        /**
+         * Optional passphrase.
+         */
+        passphrase?: string;
+    }
+
     interface TLSSocketOptions extends SecureContextOptions, CommonConnectionOptions {
         /**
          * If true the TLS socket will be instantiated in server-mode.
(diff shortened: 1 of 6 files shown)


types/node/v10/tls.d.ts
```diff
@@ -156,6 +171,32 @@ declare module "tls" {
...
+         * Returns an object representing the type, name, and size of parameter
+         * of an ephemeral key exchange in Perfect Forward Secrecy on a client
+         * connection. It returns an empty object when the key exchange is not
+         * ephemeral. As this is only supported on a client socket; null is
+         * returned if called on a server socket. The supported types are 'DH'
+         * and 'ECDH'. The name property is available only when type is 'ECDH'.
+         *
+         * For example: { type: 'ECDH', name: 'prime256v1', size: 256 }.
+         */
+        getEphemeralKeyInfo(): EphemeralKeyInfo | {} | null;
```
SimonSchick: Why `{}`?

types/node/tls.d.ts
```diff
@@ -114,12 +151,50 @@ declare module "tls" {
...
+         * object has some properties corresponding to the fields of the
+         * certificate.
+         *
+         * See tls.TLSSocket.getPeerCertificate() for an example of the
+         * certificate structure.
+         *
+         * If there is no local certificate, an empty object will be returned.
+         * If the socket has been destroyed, null will be returned.
+         */
+        getCertificate(): PeerCertificate | {} | null;
```
SimonSchick: Same here, should probably be `object`.

types/node/tls.d.ts
```diff
@@ -285,7 +405,29 @@ declare module "tls" {
...
+         * used if the client request's SNI name matches the supplied hostname
+         * (or wildcard).
+         */
         addContext(hostName: string, credentials: SecureContextOptions): void;
+        /**
+         * Returns the session ticket keys.
+         */
+        getTicketKeys(): Buffer;
+        /**
+         *
```
SimonSchick: Nit: Remove this line

SimonSchick (changes requested):
You can skip updating v11, it's EOL.

types/node/v10/tls.d.ts
dex4er: Object can be empty. Function can return ie. `{ type: 'ECDH', name: 'prime256v1', size: 256 }` or `{}` or `null`. I think it is totally different than returning `{ type?: string, name?: string, size?: number }`.

Do you have any alternate syntax for "empty object" type?

dex4er: I think it works:

```ts
const keyInfo: EphemeralKeyInfo | {} | null = tlsSocket.getEphemeralKeyInfo();
if (!keyInfo) {
    console.log(keyInfo, "is null")
} else if (!("type" in keyInfo)) {
    console.log(keyInfo, "is {}")
} else {
    console.log(keyInfo.name);
    console.log(keyInfo.type);
    console.log(keyInfo.size);
}
```

but I can replace EphemeralKeyInfo to `{ type?: string, name?: string, size?: number }` if you think that it will simplify this case.

types/node/tls.d.ts
dex4er: Ok, it looks like it should be `object` instead `{}` really. Then I'll get

```
Property 'name' does not exist on type 'object | EphemeralKeyInfo'.
  Property 'name' does not exist on type 'object'.ts(2339)
```

without proper type guard.

types/node/tls.d.ts
dex4er: What do you mean? https://nodejs.org/dist/latest-v12.x/docs/api/tls.html#tls_server_getticketkeys says this is real method and my test proves that it returns a `Buffer` indeed.

dex4er:
@SimonSchick ping?

galkin (approved):
I checked this pull and documentation. The code is look good for me.
I don't have any opinion regarding storing documentation at comments. It has proc and cons, so I will 
escalate this question as separate issue.

dex4er:
@galkin A lot of methods and properties had jsdoc already. If the decision will be that documentation is unnecessary I think it all might be removed in next PR.

galkin:
I thought creating a separate issue might be premature. Let's invite @sandersn to this issue.

This pull request is good example of the issue with comments at `@types/node`. The comments are copy-pasted documentation. This approach has such proc and cons:

- 👍 IDE provides one-click documentation
- 👎 Changes at documentation will require changes at `@types/node`. We know, that https://github.com/nodejs/node/edit/master/doc/api/index.md  is a single source of truth.
- ❓Maybe, similar to `@deprecated`/ `@experimental` we should have `@see` with copy-pasted block? Or instead?

Any approach is a trade-off, but we should have consistency at least at the package level.

I think it is nice to have such a decision for the whole `@types` namespace as part of [FAQ](https://github.com/DefinitelyTyped/DefinitelyTyped#faq). Could you provide your feedback? Or should I create a separate issue for this discussion?

dex4er:
I had the same dilemma for `@types/nodemailer` and I put documentation there as far as VS Code uses it and shows it in tooltips. Of course I can be aligned with some global recommendation from FAQ and ie. remove all jsdocs from my other types.

I don't think that we need to change jsdoc in `@types/node` every time when documentation will be changed on http://nodejs.org/ Probably is important only when we need to modify typings for single module because new feature is added or old is removed.

SimonSchick (approved):

sandersn:
I'd prefer to have jsdoc whenever it's available. Especially for popular packages, there are enough human authors that I don't think it'll be a problem to keep it up to date. And for less popular packages, documentation is less likely to update as often anyway.

For the most popular packages, it might be a good idea to make the task easier with a script that copies over the comments. Or, even better, to have a regular update script that opens a PR on DT.

SimonSchick:
@sandersn I usually try to copy (and crop redundant type info) documentation from the nodejs docs.
From personal experience I'm going to say automating this is going to be stupidly complicated (naterual language processing is annoying).

dex4er:
@galkin So what is a final conclusion? Do we wait with merge or it is already accepted?

galkin:
@dex4er, I think we have a green light for merging.

sandersn:
I'm going to merge this, but you may need to watch DT issues for failures since master may have drifted some distance since the beginning of this branch.

Solonotix:
I just discovered this today, but was it intentional to represent the PKCS#12 interface as `PxfObject`? According to what I could find on the history of naming conventions, `*.p12` was a Netscape convention, and `*.pfx` was the Microsoft convention that merged into the single standard, and [PFX stands for Personal Information Exchange](https://docs.microsoft.com/en-us/windows-hardware/drivers/install/personal-information-exchange---pfx--files).

I'm specifically calling attention to the spelling error, as to whether it was intentional or not, and if maybe it should be addressed in a patch.

PR #40558, opened 2019-11-21
Add missing function `getPublicKeyFingerprint` in namespace `pki` in `@types/node-forge`

piotar:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/digitalbazaar/forge/blob/master/lib/x509.js#L821 and https://www.npmjs.com/package/node-forge#x509

### DIFF
--- a/types/node-forge/README.md
+++ b/types/node-forge/README.md
@@ -28,3 +28,4 @@ BSD-2-Clause
 [Rafal2228](https://github.com/rafal2228)
 [Beeno Tung](https://github.com/beenotung)
 [Nikita Koryabkin](https://github.com/Apologiz)
+[Piotr Tarasiuk](https://github.com/piotar)
--- a/types/node-forge/index.d.ts
+++ b/types/node-forge/index.d.ts
@@ -261,6 +261,25 @@ declare module "node-forge" {
             getAttribute(opts: string | GetAttributeOpts): Attribute | null;
         }
 
+        interface FingerprintOptions {
+            /**
+             * @description the delimiter to use between bytes for `hex` encoded output
+             */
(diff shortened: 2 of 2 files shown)


joeflateau (changes requested):
At a minimum, this PR needs tests added. But it would also be better if the function signature was overloaded depending on the encoding option since its return type is determined by that option...

something like this (not tested so probably doesn't work as-is): 

```

interface FingerprintOptions {
    /**
        * @description the delimiter to use between bytes for `hex` encoded output
        */
    delimiter?: string;
    /**
        * @description if not specified, the function will return `ByteStringBuffer`
        */
    encoding?: 'hex' | 'binary';
    /**
        * @description if not specified, the default publick key type is `RSAPublicKey`
        */
    type?: 'RSAPublicKey' | 'SubjectPublicKeyInfo';
    /**
        * @description if not specified defaults to `md.sha1`
        */
    md?: md.MessageDigest;
}

interface HexFingerprintOptions extends FingerprintOptions {
    encoding: 'hex';
}

interface StringFingerprintOptions extends FingerprintOptions {
    encoding: undefined;
}

interface ByteStringFingerprintOptions extends FingerprintOptions {
    encoding: 'binary';
}

function getPublicKeyFingerprint(publicKey: pki.PublicKey, options?: HexFingerprintOptions): Hex;
function getPublicKeyFingerprint(publicKey: pki.PublicKey, options?: StringFingerprintOptions): string;
function getPublicKeyFingerprint(publicKey: pki.PublicKey, options?: ByteStringBufferFingerprintOptions): util.ByteStringBuffer;
```

PR #40560, opened 2019-11-21
NodeJS child_process signals to be Signals

qwelias:
Please fill in this template.

### DIFF
--- a/types/node/child_process.d.ts
+++ b/types/node/child_process.d.ts
@@ -18,7 +18,7 @@ declare module "child_process" {
         readonly killed: boolean;
         readonly pid: number;
         readonly connected: boolean;
-        kill(signal?: string): void;
+        kill(signal?: NodeJS.Signals): void;
         send(message: any, callback?: (error: Error | null) => void): boolean;
         send(message: any, sendHandle?: net.Socket | net.Server, callback?: (error: Error | null) => void): boolean;
         send(message: any, sendHandle?: net.Socket | net.Server, options?: MessageOptions, callback?: (error: Error | null) => void): boolean;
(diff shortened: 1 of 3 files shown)


types/node/child_process.d.ts
```diff
@@ -18,7 +18,7 @@ declare module "child_process" {
         readonly killed: boolean;
         readonly pid: number;
         readonly connected: boolean;
-        kill(signal?: string): void;
+        kill(signal?: NodeJS.Signals): void;
```
Flarna: Accodring to docs this can be a string from `NodeJS.Signals` or a number.

types/node/child_process.d.ts
```diff
@@ -259,7 +259,7 @@ declare module "child_process" {
         cmd?: string;
         killed?: boolean;
         code?: number;
-        signal?: string;
+        signal?: NodeJS.Signals;
```
Flarna: I think `number` is also ok here.

types/node/child_process.d.ts
```diff
@@ -300,7 +300,7 @@ declare module "child_process" {
 
     interface ExecFileOptions extends CommonOptions {
         maxBuffer?: number;
-        killSignal?: string;
+        killSignal?: NodeJS.Signals;
```
Flarna: and once agen I think `number` is also ok here.

types/node/child_process.d.ts
qwelias: @Flarna docs does not eplicitly state that.

Flarna: I found it there: https://nodejs.org/dist/latest-v12.x/docs/api/child_process.html#child_process_subprocess_kill_signal `signal <number> | <string>` . Seems it's allowed everywhere the user passes something in but in events you always get one of the `NodeJS.Signals` strings.

Flarna (approved):

r3nya (approved):

PR #40561, opened 2019-11-21
[jws] Allow JOSE header parameter types other than string, as defined in RFC

grzegorzjudas:
- According to [RFC7515 4.1.11](https://tools.ietf.org/html/rfc7515#section-4.1.11) JOSE headers in JWS can be an array of strings (for `crit` property)
- The same point, as well as [RFC7519 4.1.4](https://tools.ietf.org/html/rfc7519#section-4.1.4) shows that using a number as header value for `exp` is possible
- To allow more dynamic structures and better code, allowing undefined, as it does not result in the resulting signature and/or JOSE header (after decoding back to JSON) to be different. Consider this:
```ts
type JsonWebSignatureHeader = {
    alg: JsonWebSignatureAlgorithm;
    jku?: string;
};

const header: JsonWebSignatureHeader = {
    alg: 'RS256'
};

createSign({
    header
}).on('done', (signature: string) => {
    console.log(signature);
});
```
this would result in an error `Type 'string | undefined' is not assignable to type 'string'.` because `jku` is optional property of `JsonWebSignatureHeader`, while still being correct header object for JWS.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/brianloveswords/node-jws
- [x] ~If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.~ Fixes existing version types
- [x] ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.~ no substantial changes

### DIFF
--- a/types/jws/index.d.ts
+++ b/types/jws/index.d.ts
@@ -150,5 +150,5 @@ export type Algorithm = 'HS256' | 'HS384' | 'HS512' | 'RS256' |
 
 export interface Header {
     alg: Algorithm;
-    [name: string]: string | string[] | number | undefined;
+    [name: string]: string | ReadonlyArray<string> | number | undefined;
 }
--- a/types/jws/index.d.ts
+++ b/types/jws/index.d.ts
@@ -150,5 +150,5 @@ export type Algorithm = 'HS256' | 'HS384' | 'HS512' | 'RS256' |
 
 export interface Header {
     alg: Algorithm;
-    [name: string]: string;
+    [name: string]: string | string[] | number | undefined;
 }
--- a/types/jws/index.d.ts
+++ b/types/jws/index.d.ts
@@ -172,5 +172,5 @@ export interface CertificateProperties extends PrivateProperties {
 }
 
 export interface PrivateProperties {
-    [name: string]: any
+    [name: string]: any;
 }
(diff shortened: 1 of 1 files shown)


JustinBeckwith (approved):

grzegorzjudas:
@JustinBeckwith I've specified the header properties in more depth according to that RFC. Was forced to do that when I wanted to use "jwk" property that turned out to be an object (see: [RFC7517 4.1.3](https://tools.ietf.org/html/rfc7515#section-4.1.3) that describes it as having multiple properties itself).

That, plus allowing custom (private) parameters, as [RFC7515 4.3](https://tools.ietf.org/html/rfc7515#section-4.3) describes.

types/jws/index.d.ts
```diff
@@ -148,7 +148,27 @@ export type Algorithm = 'HS256' | 'HS384' | 'HS512' | 'RS256' |
...
 
-export interface Header {
+export type Header = {
     alg: Algorithm;
-    [name: string]: string;
+    jwk?: JWK;
+    typ?: string;
+    cty?: string;
+    crit?: ReadonlyArray<string>;
+} & CertificateProperties & {
```
sheetalkamat: Instead of using intersection type here, i would recommend use `export interface Header extends CertificateProperties` instead

sheetalkamat (changes requested):

PR #40564, opened 2019-11-21
fast-text-encoding: Add v1 types

inglec-arista:
fast-text-encoding updates browser globals, hence no types are exported and no tests are provided.


sheetalkamat (changes requested):
This needs the definition of TextEncoder and TextDecoder  as globals.. Also add test.. 
Refer to https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html

PR #40567, opened 2019-11-21
Stripe: Add Type for FileLinks

Aseelaldallal:
Add Type for FileLinks
https://stripe.com/docs/api/file_links

If adding a new definition:

### DIFF
--- a/types/stripe/index.d.ts
+++ b/types/stripe/index.d.ts
@@ -4035,7 +4035,7 @@ declare namespace Stripe {
              * Only return links for the given file.
              */
             file?: string;
-            
+
             /**
              * Filter links by their expiration status. By default, all links are returned.
              */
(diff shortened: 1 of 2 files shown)


types/stripe/index.d.ts
```diff
@@ -3969,6 +3971,69 @@ declare namespace Stripe {
             | 'incorporation_document';
     }
 
+    namespace fileLinks {
+        interface IFileLink extends IResourceObject {
+            /**
+             * Value is 'file_link'
+             */
+            object: 'file_link';
+            /**
```
pushplay: Can you separate all the comments and properties with a newline like so?

```
            /**
             * Value is 'file_link'
             */
            object: 'file_link';

            /**
             * Time at which the object was created. Measured in seconds since the Unix epoch.
             */
            created: number;
```

pushplay (changes requested):
Looks really good, thanks for the tests.  Just 1 issue on formatting.

types/stripe/index.d.ts
Aseelaldallal: done :) @pushplay 
thanks!

pushplay (approved):
Awesome.

PR #40571, opened 2019-11-21
rename interface Paramater to Parameter

ziyaddin:


amcasey (approved):
Note that renaming an exported type will break anyone consuming the type.  It's nice to fold such changes into version bumps, but I can see why it would be irritating enough to fix now.

PR #40577, opened 2019-11-22
Add types for zfont

### LINKED ISSUE #6: TypeScript Support
[@types/zdog](https://www.npmjs.com/package/@types/zdog) already exists. Let's make definitions (@types/zfont) for this package?

demensky:
Fixes jaames/zfont#6

### DIFF
--- a/types/zfont/index.d.ts
+++ b/types/zfont/index.d.ts
@@ -1,0 +1,226 @@
+// Type definitions for zfont 1.2
+// Project: https://github.com/jaames/zfont
+// Definitions by: Dmitry Demensky <https://github.com/demensky>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.4
+
+import * as zdog from 'zdog';
+
+export as namespace Zfont;
+
+export function init(x: typeof zdog): typeof zdog;
+
+declare module 'zdog' {
+    /** @see {@link https://github.com/jaames/zfont#multiline-text Zfont API} */
+    type MultilineText = string | ReadonlyArray<string>;
+
+    /**
+     * Horizontal text alignment (equivalent to the CSS `text-align` property).
+     * @see {@link https://github.com/jaames/zfont#textalign Zfont API} (Text)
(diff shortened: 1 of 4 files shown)


types/zfont/index.d.ts
```diff
@@ -0,0 +1,226 @@
...
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.4
+
+import * as zdog from 'zdog';
+
+export as namespace Zfont;
+
+export function init(x: typeof zdog): typeof zdog;
+
+declare module 'zdog' {
```
sheetalkamat: Instead of declaring this as ambient module here please add zdog as another package and move definitions there. 
Thank you.

sheetalkamat (changes requested):

types/zfont/index.d.ts
sheetalkamat: If `zdog` package already present you just need to reference it instead of redefining it

demensky: @sheetalkamat. `zdog` is a separate library (I also wrote the definitions for it).
 I also do not understand why the author (@jaames) simply did not export entities ([usage example](https://github.com/jaames/zfont#hello-world)).

@sheetalkamat please provide an example of how best to do this.

sheetalkamat: please take a look at https://github.com/DefinitelyTyped/DefinitelyTyped#im-writing-a-definition-that-depends-on-another-definition-should-i-use-reference-types--or-an-import 
If something is missing or wrong in https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/zdog/index.d.ts you can fix it instead and please remove the definition of zdog from here.

demensky: @sheetalkamat [`zdog`](https://www.npmjs.com/package/zdog) & [`zfont`](https://www.npmjs.com/package/zfont) are different packages. `@types/zdog` is implemented correctly.

`zfont` patch the `zdog` API.

sheetalkamat: Got it .. I didn't realise that it was augmentation instead of definition. Thanks

PR #40582, opened 2019-11-22
[slonik] Update types to v21.4

bram-l:
Please fill in this template.

https://github.com/gajus/slonik/compare/e7126935ec8642c4f01a9aead0498a16dfa99151...master#diff-3a699a0719518f967e9b2df3f5549e02

### DIFF
--- a/types/slonik/index.d.ts
+++ b/types/slonik/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for slonik 19.0
+// Type definitions for slonik 21.4
 // Project: https://github.com/gajus/slonik#readme
 // Definitions by: Sebastian Sebald <https://github.com/sebald>
 //                 Misha Kaletsky <https://github.com/mmkal>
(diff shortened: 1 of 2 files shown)


types/slonik/index.d.ts
```diff
@@ -265,11 +267,18 @@ export interface QueryContextType {
...
     queryId: QueryIdType;
-    stackTrace: CallSiteType[] | null;
+
     /**
      * `process.hrtime.bigint()` for when query was received.
      */
     queryInputTime: number;
+
+    /** Object used by interceptors to assign interceptor-specific, query-specific context. */
+    // tslint:disable-next-line ban-types
```
mmkal: Why not use `object`?

types/slonik/index.d.ts
```diff
@@ -367,28 +376,37 @@ export interface InterceptorType {
...
+    beforeTransformQuery?: (
+        queryContext: QueryContextType,
+        query: QueryType
+    ) => MaybePromiseType<null>;
     queryExecutionError?: (
         queryContext: QueryContextType,
         query: QueryType,
         error: SlonikError
-    ) => MaybePromiseType<void>;
+    ) => MaybePromiseType<null>;
```
mmkal: Is this a change from v21? Could you link to the docs?

mmkal (commented):
Mostly looks good, but I haven't used v21 yet so maybe @sebald or @gajus could weigh in?

types/slonik/index.d.ts
gajus: https://github.com/gajus/slonik/blob/1ec5326171fa79135db93f92d35c5f8c628eb7fd/src/types.js#L468-L472

That is correct. It was one of the breaking changes.

types/slonik/index.d.ts
gajus: I am not familiar with TS enough to comment, but pretty sure it should be `any`.

It is basically an object that interceptors use to assign any arbitrary properties associated with the query.

bram-l: I've updated the `sandbox` type with `Record<string, any>`, which provides slightly better type safety compared to `any`

mmkal (approved):

PR #40591, opened 2019-11-22
feat: 🎸 add types for express-healthcheck

o8e:
## [express-healthcheck](https://github.com/lennym/express-healthcheck)

- [x]  Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

### New definition

### DIFF
--- a/types/express-healthcheck/index.d.ts
+++ b/types/express-healthcheck/index.d.ts
@@ -1,0 +1,13 @@
+// Type definitions for express-healthcheck 0.1
+// Project: https://github.com/lennym/express-healthcheck/
+// Definitions by: Ollie Tribe <https://github.com/o8e>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import { RequestHandlerParams } from 'express-serve-static-core';
+
+export interface Options {
+    test?: () => any;
+    healthy?: () => any;
+}
+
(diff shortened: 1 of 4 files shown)


types/express-healthcheck/index.d.ts
```diff
@@ -0,0 +1,13 @@
...
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import { RequestHandlerParams } from 'express-serve-static-core';
+
+export interface Options {
+    test?: () => any;
+    healthy?: () => any;
+}
+
+export default function fn(options?: Options): RequestHandlerParams;
```
sheetalkamat: This needs `export =` instead of `export default`. Refer to https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default

sheetalkamat (changes requested):

types/express-healthcheck/index.d.ts
o8e: To be honest, I couldn't get it to work with `export =` as tslint didn't like it, and end up getting stuck in a linting loop.

I'm a novice at this, so if you want to take the time to export them properly I would be very grateful. Otherwise I'm probably not going to invest more time in fighting the linter when I have no clue.

PR #40593, opened 2019-11-22
[plotly.js] Added polar layout attributes

mtadams007:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://plot.ly/javascript/reference/#layout-polar>>

### DIFF
--- a/types/plotly.js/index.d.ts
+++ b/types/plotly.js/index.d.ts
@@ -209,7 +209,7 @@ export interface LayoutPolar {
 		categoryarray: string[];
 		angle: number;
 		side: 'clockwise' | 'counterclockwise';
-		title: {
+		title: Partial<{
 			text: string;
 			font: Partial<Font>;
 			hoverformat: string;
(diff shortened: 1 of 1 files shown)


types/plotly.js/index.d.ts
```diff
@@ -282,6 +282,12 @@ export interface Layout {
 	bargroupgap: number;
 	selectdirection: 'h' | 'v' | 'd' | 'any';
 	hiddenlabels: string[];
+	polar: {
```
szechyjs: based on the docs there are many more properties than `bgcolor` and `radialaxis.color`. Could you make this feature complete? Something like:

```ts
polar: Partial<LayoutPolar>;
```

https://plot.ly/javascript/reference/#layout-polar

Also, as is, both of those properties would be required when polar is defined, as the `?` is missing on each.

szechyjs (changes requested):

types/plotly.js/index.d.ts
mtadams007: Sure thing. Will get on it.

types/plotly.js/index.d.ts
```diff
@@ -186,6 +186,80 @@ export interface DownloadImgopts {
...
+	domain: {
+		x: number[];
+		y: number[];
+		row: number;
+		column: number;
+	};
+	sector: number[];
+	hole: number;
+	bgcolor: Color;
+	radialaxis: {
```
szechyjs: all of these nested properties should be partials...

```ts
radialaxis: Partial<{
  visible: boolean;
  range: any[];
}>;

mtadams007: I made the radialaxis into a partial. I'm not super familiar with using Partials - should I make all the nested features within it also a partial (e.g. `title`)?

mtadams007:
Closing in favor of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42832

PR #40610, opened 2019-11-23
Fix circular dependency bug in @types/validator

bondz:
With these two lines, typescript errors with

`Circular definition of import alias 'Validator'`

Without them, the library works just fine.

Please fill in this template.

### DIFF
--- a/types/validator/index.d.ts
+++ b/types/validator/index.d.ts
@@ -1243,7 +1243,3 @@ export function whitelist(input: string, chars: string): string;
  * Converts to string.
  */
 export function toString(input: any): string;
-
-import * as Validator from './';
-
-export default Validator;


chidiogo-akunne (approved):

louy:
I don’t think this change matches the actual export signature of the validator package. In the source code, validator is exported as `default` not as `module.exports`

louy:
Take a look at #40491

bondz:
Oh, right @louy I think your PR is more comprehensive and fixes the actual issue. I'm unable to use typings for sequelize because of the broken typings for validator. Currently, I get

image

I'll be closing this in favour of yours but it seems to be blocked.

louy:
I’ll try to get it released ASAP. My point was that even though this change might solve the types issues I think it’ll lead to a broken implementation since the types don’t match the underlying module, but I could be wrong

bondz:
I agree, although this fixes the bug. Doesn't change the fact that the typings are wrong in the first place and your PR covers that. I hope it gets released soon though.

mehulrkaj:
Hi, I'm having the same issue. I'm not able to resolve it.
I tried upgrading the version of @types/sequelize, the problem still persist. How can I solve this?

Screenshot 2019-11-26 at 11 32 07 AM

bondz:
@mehulrkaj You can temporarily fix this by putting a field in your `package.json` if you're using yarn.

```json
"resolutions": {
    "@types/validator": "10.11.3",
  }
```

You can remove that when #40491 is published.

bondz:
@mehulrkaj I was mistaken, this is fixed. Use `"@types/sequelize": "4.28.8"`

mehulrkaj:
@bondz The issue still exists, after changing `"@types/sequelize": "4.28.8"` . Please find the attachment below.
Screenshot 2019-11-26 at 2 19 32 PM

PR #40619, opened 2019-11-24
Async | Added retry function returning a promise

raythurnvoid:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://caolan.github.io/async/v3/docs.html#retry>>

### DIFF
--- a/types/async/index.d.ts
+++ b/types/async/index.d.ts
@@ -221,6 +221,14 @@ export function retry<T, E = Error>(
     task: (callback: AsyncResultCallback<T, E>, results: any) => void,
     callback: AsyncResultCallback<any, E>
     ): void;
+export function retry<T, E = Error>(
+    opts: number | {
+        times: number,
+        interval: number | ((retryCount: number) => number),
+        errorFilter?: (error: Error) => boolean
+    },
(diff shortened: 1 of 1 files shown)


types/async/index.d.ts
```diff
@@ -221,6 +221,14 @@ export function retry<T, E = Error>(
     task: (callback: AsyncResultCallback<T, E>, results: any) => void,
     callback: AsyncResultCallback<any, E>
     ): void;
+export function retry<T, E = Error>(
+    opts: number | {
+        times: number,
```
rbuckton: According to the example in the linked documentation, `times` and `interval` are both optional:

```js
// try calling apiMethod only when error condition satisfies, all other
// errors will abort the retry control flow and return to final callback
async.retry({
  errorFilter: function(err) {
    return err.message === 'Temporary error'; // only retry on a specific error
  }
}, apiMethod, function(err, result) {
    // do something with the result
});
```

Also, the first argument itself is entirely optional:

```js
// try calling apiMethod the default 5 times no delay between each retry
async.retry(apiMethod, function(err, result) {
    // do something with the result
});
```

rbuckton (changes requested):

PR #40624, opened 2019-11-25
Titanium updates

ffMathy:
made fireEvent event arg optional.


amcasey (commented):
There appear to be identical signatures on a bunch of other types.  Is this the only one for which the "event" parameter is optional?

ffMathy:
Not sure. I'm just improving a few things at a time as I go along.

amcasey:
Okay.  Let's give @appcelerator and @janvennemann some time to respond.

janvennemann:
@ffMathy the Titanium typings will be auto generated. Is there a matching PR on our side to reflect this change?

ffMathy:
Nope, I didn't know these were auto-generated.

janvennemann:
The typings are auto generated from our [apidoc](https://github.com/appcelerator/titanium_mobile/tree/master/apidoc/Titanium) folder.

For this change, for example, you need to modify [Proxy.fireEvent](https://github.com/appcelerator/titanium_mobile/blob/40bd1d92749857a37ebea2fa87c2f064d1775a51/apidoc/Titanium/Proxy.yml#L60) and make the parameter optional by adding `optional: true`. Then you can update typings with `npm run build:docs -- --format typescript --output .`, which will create the updated `index.d.ts` in the root dir of `titanium_mobile`.

amcasey (changes requested):
Sounds like a change to the upstream project is required first?

ffMathy:
I made a PR for it here: https://github.com/appcelerator/titanium_mobile/pull/11371

janvennemann (approved):
Upstream PR was merged so this is good to go!

amcasey (approved):

PR #40630, opened 2019-11-25
fast-text-encoding: Fix global name collisions

inglec-arista:
Global var definitions for TextDecoder and TextEncoder collide with the
globals defined by "lib.dom.d.ts". Ideally we would be able to
conditionally define TextDecoder and TextEncoder if they are not
already defined in the current context. Since we cannot do conditional
definitions in an ambient context, we export these types so the user
can do this themselves (see tests).

- Move global type definitions to namespace to prevent collisions with
  "lib.dom.d.ts" TextDecoder and TextEncoder.
- Update name of namespace from "fastTextEncoder" to "fastTextEncoding"
  to match package name.
- Update tests to manually define global vars TextDecoder and
  TextEncoder.

### DIFF
--- a/types/fast-text-encoding/fast-text-encoding-tests.ts
+++ b/types/fast-text-encoding/fast-text-encoding-tests.ts
@@ -1,4 +1,6 @@
-import 'fast-text-encoding';
+// Declare globals in non-DOM environment
+declare let TextDecoder: fastTextEncoding.TextDecoder;
+declare let TextEncoder: fastTextEncoding.TextEncoder;
 
 const encoder1 = new TextEncoder();
 const _encoder2 = new TextEncoder('utf-8');
--- a/types/fast-text-encoding/index.d.ts
+++ b/types/fast-text-encoding/index.d.ts
@@ -1,28 +1,66 @@
 // Type definitions for fast-text-encoding 1.0
-// Project: https://github.com/samthor/fast-text-encoder#readme
+// Project: https://github.com/samthor/fast-text-encoding#readme
 // Definitions by: Ciarán Ingle <https://github.com/inglec-arista>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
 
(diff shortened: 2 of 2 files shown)


types/fast-text-encoding/index.d.ts
```diff
@@ -2,27 +2,65 @@
 // Project: https://github.com/samthor/fast-text-encoder#readme
```
amcasey: Typo: "encoding"

types/fast-text-encoding/fast-text-encoding-tests.ts
```diff
@@ -1,5 +1,9 @@
 import 'fast-text-encoding';
```
amcasey: This doesn't appear to have any effect.

amcasey (changes requested):
It looks like the package name might have changed at some point and the old name might still be used in a couple places.

Requesting changes for bookkeeping purposes, but I'll merge if you want to go ahead.

types/fast-text-encoding/fast-text-encoding-tests.ts
inglec-arista: Ah, I didn't realise that you access the namespace without importing, thanks!

types/fast-text-encoding/index.d.ts
inglec-arista: Fixed

types/fast-text-encoding/index.d.ts
```diff
@@ -2,27 +2,65 @@
...
 // Definitions by: Ciarán Ingle <https://github.com/inglec-arista>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.2
 
-declare namespace fastTextEncoder {
-    interface Options {
+/**
+ * fast-text-encoder does not export any members, but defines the global classes
```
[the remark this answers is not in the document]
inglec-arista: Fixed

amcasey (approved):
Thanks!

PR #40631, opened 2019-11-25
[@types/google.visualization] adding missing tooltip action methods

tim-kuteev:
adding missing tooltip action methods for charts / removing nonexistent methods

### DIFF
--- a/types/google.visualization/index.d.ts
+++ b/types/google.visualization/index.d.ts
@@ -327,7 +327,7 @@ declare namespace google {
         //#region GeoChart
 
         // https://developers.google.com/chart/interactive/docs/gallery/geochart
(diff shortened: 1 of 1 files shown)


dckorben:
I don't see anything of particular issue. Better fidelity on properties/methods for each chart type actually supports is good.

rbuckton:
I'm a bit concerned about the definitions for base classes that don't actually exist at runtime (such as `google.visualization.ChartBaseRenderable`). However, since there are already other definitions in the declaration that don't actually exist (such as `google.visualization.ChartBaseRenderable`), and the rest of the PR looks fine to me, I don't see any other reason not to merge this.

PR #40649, opened 2019-11-26
[@types/dc] updated dc type definitions to latest version

privateOmega:
Please fill in this template.

### DIFF
--- a/types/dc/index.d.ts
+++ b/types/dc/index.d.ts
@@ -78,10 +78,10 @@ declare namespace dc {
     }
 
     export interface Filters {
-        RangedFilter(low: number, high: number): Array<number>;
-        TwoDimensionalFilter(filter: Array<number>): Array<number>;
-        RangedTwoDimensionalFilter(arr: Array<Array<number>>): Array<Array<number>>;
-        HierarchyFilter(path: string): Array<string>;
+        RangedFilter(low: number, high: number): Filter;
+        TwoDimensionalFilter(filter: Array<number>): Filter;
+        RangedTwoDimensionalFilter(arr: Array<Array<number>>): Filter;
+        HierarchyFilter(path: string):  Filter;
     }
 
     export interface Logger {
--- a/types/dc/index.d.ts
+++ b/types/dc/index.d.ts
@@ -3,6 +3,7 @@
 // Definitions by: hans windhoff <https://github.com/hansrwindhoff>
 //                 matt traynham <https://github.com/mtraynham>
 //                 matthias jobst <https://github.com/MatthiasJobst>
+//                 kiran mathew mohan <https://github.com/privateOmega>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 import * as d3 from "d3";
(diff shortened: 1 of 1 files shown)


privateOmega:
@gordonwoodhull PTAL when you have bandwidth and I would really appreciate it if you could provide change suggestions on where you deem necessary.

types/dc/index.d.ts
```diff
@@ -63,34 +64,42 @@ declare namespace dc {
...
 
     export interface Filter {
         isFiltered(value: any): boolean;
     }
 
     export interface Filters {
-        RangedFilter(low: any, high: any): Filter;
-        TwoDimensionalFilter(arr: Array<any>): Filter;
-        RangedTwoDimensionalFilter(arr: Array<any>): Filter;
+        RangedFilter(low: number, high: number): Array<number>;
```
hansrwindhoff: this seems incorrect, the return of these functions is an Array that has a isFiltered method, see the implementation:

the jsdoc annnotation in the code is incomplete...

/**
 * RangedFilter is a filter which accepts keys between `low` and `high`.  It is used to implement X
 * axis brushing for the {@link dc.coordinateGridMixin coordinate grid charts}.
 *
 * Its `filterType` is 'RangedFilter'
 * @name RangedFilter
 * @memberof dc.filters
 * @param {Number} low
 * @param {Number} high
 * @returns {Array<Number>}
 * @constructor
 */
dc.filters.RangedFilter = function (low, high) {
    var range = new Array(low, high);
    range.isFiltered = function (value) {
        return value >= this[0] && value < this[1];
    };
    range.filterType = 'RangedFilter';

    return range;
};

types/dc/index.d.ts
```diff
@@ -63,34 +64,42 @@ declare namespace dc {
     }
 
     export interface Events {
-        trigger(fn: () => void, delay?: number): void;
+        trigger(closure: () => void, delay?: number): void;
```
hansrwindhoff: why the rename?

hansrwindhoff (changes requested):

types/dc/index.d.ts
privateOmega: @hansrwindhoff True, I just went off modifying the type definitions just based on jsdoc annotation and completely ignored the actual code. PTAL after my update.

gordonwoodhull:
Hi @privateOmega. Thanks for putting in the effort here. The typings for dc.js have been severely out of date for a long time and I’ve tried to help a lot of people struggling with them.

Unfortunately I don’t currently use typescript and I think it would take a while for me to figure this out. There are also a number of edge cases like filters which I think may be difficult.

I definitely want to help, because this is important, but I’m not sure what to do.

types/dc/index.d.ts
privateOmega: In [Line 1231](https://github.com/dc-js/dc.js/blob/56df1dcd3f67201b750ef989aaa1c3e15632096d/dc.js#L1231),  apparently the function argument was renamed.

privateOmega:
@gordonwoodhull TBH this would be my first use of typescript per se, so I am not really acquainted with the best practices and all.
Also I updated the type definitions just based on jsdoc comments, there's a high chance some of it are wrong. :D

privateOmega:
@hansrwindhoff It would be great if you could review the changes and merge the PR if there's no more issues with it.

hansrwindhoff (approved):
Thanks for the work!

orta:
Sorry for the delay! Looks good to me 👍

PR #40658, opened 2019-11-26
@types/react: Mark booleanish HTML/SVG JSX attributes as such

fregante:
These attributes are "booleanish" as described by React’s own code: https://github.com/facebook/react/blob/a7d07ff24d8d3d925ab45f5eb9be42a62a345dad/packages/react-dom/src/shared/DOMProperty.js#L257-L269

---

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/react/blob/a7d07ff24d8d3d925ab45f5eb9be42a62a345dad/packages/react-dom/src/shared/DOMProperty.js#L257-L269

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -45,6 +45,7 @@ type NativePointerEvent = PointerEvent;
 type NativeTransitionEvent = TransitionEvent;
 type NativeUIEvent = UIEvent;
 type NativeWheelEvent = WheelEvent;
+type Booleanish = boolean | 'true' | 'false';
 
 /**
  * defined in scheduler/tracing
@@ -1656,16 +1657,16 @@ declare namespace React {
         // Standard HTML Attributes
         accessKey?: string;
         className?: string;
-        contentEditable?: boolean;
+        contentEditable?: Booleanish;
         contextMenu?: string;
         dir?: string;
-        draggable?: boolean;
+        draggable?: Booleanish;
         hidden?: boolean;
         id?: string;
         lang?: string;
         placeholder?: string;
         slot?: string;
-        spellCheck?: boolean;
+        spellCheck?: Booleanish;
         style?: CSSProperties;
         tabIndex?: number;
         title?: string;
@@ -2272,7 +2273,7 @@ declare namespace React {
         ascent?: number | string;
         attributeName?: string;
         attributeType?: string;
-        autoReverse?: number | string;
+        autoReverse?: Booleanish;
         azimuth?: number | string;
         baseFrequency?: number | string;
         baselineShift?: number | string;
@@ -2312,7 +2313,7 @@ declare namespace React {
         enableBackground?: number | string;
         end?: number | string;
         exponent?: number | string;
-        externalResourcesRequired?: number | string;
+        externalResourcesRequired?: Booleanish;
         fill?: string;
         fillOpacity?: number | string;
         fillRule?: "nonzero" | "evenodd" | "inherit";
@@ -2321,7 +2322,7 @@ declare namespace React {
         filterUnits?: number | string;
         floodColor?: number | string;
         floodOpacity?: number | string;
-        focusable?: number | string;
+        focusable?: Booleanish;
         fontFamily?: string;
         fontSize?: number | string;
         fontSizeAdjust?: number | string;
@@ -2399,7 +2400,7 @@ declare namespace React {
         pointsAtX?: number | string;
         pointsAtY?: number | string;
         pointsAtZ?: number | string;
-        preserveAlpha?: number | string;
+        preserveAlpha?: Booleanish;
         preserveAspectRatio?: string;
         primitiveUnits?: number | string;
         r?: number | string;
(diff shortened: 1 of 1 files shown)


eps1lon (commented):
Could you do me a favor an provide a small codesandbox showcasing what happens with `true`, `'true'` and `'tru'` (sic!)?

fregante:
React applies them as they are: https://codesandbox.io/s/vigilant-tdd-1jh5m

The HTML spec says that they are ignored and the attribute uses its defaults:
> `<input spellcheck=" false" id="b"></label>`
> The element with ID "b" in the following example would have checking enabled (**the leading space character in the attribute's value on the input element causes the attribute to be ignored**, so the ancestor's value is used instead, regardless of the default).
https://html.spec.whatwg.org/multipage/interaction.html#spelling-and-grammar-checking

eps1lon (approved):
Makes sense, thanks!

ferdaber (approved):

types/react/index.d.ts
```diff
@@ -2272,7 +2273,7 @@ declare namespace React {
         ascent?: number | string;
         attributeName?: string;
         attributeType?: string;
-        autoReverse?: number | string;
```
amcasey: Any idea if this is supposed to accept `0` or `1`?  I don't see anything in the code you linked, but I assume there was some reason for allowing `number` in the old type.

amcasey (changes requested):
Non-blocking question - just requesting changes for bookkeeping reasons.

types/react/index.d.ts
fregante: I changed this as well because the w3.org recommendation explicitly lists `"true" | "false"`

https://www.w3.org/TR/REC-smil/smil-timemanip.html#adef-autoReverse

types/react/index.d.ts
```diff
@@ -2312,7 +2313,7 @@ declare namespace React {
         enableBackground?: number | string;
         end?: number | string;
         exponent?: number | string;
-        externalResourcesRequired?: number | string;
+        externalResourcesRequired?: Booleanish;
```
fregante: https://www.w3.org/TR/SVG11/struct.html#ExternalResourcesRequiredAttribute

types/react/index.d.ts
```diff
@@ -2321,7 +2322,7 @@ declare namespace React {
         filterUnits?: number | string;
         floodColor?: number | string;
         floodOpacity?: number | string;
-        focusable?: number | string;
+        focusable?: Booleanish;
```
fregante: https://www.w3.org/TR/SVGTiny12/interact.html#focusable-attr

```suggestion
        focusable?: Booleanish | "auto";
```

types/react/index.d.ts
```diff
@@ -2399,7 +2400,7 @@ declare namespace React {
         pointsAtX?: number | string;
         pointsAtY?: number | string;
         pointsAtZ?: number | string;
-        preserveAlpha?: number | string;
+        preserveAlpha?: Booleanish;
```
fregante: https://www.w3.org/TR/SVG11/filters.html#feConvolveMatrixElementPreserveAlphaAttribute

types/react/index.d.ts
```diff
@@ -1656,16 +1657,16 @@ declare namespace React {
         // Standard HTML Attributes
         accessKey?: string;
         className?: string;
-        contentEditable?: boolean;
+        contentEditable?: Booleanish;
```
fregante: https://html.spec.whatwg.org/#dom-contenteditable

```suggestion
        contentEditable?: Booleanish | "inherit";
```

types/react/index.d.ts
```diff
@@ -1656,16 +1657,16 @@ declare namespace React {
...
         contextMenu?: string;
         dir?: string;
-        draggable?: boolean;
+        draggable?: Booleanish;
```
fregante: https://html.spec.whatwg.org/#the-draggable-attribute

types/react/index.d.ts
```diff
@@ -1656,16 +1657,16 @@ declare namespace React {
...
         hidden?: boolean;
         id?: string;
         lang?: string;
         placeholder?: string;
         slot?: string;
-        spellCheck?: boolean;
+        spellCheck?: Booleanish;
```
fregante: https://html.spec.whatwg.org/#dom-spellcheck

fregante (commented):
Here's all the related docs. Most of them seem right but a couple of them accept `"auto"` or `"inherit"`

Comment/react on those 2 to confirm the types

types/react/index.d.ts
amcasey: I like it, but I don't have permission to update your fork.  Please go ahead and add it to the PR.

types/react/index.d.ts
amcasey: I like it, but I don't have permission to update your fork.  Please go ahead and add it to the PR.

amcasey (approved):
LGTM.  Thanks for confirming!

amcasey:
@fregante do you want to make those changes you propose or shall I merge?

fregante:
All good

PR #18329, opened 2017-07-23
[apn] remove typings provided by package directly

BendingBender:
Fixes #11751.
One of the packages to remove from the list in #13364

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [ ] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If removing a declaration:
- [ ] If a package was never on DefinitelyTyped, you don't need to do anything. (If you wrote a package and provided types, you don't need to register it with us.)
- [x] Delete the package's directory.
- [x] Add it to `notNeededPackages.json`.

PR #40660, opened 2019-11-26
Import Boom type only.

Alkasai:
When I try to build my project locally with `@types/hapi__hapi` included as a dependency, I get the following error:
`Cannot use namespace 'Boom' as a type.` lines 514 & 4057

I'm also running TypeScript 3.7.2 on that project. I think this error is only visible on this version of TypeScript.

image

### DIFF
--- a/types/hapi__hapi/index.d.ts
+++ b/types/hapi__hapi/index.d.ts
@@ -20,7 +20,7 @@
 
 /// <reference types='node' />
 
-import * as Boom from '@hapi/boom';
+import { Boom } from '@hapi/boom';
 import * as http from 'http';
 import * as https from 'https';
 import * as Shot from '@hapi/shot';


saboya (approved):

SimonSchick:
This is problematic as this is only true for the types shipped by `@hapi/boom` specifically now.
We should probably remove the boom typings from DT and/or use the `@hapi/boom` ones.

saboya:
> This is problematic as this is only true for the types shipped by `@hapi/boom` specifically now.
> We should probably remove the boom typings from DT and/or use the `@hapi/boom` ones.

Oh, I thought the package typings were being used already. You're right. We should probably remove the typings from DT then.

Alkasai:
> > This is problematic as this is only true for the types shipped by `@hapi/boom` specifically now.
> > We should probably remove the boom typings from DT and/or use the `@hapi/boom` ones.
> 
> Oh, I thought the package typings were being used already. You're right. We should probably remove the typings from DT then.

How do you do that? Anything I can help with? I just saw this error when I upgraded to TypeScript 3.7.2. Not sure how to address that properly.

saboya:
> > > This is problematic as this is only true for the types shipped by `@hapi/boom` specifically now.
> > > We should probably remove the boom typings from DT and/or use the `@hapi/boom` ones.
> > 
> > 
> > Oh, I thought the package typings were being used already. You're right. We should probably remove the typings from DT then.
> 
> How do you do that? Anything I can help with? I just saw this error when I upgraded to TypeScript 3.7.2. Not sure how to address that properly.

It's as simple as deleting the files and adding the correct entry in the `notNeededPackages.json` file. Example: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/18329

PR #40669, opened 2019-11-26
react-dom tests: Add missing import of 'experimental'

sandersn:
This worked by mistake because experimental.d.ts is in tsconfig.json, but consumers of `@types/react-dom` will need `import 'react-dom/experimental'` so the test should reflect that.


sandersn:
Since this is just a test change and my upcoming rework of tsconfigs on DT depends on it, I'm going to merge it, but please do comment if it needs further changes. I'll be watching DT closely today since I'm making a major change. (See https://github.com/microsoft/types-publisher/issues/708 for details)

PR #40676, opened 2019-11-26
Trim tsconfig files; move untested files to OTHER_FILES.txt

sandersn:
This is the last major step of https://github.com/microsoft/types-publisher/issues/708. It does two things:

1. Trim the files list of all tsconfig.json down to test files and `index.d.ts`. Any other d.ts file needs to be referenced [transitively] either from `index.d.ts` or from a test file.
2. Add unreferenced d.ts files to OTHER_FILES.txt

Notably, it does *not* revise the instructions for what to put in tsconfig/OTHER_FILES. I'll do that in a followup PR.

In a few case, I had to manually add references. These are the only d.ts files that are changed; everything else is just tsconfig.json/OTHER_FILES.txt:

1. codemirror, plus a dependency react-codemirror
2. dwt, just v13
3. rangy
3. slickgrid
4. strophe.js
5. strophejs, plus a dependency strophejs-plugin-roster
6. waypoints

These packages do not use imports, so I added `<reference types=...` to the tests. Unfortunately, due to a bug in name resolution, `types` references don't respect baseUrl/typeRoots the way imports do, so I had to use a roundabout reference that only works with Typescript 3.2 and higher: 

```ts
/// <reference types="../rangy/rangy-classapplier" />
```

That is, I used a relative reference which included the package name. I changed types-publisher so that CI tests allow this kind of reference in tests.

Codemirror is the only widely used package in the list, and it already required Typescript 2.8, so I think the jump to 3.2 is not horrible.


ExE-Boss (approved):
I&nbsp;believe that&nbsp;this is&nbsp;a&nbsp;good&nbsp;idea.

sandersn:
Times out on Travis =(
Still running on Azure.

antoinerousseau:
Does @typescript-bot really have to ping the whole planet for a review? :/

types/algoliasearch/tsconfig.json
```diff
@@ -18,7 +18,6 @@
     },
     "files": [
         "index.d.ts",
-        "lite/index.d.ts",
```
Haroenv: What does this mean if a user imports `algoliasearch/lite` now? I think I'm a little lost :)

sandersn: Nothing changes for users &mdash; notice that algoliasearch wasn't one of the packages that published a new version. 

In fact, nothing really changes for algoliasearch since its tests already import and test algoliasearch/lite. `lite/index.d.ts` just isn't in tsconfig.json anymore, which means that it isn't automatically published. You either have to test it &mdash; what algoliasearch does &mdash; or add it to OTHER_FILES.txt to indicate that it should be published untested. Only about 100 packages do the latter.

Haroenv: Okay, that makes lots of sense, and makes it simpler to add "common" files which aren't exports. Thanks!

PR #40681, opened 2019-11-26
[js-cookie] Remove getJson method

nreynis:
The `getJson` method has been dropped from the library.

https://github.com/js-cookie/js-cookie/commit/4b79290b98d7fbf1ab493a7f9e1619418ac01e45

### DIFF
--- a/types/js-cookie/index.d.ts
+++ b/types/js-cookie/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for js-cookie 2.2
+// Type definitions for js-cookie 3.0
 // Project: https://github.com/js-cookie/js-cookie
 // Definitions by: Theodore Brown <https://github.com/theodorejb>
 //                 BendingBender <https://github.com/BendingBender>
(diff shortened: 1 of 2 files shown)


types/js-cookie/index.d.ts
```diff
@@ -69,18 +69,6 @@ declare namespace Cookies {
          */
         get(): {[key: string]: string};
 
-        /**
```
amcasey: Were these removed in 3.0?  If so, does the version number at the top need to be updated?

amcasey (changes requested):

types/js-cookie/index.d.ts
nreynis: You're right. I was confused by the fact that version umber was bumped two commits after. Will fix that ASAP.

nreynis: I bumped the version number, but I feel like this should wait the end of the beta before being merged.

Should I close the PR and resub;it it later ?

nreynis: Also, the generic on `CookieStatic` should expand `string` and not `object` as set doesn't accept array or literals anymore.

amcasey: If you like, you can mark it as Draft until you're ready to go.

amcasey (approved):
LGTM, but I understand we're waiting for 3.0 to leave beta.

types/js-cookie/index.d.ts
nreynis: Well, I have another one to submit for branch 2.x, so this one will be rebased. It will be more simple to just resubmit it later.

Thank you for the review!

PR #40685, opened 2019-11-26
Normalise cli-color files, fixup javascript-time-ago

sandersn:


types/javascript-time-ago/OTHER_FILES.txt
```diff
@@ -0,0 +1,131 @@
+locale/lb/index.d.ts
```
sandersn: Only 3 are ever tested, which is reasonable since all the files are nearly identical.

PR #40688, opened 2019-11-27
Fix missing module export for grecaptcha

swiftlee:
With this minor fix, one would now be able to properly use ES6 style imports to access the grecaptcha typings.

Previously:

`import grecaptcha from 'grecaptcha';`
The ES6 style import statement would throw a compile-time error stating the the module did not exist in `@types/grecaptcha/index.d.ts`

After the fix:

`import grecaptcha from 'grecaptcha';`
The ES6 style import statement _no longer_ presents developers with a compile-time error and can properly use the typings provided by the `index.d.ts`.

### DIFF
--- a/types/grecaptcha/index.d.ts
+++ b/types/grecaptcha/index.d.ts
@@ -8,6 +8,7 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 declare var grecaptcha: ReCaptchaV2.ReCaptcha;
+export = grecaptcha;
 
 declare namespace ReCaptchaV2 {
   interface ReCaptcha {


types/grecaptcha/index.d.ts
```diff
@@ -8,6 +8,7 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 declare var grecaptcha: ReCaptchaV2.ReCaptcha;
+export = grecaptcha;
```
rafaeltavares: This type definition is intended to map the client script of Google reCaptcha.
The script's API is provided as a global variable, not as a module.

swiftlee: I suppose that makes sense. I had quite a few issues trying to access it properly while still having the standard grecaptcha (non-ts library) installed. I think the var names were duplicate in the global namespace from both the types and the non-typed.

Side note: I could be totally wrong, I just had issues with this :)

PR #40699, opened 2019-11-27
fix: incorrect signature cause ts2345

stevenhankin:
Test incorrectly used a static Axios config and needed to pass an Axios Instance
since the Axios module is updated to inject an interceptor

Please fill in this template.

- [y ] Use a meaningful title for the pull request. Include the name of the package modified.
- [y ] Test the change in your own code. (Compile and run.)
- [ y] Add or edit tests to reflect the change. (Run with `npm test`.)
- [ y] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [y ] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [ y] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If adding a new definition:

- [ n] Provide a URL to documentation or source code which provides context for the suggested changes: <<url here>>
- [ n] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ n] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

If removing a declaration:

### DIFF
--- a/types/axios-curlirize/index.d.ts
+++ b/types/axios-curlirize/index.d.ts
@@ -5,12 +5,16 @@
 
 export = AxiosCurlirize;
 
-import { AxiosStatic } from 'axios';
+import { AxiosInstance } from 'axios';
 
-interface Callback {
-    (error: Error, result?: number): void;
+interface Result {
+    command: string;
 }
 
-declare function AxiosCurlirize(instance: AxiosStatic, callback?: Callback): void;
+interface Callback {
+    (result: Result, error: Error): void;
+}
+
+declare function AxiosCurlirize(instance: AxiosInstance, callback?: Callback): void;
 
 declare namespace AxiosCurlirize { }
--- a/types/axios-curlirize/index.d.ts
+++ b/types/axios-curlirize/index.d.ts
@@ -1,6 +1,6 @@
 // Type definitions for axios-curlirize 1.3
 // Project: https://github.com/delirius325/axios-curlirize#readme
-// Definitions by: Steven Hankin <https://github.com/me>
+// Definitions by: Steven Hankin <https://github.com/stevenhankin>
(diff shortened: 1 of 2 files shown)


types/axios-curlirize/index.d.ts
```diff
@@ -5,12 +5,16 @@
 
 export = AxiosCurlirize;
```
amcasey: Normally, this would have been fast-tracked, since you're the owner.  However, it looks like the header template wasn't updated, so the bot couldn't identify you.  Was that supposed to be https://github.com/stevenhankin?

stevenhankin: Hello @amcasey !  Yes, I'm the owner of the template (but not of the referenced axios-curlirize module)

amcasey: Sorry, I think my question was unclear.  At the top of this file, it says 
```ts
// Definitions by: Steven Hankin <https://github.com/me>
```
Should it instead say
```ts
// Definitions by: Steven Hankin <https://github.com/stevenhankin>
```

stevenhankin: Lol...sorry, I'm new to this stuff.  Yes it should; I'll update the PR in a few mins...

amcasey: No worries.  Once you correct it, future changes you make to this type definition should be fast-tracked by the bot.

amcasey (approved):

PR #40700, opened 2019-11-27
[@types/cesium] Add some missing class and update to Cesium 1.63

Crocsx:
*edit: sorry I made a mess I duplicated the PR*

class Additions are took from :

https://cesium.com/docs/cesiumjs-ref-doc/Ion.html?classFilter=Ion
https://cesium.com/docs/cesiumjs-ref-doc/PostProcessStageLibrary.html
https://cesium.com/docs/cesiumjs-ref-doc/PostProcessStageComposite.html

change to existing class took from :

https://github.com/AnalyticalGraphicsInc/cesium/blob/master/CHANGES.md

Not in doc  :  
`Cesium.buildModuleUrl` and `Cesium.when` exist but are not documented :  
exemple (look at the doc code exemple, it contains the method) : 
Cesium.when => https://cesium.com/docs/cesiumjs-ref-doc/sampleTerrainMostDetailed.html
*but I do not know exactly how to type that one, so I made a generic, if you could help for that one*

Cesium.buildModuleUrl => https://cesium.com/docs/cesiumjs-ref-doc/BaseLayerPicker.html

### DIFF
--- a/types/cesium/index.d.ts
+++ b/types/cesium/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for cesium 1.59
+// Type definitions for cesium 1.63
 // Project: http://cesiumjs.org
 // Definitions by: Aigars Zeiza <https://github.com/Zuzon>
 //                 Harry Nicholls <https://github.com/hnipps>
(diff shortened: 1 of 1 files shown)


Crocsx:
concerning the Compilation fail, Sorry but I do not really understand how to fix this 
I have 

```
    class Ion {
        static defaultAccessToken: string;
        static defaultServer: string | Resource;
    }
```

changing it to 

```
    namespace Ion {
        type defaultAccessToken = string;
        type defaultServer = string | Resource;
    }
``` 
or 
```
    namespace Ion {
        interface Props {
            defaultAccessToken: string;
            defaultServer: string | Resource;
        }
    }
```
make it undefined

```Property 'Ion' does not exist on type 'typeof Cesium'```

Is there a way I do not know ?

PR #40701, opened 2019-11-27
feat: Add dialogflow-fulfillment

nightillusions:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/dialogflow-fulfillment/index.d.ts
+++ b/types/dialogflow-fulfillment/index.d.ts
@@ -10,7 +10,7 @@
 import { DialogflowConversation } from 'actions-on-google';
 import { Request, Response } from 'express';
 
-export enum Platform {
+export enum Platforms {
     UNSPECIFIED = 'PLATFORM_UNSPECIFIED',
     FACEBOOK = 'FACEBOOK',
     SLACK = 'SLACK',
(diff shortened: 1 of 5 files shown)


types/dialogflow-fulfillment/index.d.ts
```diff
@@ -0,0 +1,217 @@
...
+//                 SlowTick <https://github.com/slowtick>
+//                 Diego Dal Bosco <https://github.com/diegodalbosco>
+//                 Marcel <https://github.com/MTRNord>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.1
+
+import { DialogflowConversation } from 'actions-on-google';
+import { Request, Response } from 'express';
+
+export enum Platform {
```
amcasey: It looks like the underlying type (i.e. the one available at runtime) might be called "PLATFORMS": https://github.com/dialogflow/dialogflow-fulfillment-nodejs/blob/8e7b12bcea64fc3b923b1845eb4c280d610825d0/src/rich-responses/rich-response.js#L18

amcasey: (i.e. it seems like `Platform.UNSPECIFIED` won't resolve at runtime)

amcasey (changes requested):

types/dialogflow-fulfillment/index.d.ts
nightillusions: I renamed `Platform` to `Platforms`

amcasey: I think it might need to be in all-caps, as in the underlying code.  You can confirm by referencing `Platforms.UNSPECIFIED` in your code and checking whether it's defined at runtime.

PR #40704, opened 2019-11-27
Add sticky-session to types

vanny96:
Please fill in this template.

If adding a new definition:


package.json
```diff
@@ -22,6 +22,7 @@
         "lint": "dtslint types"
     },
     "devDependencies": {
+        "@types/node": "^12.12.14",
```
amcasey: Why was this needed?

amcasey (changes requested):

vanny96:
@amcasey sorry for the mistake! It's my first time contributing so I made some mistakes, if there is another error please tell me!

package.json
vanny96: It wasn't sorry! I updated the pull request, let me know if I have to update anything!

types/sticky-session/index.d.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * Method to create the sticky session
+ * @param server Server used to create the session
+ * @param port Port where the app will be listening at
+ * @param options Options to customize the sticky session
+ */
+export function listen(
+    server: Server,
+    port: number,
+    options: {
```
demensky: `options` is [optional](https://github.com/indutny/sticky-session/blob/master/lib/sticky/api.js#L12).

types/sticky-session/index.d.ts
```diff
@@ -0,0 +1,22 @@
...
+        workers: number
```
demensky: `workers` is [optional](https://github.com/indutny/sticky-session/blob/master/lib/sticky/api.js#L15).

types/sticky-session/index.d.ts
demensky: Please add [`env`](https://github.com/indutny/sticky-session/blob/master/lib/sticky/api.js#L17).

types/sticky-session/index.d.ts
```diff
@@ -0,0 +1,22 @@
...
+
+/**
+ * Method to create the sticky session
+ * @param server Server used to create the session
+ * @param port Port where the app will be listening at
+ * @param options Options to customize the sticky session
+ */
+export function listen(
+    server: Server,
+    port: number,
```
demensky: `port` is [optional](https://github.com/indutny/sticky-session/blob/master/lib/sticky/api.js#L18) ([`net.Server#listen`](https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback)).

types/sticky-session/index.d.ts
demensky: I propose to add `readonly`.

demensky (changes requested):

package.json
amcasey: Thanks!

types/sticky-session/index.d.ts
vanny96: I added ```env``` with ```any``` type because this is what the ```cluster.fork(env)``` method uses. If there is any correction to do let me know!

demensky: `any` here is probably better.

types/sticky-session/index.d.ts
```diff
@@ -0,0 +1,24 @@
+// Type definitions for sticky-session 1.1
+// Project: https://github.com/indutny/sticky-session
+// Definitions by: Giovanni Barbaro <https://github.com/vanny96>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+/// <reference types="node" />
+
+import { Server } from "http";
+import { ProcessEnvOptions } from "child_process";
```
demensky:
```suggestion
```

You just need to remove the `ProcessEnvOptions` which is probably left by accident.

demensky (changes requested):

amcasey (approved):
LGTM after making the last fix suggested by @demensky

demensky (approved):

demensky (changes requested):
Ups... wrong button.

vanny96:
Sorry for the delay! It was a busy week so I had no time to check back! Anyway I removed the unused import, let me know if there is anything more I could do!

types/sticky-session/index.d.ts
vanny96: Yes I left them when I was playing around with the `env` type, I then opted for `any` instead but forgot to remove the import! I have updated the pull request! Let me know if there is anything more I have to change!

demensky (approved):

orta:
Alright, this looks good 👍

PR #40710, opened 2019-11-27
js-cookie: better default generic and converter typing

nreynis:
I made several changes to how generics are used:

- CookieStatic shouldn't extends `object` because you wan't to be able to pass `string`. The fact that it didn't accept `string` is why you were forced to use the union type `string | T`. 
- CookieStatic default generic value can't be `object`, the library only meant to accept arrays, object litterals or plain old string: https://github.com/js-cookie/js-cookie/tree/1698b1ae5fee119aa3d02f3be8e31260e02f1475#json
- CookieStatic now has two generics T for input and U for output (note that when branch 3.x, which drop Json support, leave beta this could be simplified back to a single generic)
- `get` returns `U` instead of hardcoded `string`
- `CookieReadConverter` use the same generic than `CookieWriteConverter`
- The `read` method isn't optional when using `withConverter` expanded syntax

### DIFF
--- a/types/js-cookie/index.d.ts
+++ b/types/js-cookie/index.d.ts
@@ -48,7 +48,7 @@ declare namespace Cookies {
         [property: string]: any;
     }
 
-    interface CookiesStatic<T extends object = object> {
+    interface CookiesStatic<T, U> {
         /**
          * Allows default cookie attributes to be accessed, changed, or reset
          */
@@ -57,17 +57,17 @@ declare namespace Cookies {
         /**
          * Create a cookie
          */
-        set(name: string, value: string | T, options?: CookieAttributes): void;
+        set(name: string, value: T, options?: CookieAttributes): void;
 
         /**
          * Read cookie
          */
-        get(name: string): string | undefined;
+        get(name: string): U | undefined;
 
         /**
          * Read all available cookies
          */
-        get(): {[key: string]: string};
+        get(): {[key: string]: U};
 
         /**
          * Returns the parsed representation of the string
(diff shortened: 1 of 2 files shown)


rbuckton:
The built-in JSON support causes problems with being able to intuit what is actually happening under the covers. It seems like any kind of `string -> T` or `T -> string` conversion is inherently unsound due to how the library is written. The only *sound* converters possible right now are `string -> string`, due to the fact that `set` calls `JSON.stringify` *before* calling a write converter, and `get` will return the decoded cookie value if the read-converter returns something falsey (i.e. `null`, `undefined`, `0`, `false`, or `""`).

I'm not sure I see the value of adding a second generic given this unsoundness. I will need to review this further to see if I have any other advice I can give.

nreynis:
@rbuckton You are totally right. I overlooked this, it works only for object which are not JSON serializable (ex: https://jsfiddle.net/89dpo21f/).

So I guess this would have to wait for version 3.0 which drop JSON support.
This is a pity because in my opinion current generics is also broken and don't make sense.  
When I see in the tests things like:
```ts
const PHPCookies = Cookies.withConverter<object>({
    write(value) {
        value; // $ExpectType string | object
        return encodeURIComponent(value as string)
            .replace(/%(23|24|26|3A|3C|3E|3D|2F|3F|40|5B|5D|5E|60|7B|7D|7C)/g, decodeURIComponent);
    },
    read(value) {
        value; // $ExpectType string
        return value
            .replace(/\+/g, ' ')
            .replace(/(%[0-9A-Z]{2})+/g, decodeURIComponent);
    }
});
```
You just specify an `<object>` for totally no reason or added value. It's even confusing because you'll likely never see any object in input of your converter and can't output one either.

Is it possible to make a type definition for a version in beta (3.0.0-beta.0) or does DefinitelyTyped require strict X.Y.Z semver numbers ?

rbuckton:
DT packages are based on the `major.minor` part of a published version. You could create a 3.0 version of the package types side by side with the current version using the steps outlined here: https://github.com/DefinitelyTyped/DefinitelyTyped#if-a-library-is-updated-to-a-new-major-version-with-breaking-changes-how-should-i-update-its-type-declaration-package

nreynis:
Thank you for the review. I'll close that and resubmit an updated PR when version 3 do stabilize.

PR #40716, opened 2019-11-28
Add akamai-edgeworkers

evan-hughes:
Please fill in this template.

If adding a new definition:


types/akamai-edgeworkers/index.d.ts
```diff
@@ -0,0 +1,296 @@
...
+         */
+        zipCode: string | undefined;
+    }
+
+    /**
+     * Notes:
+     * * If device properties can not be supplied for any reason,
+     *   undefined is returned for each property
+     */
+    class Device {
```
rbuckton: From what I can gather, the `EW` namespace in this file doesn't actually exist at runtime, and is provided solely as a place to define the types needed for the four methods mentioned in the README (and at https://developer.akamai.com/api/web_performance/edgeworkers/v1.html#edgeworkersbundleformat). By defining `Device` as a `class`, that changes the `EW` namespace from a "type-only" namespace to a "value" namespace (since classes are both types *and* values), meaning that `EW` will appear to exist as a valid expression. I would recommend changing `Device` to an interface to avoid this:

```suggestion
    interface Device {
```

types/akamai-edgeworkers/test/cookies-tests.ts
```diff
@@ -0,0 +1,51 @@
+import { Cookies, SetCookie } from 'akamai-edgeworkers/cookies';
```
rbuckton: This doesn't seem to match how these would be referenced in the akamai edgeworkers samples: https://github.com/akamai/edgeworkers-examples/blob/f273d82ca7186e08345ab65d383db4e723a03282/cookie-accrual/main.js#L11

types/akamai-edgeworkers/test/url-search-params-tests.ts
```diff
@@ -0,0 +1,63 @@
+import URLSearchParams from 'akamai-edgeworkers/url-search-params';
```
rbuckton: This doesn't match how this module is referenced from the akamai edgeworkers samples: https://github.com/akamai/edgeworkers-examples/blob/f273d82ca7186e08345ab65d383db4e723a03282/conference-details/main.js#L8

types/akamai-edgeworkers/README.md
```diff
@@ -0,0 +1,62 @@
...
+export function onClientRequest(request : EW.MutableRequest & EW.HasRespondWith){}
+export function onOriginRequest(request : EW.MutableRequest) {}
+export function onOriginResponse(request : EW.ImmutableRequest & EW.HasRespondWith, response : EW.Response) {}
+export function onClientResponse(request : EW.ImmutableRequest, response : EW.Response) {} 
+```
+
+The triple-slashed first line references this package and pulls `EW` into your 
+namespace. 
+
+## Using Built-In Modules
```
rbuckton: Instead of using a `"paths"` mapping, you could move the contents of `cookies.d.ts` and `url-search-params.d.ts` into `index.d.ts` as ambient modules, similar to how NodeJS handles its module definitions. I've added more specifics about this in the relevant files.

types/akamai-edgeworkers/cookies.d.ts
```diff
@@ -0,0 +1,97 @@
+/**
```
rbuckton: Instead of using a `"paths"` mapping, you could move the contents of `cookies.d.ts` into `index.d.ts` as an ambient module, similar to how NodeJS handles its module definitions:

```ts
declare module "cookies" {
    export class Cookies { ... }
    ...
}
```

types/akamai-edgeworkers/url-search-params.d.ts
```diff
@@ -0,0 +1,56 @@
+export default class URLSearchParams {
```
rbuckton: Instead of using a `"paths"` mapping, you could move the contents of `url-search-params.d.ts` into `index.d.ts` as an ambient module, similar to how NodeJS handles its module definitions:

```ts
declare module "url-search-params" {
    export class URLSearchParams { ... }
    ...
}
```

types/akamai-edgeworkers/README.md
```diff
@@ -0,0 +1,62 @@
...
+```    
+
+## Using the `Request` and `Response` Objects
+
+The predefined EdgeWorker callbacks take Request and Response objects as 
+arguments. After you have installed this package, you can create a `main.ts`
+with the following stubs:
+
+```typescript
+/// <reference types="akamai-edgeworkers-tsc"/>
```
rbuckton: Going by the path to this package, I think the `-tsc` is extraneous:

```suggestion
/// <reference types="akamai-edgeworkers"/>
```

rbuckton (changes requested):

types/akamai-edgeworkers/index.d.ts
evan-hughes: You're correct - I've made the change. Thanks for the background.

types/akamai-edgeworkers/test/cookies-tests.ts
evan-hughes: I believe I ran into problems with the build when I included a path mapping in the tsconfig.json. I'll try switching it back to the regular convention and see what Travis has to say.

types/akamai-edgeworkers/README.md
evan-hughes: That's much better than messing with paths. Thanks for the tip!

types/akamai-edgeworkers/test/cookies-tests.ts
evan-hughes: Now that the extra modules are defined in the index.d.ts, they can be referenced without the extra class. Please ignore my previous comment.

rbuckton (approved):

PR #40717, opened 2019-11-28
[victory] Add 'VictoryClipContainer' and 'Flyout' types and fixing theme types

Sophmrs:
Please fill in this template.

The `VictoryThemeDefinition` type changes make it compatible with the default themes:
[greyscale](https://github.com/FormidableLabs/victory/blob/da57673edd8963b4c6549a3963e5e6b1d2ed0d0d/packages/victory-core/src/victory-theme/grayscale.js)
[material](https://github.com/FormidableLabs/victory/blob/da57673edd8963b4c6549a3963e5e6b1d2ed0d0d/packages/victory-core/src/victory-theme/material.js)


types/victory/index.d.ts
```diff
@@ -327,6 +327,112 @@ declare module 'victory' {
         y: number;
     };
 
+    export interface VictoryClipContainerProps {
```
ibarrae: Maybe it's missing `extends VictoryContainerProps` due to [common container props](https://formidable.com/open-source/victory/docs/common-container-props)? Then this interface should only have the props specific to it. WDYT @Theomg

types/victory/index.d.ts
```diff
@@ -327,6 +327,112 @@ declare module 'victory' {
...
+         */
+        clipWidth?: number;
+        /**
+         * The `events` prop attaches arbitrary event handlers to the group element.
+         * This prop should be given as an object of event names and corresponding event handlers.
+         * When events are provided via Victory's event system, event handlers will be called with
+         * the event, the props of the component it is attached to, and an `eventKey` when applicable.
+         * @example events={{onClick: (evt) => alert("x: " + evt.clientX)}}
+         */
+        events?: object;
```
ibarrae: Maybe it's worth using the same type as [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/e02df99913d0c992415eec15a2cb1ba33e8ccd6c/types/victory/index.d.ts#L181)?

types/victory/index.d.ts
```diff
@@ -327,6 +327,112 @@ declare module 'victory' {
...
+         * It is usually necessary to set deterministic ids for automated testing.
+         */
+        clipId?: number | string;
+        /**
+         * The `clipPadding` prop is used when the clipped area should be larger than the range of a chart.
+         * This prop should be given as an object with `top`, `bottom`, `left`, and `right` properties. Set
+         * the `clipPadding` prop is useful for extending the visible area of a chart in some dimension so
+         * that data or labels are not cut off.
+         */
+        clipPadding?: {
```
ibarrae: There's an existing type called `BlockProps` with the same definitions, I think it's worth using it :smile:

ibarrae (commented):
Looks great! Just left some suggestions :smiley:

types/victory/index.d.ts
Theomg: Great idea, didn't know about `DOMAttributes`, it's a really useful type.

types/victory/index.d.ts
Theomg: Agreed.

types/victory/index.d.ts
Theomg: I don't think `VictoryClipContainer` has anything to do with other containers though.
The code(https://github.com/FormidableLabs/victory-core/blob/master/src/victory-clip-container/victory-clip-container.js#L10) makes no references to the `VictoryContainer` used as a base for other containers such as `VictoryVoronoiContainer` (https://github.com/FormidableLabs/victory/blob/master/packages/victory-voronoi-container/src/victory-voronoi-container.js#L196), which is where they get the `VictoryContainerProps` from.
It would deduplicate `children`, `className` and `events`, but I'm unsure if it would really be the correct thing to do here. Do you think it's worth doing it anyways?

ibarrae: I agree with not extending the whole `VictoryContainerProps` interface, but since you mention it would reuse `children`, `className` and `events` maybe we could use [pick](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktk).

``` typescript
... extends Pick<VictoryContainerProps, "children" | "className" | "events">
```
WDYT?

Theomg: I was going by the types in the documentation, but it seems like in the actual type definitions neither `VictoryContainerProps` nor `VictoryCommonProps` have `className` or `children`, and other types like `VictoryLabelProps` implement them themselves. `children` could be left to React.Component but would lose its specificity and its JSDoc.
Only `events` can be picked from `VictoryContainerProps`, but I personally wouldn't use that as the component itself doesn't inherit from it, but if you think it's worth it adding the `Pick<VictoryContainerProps, 'events'>` I could add it. Do you think it's better to add the Pick with only one property?

ibarrae: :thinking: I don't think we should use `Pick` for just one prop, I think we should go with `DOMAttributes` once again.

Theomg: Alright, already using the `DOMAttributes` on the events.
I believe everything is according to your suggestions then.

sandersn:
@ibarrae are all your comments addressed? I think this is ready to merge if so.

ibarrae:
Yeah, everything looks great to me @sandersn

PR #40718, opened 2019-11-28
[dropzone] Update types for success callback

aried3r:
Please fill in this template.

- [x]  Use a meaningful title for the pull request. Include the name of the package modified.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: `on("success")` callback:
https://www.dropzonejs.com/#event-success
https://gitlab.com/meno/dropzone/blob/abe7a2af6873feaacec407c0c64cc3842211ad8f/src/dropzone.js#L2239
`success` config:
https://gitlab.com/meno/dropzone/blob/abe7a2af6873feaacec407c0c64cc3842211ad8f/src/dropzone.js#L870-876

### DIFF
--- a/types/dropzone/index.d.ts
+++ b/types/dropzone/index.d.ts
@@ -131,7 +131,7 @@ declare namespace Dropzone {
         sending?(file: DropzoneFile, xhr: XMLHttpRequest, formData: FormData): void;
         sendingmultiple?(files: DropzoneFile[], xhr: XMLHttpRequest, formData: FormData): void;
 
-        success?(file: DropzoneFile, response: Object | string): void;
+        success?(file: DropzoneFile): void;
         successmultiple?(files: DropzoneFile[], responseText: string): void;
 
         canceled?(file: DropzoneFile): void;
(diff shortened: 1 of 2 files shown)


types/dropzone/index.d.ts
```diff
@@ -131,7 +131,7 @@ declare namespace Dropzone {
         sending?(file: DropzoneFile, xhr: XMLHttpRequest, formData: FormData): void;
         sendingmultiple?(files: DropzoneFile[], xhr: XMLHttpRequest, formData: FormData): void;
 
-        success?(file: DropzoneFile, response: Object | string): void;
```
PikachuEXE: image
The file has been uploaded successfully. Gets the server response as second argument. (This event was called finished previously)
So the option `success` should still have `response` as 2nd argument
It's just that the default option does not need `response` at all

Should revert change for test too

PikachuEXE (changes requested):

types/dropzone/index.d.ts
aried3r: Hey! Thanks for your review!

I believe this is not the case. What you're commenting on is the _configuration option_ in the types, but the screenshot refers to the _events_, see [here](https://gitlab.com/meno/dropzone/blob/abe7a2af6873feaacec407c0c64cc3842211ad8f/src/dropzone.js#L870-876) for the config option which indeed does only receive `file`. The event list receives both, `file` and `response`, see [here](https://gitlab.com/meno/dropzone/blob/abe7a2af6873feaacec407c0c64cc3842211ad8f/src/dropzone.js#L2239).

Sadly the dropzone.js website is missing the config option in their documentation. But maybe this is indeed on purpose as the [event list documentation](https://www.dropzonejs.com/#event-list) states:

> Do **not** overwrite those as configuration options, unless you know what you're doing.

PikachuEXE (approved):
After a closer look on the code I can't even see where is the option `success` is used
I am fine with sticking to default options

orta:
Looks good, thanks for the PR @aried3r - sorry it took some time to get around to this

PR #40720, opened 2019-11-28
fix(ioredis): make types more accurate

SimonSchick:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://redis.io/commands

This is mostly narrowing almost all instances of `any` to something more useful.
There are still quite a few things missing (due to methods being autogenerated in the package) but this should help a few developers :)

Also remove some private members from interfaces.

### DIFF
--- a/types/ioredis/ioredis-tests.ts
+++ b/types/ioredis/ioredis-tests.ts
@@ -80,15 +80,8 @@ new Redis({
 });
 
 const pub = new Redis();
-redis.subscribe('news', 'music', (err: any, count: any) => {
-    // Now we are subscribed to both the 'news' and 'music' channels.
-    // `count` represents the number of channels we are currently subscribed to.
 
-    pub.publish('news', 'Hello world!');
-    pub.publishBuffer('music', Buffer.from('Hello again!'));
-});
-
-redis.on('message', (channel: any, message: any) => {
+pub.on('message', (channel: any, message: any) => {
     // Receive message Hello world! from channel news
     // Receive message Hello again! from channel music
     console.log('Receive message %s from channel %s', message, channel);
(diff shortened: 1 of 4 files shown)


types/ioredis/ioredis-tests.ts
```diff
@@ -80,23 +80,16 @@ new Redis({
 });
 
 const pub = new Redis();
-redis.subscribe('news', 'music', (err: any, count: any) => {
```
plantain-00: Why this removed?

plantain-00:
Other changes look good, basically `any` to more accurate type, it's breaking change but worth it.
Need review from other people.

types/ioredis/ioredis-tests.ts
SimonSchick: Right I meant to ask about this, the callback technically comes after the vararg which is not possible to model in TS (yet) so I removed this test as I couldn't satisfy it.
The possible options here are to add multiple overloads and an `array` variant instead eg.

```ts
subscribe(sub1: string, cb: (err: Error, count: number) => void): Promise<number>;
...
subscribe(subs: string[], cb: (err: Error, count: number) => void): Promise<number>;
```

Since ioredis flattens all arguments by default, this would be a valid (and usable) definition.

SimonSchick: @plantain-00 ping

plantain-00 (approved):

JanFellner:
I have issues using the new d.ts.
```
const redis = new IORedis(params);
redis.client("list", (error: any, result: string): void => {
     // whatever
});

```

Argument of type '(error: any, result: string) => void' is not assignable to parameter of type 'ValueType'.
  Type '(error: any, result: string) => void' is missing the following properties from type 'Buffer': write, toJSON, equals, compare, and 77 more.

From the d.ts. ValueType is either a dedicated type or an array of any[].

Where am i wrong?

SimonSchick:
The problem is that TS (and technically JS) do not support a single argument after a vararg making this kind of call pattern notoriously hard to model in typings.
The callback variant was never explicitly defined therefor this broke stuff.
In fact, looking at the definitions, most of them appear to be wrong since even if you provide a callback you still get a promise......

For now I'd suggest you just cast to `any` while I try to figure out how to improve these things.

PR #15858, opened 2017-04-14
fix(ioredis): allow RedisOptions.retryStrategy to return false

satoshun00:
- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [x] Provide a URL to documentation or source code which provides context for the suggested changes:
  - https://github.com/luin/ioredis/tree/master#auto-reconnect
  - > When the return value isn't a number, ioredis will stop trying to reconnect
- [ ] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "../tslint.json" }`.

PR #40726, opened 2019-11-29
Update ioredis sentinel retry strategy

vecerek:
The new declaration of the `sentinelRetryStrategy` option is:
```ts
(retryAttempts: number) => number | void | null;
```

~But for consistency with `retryStrategy`, I decided to use:~
```ts
(times: number) => number | false;
```

Instead of ☝️ it has been decided to change `retryStrategy`'s declaration to `(times: number) => number | void | null;` in order to stay consistent with the original implementation. See https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40726#issuecomment-559893144.

---

Please fill in this template.

   - https://github.com/luin/ioredis/compare/v4.14.4...v4.15.0
   - https://github.com/luin/ioredis/blob/06b28e1c0a7d9ce163037d4c569b6b94970d99d5/lib/connectors/SentinelConnector/index.ts#L40

### DIFF
--- a/types/ioredis/index.d.ts
+++ b/types/ioredis/index.d.ts
@@ -1014,7 +1014,7 @@ declare namespace IORedis {
          * When the return value isn't a number, ioredis will stop trying to reconnect.
          * Fixed in: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/15858
          */
-        retryStrategy?(times: number): number | false;
+        retryStrategy?(times: number): number | void | null;
         /**
          * By default, all pending commands will be flushed with an error every
          * 20 retry attempts. That makes sure commands won't wait forever when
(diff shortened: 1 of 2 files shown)


plantain-00:
I think it should be `number | void | null` for the both.

vecerek:
It would definitely be more aligned with the package. Wouldn’t it be a breaking change for `retryStrategy`, though? Or it doesn’t matter too much because the compilers would catch that anyway and it wouldn’t be much work to adjust to the new declaration?

plantain-00:
The change was introduced 2 years ago: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/15858
and it's v3: https://github.com/luin/ioredis/releases/tag/v3.2.2
so a breaking change should be OK.

plantain-00 (approved):

PR #40728, opened 2019-11-29
[pg] Updated Client and Pool config types according to docs

schibrikov:
I found that typings are incorrect according to the documentation.
query_timeout option is missing, but connectionTimeoutMillis is present for Client config, which is not right and doesn't work.

https://node-postgres.com/api/client
https://node-postgres.com/api/pool

### DIFF
--- a/types/pg/index.d.ts
+++ b/types/pg/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for pg 7.11
+// Type definitions for pg 7.14
 // Project: http://github.com/brianc/node-postgres
 // Definitions by: Phips Peter <https://github.com/pspeter3>, Ravi van Rooijen <https://github.com/HoldYourWaffle>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
@@ -10,7 +10,9 @@ import events = require("events");
 import stream = require("stream");
 import pgTypes = require("pg-types");
 
-export interface ConnectionConfig {
+import { ConnectionOptions } from "tls";
+
+export interface ClientConfig {
     user?: string;
     database?: string;
     password?: string;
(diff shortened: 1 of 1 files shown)


types/pg/index.d.ts
```diff
@@ -1,8 +1,8 @@
-// Type definitions for pg 7.11
+// Type definitions for pg 7.14
 // Project: http://github.com/brianc/node-postgres
 // Definitions by: Phips Peter <https://github.com/pspeter3>, Ravi van Rooijen <https://github.com/HoldYourWaffle>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+// TypeScript Version: 3.7
```
HoldYourWaffle: Why was this necessary?

sHooKDT: Not necessary at all I believe. Changed back.

HoldYourWaffle (changes requested):

types/pg/index.d.ts
```diff
@@ -34,9 +35,7 @@ export interface Defaults extends ConnectionConfig {
 
 import { ConnectionOptions } from "tls";
 
-export interface ClientConfig extends ConnectionConfig {
-    ssl?: boolean | ConnectionOptions;
-}
+export interface ClientConfig extends ConnectionConfig {}
```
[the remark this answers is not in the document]
sHooKDT: Yes, me neither. Though I wasn't sure I can just remove something, because it could be used by someone. 
If you think it's okay, I will fix it soon.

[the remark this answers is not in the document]
sHooKDT: Seems that it's even used in other typings for some reason. I'll get this interface back and disable linter rule then.

types/pg/index.d.ts
```diff
@@ -20,24 +22,21 @@ export interface ConnectionConfig {
...
     stream?: stream.Duplex;
     statement_timeout?: false | number;
-    connectionTimeoutMillis?: number;
+    ssl?: boolean | ConnectionOptions;
+    query_timeout?: number;
     keepAliveInitialDelayMillis?: number;
 }
 
-export interface Defaults extends ConnectionConfig {
+export type ConnectionConfig = ClientConfig;
```
HoldYourWaffle: I really like this approach, but I believe this breaks userland interfaces that *extend* `ConnectionConfig`. I don't know if this is an actual problem or if this is 'niche' enough to be safely changed, either way I don't think I'm the right person to decide that.
Are there DT guidelines for these kind of situations?

sHooKDT: Extending types seems to work, at least in the newest TS versions https://www.typescriptlang.org/play/index.html?ssl=1&ssc=1&pln=12&pc=2#code/C4TwDgpgBAglC8UDeAoK6oEMBcUB2ArgLYBGEATgNwoC+KKAlnsBQGaYDG0AQlBAB4s8AEwDOsZGgwlco4OSYBzanRQcA9njlQAbpgWYSAGwi5eiVBiy4ATAGYALABop6GVADko4aw+0gA
What are the problems you aware of?

sHooKDT:
Bump, I believe it's ready for merge. Can someone please have a look once more?

types/pg/index.d.ts
HoldYourWaffle: I guess I was wrong then, perhaps it was changed in a recent version.

HoldYourWaffle (approved):

PR #40731, opened 2019-11-29
power-assert: update assert function

falsandtru:
Use http://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions


types/power-assert/ts3.7/index.d.ts
```diff
@@ -0,0 +1,36 @@
...
+import * as module from "../index";
+
+export = assert;
+export as namespace assert;
+
+declare function assert(value: any, message?: string): asserts value;
+declare namespace assert {
+    export import AssertionError = module.AssertionError;
+
+    export import fail = module.fail;
```
rbuckton: Would it make more sense to have fail be the following?
```suggestion
    export function fail(actual?: any, expected?: any, message?: string, operator?: string): never;    
```

types/power-assert/ts3.7/index.d.ts
```diff
@@ -0,0 +1,36 @@
...
+    export import ok = module.ok;
```
rbuckton: Would it make more sense to have `ok` be the following?
```suggestion
    export function ok(value: any, message?: string): asserts value;
```

types/power-assert/ts3.7/index.d.ts
falsandtru: Why? All the others are the same?

falsandtru: I didn't see the return type. I'll fix the original definition.

falsandtru: done

types/power-assert/ts3.7/index.d.ts
falsandtru: done.

rbuckton (approved):

PR #40738, opened 2019-11-29
[ramda] cleanup

millsp:
Please fill in this template.

* Updated `ts-toolbelt`
* Fixed `match` return type
* Moved all type tools to `tools.d.ts`
* Better typings for `merge` function family
* Flattening the declaration file
* Moved references to `OTHER_FILES.txt`


pirix-gh:
@typescript-bot test perf

Philippe-mills (approved):

types/ramda/index.d.ts
```diff
@@ -540,246 +539,37 @@
...
+    ValueOfRecord,
+} from './tools';
 
 declare let R: R.Static;
 
 declare namespace R {
-    import ValueOfRecord = Tools.ValueOfRecord;
-    type Omit<T, K extends string> = Pick<T, Exclude<keyof T, K>>;
-
-    type Ord = number | string | boolean | Date;
```
rbuckton: Will removing these from `R` break anyone using `R.Ord`, etc.?

pirix-gh: Hi Ron, thanks for pointing this out. I will look for an alternative solution.

pirix-gh:
@rbuckton I re-exported the types from the namespace. It's not exactly how I wanted to do this but I couldn't find a way to export the types generically from the `tools.d.ts` file. I could use the namespace but we're limited because of the `import` (...but maybe, you know how to solve this :smiley_cat: ).

Otherwise we can leave this as it is now, it is still cleaner than before.

types/ramda/index.d.ts
```diff
@@ -540,245 +539,36 @@
...
-        <T, U>(val: T, obj: U): Record<K, T> & U;
-    }
+    type Arity0Fn                                                                   = Tools.Arity0Fn;
+    type Arity1Fn                                                                   = Tools.Arity1Fn;
+    type Arity2Fn                                                                   = Tools.Arity2Fn;
+    type AssocPartialOne<K extends keyof any>                                       = Tools.AssocPartialOne<K>;
+    type ComposeWithFns<V0, R>                                                      = Tools.ComposeWithFns<V0, R>;
+    type Dictionary<A>                                                              = Tools.Dictionary<A>;
+    type Evolvable<E extends Evolver>                                               = Tools.Evolvable<E>;
+    type Evolve<O extends Evolvable<E>,  E extends Evolver>                         = Tools.Evolve<O, E>;
```
rbuckton: I believe you can use a TS namespace-import for most of these. That syntax means you could avoid retyping the various type parameters and constraints. The only time you wouldn't be able to use it are cases where there's a *value* with the same name.

```suggestion
    export import Evolve = Tools.Evolve;
```

This kind of change isn't necessary for me to approve this, so if you'd rather leave this as-is let me know and I can merge.

pirix-gh: I forgot to mention that I tried this in the beginning and that `declare let R: R.Static;` is blocking here. I believe this is kind of a trick to enable importing of a few utilities at a time (eg. `import {map} from "ramda"`) while keeping an interface (for convenience, I guess).

pirix-gh:
So I worked on flattening the declaration, which removed legacy namespace & `Static` interface.

types/ramda/index.d.ts
```diff
@@ -33,3136 +33,2110 @@
...
+export function join(x: string, xs: readonly any[]): string;
+export function join(x: string): (xs: readonly any[]) => string;
+
+/**
+ * Applies a list of functions to a list of values.
+ */
+export function juxt<A extends any[], R1, R2>(fns: [(...a: A) => R1, (...a: A) => R2]): (...a: A) => [R1, R2];
+export function juxt<A extends any[], R1, R2, R3>(fns: [(...a: A) => R1, (...a: A) => R2, (...a: A) => R3]): (...a: A) => [R1, R2, R3];
+export function juxt<A extends any[], R1, R2, R3, R4>(fns: [(...a: A) => R1, (...a: A) => R2, (...a: A) => R3, (...a: A) => R4]): (...a: A) => [R1, R2, R3, R4];
+export function juxt<A extends any[], R1, R2, R3, R4, R5>(fns: [(...a: A) => R1, (...a: A) =>
```
rbuckton: I don't see an `export as namespace R` in the updated code.

delanni:
@pirix-gh 
I just want to contest the statement ` * => ramda's `merge` functions are 100% properly typed.` in the tools.d.ts file around Merge. They're not a 100% properly typed, just after this cleanup, `mergeDeepRight` stopped accepting proper right mergeable types. 

My best guess is the type for `mergeDeepRight` changed from `O.Ensure<O.MergeUp<O2, O1, 'deep'>>` to Tools/Merge which is `O.MergeUp<T.ObjectOf<O1>, T.ObjectOf<O2>, Depth>;` which I think the type for `mergeDeepRight` was before another fix (https://github.com/DefinitelyTyped/DefinitelyTyped/pull/38683)

Is it possible that you had some oversight and went back to a less accurate type while moving stuff into the tools?

pirix-gh:
@delanni

I think what you are complaining about is the display of the type that is not as neat as before? Or would you mind providing an example?

delanni:
I wasn't complaining, I just called out, that you apparently changed the type definition for `mergeDeepRight` when you moved it to the cleaned-up format.
Earlier it used to be:
```
        mergeDeepRight<O1 extends object, O2 extends object>(o1: O1, o2: O2): O.Ensure<O.MergeUp<O2, O1, 'deep'>>;
```
Now it is
```
export function mergeDeepRight<O1 extends object, O2 extends object>(o1: O1, o2: O2): Merge<O2, O1, 'deep'>;
// Where Merge comes from tools.d.ts, and it is:
export type Merge<O1 extends object, O2 extends object, Depth extends 'flat' | 'deep'> =
    O.MergeUp<T.ObjectOf<O1>, T.ObjectOf<O2>, Depth>;
```
The version with `O.Ensure<O.MergeUp...` version was introduced in (#38683) - moving away from the version that's now brought back.

I just found this interesting, and a potential cause for some type errors in my our codebase with the latest and greatest typings.

pirix-gh:
@delanni would you mind posting a simple repro where this fails?

delanni:
I can't bring a simple case. Just consider why you'd changed it in the previous case and whether the changing back is intentional.

pirix-gh:
Well, the reason of the change is that ramda discards array props when merging.

https://ramdajs.com/repl/#?R.mergeDeepLeft%28%5B1%5D%2C%20%5B1%2C%202%5D%29

This is what `O.MergeUp<T.ObjectOf<O1>, T.ObjectOf<O2>, Depth>` does.

`T.ObjectOf<*>` takes any object, and if it is really an array it plucks out its array properties, thus making it a simple object (like ramda does).

`O.MergeUp<*, *>` does the mergings like the `...` operator but deeply.

delanni:
All right, then not using O.Ensure should be replaced by ObjectOf for the two merged types, and the change should be equivalent - but ultimately something is not right, but then I'll try to pinpoint what's not quite right over the weekend.
Thanks

PR #40739, opened 2019-11-29
[parquetjs] optional/repeated are always usable

dobesv:
The library will accept optional on a struct (with fields) and
repeated on a normal field (without fields).

For example, a struct could be optional, or you can have an array
of integers.


jon-fearer (approved):
Yeah this makes total sense. Oddly I never received an email to review 5 days ago, otherwise would have reviewed it then.

PR #40742, opened 2019-11-30
[mongoose] export Document class, add DocumentQuery.map()

lkho:
Reopen #38999


types/mongoose/index.d.ts
```diff
@@ -3322,6 +3329,7 @@ declare module "mongoose" {
     where(path: string, val?: any): Query<any> & QueryHelpers;
   }
 
+  class Document {}
```
DanielRosenwasser: Can you add a test for instantiating `Document?

lkho: added

andrewbranch (approved):

PR #40745, opened 2019-11-30
[graphql-schema-linter] Add type definitions for graphql-schema-linter

gagoar:


types/graphql-schema-linter/lib/configuration/index.d.ts
```diff
@@ -0,0 +1,8 @@
+import { ValidationContext, ASTKindToNode, Visitor, GraphQLError } from 'graphql';
```
rbuckton: In the source, there is no `configuration/index.js`, but rather just `configuration.js`. Since an import of `graphql-schema-linter/lib/configuration/index` would fail at runtime, I would recommend moving/renaming this file to `lib/configuration.d.ts`.

rbuckton: The same seems to be true of *every* file under `lib/**`.

types/graphql-schema-linter/lib/validator/index.d.ts
```diff
@@ -0,0 +1,7 @@
+import { Configuration, Rule } from '../configuration';
+import { ValidationError } from '../validation_error';
+export function validateSchemaDefinition(
+  schemaDefinition: string,
+  rules: ReadonlyArray<Rule>,
+  configuration: InstanceType<typeof Configuration>,
```
rbuckton:
```suggestion
  configuration: Configuration,
```

types/graphql-schema-linter/lib/validator/index.d.ts
```diff
@@ -0,0 +1,7 @@
...
+): ReadonlyArray<InstanceType<typeof ValidationError>>;
```
rbuckton:
```suggestion
): ReadonlyArray<ValidationError>;
```

rbuckton (changes requested):
All of the files under `lib/**` are incorrectly named `lib/**/<module>/index.d.ts` when they should be `lib/**/<module>.d.ts`.

types/graphql-schema-linter/lib/configuration/index.d.ts
```diff
@@ -0,0 +1,8 @@
...
+
+export type Rule = (context: ValidationContext) => Visitor<ASTKindToNode>;
+export class Configuration {
+  constructor(options: GraphqlSchemaLinter.Configuration.Options, stdinFd: string | null);
+  getRules(): ReadonlyArray<Rule>;
```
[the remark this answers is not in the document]
gagoar: Good question! 
I've typed the methods that are often used as `public methods`, every other method in the Configuration class are used within itself.

types/graphql-schema-linter/index.d.ts
```diff
@@ -0,0 +1,18 @@
+// Type definitions for graphql-schema-linter 0.2
+// Project: https://github.com/cjoudrey/graphql-schema-linter
+// Definitions by: Gago <https://github.com/gagoar>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare namespace GraphqlSchemaLinter {
```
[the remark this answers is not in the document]
gagoar: I wrote it there, because I was reusing it in other types. I'm correcting this, and avoiding the export, and just have the interface to use inside the `Configuration.d.ts`. 
leme know if there's a better or more cannonical way to do it.

gagoar:
Hi! any news with the necessary review for this PR? ping @rbuckton

types/graphql-schema-linter/index.d.ts
```diff
@@ -0,0 +1,15 @@
+// Type definitions for graphql-schema-linter 0.2
+// Project: https://github.com/cjoudrey/graphql-schema-linter
+// Definitions by: Gago <https://github.com/gagoar>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import { Configuration } from './lib/configuration';
+import { validateSchemaDefinition } from './lib/validator';
+
+export interface configuration {
```
sheetalkamat: This needs to be value and not interface based on https://github.com/cjoudrey/graphql-schema-linter/blob/402b0a6dd610b00ba346c422e8558dd4dc15b634/src/index.js#L3

sheetalkamat (changes requested):

PR #40753, opened 2019-12-01
[react-syntax-highlighter] Fix lineProps and codeTagProps

michalciechan:
Previously good code wouldn't compile because both `props` used
definitions without `React.HTMLProps` which is the correct definition for
the `props object` mentioned in https://github.com/conorhastings/react-syntax-highlighter#props.

All `lineProps` and `codeTagProps` tests have now explicitly defined
`prop` type.

Also code is now formatted with prettier.

### DIFF
--- a/types/react-syntax-highlighter/index.d.ts
+++ b/types/react-syntax-highlighter/index.d.ts
@@ -7,17 +7,15 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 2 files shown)


masterzenox:
@NoHomey, @guoyunhe, @anirban09 take a look please (bot didn't mention you).

guoyunhe (changes requested):
Please don't introduce unnecessary auto-format. You can disable Prettier or similar extensions. This PR is not valid for review.

guoyunhe (approved):
LGTM

PR #40754, opened 2019-12-01
Types for existing npm package `subleveldown` with tests

carsonfarmer:

If adding a new definition:

### DIFF
--- a/types/subleveldown/index.d.ts
+++ b/types/subleveldown/index.d.ts
@@ -5,12 +5,14 @@
 
 import { LevelUp } from 'levelup';
 import { AbstractLevelDOWN, AbstractIterator, ErrorCallback } from 'abstract-leveldown';
+import { CodecOptions } from 'level-codec';
 
 type SubDownOpenHook = (callback: ErrorCallback) => void;
 
-interface SubDownOptions {
+interface SubDownOptions extends CodecOptions {
     separator?: string;
     open?: SubDownOpenHook;
+    // TODO: Remove and inherit from constructor options from levelup package
     [key: string]: any;
 }
 
--- a/types/subleveldown/index.d.ts
+++ b/types/subleveldown/index.d.ts
@@ -19,7 +19,7 @@ declare namespace sub {
         ((options?: O) => Promise<V>);
 
     type InferDBClear<DB> =
-        DB extends { get: (options: infer O, callback: ErrorValueCallback<infer V>) => void } ?
+        DB extends { clear: (options: infer O, callback: ErrorValueCallback<infer V>) => void } ?
         LevelUpClear<V, O> :
         LevelUpClear<any, AbstractGetOptions>;
 
--- a/types/subleveldown/index.d.ts
+++ b/types/subleveldown/index.d.ts
@@ -13,8 +13,14 @@ declare namespace sub {
         // Any other options are passed along to the underlying levelup and encoding-down constructors.
         [key: string]: any;
     }
+
+    class SubIterator<K = any, V = any> extends AbstractIterator<K, V> {
+        iterator: AbstractIterator<K, V>;
+        prefix: string;
+        constructor(db: LevelUp<AbstractLevelDOWN<K, V>, AbstractIterator<K, V>>, ite: AbstractIterator<K, V>, prefix: string);
+    }
 }
 
-declare function sub<K = any, V = any>(db: LevelUp, prefix?: string, opts?: sub.Options | string): LevelUp<AbstractLevelDOWN<K, V>, AbstractIterator<K, V>>;
+declare function sub<K = any, V = any>(db: LevelUp, prefix?: string, opts?: sub.Options | string): LevelUp<AbstractLevelDOWN<K, V>, sub.SubIterator<K, V>>;
 
 export = sub;
--- a/types/subleveldown/index.d.ts
+++ b/types/subleveldown/index.d.ts
@@ -6,10 +6,11 @@
 import { LevelUp } from 'levelup';
 import { AbstractLevelDOWN, AbstractIterator, ErrorCallback } from 'abstract-leveldown';
 
+type SubDownOpenHook = (callback: ErrorCallback) => void;
+
 interface SubDownOptions {
     separator?: string;
-    open?: (callback: ErrorCallback) => void;
-    // Any other options are passed along to the underlying levelup and encoding-down constructors.
+    open?: SubDownOpenHook;
     [key: string]: any;
 }
 
--- a/types/subleveldown/index.d.ts
+++ b/types/subleveldown/index.d.ts
@@ -1,6 +1,7 @@
 // Type definitions for subleveldown 4.1
 // Project: https://github.com/level/subleveldown
 // Definitions by: Carson Farmer <https://github.com/carsonfarmer>
+//                 Dmitry Demensky <https://github.com/demensky>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 import { LevelUp } from 'levelup';
--- a/types/subleveldown/index.d.ts
+++ b/types/subleveldown/index.d.ts
@@ -6,21 +6,13 @@
 import { LevelUp } from 'levelup';
 import { AbstractLevelDOWN, AbstractIterator, ErrorCallback } from 'abstract-leveldown';
 
-declare namespace sub {
-    interface Options {
-        separator?: string;
-        open?: ((callback: ErrorCallback) => void);
-        // Any other options are passed along to the underlying levelup and encoding-down constructors.
-        [key: string]: any;
-    }
-
-    class SubIterator<K = any, V = any> extends AbstractIterator<K, V> {
-        iterator: AbstractIterator<K, V>;
-        prefix: string;
-        constructor(db: LevelUp<AbstractLevelDOWN<K, V>, AbstractIterator<K, V>>, ite: AbstractIterator<K, V>, prefix: string);
-    }
+interface SubDownOptions {
+    separator?: string;
+    open?: ((callback: ErrorCallback) => void);
+    // Any other options are passed along to the underlying levelup and encoding-down constructors.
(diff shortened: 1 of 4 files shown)


types/subleveldown/index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+// Definitions by: Carson Farmer <https://github.com/carsonfarmer>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import { LevelUp } from 'levelup';
+import { ErrorValueCallback, ErrorCallback, AbstractGetOptions } from 'abstract-leveldown';
+
+declare namespace sub {
+    interface Options {
+        separator?: string;
+        open?: ((callback?: ErrorCallback) => any);
```
demensky: Please Use `void` as [return type of callback][callback].

[callback]: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#return-types-of-callbacks

types/subleveldown/index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+        DB extends { get: (options: infer O, callback: ErrorValueCallback<infer V>) => void } ?
+        LevelUpClear<V, O> :
+        LevelUpClear<any, AbstractGetOptions>;
+
+    interface SublevelDown<DB, Iterator> extends LevelUp<DB, Iterator> {
+        clear: InferDBClear<DB>;
+    }
+}
+
+declare function sub<DB, Iterator>(db: LevelUp < DB, Iterator >, prefix ?: string, options ?: sub.Options): sub.SublevelDown<DB, Iterator>;
```
demensky: [`opts` may be a `string`](https://github.com/Level/subleveldown/blob/master/leveldown.js#L48).

types/subleveldown/index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import { LevelUp } from 'levelup';
+import { ErrorValueCallback, ErrorCallback, AbstractGetOptions } from 'abstract-leveldown';
+
+declare namespace sub {
+    interface Options {
+        separator?: string;
+        open?: ((callback?: ErrorCallback) => any);
+        [key: string]: any;
```
demensky: The source code uses only [separator] and [open]. Why is it here?

[separator]: https://github.com/Level/subleveldown/blob/master/leveldown.js#L51

[open]: https://github.com/Level/subleveldown/blob/master/leveldown.js#L62

types/subleveldown/index.d.ts
demensky: [Why is callback listed as optional?][optional-parameters-in-callbacks] In [source code][_open], it is always passed.

[optional-parameters-in-callbacks]: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#optional-parameters-in-callbacks
[_open]: https://github.com/Level/abstract-leveldown/blob/52d0179e272f2f5c0420f3e24415510aa63e042b/abstract-leveldown.js#L34

demensky (changes requested):

types/subleveldown/index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+        DB extends { clear: (options: infer O, callback: ErrorValueCallback<infer V>) => void } ?
+        LevelUpClear<V, O> :
+        LevelUpClear<any, AbstractGetOptions>;
+
+    interface SublevelDown<DB, Iterator> extends LevelUp<DB, Iterator> {
+        clear: InferDBClear<DB>;
+    }
+}
+
+declare function sub<DB, Iterator>(db: LevelUp < DB, Iterator >, prefix ?: string, options ?: sub.Options): sub.SublevelDown<DB, Iterator>;
```
demensky: `SubIterator` is actually a class. I suggest using functions and ambient classes. [Example].

[Example]: https://www.typescriptlang.org/play/#code/MYGwhgzhAECCBGEAuAnMwkEkkFM1IHsUAeAEQCEAaabPMQlAPmgG8BfAWAChuBLAO1woAZuhzQAMjgBuOEAFUADmSo0h9Is3Z9BeUcHEB5RUl4F+MbTy4B6G9AC0T5y4fduAExygwKcT6hoAGUAV3hafCIVagiNJmgcAA9cfg8YBGQ0DFiGaLU6BkYWbmhS6CQAW0UALmgwfgBPAG53LjLoRTCQXmBoYHNMkIwiAAoPeFqpWQVFaDyczWpFP2FeRIB+WsyBAHNqAhMzC2hN6GNTAYBKFq5Oay8fP2hhEP4MI+CwhZIKGPVCsYTSQyORKeb-RYdFZrU7bfh7aAHC4WU7nI4QS61ULhCE-VTfRg3bj9CxIco4ZAARmgAF5oPwcAB3T44gqjdjXaB2FnfVok5Dk5AAJlpPNxIw5TS59mxvOsuCpADpKooiVwFUghcqqk0gA

demensky (changes requested):

carsonfarmer:
Thank you @demensky for your thorough and helpful review! I've addressed most of your comments, and have one open question left for some clarification.

types/subleveldown/index.d.ts
carsonfarmer: Done

types/subleveldown/index.d.ts
carsonfarmer: Good catch, added!

types/subleveldown/index.d.ts
carsonfarmer: That was a relic of some old work, thanks for catching! Fixed.

types/subleveldown/index.d.ts
carsonfarmer: Good point, and thanks for the link! Fixed.

types/subleveldown/index.d.ts
carsonfarmer: This sounds reasonable. In this case, do you mean the main `SubDown` function/class, or the 'internal' `SubIterator`? In the source code, `SubIterator` is not exported, though it obviously is returned from the `iterator` method. Are you suggesting:
- [ ] Add this `SubIterator` class to the definition as well?
- [ ] Simply implement the main `SubDown` using a function and ambient class?
- [ ] Both of these?

Note that I updated the name of the main interface (`SubDown`) to better match the source code.

carsonfarmer: Either way, your suggestion is a good one for the main class/function/interface. So I will certainly add this.

types/subleveldown/index.d.ts
```diff
@@ -4,27 +4,18 @@
...
-
-    type InferDBClear<DB> =
-        DB extends { clear: (options: infer O, callback: ErrorValueCallback<infer V>) => void } ?
-        LevelUpClear<V, O> :
-        LevelUpClear<any, AbstractGetOptions>;
-
-    interface SubDown<DB, Iterator> extends LevelUp<DB, Iterator> {
-        clear: InferDBClear<DB>;
-    }
+    interface SubDown<DB, Iterator> extends LevelUp<DB, Iterator> {}
```
carsonfarmer: Removed clear definitions in favour of a separate PR to levelup (where they actually belong): https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40792

carsonfarmer (commented):
I hope these updates are 'ok' (they are a departure from the original submission). Additional changes are also expected upon answering one remaining question.

carsonfarmer:
More updates, I'm still new to some aspects of Typescript defs, so I hope this is now pretty clear.

The most recent updates allow the types to reflect arbitrary key/value types in each sublevel, which is an explicit requirement of the package: https://github.com/Level/subleveldown#background. The tests have been updated to explicitly check for this.

types/subleveldown/index.d.ts
```diff
@@ -0,0 +1,26 @@
+// Type definitions for subleveldown 4.1
+// Project: https://github.com/level/subleveldown
+// Definitions by: Carson Farmer <https://github.com/carsonfarmer>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import { LevelUp } from 'levelup';
+import { AbstractLevelDOWN, AbstractIterator, ErrorCallback } from 'abstract-leveldown';
+
+declare namespace sub {
```
carsonfarmer: Is this the correct use of a namespace here?

carsonfarmer: As in, should this be named `SubDown`?

demensky: I think not.  In the [source code](https://github.com/Level/subleveldown/blob/master/index.js#L5), a function is exported, not a class. The class instance only returns from the function.

I probably made some unintentional mistakes in some comments because I got confused in the source code of the library.

I can review this PR in more detail and thoughtfully. What do you say about this?

carsonfarmer: Yes please, and not a problem at all!

carsonfarmer:
Thanks again to @demensky! I look forward to your (hopefully final) review, and am happy to make any additional changes as needed. I suspect we can actually simplify things a great deal here.

carsonfarmer:
Apologies to anyone following this work, I accidentally pushed a ton of commits to this branch. I've now force-pushed the the branch from 98f2143 to 3ba1bbe... hopefully this doesn't cause you any issues!

types/subleveldown/index.d.ts
demensky: I suggest such improvements: carsonfarmer/DefinitelyTyped#1.

carsonfarmer:
Ok, I'm happy with the final product here @demensky! I've made one last commit to add your name to the header, I hope this is ok? Otherwise, I think we're good to go here?

demensky (approved):

carsonfarmer:
I think this one is ready to go. It seems like the Revision needed label got left on for some reason?

PR #40756, opened 2019-12-01
fix(airtable): return types for Table.create

Merott:
When `create` is called with a single object to create, the return type is not an array, but a single object. `Response<TFields>` is defined as `ReadonlyArray<Row<TFields>>` (note: `Array`), which isn't correct for a single object. Likewise, `Promise<Array<Response<TFields>>>` implies a multidimensional array, which again isn't correct.

Please fill in this template.

### DIFF
--- a/types/airtable/index.d.ts
+++ b/types/airtable/index.d.ts
@@ -42,8 +42,8 @@ declare global {
         interface Table<TFields extends FieldSet> {
             select(opt?: SelectOptions): Query<TFields>;
             find(id: string): Promise<Response<TFields>>;
-            create(record: TFields, opts?: { typecast: boolean }): Promise<Response<TFields>>;
-            create(records: TFields[], opts?: { typecast: boolean }): Promise<Array<Response<TFields>>>;
+            create(record: TFields, opts?: { typecast: boolean }): Promise<ResponseSingle<TFields>>;
+            create(records: TFields[], opts?: { typecast: boolean }): Promise<Response<TFields>>;
             update(...args: any[]): Promise<any>;
             replace(...args: any[]): Promise<any>;
             destroy(...args: any[]): Promise<any>;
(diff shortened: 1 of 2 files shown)


bvalosek (approved):

types/airtable/index.d.ts
```diff
@@ -73,6 +73,7 @@ declare global {
         }
 
         type Response<TFields> = ReadonlyArray<Row<TFields>>;
+        type ResponseSingle<TFields> = Readonly<Row<TFields>>;
```
rbuckton: Why is the single response a `Readonly<Row<...>>`, but an array response is only a `Row<...>`? `ReadonlyArray` is not recursively read-only, it only makes the *array* readonly, not its contents.

types/airtable/index.d.ts
```diff
@@ -41,9 +41,9 @@ declare global {
 
         interface Table<TFields extends FieldSet> {
             select(opt?: SelectOptions): Query<TFields>;
-            find(id: string): Promise<Response<TFields>>;
-            create(record: TFields, opts?: { typecast: boolean }): Promise<Response<TFields>>;
-            create(records: TFields[], opts?: { typecast: boolean }): Promise<Array<Response<TFields>>>;
+            find(id: string): Promise<Record<TFields>>;
+            create(record: TFields, opts?: { typecast: boolean }): Promise<Record<TFields>>;
```
Merott: `find(id)` and `create(record)` both return a single record.

types/airtable/index.d.ts
Merott: You're right. I updated the interface names and the definitions as I think it was confusing before.

rbuckton (approved):

rbuckton:
Thanks for the contribution!

PR #40757, opened 2019-12-01
Added negotiateLocale(key) to angular-translate

yrtimiD:
Adds missing method.
From [official documentation](https://angular-translate.github.io/docs/#/api/pascalprecht.translate.$translate) scroll to negotiateLocale([key])
or [source code](https://github.com/angular-translate/angular-translate/blob/aafbc39/src/service/translate.js#L1883)
JSDocs are from original lib code
Exists in 2.16 (no need to change version)

### DIFF
--- a/types/angular-translate/index.d.ts
+++ b/types/angular-translate/index.d.ts
@@ -69,8 +69,8 @@ declare module 'angular' {
              *
              * If no or a falsy key is given, returns undefined.
              *
-             * @param {string} [key] Language key
-             * @return {string|undefined} Language key or undefined if no language key is found.
+             * @param [key] Language key
+             * @return Language key or undefined if no language key is found.
              */
             negotiateLocale(key?: string): string | undefined;
             preferredLanguage(langKey?: string): string;
--- a/types/angular-translate/index.d.ts
+++ b/types/angular-translate/index.d.ts
@@ -58,9 +58,19 @@ declare module 'angular' {
             instant(translationId: string, interpolateParams?: any, interpolationId?: string, forceLanguage?: string, sanitizeStrategy?: string): string;
             instant(translationId: string[], interpolateParams?: any, interpolationId?: string, forceLanguage?: string, sanitizeStrategy?: string): { [key: string]: string };
             isPostCompilingEnabled(): boolean;
-            /** Returns a language key based on available languages and language aliases. If a language key cannot be resolved, returns undefined.
+            /**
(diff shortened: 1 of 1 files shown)


types/angular-translate/index.d.ts
```diff
@@ -56,6 +58,21 @@ declare module 'angular' {
...
+             * @methodOf pascalprecht.translate.$translate
+             *
+             * @description
+             * Returns a language key based on available languages and language aliases. If a
+             * language key cannot be resolved, returns undefined.
+             *
+             * If no or a falsy key is given, returns undefined.
+             *
+             * @param {string} [key] Language key
+             * @return {string|undefined} Language key or undefined if no language key is found.
```
demensky: The types in the JSDoc in `d.ts` are overengineering. I propose to simplify.
```
             * @param key Language key
             * @return Language key or undefined if no language key is found.
```

demensky (changes requested):

types/angular-translate/index.d.ts
yrtimiD: That's just a copy-paste from the original code. But agree, it's repetition :)

yrtimiD: Removed

demensky: @yrtimiD you forgot about brackets (`[]`).

demensky: @yrtimiD you forgot about brackets (`[]`).

types/angular-translate/index.d.ts
```diff
@@ -56,6 +58,21 @@ declare module 'angular' {
...
+             * @name pascalprecht.translate.$translate#negotiateLocale
+             * @methodOf pascalprecht.translate.$translate
+             *
+             * @description
+             * Returns a language key based on available languages and language aliases. If a
+             * language key cannot be resolved, returns undefined.
+             *
+             * If no or a falsy key is given, returns undefined.
+             *
+             * @param [key] Language key
```
GabrielGil: I think what @demensky meant was for you to remove the brackets that in JSDoc scope mean optional, but in typescript that info is already in the signature.

```suggestion
             * @param key Language key
```

GabrielGil (changes requested):

demensky (approved):

GabrielGil (approved):

orta:
👍 Looks good

PR #40765, opened 2019-12-02
Added jqwidgets TypeScript definitions

DimitarD:
Please fill in this template.

If adding a new definition:


orta:
Hi, given that this comes form the jqwidgets team is there a reason why the d.ts files are not shipped with the library and then will be accurate to the version on npm?

DimitarD:
The d.ts file is actually included in our npm package (jqwidgets-scripts). However, we would like for it to be featured here too, this repository being an exhaustive list of TypeScript type definitions.

types/jqwidgets/index.d.ts
```diff
@@ -0,0 +1,4896 @@
+// Type definitions for jqwidgets-scripts
```
DanielRosenwasser: Looks like our lint rules are disabled for this package. Can you enable the lints in our `tslint.json` and fix up the version string to be a `MAJOR.MINOR` version (e.g. `8.3`)?

Also, it looks like this containing folder was incorrectly named. jqwidgets doesn't exist on npm, but jqwidgets-scripts does.

DanielRosenwasser (changes requested):

DanielRosenwasser:
> The d.ts file is actually included in our npm package (jqwidgets-scripts). However, we would like for it to be featured here too, this repository being an exhaustive list of TypeScript type definition

I see; however, that's not the purpose of DefinitelyTyped. DefinitelyTyped exists to fill the gap for packages that don't contain `.d.ts` files. Thanks for taking the time though.

PR #40776, opened 2019-12-02
[connect] Add missing property originalUrl to req parameter

trygveaa:
This adds a new class IncomingMessage which extends from
http.IncomingMessage and adds the originalUrl property which was
previously missing. This new class is used for the req parameter in
HandleFunction.

The originalUrl property is added to req here:
https://github.com/senchalabs/connect/blob/3.4.0/index.js#L133

`originalUrl` is present in the version specified in these type definitions, so I didn't change the version.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/senchalabs/connect/blob/3.4.0/index.js#L133

### DIFF
--- a/types/connect/index.d.ts
+++ b/types/connect/index.d.ts
@@ -18,11 +18,15 @@ declare function createServer(): createServer.Server;
 declare namespace createServer {
     export type ServerHandle = HandleFunction | http.Server;
 
+    export class IncomingMessage extends http.IncomingMessage {
+        originalUrl?: http.IncomingMessage["url"];
+    }
+
     type NextFunction = (err?: any) => void;
 
-    export type SimpleHandleFunction = (req: http.IncomingMessage, res: http.ServerResponse) => void;
-    export type NextHandleFunction = (req: http.IncomingMessage, res: http.ServerResponse, next: NextFunction) => void;
-    export type ErrorHandleFunction = (err: any, req: http.IncomingMessage, res: http.ServerResponse, next: NextFunction) => void;
+    export type SimpleHandleFunction = (req: IncomingMessage, res: http.ServerResponse) => void;
+    export type NextHandleFunction = (req: IncomingMessage, res: http.ServerResponse, next: NextFunction) => void;
+    export type ErrorHandleFunction = (err: any, req: IncomingMessage, res: http.ServerResponse, next: NextFunction) => void;
     export type HandleFunction = SimpleHandleFunction | NextHandleFunction | ErrorHandleFunction;
 
     export interface ServerStackItem {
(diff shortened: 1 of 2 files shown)


EvanHahn (changes requested):
Rejecting because I have a question:

Should we test that a request handler still works with `http.IncomingMessage`? For example, should we have a test like this?

```ts
app.use((req: http.IncomingMessage, res: http.ServerResponse) => {
  console.log(req, res);
  res.end();
});

**Comment by trygveaa** [comment]
@EvanHahn: Not sure if it's a question for me, or another maintainer. But sure, I can add that test if you want. I agree that it's a good idea.

**Comment by trygveaa** [comment]
@EvanHahn: I've added the test.

File: types/connect/index.d.ts [hunk 1]
```diff
@@ -18,11 +18,15 @@ declare function createServer(): createServer.Server;
 declare namespace createServer {
     export type ServerHandle = HandleFunction | http.Server;
 
+    export class IncomingMessage extends http.IncomingMessage {
+        originalUrl?: http.IncomingMessage["url"];
```
EvanHahn: I'm not a TypeScript expert. What does this line do?

EvanHahn (approved):

types/connect/index.d.ts
trygveaa: `http.IncomingMessage["url"]` is a reference to the type of the `url` property in `http.IncomingMessage`, so it's setting the type of `originalUrl` to be the same as the type of `url`. It's called lookup types, [documented here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html#keyof-and-lookup-types).

orta:
Looks good and thanks for the test to verify existing code doesn't break 👍

types/connect/index.d.ts
```diff
@@ -18,11 +18,15 @@ declare function createServer(): createServer.Server;
 declare namespace createServer {
     export type ServerHandle = HandleFunction | http.Server;
 
+    export class IncomingMessage extends http.IncomingMessage {
```
siraajahmed: This should be "implements" as http.IncomingMessage is an interface.

My code will no longer compile in typescript 3.7.2 because of this issue.

Full error is "node_modules/@types/connect/index.d.ts(21,42): error TS2689: Cannot extend an in
terface 'http.IncomingMessage'. Did you mean 'implements'?"

Please can we fix this ASAP.

trygveaa: According to [the types in this repo](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/http.d.ts#L269), `http.IncomingMessage` is a class. Where are you getting your types from?

siraajahmed: Ah thanks for that. Turns out we were using an older version of node @types where the `http.IncomingMessage` was an interface. Thanks for your prompt response

PR #40779, opened 2019-12-02
[react-virtualized] Correcting Table.noRowsRenderer property function return type

Ghary:
`noRowsRenderer` takes a function which will return an element to render when no rows are found. The default value for this prop is `() => null` which also does not fit the schema of `() => void`.

Here is the line of code passing `noRowsRenderer` to the underlying `Grid`:
https://github.com/bvaughn/react-virtualized/blob/ab36d92cb36428cfee4b733f35dbe6a191168c6e/source/Table/Table.js#L438

Here is the `noRowsRenderer` function passed in being executed to render an element:
https://github.com/bvaughn/react-virtualized/blob/master/source/Collection/CollectionView.js#L411

Tests showing it should return an element:
https://github.com/bvaughn/react-virtualized/blob/ab36d92cb36428cfee4b733f35dbe6a191168c6e/source/Grid/Grid.jest.js#L832

And the types.js file with the correct type:
https://github.com/bvaughn/react-virtualized/blob/8738097a6ead422f74590d6a631bd5fd83ebc831/source/Grid/types.js#L53

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: see above

### DIFF
--- a/types/react-virtualized/dist/es/Table.d.ts
+++ b/types/react-virtualized/dist/es/Table.d.ts
@@ -229,7 +229,7 @@ export type TableProps = GridCoreProps & {
     /** Optional id */
     id?: string;
     /** Optional renderer to be used in place of table body rows when rowCount is 0 */
-    noRowsRenderer?: () => JSX.Element;
+    noRowsRenderer?: () => JSX.Element | null;
     /**
      * Optional callback when a column's header is clicked.
      * ({ columnData: any, dataKey: string }): void
--- a/types/react-virtualized/dist/es/Table.d.ts
+++ b/types/react-virtualized/dist/es/Table.d.ts
@@ -229,7 +229,7 @@ export type TableProps = GridCoreProps & {
     /** Optional id */
     id?: string;
     /** Optional renderer to be used in place of table body rows when rowCount is 0 */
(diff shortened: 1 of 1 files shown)


Stevearzh (approved):
LGTM

Stevearzh (approved):

types/react-virtualized/dist/es/Table.d.ts
```diff
@@ -229,7 +229,7 @@ export type TableProps = GridCoreProps & {
     /** Optional id */
     id?: string;
     /** Optional renderer to be used in place of table body rows when rowCount is 0 */
-    noRowsRenderer?: () => void;
+    noRowsRenderer?: () => JSX.Element;
```
rbuckton: If the default is `() => null`, wouldn't this be more correct when used with `--strictNullChecks`?

```suggestion
    noRowsRenderer?: () => JSX.Element | null;
```

Ghary: Good feedback! I've committed your suggested change. TY

rbuckton (approved):

rbuckton:
Thanks for the contribution!

PR #40784, opened 2019-12-02
custom values for marker option @types/leaflet

soupay:
Adding type support for customValues to be passed in MarkerOptions for different types of custom functionality

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/DefinitelyTyped/DefinitelyTyped/issues/40787>>

### DIFF
--- a/types/leaflet/index.d.ts
+++ b/types/leaflet/index.d.ts
@@ -1884,7 +1884,7 @@ export interface MarkerOptions extends InteractiveLayerOptions {
     /** Number of pixels the map should pan by. */
     autoPanSpeed?: number;
     /** Optional object to pass in values as when required for custom functionality */
-    customValues?: Object;
+    customValues?: object;
 }
 
 export class Marker<P = any> extends Layer {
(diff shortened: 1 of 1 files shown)


ronikar (commented):
I didn't find `customValue` in leaflet documentation.
Can you add a reference to it in leaflet code?

soupay:
> I didn't find `customValue` in leaflet documentation.
> Can you add a reference to it in leaflet code?

Yeah I figured its not an existing leaflet functionality. Let me talk to them to get it included as a parameter.

ronikar:
You should do it in your types, locally

henrythasler:
While I can understand that one might need something like that, I don't think it should be included in the official typings as it is unrelated to the current leaflet code.

henrythasler (commented):
While I can understand that one might need something like that, I don't think it should be included in the official typings as it is unrelated to the current leaflet code.

sandersn (changes requested):
Sounds like the type authors have some concerns about adding this. Requesting changes so the bot can understand the state.

henrythasler:
I don't know how to tell that to the typescript-bot, but I'd rather reject this PR.

PR #40791, opened 2019-12-03
Add types for change-case-object

jameschao:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/change-case-object/index.d.ts
+++ b/types/change-case-object/index.d.ts
@@ -4,25 +4,25 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 export function camelCase(s: string): string;
-export function camelCase<T>(arr: ReadonlyArray<T>): ReadonlyArray<T>;
+export function camelCase<T>(arr: ReadonlyArray<T>): T[];
 export function camelCase(obj: object): object;
 
 export function snakeCase(s: string): string;
-export function snakeCase<T>(arr: ReadonlyArray<T>): ReadonlyArray<T>;
+export function snakeCase<T>(arr: ReadonlyArray<T>): T[];
 export function snakeCase(obj: object): object;
 
 export function paramCase(s: string): string;
-export function paramCase<T>(arr: ReadonlyArray<T>): ReadonlyArray<T>;
+export function paramCase<T>(arr: ReadonlyArray<T>): T[];
 export function paramCase(obj: object): object;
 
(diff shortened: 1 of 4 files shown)


types/change-case-object/index.d.ts
```diff
@@ -0,0 +1,28 @@
+// Type definitions for change-case-object 2.0
+// Project: https://github.com/BinaryThumb/change-case-object#readme
+// Definitions by: James Chao <https://github.com/jameschao>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export function camelCase(s: string): string;
+export function camelCase<T>(arr: ReadonlyArray<T>): ReadonlyArray<T>;
```
demensky: If you pass an array of objects, the result will not be equivalent. Example ([RunKit](https://runkit.com/demensky/5de9bc1d3d6662001a19a5c8)):

```js
const changeCaseObject = require("change-case-object");

changeCaseObject.camelCase([{a_b_c: 'e_f_g'}]); // [{aBC: 'e_f_g'}]
```

I suggest this option:

```suggestion
export function camelCase(arr: ReadonlyArray<string>): ReadonlyArray<string>;
export function camelCase(arr: ReadonlyArray<object>): ReadonlyArray<object>;
```

demensky: Taking a read-only array is a good idea. But why indicate this for an array that is created each time this function is called?

```suggestion
export function camelCase<T>(arr: ReadonlyArray<T>): T[];
```

demensky (changes requested):
I left comments only for `camelCase`, but they are relevant for all functions.

types/change-case-object/index.d.ts
jameschao: Good catch @demensky. That was unintentional.

jameschao: 👍

jameschao:
Thanks for the feedback @demensky and I've fixed the issues you identified. Would you mind taking another look? Thanks.

demensky (approved):

demensky (approved):

orta:
Looks good to me too 👍

jameschao:
Thanks @orta! Would this automatically create and publish a new `@types/changed-case-object` package?

orta:
yes

jameschao:
Thanks! I'm using it in my project now.

PR #40794, opened 2019-12-03
[rdf-js] Add Dataset interface

thewilkybarkid:
Adds `Dataset` and `DatasetFactory` from the [RDF/JS: Dataset specification 1.0](https://rdf.js.org/dataset-spec/).

I didn't add this in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40657 due to the 'experimental' warning, but adding now following https://gitter.im/rdfjs/public?at=5de4fe2ac3d6795b9f0c1564.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://rdf.js.org/dataset-spec/
- ~If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.~
- ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.~

### DIFF
--- a/types/rdf-js/index.d.ts
+++ b/types/rdf-js/index.d.ts
@@ -552,7 +552,7 @@ export interface Dataset<Q extends BaseQuad = Quad> extends DatasetCore<Q> {
      *
      * The stream events `end` and `error` are wrapped in a Promise.
      */
-    import(stream: Stream<Q>): Promise<Dataset<Q>>;
+    import(stream: Stream<Q>): Promise<this>;
 
     /**
      * Returns a new dataset containing alls quads from the current dataset that are also included in the given dataset.
--- a/types/rdf-js/index.d.ts
+++ b/types/rdf-js/index.d.ts
@@ -465,7 +465,7 @@ export interface DatasetCore<Q extends BaseQuad = Quad> {
      * @param object    The optional exact object to match.
      * @param graph     The optional exact graph to match.
      */
-    match(subject?: Term, predicate?: Term, object?: Term, graph?: Term): DatasetCore<Q>;
+    match(subject?: Term, predicate?: Term, object?: Term, graph?: Term): this;
 
     [Symbol.iterator](): Iterator<Q>;
 }
@@ -478,8 +478,6 @@ export interface DatasetCoreFactory<Q extends BaseQuad = Quad> {
 }
 
 export interface Dataset<Q extends BaseQuad = Quad> extends DatasetCore<Q> {
-    match(subject?: Term, predicate?: Term, object?: Term, graph?: Term): Dataset<Q>;
-
     /**
      * Imports the quads into this dataset.
      *
@@ -512,7 +510,7 @@ export interface Dataset<Q extends BaseQuad = Quad> extends DatasetCore<Q> {
     /**
      * Returns a new dataset that contains all quads from the current dataset, not included in the given dataset.
      */
-    difference(other: Dataset<Q>): Dataset<Q>;
+    difference(other: Dataset<Q>): this;
 
     /**
      * Returns true if the current instance contains the same graph structure as the given dataset.
@@ -538,7 +536,7 @@ export interface Dataset<Q extends BaseQuad = Quad> extends DatasetCore<Q> {
      *
      * This method is aligned with Array.prototype.filter() in ECMAScript-262.
      */
-    filter(iteratee: QuadFilterIteratee<Q>): Dataset<Q>;
+    filter(iteratee: QuadFilterIteratee<Q>): this;
 
     /**
      * Executes the provided `iteratee` once on each quad in the dataset.
@@ -557,12 +555,12 @@ export interface Dataset<Q extends BaseQuad = Quad> extends DatasetCore<Q> {
     /**
      * Returns a new dataset containing alls quads from the current dataset that are also included in the given dataset.
      */
-    intersection(other: Dataset<Q>): Dataset<Q>;
+    intersection(other: Dataset<Q>): this;
 
     /**
      * Returns a new dataset containing all quads returned by applying `iteratee` to each quad in the current dataset.
      */
-    map(iteratee: QuadMapIteratee<Q>): Dataset<Q>;
+    map(iteratee: QuadMapIteratee<Q>): this;
 
     /**
      * This method calls the `iteratee` on each `quad` of the `Dataset`. The first time the `iteratee` is called, the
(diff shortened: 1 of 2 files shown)


types/rdf-js/index.d.ts
```diff
@@ -476,3 +476,183 @@ export interface DatasetCoreFactory<Q extends BaseQuad = Quad> {
...
+     * The returned quad can be the given quad or a new one.
+     */
+    map(quad: Q, dataset: Dataset<Q>): Q;
+}
+
+export interface QuadReduceIteratee<Q extends BaseQuad = Quad, A = any> {
+    /**
+     * A callable function that can be executed on an accumulator and quad and returns a new accumulator.
+     */
+    run(accumulator: A, quad: Q, dataset: Dataset<Q>): A;
```
thewilkybarkid: The spec has a `void` return type, as well as saying it 'returns a new accumulator' (the former must be a mistake).

types/rdf-js/index.d.ts
```diff
@@ -476,3 +476,179 @@ export interface DatasetCoreFactory<Q extends BaseQuad = Quad> {
...
 }
+
+export interface Dataset<Q extends BaseQuad = Quad> extends DatasetCore<Q> {
+    /**
+     * Imports the quads into this dataset.
+     *
+     * This method differs from `Dataset.union` in that it adds all `quads` to the current instance, rather than
+     * combining `quads` and the current instance to create a new instance.
+     */
+    addAll(quads: Dataset<Q>|Iterator<Q>): this;
```
thewilkybarkid: Need to check if 'sequence' in the spec refers to a plain array.

types/rdf-js/index.d.ts
```diff
@@ -476,3 +476,179 @@ export interface DatasetCoreFactory<Q extends BaseQuad = Quad> {
...
+     */
+    toString(): string;
+
+    /**
+     * Returns a new `Dataset` that is a concatenation of this dataset and the quads given as an argument.
+     */
+    union(quads: Dataset<Q>): Dataset<Q>;
+}
+
+export interface DatasetFactory extends DataFactory {
```
thewilkybarkid: Is this meant to extend `DataFactory`? `DatasetCoreFactory` doesn't, and https://github.com/rdf-ext/rdf-dataset-indexed/blob/master/index.js doesn't implement it.

types/rdf-js/index.d.ts
vhf: Good catch: https://github.com/rdfjs/dataset-spec/pull/56

types/rdf-js/index.d.ts
rubensworks: Sequence should be an array I think, as it refers to .pop(), .splice(), .shift().

types/rdf-js/index.d.ts
rubensworks: That looks like a typo in the spec (feel free to open a PR for it). I think that should become `extends DatasetCoreFactory`.

thewilkybarkid: :+1: Makes more sense.

rubensworks (approved):
Looks good to me!

types/rdf-js/index.d.ts
thewilkybarkid: https://github.com/rdfjs/dataset-spec/pull/57

thewilkybarkid:
Note to self: don't open draft pull requests here. 😅

thewilkybarkid:
Should be ready now; waiting on a resolution on https://github.com/rdfjs/dataset-spec/pull/57.

Worth another look though @rubensworks.

types/rdf-js/index.d.ts
```diff
@@ -476,3 +476,181 @@ export interface DatasetCoreFactory<Q extends BaseQuad = Quad> {
      */
     dataset(quads?: Q[]): DatasetCore<Q>;
 }
+
+export interface Dataset<Q extends BaseQuad = Quad> extends DatasetCore<Q> {
+    match(subject?: Term, predicate?: Term, object?: Term, graph?: Term): Dataset<Q>;
```
thewilkybarkid: Not aware of a way to have the return type on `DatasetCore.match` be the same as the current type (`this` means the same instance?).

rubensworks: `Dataset<Q>` should be fine.

thewilkybarkid: Reading https://www.typescriptlang.org/docs/handbook/advanced-types.html#polymorphic-this-types, `this` doesn't actually seem to mean the same instance. So can remove this and update `DatasetCore.match` to return `this`.

types/rdf-js/index.d.ts
```diff
@@ -476,3 +476,181 @@ export interface DatasetCoreFactory<Q extends BaseQuad = Quad> {
...
+     *
+     * The logic described in {@link https://rdf.js.org/dataset-spec/#quad-matching|Quad Matching} is applied for each
+     * quad in this dataset to select the quads which will be deleted.
+     *
+     * @param subject   The optional exact subject to match.
+     * @param predicate The optional exact predicate to match.
+     * @param object    The optional exact object to match.
+     * @param graph     The optional exact graph to match.
+     */
+    deleteMatches(subject?: Term, predicate?: Term, object?: Term, graph?: Term): this;
```
thewilkybarkid: Will be updated if https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40821 is approved.

types/rdf-js/index.d.ts
thewilkybarkid: Though the return types on https://rdf.js.org/dataset-spec/#datasetcore-interface is confusing. Took `add`/`delete` to mean it returns the same instance (so `DatasetCore`, `Dataset`, or some other subtype), and `match` to return a new instance of the same type (even though the IDL says `Dataset` and the description 'This method always returns a new `DatasetCore`').

thewilkybarkid: Done in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40794/commits/5b643c79c97291f68f4048091be841e438058897 (could still use confirming that there's no way to separate the two cases).

rbuckton:
@thewilkybarkid Is this PR ready for review?

thewilkybarkid:
@rbuckton I think so. https://github.com/rdfjs/dataset-spec/pull/57 is merged yet, but believe it's ok.

PR #40803, opened 2019-12-03
[PapaParse] Add delimitersToGuess to ParseConfig definition

rbarbazz:
Please fill in this template.

- Not sure what to test here, the property was just missing

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [https://www.papaparse.com/docs#config](url)

### DIFF
--- a/types/papaparse/index.d.ts
+++ b/types/papaparse/index.d.ts
@@ -36,10 +36,12 @@ export const unparse: (data: Array<Object> | Array<Array<any>> | UnparseObject, 
 export const BAD_DELIMETERS: Array<string>;
 
 // The true delimiter. Invisible. ASCII code 30. Should be doing the job we strangely rely upon commas and tabs for.
-export const RECORD_SEP: string;
+declare type RECORD_SEP_TYPE = '';
+export const RECORD_SEP = '';
 
 // Also sometimes used as a delimiting character. ASCII code 31.
-export const UNIT_SEP: string;
+declare type UNIT_SEP_TYPE = '';
+export const UNIT_SEP = '';
 
 // Whether or not the browser supports HTML5 Web Workers. If false, worker: true will have no effect.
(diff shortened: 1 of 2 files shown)


types/papaparse/index.d.ts
```diff
@@ -104,6 +105,7 @@ export interface ParseConfig {
     skipEmptyLines?: boolean | 'greedy'; // default: false
     fastMode?: boolean; // default: undefined
     withCredentials?: boolean; // default: undefined
+    delimitersToGuess?: string[]; // default: [',', '\t', '|', ';', Papa.RECORD_SEP, Papa.UNIT_SEP]
```
rainshen49: Since this option has finite possible values, I would recommend making a separate type so that editors can provide better hints:
```typescript
...
type GuessableDelimiters = ','| '\t'| '|'| ';' | Papa.RECORD_SEP | Papa.UNIT_SEP
...
    delimitersToGuess?: GuessableDelimiters[]
...
```

rainshen49 (changes requested):
Thanks for your contribution! I have a comment in the code. Also you need to add testing here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ca7a684efbb93f7ee7834ce92df461e0c5db6c8f/types/papaparse/papaparse-tests.ts

In a nutshell, you would add a call to `.parse(` with the `delimitersToGuess` option set to some example values. After that, running `tsc` should not incur any errors/warnings, or you can just leave the typescript-bot to do it. Let me know if you need more clarification!

types/papaparse/index.d.ts
rbarbazz: Thank you for your detailed feedback. It's my first ever contribution so I really appreciate. 

I will make the required changes and re-submit.

rainshen49 (approved):
Nice! I'm happy with this one.

orta:
Thanks!

PR #40810, opened 2019-12-04
[styled-components] Allow static properties to be passed through

### LINKED ISSUE #32048: Extending styled-components interface with static properties
I am trying to add properties to the constructor produced by styled components using TypeScript.

What I want to do:

```tsx
type JSXButtonProps = JSX.IntrinsicElements['button'];

export interface ButtonProps extends JSXButtonProps {
  theme?: ThemeBase;
}

const ButtonInner = ({ theme, ...rest }: ButtonProps) => <button {...rest} />;

export const Button = styled(ButtonInner)(
  (props: ButtonProps): CSSEntry => get(props.theme, 'button.extend'),
);

Button.Primary = props => <Button />;
```

Which produces the error:

```
error TS2339: Property 'Primary' does not exist on type 'StyledComponent<FunctionComponent<ButtonProps>, ThemeBase, {}, never>'.
```

What works (casting to any):

```tsx
export const Button = styled(ButtonInner)(
  (props: ButtonProps): CSSEntry => get(props.theme, 'button.extend'),
);

(Button as any).Primary = props => <Button />;
```

I'd like to avoid this, so I looked into re-creating my own `ButtonInterface`, but the typings are so complex, I can't figure out where to start. For instance I see `StyledComponent<...>` but this is a generic type which cannot be extended, so that won't work. I've tried making the type and joining in the additional static class properties, but that hasn't worked either.

My latest attempt has been to do:

```tsx
interface ButtonAliases extends StyledComponentBase<any, any, any, any> {
  Primary: any;
  Secondary: any;
  Tertiary: any;
  Icon: any;
}

export const Button: ButtonAliases = styled(ButtonInner)(...);
```

But this does not work either, I get the error:

```
error TS2322: Type 'StyledComponent<({  theme, ...rest }: ButtonProps) => Element, ThemeBase, {}, never>' is not assignable to type 'ButtonAliases'.
```

Any help or suggestions would be great, thank you!

Lazyuki:
Allows styled components to retain static fields of the original component, which fixes #32048
Right now 
```jsx
class A extends React.Component {}
class B extends React.Component {
    static A = A;
}
const StyledB = styled(B)``;
<StyledB.A />
```
raises an error that says `Property 'A' does not exist on type 'StyledComponent<typeof B, DefaultTheme, {}, never>'` but if this was plain JS it works as intended. 

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: #32048

### DIFF
--- a/types/styled-components/index.d.ts
+++ b/types/styled-components/index.d.ts
@@ -22,6 +22,7 @@ declare global {
 
 import * as CSS from "csstype";
 import * as React from "react";
+import hoistNonReactStatics = require('hoist-non-react-statics');
 
 export type CSSProperties = CSS.Properties<string | number>;
 
@@ -160,7 +161,7 @@ export type StyledComponent<
     A extends keyof any = never
 > = // the "string" allows this to be used as an object key
     // I really want to avoid this if possible but it's the only way to use nesting with object styles...
-    string & StyledComponentBase<C, T, O, A> & { [K in keyof C]: C[K] };
+    string & StyledComponentBase<C, T, O, A> & hoistNonReactStatics.NonReactStatics<C extends React.ComponentType<any> ? C : never>;
 
 export interface StyledComponentBase<
     C extends keyof JSX.IntrinsicElements | React.ComponentType<any>,
--- a/types/styled-components/index.d.ts
+++ b/types/styled-components/index.d.ts
@@ -22,7 +22,7 @@ declare global {
 
 import * as CSS from "csstype";
 import * as React from "react";
-import hoistNonReactStatics = require('hoist-non-react-statics');
+import * as hoistNonReactStatics from 'hoist-non-react-statics';
 
 export type CSSProperties = CSS.Properties<string | number>;
 
--- a/types/styled-components/index.d.ts
+++ b/types/styled-components/index.d.ts
@@ -8,6 +8,7 @@
 //                 Sebastian Silbermann <https://github.com/eps1lon>
 //                 David Ruisinger <https://github.com/flavordaaave>
 //                 Matthew Wagerfield <https://github.com/wagerfield>
+//                 Yuki Ito <https://github.com/Lazyuki>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.9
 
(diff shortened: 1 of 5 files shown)


eps1lon (commented):
I guess this is fine. Though it seems like react statics like `propTypes` and `getDerivedStateFromProps` are now hoisted as well to the resulting type.

Would you be so kind and link the official documentation that describes hoisting of statics?

Lazyuki:
I suppose since `styled-components` already uses `hoist-non-react-statics` I should also do that here.

eps1lon:
> I suppose since `styled-components` already uses `hoist-non-react-statics` I should also do that here.

It was more about referencing that feature. I never used statics on styled-components so I'm not aware that they even do that.

types/styled-components/index.d.ts
```diff
@@ -21,6 +22,7 @@ declare global {
 
 import * as CSS from "csstype";
 import * as React from "react";
+import hoistNonReactStatics = require('hoist-non-react-statics');
```
eps1lon: Prefer `import ... from` to be consistent with previous imports.

eps1lon (changes requested):

types/styled-components/index.d.ts
Jessidhia: Actually I'd prefer to change the other imports to `import = require`. `react` is not an ES module, and as long as `esModuleInterop` doesn't default to `true` for `.d.ts` parsing it's better to be explicit about the commonjs import.

eps1lon: I'd agree but that is a) part of another change and b) inconsistent with virtually everybody else. That boat has long sailed where we can insist to not import react as an esmodule. `import * as React` is close enough. `import React = require()` is very uncommon syntax.

eps1lon (approved):

eps1lon:
@Lazyuki `npm run test styled-components` should reveal potential issues. Just a tiny syntax error.

Lazyuki:
@eps1lon Yeah I tried that but it errors out every time with `Error: .DS_Store: filesystem doesn't support paths of the form './x'.` and I can't seem to find any reference to this issue online. I tried deleting `.DS_Store` and also modifying the test script but didn't work...

Lazyuki:
Could I get the `Revision needed` label removed? Seems like it hasn't been removed automatically after the revision was done.

orta:
Looks good 👍

PR #40813, opened 2019-12-04
[@types/elasticsearch] hits.total is now an object in the search response

vietbui:
Breaking change introduced by Elasticsearch 7.0: https://www.elastic.co/guide/en/elasticsearch/reference/7.0/breaking-changes-7.0.html#hits-total-now-object-search-response

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.elastic.co/guide/en/elasticsearch/reference/7.0/breaking-changes-7.0.html#hits-total-now-object-search-response

### DIFF
--- a/types/elasticsearch/elasticsearch-tests.ts
+++ b/types/elasticsearch/elasticsearch-tests.ts
@@ -299,7 +299,7 @@ client.search({
     allTitles.push(hit.fields.title);
   });
 
-  if (response.hits.total !== allTitles.length) {
+  if (response.hits.total.value !== allTitles.length) {
     // now we can call scroll over and over
     client.scroll({
       scrollId: response._scroll_id!,
--- a/types/elasticsearch/elasticsearch-tests.ts
+++ b/types/elasticsearch/elasticsearch-tests.ts
@@ -299,7 +299,11 @@ client.search({
     allTitles.push(hit.fields.title);
   });
 
-  if (response.hits.total.value !== allTitles.length) {
+  const hitsTotal = typeof response.hits.total === 'object'
+    ? response.hits.total.value
+    : response.hits.total;
+
+  if (hitsTotal !== allTitles.length) {
     // now we can call scroll over and over
     client.scroll({
       scrollId: response._scroll_id!,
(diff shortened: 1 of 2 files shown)


dasdachs (changes requested):
Thanks for that! As far as I understand it defaults to an object, but can be if you add `rest_total_hits_as_int=true` in the search params. Right? Could you update you PR to account for that.

Reference: [ElasticNode API for MSearch](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/api-reference.html#_msearch)

vietbui:
@dasdachs Thanks for the review. I definitely missed that in the documentation. This is ready to be reviewed again. By the way I have a question:
- What's the level of testing do we expect for this?
- Regarding to the params, I see a mix usage of camel case and snake case time to time. E.g. in MSearchParams interface we have `search_type` and `maxConcurrentSearches`. I guess that we'll use camel case moving forward. However there's no attempt fix up existing snake case params as it'll create breaking change. Is my assumption correct?

Thanks again.

dasdachs (approved):
Thanks for the follow up!

dasdachs:
@vietbui I am relatively new to the project, so if any of the more senior contributes jump in and expand or correct my answer, I would be 

> What's the level of testing do we expect for this?

TBH I do not know. 

> Regarding to the params, I see a mix usage of camel case and snake case

For now, yes, we should leave it as it is. One solution could be to duplicate the types or interface properties and slowly transition to a unified camelCase naming conventions. One that does not force us to disable linters for certain lines and modules :)

Maybe ES 8.0 will be a good point to do some breaking changes. What are you thoughts on that?

rbuckton (changes requested):
Given this is a major version change as well as a breaking change, it might make sense to create a copy of the current definitions in a `v5` subfolder following the guidance in the Readme in the project root.

qdelettre:
@rbuckton That's not really a breaking until elastic 8.0.0, because there is an option that enable users to still get it not as an object ([cf commit message ](https://github.com/elastic/elasticsearch/pull/35849)) . So the types today are not right, this MR is totally valid until ES 8.0 imo.

PR #40816, opened 2019-12-04
react-table: port v7 types over from react-table repo

ggascoigne:
Since the v7 version of react-table has a completely incompatible
interface to v6, and is also no longer including types in its v7
package, this changeset ports over the v7 types and moves the v6 types
into a sub-folder.

The interface isn't ideal, it depends upon interface merging to be
useful.  If anyone can offer suggestions as to how to fix this without
falling into the recursive definition problems of generating types based
upon plugins which themselves change the types that they are being used
to define, please let me know.

more details can be found at https://gist.github.com/ggascoigne/646e14c9d54258e40588a13aabf0102d

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/tannerlinsley/react-table/blob/master/docs/api.md

regarding tslint:disable:no-unnecessary-generics
no-unnecessary-generics is disabled because many of these definitions are either used in a generic context or the signatures are required to match for declaration merging.


tannerlinsley (approved):

xPapla (approved):

kjaksik (approved):

antonjoel82:
Hi, everyone:

First off, thank you all for your contributions on this! I'm new to Typescript, but hope to be able to contribute down the line.

Please let me know if this is not the place to discuss this so I can relocate it, but I'm currently getting errors when I attempt to utilize some of the hooks with `useTable`. I've looked over the test file based on the "kitchen sink" example, and don't see any major differences. Here's my initialization with the hooks: 
```JavaScript
const {
        getTableProps,
        getTableBodyProps,
        headerGroups,
        prepareRow,
        page,
        canPreviousPage,
        canNextPage,
        pageOptions,
        pageCount,
        gotoPage,
        nextPage,
        previousPage,
        setPageSize,
        state: { expanded, pageIndex, pageSize },
    } = useTable<Task>({ 
            columns,
            data: tasks,
            getSubRows: getSubTaskRows,
            initialState: { pageIndex: 2 }
        },         
        useExpanded,
        usePagination
    );
```

This results in the following Typescript errors:
- Property page, canPreviousPage, canNextPage, pageOptions, pageCount, gotoPage, nextPage, previousPage, setPageSize does not exist on type 'TableState<Task>'.ts(2339)
- Property 'expanded, pageIndex, pageSize' does not exist on type 'TableState<Task>'.ts(2339)

I've been able to temporarily work-around this issue by modifying the index.d.ts file and extending other interfaces. Is there something I should do differently to fix this in my own code? Thanks!

tannerlinsley (approved):

rostero1:
Thanks for doing this.

I was trying to use these definitions with:  https://github.com/tannerlinsley/react-table/blob/master/examples/pagination-controlled/src/App.js, but ran into a few type errors around `initialState` and `manualPagination`.

ggascoigne:
> Thanks for doing this.
> 
> I was trying to use these definitions with: https://github.com/tannerlinsley/react-table/blob/master/examples/pagination-controlled/src/App.js, but ran into a few type errors around `initialState` and `manualPagination`.

These types depend on declaration merging, see https://github.com/ggascoigne/react-table/blob/ts-kitchen-sink-conrolled/examples/kitchen-sink-controlled-ts/src/types/react-table-config.d.ts for an example.

androsj:
@tannerlinsley is working so fast it's not giving this a break to merge xD

ggascoigne:
Please don't merge this yet.  I'd really like the interface to stabilize for more than a day before calling this good.

types/react-table/index.d.ts
```diff
@@ -1,746 +1,673 @@
...
+    columnsDeps: Array<(deps: any[], instance: TableInstance<D>) => any[]>;
+    flatColumns: Array<(flatColumns: Array<Column<D>>, instance: TableInstance<D>) => Array<Column<D>>>;
+    flatColumnsDeps: Array<(deps: any[], instance: TableInstance<D>) => any[]>;
+    headerGroups: Array<(flatColumns: Array<HeaderGroup<D>>, instance: TableInstance<D>) => Array<HeaderGroup<D>>>;
+    headerGroupDeps: Array<(deps: any[], instance: TableInstance<D>) => any[]>;
+    useInstanceBeforeDimensions: Array<(instance: TableInstance<D>) => void>;
+    useInstance: Array<(instance: TableInstance<D>) => void>;
+    useRows: Array<(rows: Array<Row<D>>, instance: TableInstance<D>) => Array<Row<D>>>;
+    prepareRow: Array<(row: Row<D>, instance: TableInstance<D>) => Row<D>>;
+    getTableProps: Array<(props: object, instance: TableInstance<D>) => object>;
```
kizu: This and other prop getters below now can be also just objects or arrays of objects instead of callbacks, and the callbacks can also return arrays of objects, see more there: https://github.com/tannerlinsley/react-table/issues/1749#issuecomment-564177091

types/react-table/index.d.ts
```diff
@@ -1,746 +1,673 @@
...
+    columns: Array<Column<D>>;
+    data: D[];
+} & Partial<{
+    initialState: Partial<TableState<D>>;
+    reducer: (newState: TableState<D>, action: string, previousState: TableState<D>) => TableState<D>;
+    useControlledState: (state: TableState<D>) => TableState<D>;
+    defaultColumn: Partial<Column<D>>;
+    initialRowStateKey: IdType<D>;
+    getSubRows: (originalRow: D, relativeIndex: number) => D[];
+    getRowId: (originalRow: D, relativeIndex: number) => IdType<D>;
```
androsj: To match type of selectedRowIds: `Set<string>`

```suggestion
    getRowId: (originalRow: D, relativeIndex: number) => string;
```

androsj:
`autoResetSelectedRows` is missing from `UseRowSelectOptions`

types/react-table/index.d.ts
```diff
@@ -1,746 +1,673 @@
...
+export interface UseTableColumnProps<D extends object> {
+    id: IdType<D>;
+    columns: Array<ColumnInstance<D>>;
+    isVisible: boolean;
+    render: (type: 'Header' | 'Footer' | string, props?: object) => ReactNode;
+    totalLeft: number;
+    totalWidth: number;
+    getHeaderProps: (props?: object) => object;
+    getFooterProps: (props?: object) => object;
+    toggleHiden: (value?: boolean) => void;
```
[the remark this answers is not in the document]
ggascoigne: damnit, I thought I'd fixed this one :)

Thank you for taking the time to look at this and give your feedback.

types/react-table/index.d.ts
ggascoigne: yeah, I just noticed this.

Oleg-Ovcharenko:
In react-table we have setHiddenColumns  but not setColumnHidden

types/react-table/index.d.ts
```diff
@@ -1,746 +1,673 @@
...
 
-export interface SortingRule {
+export interface UseTableColumnProps<D extends object> {
+    id: IdType<D>;
+    columns: Array<ColumnInstance<D>>;
+    isVisible: boolean;
+    render: (type: 'Header' | 'Footer' | string, props?: object) => ReactNode;
+    totalLeft: number;
+    totalWidth: number;
+    getHeaderProps: (props?: object) => object;
```
tpict: for prop getters like this, can we type the response as `object & { key: string }` in lieu of a more comprehensive interface?

kizu: I think we could maybe declare separate types outside, one for each prop getter, like `TableHeaderProps`, `TableRowProps`, `TableCellProps` etc. and then use those in prop getters? Like

```ts
export interface TableCellProps {
  key?: React.Key;
}

// …

getCellProps: (props?: TableCellProps) => TableCellProps;
```

This would allow us to use declaration merging to extend those interfaces easily, adding the props that we would be using?

tpict: Yeah absolutely. I just thought it was worth adding `key` at the very least since that needs to passed explicitly if you're enforcing `react/jsx-key` in your eslint config.

kizu: Ah, yes, hmm, it would be nice to make the key mandatory in the output, but optional in the input!

types/react-table/index.d.ts
```diff
@@ -1,746 +1,673 @@
...
+    getSubRows: (originalRow: D, relativeIndex: number) => D[];
+    getRowId: (originalRow: D, relativeIndex: number) => IdType<D>;
+}>;
+
+export interface UseTableHooks<D extends object> {
+    useOptions: Array<(options: TableOptions<D>, args: TableOptions<D>) => TableOptions<D>>;
+    stateReducers: Array<
+        (
+            newState: TableState<D>,
+            action: string,
```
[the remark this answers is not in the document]
ggascoigne: actions are typed now:

```tsx
export type ActionType = { type: string } & Record<string, any>;
```

types/react-table/index.d.ts
ggascoigne: I refactored the types for all of the property getters, there might be a more compact way to have done it, but what's there is at least correct, I hope :)

Happy to take any suggestions about how to make them cleaner.

types/react-table/index.d.ts
ggascoigne: done, separate types for each one.

types/react-table/index.d.ts
kizu: Here is my PR for your PR :) https://github.com/ggascoigne/DefinitelyTyped/pull/1 — adds a generic that makes it much easier to create all the getters (and also fixes their signatures etc.)

ggascoigne:
and by `intermediate`, I really meant `indeterminate` - bah.

types/react-table/index.d.ts
```diff
@@ -538,12 +542,16 @@ export interface TableToggleAllRowsSelectedProps {
     style?: { cursor: string };
     checked?: boolean;
     title?: string;
+    indeterminate?: boolean;
 }
 
-export interface TableToggleRowsSelectedProps {}
+export interface TableToggleRowsSelectedProps {
```
androsj: This has the same default props as `TableToggleAllRowsSelectedProps `

types/react-table/index.d.ts
```diff
@@ -167,6 +167,10 @@ export interface UseTableColumnOptions<D extends object>
 
 type UpdateHiddenColumns<D extends object> = (oldHidden: Array<IdType<D>>) => Array<IdType<D>>;
 
+export interface TableToggleHideAllColumnProps {
```
androsj: These default props are identical to `TableToggleRowsSelectedProps` and `TableToggleAllRowsSelectedProps`.
How about making a common interface then? Something like `TableToggleCommonProps`

types/react-table/index.d.ts
```diff
@@ -538,12 +542,16 @@ export interface TableToggleAllRowsSelectedProps {
...
 }
 
-export interface TableToggleRowsSelectedProps {}
+export interface TableToggleRowsSelectedProps {
+    indeterminate?: boolean;
+}
 
 export type UseRowSelectOptions<D extends object> = Partial<{
     manualRowSelectedKey: IdType<D>;
+    autoResetSelectedRows: boolean;
```
androsj: Should be optional`?` like the other autoResets ?

andrewbranch:
It looks like this is still in flux and not ready to merge; is that right @ggascoigne?

ggascoigne:
I can’t say that there will be no further changes before v7 is released, but the last few sets of changes have been around the plugin implementation interface rather than to the more public interfaces. I.e. the types for rc.12 would work pretty well with rc.15 unless you write your own plugins. 

So yes there might be changes, but hopefully they will be minor.

andrewbranch:
You’ve clearly put a ton of work into this, had a lot of eyes on it, and already have several community approvals, so I’m happy to merge whenever you’re ready.

JimmyMultani:
I would love to see this merged as is for now, with another PR created for when v7 is released. A lot of developers are stuck on beta.23 since it was the last version to actually have typescript defs in it.

ggascoigne:
Ok. Turns out that there are a few more changes on the way. So we’d best hold off.

ggascoigne:
@jimmymultani If you want to try these out early, I’d use patch-package to install them.

ChrisB9:
Are you able to roughly estimate how long you need to hold it?

ggascoigne:
@ChrisB9 Honestly I don't, I'm not really involved in the main lib, other than the odd small PR.  There are some more changes being worked on, I don't know how impactful they will be.

But I'm also not wedded to the idea that this PR doesn't get merged until v7 releases.  I think that it matches the current rc pretty well.

It seems that there is interest in merging it now and me opening a separate PR for any subsequent changes, I'm good with that.

@andrewbranch can you merge this when you get a chance.  Thank you.

PR #23940, opened 2018-02-27
[react-native] Add YellowBox API

timwangdev:
- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/facebook/react-native/blob/26684cf3adf4094eb6c405d345a75bf8c7c0bf88/Libraries/ReactNative/YellowBox.js> <https://github.com/facebook/react-native/commit/26038f5>
- [x] Increase the version number in the header if appropriate.
- [x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #40817, opened 2019-12-04
[react-naitve] fix YellowBox type

chinesedfan:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/facebook/react-native/blob/26684cf3adf4094eb6c405d345a75bf8c7c0bf88/Libraries/ReactNative/YellowBox.js#L312>>

---

@timwangdev This PR fixes #23940. `React.Component` means the component instance, while `YellowBox` is a class.

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -26,6 +26,7 @@
 //                 Romain Faust <https://github.com/romain-faust>
 //                 Be Birchall <https://github.com/bebebebebe>
 //                 Jesse Katsumata <https://github.com/Naturalclar>
+//                 Xianming Zhong <https://github.com/chinesedfan>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
@@ -8826,7 +8827,7 @@ export function findNodeHandle(
 
 export function processColor(color: any): number;
 
-export const YellowBox: React.Component<any, any> & { ignoreWarnings: (warnings: string[]) => void };
+export const YellowBox: React.ComponentClass<any, any> & { ignoreWarnings: (warnings: string[]) => void };
 
 //////////////////////////////////////////////////////////////////////////
 //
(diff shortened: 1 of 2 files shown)


jakebloom (commented):
Thanks for the PR! Could you add in your description why `React.Component` doesn't suffice here? What is it breaking that it needs to be changed to `React.ComponentClass`?

Given that the source code also [has an explicit type for state](https://github.com/facebook/react-native/blob/26684cf3adf4094eb6c405d345a75bf8c7c0bf88/Libraries/ReactNative/YellowBox.js#L314-L318), it would be nice to add that here too

chinesedfan:
> React.Component means the component instance, while YellowBox is a class.

@jakebloom I mentioned the reason in the PR description. Let me explain a little more.

```
// when some class `A` is defined, it can be used as a type or a value
class A {}
const A1: A = new A();       // the left `A` is a type, and the right one is a value

declare const A2: typeof A;  // for types, `A` is different with `typeof A`
new A2();
declare const A3: A;
new A3();                    // error

// so in the world of React
class C extends React.Component {}

declare const C1: typeof C;
<C1 />;                      // JSX is a kind of syntactic sugar, which means `React.createElement(C1)`

declare const C2: React.Component;
<C2 />;                      // error
```

The type definition of `React.createElement` can be found in [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L257). You can see it requires `ComponentClass` as the first parameter. In fact, `ComponentClass` is something like `typeof C` in the above example. And I have added similar test case for YellowBox.

```
function createElement<P extends {}>(
        type: FunctionComponent<P> | ComponentClass<P> | string,
        props?: Attributes & P | null,
        ...children: ReactNode[]): ReactElement<P>;
```

> Given that the source code also has an explicit type for state, it would be nice to add that here too

Usually, users don't care about the state of a component. Because it is a kind of internal implementation information. That's why I kept it as `any`. There are also other improvements can be done for this `react-native` typings file. I want to focus on a single `React.ComponentClass` problem now.

orta:
Alrighty, this feels reasonable to me - the types can be expanded if needed.

PR #40821, opened 2019-12-04
[rdf-js] Use term types consistently

thewilkybarkid:
Quads can be constructed using different types for each of the parts, but some of the Stream/Dataset methods don't make use of them and just use `Term`. This should mean know that you couldn't try and `match` a subject on a `DatasetCore` with a `DefaultGraph`.

- ~Provide a URL to documentation or source code which provides context for the suggested changes:~
- ~If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.~
- ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.~


rubensworks (changes requested):
I think the current types were intentional, where you could match any S/P/O/G type, independent of the store's quad type. Right @LaurensRietveld?

LaurensRietveld:
As I remember that was our train of thought indeed. I'm unfamiliar with the latest changes of the spec, so don't know whether this reasoning still stands

thewilkybarkid:
Hmm, trying to match on some other type should always not return anything?

rubensworks:
> Hmm, trying to match on some other type should always not return anything?

Indeed. But this PR will make this stricter than that, and will result in compilation errors when you try to match a quad with different types. I think we should remain liberal here, and accept any type, even if it doesn't match with the store's type.

thewilkybarkid:
Ok, I see that as being useful, but isn't a problem to close.

PR #40835, opened 2019-12-04
[react-native] Fix SectionList component type and add missing methods

newyankeecodeshop:
The `SectionList` type cannot be used with `React.createRef()` because it extends the wrong component type. (`ComponentClass` instead of `Component`). Extending `React.Component` is what `FlatList` and other similar components do. 
I updated the SectionList test to use the `createRef()` API to create the reference - everything compiles.
I left `SectionListStatic` in for compatibility with people that use that import.

This change could also fix issue #37077.

_This change likely breaks compatibility for some users of `SectionList`. I'm not sure what to do about that._

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://facebook.github.io/react-native/docs/sectionlist#methods-1

### DIFF
--- a/types/react-native/test/index.tsx
+++ b/types/react-native/test/index.tsx
@@ -368,10 +368,15 @@ export class FlatListTest extends React.Component<FlatListProps<number>, {}> {
 }
 
 export class SectionListTest extends React.Component<SectionListProps<string>, {}> {
-    myList: SectionList<any>;
+    myList: React.RefObject<SectionList<any>>;
+
+    constructor(props: SectionListProps<string>) {
+        super(props);
+        this.myList = React.createRef();
+    }
 
     scrollMe = () => {
-        this.myList.scrollToLocation({ itemIndex: 0, sectionIndex: 1 });
+        this.myList.current.scrollToLocation({ itemIndex: 0, sectionIndex: 1 });
     };
 
     render() {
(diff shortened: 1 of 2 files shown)


types/react-native/test/index.tsx
```diff
@@ -368,10 +368,15 @@ export class FlatListTest extends React.Component<FlatListProps<number>, {}> {
 }
 
 export class SectionListTest extends React.Component<SectionListProps<string>, {}> {
-    myList: SectionList<any>;
+    myList: React.RefObject<SectionList<any>>;
```
jakebloom: Any way we can make this more specific than `any`?

jakebloom (approved):
Looks okay to me!

thumbs up

types/react-native/test/index.tsx
newyankeecodeshop: Good call

newyankeecodeshop:
@jakebloom I removed the `any` but the Travis build failed. I can't tell what the failure is caused by. Can you re-run the Travis task and see if it was just a transient error?

el-angel:
@newyankeecodeshop  Can you also add `getScrollResponder()` and `getScrollableNode()`. Those are missing as well.

newyankeecodeshop:
@el-angel Add those - thanks. FWIW, they're missing from FlatList as well.

smaccoun:
Note that `getScrollResponder` is missing on `ScrollView` as well. Should I open a separate issue for that?

PR #40839, opened 2019-12-05
@types/mui-datatables - Change BlobPart back to string type

dandrews:
BlobPart is undefined/not found. The previous return type was a string.

Please fill in this template.

### DIFF
--- a/types/mui-datatables/index.d.ts
+++ b/types/mui-datatables/index.d.ts
@@ -199,7 +199,7 @@ export interface MUIDataTableOptions {
         buildBody: (data: any) => string,
         columns: any,
         data: any
-    ) => BlobPart;
+    ) => string;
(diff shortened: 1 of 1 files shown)


favna (approved):
LGTM. Should be string when looking at the lib docs:
> A callback function that triggers when the user downloads the CSV file. In the callback, you can control what is written to the CSV file. function(buildHead: (columns) => string, buildBody: (data) => string, columns, data) => string. Return false to cancel download of file.

Not sure exactly when BlobPart slipped through the cracks and why it passed tests when it did if it truly doesn't exist in at least @types/node....

---

Can you untick / remove this option? It doesn't apply. 
> If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.

PR #40840, opened 2019-12-05
Fix missing argument defintion in nuxtjs__auth

aruneko:
The `setUserToken` function requires `token` argument. But there is no definition in this code. I fixed this problem.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/nuxt-community/auth-module/blob/ea08d39b849b882ea32da7a9db9c338bef210db5/lib/core/auth.js#L162>>

### DIFF
--- a/types/nuxtjs__auth/index.d.ts
+++ b/types/nuxtjs__auth/index.d.ts
@@ -47,7 +47,7 @@ export interface Auth<T = any> {
   strategy(): string;
   registerStrategy(strategyName: string, strategy: object): void;
   setStrategy(strategyName: string): void;
-  setUserToken(): Promise<never>;
+  setUserToken(token: string): Promise<never>;
   getRefreshToken(strategyName: string): string;
   setRefreshToken(strategyName: string, token?: string): string;
   syncRefreshToken(strategyName: string): string;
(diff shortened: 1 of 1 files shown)


jonnyparris (approved):

types/nuxtjs__auth/index.d.ts
```diff
@@ -47,7 +47,7 @@ export interface Auth<T = any> {
   strategy(): string;
   registerStrategy(strategyName: string, strategy: object): void;
   setStrategy(strategyName: string): void;
-  setUserToken(): Promise<never>;
+  setUserToken(token: string): Promise<never>;
```
rbuckton: `Promise<never>` does not seem like the right type (unless the Promise is guaranteed to throw). I would expect this to be `Promise<void>` if its a promise whose value is unused.

aruneko: @rbuckton Thanks for your reviewing! I hadn't noticed that. Now, I've just fixed this type definition.

rbuckton (approved):

rbuckton:
Thanks for the contribution!

PR #40844, opened 2019-12-05
[rc-tooltip] Make `overlay` non-nullable

fxlrnrpt:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/react-component/tooltip/blob/3.7.3/src/Tooltip.jsx#L21 - it's required there

### DIFF
--- a/types/rc-tooltip/index.d.ts
+++ b/types/rc-tooltip/index.d.ts
@@ -3,6 +3,7 @@
 // Definitions by: rhysd <https://github.com/rhysd>
 //                 ahstro <https://github.com/ahstro>
 //                 vsaarinen <https://github.com/vsaarinen>
+//                 aigoncharov <https://github.com/aigoncharov>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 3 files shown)


types/rc-tooltip/index.d.ts
```diff
@@ -31,7 +32,7 @@ declare namespace RCTooltip {
 		placement?: Placement | Object;
 		align?: Object;
 		onPopupAlign?: (popupDomNode: Element, align: Object) => void;
-		overlay: React.ReactNode;
+		overlay: React.ReactChild | React.ReactFragment | React.ReactPortal;
```
rhysd: From [documentation](https://github.com/react-component/tooltip#props), I think function type should be included in this union.

> `React.Element | () => React.Element`

rhysd (changes requested):

types/rc-tooltip/index.d.ts
aigoncharov: Great catch! Addressed!

rhysd (approved):

orta:
Cool, looks good to me 👍

PR #40846, opened 2019-12-05
[@types/stripe] Added config param to constructor

allouis:
Please fill in this template.

~^ Running `npm run lint stripe` gives back an error like:~
```
The 'no-unnecessary-generics' rule threw an error in '/usr/src/js/DefinitelyTyped/types/stripe/index.d.ts':
TypeError: sig.parameters is not iterable
```

~:thinking: It doesn't seem like this is related to this change - but I'm not sure how to fix it :|~

Looks like it passes on CI :relaxed: 

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/stripe/stripe-node#initialize-with-config-object
- [ ] ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.~

### DIFF
--- a/types/stripe/index.d.ts
+++ b/types/stripe/index.d.ts
@@ -64,9 +64,9 @@ interface ResponseEvent {
 }
 
 interface StripeConfig {
-    apiVersion?: string;
+    apiVersion?: string | null;
     maxNetworkRetries?: number;
-    httpAgent?: Agent;
+    httpAgent?: Agent | null;
     timeout?: number;
     host?: string;
     port?: number;
--- a/types/stripe/index.d.ts
+++ b/types/stripe/index.d.ts
@@ -64,13 +64,13 @@ interface ResponseEvent {
 }
 
 interface StripeConfig {
-    apiVersion: string;
-    maxNetworkRetries: number;
-    httpAgent: Agent;
-    timeout: number;
-    host: string;
-    port: number;
-    telemetry: boolean;
+    apiVersion?: string;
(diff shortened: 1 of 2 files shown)


acomagu:
Oh, dups with mine #40841...

acomagu:
Your pull is better than mine because yours keeps the line `constructor(apiKey: string, version?: string);` which is fit to the real implementation.
So I closed mine. Hope to merge soon!(I'm beginner to contributing DefinitlyTyped so I can't review and approve, sorry)

types/stripe/index.d.ts
```diff
@@ -63,6 +63,16 @@ interface ResponseEvent {
     request_end_time: number;
 }
 
+interface StripeConfig {
+    apiVersion: string;
```
pushplay: From the doc it looks like all the properties here should be marked optional.  As is all properties have to be specified.

`apiVersion` and `httpAgent` also allow null so their types should be `string | null` and `Agent | null` respectively.

pushplay (changes requested):
Thanks for improving the tests.

allouis:
Thanks for the review! I've updated the types and added another test :relaxed:

pushplay (approved):
Thanks for the updates.

orta:
Looks like additions only to me, thanks!

PR #40851, opened 2019-12-05
[mapbox-gl] Update mapbox-gl to v1.6.0

dmytro-gokun:


clshortfuse:
@dmytro-gokun It's missing `setRotation()`, `setRotationAlignment()` and `setPitchAlignment()`. Thanks!

macobo:
@cls

macobo:
@clshortfuse feel free to create your own follow-up PR on top of this. :)

dmytro-gokun:
@clshortfuse Looks like those methods existed before 1.6.0, but were missed somehow. Maybe they are rarely used. In any case, feel free to create a PR and i promise to review it quickly

clshortfuse:
@dmytro-gokun They're actually new in 1.6.0 since they are related to setting the new variables as in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40851/files#diff-c53d5b80c827351f33db0313e7695631R1256

I can take a stab a creating another PR, but I saw you were updating multiple branches related to MapBox, so I figured I should also bring that to your attention, in case you missed them elsewhere as well.

PR #40854, opened 2019-12-05
fix: mask must accept Regex[] type - react-input-mask

lucasraziel:
Please fill in this template.

- fix: mask must accept Regex[] type - react-input-mask

mask was only accepting string, but according to the documentation of the package (link below), it also accepts a string with Regexp.

- <<https://github.com/sanniassin/react-input-mask>>


orta:
Looks good - thanks

PR #40859, opened 2019-12-05
Adds definition for Message to intercom-client

mlucic:
Messages type was missing from intercom-client definition.

### DIFF
--- a/types/intercom-client/index.d.ts
+++ b/types/intercom-client/index.d.ts
@@ -181,7 +181,4 @@ export class Events {
 export class Messages {
     create(message: Partial<Message>): Promise<IncomingMessage>;
     create(message: Partial<Message>, cb: callback<IncomingMessage>): void;
-
-    listBy(params: EventListParam): Promise<ApiResponse<CompanyList>>;
-    listBy(params: EventListParam, cb: callback<ApiResponse<CompanyList>>): void;
 }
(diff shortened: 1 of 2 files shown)


types/intercom-client/index.d.ts
```diff
@@ -175,3 +178,8 @@ export class Events {
     listBy(params: EventListParam): Promise<ApiResponse<CompanyList>>;
     listBy(params: EventListParam, cb: callback<ApiResponse<CompanyList>>): void;
 }
+
+export class Messages {
```
DanielRosenwasser: There's no value named `Messages` in `index.js` of the original project. This doesn't quite look correct to me.

mlucic: It's inside of lib/client.js

https://developerblog.intercom.com/how-to-send-html-emails-with-the-intercom-conversations-api-748eb2784b1

andrewbranch: Ok, but it’s not exported from index.js, so this declaration is incorrect: 

image

andrewbranch (changes requested):

types/intercom-client/index.d.ts
mlucic: @andrewbranch The other values such as Companies do not exist in the index of the original either, I am simply following the precedent set by the existing definition.

andrewbranch: Hm, you’re right. Those should be changed too.

PR #40864, opened 2019-12-05
React native sqlite storage generics

Elias-Graf:
I tried to add generic typing to the `executeSql` method.


Elias-Graf:
Hm, I'm not exactly sure how to fix this. Even opened an issue on [SO](https://stackoverflow.com/questions/59202664/fix-no-unnecessary-generics/59203417#59203417).

andrewbranch:
These functions are not generic, and typing them as such is incorrect; that’s why the `no-unnecessary-generics` rule exists. You can see more answer at https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36868#issuecomment-512989650 for more explanation.

PR #40869, opened 2019-12-06
Add types for php-serialize

blurfx:
Please fill in this template.

If adding a new definition:


types/php-serialize/index.d.ts
```diff
@@ -0,0 +1,11 @@
+// Type definitions for php-serialize 3.0
+// Project: https://github.com/steelbrain/php-serialize
+// Definitions by: Changhui Lee <https://github.com/blurfx>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.7
+
+export function isSerialized(givenItem: any, strict: any): any;
```
demensky:
```suggestion
export function isSerialized(givenItem: string, strict: boolean): boolean;
```

types/php-serialize/index.d.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export function serialize(item: any, scope?: any, givenOptions?: any): any;
```
demensky:
```d.ts
export interface SerializeOptions {
    /** @default 'utf8' */
    readonly encoding?: 'utf8' | 'binary';
}
```

```suggestion
export function serialize(item: any, scope?: any, givenOptions?: SerializeOptions): string;
```

types/php-serialize/index.d.ts
```diff
@@ -0,0 +1,11 @@
...
+
+export function unserialize(item: any, scope?: any, givenOptions?: any): any;
```
demensky:
```d.ts
export interface UnserializeOptions {
    /** @default true' */
    readonly strict?: boolean;
    /** @default 'utf8' */
    readonly encoding?: 'utf8' | 'binary';
}
```

```suggestion
export function unserialize(item: string | Buffer, scope?: any, givenOptions?: UnserializeOptions): any;
```

types/php-serialize/index.d.ts
demensky: [`strict` is optional](https://github.com/steelbrain/php-serialize/blob/master/src/isSerialized.js#L8).

```suggestion
export function isSerialized(givenItem: any, strict?: any): any;
```

demensky (changes requested):
The source code and documentation have a description of the types. I suggest reducing the amount of `any`.

demensky (approved):

PR #40873, opened 2019-12-06
add types for windows-mutex

bpasero:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/windows-mutex/index.d.ts
+++ b/types/windows-mutex/index.d.ts
@@ -1,0 +1,12 @@
+// Type definitions for windows-mutex 0.4
+// Project: https://github.com/Microsoft/node-windows-mutex
+// Definitions by: Benjamin Pasero <https://github.com/bpasero>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export class Mutex {
+    constructor(name: string);
+    isActive(): boolean;
+    release(): void;
+}
+
+export function isActive(name: string): boolean;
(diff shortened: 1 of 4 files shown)


demensky:
The package [already provide its own types](https://github.com/microsoft/node-windows-mutex/blob/master/index.d.ts).

bpasero:
@demensky that is right, however for our use in VSCode, the module is an [optional dependency](https://github.com/microsoft/vscode/blob/a416c77e56ef0314ae00633faa04878151610de8/package.json#L164) as it only works on Windows. As such, we cannot use the types from its module because the module does not get installed on macOS or Linux. Thus, it seems the only solution is to install the types explicitly.

bpasero:
@demensky ping (also related https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40962)

demensky:
It is difficult to offer something universal.

If installed via `--force` then the types work:

Снимок экрана 2019-12-12 в 13 33 15

Probably need to create a related issue at [node-windows-mutex](https://github.com/Microsoft/node-windows-mutex).

bpasero:
@demensky sorry, what is your suggestion?

demensky:
Remove types from [node-windows-mutex](https://github.com/microsoft/node-windows-mutex/blob/master/index.d.ts) (create a PR there) and add types here (you are already doing this).

bpasero:
@demensky sure, done: https://github.com/microsoft/node-windows-mutex/commit/a9f13b90fba1144ee202072e0acfce220737a54d

demensky (approved):

orta:
OK, then this looks good to me now that is deployed

bpasero:
@orta thanks, can you check on https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40962 as well?

PR #40875, opened 2019-12-06
[airtable] Update definitions for version 0.8

samtgarson:
[airtable] Update definitions for version 0.8

Please fill in this template.

If adding a new definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [airtable.js](https://github.com/Airtable/airtable.js/tree/master/lib)

### DIFF
--- a/types/airtable/index.d.ts
+++ b/types/airtable/index.d.ts
@@ -5,103 +5,97 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.2
 
-export = Airtable;
+export class Airtable {
+    constructor(options?: Airtable.AirtableOptions);
+    base(appId: string): Airtable.Base;
+}
 
-declare global {
-    class Airtable {
-        constructor(options?: Airtable.AirtableOptions);
-        base(appId: string): Airtable.Base;
-    }
+export interface FieldSet {
+    [key: string]:
+        | undefined
+        | string
+        | number
+        | boolean
+        | Collaborator
+        | ReadonlyArray<Collaborator>
+        | ReadonlyArray<string>
+        | ReadonlyArray<Attachment>;
+}
 
-    namespace Airtable {
-        interface FieldSet {
-            [key: string]:
-                | undefined
-                | string
-                | number
-                | boolean
-                | Collaborator
-                | ReadonlyArray<Collaborator>
-                | ReadonlyArray<string>
-                | ReadonlyArray<Attachment>;
-        }
+export interface AirtableOptions {
+    apiKey?: string;
+    endpointUrl?: string;
+    apiVersion?: string;
+    allowUnauthorizedSsl?: boolean;
+    noRetryIfRateLimited?: boolean;
+    requestTimeout?: number;
+}
 
-        interface AirtableOptions {
-            apiKey?: string;
-            endpointUrl?: string;
-            apiVersion?: string;
-            allowUnauthorizedSsl?: boolean;
-            noRetryIfRateLimited?: boolean;
-            requestTimeout?: number;
-        }
+export interface Base {
+    (tableName: string): Table<{}>;
+}
 
-        interface Base {
-            (tableName: string): Table<{}>;
-        }
+export interface Table<TFields extends FieldSet> {
+    select(opt?: SelectOptions): Query<TFields>;
+    find(id: string): Promise<Record<TFields>>;
+    create(record: TFields, opts?: { typecast: boolean }): Promise<Record<TFields>>;
+    create(records: TFields[], opts?: { typecast: boolean }): Promise<Records<TFields>>;
+    update(...args: any[]): Promise<any>;
+    replace(...args: any[]): Promise<any>;
+    destroy(...args: any[]): Promise<any>;
+}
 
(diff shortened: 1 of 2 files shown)


samtgarson:
I believe @EvanHahn is the maintainer on this 🙌

EvanHahn (changes requested):
Rejecting just because I have a couple of questions:

First, this change seems to update nearly every line in the type definitions and the tests, including updates to code style. What's the motivation for such a big change?

Second, could these changes be split up into smaller ones for easier review, clearer history, and less risk?

samtgarson:
Thanks for the review @EvanHahn, fair comments.

The test formatting change was because some clauses were wrapped in `{}`, which seemed unconventional to me—so it's just a de-indentation, although I admit the Github diff makes this hard to see.

The big-ish changes to the declaration file were also because the format seemed unconventional:
- `declare global` doesn't seem super appropriate here
- `interface` has been used for objects that are actual classes in the real library
- the test imports using `import * as Airtable from 'airtable'`, which doesn't actually work with the real library—the default export of the lib is a function and you can import the other named exports.

Hence I decided to make some changes which aren't that substantial but in a git diff do appear significant (I don't think there's any actual code style changes other than the `{}` in the test, but maybe I let some bias slip in so please let me know if you see something specific)

I have been using these types on a personal project for a while and it's all ok—but I will attempt break this into separate commits to make things a little easier to review.

All this to say—I'm fairly new to Typescript so I may have some core misunderstandings which I'm happy to be corrected on 😅

samtgarson:
@EvanHahn I've broken this into two commits, and also put the definitions back in their original order which should make it clearer what's actually changed.

types/airtable/index.d.ts
```diff
@@ -1,13 +1,23 @@
-// Type definitions for airtable 0.5
+// Type definitions for airtable 0.8
```
EvanHahn: Should this specify an exact version?

types/airtable/index.d.ts
```diff
@@ -1,13 +1,23 @@
...
+// TypeScript Version: 3.7
 
-export class Airtable {
-    constructor(options?: Airtable.AirtableOptions);
-    base(appId: string): Airtable.Base;
+export type baseCreateFunctor = (tableName: string) => Table;
+
+export default class Airtable {
+    constructor(options?: AirtableOptions);
+    base(appId: string): baseCreateFunctor;
```
EvanHahn: Should this return a `Base` instead?

types/airtable/index.d.ts
```diff
@@ -1,13 +1,23 @@
...
 //                 Max Chehab <https://github.com/maxchehab>
+//                 Sam Garson <https://github.com/samtgarson>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.2
+// TypeScript Version: 3.7
 
-export class Airtable {
-    constructor(options?: Airtable.AirtableOptions);
-    base(appId: string): Airtable.Base;
+export type baseCreateFunctor = (tableName: string) => Table;
```
EvanHahn: Does this type need to be exported?

In general, I think we might be exporting more things than need exporting. What do you think?

types/airtable/index.d.ts
```diff
@@ -1,107 +1,131 @@
...
+
+export interface NewFieldData<TFields> {
+    fields: Partial<TFields>;
+}
+
+export interface ExistingFieldData<TFields> extends NewFieldData<TFields> {
+    id: string;
+}
+
+export interface AirtableOptions {
```
EvanHahn: 1. Does this need to be exported?
1. I believe `apiKey` should be required.
1. The `allowUnauthorizedSsl` option was removed in v0.8.0, so it should be removed here.

types/airtable/index.d.ts
```diff
@@ -1,107 +1,131 @@
...
+    apiKey?: string;
+    endpointUrl?: string;
+    apiVersion?: string;
+    allowUnauthorizedSsl?: boolean;
+    noRetryIfRateLimited?: boolean;
+    requestTimeout?: number;
+}
+
+export class Base {
```
EvanHahn: Users aren't really meant to instantiate this themselves. Could we remove the export (or maybe mark the constructor as private, if TypeScript allows such a thing)?

The same comment applies to `Table` and `Record`.

types/airtable/index.d.ts
```diff
@@ -1,107 +1,131 @@
...
+export type baseCreateFunctor = (tableName: string) => Table;
+
+export default class Airtable {
+    constructor(options?: AirtableOptions);
+    base(appId: string): baseCreateFunctor;
+    static base(appId: string): baseCreateFunctor;
+    static configure(options: AirtableOptions): void;
+}
+
+export class Error {
```
EvanHahn: Internally, this is a class called `AirtableError`. Why was this renamed to `Error` here?

EvanHahn (changes requested):
Didn't do a very thorough review, but left a few comments.

Also, as an aside: will this break anyone who is using the existing types? In other words, will they need to change their type imports? If so, what is DefinitelyTyped's process here?

PR #40876, opened 2019-12-06
aws-lambda - removes the wrong assumption that headers are always set

hedefalk:
Please fill in this template.

Sorry, at the time, I can't find any good docs for this, better than https://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-handler.html

which states that "the event structure varies by service.".

However, the reason fot the PR is by imperical/anecdotal evidence :) When testing a lambda in aws console without providing any headers, null was passed giving my typechecked code a NPE.

### DIFF
--- a/types/aws-lambda/index.d.ts
+++ b/types/aws-lambda/index.d.ts
@@ -79,7 +79,7 @@ export interface APIGatewayEventRequestContext {
 // API Gateway "event"
 export interface APIGatewayProxyEvent {
     body: string | null;
-    headers: { [name: string]: string };
+    headers?: { [name: string]: string };
     multiValueHeaders: { [name: string]: string[] };
     httpMethod: string;
     isBase64Encoded: boolean;
--- a/types/aws-lambda/aws-lambda-tests.ts
+++ b/types/aws-lambda/aws-lambda-tests.ts
@@ -125,7 +125,7 @@ strOrUndefined = apiGwEvtReqCtx.routeKey;
 
 /* API Gateway Event */
 strOrNull = apiGwEvt.body;
-str = apiGwEvt.headers['example'];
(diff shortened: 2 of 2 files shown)


types/aws-lambda/index.d.ts
```diff
@@ -79,7 +79,7 @@ export interface APIGatewayEventRequestContext {
 // API Gateway "event"
 export interface APIGatewayProxyEvent {
     body: string | null;
-    headers: { [name: string]: string };
+    headers?: { [name: string]: string };
     multiValueHeaders: { [name: string]: string[] };
```
alex-bolenok-centralreach: I assume those are always set?

alex-bolenok-centralreach (approved):

simonbuchan:
Hmm, making headers optional is a breaking change, I'm suspecting we're going to get complaints to change this back. I think they will never be non-null in practice, as you will always at least get the `Host` header from HTTP/1.1, so I would put this down as an unfortunate bug in the AWS console.

That said, it's not too hard a workaround, and this is already merged, so I guess wait to see if there are complaints?

brianeletype:
> Hmm, making headers optional is a breaking change, I'm suspecting we're going to get complaints to change this back. I think they will never be non-null in practice, as you will always at least get the `Host` header from HTTP/1.1, so I would put this down as an unfortunate bug in the AWS console.
> 
> That said, it's not too hard a workaround, and this is already merged, so I guess wait to see if there are complaints?

I'll submit the first complaint.  This broke a bunch of code where we're getting the Host header.  Why would a breaking change like get through?

simonbuchan:
@brianeletype and others, I've opened a revert MR 👆 in #40929

PR #38878, opened 2019-10-04
Feature/parse/various

RaschidJFR:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: 
    * `Parse.File`: https://github.com/parse-community/Parse-SDK-JS/blob/fee4976ba4230e929093a91f095b3731d5917bdf/src/ParseFile.js#L102
    * `Query`: https://github.com/parse-community/Parse-SDK-JS/blob/fee4976ba4230e929093a91f095b3731d5917bdf/src/ParseFile.js#L102
    * `LiveQuerySubscription`: https://github.com/parse-community/Parse-SDK-JS/blob/master/src/LiveQuerySubscription.js
    * `Parse.Config`: https://github.com/parse-community/Parse-SDK-JS/blob/fee4976ba4230e929093a91f095b3731d5917bdf/src/ParseConfig.js#L87
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

PR #40878, opened 2019-12-06
[@types/parse] 2.10: Stricter Parse.Object attributes type, improve linting, and add tests

rdhelms:

### Changelog
**Updated**
- Type definitions for parse 2.10
- TypeScript Version 3.3
    - Now using conditional types from TS 2.8
    - More complex generic inferences from TS 3.3
- Made `Parse.Object` a generic in order to type its attributes
    - In order to handle `Parse.Object`'s complex construct signature, `class Object` was switched to `const Object: ObjectConstructor`. This is modeled based on the native `const Array: ArrayConstructor` type pattern.
    - The default `attributes` type is intentionally broad to minimize breaking changes: `{ [key: string]: any }`. Users are able to "opt in" to stricter typing by passing a type parameter to `Parse.Object`
    - All classes that extend `Parse.Object` are also generic.
- Added many more tests, and made use of `// $ExpectType` and `// $ExpectError` comments for stricter assertions

**Fixed**
- Removed all but two tslint rule exceptions
    - `ban-types: false` is still necessary because Parse has its own `Object` interface that tslint thinks is a banned type
    - `no-unnecessary-generics` is still necessary because we allow users to type various return types via a type parameter
    - `parse/node` and `parse/react-native` module declarations are now in their own files, per tslint preference
    - Various other lint fixes: semi-colons, combined overloads, whitespace, comment formatting
- `ErrorCode` was a global type, but is not actually a global in the parse library. It is no longer global.
- The `BaseObject` interface and class were removed
    - The parse library did not export any such class
    - It was only being used as a convenience to add `toJSON(): any` as a method on other interfaces, but in reality those various `toJSON` methods will not have the same return type. Each interface now has its own `toJSON` method, so that in the future it will be easier to provide better typings for them individually.
- Classes that extended `Parse.Object` had constructors that were typed incorrectly. `Parse.User`, `Parse.Session`, `Parse.Installation`, and `Parse.Role` now have construct signatures that match the documentation.
- `Parse.Config` does not actually extend `Parse.Object` - it is no longer typed as such.
- Some types were removed that are not documented or found in parse's source code
    - `Parse.VERSION`
    - `Parse.Object.cid` and `Parse.Object.changed`


RaschidJFR:
Hey @rdhelms thanks for your work. Love what you did with the tests. I think this PR is conflicting with #40429 you may want to check that. Also the review has become hard because of so much noise in whitespace change. Are you using the same indentation settings as the previous commit?

rdhelms:
Hi @RaschidJFR - thanks for taking a look. 

I actually worked with @pocketcolin on #40429 😀 we were hoping that PR would help setup the foundation for this PR. That one allowed for a generic Parse.Object with minimal other changes, while this one is a bit larger.

I agree the diff is tricky - the main reason being that one of the lint fixes involved re-organizing how the global types were declared, and that’s what changed the indentation by one level. If it would be preferable, I could re-disable that rule again for now and address that in another PR by itself. That might help make it easier to evaluate the other changes.

I’ll address the merge conflict on Monday - a smaller PR was merged a couple days ago but the conflict doesn’t look too bad. If anyone has any other comments/reviews I can address those tomorrow as well.

RaschidJFR:
Oh I love what you guys did there. Personally I think is one of the greatest contributions to the repo!

rdhelms:
@RaschidJFR I resolved the merge conflicts - all your work from #38878 should be included now, but it might be good to double check that.

I also reverted to the old way of declaring the modules (and re-added the exceptions to those lint rules) so that the diff should be much easier to read now. This means that the `ErrorCode` type is global again which I would consider to be a bug, but it will be easier to address that in a separate PR by itself.

orta:
Let me know when you're happy to merge 👍

rdhelms:
@RaschidJFR do you have any other thoughts on this? It’s good to go as far as I know.

RaschidJFR (approved):
LGTM 👍

PR #40880, opened 2019-12-06
[react] Add crossOrigin to SVGAttributes

icosahebron:
Please fill in this template.

React supports SVG `<image>` elements with a `crossOrigin` prop. It even special-cases that prop so that it's rendered as an all-lowercase attribute in HTML (see https://github.com/facebook/react/pull/14832). 

This PR updates `@types/react` to reflect this, letting us write:

```tsx
<image crossOrigin="anonymous"/>
```

instead of workarounds like:

```tsx
<image {...{ crossOrigin: "anonymous" }}/>
```

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -2259,6 +2259,7 @@ declare namespace React {
         // Other HTML properties supported by SVG elements in browsers
         role?: string;
         tabIndex?: number;
+        crossOrigin?: "anonymous" | "use-credentials" | "";
 
         // SVG Specific attributes
         accentHeight?: number | string;
--- a/types/react/test/elementAttributes.tsx
+++ b/types/react/test/elementAttributes.tsx
@@ -26,4 +26,5 @@ const testCases = [
     <span autoCorrect="on" />,
     <span translate="no" />,
     <span translate="yes" />,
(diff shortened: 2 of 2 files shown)


types/react/index.d.ts
```diff
@@ -2259,6 +2259,7 @@ declare namespace React {
         // Other HTML properties supported by SVG elements in browsers
         role?: string;
         tabIndex?: number;
+        crossOrigin?: "anonymous" | "use-credentials" | "";
```
ferdaber: What's the purpose of the blank string as an allowed value here?

icosahebron: The blank string appears in the `` equivalent: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/833391a966801f2e2c4ca25466d81b5265e1f4f7/types/react/index.d.ts#L1957

and when used as an attribute value, [is treated the same as `anonymous`](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin)

> Setting the attribute name to an empty value, like crossorigin or crossorigin="", is the same as anonymous.

eps1lon (approved):

orta:
👍

PR #40890, opened 2019-12-07
[parquetjs] Add useDataPageV2 option, deduplicate opts to interface

dobesv:
This declares the `useDataPageV2` option, which must be set to `false` for compatibility with the C++ (and possibly other) implementations of parquet.  BigQuery, for example, does not seem to support the V2 data page.

At the same time I factored out the options to an interface, to reduce duplication.

See https://github.com/neilcoinspectapp/parquetjs#notes for the mention of this option in the readme.


jon-fearer (approved):
Looks good. Thanks for the improvements!

orta:
👍

PR #40891, opened 2019-12-07
react-map-gl - Added missing types

Arman92:
Fixed missing types according to: https://github.com/uber/react-map-gl/blob/5.1-release/examples/viewport-animation/src/app.js
FlyToInterpolator class was missing it's constructor props and transitionDuration property couldn't have 'auto' value

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/uber/react-map-gl/blob/3e35c9b78430cdc19f1072806ff2545966ac4542/examples/viewport-animation/src/app.js#L30

### DIFF
--- a/types/react-map-gl/index.d.ts
+++ b/types/react-map-gl/index.d.ts
@@ -221,7 +221,7 @@ export interface FlyToInterpolatorProps {
     speed?: number;
     screenSpeed?: number;
     maxDuraiton?: number;
-};
+}
 
 export class FlyToInterpolator extends TransitionInterpolator {
     constructor(props?: FlyToInterpolatorProps);
--- a/types/react-map-gl/index.d.ts
+++ b/types/react-map-gl/index.d.ts
@@ -80,10 +80,7 @@ export interface QueryRenderedFeaturesParams {
 
 export class StaticMap extends React.PureComponent<StaticMapProps> {
     getMap(): MapboxGL.Map;
-    queryRenderedFeatures(
-        geometry?: MapboxGL.PointLike | MapboxGL.PointLike[],
-        parameters?: QueryRenderedFeaturesParams,
-    ): Array<GeoJSON.Feature<GeoJSON.GeometryObject>>;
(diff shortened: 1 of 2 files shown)


oturpe:
The main content looks good, thank you @Arman92 !

Both I and the test automation have the complaint that some autoformatter has been run on top of the codebase. Could you revert any autoformatting and only include changes related to the new type, using the style that is already used?

types/react-map-gl/index.d.ts
```diff
@@ -215,7 +223,16 @@ export class LinearInterpolator extends TransitionInterpolator {
     constructor(transitionProps?: string[]);
 }
 
-export class FlyToInterpolator extends TransitionInterpolator {}
+export type FlyToInterpolatorProps = {
```
fnberta: Can we make this an interface instead? The rest of the types use interfaces when possible.

fnberta (changes requested):
Thanks for the PR, looks good, only left one minor comment. Regarding the formatting, I actually welcome the changes. But maybe have them as a separate PR, that makes it clearer what was changed.

types/react-map-gl/index.d.ts
Arman92: Sure, Actually the builds were failing because I used 'type' instead of 'interface', just fixed it in a new commit.

oturpe:
Just to make it clear, I actually do not have any position regarding how the codebase should be formatted. The current style does for me as well as any other would. My concern was related to mixing type updates and reformatting in a single MR.

Arman92:
@oturpe Sure, you are absolutely right. I didn't notice that my prettier tool did a reformat while I saved files, I do agree with you; code reformatting should be in a different PR.

fnberta:
There are still formatting changes in the tests file. Could you revert these as well and put in a separate PR? Thanks!

fnberta (approved):

Arman92:
This PR is still waiting for someone to kindly apply the merge!

Arman92:
@oturpe, @fnberta This PR is still pending, maybe it's because it's still labeled 'Revision Needed'.

Can you remove this label? Or should I ask a maintainer to do this?

fnberta:
Hmm I dont think I can do anything besides approving which I already did.

PR #40892, opened 2019-12-07
fix(request): add maxSockets params at pool option

sshaplygin:
Please fill in this template.

- [] Use a meaningful title for the pull request. Include the name of the package modified.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [documentation](https://github.com/request/request/blob/master/README.md) and [code](https://github.com/request/request/blob/master/request.js#L700)

If removing a declaration:

### DIFF
--- a/types/request/index.d.ts
+++ b/types/request/index.d.ts
@@ -149,7 +149,7 @@ declare namespace request {
         maxRedirects?: number;
         removeRefererHeader?: boolean;
         encoding?: string | null;
-        pool?: false | { [key: string]: http.Agent | https.Agent } & PoolOptions | PoolOptions;
+        pool?: PoolOptions;
         timeout?: number;
         localAddress?: string;
         proxy?: any;
@@ -197,9 +197,11 @@ declare namespace request {
         };
     }
 
-    interface PoolOptions {
+    interface ExtraPoolOptions {
         maxSockets?: number;
     }
+
+    type PoolOptions = false | { [key: string]: http.Agent | https.Agent } & ExtraPoolOptions | ExtraPoolOptions;
 
     interface NameValuePair {
         name: string;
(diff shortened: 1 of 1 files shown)


types/request/index.d.ts
```diff
@@ -197,6 +197,10 @@ declare namespace request {
         };
     }
 
+    interface PoolOptions {
+        maxSockets?: number;
```
mastermatt: Instead of having the two long one-liners, does make more sense to move the `[key: string]: http.Agent | https.Agent` into this interface?

types/request/index.d.ts
```diff
@@ -280,7 +284,7 @@ declare namespace request {
         // set in `Request.prototype.init`
         headers: Headers;
         method: string;
-        pool: false | { [key: string]: http.Agent | https.Agent };
+        pool?: false | { [key: string]: http.Agent | https.Agent } & PoolOptions | PoolOptions;
```
mastermatt: This `?` shouldn't be added. `pool` is always set. `Request.prototype.init` sets it to the global pool if it's not provided in the options.
https://github.com/request/request/blob/master/request.js#L171-L173

mastermatt (changes requested):

types/request/index.d.ts
mrfoe7: Sure, it is my fault. I will fix to it

types/request/index.d.ts
mrfoe7: Your suggest is does not work, because `maxSockets` consider with `[key: string]` and compiler check to it case.

mastermatt: Yep, you're right.

What about:
```ts

interface ExtraPoolOptions {
  maxSockets?: number;
}

type PoolOptions = false | { [key: string]: http.Agent | https.Agent } & ExtraPoolOptions | ExtraPoolOptions;
```
I'd still like to reduce the two duplicate lines if possible.

mrfoe7: Thanks for your help

mastermatt (approved):
Thanks for your time on this @mrfoe7

orta:
Cool, this PR looks good 👍

PR #27708, opened 2018-07-30
[@types/node] Add missing props to error in `child_process.exec`

sorenlouv:
Closes #27660

PR #40896, opened 2019-12-07
[@types/node] Use ExecException for child_process.execFile()

thinca:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<following>>

https://github.com/DefinitelyTyped/DefinitelyTyped/pull/27708 changed type of error in `child_process.exec`, but `child_process.execFile` should be also change.
[`exec` just calls `execFile`](https://github.com/nodejs/node/blob/147317f60c41acdd934c6da3868f442e05226e37/lib/child_process.js#L149-L154).  They have same callbacks.

### DIFF
--- a/types/node/child_process.d.ts
+++ b/types/node/child_process.d.ts
@@ -320,25 +320,25 @@ declare module "child_process" {
     function execFile(file: string, args: ReadonlyArray<string> | undefined | null, options: ({ encoding?: string | null } & ExecFileOptions) | undefined | null): ChildProcess;
 
     // no `options` definitely means stdout/stderr are `string`.
-    function execFile(file: string, callback: (error: Error | null, stdout: string, stderr: string) => void): ChildProcess;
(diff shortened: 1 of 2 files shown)


galkin:
@thinca, could you also make changes at Node.js v10?

thinca:
@galkin OK.  I updted with force-pushed.  What about v11?

galkin:
@thinca, v11 is EOL, so we don't update @types for such versions.

galkin (approved):
Thank you for your work!

thinca:
Thank you for the review 😃

thinca:
CI seems to have failed regardless of this change.
I force-pushed a new commit without change to re-run the CI.

orta:
Hrm, I'm not sold on the accuracy of the perf hit in the stats above - given that both files have the same changes and the reported identifiers look pretty un-related. Merging.

PR #40900, opened 2019-12-07
Avoid error when dynamically generating routes

mcaneris:
This is supposed to avoid throwing an error when the routes object is the return value of a function passing additional params, rather than creating it as an object.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<<https://github.com/Paratron/hookrouter>>

### DIFF
--- a/types/hookrouter/index.d.ts
+++ b/types/hookrouter/index.d.ts
@@ -8,7 +8,7 @@ export namespace HookRouter {
 	type InterceptedPath = string | null;
 	interface AProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> { href: string; }
 	interface QueryParams { [key: string]: any; }
-	interface RouteObject { [key: string]: (params: QueryParams) => any; }
+	interface RouteObject { [key: string]: (params: QueryParams & string) => any; }
 }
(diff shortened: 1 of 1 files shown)


orta:
This doesn't feel right to me - do you maybe mean `| string` to mean that it could also be a string?

types/hookrouter/index.d.ts
```diff
@@ -8,7 +8,7 @@ export namespace HookRouter {
 	type InterceptedPath = string | null;
 	interface AProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> { href: string; }
 	interface QueryParams { [key: string]: any; }
-	interface RouteObject { [key: string]: (params: QueryParams) => any; }
+	interface RouteObject { [key: string]: (params: QueryParams & string) => any; }
```
DanielRosenwasser: This is an impossible type - what does this even mean?

DanielRosenwasser (changes requested):

mcaneris:
This was a patch request that unfortunately I didn't have time to fully review, which is my bad. 

I still don't have the necessary time to delve into this, so I'll hopefully revisit this in the future. Closing for now.

PR #40901, opened 2019-12-07
react-router: Add array option to params while generating path

sarioglu:
Add array option to params while generating path

This PR provides ability to use array type params since it is possible to define paths using modifiers like * and +

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/pillarjs/path-to-regexp#zero-or-more

### DIFF
--- a/types/react-router/index.d.ts
+++ b/types/react-router/index.d.ts
@@ -20,6 +20,7 @@
 //                 Wesley Tsai <https://github.com/wezleytsai>
 //                 Sebastian Silbermann <https://github.com/eps1lon>
 //                 Nicholas Hehr <https://github.com/HipsterBrown>
+//                 Mehmet Sarıoğlu <https://github.com/sarioglu>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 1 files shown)


eps1lon (commented):
Could you link to react-router docs or provide a small codesandbox to confirm that this is indeed supported? It's not obvious that there's a direct mapping from `generatePath` to `path-to-regexp`.

sarioglu:
> Could you link to react-router docs or provide a small codesandbox to confirm that this is indeed supported? It's not obvious that there's a direct mapping from `generatePath` to `path-to-regexp`.

Unfortunately `react-router` docs don't include an explicit example of that usage. So, I've prepared an example. You may find it via (https://codesandbox.io/s/angry-williams-ckxg0). I wish you won't judge my artistic abilities :).  `generatePath` passes `params` to `path-to-regexp` without any preprocessing. Therefore I think we can assume that there is a direct mapping between them.

eps1lon:
> Therefore I think we can assume that there is a direct mapping between them.

We can't though. Could you open an issue in react-router and ask if that is supported? I think the best way is to open a PR with a test case that is essentially doing what your codesandbox does. 

Otherwise I'm not comfortable adding types for a behavior that's neither documented nor tested nor endorsed.

sarioglu:
Makes sense 👍 I'll write some tests for both `react-router` and `@types/react-router` to make sure that it is an intentional feature.

RyanCavanaugh:
@sarioglu I'm going to close this for housekeeping purposes -- please open a fresh PR if/when you get an answer from the `react-router` folks on the correct usage of the API., Thanks!

PR #40910, opened 2019-12-09
[@types/react-query] Change type on getCanFetchMore props (allPages, lastPage)

### LINKED ISSUE #40899: [@types/react-query] Incorrect Type on Props for canGetFetchMore
I get a type error when using the `@types/react-query` package with the paginated canGetFetchMore function.

The type for QueryOptionsPaginated defines the props for canGetFetchMore (lastPage, allPages) as numbers:

```
export interface QueryOptionsPaginated<TResult> extends QueryOptions<TResult> {
    paginated: true;
    getCanFetchMore: (lastPage: number, allPages: number) => boolean;
}
```

But the source shows that the full data object for lastPage and allPages is sent as the props to canGetFetchMore:

```
...(query.config.paginated && {
                canFetchMore: query.config.getCanFetchMore(
                  data[data.length - 1],
                  data
                ),
```

Similarly, in the paginated examples, the tutorial shows the nextId prop being passed in lastPage to the fetch function, which wouldn't make sense if the prop is a number:

```
    {
      paginated: true,
      getCanFetchMore: (lastPage, allPages) => lastPage.nextId,
    }
```

Thanks!

@bostondv @lukaszfiszer @tannerlinsley

litewarp:
Change type on props in getCanFetchMore. Fixes #40899.

### DIFF
--- a/types/react-query/index.d.ts
+++ b/types/react-query/index.d.ts
@@ -38,7 +38,7 @@ export interface QueryOptions<TResult> {
 
 export interface QueryOptionsPaginated<TResult> extends QueryOptions<TResult> {
     paginated: true;
-    getCanFetchMore: (lastPage: number, allPages: number) => boolean;
+    getCanFetchMore: (lastPage: TResult, allPages: TResult[]) => boolean;
 }
 
 export interface QueryResult<TResult, TVariables> {


lukaszfiszer (changes requested):
Can you add a test for this type in paginated mode? See react-query-test.ts file

PR #40915, opened 2019-12-09
Fixing a property does not exist on type 'UploadedFile[]' for EXPRESS-FILEUPLOAD

dre4success:
This is the URL of the package -   https://github.com/richardgirges/express-fileupload

Any of this properties 
**name**: A name of the file for upload
**mv**: A function to move the file elsewhere on your server
**mimetype**: The mimetype of your file
**data**: A buffer representation of your file, returns empty buffer in case useTempFiles option was set to true

Does not exist on UploadedFile. Type script throws back an error of something like this, 
Property 'mimetype' does not exist on type 'UploadedFile | UploadedFile[]'.

Here's a link to a sample issue
https://github.com/richardgirges/express-fileupload/issues/175.

Removing | UploadedFile[] fixes the issue.

### DIFF
--- a/types/express-fileupload/index.d.ts
+++ b/types/express-fileupload/index.d.ts
@@ -22,7 +22,7 @@ declare function fileUpload(options?: fileUpload.Options): express.RequestHandle
 
 declare namespace fileUpload {
     class FileArray {
-        [index: string]: UploadedFile | UploadedFile[]
+        [index: string]: UploadedFile
     }
 
     interface UploadedFile {


Naktibalda (changes requested):
I haven't used express-fileupload since I created this definition more than 2 years ago.
For some reason I thought that it was possible to upload a list of files under one name, but I haven't used anything like that in the code I worked on back then.

If it is impossible to upload a list of files, please remove this part of tests.ts too:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3aa84acdd6feb2c1b317069e067f96b56d03a426/types/express-fileupload/express-fileupload-tests.ts#L27-L32

dre4success:
It is possible to upload a list of files. You just have to loop through them. So that part of the test still stands. 

This is a sample of multiple uploading 
```js
  if (Array.isArray(req.files.picture)) {
    if (req.files.picture.length > 6)
      return res.status(400).json({
        status: 400,
        message: `You can't upload more than 6 images at once`
      });

    for (let picture of req.files.picture) {
      if (!picture.mimetype.includes('image/'))
        return res.status(400).json({
          status: 400,
          message: `only images allowed for upload`
        });
    }
    let gallery = await Promise.all(
      req.files.picture.map(async single => {
        let image = await imageUploader(single);
        return image;
      })
    );
    let many = await galleryCollection.insertOne({
      title: req.body.title,
      image: gallery
    });
    return res.status(200).json({
      status: 200,
      data: {
        many,
        message: `Image uploaded successfully`
      }
    });
  }

async function imageUploader(picture: any) {
  const image = await Jimp.read(picture.data);
  let buffer = await image
    .resize(1200, Jimp.AUTO)
    .quality(90)
    .getBufferAsync(Jimp.MIME_JPEG);

  let resource = `${await crypto.randomBytes(8).toString('hex')}.jpeg`;

  const params = {
    Bucket: 'AWS-Bucket',
    ACL: 'public-read',
    Key: resource,
    Body: buffer,
    ContentType: picture.mimetype
  };
  const data = await s3.upload(params).promise();

  return `${cloudfront}/${data.Key}`;
}
```

Naktibalda:
@dre4success then I think that removing `| UploadFile[]` from dictionary is incorrect.

The app code must identify what kind of input it got by either running `Array.isArray`
or `isUploadedFile` function similar to the one I used in the test file:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3aa84acdd6feb2c1b317069e067f96b56d03a426/types/express-fileupload/express-fileupload-tests.ts#L11-L13

Naktibalda:
or cast the variable explicitly as in the linked issue if lists of files are not expected:
```js
const foo = req.files.foo as UploadedFile
```

dre4success:
Casting the variable explicitly for a single file works.

Naktibalda:
@dre4success If I were you, I would test what happens when somebody sends a list of files where a single file is expected. 
Probably it can crash your app, it would be safer to check with `Array.isArray` and handle array and single file cases differently, even if all you do is:
```js
if (Array.isArray(req.files.foo)) {
  throw new Error('Please upload only one file');
}
```

PR #40916, opened 2019-12-09
fix(react-tracking): Allow tracking info function to return Falsy value

drbalm:
React Tracking allows us to pass a function to the `@track` annotation. The function is passed props, state, etc. and returns tracking data.

Per the [React Tracking documentation](https://github.com/nytimes/react-tracking), this function can optionally return a falsy value in which case no tracking is performed: 

> If the function returns a falsy value (e.g. false, null or undefined) then the tracking call will not be made.

The current typing does not seem to support this case because the function is expected to return a value of type `T`. This PR updates the type to also support a Falsy return. (This is similar to what has been done already in the `process` function of the `Options` interface.)

I'm seeing an error even against master when I run the linting. Perhaps we need a separate PR to address that?

> Error: Errors in typescript@next for external dependencies:
../react/index.d.ts(33,22): error TS2307: Cannot find module 'csstype'.

We have been running with this change in our project for some time without any issues. I have also verified the expected behavior in VSCode without and then with this change:

**With Types Before Fix Results in Error:**
Screen Shot 2019-12-06 at 10 02 36 AM

**Without Types Before Fix Results in Error:**
Screen Shot 2019-12-06 at 10 02 58 AM

**No Error With Types After Fix :**
Screen Shot 2019-12-06 at 10 03 24 AM

**No Error Without Types After Fix :**
Screen Shot 2019-12-06 at 10 03 27 AM

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nytimes/react-tracking (see quote above)


types/react-tracking/test/react-tracking-without-types-tests.tsx
```diff
@@ -13,6 +13,11 @@ function customEventReporter(data: { page?: string }) {}
 )
 class ClassPage extends React.Component<any> {
     @track({ event: 'Clicked' })
+    @track((_props, _state, [e]: [React.MouseEvent]) => {
```
damassi: @drbalm - this file is supposed to test functionality without types.

damassi (approved):
@drbalm - One minor change but otherwise LGTM 👍

damassi:
@alloy - if you have a sec!

types/react-tracking/test/react-tracking-without-types-tests.tsx
drbalm: @damassi, *doh*, thanks for catching this! 😄 

I removed the typing in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40916/commits/3f373a780d6809d7808bd3e15b4247d7cb171e61

orta:
👋 - I'll give @alloy some time, but if not this does look good from my perspective too

alloy:
Yup, great–thanks! 🙏

PR #40923, opened 2019-12-09
Create separate versions for different react-native versions

acoates-ms:
Moving to the structure that allows us to have separate type definitions for multiple versions.  This will allow us to update the root types for the next version when changes are made to react-native master.  Rather than having to wait until the next release of react-native, when type changes have been forgotten.

Other than copying files, the only changes I've made is to the version numbers in index.d.ts in the various folders.  (0.61 in the v0.61 folder, and 0.62 in the v0.62)

Please fill in this template.

- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present). - Already fails prior to change

### DIFF
--- a/types/react-native/v0.60/BatchedBridge.d.ts
+++ b/types/react-native/v0.60/BatchedBridge.d.ts
@@ -1,0 +1,23 @@
+interface SpyData {
+    type: number;
+    module?: string;
+    method: string | number;
+    args: any[];
+}
+
+declare class MessageQueue {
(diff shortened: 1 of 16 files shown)


types/react-native/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for react-native 0.60
+// Type definitions for react-native 0.62
```
alloy: We can’t make this change yet until 0.62 has a stable release.

PR #40912, opened 2019-12-09
[@types/node] Add close event to FSWatcher

JIACHENG9:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [ ] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes.

> Event: 'close'#
Added in: v10.0.0
Emitted when the watcher stops watching for changes. The closed fs.FSWatcher object is no longer usable in the event handler.

https://nodejs.org/dist/latest-v13.x/docs/api/fs.html#fs_event_close
https://nodejs.org/dist/latest-v12.x/docs/api/fs.html#fs_event_close
https://nodejs.org/dist/latest-v11.x/docs/api/fs.html#fs_event_close
https://nodejs.org/dist/latest-v10.x/docs/api/fs.html#fs_event_close

PR #26034, opened 2018-05-25
declare SymbolConstructor.observable

phra:
if `SymbolConstructor.observable` symbol is not defined in the types, the inclusion of the latest rxjs (v6) breaks the compilation in a strict environment.

Please fill in this template.

- [ ] Use a meaningful title for the pull request. Include the name of the package modified.
- [ ] Test the change in your own code. (Compile and run.)
- [ ] Add or edit tests to reflect the change. (Run with `npm test`.)
- [ ] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [ ] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [ ] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: see https://github.com/ReactiveX/rxjs/issues/3674
- [ ] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #40927, opened 2019-12-09
feat(node): v13

SimonSchick:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V13.md#13.1.0

This ended up being quite a large PR and I want some feedback on if we can proceed with this or if this needs to be broken down further.

General list of changes:
- Changed constants to be re-exported in `constants` module to avoid duplication.
- BREAKING: Removed various es2015, 16, 17 shims (requires es2018 lib now).
- Various type restrictions in API that could be considered breaking changes.
- Aliased various global interface into `NodeJS` namespace.
- Scheduling in `cluster`
- Use of `unique symbol`
- Removal of redundant interfaces
- BREAKING: Replaced global `EventEmitter` class with interface as it's purely virtual and only there for global declarations.
  - This required various updates in other package since they sometimes used `extends EventEmitter` or `extends NodeJS.EventEmitter`.

Retry of https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40438


SimonSchick:
@orta or anyone else, can you mark this as actually *passing*, the parallel npm install keeps breaking and it's incredibly frustrating.

SimonSchick:
@Flarna for another round, nothing has really changed.

Flarna:
Some dgram and customRequire test are missing in node12 folder. Also the #40912 merged a few minutes is missing.

SimonSchick:
Hnnng.

SimonSchick:
@Flarna can you elaborate on what is missing exactly?

types/node/child_process.d.ts
```diff
@@ -3,6 +3,16 @@ declare module "child_process" {
     import * as net from "net";
     import { Writable, Readable, Stream, Pipe } from "stream";
 
+    type Serializable = string | object | number | boolean;
+    type SendHandle = net.Socket | net.Server;
+
+    interface ChildProcessIPC extends ChildProcess {
```
Flarna: it's possible to create IPC channels also via `spawn()` by specifying e.g `[0, 1, 2, 'ipc']`, therefore `ChildProcessIPC` is not a special case for `fork()` only

Flarna:
I expect  #40912 is quite clear. the remaining part is:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/node-tests.ts#L13
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/node-tests.ts#L821-L837
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/node-tests.ts#L859-L939

SimonSchick:
The dgram changes are not applicable since these are v13 specific.

SimonSchick:
@Flarna can you perhaps make a PR against my branch, my head is spinning a little 😛

SimonSchick:
Nvm I think I got it together now.

types/node/worker_threads.d.ts
```diff
@@ -60,13 +60,21 @@ declare module "worker_threads" {
...
+        maxOldGenerationSizeMb: number;
+        codeRangeSizeMb: number;
     }
 
     class Worker extends EventEmitter {
         readonly stdin: Writable | null;
         readonly stdout: Readable;
         readonly stderr: Readable;
         readonly threadId: number;
+        readonly resourceLimits?: ResourceLimits;
```
Flarna: I think `resourceLimits` is always present but each property may be missing (you get `{}` if no limit was given).
Therefore change `ResourceLimits ` also to
```
interface ResourceLimits {
    maxYoungGenerationSizeMb?: number;
    maxOldGenerationSizeMb?: number;
    codeRangeSizeMb?: number;
}
```
Once again this fits fine to options as the whole `resourceLimits` property is optional but if it is present each limit property is optional.

SimonSchick: So it would be `readonly resourceLimits: Partial<ResourceLimits>;`?

Flarna: no, `ResourceLimits` should be changed to all properties optional. This matches in both cases.

types/node/module.d.ts
```diff
@@ -1,3 +1,19 @@
 declare module "module" {
-    export = NodeJS.Module;
+    import { URL } from "url";
+    interface Module extends NodeJS.Module {}
```
Flarna: Seems the move of `Module` causes problems in one of my applications.
I augment `Module` with internals but I'm not able to adapt it in a way compatible to your changes:
```
declare namespace NodeJS {
  namespace Module {
    function _load(request: string, parent: NodeModule | undefined, isMain?: boolean): any;
    function _resolveFilename(request: string, parent: NodeModule | undefined): string;
    var _cache: ModuleCache;
  }
}
```

Flarna: I'm able to adapt my code by moving the augmentation into `module "module"` and change from `import Module = require("module");` to `import { Module } from "module";`
```
declare module "module" {
  namespace Module {
    function _load(request: string, parent: NodeModule | undefined, isMain?: boolean): any;
    function _resolveFilename(request: string, parent: NodeModule | undefined): string;
    var _cache: NodeJS.ModuleCache;
  }
} 
```

SimonSchick:
@rbuckton plss

SimonSchick:
No idea what the deal with the build failure with `parse` is, it worked a week ago and I have made no further changes there.

Flarna (approved):

SimonSchick:
I'm going to need some assistance here, this PR only covers up until 13.2 and 13.4 was already released, I really want to avoid falling behind but the lack of support is making this very hard.

Flarna:
I think the best approach is to get the basics merged and add more changes later. The current PR is already very big and therefore hard to keep merge able.

orta:
Can you try rebase? I know there were some timeout fixes for CI last week

SimonSchick:
@orta I rebased this with my last push

SimonSchick:
Some other PRs had been merged since my last push, namely https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41170 (cc @yahehe) and https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41029 (cc @andrewbranch).
These changes might be lost for the v12 typings and have to be re-applied in the `v12` folder accordingly.

I tried to avoid this by constantly rebasing but didn't have the time in the last couple days.

andrewbranch:
@SimonSchick I’ll be around until ~3 PM PST today—if you need to do a time-sensitive update, ping me and I’ll prioritize it.

SimonSchick:
@andrewbranch not until after christmas, if you have the time I suggest you simply apply the 2 commits manually/directly to the v12 folder, should be ~5-10min of work.

andrewbranch:
There’s quite a long backlog of PRs to get through today, so I probably won’t get around to that. Trying to knock it down as low as I can before taking the next two days off.

SimonSchick:
@andrewbranch nope, only PRs that affected node are:

```
1d9e9b1590 [@types/node] Add missing `method` property to `http.ClientRequest` (#41029)
46370cf94a Fix node/crypto.verify return type to boolean from Buffer (#41170)
```

andrewbranch:
Sorry, I meant that I have a long backlog of PRs to review across all of DT 😄

SimonSchick:
Ah ok :)

orgads:
@SimonSchick Is the removal of `SymbolConstructor.observable` in globals.d.ts intentional? It was added in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/26034 and moved to v12 in this change, removing support on node 14.

PR #40929, opened 2019-12-09
Revert "aws-lambda - removes the wrong assumption that headers are always set"

simonbuchan:
Reverts DefinitelyTyped/DefinitelyTyped#40876 due to breaking change complaints.

I happen to think this is the right thing to do, but I'm happy to hear this out.

### DIFF
--- a/types/aws-lambda/aws-lambda-tests.ts
+++ b/types/aws-lambda/aws-lambda-tests.ts
@@ -125,7 +125,7 @@ strOrUndefined = apiGwEvtReqCtx.routeKey;
 
 /* API Gateway Event */
 strOrNull = apiGwEvt.body;
-str = apiGwEvt.headers!['example'];
+str = apiGwEvt.headers['example'];
 str = apiGwEvt.multiValueHeaders['example'][0];
 str = apiGwEvt.httpMethod;
 bool = apiGwEvt.isBase64Encoded;
--- a/types/aws-lambda/index.d.ts
+++ b/types/aws-lambda/index.d.ts
@@ -79,7 +79,7 @@ export interface APIGatewayEventRequestContext {
 // API Gateway "event"
 export interface APIGatewayProxyEvent {
     body: string | null;
-    headers?: { [name: string]: string };
+    headers: { [name: string]: string };
     multiValueHeaders: { [name: string]: string[] };
     httpMethod: string;
     isBase64Encoded: boolean;


simonbuchan:
Maybe give this one at least a day for feedback?

hedefalk:
Hi,

I did the PR that this reverts so I guess I should chime in. I might have missed to be clear on the fact that it was a breaking change. I guess therefore the version wasn't upped appropriately and therefore merits complaints? 

I however still firmly believe this is a change that need to get in there. The point of having types is to catch these kinds of problems, right? Instead of a compile error I got a runtime NPE. That's exactly the kind of things I want to catch at compile time. 

Saying that the  AWS console is doing it wrong by passing null - I definitely agree they conceptually might be, but they are simply one of the _truths_ though?

alex-bolenok-centralreach:
AWS in general is sloppy about nulls, required/optional values and such, especially when multiple event sources are involved.

For instance, an SNS event is one thing when you subscribe a lambda directly to SNS and a whole nother thing when you pipe it though SQS. It's almost the same, but some properties are missing and capitalization in property names is off. A major PITA!

We can try to keep up with them (say, define different types for different event sources, like in this case), or, alternatively, just make the types as permissive as possible, like in the original PR.

This would require writing more runtime type guards and assertions than usual, but this is probably a good thing anyway, given that you can't really trust anything that comes in.

orta:
OK, conversation looks like it's landed on reverting is the right thing to do here. Merging 👍

hedefalk:
@orta, My reading is that there is still the original change should be there, but not just in a point release, should I re-PR for a major release?

orta:
Doesn't look like the types semver is trying to stay in-step with the [SDK's](https://github.com/aws/aws-sdk-js/releases) - seems fine to me

dnalborczyk:
@orta  I don't think `@types/aws-lambda` is tracking [aws-sdk-js](url) (which provide it's own types) nor any npm module.

simonbuchan:
It's tracking lambda node runtime changes, in that those are when the handler API has changed before, but AWS didn't change anything in v10. I'm considering breaking up the package into a bunch of per-service packages and a core handler package, which doesn't seem to be too much work, but the main issue is that it loses all attribution at that point, without a bunch of real ugly git hackery. If I did that though, it would make it far easier to making even more sweeping breaking changes than this.

PR #40887, opened 2019-12-06
chore(ioredis): merge callback and promise signatures

SimonSchick:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/luin/ioredis
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

This is still somewhat WIP.

I'm trying to fix various issues while retaining type safety.
Unfortunately many of those changes result in breaking changes due to TS limitations.

Fixes #40324
Fixes #35566
FIxes #34662
Fixes #28220
Fixes #40930
Fixes #27784 (technically fixed by previous PR but hey :) )
Fixes #27391 (Not fixed by me but this but it's easier to bulk close
Fixes #23637 (This was also fixed a while ago but can be closed)

PR #40930, opened 2019-12-10
[@types/ioredis]: Add additional definitions for Cluster

kyleamazza:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://npmdoc.github.io/node-npmdoc-ioredis/build/apidoc.html
https://github.com/luin/ioredis

### DIFF
--- a/types/ioredis/index.d.ts
+++ b/types/ioredis/index.d.ts
@@ -963,6 +963,8 @@ declare namespace IORedis {
 
         del(...keys: KeyType[]): Promise<number>;
 
+        unlink(...keys: KeyType[]): Promise<number>;
+
         incr(key: KeyType, callback: (err: Error, res: number) => void): void;
         incr(key: KeyType): Promise<number>;
 
--- a/types/ioredis/index.d.ts
+++ b/types/ioredis/index.d.ts
@@ -965,6 +965,9 @@ declare namespace IORedis {
 
         unlink(...keys: KeyType[]): Promise<number>;
 
+        substr(key: KeyType, start: number, end: number, callback: (err: Error, res: string) => void): void;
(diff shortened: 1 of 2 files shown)


plantain-00 (approved):

types/ioredis/index.d.ts
```diff
@@ -948,14 +949,33 @@ declare namespace IORedis {
         setnx(key: KeyType, value: ValueType, callback: (err: Error, res: number) => void): void;
         setnx(key: KeyType, value: ValueType): Promise<BooleanResponse>;
 
+        setex(key: KeyType, seconds: number, value: ValueType, callback: (err: Error, res: Ok) => void): void;
+        setex(key: KeyType, seconds: number, value: ValueType): Promise<Ok>;
```
SimonSchick: These are technically wrong, you can provide a callback AND get a promise at the same time, see my PR.

SimonSchick (changes requested):

types/ioredis/index.d.ts
SimonSchick: Also it would probably make more sense to move these methods into a common interface for Redis and Cluster rather than duplicating them.

SimonSchick (changes requested):

types/ioredis/index.d.ts
kyleamazza: Ah I see, I could wait for #40887 to be merged, then integrate those changes into this branch?

Also, just to be clear on the common interface, you mean creating a separate interface, and having Redis and Cluster extend them, right?

SimonSchick: Yes.

PR #40934, opened 2019-12-10
fix(ember-data): improve belongsTo and hasMany relationship typing

makepanic:
Please fill in this template.

  * `remoteType` return type: [hasMany](https://github.com/emberjs/data/blob/v3.14.0/packages/store/addon/-private/system/references/has-many.js#L76-L81), [belongsTo](https://github.com/emberjs/data/blob/v3.14.0/packages/store/addon/-private/system/references/reference.ts#L76-L82) (Note the ember-data type also declares `identity` but it doesn't seem to be used)
  * I made BelongsToReference and HasManyReference generic (HasManyReference had some generic parts already). We might be able to narrow the generic type further by asserting that `T extends Model` but I'm not sure if it's too strict.
  * `hasMany` and `belongsTo` now extracts the type by extracting it from the instance. This can create false-positives as it can't decide if a field is an attribute or a relationship. This drawback [already existed previously](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/ember-data/index.d.ts#L26).
  * __Note__: this also adjusts the tsconfig to enable checking of `belongsTo` and `hasMany` when used as a decorator

### DIFF
--- a/types/ember-data/index.d.ts
+++ b/types/ember-data/index.d.ts
@@ -25,6 +25,18 @@ type ModelKeys<Model extends DS.Model> = Exclude<keyof Model, keyof DS.Model>;
 type AttributesFor<Model extends DS.Model> = ModelKeys<Model>; // TODO: filter to attr properties only (TS 2.8)
 type RelationshipsFor<Model extends DS.Model> = ModelKeys<Model>; // TODO: filter to hasMany/belongsTo properties only (TS 2.8)
 
+// infer the generic type of hasMany when used as computed or decorator
+type UnpackManyArray<T> =
+    T extends DS.ManyArray<infer X> ? X :
+    T extends DS.PromiseManyArray<infer Y> ? Y :
+    T extends Ember.ComputedProperty<DS.PromiseManyArray<infer Z>, any> ? Z :
+    T;
+
+// used to infer the generic type of belongsTo when used as computed or decorator
+type UnpackComputedProperty<T> =
+    T extends Ember.ComputedProperty<infer Y> ? Y:
+    T;
+
 export interface ChangedAttributes {
     [key: string]: [any, any] | undefined;
 }
@@ -546,11 +558,11 @@ export namespace DS {
         /**
          * Get the reference for the specified belongsTo relationship.
          */
-        belongsTo(name: RelationshipsFor<this>): BelongsToReference;
+        belongsTo<T extends keyof this = RelationshipsFor<this>>(name: T): BelongsToReference<UnpackComputedProperty<this[T]>>;
         /**
          * Get the reference for the specified hasMany relationship.
          */
-        hasMany(name: RelationshipsFor<this>): HasManyReference<any>;
+        hasMany<T extends keyof this = RelationshipsFor<this>>(name: T): HasManyReference<UnpackManyArray<this[T]>>;
         /**
          * Given a callback, iterates over each of the relationships in the model,
          * invoking the callback with the name of each relationship and its relationship
@@ -725,13 +737,13 @@ export namespace DS {
      * addon author to perform meta-operations on a belongs-to
      * relationship.
      */
-    class BelongsToReference {
+    class BelongsToReference<T> {
         /**
          * This returns a string that represents how the reference will be
          * looked up when it is loaded. If the relationship has a link it will
          * use the "link" otherwise it defaults to "id".
          */
-        remoteType(): string;
+        remoteType(): 'id' | 'link';
         /**
          * The `id` of the record that this reference refers to. Together, the
          * `type()` and `id()` methods form a composite key for the identity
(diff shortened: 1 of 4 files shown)


chriskrycho (changes requested):
Thanks for opening this. Happy to iterate on it with you!  👍 At a minimum we need *much* more type-test coverage over this. I'm marking as **Request Changes** until—

- @mike-north and I have both had a chance to test and review
- additional test coverage is in place
- we as a community make sure that this doesn't introduce performance regressions when used in large real-world apps (a real danger with sophisticated conditional types)

Again, thank you for picking this up and getting some motion on it—I've wanted this particular thing to move forward for a long time!

makepanic:
as mentioned in discord, I'd also be ok with removing the "cleverness" and omit automatic type extraction. 

Having a generic `BelongsToReference` would make things easier than they currently are

types/ember-data/index.d.ts
```diff
@@ -546,11 +558,11 @@ export namespace DS {
         /**
          * Get the reference for the specified belongsTo relationship.
          */
-        belongsTo(name: RelationshipsFor<this>): BelongsToReference;
+        belongsTo<T extends keyof this = RelationshipsFor<this>>(name: T): BelongsToReference<UnpackComputedProperty<this[T]>>;
```
mike-north: Unfortunately `this` used as a typeparam has been a source of _significant breaking changes_  to our types as new compiler versions are released. The fact that it includes `protected` and `private` members, when used internally, but those _should not be visible externally_ results in some significant nuance that few other type packages seem to encounter (i.e., the ember types have been one of only 5-6 packages that break routinely around compiler changes that affect `this`)

As the person who often has to urgently dig us out of these kinds of problems, I'd prefer to 

(a) not use `<this>` or
(b) be ready to, at a moments notice, eject out to a _less clever_ typing in the event a new TS nightly build starts to block our CI pipeline

types/ember-data/test/belongs-to.ts
```diff
@@ -7,11 +7,17 @@ class Folder extends DS.Model {
...
+    @DS.belongsTo('owner') owner!: Owner;
+}
+
+class Owner extends DS.Model {
 }
 
 declare module 'ember-data/types/registries/model' {
     export default interface ModelRegistry {
         folder: Folder;
+        owner: Owner;
```
mike-north: Could we name this something else? I had to hunt around to make sure that this wasn't the `owner` from RFC-232 tests (i.e., `this.owner` (ApplicationInstance) )

mike-north (changes requested):

types/ember-data/index.d.ts
makepanic: What do you think about removing the `this` usage for belongsTo/hasMany and simply make HasManyReference and BelongsToReference generic?

I also hit some issues with the old `RelationshipsFor<this>` 

e.g.
```
belongsTo<T extends DS.Model>(name: string): BelongsToReference<T>;
hasMany<T extends DS.Model>(name: string): HasManyReference<T>;
```

types/ember-data/test/belongs-to.ts
makepanic: Sure. I've renamed it.

types/ember-data/index.d.ts
mike-north: I'd consider this a possible misuse of generics. Generally they're intended to be used as a tool for expressing a relationship between two or more types (expressed in terms of one or more typeparams), and in the case above you use it in only once.

What you have here is just a casting mechanism where we'd have to provide a typeparam at every usage site. If we're going to encourage this kind of casting, I'd prefer that it _looks like_ casting (i.e., using the `as` keyword)

From a type safety standpoint,
```ts
belongsTo<T extends DS.Model>(name: string): BelongsToReference<T>;

this.belongsTo<Book>('book');
```
is no more type safe than 
```ts
belongsTo(name: string): BelongsToReference<DS.Model>;

this.belongsTo('book') as Book;
```

makepanic: Yes that makes sense. 

For us there were multiple places where the current api is really verbose if one wants to type it. e.g. the following is needed right now:

```
const ref = model.belongsTo('rel') 
ref.load() as Promise<Rel | null>;
ref.value() as Rel | null;
ref.reload() as Promise<Rel | null>;
```

Which could be changed to:
```
const ref = model.belongsTo('rel') as BelongsToRef<Rel>;
ref.load() // Promise<Rel | null>;
ref.value() // Rel | null;
ref.reload() // Promise<Rel | null>;
```

Which would be possible if we make `BelongsToReference` generic in the same way `HasManyReference` is currently generic. 

Would you be ok with that?

makepanic:
I've updated the PR and removed all the type inference.

Now the PR only changes BelongsToReference to be generic in the same way HasManyReference already was.

sandersn:
@mike-north mind taking another look? I kind of agree with you that a type parameter on BelongsToReference is suspicious, but @makepanic has a good point that the current code leads to a lot of casting. But the decision to merge or not is up to you.

sandersn (changes requested):
Requesting changes so the bot understands the current state.

chriskrycho:
cc. also @jamescdavis

PR #40939, opened 2019-12-10
Node: `UrlObject`: omit `path`

OliverJAsh:
Incorrect:

```ts
url.format({ path: '/foo?a=b' })
// => ''
```

Correct:

```ts
url.format({ pathname: '/foo', search: '?a=b' })
// => '/foo?a=b'
```

---

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://devdocs.io/node/url#url_url_format_urlobject


islishude (changes requested):
see https://nodejs.org/dist/latest-v12.x/docs/api/url.html#url_urlobject_path

OliverJAsh:
@islishude IIUC, `path` exists on the _output_ of `url.parse`, but it is never used as a parameter inside of `url.format`. In that case, `search` and `pathname` must be used instead.

The `UrlObject` type in this project reflects the latter part only, as indicated by the comment immediately above the type.

For that reason I think `path` should be considered an invalid parameter and thus removed from the `UrlObject` type?

If `UrlObject` is used in more places than just the input of `url.format`, then we need a separate type.

islishude:
url.format accepts [url.URL class](https://nodejs.org/dist/latest-v12.x/docs/api/url.html#url_the_whatwg_url_api) instead of [url. urlObject](https://nodejs.org/dist/latest-v12.x/docs/api/url.html#url_legacy_url_api)

islishude:
FYI 
https://github.com/nodejs/node/blob/525fd9ca7c6146d15b4f73dcdb3854b9c63e8fbb/lib/url.js#L536

OliverJAsh:
That's not what it says here: https://devdocs.io/node/url#url_url_format_urlobject

I think you're talking about a different thing under the same name: https://devdocs.io/node/url#url_url_format_url_options

types/node/node-tests.ts
```diff
@@ -21,6 +21,10 @@ import Module = require("module");
     {
         url.format(url.parse('http://www.example.com/xyz'));
 
+        // TODO: why won't this work?
```
galkin: Please remove this comment

types/node/url.d.ts
```diff
@@ -8,7 +8,6 @@ declare module "url" {
         host?: string | null;
         hostname?: string | null;
         href?: string | null;
-        path?: string | null;
```
galkin: Please add similar change  to node v10

galkin (changes requested):
Thank you for your pull. According my tests and checking documentation https://nodejs.org/dist/latest-v12.x/docs/api/url.html#url_url_format_urlobject
url.format doesn't work with `path`. `pathname` should be used.

types/node/test/url.ts
```diff
@@ -0,0 +1,153 @@
+import assert = require("assert");
```
OliverJAsh: I had to move the tests here to workaround https://github.com/DefinitelyTyped/DefinitelyTyped/issues/40967

types/node/test/url.ts
```diff
@@ -0,0 +1,153 @@
...
+        pathname: 'search',
+        query: { q: "you're a lizard, gary" }
+    });
+
+    const myURL = new url.URL('https://a:b@你好你好?abc#foo');
+    url.format(myURL, { fragment: false, unicode: true, auth: false });
+
+    // `format doesn't work with `path`: use `pathname` and `search` instead
+    // $ExpectError
+    url.format({ path: '/foo' });
```
OliverJAsh: This is the only new test

OliverJAsh:
@galkin Done

types/node/v11/test/url.ts
```diff
@@ -0,0 +1,5 @@
+import * as url from 'url';
+
+// `format doesn't work with `path`: use `pathname` and `search` instead
+// $ExpectError
+url.format({ path: '/foo' });
```
galkin: We don't update EOL version. v11 is EOL. Please remove this changes.

types/node/v11/tsconfig.json
```diff
@@ -26,7 +26,8 @@
         "test/util.ts",
         "test/v8.ts",
         "test/worker_threads.ts",
-        "test/zlib.ts"
```
galkin: The same.

types/node/v11/url.d.ts
```diff
@@ -7,7 +7,6 @@ declare module "url" {
         host?: string;
         hostname?: string;
         href?: string;
-        path?: string;
```
galkin: The same.

galkin (changes requested):

galkin (approved):

islishude (approved):

trivikr (approved):

DanielRosenwasser:
The slowdown is likely a fluke - maybe we should re-run tests. Any easy way to do that @sandersn? Also, would be nice if there didn't need to be a new file to work around that linked issue.

OliverJAsh:
@DanielRosenwasser Merged master in so that should force the tests to re-run 🤞

r3nya (approved):

andrewbranch:
Merge conflicts ☹️

Issue #39358, opened 2019-10-23
@types/mongodb - Generics does not work with v3.2.x or v3.3.x


lirbank:
Using generics to set the schema does not seem to work since v3.2.x of @types/mongodb

See this example using @types/mongodb v3.1.32 (works):
https://codesandbox.io/s/typesmongodb-3132-works-nmxbc

See this example using @types/mongodb v3.3.6 (type error):
https://codesandbox.io/s/typesmongodb-336-type-error-onshs

Reproduction (same as CodeSandbox):
```ts
import { Db, Collection } from "mongodb";

export default class Model<Schema extends { _id: string }> {
  private readonly db: Db;
  private readonly collection: Collection<Schema>;

  public constructor(db: Db, collectionName: string) {
    this.db = db;
    this.collection = db.collection(collectionName);
  }

  public findOneById(id: string) {
    // This gives the type error below on @types/mongodb v3.3.6
    return this.collection.findOne({ _id: id });
  }
}
```

TS Error:
```
Argument of type '{ _id: string; }' is not assignable to parameter of type 'FilterQuery<Schema>'.
  Type '{ _id: string; }' is not assignable to type '{ [P in keyof Schema]?: MongoAltQuery<Schema[P]> | QuerySelector<MongoAltQuery<Schema[P]>> | undefined; }'.
```

Authors: @CaselIT @alanmarcell @bitjson @dante-101 @mcortesi @EnricoPicci @AJCStriker @julien-c @daprahamian @denys-bushulyak @BastienAr @sindbach @geraldinelemeur @jishi @various89 @angela-1 @hector7 @floric @erikc5000 @Manc @jloveridge @ranguna @HosseinAgha @albertossilva

HosseinAgha:
Hi @lirbank,  
This looks to me like a TypeScript [mapped types](https://www.typescriptlang.org/docs/handbook/advanced-types.html#mapped-types) bug.  
I can confirm that this does not work:  
```typescript
  const genericFunc = async <Schema extends { _id: string }>() => {
    type a = { [P in keyof Schema]: Schema[P] };
    const b: a = { _id: 'sd' };
  }
```
And this does work:  
```typescript
  type Schema2 = { _id: string };
  type a2 = { [P in keyof Schema2]: Schema2[P] };
  const b2: a2 = { _id: 'sd' };
```
We've implement `collection.find` types very similar with above code.  
[...]

PR #40943, opened 2019-12-10
feat(@types/mongodb): better support for Schemas with "any" indexed type

HosseinAgha:
Please fill in this template.

This PR fixes 2 things:
  - Add better support for Schemas with indexed type set to `any` (tests included)
```typescript
  interface IndexTypeTestModel {
    stringField: string;
    numberField?: number;
    [key: string]: any;
  }
```
  - Change default schema type from `any` into `{ [key: string]: any }` as MongoDB documents can only be objects (we accepted numbers, etc before). Some new tricky bugs in the tests surfaced as the result.

### DIFF
--- a/types/mongodb/test/collection/insertX.ts
+++ b/types/mongodb/test/collection/insertX.ts
@@ -7,15 +7,42 @@ async function run() {
   const client = await connect(connectionString);
   const db = client.db('test');
 
-  // test insertOne results
-  db.collection('test-insert').insertOne({ a: 2 }, (err, result) => {
-    result.insertedCount; // $ExpectType number
-    result.insertedId; // $ExpectType ObjectId
-    result.result.n; // $ExpectType number
-    result.result.ok; // $ExpectType number
-  });
+  const anyCollection = db.collection('test-any-type');
 
-  // test with collection type
+  // should accept any _id type when it is not provided in Schema
+
+  /**
+   * test no collection type ("any")
+   */
+  // test insertOne results
+  anyCollection.insertOne({ a: 2 }, (err, result) => {
+    result.insertedCount; // $ExpectType number
+    result.insertedId;    // $ExpectType any
+    result.ops[0].a;      // $ExpectType any
+    result.result.n;      // $ExpectType number
+    result.result.ok;     // $ExpectType number
+  });
+  // test insertMany results
+  anyCollection.insertMany([{ a: 2 }], (err, result) => {
+    result.insertedCount; // $ExpectType number
+    result.insertedIds;   // $ExpectType { [key: number]: any; }
+    result.ops[0].a;      // $ExpectType any
+    result.result.n;      // $ExpectType number
+    result.result.ok;     // $ExpectType number
+  });
+  // should accept _id with ObjectId type
+  const insertManyWithIdResult = await anyCollection.insertMany([{ _id: new ObjectId(), a: 2 }]);
+  insertManyWithIdResult.insertedCount; // $ExpectType number
+  insertManyWithIdResult.insertedIds;   // $ExpectType { [key: number]: any; }
+  insertManyWithIdResult.ops[0].a;      // $ExpectType any
+  insertManyWithIdResult.result.n;      // $ExpectType number
+  insertManyWithIdResult.result.ok;     // $ExpectType number
+  // should accept any _id type when it is not provided in Schema
+  await anyCollection.insertMany([{ _id: 12, a: 2 }]);
+
+  /**
+   * test with collection type
+   */
   interface TestModel {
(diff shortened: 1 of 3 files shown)


HosseinAgha:
This can generate quite a lot of complaints from users that relied on the default Schema being `any` and used Polymorphism (__everything extends any__) to change the type of `findX` return value:  

```typescript
// this is OK  as users types is implicitly any
const users = await collection.users.find().toArray();
// this is Wrong as `User` does not extend `Record<string, any>` or `{ [k: string]: any };`
const users: User[] = await collection.users.find().toArray();
// users should do this
const users = await collection.users.find().toArray() as User[];
```

IMHO using `any` for the returned type of `findX` is wrong as we know that it returns an object (`Record<string, any>` to be specific).

HosseinAgha:
__Maintainers I need some other owners' approval before merging this. Please don't merge until someone reviews this__

CaselIT:
does by any change this also fixes https://github.com/DefinitelyTyped/DefinitelyTyped/issues/39358 ?

CaselIT (approved):
I agree regarding the change of default.

HosseinAgha:
No I think #39368 is still a TypeScript bug. I opened a proper issue explaining the bug: https://github.com/microsoft/TypeScript/issues/35647

HosseinAgha:
Dear Maintainers, I think this is ready to merge now!

CaselIT:
ok

>  I don't know any workaround to solve it

I also don't. 

Since the typescript issue has been closed, maybe the best way to solve it would be to have an issue here where we explain the problem we are trying to solve and the current implementation and tag one of the typescript devs for an advice?

HosseinAgha:
@CaselIT Good idea. I don't think #39358 have anything to do with this PR. 
I don't think the #39358 is resolvable, I spent some time discussing the issue please take a look at my comment https://github.com/DefinitelyTyped/DefinitelyTyped/issues/39358#issuecomment-566574462 and my issue https://github.com/microsoft/TypeScript/issues/35647#issuecomment-566584170

HosseinAgha:
Regarding this PR, there were some issues after changing DefaultSchema from `any` to `{ [key: string]: any }`. I've added some more test to have more coverage. I'm going to push some fixes today.

HosseinAgha:
I reverted my code to use `any` as `DefaultSchema` I think it generates too much trouble for us.  
I also added some tests for some edge cases. The only scenario that I could not cover is when user provides a Schema with indexed types and a required _id.  
```ts
  interface IndexTypeTestModelWithObjectId {
    _id: ObjectId;
    stringField: string;
    numberField?: number;
    [key: string]: any;
  }
```
I could not make `_id` optional in this scenario (for insertX methods).
I think this is almost impossible to support with the current version of TypeScript.

HosseinAgha:
Anyway, I think we are ready to merge! Thank you for your review @CaselIT.

types/mongodb/test/collection/insertX.ts
```diff
@@ -45,21 +78,154 @@ async function run() {
...
+    randomFiel2: 32,
+  });
+  const indexTypeResultMany1 = await indexTypeCollection1.insertMany([
+    { stringField: 'hola', numberField: 0 },
+    { _id: new ObjectId, stringField: 'hola', randomField: [34, 54, 32] },
+  ]);
+
+  // should accept wrong _id type
+  // $ExpectError
+  await indexTypeCollection1.insertMany([ { _id: 12, stringField: 'hola', numberField: 0 } ]);
```
CaselIT: I think the comment should be `// should not accept wrong _id type`

types/mongodb/test/collection/insertX.ts
```diff
@@ -45,21 +78,154 @@ async function run() {
...
+  interface IndexTypeTestModelWithObjectId {
+    _id: ObjectId;
+    stringField: string;
+    numberField?: number;
+    [key: string]: any;
+  }
+  const indexTypeCollection3 = db.collection<IndexTypeTestModelWithObjectId>('testCollection');
+
+  // TODO: should not demand _id if it is ObjectId
+  // await indexTypeCollection3.insertOne({ stringField: 'hola', numberField: 23, randomField: [34, 54, 32], randomFiel2: 32 });
```
CaselIT: Do you need to add these tests or they will be added at a later time?

CaselIT:
I've added a couple of minor comments. but they are nice to have :)

types/mongodb/test/collection/insertX.ts
HosseinAgha: This is a valid use case but as I said above I couldn't add support for it with the current TS features

PR #40950, opened 2019-12-10
[react-native] adds StyleSheet.compose definition

vtolochk:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://facebook.github.io/react-native/docs/stylesheet#compose>>

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -1965,7 +1965,7 @@ type Falsy = undefined | null | false;
 interface RecursiveArray<T> extends Array<T | RecursiveArray<T>> {}
 /** Keep a brand of 'T' so that calls to `StyleSheet.flatten` can take `RegisteredStyle<T>` and return `T`. */
 type RegisteredStyle<T> = number & { __registeredStyleBrand: T };
-export type StyleProp<T> = T | RegisteredStyle<T> | RecursiveArray<T | RegisteredStyle<T> | Falsy> | Falsy;
+export type StyleProp<T> = T | RegisteredStyle<T> | ReadonlyArray<T> | RecursiveArray<T | RegisteredStyle<T> | Falsy> | Falsy;
 
 /**
  * @see https://facebook.github.io/react-native/docs/accessibility.html#accessibility-properties
@@ -5135,7 +5135,7 @@ export namespace StyleSheet {
      * an array, saving allocations and maintaining reference equality for
      * PureComponent checks.
      */
-    export function compose<T>(style1?: StyleProp<T>, style2?: StyleProp<T>): ReadonlyArray<T> | undefined;
+    export function compose<T>(style1: StyleProp<T>, style2: StyleProp<T>): StyleProp<T>;
 
     /**
      * WARNING: EXPERIMENTAL. Breaking changes will probably happen a lot and will
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -27,6 +27,7 @@
 //                 Be Birchall <https://github.com/bebebebebe>
 //                 Jesse Katsumata <https://github.com/Naturalclar>
 //                 Xianming Zhong <https://github.com/chinesedfan>
+//                 Valentyn Tolochko <https://github.com/vtolochk>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
@@ -5127,6 +5128,14 @@ export namespace StyleSheet {
      * the alternative use.
      */
     export function flatten<T>(style?: StyleProp<T>): T;
+
+    /**
+     * Combines two styles such that style2 will override any styles in style1.
+     * If either style is falsy, the other one is returned without allocating
+     * an array, saving allocations and maintaining reference equality for
+     * PureComponent checks.
+     */
+    export function compose<T>(style1?: StyleProp<T>, style2?: StyleProp<T>): ReadonlyArray<T> | undefined;
 
     /**
      * WARNING: EXPERIMENTAL. Breaking changes will probably happen a lot and will
--- a/types/react-native/test/index.tsx
+++ b/types/react-native/test/index.tsx
@@ -195,6 +195,50 @@ const s = StyleSheet.create({
 });
 const f1: TextStyle = s.shouldWork;
 
+// StyleSheet.compose
+// It creates a new style object by composing two existing styles
+const composeTextStyle1: StyleProp<TextStyle> = {
+    color: '#000000',
+    fontSize: 20,
+};
+
+const composeTextStyle2: StyleProp<TextStyle> = {
+    fontSize: 12,
+};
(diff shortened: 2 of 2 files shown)


elicwhite (changes requested):
Are you sure this type is correct?

[This is the implementation in Flow](https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/StyleSheet.js#L264-L273):
```
compose<T: DangerouslyImpreciseStyleProp>(
    style1: ?T,
    style2: ?T,
  ): ?T | $ReadOnlyArray<T> {
    if (style1 != null && style2 != null) {
      return ([style1, style2]: $ReadOnlyArray<T>);
    } else {
      return style1 != null ? style1 : style2;
    }
  },
```

Which I would think would be something like this:

```
export function compose<T>(style1?: StyleProp<T>, style2?: StyleProp<T>): ?StyleProp<T> | ReadonlyArray<StyleProp<T>>
```
Although it is possible I'm wrong, I use Flow, and don't know TypeScript. :)

The flow implementation in core has a [bunch of flow tests](https://github.com/facebook/react-native/blob/37df2b28ad8a3c698e67250c2464feba78f7832a/Libraries/StyleSheet/__flowtests__/StyleSheet-flowtest.js#L19-L79) that enforces some invariants around this function. Can you copy those over here and make sure they fail/succeed the same way? That will help make sure this matches the same semantics.

vtolochk:
I've updated the PR, added **ReadonlyArray<T>** to the StyleProp<T> type, removed **?** from the parameters because StyleProp<T> type has the **Falsy** type and ? means nothing there.

I copied tests from the flow as u suggested and it works well for me.
I run the `npm lint` and `npm test` commands one more time and everything okay.

types/react-native/test/index.tsx
```diff
@@ -195,6 +195,74 @@ const s = StyleSheet.create({
...
+    composeTextStyle,
+    Math.random() < 0.5 ? composeTextStyle : null,
+);
+
+const combinedStyle6 = StyleSheet.compose(
+    null,
+    null,
+);
+
+// The following use of the compose method is invalid:
```
alloy: You can uncomment these and add a `$ExpectError` comment above them to have the test assertion actually being verified.

alloy (changes requested):
I’m unsure what part you’re referring to by ”`?` being removed from the parameter”, but this PR seems fine to me now, just a minor improvement request to the test.

types/react-native/test/index.tsx
vtolochk: Aha, thanks, will change it.

About `?` - it means an optional parameter or it just simply adds ` | undefined` to the type.

I have used it in the previous commit.

If you will go to the StyleProp<T> type definition, you will see that it includes `Falsy` type.

The `Falsy` type itself has the following definition - `type Falsy = undefined | null | false;` so the `?` doesn't make sense.

alloy: Re `?`, I see, I was just looking at the complete diff. Looking at [the Flow typing](https://github.com/facebook/react-native/blob/master/Libraries/StyleSheet/StyleSheet.js#L265-L266) of the API, the current version of the PR does look correct.

Just one note, in case there was some misunderstanding there, but wether or not `Falsy` includes `undefined` doesn’t matter in regards to `?`, as that is meant to signify an optional parameter, which just happens to become `undefined`.

```ts
// Required parameter that may be undefined
function foo(x: string | undefined) {}

// Optional parameter
function foo(x?: string) {}
```

types/react-native/test/index.tsx
```diff
@@ -195,6 +195,74 @@ const s = StyleSheet.create({
...
+    color: '#000000',
+    fontSize: 20,
+};
+
+const composeImageStyle: StyleProp<ImageStyle> = {
+    resizeMode: 'contain',
+};
+
+// The following use of the compose method is valid
+const combinedStyle = StyleSheet.compose(
```
TheSavior: Can you add the cast to these so that it matches the expected return type? These casts and expected return types are in the original flow test

elicwhite (changes requested):

types/react-native/test/index.tsx
vtolochk: Done!

types/react-native/test/index.tsx
vtolochk: Done!

alloy:
Thanks!

alloy (approved):

lafiosca:
@alloy @vtolochk I'm a bit late, and I'm not sure if this is the place to bring this up, but the addition of `ReadonlyArray<T>` to the `StyleProp` definition breaks my use case of flattening multiple style props to extract final values. Here is a minimal reproduction:

```
const styles: StyleProp<ViewStyle>[] = [];
const combinedStyle = StyleSheet.flatten(styles);
const { top } = combinedStyle;
```

Prior to the change, the type of `combinedStyle` was correctly inferred as `ViewStyle`, and `top` was inferred as `string | number | undefined`. With the change, `combinedStyle` is instead inferred as `StyleProp<ViewStyle>`, and this causes the third line to complain that the property `top` does not exist.

TheSavior:
This should probably be reverted and type tests for flatten should be added.

vtolochk:
I am ready to fix it and add a test for the flatten use case, should I make another PR?

vtolochk:
Flow types suggest returning `ReadonlyArray<T>` for the compose function but compose and flatten functions almost does the same  thing and should have the same return type (am I wrong here ?), what if I will remove `ReadonlyArray<T>` from the return type and it will give the same feature for the compose function to get properties like - `const { top } = combinedStyle;`

It will look something like this:
Screenshot 2019-12-25 at 08 37 31

TheSavior:
Compose is not meant to flatten. Compose is to make sure that if you only pass in one argument that you get that argument back and not an array (for improved equality checks for React).

Flatten merges a style prop and gives you back a plain object guaranteed.

alloy:
@vtolochk That would be great 👌 If you won’t be able to fix this over the next few days, then please first submit a PR that reverts the current change.

types/react-native/index.d.ts
```diff
@@ -1964,7 +1965,7 @@ type Falsy = undefined | null | false;
 interface RecursiveArray<T> extends Array<T | RecursiveArray<T>> {}
```
leonid-shevtsov: Could extending `RecursiveArray` from `ReadonlyArray`, instead of `Array` solve this issue? Don't see how it can be written to.

vtolochk: Unfortunately no :(

vtolochk:
@alloy I've fixed, what is the right way to push the commit for review? Should I create another PR?

TheSavior:
Another PR is good

vtolochk:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41408

PR #40954, opened 2019-12-10
twilio - Fixed incorrect callback signature

CatGuardian:
Please fill in this template.

- [ x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x ] Test the change in your own code. (Compile and run.)
- [x ] Add or edit tests to reflect the change. (Run with `npm test`.)

- [x ] Provide a URL to documentation or source code which provides context for the suggested changes: Line 169 of node_modules/twilio/lib/Client.js when twilio@2.11.1 is installed
- [x ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ x] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

### DIFF
--- a/types/twilio/index.d.ts
+++ b/types/twilio/index.d.ts
@@ -1,6 +1,8 @@
-// Type definitions for twilio
+// Type definitions for twilio 2.11
 // Project: https://github.com/twilio/twilio-node
-// Definitions by: nickiannone <https://github.com/nickiannone>, Ashley Brener <https://github.com/ashleybrener>
+// Definitions by: nickiannone <https://github.com/nickiannone>
+//                 Ashley Brener <https://github.com/ashleybrener>
+//                 Anthony Messerschmidt <https://github.com/CatGuardian>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.3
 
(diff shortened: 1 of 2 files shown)


CatGuardian:
@nickiannone

types/twilio/index.d.ts
```diff
@@ -1,6 +1,8 @@
-// Type definitions for twilio
+// Type definitions for twilio 2.11
```
DanielRosenwasser: Current version is 3.39 on npm. Any clue on significant changes? Should it just be marked as 3.39?

CatGuardian: Twilio 3+ has its own typings defined in the repo/npm package.
https://github.com/twilio/twilio-node/blob/master/package.json#L63

And yes this is majorly different than Twilio v3. They pretty much did a complete rewrite to the library.

So my intentions is to fix up the typings for the version of Twilio I'm using since either upgrading TypeScript to the latest version or upgrading my node version to node 12 (and thusly the @types/node) caused an error with these Typings. Not sure since I upgraded TypeScript and Node at the same time.

DanielRosenwasser: Gotcha, thanks for the response!

CatGuardian:
Thank you @DanielRosenwasser for looking at this and merging!

PR #40958, opened 2019-12-11
@types/braintree webhooks

ACDR:
Braintree webhook testing types

Please fill in this template.

- [Y] Use a meaningful title for the pull request. Include the name of the package modified.
- [Y] Test the change in your own code. (Compile and run.)
- [Y] Add or edit tests to reflect the change. (Run with `npm test`.)
- [Y] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [Y] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [Y] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [Y] Provide a URL to documentation or source code which provides context for the suggested changes: https://developers.braintreepayments.com/guides/webhooks/testing-go-live/node

- [N/A] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.

- [N/A] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

### DIFF
--- a/types/braintree/index.d.ts
+++ b/types/braintree/index.d.ts
@@ -838,6 +838,7 @@ declare namespace braintree {
 
     export class WebhookNotification {
         kind: WebhookNotificationKind;
+        timestamp: Date;
         subscription?: Subscription;
         merchantAccount?: MerchantAccount;
         transaction?: Transaction;
--- a/types/braintree/index.d.ts
+++ b/types/braintree/index.d.ts
@@ -1,7 +1,8 @@
 // Type definitions for braintree 2.20
 // Project: https://github.com/braintree/braintree_node
 // Definitions by: Sam Rubin <https://github.com/smrubin>,
-//                 Mohamed Elsharnouby <https://github.com/sharno>
+//                 Mohamed Elsharnouby <https://github.com/sharno>,
+//                 Aaron Rose <https://github.com/acdr>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.6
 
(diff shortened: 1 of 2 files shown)


types/braintree/index.d.ts
```diff
@@ -816,6 +827,55 @@ declare namespace braintree {
...
+    /**
+     * Webhooks
+     */
+
+    export class SampleNotification {
+        bt_signature: string;
+        bt_payload: string;
+    }
+
+    export class WebhookNotification {
```
smrubin: Looks like `timestamp` is part of the WebhookNotification as well, can we add it here? https://developers.braintreepayments.com/reference/general/webhooks/overview

types/braintree/index.d.ts
```diff
@@ -816,6 +827,55 @@ declare namespace braintree {
...
+        kind: WebhookNotificationKind;
+        subscription?: Subscription;
+        merchantAccount?: MerchantAccount;
+        transaction?: Transaction;
+        dispute?: Dispute;
```
smrubin: Looks like there are some additional kinds of webhook notifications as well that yield additional attributes (https://developers.braintreepayments.com/reference/general/webhooks/overview#notification-kinds). Think this PR is a great start to implementing webhooks though, so it shouldn't be a blocker. Feel free to submit another PR for those if you're up for it!

smrubin (changes requested):
@ACDR Looks good and thanks for working on the webhooks (and not to mention adding types for the testing). I'm only requesting changes to add the `timestamp` field to the `WebhookNotification` class. After that change, should be good to ship.

smrubin (approved):
Thanks @ACDR!

orta:
👍

PR #40961, opened 2019-12-11
[@types/react]Dispatch in useReducer could accept 0 arguments

dancerphil:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://gist.github.com/dancerphil/86dee843b1b9bbebfb4a5bb3f138dd5f

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -900,7 +900,8 @@ declare namespace React {
     // overload where dispatch could accept 0 arguments.
     function useReducer<R extends ReducerWithoutAction<any>>(
         reducer: R,
-        initializerArg: ReducerStateWithoutAction<R>
+        initializerArg: ReducerStateWithoutAction<R>,
+        initializer?: undefined
     ): [ReducerStateWithoutAction<R>, DispatchWithoutAction];
     /**
      * An alternative to `useState`.
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -21,6 +21,7 @@
 //                 Kanitkorn Sujautra <https://github.com/lukyth>
 //                 Sebastian Silbermann <https://github.com/eps1lon>
 //                 Kyle Scully <https://github.com/zieka>
+//                 Cong Zhang <https://github.com/dancerphil>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
@@ -821,7 +822,8 @@ declare namespace React {
     type SetStateAction<S> = S | ((prevState: S) => S);
     // this technically does accept a second argument, but it's already under a deprecation warning
     // and it's not even released so probably better to not define it.
-    type Dispatch<A> = (value: A) => void;
+    // Since action _can_ be undefined, dispatch may be called without any parameters.
+    type Dispatch<A> = (value?: A) => void;
     // Unlike redux, the actions _can_ be anything
     type Reducer<S, A> = (prevState: S, action: A) => S;
     // types used to try and prevent the compiler from reducing S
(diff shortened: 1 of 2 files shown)


types/react/test/hooks.tsx
```diff
@@ -158,6 +160,7 @@ function useEveryHook(ref: React.Ref<{ id: number }>|undefined): () => boolean {
     }, []);
     React.useEffect(() => {
         dispatch({ type: 'getOlder' });
```
eps1lon: I think this dispatch would now also allow zero arguments. Since this is the most common use case it should be fully supported instead of a nichè case. 

I think we need an overload to make zero-arguments dispatch safe.

eps1lon (changes requested):

types/react/test/hooks.tsx
dancerphil: Any idea how to implement the overload?

eps1lon: I haven't put any thought into it. As I said it's a very rare use case. Maybe dispatch `undefined` explicitly?

ferdaber: The dispatch type is reliant on the reducer, so if the reducer is 1-ary, then the dispatch necessarily needs to be 0-ary, @dancerphil you can start with that.

dancerphil: Thank a lot, i am working on it: https://gist.github.com/dancerphil/060d5aeca3c47be2e5d9ba368796b43f

I fully agree that we should not allow both defined and undefined values.

dancerphil:
@eps1lon Hello, would you like review this PR again?

eps1lon (approved):
This looks good, thanks!

types/react/index.d.ts
```diff
@@ -864,6 +871,38 @@ declare namespace React {
...
+     * updates because you can pass `dispatch` down instead of callbacks.
+     *
+     * @version 16.8.0
+     * @see https://reactjs.org/docs/hooks-reference.html#usereducer
+     */
+    // overload where dispatch could accept 0 arguments.
+    function useReducer<R extends ReducerWithoutAction<any>>(
+        reducer: R,
+        initializerArg: ReducerStateWithoutAction<R>,
+        initializer?: undefined
```
ferdaber: Should this just be omitted?

dancerphil: You mean this line?

I just follow the similar line [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L924).

Looks it can be omitted. I'm not quite sure.

dancerphil: Tell me if you are sure about this.

dancerphil: I have test it. It passed.

ferdaber: ughhh I remember talking to Jessica about this, and I think it was necessary at the time...? It may be fixed now but honestly I'd rather just keep it consistent, so you can keep it the way it was before.

ferdaber (approved):
Thanks for contributing!

DanielRosenwasser:
Thank you!

dancerphil:
Thanks you! @eps1lon @ferdaber @DanielRosenwasser 

Also thanks for the help of my friends, @LeuisKen, @yuxuan, @NullMDR

billyjanitsch:
@dancerphil @ferdaber it looks like this PR broke the following use case:

```js
const reducer = (state: Set<number>, action: number): Set<number> => state
React.useReducer(reducer, undefined, () => new Set())
```

This code now yields this error:

```
  The last overload gave the following error.
    Argument of type 'undefined' is not assignable to parameter of type 'Set<number>'.ts(2769)
```

Reverting to `@types/react@16.9.16` fixes the issue.

dancerphil:
@billyjanitsch OK, I will check this.

If I can solve this, I will raise another PR and mention you.

PR #40966, opened 2019-12-11
[@types/ioredis] hgetall returns object

massivefermion:
changed type of hgetall result from string[] to object

### DIFF
--- a/types/ioredis/index.d.ts
+++ b/types/ioredis/index.d.ts
@@ -323,8 +323,8 @@ declare namespace IORedis {
         hvals(key: KeyType, callback: (err: Error, res: string[]) => void): void;
         hvals(key: KeyType): Promise<string[]>;
 
-        hgetall(key: KeyType, callback: (err: Error, res: object) => void): void;
-        hgetall(key: KeyType): Promise<object>;
+        hgetall(key: KeyType, callback: (err: Error, res: { [key: string]: string | number }) => void): void;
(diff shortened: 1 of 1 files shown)


types/ioredis/index.d.ts
```diff
@@ -323,8 +323,8 @@ declare namespace IORedis {
         hvals(key: KeyType, callback: (err: Error, res: string[]) => void): void;
         hvals(key: KeyType): Promise<string[]>;
```
SimonSchick: That should be `{ [key: string]: string | number }` iirc

MassiveFermion: fixed it

SimonSchick (approved):

orta:
Cool, looks good to me

demian85:
Can someone please explain this commit: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40966/commits/f3e255236d34ab03de32dd3581c0737575955311

Why does `hgetall` return `string | number`? Redis stores all its data as strings, it just makes no sense for a key to have a number type. It is breaking my types. Did anyone validate this before approving such thing?

types/ioredis/index.d.ts
demian85: This is completely wrong. Should I create another pull request fixing an already merged pull request?

SimonSchick: No you would open a new one.
My bad for suggesting the wrong type here, got it mixed up with something else.

PR #40968, opened 2019-12-11
[@types/react-native] WebSocket

SychevSP:
Added type declaration for WebSocket

### DIFF
--- a/types/react-native/globals.d.ts
+++ b/types/react-native/globals.d.ts
@@ -296,5 +296,12 @@ interface WebSocket extends EventTarget {
 
 declare var WebSocket: {
     prototype: WebSocket;
-    new (uri: string, protocols?: string | string[], options?: {headers: {[headerName: string]: string}}): WebSocket;
+    new (
+        uri: string,
+        protocols?: string | string[] | null,
+        options?: {
+            headers: {[headerName: string]: string};
+            [optionName: string]: any;
+        } | null,
+    ): WebSocket;
 };
--- a/types/react-native/globals.d.ts
+++ b/types/react-native/globals.d.ts
@@ -274,7 +274,7 @@ declare class URLSearchParams {
 }
 
 interface WebSocketMessageEvent extends Event {
-    data?: unknown;
+    data?: any;
 }
 interface WebSocketErrorEvent extends Event {
     message: string;
@@ -286,7 +286,7 @@ interface WebSocketCloseEvent extends Event {
 
 interface WebSocket extends EventTarget {
     readonly readyState: number;
-    send(data?: any): void;
+    send(data: string | ArrayBuffer | ArrayBufferView | Blob): void;
     close(code?: number, reason?: string): void;
     onopen: (() => void) | null;
     onmessage: ((event: WebSocketMessageEvent) => void) | null;
--- a/types/react-native/globals.d.ts
+++ b/types/react-native/globals.d.ts
@@ -272,3 +272,29 @@ declare class URLSearchParams {
     keys(): IterableIterator<string>;
     values(): IterableIterator<string>;
 }
+
+interface WebSocketMessageEvent extends Event {
+    data?: unknown;
+}
+interface WebSocketErrorEvent extends Event {
+    message: string;
+}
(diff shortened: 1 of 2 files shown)


types/react-native/globals.d.ts
```diff
@@ -272,3 +272,29 @@ declare class URLSearchParams {
     keys(): IterableIterator<string>;
     values(): IterableIterator<string>;
 }
+
+interface WebSocketMessageEvent extends Event {
+    data?: any;
```
alloy: Should this perhaps be `unknown` nowadays?

alloy (changes requested):
Can you please provide links to docs/source to be able to review these additions?

SychevSP:
> Can you please provide links to docs/source to be able to review these additions?

To make these changes I used three sources:
1) Usage of WebSockets in my own RN project
2) React Native documentation
3) Flow types in React Native on GitHub

The overview of the API is in React Native documentation: 
https://facebook.github.io/react-native/docs/network#websocket-support

A more detailed description is on GitHub:
readyState is a number: 
https://github.com/facebook/react-native/blob/a62062f3c0171fb16b8ce761ce0934a4de4d296c/Libraries/WebSocket/WebSocket.js#L82

types for the send method:
https://github.com/facebook/react-native/blob/a62062f3c0171fb16b8ce761ce0934a4de4d296c/Libraries/WebSocket/WebSocket.js#L166

types for the close method:
https://github.com/facebook/react-native/blob/a62062f3c0171fb16b8ce761ce0934a4de4d296c/Libraries/WebSocket/WebSocket.js#L157

All the onX events:
https://github.com/facebook/react-native/blob/a62062f3c0171fb16b8ce761ce0934a4de4d296c/Libraries/WebSocket/WebSocket.js#L217

Constructor:
https://github.com/facebook/react-native/blob/a62062f3c0171fb16b8ce761ce0934a4de4d296c/Libraries/WebSocket/WebSocket.js#L85

types/react-native/globals.d.ts
SychevSP: I will fix it in a new commit

SychevSP: Changing from 'any' to 'unknown' requires TypeScript 3.0. The tslint configuration of the package requires TypeScript 2.9
To use unknown type I would have to specify TypeScript 3.0 for the whole file globals.d.ts.
I thought it may be a breaking change so did not do it

alloy: Makes sense 👍

types/react-native/globals.d.ts
```diff
@@ -272,3 +272,29 @@ declare class URLSearchParams {
...
+    close(code?: number, reason?: string): void;
+    onopen: (() => void) | null;
+    onmessage: ((event: WebSocketMessageEvent) => void) | null;
+    onerror: ((event: WebSocketErrorEvent) => void) | null;
+    onclose: ((event: WebSocketCloseEvent) => void) | null;
+}
+
+declare var WebSocket: {
+    prototype: WebSocket;
+    new (uri: string, protocols?: string | string[], options?: {headers: {[headerName: string]: string}}): WebSocket;
```
alloy: In Flow, a question mark to the left of a type identifier means it can be `undefined` (optional) or `null`. The `options` object has optional keys and should allow for more keys than the few it hints at.

```suggestion
    new (uri: string, protocols?: string | string[] | null, options?: { headers?: { options?: string, [key: string]: any }, [key: string]: any } | null): WebSocket;
```

alloy (changes requested):
Why is `WebSocket` not defined as a `class`?

types/react-native/globals.d.ts
SychevSP: Declaring WebSocket as a class works too.
I used type merging (interface + var) instead of class because this pattern is used all over this type declaration file. For example XMLHttpRequest and Headers are defined in this way.

Added optional keys to options. Allowed protocols and options to be null.

alloy: Fair enough; still, thanks for updating!

alloy:
Great, thanks for bearing with me! 🙏

PR #40973, opened 2019-12-11
[expect-puppeteer]: Added "default options" functions

minusfive:
Please fill in this template.

  - https://github.com/smooth-code/jest-puppeteer/tree/master/packages/expect-puppeteer#configure-default-options
  - https://github.com/smooth-code/jest-puppeteer/blob/master/packages/expect-puppeteer/src/options.js
  - https://github.com/smooth-code/jest-puppeteer/commits/master/packages/expect-puppeteer/src/options.js

---

This PR adds missing definitions for [`expect-puppeteer`](https://github.com/smooth-code/jest-puppeteer/tree/master/packages/expect-puppeteer#configure-default-options)'s default options configuration functions.

### DIFF
--- a/types/expect-puppeteer/index.d.ts
+++ b/types/expect-puppeteer/index.d.ts
@@ -67,4 +67,11 @@ declare global {
 }
 
 declare function expectPuppeteer(instance: ElementHandle | Page): ExpectPuppeteer;
-export = expectPuppeteer;
+
+interface ExpectDefaultOptions extends ExpectToClickOptions, ExpectTimingActions {}
+declare function defaultOptions(options: ExpectDefaultOptions): ExpectDefaultOptions;
+declare function setDefaultOptions(options: ExpectDefaultOptions): void;
+declare function getDefaultOptions(): ExpectDefaultOptions;
+
+export default expectPuppeteer;
+export { defaultOptions, setDefaultOptions, getDefaultOptions };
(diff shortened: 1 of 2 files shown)


minusfive:
Ping @gregberge in case you want to review?

tkrotoff (changes requested):
.

tkrotoff (approved):
Nice!

orta:
I'm a little worried about the change for: `export = expectPuppeteer` to `export default expectPuppeteer;` - which will break builds for imports. 

Are you sure that's what you want to do?

minusfive:
@orta right. I'm used to most projects nowadays automatically handling conversion from CJS to ESM. Since TS only allows 1 "=" export, I'm not clear on how to do it, but will dig into it. Perhaps as keys of the parent interface? Will update PR as soon as I'm back at comp.

types/expect-puppeteer/index.d.ts
```diff
@@ -67,4 +67,11 @@ declare global {
...
 
 declare function expectPuppeteer(instance: ElementHandle | Page): ExpectPuppeteer;
-export = expectPuppeteer;
+
+interface ExpectDefaultOptions extends ExpectToClickOptions, ExpectTimingActions {}
+declare function defaultOptions(options: ExpectDefaultOptions): ExpectDefaultOptions;
+declare function setDefaultOptions(options: ExpectDefaultOptions): void;
+declare function getDefaultOptions(): ExpectDefaultOptions;
+
+export default expectPuppeteer;
```
DanielRosenwasser: This package was declared using `module.exports = ...` or some variant. It should be declared with an `export = ...` like one of the following:

### Variables

```ts
declare var xyz: SomeType;

export = xyz;
```

### Classes

```ts
class Foo {
  // ...
}

export = Foo
```

### Functions

```ts
function foo(): any;

export = foo;
```

```ts
import foo = require("foo");
```

DanielRosenwasser (changes requested):

PR #40978, opened 2019-12-11
Update status badges

sandersn:
Point to dev.azure.com/DefinitelyTyped instead of /Microsoft. These repos are now part of the DefinitelyTyped org.


sandersn:
Manual inspection of the badges in the diff looks good.

PR #40983, opened 2019-12-12
Add RouterService `off` method definition to compliment `on` definition

rjschie:
Please fill in this template.

  - **RouterService** extends the `Evented` mixin, which implements the `on` and `off` methods for the two `routeWillChange` and `routeDidChange` events. While `on` has been defined on the RouterService class, `off` is missing.
  - https://github.com/emberjs/ember.js/blob/v3.14.1/packages/%40ember/-internals/routing/lib/services/router.ts#L410
  - http://api.emberjs.com/ember/3.12/classes/Evented/methods/off?anchor=off

### DIFF
--- a/types/ember__routing/router-service.d.ts
+++ b/types/ember__routing/router-service.d.ts
@@ -220,7 +220,8 @@ export default class RouterService extends Service {
         options?: { queryParams: object }
     ): string;
 
-    // https://api.emberjs.com/ember/3.6/classes/RouterService/events/routeDidChange?anchor=routeDidChange
+    // https://api.emberjs.com/ember/3.12/classes/RouterService/events/routeDidChange?anchor=routeDidChange
+    // http://api.emberjs.com/ember/3.12/classes/Evented/methods/on?anchor=on
     /**
      * Register a callback for an event.
      *
@@ -237,5 +238,21 @@ export default class RouterService extends Service {
     on(
         name: 'routeDidChange' | 'routeWillChange',
         callback: (transition: Transition) => void
+    ): RouterService;
+
+    // https://api.emberjs.com/ember/3.12/classes/RouterService/events/routeDidChange?anchor=routeDidChange
+    // http://api.emberjs.com/ember/3.12/classes/Evented/methods/off?anchor=off
+    /**
+     * Remove a registered callback for an event.
+     *
+     * To remove the registered callback, the function you pass must be the
+     * same function as the original callback that was registered.
+     *
+     * @param name     the name of the event
+     * @param callback the same function of the original callback
+     */
+    off(
+      name: 'routeDidChange' | 'routeWillChange',
+      callback: (transition: Transition) => void
     ): RouterService;
 }
--- a/types/ember__routing/test/router.ts
+++ b/types/ember__routing/test/router.ts
@@ -80,4 +80,32 @@ const RouterServiceConsumer = Service.extend({
                 }
             });
     },
+    offAndRouteInfo() {
+        const router = get(this, 'router');
+        router
+            .off('routeWillChange', transition => {
+                const to = transition.to;
+                to.child; // $ExpectType RouteInfo | null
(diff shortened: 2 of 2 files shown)


chriskrycho (changes requested):
One question before I approve this – I think we could simplify the tests a bit – but *thank you* for doing this!

types/ember__routing/router-service.d.ts
```diff
@@ -238,4 +239,20 @@ export default class RouterService extends Service {
...
+     *
+     * To remove the registered callback, the function you pass must be the
+     * same function as the original callback that was registered.
+     *
+     * @param name     the name of the event
+     * @param callback the same function of the original callback
+     */
+    off(
+      name: 'routeDidChange' | 'routeWillChange',
+      callback: (transition: Transition) => void
```
jamescdavis: Should we also support the three-argument version where the `this` context is specified as the second argument and the third parameter is the name of a method on the target? Would mean overloading with something like:
```ts
off<T extends object = object(
    name: 'routeDidChange' | 'routeWillChange',
    target: T,
    method: CallbackMethod<T>,
): RouterService;
```
Where `CalllbackMethod` is something like:
```ts
type CallbackMethod<T> = {
     [K in keyof T]: T[K] extends (transition: Transition) => void ? K : never
}[keyof T];
```
(and update `on` to support this as well.)

jamescdavis: Please use 4-space indents for consistency.

chriskrycho:
No worries! You should use `$ExpectType Transition` over `assertType<Transition>` because `$ExpectType` will error if it's too *wide* as well as if it's too narrow a type. `assertType<Transition>` will happily pass if what gets handed back happens to be `any`; `$ExpectType` will *not*.

types/ember__routing/router-service.d.ts
rjschie: Good point, thanks @jamescdavis ! I'll make that change.

mike-north (changes requested):
Aside from the particulars of the implementation, I think we may be missing something here

The ember API docs do not seem to indicate that the fact that the `RouterService` happens to use the `Evented` mixin is part of the public API.

Here's an example of `Route`, which clearly says "Uses: Evented"
Screen Shot 2019-12-15 at 10 56 10 PM

And the `RouterService` has no such indication
Screen Shot 2019-12-15 at 10 56 32 PM

Now, clearly the [`routeDidChange`](https://api.emberjs.com/ember/3.15/classes/RouterService/events?show=inherited%2Cprotected%2Cprivate%2Cdeprecated) documentation illustrates use of an `.on(` method, so it's pretty likely that this is a bug in the docs.

If `RouterService` is _deliberately not indicating_ that it *uses* this mixin in the API docs, we should not make this change at all (I doubt that this is the case).

If we do find that this is a bug in the API docs, I believe the proper fix here is just
```diff
+ import Evented from '@ember/object/evented';

- export default class RouterService extends Service {
+ export default class RouterService extends Service implements Evented {
```

To accurately represent the presence of this mixin, we can't really constrain `on` and `off` to specific event types. [`Evented#trigger`](https://api.emberjs.com/ember/3.15/classes/Evented/methods/trigger?anchor=trigger) is a public method, which allows for _arbitrarily named events_ to be fired (and listened for).

The following would be a valid (but admittedly unusual) use case, and we should add tests for it if we find that `RouterService` is supposed to include `Evented`

```ts
 router.on("foo", () => {});
 router.trigger("foo");
 router.off("foo", () => {});
```

Through following [@jamescdavis' suggestions](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40983/files#r357742725) it would be possible to define some override signatures that are nicely typed, while still allowing the example above with the `"foo"` event to function in a less-nicely-typed way (i.e., the callback arg might be an `{}` or something -- `unknown` when we can drop TS 2.8 and 2.9)

PR #40989, opened 2019-12-12
fix: align callback functions of dataset with spec

tpluscode:
Please fill in this template.

- [x] Provide a URL to documentation: https://rdf.js.org/dataset-spec/#quadmapiteratee-interface

### DIFF
--- a/types/rdf-js/rdf-js-tests.ts
+++ b/types/rdf-js/rdf-js-tests.ts
@@ -177,27 +177,11 @@ function test_dataset() {
     const stream1: Stream = <any> {};
     const stream2: Stream<QuadBnode> = <any> {};
 
-    const quadFilterIteratee = (quad: Quad, dataset: Dataset): boolean => {
-        return dataset.has(quad);
-    };
-    const quadMapIteratee = (quad: Quad, dataset: Dataset): Quad => {
-        if (dataset.has(quad))
-            return quad;
-
-        return factory.quad(quad.subject, quad.predicate, quad.object, quad.graph);
-    };
-    const quadReduceToStringIteratee = (reduced: string, quad: Quad): string => {
-        return reduced + quad.subject.termType;
-    };
-    const quadReduceToArrayIteratee = (arr: boolean[], quad: Quad, dataset: Dataset): boolean[] => {
-        return [
-            ...arr,
-            dataset.has(quad)
-        ];
-    };
-    const quadForEachIteratee = (quad: Quad, dataset: Dataset): void => {
-        console.log(dataset.has(quad) ? 'it really has it' : 'what?');
-    };
+    const quadFilterIteratee: (quad: Quad, dataset: Dataset) => boolean = <any> {};
+    const quadMapIteratee: (quad: Quad, dataset: Dataset) => Quad = <any> {};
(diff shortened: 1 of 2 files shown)


types/rdf-js/rdf-js-tests.ts
```diff
@@ -176,19 +177,27 @@ function test_dataset() {
...
-    const quadMapIteratee2: QuadMapIteratee<QuadBnode> = <any> {};
-
-    const quadReduceIteratee1: QuadReduceIteratee = <any> {};
-    const quadReduceIteratee2: QuadReduceIteratee<string> = <any> {};
-    const quadReduceIteratee3: QuadReduceIteratee<any, QuadBnode> = <any> {};
-    const quadReduceIteratee4: QuadReduceIteratee<string, QuadBnode> = <any> {};
-
-    const quadRunIteratee1: QuadRunIteratee = <any> {};
-    const quadRunIteratee2: QuadRunIteratee<QuadBnode> = <any> {};
+    const quadFilterIteratee = (quad: Quad, dataset: Dataset) => {
```
rubensworks: Could you add the explicit types for each iteratee?
It's good to have the tests as strongly typed as possible, to avoid unintentonal future breakages.

rubensworks (changes requested):

types/rdf-js/rdf-js-tests.ts
tpluscode: I added the return types to callback signatures. You think that enough?

rubensworks: While that's also good to have, I was referring to the types of the iteratees itself. So `const quadFilterIteratee: someType = ...`

tpluscode: I'm not convinced.

The explicit type makes sense when verifying the return of tested call. In this case the type inference should be enough:

```ts
const quadFilterIteratee: (quad: Quad, dataset: Dataset) => boolean 
  = (quad: Quad, dataset: Dataset): boolean {}
```

I think this is unnecessarily redundant.

rubensworks: Yes, it's redundant, but that's the whole point of these tests 🙂.
Things here should be as strongly typed as possible to avoid future regressions.

tpluscode: Not really. This is not the subject under test. The actual interface methods are and there v everything is explicitly typed.

Here nothing is really implicit.

I'm out now so if you insist I can do b that v later. But seriously, would you also have me type

```js
const x: number = 5;
const y: string = "there is nothing implicit about this" 
```

?

tpluscode: Alright, having re-read my own argumentation I realised that the implementation of those callback methods is just a distraction and clutter anyway. I replaced them with just variables typed to what I expect them to be.

rubensworks (approved):
LGTM!

tpluscode:
Hi @armanio123. Sorry to bother you directly but you have been reviewing my other PRs before. Would you mind looking at this one? 🙏  Seems that the red label got it stuck?

Also, I have two other, significantly larger PRs. Is there any official way to draw maintainers' attention to them? 😞

PR #41002, opened 2019-12-12
Lodash: add _.camelCase string array support 

mouadennaciri:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://lodash.com/docs/4.17.15#camelCase>

### DIFF
--- a/types/lodash/lodash-tests.ts
+++ b/types/lodash/lodash-tests.ts
@@ -5854,8 +5854,8 @@ fp.now(); // $ExpectType number
         _.transform<number, AbcObject>(dictionary, iterator); // $ExpectType Dictionary<AbcObject>
         _.transform<number, AbcObject>(dictionary, iterator, accumulator); // $ExpectType Dictionary<AbcObject>
         _(dictionary).transform(); // $ExpectType LoDashImplicitWrapper<Dictionary<any>>
-        _(dictionary).transform<number, AbcObject>(iterator); // $ExpectType LoDashImplicitWrapper<Dictionary<AbcObject>>
-        _(dictionary).transform<number, AbcObject>(iterator, accumulator); // $ExpectType LoDashImplicitWrapper<Dictionary<AbcObject>>
+        _(dictionary).transform(iterator); // $ExpectType LoDashImplicitWrapper<Dictionary<AbcObject>>
+        _(dictionary).transform(iterator, accumulator); // $ExpectType LoDashImplicitWrapper<Dictionary<AbcObject>>
     }
 
     {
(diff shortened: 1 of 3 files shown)


types/lodash/lodash-tests.ts
```diff
@@ -5854,17 +5854,17 @@ fp.now(); // $ExpectType number
         _.transform<number, AbcObject>(dictionary, iterator); // $ExpectType Dictionary<AbcObject>
         _.transform<number, AbcObject>(dictionary, iterator, accumulator); // $ExpectType Dictionary<AbcObject>
         _(dictionary).transform(); // $ExpectType LoDashImplicitWrapper<Dictionary<any>>
-        _(dictionary).transform<number, AbcObject>(iterator); // $ExpectType LoDashImplicitWrapper<Dictionary<AbcObject>>
-        _(dictionary).transform<number, AbcObject>(iterator, accumulator); // $ExpectType LoDashImplicitWrapper<Dictionary<AbcObject>>
+        _(dictionary).transform(iterator); // $ExpectType LoDashImplicitWrapper<Dictionary<AbcObject>>
+        _(dictionary).transform(iterator, accumulator); // $ExpectType LoDashImplicitWrapper<Dictionary<AbcObject>>
```
mouadennaciri: Fix a strange error when I run `npm run lint lodash`
```
The 'use-default-type-parameter' rule threw an error in '.../DefinitelyTyped/types/lodash/lodash-tests.ts':
TypeError: Cannot read property 'default' of undefined
```

thorn0:
As per the docs the parameter is `string`. While it indeed works for `string[]` too because Lodash calls `.toString()` on the argument, generally we don't add undocumented signatures to the type definitions.

thorn0 (changes requested):
As per the docs the parameter is string. While it indeed works for string[] too because Lodash calls .toString() on the argument, generally we don't add undocumented signatures to the type definitions.

So I don't think this PR can be merged.

mouadennaciri:
> As per the docs the parameter is string. While it indeed works for string[] too because Lodash calls .toString() on the argument, generally we don't add undocumented signatures to the type definitions.
> 
> So I don't think this PR can be merged.

@thorn0 According to the docs _.camelCase can take a string array.

> Arguments
> [string=''] (string): The string to convert.

[_.parseInt](https://lodash.com/docs/4.17.15#parseInt) is for example a utility which takes only string.

> Arguments
> string (string): The string to convert.

thorn0:
It's not a string array, this notation means that it's an optional parameter.

mouadennaciri:
> It's not a string array, this notation means that it's an optional parameter.

@thorn0 You're right, ok I close the PR.

PR #41015, opened 2019-12-13
fix(@types/weixin-app): add missing createCanvasContext function

ChrisCindy:


taoqf:
https://developers.weixin.qq.com/miniprogram/dev/api/canvas/wx.createCanvasContext.html

types/weixin-app/index.d.ts
```diff
@@ -3023,7 +3023,8 @@ declare namespace wx {
 	 * ontext跟<canvas/>不存在对应关系，
 	 * 一个context生成画布的绘制动作数组可以应用于多个<canvas/>。
 	 */
-	function createContext(): CanvasContext;
+    function createContext(): CanvasContext;
+    function createCanvasContext(canvasId: string, instance?: object): CanvasContext;
```
taoqf: Could you add some tests about this? I am not sure about type `object`.

taoqf (changes requested):
Make sure type `object` works. maybe you would add some tests.

PR #41021, opened 2019-12-13
[@types/react] @types/react state Readonly - this is incorrect (#41020)

vyemialyanchyk:
@Microsoft

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -469,7 +469,7 @@ declare namespace React {
         // In the future, if we can define its call signature conditionally
         // on the existence of `children` in `P`, then we should remove this.
         readonly props: Readonly<P> & Readonly<{ children?: ReactNode }>;
-        state: Readonly<S>;
+        state: S;
         /**
          * @deprecated
          * https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs


ferdaber (changes requested):
What is the reasoning behind this change? `state` is deliberately set to readonly so that you do not accidentally mutate state as per the React team's recommendation. State changes should go through `this.setState` instead.

vyemialyanchyk:
> What is the reasoning behind this change? `state` is deliberately set to readonly so that you do not accidentally mutate state as per the React team's recommendation. State changes should go through `this.setState` instead.

For me it is not clear how this is related to this particular fix. React team recomendation is for developers, but there are plenty of ways to manage state. if I define class for my component state, I need to write setter to update property directly. 
'props' is definetly Readonly, 'state' is a state of component, it should be changable, how to manage component state should be per developer decision, no sense to write setters or search other workarounds (like (this.state as T).id=x for example), just because someone decided to make immutable initially mutable property. By it's intention state is mutable. How to mute it - this is just recomendation.

eps1lon:
> How to mute it - this is just recommendation.

By that logic anything written in the reactjs docs is "just recommendation". At that point we don't need any type definitions at all because e.g. using `render` to return the rendered content is "just recommendation".

The docs don't just recommend it but explicitly state you should not mutate it:

> Do Not Modify State Directly

-- https://reactjs.org/docs/state-and-lifecycle.html#do-not-modify-state-directly

If you want to use some other mechanism that doesn't cause a re-render when changed, you could use class fields.

vyemialyanchyk:
> > How to mute it - this is just recommendation.
> 
> By that logic anything written in the reactjs docs is "just recommendation". At that point we don't need any type definitions at all because e.g. using `render` to return the rendered content is "just recommendation".
> 
> The docs don't just recommend it but explicitly state you should not mutate it:
> 
> > Do Not Modify State Directly
> 
> -- https://reactjs.org/docs/state-and-lifecycle.html#do-not-modify-state-directly
> 
> If you want to use some other mechanism that doesn't cause a re-render when changed, you could use class fields.

by the reference you provided, the main statement is:
**"For example, this will not re-render a component:"**
i.e. proper interpretation of is:
if you do 
this.state.comment = 'Hello';
- this will not re-render a component.

BUT if we use MobX and marked comment as observable
this.state.comment = 'Hello';
is easy fair way to re-render component.

So decision to make state as Readonly, prevent many developers to use best technics.
React does not impose use of setState, they just state:
this.state.comment = 'Hello';
will not re-render a component with simple JS implementation.

eps1lon:
> React does not impose use of setState, they just state:

This is incomplete. While they do state that it won't re-render they also say

> Do Not Modify State Directly

and 

> // Wrong
> this.state.comment = 'Hello';

So if you do change how components work with decorators then you have to change the typings as well. In contrast by changing the code to conform to mobx we make the experience for everyone else harder. We could make a pragmatic argument to "just accept" it but I don't see mobx as the de facto standard for state management in react.

vyemialyanchyk:
@eps1lon Sebastian, with my reply this will turn into linguistic argue.. imo, you are not right and not ready to hear my arguments, schade. ) 

"well-intentioned lined road to hell".  "every one else" - to make they experience smooth.. they are not children.. they also able to read React documentation and interpret it according their common sense.. 

There is a sense to ask author of #do-not-modify-state-directly directly what did he mean. Have he other way to underline, what using direct assignment will not trigger re-render for component in plain JS.

PR #41024, opened 2019-12-13
Multimap modified values and keys types

tallenstudios:
- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/villadora/multi-map>>

This change is to make it so that **keys** and **values** don't have to be typed as _any_ when using multimap.


orta:
Looks good, would love to have some tests on this in the future 👍

PR #41031, opened 2019-12-14
[react-redux] Configurable RootState with module augmentation

Lazyuki:
With this PR, users can use Typescript's [module augmentation](https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation) to specify the type of `RootState` in which all the `react-redux` functions can infer from. (I got this idea from how `styled-components` does `DefaultTheme` ) https://github.com/DefinitelyTyped/DefinitelyTyped/blob/7cb38cbc6d4f3092270331f90c16d29d1fc43cc2/types/styled-components/index.d.ts#L410-L417
All users need to do is create a `d.ts` file that extends this `RootState` with their own state. This works particularly well with redux since you are supposed to create only one "root state" and all the components share this state
e.g.
```tsx
// my-redux.d.ts
import 'react-redux';
type UserState = { name: string };
type CacheState = { favorites: string[] };

type MyState = { user: UserState, cache: CacheState };

declare module 'react-redux' {
  export interface ConfigurableRootState extends MyState {}
}
```

Then, anywhere in their code, they could simply do 
```tsx
connect(
    ({ user, cache }) => (
        { 
            username: user.name,
            favorites: cache.favorites,
        }
    ))(Component);
```
without manually specifying types.
The root state type would be automatically inferred in other functions like
```tsx
const username = useSelector(state => state.user.name);
```
as well. 

I've tested this in my own project and it works great, but I may have overlooked unintended side effects. Let me know if this is something that you guys tried but decided against it. 

Please fill in this template.

- [x] Test the change in your own code. (Compile and run.) **Can't test module augmentation in those test files**

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <see styled-components link>

### DIFF
--- a/types/react-redux/index.d.ts
+++ b/types/react-redux/index.d.ts
@@ -46,7 +46,7 @@ import hoistNonReactStatics = require('hoist-non-react-statics');
 export interface ConfigurableRootState {}
 
 export type AnyIfEmpty<T extends object> = keyof T extends never ? any : T;
-export type SafeRoofState = AnyIfEmpty<ConfigurableRootState>;
+export type SafeRootState = AnyIfEmpty<ConfigurableRootState>;
 
 // Omit taken from https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html
 export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
@@ -188,7 +188,7 @@ export interface Connect {
     // tslint:disable:no-unnecessary-generics
     (): InferableComponentEnhancer<DispatchProp>;
 
-    <TStateProps = {}, no_dispatch = {}, TOwnProps = {}, State = SafeRoofState>(
+    <TStateProps = {}, no_dispatch = {}, TOwnProps = {}, State = SafeRootState>(
         mapStateToProps: MapStateToPropsParam<TStateProps, TOwnProps, State>
     ): InferableComponentEnhancerWithProps<TStateProps & DispatchProp, TOwnProps>;
 
@@ -205,12 +205,12 @@ export interface Connect {
         TOwnProps
     >;
 
-    <TStateProps = {}, TDispatchProps = {}, TOwnProps = {}, State = SafeRoofState>(
+    <TStateProps = {}, TDispatchProps = {}, TOwnProps = {}, State = SafeRootState>(
         mapStateToProps: MapStateToPropsParam<TStateProps, TOwnProps, State>,
         mapDispatchToProps: MapDispatchToPropsNonObject<TDispatchProps, TOwnProps>
     ): InferableComponentEnhancerWithProps<TStateProps & TDispatchProps, TOwnProps>;
 
-    <TStateProps = {}, TDispatchProps = {}, TOwnProps = {}, State = SafeRoofState>(
+    <TStateProps = {}, TDispatchProps = {}, TOwnProps = {}, State = SafeRootState>(
         mapStateToProps: MapStateToPropsParam<TStateProps, TOwnProps, State>,
         mapDispatchToProps: MapDispatchToPropsParam<TDispatchProps, TOwnProps>,
     ): InferableComponentEnhancerWithProps<
@@ -224,7 +224,7 @@ export interface Connect {
         mergeProps: MergeProps<undefined, undefined, TOwnProps, TMergedProps>,
     ): InferableComponentEnhancerWithProps<TMergedProps, TOwnProps>;
 
-    <TStateProps = {}, no_dispatch = {}, TOwnProps = {}, TMergedProps = {}, State = SafeRoofState>(
+    <TStateProps = {}, no_dispatch = {}, TOwnProps = {}, TMergedProps = {}, State = SafeRootState>(
         mapStateToProps: MapStateToPropsParam<TStateProps, TOwnProps, State>,
         mapDispatchToProps: null | undefined,
         mergeProps: MergeProps<TStateProps, undefined, TOwnProps, TMergedProps>,
@@ -236,7 +236,7 @@ export interface Connect {
         mergeProps: MergeProps<undefined, TDispatchProps, TOwnProps, TMergedProps>,
     ): InferableComponentEnhancerWithProps<TMergedProps, TOwnProps>;
 
-    <TStateProps = {}, no_dispatch = {}, TOwnProps = {}, State = SafeRoofState>(
+    <TStateProps = {}, no_dispatch = {}, TOwnProps = {}, State = SafeRootState>(
         mapStateToProps: MapStateToPropsParam<TStateProps, TOwnProps, State>,
         mapDispatchToProps: null | undefined,
         mergeProps: null | undefined,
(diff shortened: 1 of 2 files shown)


andrewbranch:
I’d like to wait for feedback from the package owners on this. Thanks for your patience with the holidays. In the meantime, it would be advisable to add a test for this functionality.

Lazyuki:
@andrewbranch thanks for the update. However if I understand it correctly, there's no way to test module augmentation feature at least within DefinitelyTyped? If I augment this module anywhere in the test wouldn't it affect all the other tests?

andrewbranch:
Hm yeah, I think that’s true. I think I’d personally rather see the tests have an augmentation than not, but I guess it’s debatable. I don’t think it’s a blocker for merging. I’ll leave it up to you and the package owners.

Lazyuki:
@andrewbranch Would I need to do something to get attention of the "package owners"? I don't know who those would be.

andrewbranch:
TypeScript bot has already pinged them all: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41031#issuecomment-565658528

Normally after two weeks I would start bugging people more or just go ahead and merge it since it looks reasonable to me, but with the holidays I’m willing to give it until the end of next week.

Lazyuki:
@tkqubo @kenzierocks @clayne11 @tansongyang @NicholasBoll @mDibyo @Kallikrein @val1984 @jrakotoharisoa @apapirovski @surgeboris @soerenbf @MrWolfZ @DylanVann 
bump if you missed this during the holidays

types/react-redux/index.d.ts
```diff
@@ -537,7 +551,7 @@ export function useDispatch<A extends Action = AnyAction>(): Dispatch<A>;
  *   return <div>{counter}</div>
  * }
  */
-export function useSelector<TState, TSelected>(
+export function useSelector<TState = SafeRootState, TSelected = any>(
```
MrWolfZ: I'm not sure I agree with this change. By doing this the type of `TState` becomes `any` when no type parameters are given, while before this would result in `unknown` which seems more reasonable to me

types/react-redux/index.d.ts
```diff
@@ -35,6 +36,18 @@ import {
...
+ * This interface can be augmented by users to add default types for the root state when
+ * using `react-redux`.
+ * Use module augmentation to append your own type definition in a your_custom_type.d.ts file.
+ * https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation
+ */
+// tslint:disable-next-line:no-empty-interface
+export interface ConfigurableRootState {}
+
+export type AnyIfEmpty<T extends object> = keyof T extends never ? any : T;
+export type SafeRootState = AnyIfEmpty<ConfigurableRootState>;
```
MrWolfZ: Making `SafeRootState` `any` instead of `{}` as a default can make certain errors harder to spot

MrWolfZ (commented):
As far as I can see this change seems reasonable enough and should be "safe" in the sense that there is no breaking change for existing projects. However, certain usages of the types become more permissive by using `any` instead of `{}` as a default. Not sure if that is desirable. Since I do not use react-redux in any of my projects I have a hard time to judge how big the benefit of this is. It's probably also quite difficult to know about this feature without looking at the types themselves, so maybe we'd need some hints in the comments of some of the other types that mention this feature.

types/react-redux/index.d.ts
Lazyuki: Good point, I'll fix it

Lazyuki:
Hey @MrWolfZ I've updated some stuff. Can you review again?

types/react-redux/index.d.ts
```diff
@@ -35,6 +36,18 @@ import {
...
+/**
+ * This interface can be augmented by users to add default types for the root state when
+ * using `react-redux`.
+ * Use module augmentation to append your own type definition in a your_custom_type.d.ts file.
+ * https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation
+ */
+// tslint:disable-next-line:no-empty-interface
+export interface DefaultRootState {}
+
+export type ObjectIfEmpty<T extends object> = keyof T extends never ? {} : T;
```
MrWolfZ: Isn't this redundant? Make it an empty object if it is an empty object?

types/react-redux/index.d.ts
```diff
@@ -577,7 +591,7 @@ export interface TypedUseSelectorHook<TState> {
  *   return <div>{store.getState()}</div>
  * }
  */
-export function useStore<S = any, A extends Action = AnyAction>(): Store<S, A>;
+export function useStore<S = SafeRootState, A extends Action = AnyAction>(): Store<S, A>;
```
MrWolfZ: This is a breaking change now. I think you should keep all other places as `{}` while keeping this `any`

types/react-redux/react-redux-tests.tsx
```diff
@@ -1388,7 +1388,8 @@ function testUseStore() {
 
     const untypedStore = useStore();
     const state = untypedStore.getState();
-    state.things.stuff.anything; // any by default
+    // untyped store has {} as the state by default
```
MrWolfZ: this test change shows that this is a breaking change. please ensure this test still works without any changes to it

types/react-redux/react-redux-tests.tsx
```diff
@@ -1400,14 +1401,14 @@ function testUseStore() {
...
-    // $ExpectType <TState, TSelected>(selector: (state: TState) => TSelected, equalityFn?: ((left: TSelected, right: TSelected) => boolean) | undefined) => TSelected
+    // $ExpectType <TState = {}, TSelected = unknown>(selector: (state: TState) => TSelected, equalityFn?: ((left: TSelected, right: TSelected) => boolean) | undefined) => TSelected
     createSelectorHook();
     interface RootState {
         property: string;
     }
     // Should be able to create a version typed for a specific root state.
     const useTypedSelector: TypedUseSelectorHook<RootState> = createSelectorHook();
-    // $ExpectType <S = any, A extends Action<any> = AnyAction>() => Store<S, A>
+    // $ExpectType <S = {}, A extends Action<any> = AnyAction>() => Store<S, A>
```
MrWolfZ: this test change shows that this is a breaking change. please ensure this test still works without any changes to it

MrWolfZ (changes requested):
As written in the code comments, please ensure that your changes do not contain any breaking changes

types/react-redux/index.d.ts
Lazyuki: Yeah this is a breaking change but I'm not sure why it's `any` in the first place. I feel like it should have been an empty object.

types/react-redux/index.d.ts
Lazyuki: Oops, you are right. This can be removed

types/react-redux/index.d.ts
MrWolfZ: That may be, but unfortunately there are already too many apps out there that depend on this. Therefore, we need to revert this change.

MrWolfZ (approved):

oidualc:
This change broke the inline function definition on useSelector on JavaScript files when TypeScript check is enabled (e.g. on VSCode).
For example this expression `useSelector(state => state.myproperty)` now gives the error `Property 'myproperty' does not exist on type 'DefaultRootState'. ts(2339)` because now `state`is of type `DefaultRootState`.
Can this pull request be rethought taking this use case into account?

PR #41042, opened 2019-12-15
Add use-resize-observer types

PetrSvirak:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/use-resize-observer/index.d.ts
+++ b/types/use-resize-observer/index.d.ts
@@ -1,0 +1,105 @@
+// Type definitions for use-resize-observer 5.0
+// Project: https://github.com/ZeeCoder/use-resize-observer#readme
+// Definitions by: PetrSvirak <https://github.com/PetrSvirak>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.5
+export {};
+
+/***
+ * Allows using a ResizeObserver to measure size of an element assigned with ref returned from the hook.
+ * The hook is invoked at least once before the first measurement of actual size by the observer.
+ * @return a ref,
+ *         and ref.current's width (1 before the first measurement),
+ *         and ref.current's height (1 before the first measurement).
+ */
+declare function useResizeObserver(): useResizeObserver.ObserverResultWithDefaultSize;
+
+/***
+ * Allows using a ResizeObserver to measure size of an element assigned with ref returned from the hook.
+ * The hook is invoked at least once before the first measurement of actual size by the observer.
+ * @return a ref (optionally pass in your own),
+ *         and ref.current's width (undefined before the first measurement),
+ *         and ref.current's height (undefined before the first measurement).
+ */
+declare function useResizeObserver<TElement extends HTMLElement>(defaults: DefaultsWithoutSize<TElement>): useResizeObserver.ObserverResultWithoutDefaultSize<TElement>;
+
+/***
+ * Allows using a ResizeObserver to measure size of an element assigned with ref returned from the hook.
+ * The hook is invoked at least once before the first measurement of actual size by the observer.
+ * @return a ref (optionally pass in your own),
+ *         and ref.current's width (defaultWidth the first measurement),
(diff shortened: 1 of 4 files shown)


types/use-resize-observer/index.d.ts
```diff
@@ -0,0 +1,101 @@
+// Type definitions for use-resize-observer 5.0
+// Project: https://github.com/ZeeCoder/use-resize-observer#readme
+// Definitions by: PetrSvirak <https://github.com/PetrSvirak>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.5
```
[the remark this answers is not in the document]
PetrSvirak: I'm not completely sure whether it is [smarter type checking](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html#smarter-union-type-checking) or [better type inference](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html#higher-order-type-inference-from-generic-constructors) or even something else, however, tests for lower TS versions fail with several errors like the following one:
```
ERROR: 12:1  expect  TypeScript@3.4 expected type to be:
  ObserverResultWithoutDefaultSize<HTMLElement>
got:
  ObserverResultWithoutDefaultSize<{}>
```
Fixing this for TS 3.4 and lower breaks tests for TS@next, so I stuck with the 3.5 and newer.

types/use-resize-observer/index.d.ts
```diff
@@ -0,0 +1,101 @@
...
+export {};
+
+/***
+ * Allows using a ResizeObserver to measure size of an element assigned with ref returned from the hook.
+ * The hook is invoked at least once before the first measurement of actual size by the observer.
+ * @return a ref,
+ *         and ref.current's width (1 before the first measurement),
+ *         and ref.current's height (1 before the first measurement).
+ */
+export default function(): ObserverResultWithDefaultSize;
```
andrewbranch: The source library has a `main` entry pointing to a CommonJS module and a `module` entry pointing to an ESM module. You’ve correctly modeled the ESM module here. But, TypeScript doesn’t yet have the ability to model both formats side-by-side for the same module, so unfortunately we have to pick one and ignore the other. For now, I think that should be the `main` entry, so this should be written as a CJS module:

```ts
declare function useResizeObserver(): ObserverResultWithDefaultSize;
declare function useResizeObserver(// ... continue with overloads...

export = useResizeObserver;
```

Users can still do `import useResizeObserver from 'use-resize-observer'` if they have `esModuleInterop` turned on.

andrewbranch (changes requested):

types/use-resize-observer/index.d.ts
PetrSvirak: Makes sense, thanks 😃. 
I did not find a way to export additional interfaces, though 😢. So, I had to remove `ObserverDefaults` and omit exports of `ObserverResult*` from `index.d.ts`. I preserved tests for `ObserverDefaults` even though I had to repeat some types from `index.d.ts` in `the use-resize-observer-tests.ts` – while it is a bit more error-prone, I found the tests quite useful while defining types/expressing possible defaults' variations. See https://github.com/DefinitelyTyped/DefinitelyTyped/commit/e4ef59a31c4cbb0c4ec7f9b3f367cdb77598d10b.

andrewbranch: Oh, here’s how you do that:

```ts
declare function useResizeObserver(): useResizeObserver.ObserverResultWithDefaultSize;
declare function useResizeObserver(// ... continue with overloads...

declare namespace useResizeObserver {
  export interface ObserverResultWithDefaultSize { ... }
}

export = useResizeObserver;
```

PetrSvirak: Thanks 👍, interfaces put back in https://github.com/DefinitelyTyped/DefinitelyTyped/commit/370bfedb863aca89cacee499dfef7685cf5c775a.

andrewbranch (approved):

PR #41049, opened 2019-12-15
Added type definition for "use-position

muscla87:

If adding a new definition:

### DIFF
--- a/types/use-position/index.d.ts
+++ b/types/use-position/index.d.ts
@@ -3,32 +3,30 @@
 // Definitions by: Claudio Musumeci <https://github.com/muscla87>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export namespace UsePosition {
+/**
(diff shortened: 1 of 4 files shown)


types/use-position/index.d.ts
```diff
@@ -0,0 +1,46 @@
+// Type definitions for use-position 0.0
+// Project: https://github.com/trekhleb/use-position
+// Definitions by: Claudio Musumeci <https://github.com/muscla87>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export namespace UsePosition {
```
DanielRosenwasser: Is there a need for the namespace? You can just export `PositionOptions` at the top level. See more at this long-winded explanation https://stackoverflow.com/a/30357635/4386952

muscla87: Thank you for your review. I have removed the unnecessary namespace.

andrewbranch (approved):

PR #41051, opened 2019-12-15
feat(parse/Query): add generic type check for modifiers functions

RaschidJFR:
Follow-up PR for #40429 

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseQuery.js


RaschidJFR:
**Note**: 
please don't merge this immediatly as I'm waiting for #41022 to be merged as there will be conflicting changes.

orta:
There's your conflict ^ ;)

RaschidJFR:
Thanks @orta ready now :)

orta:
The version bump from 3.3 to 3.5 is a breaking change, I'd like at least 2 👍's from the DT maintainers before merging, but it looks pretty solid to me.

rdhelms:
Hi @RaschidJFR - again, sorry for being out of touch for a bit so I know this is a late review, but had a follow up question. I noticed you made the `attributes` parameter in the `Object` and `User` constructors optional, and removed the test for `new Parse.Object` that passed a Type Parameter but no attributes. Was there a particular reason for that? We found it advantageous to force the user to supply an initial value of attributes if they were claiming the attributes to be of a particular type.

As an example, with your changes the following would be valid:
```ts
const thing = new Parse.Object<{ example: string }>('TestObject');
const stringAttr = thing.get('example') // TS thinks this is string but it is actually undefined
```

RaschidJFR:
oh, sorry @rdhelms , I didn't know you'd just added that. But I believe this definition should resemble the original JS source; as attributes are optional in [`ParseUser`](https://github.com/parse-community/Parse-SDK-JS/blob/6dc6e5ca7625adeac1afd3f5fe287ca7ed2f7f3e/src/ParseUser.js#L45) constructor I think they should be too in this TS definition. 

Also, I don't think user should be forced to supply initial values, that should be up to them, don't you think?. for example, I noticed it because I started getting errors in various source files (just after updating the package) because my flow didn't require initial attributes, even though I implemented type parameters, which gave a huge boost to my code.

rdhelms:
Yep, so we did try out using the type parameter in the manner that you're describing, but then ultimately the example that I shared above in which `thing.get('example')` returns an unsafe type seemed like a deal breaker to us. 

When we initially wrote the `Parse.Object` constructor types, we essentially allowed for these two cases:
1) `new Parse.Object('SomeClass')` - The user can create an object without passing attributes by leaving off the type parameter (in this way the initial attributes ARE still optional)
2) `new Parse.Object<{ example: number }>('SomeClass', { example: 10 })` - If the user wants to provide a type parameter, the only way for that type to be accurate and safe is to provide an initial value for attributes.

I would be interested in knowing any use cases where you found value in having a typed Parse Object whose attributes were an empty object (since that's what parse initially sets the attributes to be), because that's exactly what we were trying to avoid.

As another example, we found that refactoring
```ts
const newObject = new Parse.Object('SomeClass')
newObject.set('someProperty', 'someValue')
```
to 
```ts
const newObject = new Parse.Object<{ someProperty: string }>('SomeClass', {
  someProperty: 'someValue'
})
```
to provide a more type-safe way of creating parse objects, and our initial types helped us find places where this needed to be done. It felt a little bit like changing
```ts
let thing
thing = 'stuff'
```
to
```ts
const thing = 'stuff'
```
which felt like an improvement.

And as a last possible workaround, if you DO truly want to create a Parse Object whose attributes are an empty object, but you want to treat its attributes as a particular type, that feels to me like a good case for a type assertion:
```ts
const newThing = new Parse.Object('Thing') as Parse.Object<{ someProperty: string }>
newThing.get('someProperty') // This would still be unsafe, but the assertion above makes that assumption a bit more explicit
newThing.set('someProperty', 'someValue')  // And you would still get the benefit of intellisense, etc here
```

Definitely still open for discussion, though, since this is all so new. The types should end up being whatever the community finds most beneficial, so if more people consider the strictness of our initial constructor types to be overbearing than that's fine. But with the current typings I am a bit concerned about the momentary unsafe state that the Parse Objects are in right after being instantiated.

LinusU:
We also have a lot of places in our code where `Object` is constructed without providing any initial attributes, but this is as you say unsafe since we can get back `undefined` when accessing an attribute later. I would really like to go and clean that up in the manner that you described!

I think that we either should **a)**

- allow not passing attributes at all,
- allow passing `Partial<Attributes>` to the constructor, and
- make `.get(key: K)` return `Attributes[K] | undefined`

or **b)**

- require passing attributes,
- only allow full `Attributes` to be passed to the constructor, and
- keep `.get(key: K)` as returning `Attributes[K]`

Personally, I think that **b** is very much preferred.

-----

It would be a nice "feature" to not requiring passing attributes at all if all properties of `Attributes` are optional. Not sure if this is expressible in TypeScript though 🤔

rdhelms:
@LinusU Your line of thinking is almost identical to what we went through 🙂We initially tried (a) but felt that when a Parse Object was typed as a function parameter, for example, we didn't want it to be partial:
```ts
function processSomeObject(someParseThing: Parse.Object<{ someProperty: string }>) {
  someParseThing.get('someProperty') // We felt that it should be safe to assume this is string instead of having to narrow string | undefined
}
```

And I really like what you mentioned about not requiring the parameter if all the properties were optional - we did attempt this and hit a wall with TS but I would love if that could be possible. Splitting the `class Object` into `interface Object` and `interface ObjectConstructor` was one example of the TS gymnastics necessary to even get (b) to work in such a way that we could keep backward compatibility for users that didn't specify a type parameter at all. Getting the "optional param when all optional props" to work seemed beyond even that.

What you describe as (b) is what we ultimately went with and felt happy with (with the additional note that the initial attributes are still optional if no type parameter is passed), although I can see and appreciate some of the inconvenience that @RaschidJFR brought up.

RaschidJFR:
Hi guys, sorry for the delay. I've been going over the commit and running some tests in my own code. I just realized that we may be talking about two different subjects here. Let me explain:

I agree with you on the constructor for `ParseObject`:

> 1. `new Parse.Object('SomeClass')` - The user can create an object without passing attributes by leaving off the type parameter (in this way the initial attributes ARE still optional)
> 2. `new Parse.Object<{ example: number }>('SomeClass', { example: 10 })` - If the user wants to provide a type parameter, the only way for that type to be accurate and safe is to provide an initial value for attributes.

I don't agree on the constructor for `ParseUser`. I don't see a reason not to let the developer create a User object passing an initial set of attributes (the original JS SDK works this way). 

That's why I added:
 
```
new(attributes?: Attributes): User;
```
instead of 
```
new(): User;
```

I added this last comment to the review on #41484 so we may continue the conversation there. See you there guys :)

PR #41065, opened 2019-12-16
expose __RouterContext in react-router-dom

artola:
Please fill in this template.

  https://github.com/ReactTraining/react-router/blob/0c058adc5b56c6d9b409af5e4d8330c645fac00c/packages/react-router-dom/modules/index.js#L1

  https://github.com/ReactTraining/react-router/blob/0c058adc5b56c6d9b409af5e4d8330c645fac00c/packages/react-router/modules/index.js#L38


eps1lon (changes requested):
The name implies that it is a private member. Could you link to documentation or an official statement that makes it clear it is indeed a public member?

artola:
> The name implies that it is a private member. Could you link to documentation or an official statement that makes it clear it is indeed a public member?

The source code exports `export { default as __RouterContext } from "./RouterContext.js";` because it was an old time issue, while prefixing with `__` is a convention for internal, there is no such when it is exported from the index. From my point of view, the typings just reflect the code, no other intention.

Might be in a near future changing again, see:
https://github.com/ReactTraining/react-router/pull/6753#issue-281365263

eps1lon:
The linked issue makes it clear that it is not part of the public API.

artola:
> The linked issue makes it clear that it is not part of the public API.

If it is not part of the API, then should not be exposed by the library, but it is at least in v5.1 (for this typings) and still until v5.12
For a couple of years we used it and we augmented the typings in our codebase.

Consideration/question: are the typings "opinionated"  or should them match the library code as accurate as possible?

p-jackson:
The problem I see with adding typings for things not in the public API is versioning. `react-router-dom` could remove `__RouterContext` in a minor version change, but removing from the typings would require a major version change.

This seems like a situation where it's better to add the necessary typings to your own project instead of the shared package. That way you can weigh whether the pros and cons make sense for your situation.

eps1lon:
> Consideration/question: are the typings "opinionated" or should them match the library code as accurate as possible?

They should match the public API.

For class methods this is may be clearer: 

```ts
class Foo {
  private run() {};
}
```

will emit no runtime error for 

```js
const foo = new Foo();
foo.run();
```

but the declaration file will not contain a type for `foo.run()`. The type checker would reject that code while the runtime wouldn't. 

`__RouterContext` is such a private member per https://github.com/ReactTraining/react-router/pull/6753#issue-281365263

Hope this clears things up. In other words: Just because something is codified doesn't mean it's part of the public API.

PR #41082, opened 2019-12-17
Add @rdfjs/namespace

thewilkybarkid:
Moved from https://github.com/rdfjs-base/namespace/pull/5. (Tried to improve it a bit, still a little confused by CommonJS exports...)

If adding a new definition:

### DIFF
--- a/types/rdfjs__namespace/index.d.ts
+++ b/types/rdfjs__namespace/index.d.ts
@@ -3,19 +3,15 @@ import { DataFactory, NamedNode } from 'rdf-js';
 declare function namespace(baseIRI: string, options?: namespace.BuilderOptions): namespace.NamespaceBuilder;
 
 declare namespace namespace {
-
-    export interface NamespaceBuilder {
-        (property: TemplateStringsArray): NamedNode;
-
-        (property: string): NamedNode;
+    interface NamespaceBuilder {
+        (property: TemplateStringsArray | string): NamedNode;
 
         readonly [property: string]: NamedNode;
     }
 
-    export interface BuilderOptions {
+    interface BuilderOptions {
         factory?: DataFactory;
     }
-
 }
 
 export = namespace;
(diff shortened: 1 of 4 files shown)


thewilkybarkid:
Not sure what I'm doing wrong. Had problems with using `dts-gen` to create both the `@rdfjs/namespace` and `rdfjs__namespace` forms.

types/rdfjs__namespace/index.d.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { DataFactory, NamedNode } from 'rdf-js';
+
+declare function namespace(baseIRI: string, options?: namespace.BuilderOptions): namespace.NamespaceBuilder;
+
+declare namespace namespace {
+    interface NamespaceBuilder {
+        (property: TemplateStringsArray | string): NamedNode;
+
+        readonly [property: string]: NamedNode;
```
thewilkybarkid: My IDE (PhpStorm) complains if using the property `name` (confuses it with `Function.name`, so thinks it's a `string` instead).

TypeScript and eslint don't have the problem though; not sure if there's a way to keep both happy.

thewilkybarkid: Adding an explicit `name` property gets round it, but  should be unnecessary.

tpluscode: My eslint setup complains about that `name` property. Here's another solution

```
interface NamespaceBuilderFunc {
  (property: TemplateStringsArray | string): NamedNode;
}
type NamespaceBuilder = Record<string, NamedNode> & Omit<NamespaceBuilderFunc, 'name'>
```

thewilkybarkid: Hmm, doesn't seem to work, no longer recognised as callable...

andrewbranch (approved):

types/rdfjs__namespace/index.d.ts
stephenwf: @thewilkybarkid I think it's unavoidable to redeclare the properties from the `Function.prototype` (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/prototype)

If Typescript did override, then it would have a knock on effect on things like objects:
```typescript
type A = {
 [property: string]: SomeType;
};

const t: A = { };

t.toString(); // is this a function or `SomeType`
```

thewilkybarkid: Opened https://stackoverflow.com/questions/59560965/name-property-when-a-type-is-both-indexable-and-callable

PR #41084, opened 2019-12-17
17.4.39 added

Syncfusion-JavaScript:


andrewbranch (changes requested):
@Syncfusion-JavaScript if this account is an owner of these definitions, please add it to the `Definitions by:` in the header so future PRs will get fast tracked.

PR #41085, opened 2019-12-17
[@types/ioredis] zrangebyscore and zrevrangebyscore accept strings as min/max

constb:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://redis.io/commands/zrangebyscore


plantain-00 (approved):

orta:
Looks good, would be nice to get some tests with changes like these in the future but not a blocker 👍

demian85:
Hi, can you please explain why `hgetall` now has `number` as a possible return type?
```
hgetall(key: KeyType, callback: (err: Error, res: { [key: string]: string | number }) => void): void;
hgetall(key: KeyType): Promise<{ [key: string]: string | number }>;
```
It does not make sense. Redis stores strings only, how can ioredis know if key has a number? What kind of parsing is it doing there? It's confusing because a few types broke in my app. I was expecting strings but know all of a sudden you say it can return numbers.

PR #41087, opened 2019-12-17
recharts: Add ReferenceLine position prop

pgjones:
Recharts' ReferenceLines have gained the ability for a position
to be specified, https://github.com/recharts/recharts/pull/1685.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/recharts/recharts/pull/1685

### DIFF
--- a/types/recharts/index.d.ts
+++ b/types/recharts/index.d.ts
@@ -55,6 +55,7 @@ export type LineType =
 export type IfOverflowType = 'hidden' | 'visible' | 'discard' | 'extendDomain';
 export type AxisInterval = number | 'preserveStart' | 'preserveEnd' | 'preserveStartEnd';
 export type BaseValueType = number | 'auto' | 'dataMin' | 'dataMax';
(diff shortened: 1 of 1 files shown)


gndelia:
Thanks for your PR! 
Docs from `recharts` are always out of date :(

I see those `position` values are also used in another element. Could we create a type from them, in order to reuse them? Like

```ts
type Position = 'start' | 'middle |, 'end'

// then in reference line
position?: Position

// and then in TextProps
textAnchor?: Position | typeof 'inherit';
verticalAnchor?: Position;
```

or something like that

pgjones:
@gndelia looks like recharts is switching to Typescript with version 2, is this change worth merging in this case?

gndelia:
@pgjones  I would merge it either way because there might be some devs that cannot migrate to v2 (for whatever reason) and this might still be useful for them. Then we can create a stub for next releases

andrewbranch (approved):

PR #41102, opened 2019-12-17
Update Figma API typings to version 9

rudi-c:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.figma.com/plugin-docs/api/typings/

### DIFF
--- a/types/figma/index.d.ts
+++ b/types/figma/index.d.ts
@@ -7,12 +7,14 @@
 //   rudi@figma.com, jonas@figma.com
 // While changes to the types aren't breaking changes to how plugins run,
 // it is preferable to coordinate breaking type changes with breaking API changes.
+// TypeScript Version: 2.7
 
-// Figma Plugin API version 1, update 5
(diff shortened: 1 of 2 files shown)


types/figma/index.d.ts
```diff
@@ -713,3 +808,6 @@ interface Image {
     readonly hash: string;
     getBytesAsync(): Promise<Uint8Array>;
 }
+} // declare global
+
+export {};
```
andrewbranch: Just curious, why the change to a global-augmenting module from the global declarations?

rudi-c: The global `figma` variable was no longer being automatically picked up by `tsc` and needed an import statement, this happened to fix it.

andrewbranch: I would think that _after_ this change, it would require an `import 'figma'` to get the globals included. If you’re trying to get the globals to show up without an import declaration, can you send me a repro of it not being picked up so I can dig into that?

PR #41105, opened 2019-12-18
Adding support for viz.js

mckaysalisbury:
Please fill in this template.

- [☑️] Use a meaningful title for the pull request. Include the name of the package modified.
- [✔️] Test the change in your own code. (Compile and run.)
- [✅] Add or edit tests to reflect the change. (Run with `npm test`.)
- [✓] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [☑] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [✔] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If adding a new definition:
- [✅] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration`
- [✓] If this is for an NPM package, match the name. If not, do not conflict with the name of an NPM package.
- [☑️] Create it with `dts-gen --dt`, not by basing it on an existing project.
- [Accurate, but not complete (see comment in `common.d.ts`)] Represents shape of module/library [correctly](https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html)
- [✔️] `tslint.json` should be present and it shouldn't have any additional or disabling of rules. Just content as `{ "extends": "dtslint/dt.json" }`. If for reason the some rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]`  and not for whole package so that the need for disabling can be reviewed.
- [✅] `tsconfig.json` should have `noImplicitAny`, `noImplicitThis`, `strictNullChecks`, and `strictFunctionTypes` set to `true`.

### DIFF
--- a/types/viz.js/common.d.ts
+++ b/types/viz.js/common.d.ts
@@ -1,0 +1,23 @@
+interface Options {
+  format?: string;
+  engine?: string;
+  files?: string[];
+  images?: string[];
+  yInvert?: boolean;
+}
+
+/**
+ * This interface defines the shape of an object that is held by the caller.
+ * This `Module` was created by emscripten, and is therefore largely arcane.
+ * This currently just lists a subset of what is defined in `Module`.
+ */
+interface Module {
+  run(): void;
+}
+
+type renderFunction = (instance: Module, src: string, options: Options) => string;
+
+declare class Viz {
+  constructor(arg: { Module: Module; render: renderFunction });
+  renderString(src: string, options?: Options): Promise<string>;
+}
--- a/types/viz.js/common.d.ts
+++ b/types/viz.js/common.d.ts
@@ -15,9 +15,9 @@ interface Module {
   run(): void;
 }
 
-type renderFunction = (instance: Module, src: string, options: Options) => string;
+type RenderFunction = (instance: Module, src: string, options: Options) => string;
 
 declare class Viz {
-  constructor(arg: { Module: Module; render: renderFunction });
+  constructor(arg: { Module: Module; render: RenderFunction });
   renderString(src: string, options?: Options): Promise<string>;
 }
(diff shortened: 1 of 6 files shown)


types/viz.js/full.render.js.d.ts
```diff
@@ -0,0 +1,3 @@
+/// <reference path="common.d.ts" />
```
DanielRosenwasser: Do these files need to exist? Can you consolidate into one file to reflect the package structure, or at least just use modules so that the declarations within those files don't leak into the global scope?

types/viz.js/common.d.ts
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * This interface defines the shape of an object that is held by the caller.
+ * This `Module` was created by emscripten, and is therefore largely arcane.
+ * This currently just lists a subset of what is defined in `Module`.
+ */
+interface Module {
+  run(): void;
+}
+
+type renderFunction = (instance: Module, src: string, options: Options) => string;
```
DanielRosenwasser: Nit: Most type names start with a capital letter by convention.

DanielRosenwasser (changes requested):

types/viz.js/common.d.ts
mckaysalisbury: Resolved.

types/viz.js/full.render.js.d.ts
mckaysalisbury: When I had it as a single file, the DefinitelyTyped linter complained because I was declaring modules, so I split it out. So, I guess I'm not sure how to do what you're requesting and still keep the linter happy.

```
declare module 'viz.js' {
  export = Viz;
}

declare module 'viz.js/full.render.js' {
  export const Module: IModule;
  export function render(instance: IModule, src: string, options: IOptions): string;
}
```

mckaysalisbury:
@DanielRosenwasser Is there anything else I need to do to move forward? This has been stuck for a few months.

PR #41121, opened 2019-12-18
Add mitsobox

mrclemds:
Add mitsobox


types/mitsobox/tslint.json
```diff
@@ -0,0 +1,80 @@
...
+        "space-before-function-paren": false,
+        "space-within-parens": false,
+        "strict-export-declare-modifiers": false,
+        "trim-file": false,
+        "triple-equals": false,
+        "typedef-whitespace": false,
+        "unified-signatures": false,
+        "void-return": false,
+        "whitespace": false
+    }
```
andrewbranch: image

andrewbranch (changes requested):

types/mitsobox/index.d.ts
```diff
@@ -0,0 +1,32 @@
+// Type definitions for mitsobox 0.1
+// Project: https://github.com/jimfilippou/mitsobox
+// Definitions by: MrClemds <https://github.com/MrClemds>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.7
```
andrewbranch: Why does this require TypeScript 3.7?

andrewbranch (changes requested):

types/mitsobox/index.d.ts
mrclemds: I thought it was necessary to indicate this information and so I added the one I use.

types/mitsobox/tslint.json
mrclemds: Okay, I'm sorry, this is a mistake.

andrewbranch (approved):
Thanks for the updates!

PR #41125, opened 2019-12-18
[react-datepicker] Update `@types/react-datepicker` to comply with `react-datepicker` v2.10

WtheFourth:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/Hacker0x01/react-datepicker/blob/master/src/index.jsx#L111>

### DIFF
--- a/types/react-datepicker/index.d.ts
+++ b/types/react-datepicker/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for react-datepicker 2.9
+// Type definitions for react-datepicker 2.10
 // Project: https://github.com/Hacker0x01/react-datepicker
 // Definitions by: Rajab Shakirov <https://github.com/radziksh>,
 //                 Andrey Balokha <https://github.com/andrewBalekha>,
@@ -15,7 +15,7 @@
 //                 Kerry Gougeon <https://github.com/kerry-g>
 //                 Walter Kennedy <https://github.com/wthefourth>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.1
 
 import * as React from 'react';
 import * as Popper from 'popper.js';
@@ -32,6 +32,7 @@ interface HighlightDates {
 export interface ReactDatePickerProps {
     adjustDateOnChange?: boolean;
     allowSameDay?: boolean;
+    ariaLabelledBy?: string;
     autoComplete?: string;
     autoFocus?: boolean;
     calendarClassName?: string;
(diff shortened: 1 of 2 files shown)


types/react-datepicker/index.d.ts
```diff
@@ -72,14 +74,18 @@ export interface ReactDatePickerProps {
     nextMonthButtonLabel?: string;
     nextYearButtonLabel?: string;
     onBlur?(event: React.FocusEvent<HTMLInputElement>): void;
+    onCalendarClose?(...args: any): void;
+    onCalendarOpen?(...args: any): void;
```
tu4mo: What are these `args` used for?

WtheFourth: Those functions are called [here](https://github.com/Hacker0x01/react-datepicker/blob/master/src/index.jsx#L259). The [provided `PropType` indicates those two](](https://github.com/Hacker0x01/react-datepicker/blob/master/src/index.jsx#L162)) are of type `PropTypes.func`, so I went to the type definitions for `prop-types` to get that [definition](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/prop-types/index.d.ts#L61).

I imagine these lines could be changed to:
```
onCalendarClose?(): void;
onCalendarOpen?(): void;
```
... as `react-datepicker` itself doesn't provide a callback signature, but I left it as `...args` in case the function assigned to either accepted arguments.

tu4mo: I'd change them to what you suggested. `react-datepicker` doesn't provide any arguments for the callbacks and the types should reflect that.

WtheFourth: I'll get that updated this afternoon and pushed over.

types/react-datepicker/index.d.ts
```diff
@@ -15,7 +15,7 @@
 //                 Kerry Gougeon <https://github.com/kerry-g>
 //                 Walter Kennedy <https://github.com/wthefourth>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.1
```
DanielRosenwasser: Why's this need 3.1? You can probably just remove this now and let it implicitly rely on 2.8

WtheFourth: It needed 3.1 for the `(...args: any)`, but I'm taking that part out most likely. I'll try and get a new PR submitted ASAP.

types/react-datepicker/index.d.ts
WtheFourth: I've updated the definitions to reflect suggested changes with commit 7b16773b145f8fd9c2fa58b5932178bd951c86c3 below. @tu4mo

tu4mo: LGTM 👍

tu4mo (approved):

PR #41126, opened 2019-12-18
kurento-client package definition file

jahilldev:
Created typescript definitions for the `kurento-client` package as they are currently without.

If adding a new definition:

### DIFF
--- a/types/kurento-client/index.d.ts
+++ b/types/kurento-client/index.d.ts
@@ -7,15 +7,15 @@
 type Callback<T> = (error: Error, result: T) => void;
 
 interface Options {
-    failAfter: number;
-    enableTransactions: boolean;
-    useImplicitTransactions: boolean;
-    strict: boolean;
-    request_timeout: number;
-    response_timeout: number;
-    duplicates_timeout: number;
-    access_token: string;
-    socket: any;
+    failAfter?: number;
+    enableTransactions?: boolean;
+    useImplicitTransactions?: boolean;
+    strict?: boolean;
+    request_timeout?: number;
+    response_timeout?: number;
+    duplicates_timeout?: number;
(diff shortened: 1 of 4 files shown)


types/kurento-client/index.d.ts
```diff
@@ -0,0 +1,132 @@
...
+
+interface Kurento {
+    (ws_uri: string, options?: Options): Promise<KurentoClient>;
+    getComplexType: (complex: 'IceCandidate') => (value: any) => any;
+}
+
+declare const kurento: Kurento;
+
+// tslint:disable-next-line npm-naming
+export default kurento;
```
andrewbranch: Looks like the library is a CommonJS module, not an ES Module: https://unpkg.com/kurento-client@6.13.0/lib/index.js

Should be something like

```ts
declare namespace kurento {
  // exported types go here
}

declare const kurento: kurento.Kurento;

export = kurento;
```

andrewbranch (changes requested):

types/kurento-client/index.d.ts
jhdevuk: Yeh, the reason I structured the types like this is so an end user can import isolated interfaces from the types, so this structure:

```
import kurento, { MediaPipeline } from 'kurento-client';
...
const client = kurento(...);
...
class Something {
  private pipeline: MediaPipeline;
  ...
}
```
If this is wrong I'll make the relevant changes though. Unless there's an alternative method of providing this that I'm missing?

Thanks for the feedback!

jhdevuk:
@andrewbranch Any more thoughts on this? Would love to get this PR in and published as soon as possible. Thanks! 👍

andrewbranch (approved):

PR #41138, opened 2019-12-19
[google__maps][@google/maps] Only allow calling `asPromise()` if client supports it

pheromonez:
All requests created with the `GoogleMapsClient` currently allow calling `asPromise()` even if the client was not created with the `Promise` option set.
This change will only add the `asRequest()` function to requests if they are made using a client that was created with `Promise` specified.

This makes use of a TypeScript 2.8 feature, [Conditional Types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#conditional-types), which will break compatibility with projects using older Typescript versions.

I've made the assumption that most people will be using Promises, so the behaviour of `GoogleMapsClient` will support `asPromise` by default. If anyone is using the `GoogleMapsClient` directly in their code then the type definitions will assume that promises are supported.
``` typescript
func doSomething (client: GoogleMapsClient) {
  // Promise supported by default
  client.geocode({...}).asPromise()...
}
```

On the other hand, if the user is not supporting promises then there will be a little more work, unfortunately.
``` typescript
func doSomething (client: GoogleMapsClient<"WITHOUT_PROMISE">) {
  // Promise not supported; will indicate that `asPromise` is not defined on type
  client.geocode({...}).asPromise()...
}
```

It _might_ be worth introducing a type alias for `GoogleMapsClient<"WITHOUT_PROMISE">` called `PromiselessGoogleMapsClient` or something, but ideally people wouldn't need to be using these types in their own code, and they would just be using the client in the way shown in the tests.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [Source code](https://github.com/googlemaps/google-maps-services-js/blob/1e20ccc7264e8858723dcf4971d9ad10dc9e5ff2/lib/internal/make-api-call.js#L149); `asPromise` only added if `options.Promise` is defined

### DIFF
--- a/types/google__maps/index.d.ts
+++ b/types/google__maps/index.d.ts
@@ -1,13 +1,14 @@
 // Type definitions for @google/maps 0.5
 // Project: https://github.com/googlemaps/google-maps-services-js
 // Definitions by: Indri Muska <https://github.com/indrimuska>
+//                 Paul Huynh <https://github.com/pheromonez>
 // Definitions: https://github.com/indrimuska/google-maps-api-typings
-// TypeScript Version: 2.3
+// TypeScript Version: 2.8
 
 /**
  * Creates a Google Maps client. The client object contains all the API methods.
  */
-export interface CreateClientOptions {
+export interface CreateClientOptionsBase {
     /** API key (required, unless clientID and clientSecret provided). */
     key: string;
     /** Maps API for Work client ID. */
@@ -20,13 +21,20 @@ export interface CreateClientOptions {
     timeout?: number;
     /** Default language for all queries. */
     language?: Language;
-    /** Promise constructor (optional). */
-    Promise?: PromiseConstructor;
     /** Rate options. */
     rate?: RateOptions;
     /** Retry options. */
     retryOptions?: RetryOptions;
 }
+/** Create a Google Maps client, with Promise support. */
+export interface CreateClientOptionsWithPromise extends CreateClientOptionsBase {
+    /** Promise constructor */
+    Promise: PromiseConstructor;
+}
+/**
+ * Creates a Google Maps client. The client object contains all the API methods.
+ */
+export type CreateClientOptions = CreateClientOptionsBase | CreateClientOptionsWithPromise;
 
 export interface RateOptions {
     /** Controls rate-limiting of requests. Maximum number of requests per period. (Default: 50). */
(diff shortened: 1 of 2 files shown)


indrimuska:
Hi @pheromonez, you could try using type union (`ClientWithPromise | ClientWithouthPromise`) to any type of client with no need of explicitly specifying the parameter type to understand if you want to use promises or not.
Also this way you don’t need to require a newer version of ts.

pheromonez:
Hi @indrimuska, thanks for the feedback.

If I understand you correctly, you are suggesting that we could do something like the following:
``` typescript
function createClient(options: Options): ClientWithPromise | ClientWithoutPromise;
```
This would allow the caller to manually cast the resulting client to use promise or not.
``` typescript
const client = createClient({...}) as ClientWithoutPromise;
```

The issue I have with this is the the caller may attempt to use the client with the assumption that promises are supported, when in fact the client does not support it.
``` typescript
// Note: created without promise support
const client = createClient({ key }) as ClientWithPromise;
// Runtime error; function not defined!
client.geocode({...}).asPromise().then(...)
```

 Please feel free to correct me if I have misunderstood you.

indrimuska:
Exactly, you get the point! You don't need to cast the client, if you properly provide different version of the client, something like:

```ts
interface Options { a, b, c }
interface OptionsWithPromise extends Options {
   Promise: PromiseConstructor;
}

declare function createClient(options: Options): Client;
declare function createClient(options: OptionsWithPromise): ClientWithPromise;
```

pheromonez:
Hi @indrimuska, I've updated this PR with the changes you've suggested. I tried using this with Typescript 2.3 (the original version in the type definitions) and it seems to be working okay.

types/google__maps/index.d.ts
```diff
@@ -1,6 +1,7 @@
 // Type definitions for @google/maps 0.5
 // Project: https://github.com/googlemaps/google-maps-services-js
 // Definitions by: Indri Muska <https://github.com/indrimuska>
+//                 Paul Huynh <https://github.com/pheromonez>
```
indrimuska: only the main maintainer should be present in the credits section

types/google__maps/index.d.ts
```diff
@@ -255,7 +264,12 @@ export type Language = (
 );
 
 export type GoogleMapsClientEndpoint<Request, Response> = (query: Request, callback?: ResponseCallback<Response>) => RequestHandle<Response>;
+export type GoogleMapsClientEndpointWithPromise<Request, Response> = (query: Request, callback?: ResponseCallback<Response>) => RequestHandleWithPromise<Response>;
 
+/**
+ * Google Maps clieant that provides access to all the APIs.
+ * @param Promisable Determine if `asPromise()` is available in request. Default is WITHOUT_PROMISE.
```
indrimuska: `GoogleMapsClient` is not a parametric type, this must be a leftover of the previous commit

indrimuska (changes requested):

indrimuska (approved):

PR #41140, opened 2019-12-19
power-assert: Revert updates

falsandtru:
@rbuckton Fixes breaks of existing tests.


types/power-assert/ts3.7/index.d.ts
```diff
@@ -27,7 +27,7 @@ declare namespace assert {
     }
 
     export function fail(actual?: any, expected?: any, message?: string, operator?: string): never;
-    export function ok(value: any, message?: string): asserts value;
```
andrewbranch: Without these, I think there’s no need for a `ts3.7` folder, right?

andrewbranch (changes requested):
Can you expound on what was breaking / why this is necessary?

types/power-assert/ts3.7/index.d.ts
falsandtru: I'm keeping this folder to introduce asserts predicate again after the problems are resolved.

falsandtru:
See https://github.com/microsoft/TypeScript/issues/35567. The previous update broke a lot of existing tests.

andrewbranch:
Hm. I’d like to wait on @vvakame on this one. It’s unclear still what’s going to happen with https://github.com/microsoft/TypeScript/issues/35567, and users can downgrade their version of these typings if needed.

falsandtru:
Probably @vvakame is no longer maintaining this definition. And fixing the version of some packages with updates is a pain. Please accept this. If anyone really needs asserts predicate, he will request it. If not, holding the breaks on the actual projects has no sense. I believe waiting for requests of the actual users is the best way.

andrewbranch:
Can you show me reports (other than your own) of people who are negatively affected by the status quo? https://github.com/microsoft/TypeScript/issues/35567 has no reactions and very few comments. Does this fix an issue filed on DefinitelyTyped? Have people been complaining about it in the source library’s repo? I just want to minimize unnecessary churn. It’s very possible that @vvakame is no longer maintaining this, but also a lot of people are taking time off for the holidays right now. I’m not rushing in to merge things I lack context on while people are on vacation.

falsandtru:
Unfortunately probably there is no other active user of this definition. But at least this new feature, asserts predicate, is not used even on https://github.com/microsoft/TypeScript. @types/node is too. Why you don't use your feature on your project before forcing users to use it? I can wait for a few weeks but please don't force users to use such an immature feature.

falsandtru:
FYI, @rbuckton failed to introduce asserts predicate to TypeScript (https://github.com/microsoft/TypeScript/issues/35638).

falsandtru:
@andrewbranch Any blocking issues without vacation?

andrewbranch:
I would personally like to see the 3.7 folder removed with this change, as microsoft/TypeScript#35567 has not yet been marked as actionable, and there’s no point in leaving useless files around when we have git. But otherwise, I expect whoever is on DT rotation next week will merge it then.

falsandtru:
@andrewbranch I removed all files related 3.7 folder but looks like the CI has a bug.

andrewbranch (approved):

PR #41169, opened 2019-12-20
Update jest-axe tests for new axe-core version

sandersn:
I just fixed the type error; the actual value, `[]` is not valid at runtime because it is supposed to contain a list of rules. But the jest-axe test doesn't have any rules either. So the test is already invalid at runtime.


sandersn:
This caused the overnight test to fail for a couple days; I made a minimal fix but, types authors, if you want to improve it, please do.

PR #41173, opened 2019-12-20
feat(@hapi/joi): prevent incorrectly infering Joi.object type from user provided parameters

### LINKED ISSUE #41128: [@types/hapi__joi] Definition for Joi.append() and Joi.keys() doesn't allow object extensions

- [X] I tried using the `@types/hapi__joi` package and had problems.
- [X] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
  - Authors: @Bartvds, @myers, @cglantschnig, @SW, @GaelMagnan, @ralekna, @schfkt, @rokoroku, @dankraus, @wanganjun, @rafaelkallis, @aconanlai, @zaphoyd, @thewillg, @SimonSchick, @afharo, @lenovouser, @AnandChowdhary, @myovchev, @RecuencoJones, @everhardt
 

Trying to extend an object in Joi yields Typescript errors. Sample code:

```
export const schema = Joi.object({
    a: Joi.string()
});

export const extendedSchema = schema.append({
    b: Joi.string()  <<< ERROR
});

export const extendedSchema2 = schema.keys({
    b: Joi.string()    <<< ERROR
});
```
The error reported is: 
```
Argument of type '{ b: Joi.StringSchema; }' is not assignable to parameter of type 'SchemaMap<{ a: unknown; }>'.
  Object literal may only specify known properties, and 'b' does not exist in type 'SchemaMap<{ a: unknown; }>'.ts(2345)
```

HosseinAgha:

This fixes #41128, @SimonSchick.

### DIFF
--- a/types/hapi__joi/index.d.ts
+++ b/types/hapi__joi/index.d.ts
@@ -1918,7 +1918,8 @@ declare namespace Joi {
         /**
          * Generates a schema object that matches an object data type (as well as JSON strings that have been parsed into objects).
          */
-        object<TSchema = any>(schema?: SchemaMap<TSchema>): ObjectSchema<TSchema>;
+        // tslint:disable-next-line:no-unnecessary-generics
+        object<TSchema = any, T = TSchema>(schema?: SchemaMap<T>): ObjectSchema<TSchema>;
 
         /**
          * Generates a schema object that matches a string data type. Note that empty strings are not allowed by default and must be enabled with allow('').
--- a/types/hapi__joi/hapi__joi-tests.ts
+++ b/types/hapi__joi/hapi__joi-tests.ts
@@ -1107,6 +1107,37 @@ schema = Joi.any();
 const terms = schema.$_terms;
 
 // --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- --- ---
+// Test Joi.object, Joi.append and Joi.extends (with `any` type)
+
+// should be able to append any new properties
+let anyObject = Joi.object({
+    name: Joi.string().required(),
+    family: Joi.string(),
+});
+
+anyObject = anyObject.append({
+    age: Joi.number(),
+}).append({
+    height: Joi.number(),
+});
+
+anyObject = anyObject.keys({
+    length: Joi.string()
+});
+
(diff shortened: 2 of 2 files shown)


types/hapi__joi/index.d.ts
```diff
@@ -1918,7 +1918,8 @@ declare namespace Joi {
         /**
          * Generates a schema object that matches an object data type (as well as JSON strings that have been parsed into objects).
          */
-        object<TSchema = any>(schema?: SchemaMap<TSchema>): ObjectSchema<TSchema>;
+        // tslint:disable-next-line:no-unnecessary-generics
+        object<TSchema = any, T = TSchema>(schema?: SchemaMap<T>): ObjectSchema<TSchema>;
```
rokoroku: I think it would be better to add type argument to `ObjectSchema.keys()` method, not to the object.

e.g.
```
        /**
         * Sets or extends the allowed object keys.
         */
        keys(schema?: SchemaMap<TSchema>): this;
        keys<TExtendSchema>(schema: SchemaMap<TExtend>): ObjectSchema<TSchema & TExtendSchema>;
```

HosseinAgha: Good idea, but this will be a breaking change and I could not make it work. I defined the Overwrite type as our current version of TypeScript does not support it but Because of type inference user should provide the `TExtension` if they want type checking. for example here:
The `TSchema` is `any` so the returned type of first `keys` is `ObjectSchema<Overwrite<any, User>>` (which is `any`). So other keys will be `any`.
```ts
const userSchemaError = Joi.object().keys<User>({
    name: Joi.string().required(),
    family: Joi.string(),
    height: Joi.number(), // $ExpectError
}).keys<{ more: number }>({
    notMore: Joi.number,
});
```
As users usually start their types with a `Joi.object` I think current workaround is our best option. (Define all possible future fields using `TSchema` of `Joi.object<TSchema>`).

HosseinAgha: Good Idea. I actually spent some time to implement this but due to typescript type inference user should provide the Schema generic parameter on every validator extension in order to force TS to use his/her types. IMO defining all possible future fields in the `Joi.object` is a safer pattern.

PR #41190, opened 2019-12-22
[codemirror] EditorConfiguration.scrollbarStyle

dperetti:
`EditorConfiguration.scrollbarStyle` must not be set to `undefined` (CodeMirror would fail to set up because of [how initScrollbars() works](https://github.com/codemirror/CodeMirror/blob/ed8dfeb5e2ed25b5dd1f1eccc7b757ca6dbd118d/src/display/scrollbars.js#L180), expecting a [valid key for scrollbarModel](https://github.com/codemirror/CodeMirror/blob/ed8dfeb5e2ed25b5dd1f1eccc7b757ca6dbd118d/src/display/scrollbars.js#L171)). Scrollbars should be disabled with `null`.

### DIFF
--- a/types/codemirror/index.d.ts
+++ b/types/codemirror/index.d.ts
@@ -941,7 +941,7 @@ declare namespace CodeMirror {
          * Chooses a scrollbar implementation. The default is "native", showing native scrollbars. The core library also
          * provides the "null" style, which completely hides the scrollbars. Addons can implement additional scrollbar models.
          */
-        scrollbarStyle?: string;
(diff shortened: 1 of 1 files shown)


peterblazejewicz:
@dperetti 
Can you review the failed CI build log?
https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/628243306?utm_source=github_status&utm_medium=notification
You should get the same output feedback running `npm run lint codemirror` from your branch
Thanks!

PR #41201, opened 2019-12-23
[esquery] use discriminated union for Selector

bradzacher:
Please fill in this template.

- [ ] ~~Provide a URL to documentation or source code which provides context for the suggested changes:~~ N/A
- [ ] ~~If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.~~ N/A
- [ ] ~~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.~~ N/A

---

The initial types returned the "super type" `Selector` interface. This makes it really hard to use the types, because you have to do manual assertion casts after checking the `type` prop:
```ts
const selector = esquery.parse(selectorString);
switch (selector.type) {
    case 'adjacent':
        // $ExpectError
        selector.left;

        const adjacentSelector = selector as esquery.Adjacent;
        // $ExpectType SubjectSelector
        selector.left;
        break;

    case 'attribute':
        // $ExpectError
        selector.name;

        const attributeSelector = selector as esquery.Attribute;
        // $ExpectType string
        selector.name;
}
```

This PR just introduces a discriminated union on top of the existing types, so that this now "just works":
```ts
const selector = esquery.parse(selectorString);
switch (selector.type) {
    case 'adjacent':
        // $ExpectType SubjectSelector
        selector.left;
        // $ExpectType SubjectSelector
        selector.right;
        break;

    case 'attribute':
        // $ExpectType string
        selector.name;
}
```

Also, upon inspecting the source code, it looks like `esquery.parse` can never return `undefined`, it only returns a valid selector, or it throws.

### DIFF
--- a/types/esquery/esquery-tests.ts
+++ b/types/esquery/esquery-tests.ts
@@ -10,7 +10,7 @@ function f (n) {
 
 const s = 'FunctionDeclaration !VariableDeclaration > VariableDeclarator[init.value > 3]';
 
-// $ExpectType Selector | undefined
+// $ExpectType Selector
 const selector = esquery.parse(s);
 
 // $ExpectType Node[]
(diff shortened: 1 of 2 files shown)


cherryblossom000 (approved):

cherryblossom000:
That's a much better way of doing it. Thanks!

PR #41206, opened 2019-12-23
Add types for remove-files-webpack-plugin

srgykuz:
Hello.

Here is a types for remove-files-webpack-plugin.
GitHub – https://github.com/Amaimersion/remove-files-webpack-plugin
NPM – https://www.npmjs.com/package/remove-files-webpack-plugin

# Template

If adding a new definition:
  - **Note**: current JSDoc types in the plugin file cannot be treated as a providing of it's own types. I created it only for local usage and not all editors or IDE's can treat this as a truly type.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,5 +31,9 @@
             "_comment": "This will remove husky from when we started migrating to use prettier",
             "pre-commit": "npm uninstall husky"
         }
+    },
+    "dependencies": {
+        "@types/source-map": "^0.5.7",
+        "@types/webpack": "^4.41.0"
     }
 }
--- a/package.json
+++ b/package.json
@@ -31,9 +31,5 @@
             "_comment": "This will remove husky from when we started migrating to use prettier",
             "pre-commit": "npm uninstall husky"
         }
-    },
-    "dependencies": {
-        "@types/source-map": "^0.5.7",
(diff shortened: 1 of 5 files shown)


peterblazejewicz:
@Amaimersion 
First thanks for the contribution! Just quick question - as you are the owner of the project, why here, instead of shipping directly with the plugin? A lot of the projects already do this, this makes devx easier I think:
http://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#including-declarations-in-your-npm-package

Amaimersion:
@peterblazejewicz 
Hello! 

From [your link](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html):
```
1) bundling with your npm package, or
2) publishing to the @types organization on npm.

If your package is written in TypeScript then the first approach is favored. Use the --declaration flag to generate declaration files. This way, your declarations and JavaScript will always be in sync.
If your package is not written in TypeScript then the second is the preferred approach.
```

My package not written in TypeScript, so, i picked up № 2.

peterblazejewicz:
I'll post comment on related repo, let's keep here clear.

Amaimersion:
Closed because of [this](https://github.com/Amaimersion/remove-files-webpack-plugin/issues/13#issuecomment-568871242).

PR #41207, opened 2019-12-23
[react-map-gl]Add type definitions of source and layer components

williamlabdev:
Please fill in this template.

If adding a new definition:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://uber.github.io/react-map-gl/#/Documentation/api-reference/source>>

If removing a declaration:


types/react-map-gl/index.d.ts
```diff
@@ -423,3 +424,47 @@ export interface SVGOverlayProps extends BaseControlProps {
 }
 
 export class SVGOverlay extends BaseControl<SVGOverlayProps, Element> {}
+
+export interface SourceProps {
```
fnberta: Where did you get all these props from? In the documentation and the source code, I only see `id`, `type`, and `children`. Or am I missing something?

fnberta (changes requested):
Thank you for this PR. I have one question regarding the Props of the `Source` component, see inline.

types/react-map-gl/index.d.ts
chiuhow: In the source code of react-map-gl ,sourceOptions is provdied implictly and extra properties will be added through the loop.
https://github.com/uber/react-map-gl/blob/c71b0d4b9da65bdeaa4a077d85b4658654b68100/src/components/source.js#L96
I add these extra properties and set them as optional and people could pass extra properties which are defined by mapbox-gl with type checking.

fnberta (approved):
Sorry for the delay, approving!

types/react-map-gl/index.d.ts
fnberta: Ok, cool

PR #41210, opened 2019-12-23
Add pica library type definition files

hmtylmz:
Please fill in this template.

If adding a new definition:

I write a type definition file and add pull request library itself. But author said:
https://github.com/nodeca/pica/pull/171

### DIFF
--- a/types/pica/index.d.ts
+++ b/types/pica/index.d.ts
@@ -70,7 +70,7 @@ interface PicaResizeOptions {
 
 interface PicaResizeBufferOptions {
     // Uint8Array with source data.
-    src?: number[];
+    src: number[];
     // src image width.
     width: number;
     // src image height.
--- a/types/pica/index.d.ts
+++ b/types/pica/index.d.ts
@@ -1,0 +1,85 @@
+// Type definitions for pica 5.1
+// Project: https://github.com/nodeca/pica
+// Definitions by: Hamit YILMAZ <https://github.com/hmtylmz>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare class Pica {
+    constructor(config?: PicaOptions);
+    /**
+     * Resize image from one canvas (or image) to another. Sizes are taken from source and destination objects.
+     * Result is Promise, resolved with to on success.
+     * (!) If you need to process multiple images, do it sequentially to optimize CPU & memory use. Pica already knows how to use multiple cores (if browser allows).
+     */
+    resize(
+        from: HTMLCanvasElement | HTMLImageElement | File | Blob,
+        to: HTMLCanvasElement,
+        options?: PicaResizeOptions
+    ): Promise<HTMLCanvasElement>;
+
+    /**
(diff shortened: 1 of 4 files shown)


types/pica/pica-tests.ts
```diff
@@ -0,0 +1,34 @@
+let picaOptions: PicaOptions = { features: ['js', 'wasm', 'ww', 'cib'] };
+let resizer: Pica = new Pica(picaOptions);
```
peterblazejewicz: Just suggestions, as the options are optional, could we also include the parameterless constructor test as well? Exactly as documented on the package readme (all ctros are parameterless there):
```js
const pica = Pica();
pica.resize(img, canvas).then(...);
```
Thanks!

hmtylmz: Added tests with parameterless constructor

types/pica/index.d.ts
```diff
@@ -0,0 +1,95 @@
...
+    unsharpRadius?: number;
+    // 0..255. Default = 0. Threshold for applying unsharp mask.
+    unsharpThreshold?: number;
+    // Promise instance. If defined, current operation will be terminated on rejection.
+    cancelToken?: string;
+}
+
+interface PicaResizeBufferOptions {
+    // Uint8Array with source data.
+    src?: number[];
```
peterblazejewicz: I believe this is required:
https://github.com/nodeca/pica/blob/master/README.md#resizebufferoptions---promise
It passed bits to 3rd party library:
https://github.com/nodeca/multimath#readme
Can you double-check this?
Thanks!

hmtylmz: Sorry. I fixed this.

peterblazejewicz (approved):
LGTM
Thanks for contribution!

andrewbranch (approved):

types/pica/index.d.ts
```diff
@@ -0,0 +1,95 @@
+// Type definitions for pica 5.1
+// Project: https://github.com/nodeca/pica
+// Definitions by: Hamit YILMAZ <https://github.com/hmtylmz>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare class Pica {
```
wkritzinger: Shouldn't this be exported? I get `File '/node_modules/@types/pica/index.d.ts' is not a module.  TS2306` if nothing is exported from this file.

andrewbranch: Pica is a UMD module (meaning it works as a global, an AMD module, or a CommonJS module). So while it’s not exactly _wrong_ to simply declare it as a global, it’s incomplete. There should be an `export as namespace Pica` to indicate it’s available as UMD.

namelesswc:
I'm having trouble using this.
As far as I understood, I don't need to import this like other packages `import * as Pica from "pica";`.
But directly using it like this `Pica.resize(image, canvas);` also doesn't work. I get `Property 'resize' does not exist on type 'typeof Pica'.`

peterblazejewicz:
folks, I've made a RP to mitigate discussed problems:
#44858
Thanks for the patience!

PR #41222, opened 2019-12-23
Add types for cypress-image-snapshot

Keysox:

### DIFF
--- a/types/cypress-image-snapshot/index.d.ts
+++ b/types/cypress-image-snapshot/index.d.ts
@@ -1,0 +1,14 @@
+// Type definitions for cypress-image-snapshot 3.1
+// Project: https://github.com/palmerhq/cypress-image-snapshot
+// Definitions by: Alex Kessock <https://github.com/Keysox>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare global {
+    namespace Cypress {
+        interface Chainable {
+            matchImageSnapshot: () => void;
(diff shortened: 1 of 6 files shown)


types/cypress-image-snapshot/index.d.ts
```diff
@@ -0,0 +1,14 @@
+// Type definitions for cypress-image-snapshot 3.1
+// Project: https://github.com/palmerhq/cypress-image-snapshot
+// Definitions by: Alex Kessock <https://github.com/Keysox>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare global {
+    namespace Cypress {
+        interface Chainable {
+            matchImageSnapshot: () => void;
```
peterblazejewicz: Just wondering, should that definition match command definition? So users could use:
```ts
cy.visit('/login');
// snapshot name will be the test title
cy.matchImageSnapshot();
// snapshot name will be the name passed in
cy.matchImageSnapshot('login');
// options object passed in
cy.matchImageSnapshot({});
// match element snapshot
cy.get('#login').matchImageSnapshot();
```
as per plugin docs?

Keysox: @peterblazejewicz -- good catch! I haven't passed any parameters in before but just tested and they do work so I went ahead and added those.

peterblazejewicz (approved):
LGTM
Thanks for contribution!

PR #41237, opened 2019-12-24
[testing-library__dom] Fix async bound queries

### LINKED ISSUE #366: Typings for `findBy*` do not include `waitForElement` options
- `react-testing-library` version: 7.0.0
- `react` version: 16.8.4
- `node` version: 8.12.0
- `npm` (or `yarn`) version: 6.9.0

### Relevant code or config:

```js
const { findByLabelText } = render(
  <div />,
);

expect(await findByLabelText('anything', undefined, { timeout: 100 }))
  .toHaveAttribute(
    'aria-checked',
    'mixed',
  );

```

### What you did:

[Per the docs](https://testing-library.com/docs/dom-testing-library/api-queries#findby), the *findBy* queries accept the `waitForElement` options as the last argument. The TypeScript configuration only reflects a possible 1-2 arguments instead of the expected 1-3.

I suspect this has to do with how `RenderResult` (https://github.com/testing-library/react-testing-library/blob/960451b00196352fc3085a3d1bb233b5b85f454b/typings/index.d.ts#L19-L34) is union'ed (`& {[P in keyof Q]: BoundFunction<Q[P]>`) with `queries` via `BoundFunction` (https://github.com/testing-library/dom-testing-library/blob/4fb00936cbf5b0d715f5b54d16576dc3bfe20460/typings/get-queries-for-element.d.ts#L4-L14).

I'd love to help; I'm hazy on how exactly the following works; which is where I believe the problem lies:

https://github.com/testing-library/dom-testing-library/blob/4fb00936cbf5b0d715f5b54d16576dc3bfe20460/typings/get-queries-for-element.d.ts#L4-L14

wezleytsai:
Fixes https://github.com/testing-library/react-testing-library/issues/366

Many queries accept arguments of type `container: HTMLElement, id: Matcher, options?: MatcherOptions, waitForElementOptions?: WaitForElementOptions`. The existing `BoundFunction` definition does not take this into account and matches these to `T extends (a1: any, text: infer P, options: infer Q) => infer R`, resulting in the bound query to be of type `(text: P, options?: Q) => R`, which makes it impossible to supply the `waitForElementOptions` argument in a type-safe manner. This fixes `BoundFunction` to account for the async function signature so that the bound query will have type `(text: P, options?: Q, waitForElementOptions?: W) => R`.

Please fill in this template.

### DIFF
--- a/types/testing-library__dom/get-queries-for-element.d.ts
+++ b/types/testing-library__dom/get-queries-for-element.d.ts
@@ -8,6 +8,8 @@ export type BoundFunction<T> = T extends (
     options: infer Q,
 ) => infer R
     ? (text: P, options?: Q) => R
+    : T extends (a1: any, text: infer P, options: infer Q, waitForElementOptions: infer W) => infer R
+    ? (text: P, options?: Q, waitForElementOptions?: W) => R
     : T extends (a1: any, text: infer P, options: infer Q) => infer R
     ? (text: P, options?: Q) => R
     : never;
--- a/types/testing-library__dom/index.d.ts
+++ b/types/testing-library__dom/index.d.ts
@@ -6,6 +6,7 @@
 //                 Weyert de Boer <https://github.com/weyert>
 //                 Ronald Rey <https://github.com/reyronald>
 //                 Justin Hall <https://github.com/wKovacs64>
+//                 Wesley Tsai <https://github.com/wezleytsai>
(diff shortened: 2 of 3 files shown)


eps1lon (changes requested):
Thanks for the fix.

Could you add a test to make make it obvious what's being changed here?

eps1lon (approved):

Issue #36306, opened 2019-06-20
[@types/mathjs] a suggestion of type


aoi-umi:
when i use round,
i need to write like this:
```ts
let a = 1.111;
a = mathjs.round(a, 2) as number;
```
but change like this:
```ts
round<T extends number | BigNumber | Fraction | Complex | MathArray | Matrix>(
            x: T,
            n?: number | BigNumber | MathArray
        ): T;
```
i no longer need to change the type， just
```
let a = 1.111;
a = mathjs.round(a, 2);//return type same as typeof a;
```
the same as other type

aoi-umi:
@siavol @andnp @bradbesserman

andnp:
Looks good to me. Feel free to make a PR and tag us there!

PR #41242, opened 2019-12-25
[mathjs] generic

aoi-umi:
[issus](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/36306)

@siavol @andnp @bradbesserman


RyanCavanaugh (changes requested):
This isn't a great change because TypeScript is going to think that `log(10)` is `10`. You can use a conditional type to correctly relate the input type to the output type while not preserving literal types

Please add tests!

RyanCavanaugh (approved):

uniqueiniquity:

PR #41247, opened 2019-12-25
Fix aws-param-store path parameter type 

nathanoertel:
According to the [documentation](https://docs.aws.amazon.com/cli/latest/reference/ssm/get-parameters-by-path.html#options) the --path parameter is a string not a parameter name array.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://docs.aws.amazon.com/cli/latest/reference/ssm/get-parameters-by-path.html#options

### DIFF
--- a/types/aws-param-store/aws-param-store-tests.ts
+++ b/types/aws-param-store/aws-param-store-tests.ts
@@ -12,7 +12,6 @@ import {
 
 declare let bool: boolean;
 declare let query: ParameterQuery;
-declare let path: string;
 declare let psName: SSM.Types.PSParameterName;
 declare let psNames: SSM.Types.ParameterNameList;
 declare let options: SSM.Types.ClientConfiguration;
@@ -27,7 +26,7 @@ declare let promiseAllParamResults: Promise<typeof allParamResults>;
 
 query = parameterQuery();
 
-query.path(path);
+query.path(psName);
 query.named(psName);
 query.named(psNames);
 query.decryption(bool);
@@ -66,8 +65,8 @@ promiseParamResult = getParameter(psName, options);
 promiseParamsResult = getParameters(psNames);
 promiseParamsResult = getParameters(psNames, options);
 
-promiseParamsByPathResult = getParametersByPath(path);
-promiseParamsByPathResult = getParametersByPath(path, options);
+promiseParamsByPathResult = getParametersByPath(psName);
+promiseParamsByPathResult = getParametersByPath(psName, options);
 
 paramResult = getParameterSync(psName);
 paramResult = getParameterSync(psName, options);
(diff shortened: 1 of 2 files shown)


types/aws-param-store/aws-param-store-tests.ts
```diff
@@ -65,14 +66,14 @@ promiseParamResult = getParameter(psName, options);
 promiseParamsResult = getParameters(psNames);
 promiseParamsResult = getParameters(psNames, options);
 
-promiseParamsByPathResult = getParametersByPath(psNames);
-promiseParamsByPathResult = getParametersByPath(psNames, options);
+promiseParamsByPathResult = getParametersByPath(path);
```
jasonthomasgray: Yeah this shouldn't have been psNames. I was trying to use the types that are defined for the aws SDK, so looking back at those (https://github.com/aws/aws-sdk-js/blob/a7af3ec024028681a8aa676decf0929d962b9f9a/clients/ssm.d.ts#L4580) this should be `PSParameterName` which is aliased here as `psName`

types/aws-param-store/index.d.ts
```diff
@@ -27,17 +27,17 @@ export function getParametersSync(
...
 
 export function getParametersByPathSync(
-    path: SSM.Types.ParameterNameList,
+    path: string,
     options?: SSM.Types.ClientConfiguration
 ): SSM.Types.ParameterList;
 
 export interface ParameterQuery {
-    path(path: SSM.Types.PSParameterName): ParameterQuery;
+    path(path: string): ParameterQuery;
```
jasonthomasgray: I think this can stay as `SSM.Types.PSParameterName`

types/aws-param-store/index.d.ts
```diff
@@ -27,17 +27,17 @@ export function getParametersSync(
...
 
 export function getParametersByPath(
-    path: SSM.Types.ParameterNameList,
+    path: string,
     options?: SSM.Types.ClientConfiguration
 ): Promise<SSM.Types.ParameterList>;
 
 export function getParametersByPathSync(
-    path: SSM.Types.ParameterNameList,
+    path: string,
```
jasonthomasgray: This and the async version above can use `SSM.Types.PSParameterName` which ends up as string anyway but using the SDK types seems better to me.

types/aws-param-store/aws-param-store-tests.ts
```diff
@@ -12,6 +12,7 @@ import {
 
 declare let bool: boolean;
 declare let query: ParameterQuery;
+declare let path: string;
```
jasonthomasgray: I don't think path is needed as `SSM.Types.PSParameterName` is string in the end.

jasonthomasgray (changes requested):
This is a good catch I totally missed that when combing the source and the SDK types.

If we are able to use the SDK types as I've commented then I think that's good to go.

types/aws-param-store/aws-param-store-tests.ts
nathanoertel: I went with string because PSParameterName infers that you are passing a parameter name which _path_ is not, it is any string that could match any number of parameter names.

nathanoertel:
@jasonthomasgray I've updated the types.

jasonthomasgray (approved):
Looks good to me. 😊

andrewbranch (approved):

PR #41253, opened 2019-12-26
Add @types/onecolor

ghost:
Add @types/onecolor for [one-color](https://github.com/One-com/one-color)

Please fill in this template.


peterblazejewicz (changes requested):
@Azliya 
You will need to restructure the PR, the `onecolor` module files should reside in their own directory
named after module name,
see example:
https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/base64-js
Thanks!

types/onecolor/index.d.ts
```diff
@@ -0,0 +1,34 @@
+// Type definitions for onecolor 3.1
+// Project: onecolor
```
peterblazejewicz: As there are multiple projects, can you refere via link which one these types refer to?
sample:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/base64-js/index.d.ts#L2

types/onecolor/onecolor-tests.ts
```diff
@@ -0,0 +1,58 @@
+// This contains sample code which tests the typings. This code does not run, but it is type-checked.
+
+import onecolor = require('onecolor');
+
+// Contrustor
```
peterblazejewicz: Obvious typo

Azliya: Thanks for your review @peterblazejewicz, and I've amended this typo in this [commit](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41253/commits/271db78639557c73559e4686ec4fcf29fb892ea8)

types/onecolor/index.d.ts
Azliya: Thanks @peterblazejewicz, ref link added in this [commit](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41253/commits/271db78639557c73559e4686ec4fcf29fb892ea8)

types/onecolor/index.d.ts
```diff
@@ -0,0 +1,34 @@
+// Type definitions for onecolor 3.1
+// Project: https://github.com/One-com/one-color
+// Definitions by: Azliya <https://github.com/Azliya>
+// Definitions: https://github.com/DefinitelyTyped/onecolor
+
+export = onecolor;
+declare function onecolor(clr: onecolor.ColorDefType): onecolor.Color;
+declare namespace onecolor {
```
sheetalkamat: This can be used as global as well as module.. 

sheetalkamat (changes requested):

PR #41255, opened 2019-12-26
[bn.js] Fix module import without 'allowSyntheticDefaultImports'

zone117x:
Provide the ability to import this lib using `import * as bn from 'bn.js'` syntax, with default `tsconfig.json` settings. 

Previously, the bn.js lib could only be imported when using `allowSyntheticDefaultImports`, `require('bn.js')`, or ` // @ts-ignore \n import * as bn from 'bn.js'`. 

This causes problems for projects that wish to avoid the `esModuleInterop` flag:
* Projects that aim to have maximally readable js output, and js that closely mirrors the ts code -- therefore avoid the 10+ lines of helper code generated by esModuleInterop.
* Projects that aim to prevent the side-effects that this helper code generates -- when the flag is enabled, it begins to propagate throughout the codebase and used on module imports which do not actually need it. 
* Projects that wish to avoid causing issues with downstream dependencies, e.g.:
  * If the `// @ts-ignore` is used, it often requires downstreams dependencies to enable `skipLibCheck`.
  * If the `require('bn.js')` syntax is used, it often requires disabling their eslint/tslint rules for the line, and/or require downstream dependencies to do the same. 
* Many projects import and use `bn.js` package bare, without types in order to avoid these issues. 

### DIFF
--- a/types/bn.js/bn.js-tests.ts
+++ b/types/bn.js/bn.js-tests.ts
@@ -1,23 +1,33 @@
 // test that it works as a require
-import BN = require('bn.js');
+import BN_require = require('bn.js');
+// test that it works as module import
+import * as BN_esm from 'bn.js';
+import { Endianness, IPrimeName } from 'bn.js';
 
-let bn = new BN(42);
-bn = bn.add(bn);
-bn.isZero();
-bn.byteLength;
+function runTests(BN: typeof BN_esm) {
+    let bn = new BN(42);
+    bn = bn.add(bn);
+    bn.isZero();
+    bn.byteLength;
 
-bn.toArrayLike(Buffer, 'le', 2);
-const test = new BN(1, 'le');
+    const endian: Endianness = 'le';
+    bn.toArrayLike(Buffer, endian, 2);
+    const test = new BN(1, endian);
 
-const ctx = BN.red('p224');
-ctx.prime.name;
+    const primeName: IPrimeName = 'p224';
+    const ctx = BN.red(primeName);
+    ctx.prime.name;
 
-const red = bn.toRed(ctx);
-const newRed = red.redAdd(new BN(1));
-newRed.cmp(bn);
-newRed.fromRed();
+    const red = bn.toRed(ctx);
+    const newRed = red.redAdd(new BN(1).toRed(ctx));
+    newRed.cmp(bn);
+    newRed.fromRed();
 
-const expected = new BN(0x4020);
-const actualArray = new BN([0x40, 0x20]);
-const actualUint8Array =  new BN(new Uint8Array([0x40, 0x20]));
-const actualString = new BN('0x4020');
+    const expected = new BN(0x4020);
+    const actualArray = new BN([0x40, 0x20]);
+    const actualUint8Array =  new BN(new Uint8Array([0x40, 0x20]));
+    const actualString = new BN('0x4020');
+}
+
+runTests(BN_esm);
+runTests(BN_require);
--- a/types/bn.js/index.d.ts
+++ b/types/bn.js/index.d.ts
@@ -7,59 +7,59 @@
 
 /// <reference types="node"/>
 
-type Endianness = 'le' | 'be';
-type IPrimeName = 'k256' | 'p224' | 'p192' | 'p25519';
-
-interface MPrime {
-    name: string;
-    p: BN;
-    n: number;
-    k: BN;
-}
-
-interface ReductionContext {
-    m: number;
-    prime: MPrime;
-    [key: string]: any;
-}
-
 declare namespace BN {
-    /**
-     * @description  create a reduction context
-     */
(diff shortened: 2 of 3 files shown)


Gilthoniel:
According to the FAQ, this practice is discourage. What's wrong with using `esModuleInterop` ?

zone117x:
It's unfortunate that the FAQ discourages this. The [cited stackoverflow post](https://stackoverflow.com/questions/39415661/what-does-resolves-to-a-non-module-entity-and-cannot-be-imported-using-this) from 2013 seems inaccurate today. 

> ... they might work today in your particular downleveling module bundler (e.g. rollup), but this is illusory. There aren't any ES6 module implementations in the wild yet, but that won't be true forever.
> Picture your future self, trying to run on a neato native ES6 module implementation and finding that you've set yourself up for major failure by trying to use ES6 syntax to do something that ES6 explicitly doesn't do.

This must be referring to an older version of Typescript? 
* The compiled `js` output is the same -- no bundler downleveling required. 
* The `esModuleInterop` and `allowSyntheticDefaultImports` options are not enabled by default. 
* Common eslint/tslint configurations disallow `require(...)` statements. 
* The `esModuleInterop` flag generates 10+ lines of side-effect-prone helper code (the `__importStar` and `__importDefault` functions), and the output `js` is less readable. Many projects wish to avoid this. 

With this patch, the following typical code and config generate output as expected:

`tsconfig.json`:
```jsonc
{
    "compilerOptions": {
        "module": "commonjs",
        "target": "es6",
        "strict": true
    }
}
```

`test.ts` source:
```ts
import * as BN from 'bn.js';
let bn = new BN(42);
```

`test.js` output:
```js
// with `"target": "es6"` (same for all higher targets as well, like `es2020`)
const BN = require("bn.js");
let bn = new BN(42);

// or with `"target": "es5" (same for the lower targets)`
var BN = require("bn.js");
var bn = new BN(42);
```



----

This patch still allows projects to use `esModuleInterop` or `require(...)`:

`tsconfig.json`:
```jsonc
{
    "compilerOptions": {
        "module": "commonjs",
        "target": "es6",
        "strict": true,
        "esModuleInterop": true
    }
}
```

`test.ts` source:
```ts
import BN_require = require('bn.js');
import BN from 'bn.js';

let bn = new BN(42);
let bn_require = new BN_require(42);
```

`test.js` output:
```js
var __importDefault = (this && this.__importDefault) || function (mod) {
    return (mod && mod.__esModule) ? mod : { "default": mod };
};
Object.defineProperty(exports, "__esModule", { value: true });
const BN_require = require("bn.js");
const bn_js_1 = __importDefault(require("bn.js"));
let bn = new bn_js_1.default(42);
let bn_require = new BN_require(42);
```

zone117x:
The `@types/node` definitions provide precedent for this:

The tsconfig.json used is essentially default -- no `esModuleInterop`:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/926b2c951fdfe50eaa178b3bb75c726f4c905cfb/types/node/v12/tsconfig.json#L34-L40

The type definitions import each other using `import * as x from 'x`` syntax, and always provide the capability to `import *`:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/926b2c951fdfe50eaa178b3bb75c726f4c905cfb/types/node/v12/crypto.d.ts#L1-L3

---

This `@types/bn.js` package is one of the only dependencies in several large projects that do not provide this capability. Because of this, the `bn.js` package is often used bare, without types, to avoid potential issues with `esModuleInterop`. `bitcoinjs-lib` and its family of dependencies is one example.

Gilthoniel (approved):

zone117x:
@Gilthoniel thanks for the review. FYI -- I just pushed an update which includes:
* New unit tests for both import methods. 
* Exports the Endianness and PrimeNames types (useful for downstream dependencies that want to passthrough the allowed types without having to redefine them). 
* Restored the static class methods to avoid potential breakage for consumers who do something like pass the static BN object properties around.

types/bn.js/bn.js-tests.ts
```diff
@@ -1,23 +1,33 @@
...
+    const primeName: IPrimeName = 'p224';
+    const ctx = BN.red(primeName);
+    ctx.prime.name;
 
-const red = bn.toRed(ctx);
-const newRed = red.redAdd(new BN(1));
-newRed.cmp(bn);
-newRed.fromRed();
+    const red = bn.toRed(ctx);
+    const newRed = red.redAdd(new BN(1).toRed(ctx));
```
zone117x: Reviewer note:
This throws an error since at least v4.11.0 (the version indicated in the type def).
See: https://github.com/indutny/bn.js/blob/e670ab2360a6f37994014d85a07849c81817c1b0/lib/bn.js#L2849-L2852

Added the `.toRed` to fix the error.

PR #41258, opened 2019-12-26
Add @types/rebass__forms

zinozzino:
Add @types/rebass__forms for [@rebass/forms](https://github.com/rebassjs/rebass/tree/master/packages/forms)

**This repo depends on [@types/rebass](https://www.npmjs.com/package/@types/rebass), [@types/styled-system](https://www.npmjs.com/package/@types/styled-system), [@types/styled-components](https://www.npmjs.com/package/@types/styled-components)**

### DIFF
--- a/types/rebass__forms/index.d.ts
+++ b/types/rebass__forms/index.d.ts
@@ -8,7 +8,9 @@ import * as React from 'react';
 import * as Rebass from 'rebass';
 import * as StyledSystem from 'styled-system';
 
-type Omit<T, K extends string | number | symbol> = { [P in Exclude<keyof T, K>]: T[P] };
+export {};
+
+type Omit<T, K> = Pick<T, Exclude<keyof T, K>>;
 
 interface BoxKnownProps
     extends Rebass.BaseProps,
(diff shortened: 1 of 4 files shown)


types/rebass__forms/index.d.ts
```diff
@@ -0,0 +1,54 @@
+// Type definitions for @rebass/forms 4.0
+// Project: https://github.com/rebassjs/rebass#readme
+// Definitions by: zinozzino <https://github.com/zinozzino>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.1
```
peterblazejewicz: You will have to use TS 3.5 when using the `Omit` 
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html#the-omit-helper-type
Is that OK? If so, just update the header and this would remove most of the lint errors

types/rebass__forms/index.d.ts
```diff
@@ -0,0 +1,54 @@
...
+
+import { ResponsiveStyleValue, SystemStyleObject } from '@styled-system/css';
```
peterblazejewicz: This one is not used, can be removed I think

types/rebass__forms/rebass__forms-tests.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+
+import { Label, Input, Select, Textarea, Radio, Checkbox } from '@rebass/forms';
```
peterblazejewicz: Can the `TextArea` be added below? Otherwise it is not covered by any test

zinozzino: Thank you for your response! I just added `Textarea` component below.

types/rebass__forms/index.d.ts
zinozzino: Thank you! I just remove those imports.

types/rebass__forms/index.d.ts
zinozzino: i think adding the custom omit type for compatibility is the better way. so I added the one.

andrewbranch (approved):

PR #41270, opened 2019-12-27
Improved type definitions for react-widgets

jankalfus:
Added exports to all interfaces and moved all type definitions into a single file. This way, the react-widgets components can be easily extended and the types used when extending them.

For example:
obrazek


r3nya (changes requested):
Hey! I like your idea.

But don't forget about some case – If user import components like `import DropdownList from 'react-widgets/lib/DropdownList';`, it won't support this way, right?

It would be awesome if you can add tests for `import DropdownList from 'react-widgets/lib/DropdownList';` or `import Combobox from 'react-widgets/lib/Combobox'` cases.

jankalfus:
@r3nya Yeah, that won't work. Do you have a suggestion how I can modify the code, so that we can support both use cases? Like so:
```
import { DateTimePicker, DateTimePickerProps } from "react-widgets"  // UC 1
import DateTimePicker, { DateTimePickerProps } from "react-widgets/lib/DateTimePicker"  // UC 2
```
I'm guessing I would have to put back all the definitions into separate files as they were before (that's easy :)), but I'm not sure how to reexport all interfaces into the index.d.ts file?

I mean, I could also live with having to import the props from the lib file :) It's just that the following is much nicer:
```
import { DateTimePicker, DateTimePickerProps } from "react-widgets"
```

r3nya:
@jankalfus Oh … I think you should 
- export all interfaces from `https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-widgets/lib/*.d.ts`
- and import it in `https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-widgets/index.d.ts` and export these interfaces here. 🤷‍♂ 

I guess this option will cover all cases.

Btw, if you have free time please add tests for importing 
```js
import DateTimePicker, { DateTimePickerProps } from "react-widgets/lib/DateTimePicker"
``` 
in `https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-widgets/react-widgets-tests.tsx`. (if you don't have free time, it's ok as well). :)

Thanks in advance! ❤️

r3nya:
@jankalfus 

> but I'm not sure how to reexport all interfaces into the index.d.ts file?

Hm, good question lol. I'm not sure tbh.

r3nya:
@jankalfus btw, please follow this example, I hope it would be useful 

- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-bootstrap/lib/index.d.ts
- https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-bootstrap/index.d.ts#L47

jankalfus:
@r3nya Thanks for the ideas and tips!

I've added the tests for the default exports.

I managed to change the exports so that everything remains as it was before + now the developer has the ability to import the interfaces and types. So it can be used the following way now:

```
import { Calendar } from "react-widgets";
import * as CalendarDefault from "react-widgets/lib/Calendar";
import { CalendarMessages, CalendarProps, CalendarView } from "react-widgets/lib/Calendar";

// or with esModuleInterop enabled:
import Calendar, { CalendarMessages, CalendarProps, CalendarView } from "react-widgets/lib/Calendar";
```

I'm not sure how to do both default and separate exports with commonjs, seems to be way too complicated with too small benefit. I'm pretty happy with the result the way it is.

r3nya (approved):
Great job! 👍

PR #41277, opened 2019-12-27
fix(truncate-middle): null union type for `str`

peterblazejewicz:
This fixes empty string return when param equals `null`:
https://git.io/JeNkz

Thanks!

/cc @garyking

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/kahwee/truncate-middle/blob/master/index.js#L13-L15


RyanCavanaugh:
@kahwee is this an intended use of the library?

PR #41298, opened 2019-12-29
fix: rdf-js parsers actually return a quad stream

tpluscode:
The base `EventEmitter` is not enough to write the code below:

```typescript
import stringToStream from 'string-to-stream'
import rdf from 'rdf-ext'
import { Dataset } from 'rdf-js'

export function parse(quads: string): Promise<Dataset> {
  const stream = stringToStream(quads) as any

  return rdf.dataset().import(parser.import(stream))
}
```

This code works fine and for typings to be correct the `parser.stream()` should return a stream

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: _see above_

### DIFF
--- a/types/rdfjs__parser-jsonld/rdfjs__parser-jsonld-tests.ts
+++ b/types/rdfjs__parser-jsonld/rdfjs__parser-jsonld-tests.ts
@@ -1,6 +1,6 @@
 import Parser = require('@rdfjs/parser-jsonld');
 import { Context } from 'jsonld/jsonld-spec';
-import { DataFactory, Sink, Stream } from 'rdf-js';
+import { DataFactory, Sink, Stream, BaseQuad } from 'rdf-js';
 
 const baseIRI = '';
 const context: Context = {} as any;
@@ -16,3 +16,12 @@ const sink: Sink = parser1;
 const eventEmitter1: Stream = parser1.import(stream);
 const eventEmitter2: Stream = parser1.import(stream, {});
 const eventEmitter3: Stream = parser1.import(stream, { baseIRI, context, factory });
+
+interface SpecializedQuad extends BaseQuad {
+    foo: string;
+}
+const typedStream: Stream<SpecializedQuad> = <any> {};
+const typedParser: Parser<SpecializedQuad> = <any> {};
+const typedImported: Stream<SpecializedQuad> = typedParser.import(typedStream);
+const typedImported1: Stream<SpecializedQuad> = typedParser.import(typedStream, {});
+const typedImported2: Stream<SpecializedQuad> = typedParser.import(typedStream, { baseIRI, context, factory });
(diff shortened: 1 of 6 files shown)


types/rdfjs__parser-jsonld/rdfjs__parser-jsonld-tests.ts
```diff
@@ -14,6 +13,6 @@ const parser3 = new Parser({ baseIRI, context, factory });
 
 const sink: Sink = parser1;
 
-const eventEmitter1: EventEmitter = parser1.import(stream);
-const eventEmitter2: EventEmitter = parser1.import(stream, {});
-const eventEmitter3: EventEmitter = parser1.import(stream, { baseIRI, context, factory });
+const eventEmitter1: Stream = parser1.import(stream);
+const eventEmitter2: Stream = parser1.import(stream, {});
+const eventEmitter3: Stream = parser1.import(stream, { baseIRI, context, factory });
```
thewilkybarkid: Should also test with a different quad implementation.

tpluscode: Something like this?
Have to go all the way back to a custom quad. Using `BaseQuad` is cheating imo ;)

```ts
interface SpecializedQuad extends BaseQuad {}
const typedStream: Stream<SpecializedQuad> = <any> {};
const typedParser: Parser<SpecializedQuad> = <any> {};
const typedImported: Stream<SpecializedQuad> = typedParser.import(typedStream);
const typedImported1: Stream<SpecializedQuad> = typedParser.import(typedStream, {});
const typedImported2: Stream<SpecializedQuad> = typedParser.import(typedStream, { baseIRI, context, factory });
```

thewilkybarkid: Yep!

types/rdfjs__serializer-jsonld-ext/index.d.ts
```diff
@@ -19,10 +19,10 @@ declare namespace Serializer {
     }
 }
 
-declare class Serializer implements Sink {
+declare class Serializer<Q extends BaseQuad = Quad> implements Sink<Q> {
     constructor(options?: Serializer.SerializerOptions);
 
-    import(stream: Stream, options?: Serializer.SerializerOptions): EventEmitter;
+    import(stream: Stream<Q>, options?: Serializer.SerializerOptions): EventEmitter;
```
thewilkybarkid: All extensions of `@rdfjs/sink` return a `Stream`?

tpluscode: I don't understand

thewilkybarkid: Sorry, mixed terminologies. Meant a `Readable` stream. But have just looked at the code and see `read` is a dummy function so no need to change it here.

thewilkybarkid (approved):

PR #41300, opened 2019-12-29
Update webmidi typings

calebegg:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.w3.org/TR/webmidi/


types/webmidi/index.d.ts
```diff
@@ -4,193 +4,234 @@
...
+        /**
+         * The system name of the port.
+         */
+        name?: string;
+
+        /**
+         * A descriptor property to distinguish whether the port is an input or an output
+         * port.
+         */
+        type: MIDIPortType;
```
lostfictions: Same question as above -- can't this just be removed? (Might be nice to move the docstring to both implementors though.)

types/webmidi/index.d.ts
```diff
@@ -4,193 +4,234 @@
...
+        ): void;
+
+        /**
+         * This attribute informs the user whether system exclusive support is enabled on
+         * this MIDIAccess.
+         */
+        sysexEnabled: boolean;
+    }
+
+    type MIDIPortType = 'input' | 'output';
```
lostfictions: Is this still necessary since it's now directly defined on the implementors of `MIDIPort`?

types/webmidi/index.d.ts
```diff
@@ -4,193 +4,234 @@
...
+        addEventListener(
+            type: 'statechange',
+            listener: (this: this, e: MIDIConnectionEvent) => any,
+            options?: boolean | AddEventListenerOptions,
+        ): void;
+        addEventListener(
+            type: string,
+            listener: EventListenerOrEventListenerObject,
+            options?: boolean | AddEventListenerOptions,
+        ): void;
```
lostfictions: Is it necessary to define the non-discriminated overload here? Isn't it already part of `EventTarget`?

types/webmidi/index.d.ts
```diff
@@ -4,193 +4,234 @@
...
+    }
```
lostfictions: Same Q as above re: necessity of redefining these (I genuinely don't know -- if TS requires all of them to be redeclared every time this makes sense).

types/webmidi/index.d.ts
```diff
@@ -4,193 +4,234 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
```
lostfictions: btw, not sure at what point these typings became considered applicable to https://github.com/djipco/webmidi (see line 2 folded above, which I can't directly comment on) but I think that's not the case (see discussion here: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/39571). Mind removing it?

lostfictions (changes requested):
Thanks! A little unfortunate both these files got almost entirely churned on whitespace and quotes -- was that your choice or is it a lint dts enforces?

Also, I haven't worked on WebMIDI stuff in several years and I'm not really suited to review and maintain these typings anymore. Could you replace my name on the contributors list at the top with yours?

types/webmidi/index.d.ts
calebegg: Unfortunately, this seems to be required for the "implements EventTarget" to work. Otherwise, TS complains. I copied this approach from the EventTarget implementors in lib.dom.d.ts, so I think it's the best way to do this.

types/webmidi/index.d.ts
calebegg: If this was removed, you wouldn't be able to access 'type' on e.g. MIDIConnectionEvent::port, which wouldn't make much sense.

types/webmidi/index.d.ts
calebegg: See below.

types/webmidi/index.d.ts
calebegg: Done.

calebegg:
Yeah, it seems that the project uses Prettier now for formatting. I noticed Github has an option to [ignroe whitespace](https://github.blog/2018-05-01-ignore-white-space-in-code-review/) which should help, if you hadn't spotted it.

I'm also not really suited to review changes here; I'm ultimately just making these tweaks on behalf of a friend. I removed your name so that it will go to the more general review in the future and hopefully someone else can take over ownership :)

lostfictions (approved):
@calebegg thanks! all sounds good to me. doesn't look like you pushed those line changes to the author/definitions-for comments though?

calebegg:
Oh, dang, you're right. I'm not very good at git :/

Please wait to merge this!

RyanCavanaugh (changes requested):
Marking this as needing changes per your comment

calebegg:
Thanks, updated.

calebegg:
Friendly ping! @RyanCavanaugh

sandersn:
I don't think Ryan saw the ping, but it looks like everything is good to go. Sorry about the wait!

PR #41306, opened 2019-12-30
Added react-native-share-extension

hmajid2301:
Please fill in this template.

Adding new package

### DIFF
--- a/types/react-native-share-extension/index.d.ts
+++ b/types/react-native-share-extension/index.d.ts
@@ -1,0 +1,16 @@
+// Type definitions for react-native-share-extension 2.0
+// Project: https://github.com/alinz/react-native-share-extension
+// Definitions by: Haseeb Majid <https://github.com/hmajid230>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+interface ShareData {
+    value: string;
+    type: 'text/plain' | 'images/*';
+}
+
+declare class ShareExtension {
+    data(): ShareData;
+    openURL(uri: string): void;
+}
+
+export default ShareExtension;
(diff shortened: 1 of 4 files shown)


types/react-native-share-extension/index.d.ts
```diff
@@ -0,0 +1,16 @@
...
+// Project: https://github.com/alinz/react-native-share-extension
+// Definitions by: Haseeb Majid <https://github.com/hmajid230>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+interface ShareData {
+    value: string;
+    type: 'text/plain' | 'images/*';
+}
+
+declare class ShareExtension {
```
peterblazejewicz: Can we have `close`?
https://github.com/alinz/react-native-share-extension/blob/master/lib/index.js#L8
https://github.com/alinz/react-native-share-extension/blob/master/examples/Sample1/share.ios.js#L38

peterblazejewicz (changes requested):
The linter requires tests (non-empty *-tests.tsx)

types/react-native-share-extension/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6"
+        ],
+        "noImplicitAny": true,
```
peterblazejewicz: Here the:
```json
"jsx": "react"
```
is probably to be added, if you want to use JSX tests

peterblazejewicz (approved):
LGTM
@hmajid2301 thanks for the contribution!

PR #41307, opened 2019-12-30
value validators accept number or date

mcshaz:
As can be seen  in the current vuelidate project source files, the validators handle dates as well as numbers (to be precise they handle strings which can be converted to numbers as well, but I haven't addressed this use case):

https://github.com/vuelidate/vuelidate/blob/master/src/validators/minValue.js
https://github.com/vuelidate/vuelidate/blob/master/src/validators/between.js


janesser (approved):
Looks fine to me. Thanks for your contribution.

PR #41313, opened 2019-12-30
Fix node and webpack-env compatibility

### LINKED ISSUE #9241: @storybook/react types dependency @types/webpack-env conflicting with @types/node
**Describe the bug**
`@storybook/react` types dependency `@types/webpack-env` conflicting with `@types/node`

```bash
node_modules/@types/webpack-env/index.d.ts:282:13 - error TS2403: Subsequent variable declarations must have the same type.  Variable 'require' must be of type 'Require', but here has type 'NodeRequire'.
```
**To Reproduce**
1. Setup a TS env
2. Install @storybook/react
3. Install @types/node
4. Compile project

**Screenshots**
Screenshot 2019-12-24 at 4 27 34 PM

**System:**
```
System:
    OS: macOS 10.15.1
    CPU: (8) x64 Intel(R) Core(TM) i7-3635QM CPU @ 2.40GHz
  Binaries:
    Node: 10.16.0 - ~/.nvm/versions/node/v10.16.0/bin/node
    Yarn: 1.17.3 - /usr/local/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.16.0/bin/npm
  Browsers:
    Chrome: 79.0.3945.88
    Safari: 13.0.3
```

**Additional context**
Manually editing @types/webpack-env solves this problem.
I tried adding "types" to tsconfig.json to limit type definition scope.
I am using storybook in monorepo setup using yarn.

andrewbranch:
Fixes https://github.com/storybookjs/storybook/issues/9241 (and build errors of our own) by making @types/webpack-env compatible with both v12 and v13 of @types/node. Unfortunately, I had to change which global interface `require` and `module` point at in node, which can change how packages that merge with Node’s globals can impact those types. @SimonSchick, does this look like a problem to you? Not sure what the motivation was for leaving `NodeRequire` around but changing `require` not to point to it.

### DIFF
--- a/types/webpack-env/index.d.ts
+++ b/types/webpack-env/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for webpack (module API) 1.14
+// Type definitions for webpack (module API) 1.15
 // Project: https://github.com/webpack/webpack
 // Definitions by: use-strict <https://github.com/use-strict>
 //                 rhonsby <https://github.com/rhonsby>
@@ -59,7 +59,7 @@ declare namespace __WebpackModuleApi {
          * Multiple requires to the same module result in only one module execution and only one export. Therefore a cache in the runtime exists. Removing values from this cache cause new module execution and a new export. This is only needed in rare cases (for compatibility!).
          */
         cache: {
-            [id: string]: NodeJS.Module;
+            [id: string]: NodeModule;
         }
     }
 
--- a/types/webpack-env/index.d.ts
+++ b/types/webpack-env/index.d.ts
@@ -10,6 +10,10 @@
  */
 
 declare namespace __WebpackModuleApi {
+    interface RequireResolve {
+        (id: string): string | number;
+    }
+
     interface RequireContext {
         keys(): string[];
         (id: string): any;
@@ -42,7 +46,7 @@ declare namespace __WebpackModuleApi {
          *
          * The module id is a number in webpack (in contrast to node.js where it is a string, the filename).
          */
-        resolve(path: string): number | string;
+        resolve: NodeJS.RequireResolve;
         /**
          * Like require.resolve, but doesn’t include the module into the bundle. It’s a weak dependency.
          */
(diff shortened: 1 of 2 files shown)


SimonSchick:
The motivation for the v13 changes was to cleanup globals that were not in the `NodeJS` namespace.
I left these aliases behind to provide backwards compatibility for other packages that relied on it.
What is the issue that requires the changes in the node globals?
Does decl merging not work as expected in the namespace?

andrewbranch:
> I left these aliases behind to provide backwards compatibility for other packages that relied on it.

This backward compatibility doesn’t seem to do much if the type of `require` is now `NodeJS.Require`. So declaration merging into `NodeRequire` technically is still happening, but it has no observable effect because nothing references `NodeRequire`. And if the package re-declares the global `var require: NodeRequire` (like webpack-env does), it fails to compile because that’s not the same type as @types/node says `require` is.

andrewbranch:
webpack-env can be updated to work with @types/node@13, but it was impossible to make it compatible with both v12 and v13 without this change to node. Since @types/webpack-env and @types/node are both extremely highly used, it would be nice for them to be cross-compatible at more than one version.

types/webpack-env/index.d.ts
```diff
@@ -55,19 +59,17 @@ declare namespace __WebpackModuleApi {
...
         cache: {
-            [id: string]: any;
+            [id: string]: NodeModule;
         }
     }
 
     interface Module {
         exports: any;
-        require(id: string): any;
-        require<T>(id: string): T;
```
andrewbranch: I couldn’t find from the webpack docs that this is real

types/webpack-env/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for webpack (module API) 1.14
+// Type definitions for webpack (module API) 1.15
```
andrewbranch: Does this version track something external?

SimonSchick (approved):

Issue #40650, opened 2019-11-26
[@types/styled-components] `as` prop error


mario-jerkovic:
If you know how to fix the issue, make a pull request instead.

- [x] I tried using the `@types/styled-components` package and had problems.
- [x] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [ ] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.
  - Authors: @Igorbek, @Igmat, @Lavoaster, @Jessidhia, @JKillian, @eps1lon, @flavordaaave and @wagerfield 

I am using: 
- `"@types/react": "^16.9.13`
- `"@types/styled-components": "^4.4.0"`
- `"typescript": "^3.7.2"`

I am trying to set `as` property on my component which renders `styled-component`:
```javascript

type TypographyProps<T> = React.HTMLProps<T> & {
    // ...
    component?: keyof JSX.IntrinsicElements,
    // ...
}

export const Typography = <T extends any = HTMLDivElement>(props: TypographyProps<T>) => {
    const {
        children,
        className,
        component = 'div', // <- default component type
        fontFamily = TypographyFontFamily.primary,
        gutterBottom = false,
        paragraph = false,
        underline = false,
        underlineAlignment = UnderlineAlignment.bottom,
        variant,
    } = props

    return (
        <TypographyRoot
            as={component}
            className={className}
            fontFamily={fontFamily}
            gutterBottom={gutterBottom}
            paragraph={paragraph}
            underline={underline}
            underlineAlignment={underlineAlignment}
            variant={variant}
        >
            {children}
        </TypographyRoot >
    )
}

```
The issue is that typescript throw an error:
`Type instantiation is excessively deep and possibly infinite.`

osdiab:
I am getting this as well, it's making it hard to use styled components at all

osdiab:
Additionally whenever I extend a base element's props like so:

```ts
const MyInput: React.FC<React.HTMLProps<HTMLInputElement>> = props => {
  return <input {...props} />;
}
```

Then `as` typing also causes an error, saying that `string | undefined` doesn't match `undefined` for the type of the `as` prop. `as` prop is very broken right now

orta:
[...]

Issue #40358, opened 2019-11-13
[styled-components] Problems with `as` after 4.1.22


xumix:
Have problems with `@types/styled-components`@4.1.22( also 4.4.0) and `semantic-ui-react@0.88.1` and tsc 3.7.2

```
import styled from "styled-components/macro";
import { Header } from "semantic-ui-react";
export const RegistrationLink = styled(Header)`
  color: #000 !important;
  font-weight: 400 !important;
  display: block;
  cursor: pointer;
`;
```
`<RegistrationLink as="a" textAlign="center" size="tiny" onClick={this.goToRegister} />`

Getting this error:
```
No overload matches this call.
  Overload 1 of 2, '(props: Pick<Pick<PropsWithChildren<HeaderProps>, ReactText> & Partial<Pick<PropsWithChildren<HeaderProps>, never>>, ReactText> & { ...; } & { ...; }): ReactElement<...>', gave the following error.
    Type 'string' is not assignable to type 'undefined'.
  Overload 2 of 2, '(props: StyledComponentPropsWithAs<"a", any, {}, never>): ReactElement<StyledComponentPropsWithAs<"a", any, {}, never>, string | ... 1 more ... | (new (props: any) => Component<...>)>', gave the following error.
    Type '{ children: any; as: "a"; textAlign: string; size: string; onClick: () => void; }' is not assignable to type 'IntrinsicAttributes & Pick<Pick<Pick<DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>, "children" | ... 261 more ... | "referrerPolicy"> & { ...; }, "children" | ... 262 more ... | "referrerPolicy"> & Partial<...>, "children" | ... 262 more ... | "referrerPolicy"> & { ...; } & { ...; }'.
      Property 'textAlign' does not exist on type 'IntrinsicAttributes & Pick<Pick<Pick<DetailedHTMLProps<AnchorHTMLAttributes<HTMLAnchorElement>, HTMLAnchorElement>, "children" | ... 261 more ... | "referrerPolicy"> & { ...; }, "children" | ... 262 more ... | "referrerPolicy"> & Partial<...>, "children" | ... 262 more ... | "referrerPolicy"> & { ...; } & { ...; }'.  TS2769
```

@surgeboris @sandersn @Jessidhia @smockle

bdelaforest:
[...]

PR #41316, opened 2019-12-31
[styled-components]: revert 'as' prop overload change for TS 3.7

Indimeco:
Please fill in this template.

## Context
The overloads for the 'as' prop in styled-components was recently changed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40209 which understood that
>This is potentially breaking for users that were using the as= property, but this is one of the "good breaks": the old as= behavior would use the wrong prop types.

However, the change has proven to be definitely breaking and more severe than assumed: users cannot use the 'as' prop at all because there are no types for it.
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/40358
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/40650

If this change only threw errors when 'as' causes the component to receive the wrong types due to component change then that would be understandable.

I do not have the typescript prowess to understand how close the below old code was to achieving truly polymorphic 'as' typings. If someone can make it possible then that would absolutely be the preferred resolution.
```
    // NOTE: TS <3.2 will refuse to infer the generic and this component becomes impossible to use in JSX
    // just the presence of the overload is enough to break JSX
    //
    // TODO (TypeScript 3.2): actually makes the 'as' prop polymorphic
    // (
    //     props: StyledComponentProps<C, T, O, A> & { as?: never }
    //   ): React.ReactElement<StyledComponentProps<C, T, O, A>>
    // <AsC extends keyof JSX.IntrinsicElements | React.ComponentType<any> = C>(
    //   props: StyledComponentPropsWithAs<AsC, T, O, A>
    // ): React.ReactElement<StyledComponentPropsWithAs<AsC, T, O, A>>

    // TODO (TypeScript 3.2): delete this overload
    (
        props: StyledComponentProps<C, T, O, A> & {
            /**
             * Typing Note: prefer using .withComponent for now as it is actually type-safe.
             *
             * String types need to be cast to themselves to become literal types (as={'a' as 'a'}).
             */
            as?: keyof JSX.IntrinsicElements | React.ComponentType<any>;
        }
    ): React.ReactElement<StyledComponentProps<C, T, O, A>>;
```

## Resolution
Therefore this PR simply re-instates the pre 3.7 overload for non-typesafe 'as'

### DIFF
--- a/types/styled-components/ts3.7/index.d.ts
+++ b/types/styled-components/ts3.7/index.d.ts
@@ -154,12 +154,17 @@ export interface StyledComponentBase<
     A extends keyof any = never
 > extends ForwardRefExoticBase<StyledComponentProps<C, T, O, A>> {
     // add our own fake call signature to implement the polymorphic 'as' prop
+    // TODO replace the following call signature with an actually polymorphic 'as' to make 'as' typesafe
     (
-        props: StyledComponentProps<C, T, O, A> & { as?: never }
-      ): React.ReactElement<StyledComponentProps<C, T, O, A>>;
-    <AsC extends keyof JSX.IntrinsicElements | React.ComponentType<any> = C>(
-      props: StyledComponentPropsWithAs<AsC, T, O, A>
-    ): React.ReactElement<StyledComponentPropsWithAs<AsC, T, O, A>>;
+        props: StyledComponentProps<C, T, O, A> & {
+            /**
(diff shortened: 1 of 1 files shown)


Indimeco:
Now that I can see the tests (didn't work on my machine) it appears I was mistaken. 
The scenarios failing in my app succeed in test. The issues I linked may be unrelated.

Will close this. Apologies!

PR #41321, opened 2019-12-31
[http-proxy] update TargetUrl

bodinsamuel:
## Template

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.npmjs.com/package/http-proxy#http---https-using-a-pkcs12-client-certificate

## Changes

Reflect documented actual possibility with `target`
Documented in [Readme](https://www.npmjs.com/package/http-proxy#http---https-using-a-pkcs12-client-certificate) and [test](https://github.com/http-party/node-http-proxy/blob/a3fe02d651d05d02d0ced377c22ae8345a2435a4/test/lib-http-proxy-common-test.js#L317-L334)

### DIFF
--- a/types/http-proxy/index.d.ts
+++ b/types/http-proxy/index.d.ts
@@ -17,11 +17,11 @@ import * as events from "events";
 import * as url from "url";
 import * as stream from "stream";
 
-type ProxyTarget = ProxyTargetUrl | ProxyTargetDetailled;
+type ProxyTarget = ProxyTargetUrl | ProxyTargetDetailed;
 
 type ProxyTargetUrl = string | Partial<url.Url>;
 
-interface ProxyTargetDetailled {
+interface ProxyTargetDetailed {
   host: string;
   port: number;
   protocol?: string;
--- a/types/http-proxy/index.d.ts
+++ b/types/http-proxy/index.d.ts
@@ -4,6 +4,7 @@
 //                 Florian Oellerich <https://github.com/Raigen>
 //                 Daniel Schmidt <https://github.com/DanielMSchmidt>
 //                 Jordan Abreu <https://github.com/jabreu610>
+//                 Samuel Bodin <https://github.com/bodinsamuel>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.1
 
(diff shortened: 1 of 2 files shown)


types/http-proxy/index.d.ts
```diff
@@ -16,8 +17,25 @@ import * as events from "events";
 import * as url from "url";
 import * as stream from "stream";
 
+type ProxyTarget = ProxyTargetUrl | ProxyTargetDetailled;
```
armanio123: There's a typo on `Detailled`.

types/http-proxy/index.d.ts
```diff
@@ -16,8 +17,25 @@ import * as events from "events";
...
+
 type ProxyTargetUrl = string | Partial<url.Url>;
 
+interface ProxyTargetDetailled {
```
armanio123: Could you please point me the documentation or code that uses this extra properties. The `Partial<url.Url>` already define some of this and I didn't found the details on the documentation.

armanio123 (changes requested):

types/http-proxy/index.d.ts
bodinsamuel: Yes it's partially covered by `Partial<url.Url>` but some of them are just slightly different, e.g: port is a string in url.URL and the lib allow number
I had put one link in the pr but it's indeed not enough:
- https://www.npmjs.com/package/http-proxy#http---https-using-a-pkcs12-client-certificate
- https://github.com/http-party/node-http-proxy/blob/a3fe02d651d05d02d0ced377c22ae8345a2435a4/lib/http-proxy/common.js#L37
- https://github.com/http-party/node-http-proxy/blob/a3fe02d651d05d02d0ced377c22ae8345a2435a4/test/lib-http-proxy-common-test.js#L322

types/http-proxy/index.d.ts
bodinsamuel: oh indeed 🤭

types/http-proxy/index.d.ts
armanio123: Thanks for pointing the code. Merging the PR.

armanio123 (approved):

bodinsamuel:
thanks @armanio123 🙏

PR #41322, opened 2019-12-31
ffmpeg-static-3.0.0-export path as default instead of module.path

Sharpiro:
- [ x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.npmjs.com/package/ffmpeg-static#example-usage
- [ x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.

* This PR changes the default import to a string rather than a module that has a `path` property.
* Followed naming conventions from the package documentation.
* Modified test as it was not reporting that the import property path was undefined.


Sharpiro:
I removed `ffprobe-static` typing as it seems outdated and also may have moved to its own npm package @ https://www.npmjs.com/package/ffprobe-static

Sharpiro:
It looks like @iamstevetran is no longer on github.  Tagging the package author @eugeneware if he would like to weigh in.

jumoog (approved):
looks good for me

armanio123 (changes requested):
I don't think you should remove `ffprobe-static` even if it's outdated. The library doesn't provide definitions by itself and the documentation doesn't states its not being maintained.

Sharpiro:
Ok I originally deleted the `ffprobe-static` typings b/c they were causing the tests to fail with the new `ffmpeg-static` typings.  But I think I may have found a better solution by fixing a bug in the `ffprobe-static` tests

types/ffmpeg-static/ffmpeg-static-tests.ts
```diff
@@ -1,3 +1,5 @@
-import * as ffmpegStatic from 'ffmpeg-static';
+import * as pathToFfmpeg from 'ffmpeg-static';
 
-ffmpegStatic.path;
+if (typeof pathToFfmpeg !== "string") {
```
armanio123: The test code is not actually run, only checked against the compiler. This will not prove that `pathToFfmpeg` is a string.

Instead I think you want to use:
```js
pathToFfmpeg; // $ExpectType string
```

armanio123 (changes requested):
Thanks for rolling back the deletions. Only one more comment.

Sharpiro:
That's cool, never used `dtslint` before

armanio123 (approved):

PR #41328, opened 2020-01-01
react-widgets: Fixed allowed values for onToggle and defaultOpen props

jankalfus:
- Provide a URL to documentation or source code which provides context for the suggested changes: https://jquense.github.io/react-widgets/api/DateTimePicker/#open

Fixed allowed values for onToggle and defaultOpen props to match the library docs and behavior.


r3nya (approved):
Happy new year! 🎄 

LGTM. 👍

PR #41342, opened 2020-01-02
line-by-line: enable all encodings supported by Node

winston01:
Please fill in this template.

### DIFF
--- a/types/line-by-line/index.d.ts
+++ b/types/line-by-line/index.d.ts
@@ -5,12 +5,12 @@
 
 /// <reference types="node" />
 
-import { EventEmitter } from "events";
-import { Readable } from "stream";
+import { EventEmitter } from 'events';
+import { Readable } from 'stream';
 
 interface LineByLineReaderOptions {
     /** The encoding to use. */
-    encoding?: "ascii" | "base64" | "utf8";
+    encoding?: 'ascii' | 'utf8' | 'utf16le' | 'ucs2' | 'base64' | 'latin1' | 'binary' | 'hex';
     /** If set to true, empty lines do not emit the "line" event. */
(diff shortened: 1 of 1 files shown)


types/line-by-line/index.d.ts
```diff
@@ -5,12 +5,12 @@
...
 
-import { EventEmitter } from "events";
-import { Readable } from "stream";
+import { EventEmitter } from 'events';
+import { Readable } from 'stream';
 
 interface LineByLineReaderOptions {
     /** The encoding to use. */
-    encoding?: "ascii" | "base64" | "utf8";
+    encoding?: 'ascii' | 'utf8' | 'utf16le' | 'ucs2' | 'base64' | 'latin1' | 'binary' | 'hex';
```
armanio123: Documentation only states that `"ascii" | "base64" | "utf8"` are allowed. Could you please point to where the other values are permitted?

armanio123 (changes requested):

types/line-by-line/index.d.ts
winston01: Hi @armanio123 

encoding setting is used here:

```
		readStream = this._readStream ? this._readStream :
			fs.createReadStream(this._filepath, this._streamOptions);
```

[fs.createReadStream documentation states](https://nodejs.org/api/fs.html#fs_fs_createreadstream_path_options): "The encoding can be any one of those accepted by Buffer.", which in turn are [all the ones listed in this PR](https://nodejs.org/api/buffer.html#buffer_buffers_and_character_encodings).

armanio123: That makes sense, thanks.

armanio123 (approved):

PR #41353, opened 2020-01-02
feat(node): v13.6

SimonSchick:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V13.md#13.5.0

@Flarna for early eyes, probably missed something


types/node/wasi.d.ts
```diff
@@ -0,0 +1,45 @@
...
+         * directories within the sandbox. The corresponding values in `preopens` are
+         * the real paths to those directories on the host machine.
+         */
+        preopens?: {
+            [key: string]: string;
+        };
+    }
+
+    class WASI {
+        constructor(options: WASIOptions);
```
Flarna:
```suggestion
        constructor(options?: WASIOptions);
```

types/node/wasi.d.ts
```diff
@@ -0,0 +1,45 @@
...
+         * An array of strings that the WebAssembly application will
+         * see as command line arguments. The first argument is the virtual path to the
+         * WASI command itself.
+         */
+        args?: string[];
+        /**
+         * An object similar to `process.env` that the WebAssembly
+         * application will see as its environment.
+         */
+        env?: object;
```
Flarna: maybe use `NodeJS.ProcessEnv` or similar here?

types/node/http.d.ts
```diff
@@ -77,6 +77,10 @@ declare module "http" {
         defaultPort?: number | string;
         localAddress?: string;
         socketPath?: string;
+        /**
+         * @default 8192
+         */
+        maxHeaderSize?: number;
```
Flarna:
```suggestion
        readonly maxHeaderSize: number;
```

Flarna (commented):
Left some minor comments. I have not verified that everything is included, if not in can be added later anyway.

types/node/http.d.ts
SimonSchick: These are options, not sure why these would be readonly?

Flarna: Ah sorry. You are correct. I mixed this up with the module static which is not part of this PR.

SimonSchick:
@rbuckton any concerns with moving the ts3.2 folder to 3.5 to enable WebAssembly libs?
I'd rather not do another split.

types/node/stream.d.ts
```diff
@@ -6,7 +6,9 @@ declare module "stream" {
     }
 
     namespace internal {
-        class Stream extends internal { }
+        class Stream extends internal {
+            constructor(opts?: ReadableOptions);
```
Flarna: I don't agree with this constructor of the legacy `Stream`. According to [code](https://github.com/nodejs/node/blob/1a552f614b4ec2b9faf0bf932ad4e6ef4fdcaf01/lib/internal/streams/legacy.js#L9-L11) options are just forwarded to `EventEmitter`.

SimonSchick: This seems in conflict with `readable-stream` which is kind of annoying.

SimonSchick:
ping

galkin (approved):
Next time let's split in smaller changes for fast review.

types/node/stream.d.ts
Flarna: ok, seems to be no easy workaround.

types/node/wasi.d.ts
Flarna: This comment is still valid. Any reason why `object` is used here?

types/node/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for non-npm package Node.js 13.1
+// Type definitions for non-npm package Node.js 13.5
```
Flarna: based on PR title this should be 13.6 now

SimonSchick: Welp, that was merged too quickly 😛.
Will fix with 13.7 PR :)

jwalton:
Prior to this change (in @types/node@13.1.8) you could do:

```ts
import EventEmitter from 'events';

class Foo extends EventEmitter {
}
```

and this would work.  Now it will fail with `error TS2507: Type 'typeof EventEmitter' is not a constructor function type.`

If you run into this you can fix it with:

```ts
import { EventEmitter } from 'events';

class Foo extends EventEmitter {
}
```

but I do wonder if this change was intentional?

SimonSchick:
@jwalton actually this wasn't intentional otherwise I'd have removed the `namespace` and directly exported everything. 

The diff to restore this functionality would look like this:
```diff
diff --git a/types/node/events.d.ts b/types/node/events.d.ts
index b07defc086..e6543ab822 100644
--- a/types/node/events.d.ts
+++ b/types/node/events.d.ts
@@ -1,20 +1,28 @@
 declare module "events" {
-    interface EventEmitterOptions {
-        /**
-         * Enables automatic capturing of promise rejection.
-         */
-        captureRejections?: boolean;
-    }
-
-    interface NodeEventTarget {
-        once(event: string | symbol, listener: (...args: any[]) => void): this;
-    }
-
-    interface DOMEventTarget {
-        addEventListener(event: string, listener: (...args: any[]) => void, opts?: { once: boolean }): any;
+    class EventEmitter {
+        constructor(options?: EventEmitter.EventEmitterOptions);
+        /** @deprecated since v4.0.0 */
+        static listenerCount(emitter: EventEmitter, event: string | symbol): number;
+        static defaultMaxListeners: number;
     }
 
     namespace EventEmitter {
+
+        interface EventEmitterOptions {
+            /**
+             * Enables automatic capturing of promise rejection.
+             */
+            captureRejections?: boolean;
+        }
+
+        interface NodeEventTarget {
+            once(event: string | symbol, listener: (...args: any[]) => void): this;
+        }
+
+        interface DOMEventTarget {
+            addEventListener(event: string, listener: (...args: any[]) => void, opts?: { once: boolean }): any;
+        }
+
         function once(emitter: NodeEventTarget, event: string | symbol): Promise<any[]>;
         function once(emitter: DOMEventTarget, event: string): Promise<any[]>;
         function on(emitter: EventEmitter, event: string): AsyncIterableIterator<any>;
diff --git a/types/node/test/events.ts b/types/node/test/events.ts
index 621dd68067..0d223dd46c 100644
--- a/types/node/test/events.ts
+++ b/types/node/test/events.ts
@@ -84,3 +84,7 @@ async function test() {
         console.log(e);
     }
 }
+
+{
+    const em = new events();
+}
```

I wonder if we should do this however, the `events` module is very problematic to model and update due to circular references eg. `module.exports = EventEmitter; EventEmitter.EventEmitter = EventEmitter` which usually requires an intermediate __fake__ class which caused issues and confusions before.

types/node/util.d.ts
```diff
@@ -1,5 +1,10 @@
 declare module "util" {
     interface InspectOptions extends NodeJS.InspectOptions { }
+    type Style = 'special' | 'number' | 'bigint' | 'boolean' | 'undefined' | 'null' | 'string' | 'symbol' | 'date' | 'regexp' | 'module';
```
arcanis: I think this list is missing `name`, which is used to style the property names:
https://nodejs.org/api/util.html#util_customizing_util_inspect_colors

SimonSchick: Lol yea https://github.com/nodejs/node/blob/master/lib/internal/util/inspect.js#L389
Feel free to file a PR.

evert:
Also running into this. Our package breaks with @types/node@13 due to us using the default export of `events`.

PR #41358, opened 2020-01-03
[cli-color@2.0] Add types for cli-color

OpportunityLiu:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/medikoo/cli-color>

### DIFF
--- a/types/cli-color/columns.d.ts
+++ b/types/cli-color/columns.d.ts
@@ -1,0 +1,29 @@
+declare namespace columns {
+    export interface ColumnOptions {
+        /**
+         * align: Possible options: `'left'`, `'right'` (defaults to `'left'`)
+         */
+        align?: 'left' | 'right';
+    }
+
+    export interface ColumnsOptions {
+        /**
+         *  Custom colums separator (defaults to `|`)
+         */
+        sep?: string;
+        /**
+         * columns: Per column customizations, as e.g. `[{ align: 'right' }, null, { align: 'left' }]`
+         */
+        columns?: Array<ColumnOptions | null>;
+    }
+
+    export type Row = Iterable<any> | ArrayLike<any>;
+    export type Data = Iterable<Row> | ArrayLike<Row>;
+}
+
+/**
+ * Outputs aligned table of columns.
+ */
+declare function columns(data: ReadonlyArray<ReadonlyArray<any>>, options?: columns.ColumnsOptions): string;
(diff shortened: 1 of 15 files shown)


armanio123:
I'm not very familiar with the library so I'll wait a couple more days for @ChaosinaCan to review; otherwise I'll merge it.

ChaosinaCan:
Whoops. I missed the original notification for this. I haven't used this library since I originally wrote the types for it, but I'll check this against the current documentation.

types/cli-color/columns.d.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    export type Row = Iterable<any> | ArrayLike<any>;
+    export type Data = Iterable<Row> | ArrayLike<Row>;
+}
+
+/**
+ * Outputs aligned table of columns.
+ */
+declare function columns(data: ReadonlyArray<ReadonlyArray<any>>, options?: columns.ColumnsOptions): string;
+declare function columns(data: columns.Data, options?: columns.ColumnsOptions): string;
```
ChaosinaCan: Could combine these two overloads using

```ts
declare function columns(data: columns.Data | ReadonlyArray<ReadonlyArray<any>>, ...
```

though I'm not sure if that makes the types any easier to use.

joelspadin (approved):

types/cli-color/columns.d.ts
OpportunityLiu: In fact, `ReadonlyArray<ReadonlyArray<any>>` fits signature of `columns.Data`, added it just for a hint to use arrays for `columns.Data`, should I just remove it?

ChaosinaCan: Oh, true. I'm probably not the best person to ask which way is better. Probably depends on how confusing the second overload looks in code completion.

armanio123: Is recommended to use union types: https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#use-union-types

Also put more specific types first (this is for overloads, I'm not sure if this will also apply for types) ://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#ordering

armanio123: I'll merge as is. Any improvements could be handled by another PR.

PR #41368, opened 2020-01-03
webpack: Add few missing type definitions on compilation.MainTemplate

Lapple:
- adds mainTemplate.hooks.requireEnsure hook
- adds mainTemplate.hooks.localVars hook
- adds mainTemplate.renderRequireFunctionForModule method
- adds mainTemplate.renderAddModule method

Please fill in this template.

  * [requireEnsure hook](https://github.com/webpack/webpack/blob/v4.41.5/lib/MainTemplate.js#L81-L86)
  * [localVars hook](https://github.com/webpack/webpack/blob/v4.41.5/lib/MainTemplate.js#L94)
  * [renderRequireFunctionForModule method](https://github.com/webpack/webpack/blob/v4.41.5/lib/MainTemplate.js#L467)
  * [renderAddModule method](https://github.com/webpack/webpack/blob/v4.41.5/lib/MainTemplate.js#L484) 

### DIFF
--- a/types/webpack/index.d.ts
+++ b/types/webpack/index.d.ts
@@ -1073,9 +1073,13 @@ declare namespace webpack {
           hooks: {
             jsonpScript?: SyncWaterfallHook<string, Chunk, string>;
             requireExtensions: SyncWaterfallHook<string, Chunk, string>;
+            requireEnsure: SyncWaterfallHook<string, Chunk, string>;
+            localVars: SyncWaterfallHook<string, Chunk, string>;
           };
           outputOptions: Output;
           requireFn: string;
+          renderRequireFunctionForModule(hash: string, chunk: Chunk, varModuleId?: number | string): string;
+          renderAddModule(hash: string, chunk: Chunk, varModuleId: number | string | undefined, varModule: string): string;
         }
         class ChunkTemplate extends Tapable {}
         class HotUpdateChunkTemplate extends Tapable {}
--- a/types/webpack/webpack-tests.ts
+++ b/types/webpack/webpack-tests.ts
@@ -307,6 +307,14 @@ configuration = {
         mainTemplate.hooks.requireExtensions.tap('SomePlugin', resource => {
           return resource.trimLeft();
         });
+        mainTemplate.hooks.requireEnsure.tap('SomePlugin', (resource, chunk, hash, chunkIdExpression: string) => {
+          mainTemplate.renderRequireFunctionForModule(hash, chunk, 'moduleId');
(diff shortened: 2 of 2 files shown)


types/webpack/index.d.ts
```diff
@@ -1073,9 +1073,13 @@ declare namespace webpack {
           hooks: {
             jsonpScript?: SyncWaterfallHook<string, Chunk, string>;
             requireExtensions: SyncWaterfallHook<string, Chunk, string>;
+            requireEnsure: SyncWaterfallHook<string, Chunk, string>;
+            localVars: SyncWaterfallHook<string, Chunk, string>;
           };
           outputOptions: Output;
           requireFn: string;
+          renderRequireFunctionForModule(hash: string, chunk: Chunk, varModuleId: number | string): string;
```
ofhouse: The param `varModuleId` is marked as optional in the JSDoc annotation: https://github.com/webpack/webpack/blob/v4.41.5/lib/MainTemplate.js#L464

types/webpack/index.d.ts
```diff
@@ -1073,9 +1073,13 @@ declare namespace webpack {
...
+          renderAddModule(hash: string, chunk: Chunk, varModuleId: number | string, varModule: string): string;
```
ofhouse: The param `varModuleId` is marked as optional in the JSDoc annotation:
https://github.com/webpack/webpack/blob/v4.41.5/lib/MainTemplate.js#L480

ofhouse (changes requested):
Thanks for the PR!

I've added two comments where the corresponding JSDoc annotation claims that a parameter is optional while your typing marks them as required.

types/webpack/index.d.ts
Lapple: Thanks for checking, indeed I have missed it. Changed now to:

```ts
renderRequireFunctionForModule(hash: string, chunk: Chunk, varModuleId?: number | string): string;
```

types/webpack/index.d.ts
Lapple: Had to change this one to:

```ts
renderAddModule(hash: string, chunk: Chunk, varModuleId: number | string | undefined, varModule: string): string;
```

but from the code it looks like passing undefined there would render something like `module[undefined] = myModule;`. Anyway, I assume following the JSDoc is more preferable?

ofhouse: Yeah I agree with that it makes not much sense for this param here.
But since it doesn't break anything here because it is only used to generate a string, I think it's safe to follow the JSDoc here.

ofhouse (approved):
Looks fine to me 👍

PR #41388, opened 2020-01-04
react-shadow-dom-retarget-events: update tests

sandersn:
Typescript 3.8 has a stricter type for customElements.define, so I updated the tests to match the type.


remcohaszing (changes requested):
I don’t believe this change is correct. If this change is required for TypeScript 3.8, it seems like a bug in TypeScript 3.8.

The affected line isn’t necessary to test the type definitions. It might as well be removed.

sandersn:
You are right; it's supposed to be a construct signature, not a call signature. The old type of `customElements.define` was too loose, but the new type is wrong.

PR #41397, opened 2020-01-04
Fix wrong type in blessed (screen.ignoreLocked)

danikaze:
[blessed](https://github.com/chjj/blessed) is not being maintained but is still usable
Just found a wrong typed attribute, so fixed it.

### DIFF
--- a/types/blessed/index.d.ts
+++ b/types/blessed/index.d.ts
@@ -4,6 +4,7 @@
 //                 Steve Kellock <https://github.com/skellock>
 //                 Max Brauer <https://github.com/mamachanko>
 //                 Nathan Rajlich <https://github.com/TooTallNate>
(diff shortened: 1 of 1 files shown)


skellock (approved):

danikaze:
Wasn't the previous version already `0.1.16`?
Should I had the version increased to 0.1.17 in the PR itself? :-/

armanio123:
You don't need to. The pipeline automatically increases the build number. I see on npm that the previous version, released 24 days ago was `0.1.15`, so this should be fine.

danikaze:
@armanio123 thanks for the reply. Yeah I thought so but, also found that the version I had installed before I made the PR was 0.1.16... so it felt weird...
Anyways, found other bugs so probably creating a new PR in a few days. Hopefully with a bunch of fixes not only one :)

PR #41398, opened 2020-01-04
chai-style: add types/tests

AviVahl:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/chai-style/index.d.ts
+++ b/types/chai-style/index.d.ts
@@ -1,0 +1,18 @@
+// Type definitions for chai-style 1.0
+// Project: https://github.com/darlanmendonca/chai-style
+// Definitions by: Avi Vahl <https://github.com/AviVahl>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
+
+/// <reference types="chai" />
+
+declare const chaiStyle: Chai.ChaiPlugin;
+export = chaiStyle;
+
+declare global {
+    namespace Chai {
+        interface Assertion extends LanguageChains, NumericComparison, TypeComparison {
+            style(styleName: string, styleValue?: string): Assertion;
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


types/chai-style/index.d.ts
```diff
@@ -0,0 +1,18 @@
+// Type definitions for chai-style 1.0
+// Project: https://github.com/darlanmendonca/chai-style
+// Definitions by: Avi Vahl <https://github.com/AviVahl>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.0
```
armanio123: I think you can remove this line as there's nothing using version 3.0

armanio123 (changes requested):

types/chai-style/index.d.ts
AviVahl: `@types/chai` themselves requires ts 3.0+:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/chai/index.d.ts

so I guessed any assertion plugin (that imports chai) also has that requirement. Is that not correct?

AviVahl:
Removing the min typescript version fails the tests (locally):
```
Testing 1 changed packages: chai-style
Testing 0 dependent packages: 
Error: chai-style depends on chai but has a lower required TypeScript version.
    at checkTypeScriptVersions (/home/avi/projects/DefinitelyTyped/node_modules/types-publisher/src/check-parse-results.ts:51:23)
    at Object.checkParseResults [as default] (/home/avi/projects/DefinitelyTyped/node_modules/types-publisher/src/check-parse-results.ts:19:5)
    at main (/home/avi/projects/DefinitelyTyped/node_modules/types-publisher/src/tester/test.ts:26:9)
npm ERR! Test failed.  See above for more details.
```
@armanio123 any chance you could re-review? :)

types/chai-style/index.d.ts
armanio123: Aha, makes sense. Thanks for checking.

armanio123 (approved):

PR #41404, opened 2020-01-05
Adds existing npm lib multicodec + tests

carsonfarmer:
Signed-off-by: Carson Farmer <carson.farmer@gmail.com>

Please fill in this template.

If adding a new definition:


types/multicodec/index.d.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Human friendly names for printing, e.g. in error messages
+ */
+export const print: Record<number, string>;
+
+/**
+ * Map of codecConstant to code
+ */
+export * from './constants';
```
carsonfarmer: Could this 'feature' be better represented using a namespace?

types/multicodec/index.d.ts
```diff
@@ -0,0 +1,56 @@
+// Type definitions for multicodec 1.0
+// Project: https://github.com/multiformats/js_multicodec#readme
```
armanio123: There's a typo. `js_multicodec` to `js-multicodec`

armanio123 (changes requested):
I think `multicodec` also provides a global value. See https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#global-libraries for details on how to represent this.

carsonfarmer:
Typo fixed, good catch!

> I think `multicodec` also provides a global value. See https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#global-libraries for details on how to represent this.

Maybe I'm missing something, but I don't see any evidence of this? The `aegir build` script produces a webpack umd, but doesn't look like there are any globals available..? Could you point me to where I might find this?

armanio123:
> Maybe I'm missing something, but I don't see any evidence of this? The aegir build script produces a webpack umd, but doesn't look like there are any globals available..? Could you point me to where I might find this?

Here's documentation of `multicodec` used on the on the browser: https://multiformats.github.io/js-multicodec/

If you check file https://unpkg.com/multicodec@1.0.0/dist/index.js, line 9:
```
root["Multicodec"] = factory();
```
That's where the global variable is being assigned.

carsonfarmer:
Ah I see, thanks @armanio123, I didn't read that initial link you posted properly. My apologies, yes you are right, and I think the added `export as namespace Multicodec` should cover it. Though it wasn't clear to me how best to 'test' this?

armanio123:
> Though it wasn't clear to me how best to 'test' this?
Since you're declaring a global variable by exporting the namespace, you should be able to access it's properties en methods just by using `Multicodec` on the test file (no import require).

carsonfarmer:
Cool, so that's great. But is it only available globally in a browser context? If I look at the `"main"` entry in the `package.json` I see `"src/index.js"`, which does not have anything exported globally. Is this still 'ok' from a ambient types perspective? Apologies for my ignorance here, this is apparently the first package that has a global export that I've written types for 😬.

So I guess my final question here is, is it misleading to have globally exported types in say a NodeJS context? I see that if I _do_ include the import, this masks the global value, which makes sense. However, in NodeJS, if I _don't_ import the package, I _won't_ actually have it in my global namespace...

armanio123:
>So I guess my final question here is, is it misleading to have globally exported types in say a NodeJS context? I see that if I do include the import, this masks the global value, which makes sense. However, in NodeJS, if I don't import the package, I won't actually have it in my global namespace...

I can see how this is confusing, but that's how umd modules work. It would be up to the developer to understand the library and decide the best implementation. As you correctly stated, in the case of a NodeJs environment, you require specify the import. For the case of the browser you must reference the global library by using a `script` block (I think the author documented this to be on the `dist` folder) and make the global variable available.

Btw for test purposes both scenarios should be ok. You can either decide to import or use the global scope. I'll approve and merge the PR.

armanio123 (approved):

carsonfarmer:
Thanks so much for a) explaining this, super helpful, and b) approving and merging! Cheers @armanio123!

PR #41407, opened 2020-01-05
[aws-lambda] Add AWS EventBridge handler

sheepsteak:
EventBridge is a replacement for CloudWatch Events as stated on https://docs.aws.amazon.com/eventbridge/latest/userguide/what-is-amazon-eventbridge.html

>EventBridge was formerly called Amazon CloudWatch Events. It includes new features that enable you to receive events from SaaS partners and your own applications. Existing CloudWatch Events users can access their existing default bus, rules, and events in the new EventBridge console and in the CloudWatch Events console. EventBridge uses the same CloudWatch Events API, so all of your existing CloudWatch Events API usage remains the same.

I've added an `EventBridgeEvent` interface that can be the basis of all possible events; both [AWS](https://docs.aws.amazon.com/eventbridge/latest/userguide/event-types.html) and custom ones.

`EventBridgeHandler<T>` can be used as a basis for a custom event. I've rewritten `ScheduledEvent` in terms of `EventBridgeEvent` to show how we can add all the possible AWS Events in the future. We already have `CodePipelineCloudWatchEvent` too and this could be updated to use `EventBridgeEvent` as a base.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://docs.aws.amazon.com/eventbridge/latest/userguide/what-is-amazon-eventbridge.html


elibarzilay:
Ping @skarum @tobyhede @buggy @y13i @wwwy3y3 @OrthoDex @MichaelMarner @daniel-cottone @kostya-misura @coderbyheart @palmithor @daniloraisi @simonbuchan @Haydabase @repl-chris @aneilbaboo @jeznag @louislarry @dpapukchiev @ohookins @trevor-leach @jagregory @dalen @loikg @skyzenr @redlickigrzegorz @juancarbonel @pwmcintyre @alex-bolenok-centralreach @marianzange @apepper

types/aws-lambda/index.d.ts
```diff
@@ -465,17 +466,25 @@ export type CloudFormationCustomResourceResponse =
...
-    id: string;
-    resources: string[];
+    version: string;
+}
+
+/**
+ * See https://docs.aws.amazon.com/eventbridge/latest/userguide/event-types.html#schedule-event-type
+ */
+export interface ScheduledEvent extends EventBridgeEvent<{}> {
+    'detail-type': "Scheduled Event";
```
simonbuchan: Instead of `{}`, use `any` here to preserve the previous type of `detail`: otherwise you are breaking handlers that just do `event.detail.someValue`...

Also, not sure this is an improvement or not ,but an alternative I've used is something like:

```ts
export interface EventBridgeEvent<Type, Detail> {
  "detail-type": Type;
  detail: Detail;
  ...
}
export type ScheduledEvent = EventBridgeEvent<"Scheduled Event", any>;
```

This makes it easier to declare large sets of these event types.

types/aws-lambda/index.d.ts
```diff
@@ -1153,7 +1162,7 @@ export type SQSHandler = Handler<SQSEvent, void>;
 
 export type CloudFormationCustomResourceHandler = Handler<CloudFormationCustomResourceEvent, void>;
 
-// TODO: CloudWatchEvents
+export type EventBridgeHandler<T = object> = Handler<EventBridgeEvent<T>, void>;
```
simonbuchan: Don't have defaults here (unless they are on the event type as well), and I have similar suspicion on the detail type here as above: `object` is not the same type as `{}`, and has little reason to be used in most situations. It may well be the case that the type is never going to be a primitive (which is what `object` implies), but there's little value gained by defaulting that here.

simonbuchan (changes requested):
Looks good, but a couple of quibbles to avoid breaking the past and future.

Sorry for the late review!

sheepsteak:
@simonbuchan sorry, I haven’t had time to change this yet. I’ll reopen in the next day or two.

PR #41408, opened 2020-01-05
[react-native] fixes StyleSheet.flatten use caused by StyleSheet.comp…

vtolochk:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://facebook.github.io/react-native/docs/stylesheet#compose>>

This pull request fixes StyleSheet.flatten use case of flattening multiple style props to extract final values.
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40950#issuecomment-568791033

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -1963,10 +1963,10 @@ export interface ViewPropsAndroid {
 }
 
 type Falsy = undefined | null | false;
-interface RecursiveArray<T> extends Array<T | RecursiveArray<T>> {}
+interface RecursiveArray<T> extends Array<T | ReadonlyArray<T> | RecursiveArray<T>> {}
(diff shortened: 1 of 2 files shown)


maxb94:
This still causes an error when setting the style prop of components to an array. For example:

`const foo = (style: StyleProp<ViewStyle>) => <View style={[style, style]} />`

The error goes by removing `ReadOnlyArray` from both the `StyleProp` and `RecursiveArray` type definitions.

alloy:
Skimming through the Flow code I think there’s probably room for improvements, but for now this should be good enough and at least unbreak existing code. Thanks 👍

Tetawex:
Do you think slightly changing declaration of `StyleProp<T>` to something like
```
export type StyleProp<T> = T | RegisteredStyle<T> | RecursiveArray<StyleProp<T>> | Falsy;
```
might somewhat improve readability and eliminate repeating `T | RegisteredStyle<T> | Falsy` ? (unless it degrades performance)

PR #41411, opened 2020-01-05
Add types for meyda

hughrawlinson:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/meyda/index.d.ts
+++ b/types/meyda/index.d.ts
@@ -1,0 +1,108 @@
+// Type definitions for meyda 4.3
+// Project: https://github.com/meyda/meyda
+// Definitions by: Damien Erambert <https://github.com/eramdam>
+//                 Hugh Rawlinson <https://github.com/hughrawlinson>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export as namespace Meyda;
+
+export type MeydaWindowingFunction = 'blackman' | 'sine' | 'hanning' | 'hamming';
+
+export type MeydaAudioFeature =
+    | 'amplitudeSpectrum'
+    | 'chroma'
+    | 'complexSpectrum'
+    | 'energy'
+    | 'loudness'
+    | 'mfcc'
+    | 'perceptualSharpness'
+    | 'perceptualSpread'
+    | 'powerSpectrum'
+    | 'rms'
+    | 'spectralCentroid'
+    | 'spectralFlatness'
+    | 'spectralFlux'
+    | 'spectralKurtosis'
+    | 'spectralRolloff'
+    | 'spectralSkewness'
+    | 'spectralSlope'
+    | 'spectralSpread'
+    | 'zcr'
+    | 'buffer';
+
+export interface MeydaAnalyzerOptions {
+    audioContext: AudioContext;
+    source: AudioNode;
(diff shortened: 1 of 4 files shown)


peterblazejewicz (approved):
LGTM
@hughrawlinson  thanks for the contribution!

armanio123 (changes requested):
I believe Meyda contains also a global namespace `Meyda`. If that's the case, the `d.ts` file needs to be modified to represent it. See: https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#global-libraries for details.

I'm not very familiar with the library, let me know if this PR is correct as is and I can merge it.

types/meyda/index.d.ts
```diff
@@ -0,0 +1,137 @@
...
+    audioContext: AudioContext;
+    source: MeydaSource;
+    bufferSize: number;
+    hopSize?: number;
+    windowingFunction?: MeydaWindowingFunction;
+    featureExtractors?: MeydaAudioFeature | MeydaAudioFeature[];
+    callback?: (features: Partial<MeydaFeaturesObject>) => void;
+}
+
+declare namespace Meyda {
```
armanio123: Instead of this declaration you should use `export as namespace` syntax.

types/meyda/index.d.ts
```diff
@@ -0,0 +1,137 @@
...
+    extract(
+        feature: MeydaAudioFeature | MeydaAudioFeature[],
+        signal: MeydaSignal,
+        previousSignal?: MeydaSignal,
+    ): Partial<MeydaFeaturesObject> | null;
+}
+
+export const main: Meyda;
+
+export default main;
```
armanio123: I did a quick check on the library and I didn't found where you export a default. I think this is incorrect.

types/meyda/index.d.ts
```diff
@@ -0,0 +1,137 @@
...
+    start(features: MeydaAudioFeature[]): void;
+
+    stop(): void;
+
+    setSource(source: MeydaSource): void;
+
+    get(features: MeydaAudioFeature[]): Partial<MeydaFeaturesObject> | null;
+}
+
+interface Meyda {
```
armanio123: Instead of and interface you should just export each property individually so that it's available on the namespace context. e.g:
```js
export const audioContext: AudioContext | null;
...
export function windowing(signal: MeydaSignal, windowname?: MeydaWindowingFunction): MeydaSignal;
...
```

armanio123 (changes requested):

armanio123:
I apologize if I wasn't descriptive enough. This seems like an `umd` module and such there's a way to create the `d.ts` correctly to represent the module and the global scope (You can find the details on the same link as before)

This template might help you creating the `d.ts`: https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.

hughrawlinson:
@armanio123 Sorry for all the back and forth, and thank you for the thorough review! I'm still getting familiar with ts declarations, so I really appreciate you taking the time to help get this right. I *think* all's well now once the PR passes 😄

armanio123:
I'm happy to help. This time it looks good. I'm merging the PR.

armanio123 (approved):

PR #41414, opened 2020-01-06
[meteor] Make check an assertion function

ebroder:
Meteor's check is an assertion function, and Match.test returns a type predicate. TypeScript 3.7 is expressive enough to capture this. This lets us reduce duplication in Meteor code (e.g. needing to keep function parameter types and check calls in sync with each other).

Is this a good idea? I think so - I'd find it useful in my code. Is it worth constraining to the newest version of TypeScript? I'm less sure. My project is already on 3.7 so it's not a big deal to me.

Please fill in this template.

- [:heavy_check_mark:] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [:heavy_check_mark:] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: check API is documented [here](https://docs.meteor.com/api/check.html)

### DIFF
--- a/types/meteor/check.d.ts
+++ b/types/meteor/check.d.ts
@@ -1,24 +1,45 @@
 declare module "meteor/check" {
     module Match {
-        var Any: any;
-        var String: any;
-        var Integer: any;
-        var Boolean: any;
-        var undefined: any;
-        var Object: any;
+        interface Matcher<T> { _meteorCheckMatcherBrand: void }
+        export type Pattern =
+            typeof String |
+            typeof Number |
+            typeof Boolean |
+            typeof Object |
+            typeof Function |
+            (new (...args: any[]) => any) |
+            undefined | null | string | number | boolean |
+            [Pattern] |
+            {[key: string]: Pattern} |
+            Matcher<any>;
+        export type PatternMatch<T extends Pattern> =
+            T extends Matcher<infer U> ? U :
+            T extends typeof String ? string :
+            T extends typeof Number ? number :
+            T extends typeof Boolean ? boolean :
+            T extends typeof Object ? object :
+            T extends typeof Function ? Function :
+            T extends undefined | null | string | number | boolean ? T :
+            T extends new (...args: any[]) => infer U ? U :
+            T extends [Pattern] ? PatternMatch<T[0]>[] :
+            T extends {[key: string]: Pattern} ? {[K in keyof T]: PatternMatch<T[K]>} :
+            unknown;
 
-        function Maybe(pattern: any): boolean;
+        var Any: Matcher<any>;
+        var Integer: Matcher<number>;
 
-        function Optional(pattern: any): boolean;
+        function Maybe<T extends Pattern>(pattern: T): Matcher<PatternMatch<T> | undefined | null>;
 
-        function ObjectIncluding(dico: any): boolean;
+        function Optional<T extends Pattern>(pattern: T): Matcher<PatternMatch<T> | undefined>;
 
-        function OneOf(...patterns: any[]): any;
+        function ObjectIncluding<T extends {[key: string]: Pattern}>(dico: T): Matcher<PatternMatch<T>>;
 
-        function Where(condition: any): any;
+        function OneOf<T extends Pattern>(...patterns: T[]): Matcher<PatternMatch<T>>;
 
-        function test(value: any, pattern: any): boolean;
+        function Where(condition: (val: any) => boolean): Matcher<any>;
+
+        function test<T extends Pattern>(value: any, pattern: T): value is PatternMatch<T>;
     }
 
-    function check(value: any, pattern: any): void;
+    function check<T extends Match.Pattern>(value: any, pattern: T): asserts value is Match.PatternMatch<T>;
 }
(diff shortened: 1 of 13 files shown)


types/meteor/check.d.ts
```diff
@@ -1,24 +1,45 @@
 declare module "meteor/check" {
     module Match {
-        var Any: any;
-        var String: any;
-        var Integer: any;
-        var Boolean: any;
-        var undefined: any;
-        var Object: any;
+        interface Matcher<T> { __impossibleToSatisfy: any; }
```
ebroder: I couldn't figure out a better way to express this - I'm just looking for a type that I can return but nobody else can (otherwise any type trivially satisfies the interface)

types/meteor/check.d.ts
```diff
@@ -1,24 +1,45 @@
 declare module "meteor/check" {
     module Match {
-        var Any: any;
-        var String: any;
-        var Integer: any;
-        var Boolean: any;
-        var undefined: any;
-        var Object: any;
```
ebroder: Most of these constants aren't actually defined on `Match`, so having them declared here is a bug.

perbergland:
Personally, I think moving to TS 3.7 is fine given that both adornis:typescript and meteor 1.8.3 TS use that version.

(And I like the check improvements as well - I'm looking for a chai (expect) version of them with TS support for unit tests)

andrei-markeev (changes requested):
we will have to use 3.7 sooner or later anyways, so no problem here. guys who haven't migrated to it yet can use older versions of typings.

small thing with the Matcher type, otherwise a very good change, thanks a lot! I actually want to use it right away 👍😄

alesn:
I agree on moving to TypeScript 3.7. Typescript itself - whatever version you like - can be used with Meteor. That's what I've been using before the Typescript compiler plugin for Meteor was made available (and that's what I'm still using because I never received a satisfactory response on what actual advantages using the limited Meteor TypeScript compiler plugin - rather than using 'tsc' directly - should bring). So TS 3.7.3 can be used from the latest Meteor itself using the Meteor compiler plugin or TS 3.7.4 (or any other version) can be used when using TS directly and TS brings useful improvements in every new version, so I am for moving to TS 3.7 (or 3.7.3, which is bundled in the latest Meteor compiler plugin).
One thing though, the version of TypeScript required by DefinitelyTyped definitions is specified in index.d.ts, so it should be updated to match the actually required TS version

andrei-markeev:
> the version of TypeScript required by DefinitelyTyped definitions is specified in index.d.ts, so it should be updated

he did update that, check the diff

types/meteor/check.d.ts
ebroder: Aha, "type brand" was the magic phrase I wasn't familiar with. Changed to match more common styles based on some looking around.

andrei-markeev:
Hey!

Please roll back the "Minimum" thing. Just "Typescript Version" is already enforcing the version: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#i-want-to-use-features-from-typescript-31-or-above

Run the tests & lint locally before pushing so that you know you didn't ruin the build. Remember we get email notifications about those pushes. No need to hurry 😉

Also, force-pushes are generally not the best idea after your PR was already reviewed. A single additional commit very clearly shows what have changed since last review. When force-pushing, you basically force reviewer to review the whole thing again from scratch. Also, old links no longer work since they point to no longer existing commits...

ebroder:
Sorry about the pushes, I'll stack new commits going forward. I added the minimum thing because the build was failing without it. It seems like a new change to dtslint that just got pushed: https://github.com/microsoft/dtslint/pull/265. The failure reproduces locally for me if I upgrade dtslint.

andrei-markeev:
"Minimum" - very interesting. Seems like a very recent change. Ok let's see if Travis uses same dtslint version and it actually builds :)

> Sorry about the pushes

Np 👍

I also had this habit of force pushing some time ago. Nobody ever told me anything. And then there was a guy in our team at work who would always force push... Imagine 50 files changed, and he force-pushes, and you just don't have any idea anymore what actual changes he made.

ebroder:
I think I caught a bug in that dtslint change - it was intended to accept the old syntax but doesn't. I'm happy to roll back to the old syntax once that gets fixed, but Travis definitely seems happy with the new syntax too (which I imagine will still be supported).

andrei-markeev:
> A definition owner has approved this PR

something strange is definitely going on 😄

andrei-markeev (approved):
Note for the maintainer: the bot somehow thought that definition owner has approved before anyone has approved, see above 🤔

ebroder:
Thanks for the :eyes:!

armanio123:
@andrei-markeev thanks for mentioning the bot behavior. We'll take a look, in the meanwhile I'm merging this PR.

afrokick:
@ebroder  Does it work for you with this example?

```
Match.OneOf(String, [String]);
```

The error occurred: `Argument of type 'StringConstructor[]' is not assignable to parameter of type 'StringConstructor`

ebroder:
@afrokick Good catch, there was a bug there. I've opened #44561 with a fix. I added a corresponding test from the docs, but feel free to confirm if that fixes the issue for you.

PR #41417, opened 2020-01-06
types/react-stripe-elements: simplify StripeProps

hirochachacha:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/stripe/react-stripe-elements/blob/66a2a25b80e0af31109a4fa018a407c7cec10e98/src/components/inject.js#L98-L108

### DIFF
--- a/types/react-stripe-elements/index.d.ts
+++ b/types/react-stripe-elements/index.d.ts
@@ -12,7 +12,7 @@
 //                 Hiroshi Ioka <https://github.com/hirochachacha>
 //                 Austin Turner <https://github.com/paustint>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.5
 
 /// <reference types="stripe-v3" />
 import * as React from 'react';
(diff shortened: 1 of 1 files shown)


armanio123:
LGTM, but since this is a popular package, I'll leave a couple of days open if any owners would like to jump in for the review; otherwise I'll merge it as-is.

types/react-stripe-elements/index.d.ts
```diff
@@ -12,7 +12,7 @@
 //                 Hiroshi Ioka <https://github.com/hirochachacha>
 //                 Austin Turner <https://github.com/paustint>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.8
+// TypeScript Version: 3.5
```
armanio123: This is probably not needed. I don't think you're using any 3.5 specific features.

armanio123 (changes requested):

types/react-stripe-elements/index.d.ts
hirochachacha: Thank you for the review.

Omit was introduced since 3.5.
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-5.html

That being said, we can define Omit by ourselves.
```
type Omit<T, K> = Pick<T, Exclude<keyof T, K>>;
```

Either is fine with me.  If you think keeping the version is beneficial, I'll follow the advice.

armanio123: I completely missed the `Omit`. In that case 3.5 makes sense. I think bumping up the version it's the best approach. Make note that some users might decide to lock to a specific version of the `d.ts` until they move to a newer ts version.

armanio123 (approved):

PR #41418, opened 2020-01-06
feat(html2canvas): delete @types/html2canvas

peterblazejewicz:
Starging from version 1.0.0-rc2 `html2cavans` now ships with its own type
definitions:
https://github.com/niklasvh/html2canvas/releases/tag/v1.0.0-rc.2
Recent ones:
https://unpkg.com/html2canvas@1.0.0-rc.5/dist/types/index.d.ts

Thanks!

/cc @rwhepburn @tan9 @sschocke @Ristaaf


peterblazejewicz:
It seems the `types-publisher` does not work with prerelease versions.
Should I amend the PR to shorten the version tag to `1.0.0` (which is non-prerelease)?
https://github.com/microsoft/types-publisher/blob/master/src/lib/versions.ts#L97

PR #41421, opened 2020-01-06
Adding TVEventHandler support to @types/react-native

ivarchand:
As TVEventHandler definition is not available in @types/react-native, unable to use typescript to build react-native TV apps since the metro-bundler is broken.
So, Changes are added to @types/react-native to make TVEventHandler support to develop react-native based TV apps

Thanks

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -163,7 +163,7 @@ interface EmitterSubscription extends EventSubscription {
 }
 
 interface DEPRECATED_RCTExport<T> {
-	getConstants(): Function;
+	getConstants?: () => {};
 }
 
 interface TurboModule extends DEPRECATED_RCTExport<void> {}
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -182,7 +182,7 @@ interface NativeTVNavigationEventEmitter {
 }
  
 declare class TVEventHandler {
-	new (): TVEventHandler;
+	constructor();
     enable(component: any, callback: Function): void;
     disable(): void;
 }
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -174,7 +174,7 @@ interface Spec extends TurboModule {
 }
 
 declare class TurboModuleRegistry {
-	get<TurboModule>(name: string): (legacyModule: any) => TurboModule
+	get<T extends TurboModule>(name: string): T | null | undefined
 }
 
 interface NativeTVNavigationEventEmitter {
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -23,8 +23,6 @@
 //                 Mike Martin <https://github.com/mcmar>
 //                 Theo Henry de Villeneuve <https://github.com/theohdv>
 //                 Eli White <https://github.com/TheSavior>
-
-
 //                 Romain Faust <https://github.com/romain-faust>
 //                 Be Birchall <https://github.com/bebebebebe>
 //                 Jesse Katsumata <https://github.com/Naturalclar>
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -23,6 +23,8 @@
 //                 Mike Martin <https://github.com/mcmar>
 //                 Theo Henry de Villeneuve <https://github.com/theohdv>
 //                 Eli White <https://github.com/TheSavior>
+
+
 //                 Romain Faust <https://github.com/romain-faust>
 //                 Be Birchall <https://github.com/bebebebebe>
 //                 Jesse Katsumata <https://github.com/Naturalclar>
@@ -160,6 +162,31 @@ interface EmitterSubscription extends EventSubscription {
      * for removing the subscription lies with the EventEmitter.
      */
     remove(): void;
+}
+
+interface DEPRECATED_RCTExport<T> {
+	getConstants(): Function;
+}
+
+interface TurboModule extends DEPRECATED_RCTExport<void> {}
+
+interface Spec extends TurboModule {
+    addListener(eventName: string): void;
+    removeListeners(count: number): void;
+}
+
+declare class TurboModuleRegistry {
+	get<TurboModule>(name: string): (legacyModule: any) => TurboModule
+}
+
+interface NativeTVNavigationEventEmitter {
+	addListener(eventName: string, callback: Function): NativeEventEmitter;
+}
+ 
+declare class TVEventHandler {
+	new (): TVEventHandler;
+    enable(component: any, callback: Function): void;
+    disable(): void;
 }
 
 interface EventEmitterListener {
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -183,7 +183,7 @@ interface NativeTVNavigationEventEmitter {
  
 declare class TVEventHandler {
 	constructor();
-    enable(component: any, callback: Function): void;
+    enable(component: any | null | undefined, callback: (component: any | null | undefined, data: unknown) => void): void;
     disable(): void;
 }
 


types/react-native/index.d.ts
```diff
@@ -23,6 +23,8 @@
 //                 Mike Martin <https://github.com/mcmar>
 //                 Theo Henry de Villeneuve <https://github.com/theohdv>
 //                 Eli White <https://github.com/TheSavior>
+
+
```
alloy: This is causing CI to fail.

types/react-native/index.d.ts
```diff
@@ -162,6 +164,31 @@ interface EmitterSubscription extends EventSubscription {
...
+declare class TurboModuleRegistry {
+	get<TurboModule>(name: string): (legacyModule: any) => TurboModule
+}
+
+interface NativeTVNavigationEventEmitter {
+	addListener(eventName: string, callback: Function): NativeEventEmitter;
+}
+ 
+declare class TVEventHandler {
+	new (): TVEventHandler;
```
alloy: This should be redundant.

alloy (changes requested):
Can you please link to the relevant docs/sources that describe these?

types/react-native/index.d.ts
ravichandran18390: changes are added - 0ec12eb6d3319739d960f1238a5a1495807535fe

types/react-native/index.d.ts
ravichandran18390: removed the white spaces - 933fe8b647b16a9c21b382fa190bc0bbbc182229

ravichandran18390:
> Can you please link to the relevant docs/sources that describe these?

Hi @alloy,

Used Packages and it's version:
"react-native": "0.61.0"
"@types/react-native": "0.60.23"

1. https://facebook.github.io/react-native/docs/building-for-apple-tv
var TVEventHandler = require('TVEventHandler'); - can be possible building the react-native tv app using javascript but not with typescript & it's strict ts-lint rules.

if it's used like const TVEventHandler = require('TVEventHandler'); - causes the metro bundler to be broken so index.bundle.js is not completely loaded - So Unable to run the app

2. Moreover, @types/react-native versions 0.60.23, 0.60.26 & 0.60.27 are compared only some attributes like tvParallaxProperties,  hasTVPreferredFocus  are available along with TouchableOpacity but there was no way to instantiate the TVEventHandler since it is not defined in index.d.ts

So, this solution is attained, after cross-checking the below-mentioned files of "react-native": "0.61.0"
those files are :
node_modules/react-native/Libraries/Components/AppleTV/TVEventHandler.js
node_modules/react-native/Libraries/Components/AppleTV/NativeTVNavigationEventEmitter.js
node_modules/react-native/Libraries/TurboModule/TurboModuleRegistry.js
node_modules/react-native/Libraries/TurboModule/RCTExport.js

Now, declarations are added to index.d.ts of @types/react-native. Please go through the commits which are done after your comments. and approve it

Thanks in Advance

ivarchand (commented):
Please review the changes - 933fe8b647b16a9c21b382fa190bc0bbbc182229

ivarchand (commented):
Please review the changes on commit - 0ec12eb6d3319739d960f1238a5a1495807535fe

types/react-native/index.d.ts
```diff
@@ -162,6 +162,31 @@ interface EmitterSubscription extends EventSubscription {
     remove(): void;
 }
 
+interface DEPRECATED_RCTExport<T> {
+	getConstants(): Function;
```
alloy: [The upstream Flow type](https://github.com/facebook/react-native/blob/v0.61.5/Libraries/TurboModule/RCTExport.js#L33) states that this property should be optional and is a method that returns an object:

```suggestion
	getConstants?: () => {};
```

types/react-native/index.d.ts
```diff
@@ -162,6 +162,31 @@ interface EmitterSubscription extends EventSubscription {
...
+
+interface TurboModule extends DEPRECATED_RCTExport<void> {}
+
+interface Spec extends TurboModule {
+    addListener(eventName: string): void;
+    removeListeners(count: number): void;
+}
+
+declare class TurboModuleRegistry {
+	get<TurboModule>(name: string): (legacyModule: any) => TurboModule
```
alloy: [The upstream Flow type](https://github.com/facebook/react-native/blob/v0.61.5/Libraries/TurboModule/TurboModuleRegistry.js#L19) states that the return type should _extend_ the `TurboModule` interface, in your case you simply _named_ the type parameter `TurboModule`. Also, I’m not sure where your returned function `(legacyModule: any) => TurboModule` comes from, I don’t see that in the source.

```suggestion
	get<T extends TurboModule>(name: string): T | null | undefined
```

types/react-native/index.d.ts
```diff
@@ -162,6 +162,31 @@ interface EmitterSubscription extends EventSubscription {
...
+}
+
+interface TurboModule extends DEPRECATED_RCTExport<void> {}
+
+interface Spec extends TurboModule {
+    addListener(eventName: string): void;
+    removeListeners(count: number): void;
+}
+
+declare class TurboModuleRegistry {
```
alloy: `TurboModuleRegistry` isn’t a class, but rather a module. Where is the export you’re using? Or are you importing from `Libraries/Components/AppleTV/NativeTVNavigationEventEmitter`?

types/react-native/index.d.ts
```diff
@@ -162,6 +162,31 @@ interface EmitterSubscription extends EventSubscription {
...
+	get<TurboModule>(name: string): (legacyModule: any) => TurboModule
+}
+
+interface NativeTVNavigationEventEmitter {
+	addListener(eventName: string, callback: Function): NativeEventEmitter;
+}
```
alloy: Where is this export defined?

alloy (commented):
Cool, thanks 👍 For future reference, it would be best to link to the exact files you used to base your work on to lower the barrier for people to review your work; e.g. https://github.com/facebook/react-native/blob/v0.61.5/Libraries/Components/AppleTV/TVEventHandler.js (Note that I’m linking to 0.61.5 instead, as that’s the latest available version.)

Unfortunately there’s still not enough context for me to fully review this well. Could you please provide test coverage that shows exactly how you use this at runtime? (Create a new file at `types/react-native/test/AppleTV.tsx`)

types/react-native/index.d.ts
```diff
@@ -162,6 +162,31 @@ interface EmitterSubscription extends EventSubscription {
...
+ 
+declare class TVEventHandler {
+	constructor();
```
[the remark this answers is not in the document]
ravichandran18390: Yep! I also felt the same- not necessary. As it is zero - parameterized constructor, it's not required. when we try to instantiate using "new", automatically an instance will be created.

types/react-native/index.d.ts
ravichandran18390: As there is no export defined for it specifically, no need for this interface it seems.

types/react-native/index.d.ts
ravichandran18390: yes, it's a module & I'm trying to import it from the Libraries/Components/AppleTV/NativeTVNavigationEventEmitter

alloy: Right, so we can’t have that in this root file, it needs to be in a file that follows the same package dir structure, otherwise this won’t work. Take a look at `relay-runtime` for an example of this.

ravichandran18390: Hi @alloy,
you meant that it has to be split out from index.d.ts / root file and keep it inside a lib folder?

alloy: Correct, otherwise the types won’t be picked up by TypeScript when you import from that file directly. This is also why I asked for test coverage, because you will see that if you add test real-world coverage, it should fail as-is.

alloy:
@ravichandran18390 Are you planning any follow-up to this?

PR #41424, opened 2020-01-06
fixed types for mapbox/wellknown

davide-scalzo:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: Wellknown should parse / stringify GeoJSON Geometry, link to GeoJSON spec here https://tools.ietf.org/html/rfc7946#section-3.1

### DIFF
--- a/types/wellknown/index.d.ts
+++ b/types/wellknown/index.d.ts
@@ -1,7 +1,32 @@
 // Type definitions for wellknown 0.5
 // Project: https://github.com/mapbox/wellknown#readme
-// Definitions by: Yair Tawil <https://github.com/yairtawil>
+// Definitions by: Davide Scalzo <https://github.com/davodesign84>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export function parse(input: string): {};
-export function stringify(gj: {}): string;
+export type GeoJSONPosition = [number, number] | [number, number, number];
+export interface Geometry<T, C> {
+    type: T;
+    coordinates: C;
+}
+
+export type GeoJSONPoint = Geometry<'Point', GeoJSONPosition>;
+export type GeoJSONMultiPoint = Geometry<'MultiPoint', GeoJSONPosition[]>;
+
+export type GeoJSONLineString = Geometry<'LineString', GeoJSONPosition[]>;
+
+export type GeoJSONMultiLineString = Geometry<'MultiLineString', GeoJSONPosition[][]>;
+
+export type GeoJSONPolygon = Geometry<'Polygon', GeoJSONPosition[][]>;
+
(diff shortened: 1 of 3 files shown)


types/wellknown/index.d.ts
```diff
@@ -1,7 +1,32 @@
 // Type definitions for wellknown 0.5
 // Project: https://github.com/mapbox/wellknown#readme
-// Definitions by: Yair Tawil <https://github.com/yairtawil>
+// Definitions by: Davide Scalzo <https://github.com/davodesign84>
```
peterblazejewicz: Is that indended change? Would adding a new author on the next line suffice?

types/wellknown/index.d.ts
```diff
@@ -1,7 +1,32 @@
...
+
+export type GeoJSONGeometry =
+    | GeoJSONPoint
+    | GeoJSONMultiPoint
+    | GeoJSONLineString
+    | GeoJSONMultiLineString
+    | GeoJSONPolygon
+    | GeoJSONMultiPolygon;
+
+export function parse(input: string): GeoJSONGeometry;
```
peterblazejewicz: So the documentation states it also returns `null`, when parse fails:
https://github.com/mapbox/wellknown#parsewkt

types/wellknown/wellknown-tests.ts
```diff
@@ -1,6 +1,6 @@
 import * as wellknown from 'wellknown';
 wellknown.parse("POINT(1 2)");
```
peterblazejewicz: This one should be probably updated to:
```ts
wellknown.parse("POINT(1 2)"); // $ExpectType wellknown.GeoJSONGeometry | null
```

types/wellknown/index.d.ts
davodesign84: yup, good point

types/wellknown/index.d.ts
davodesign84: ah great spot!

types/wellknown/wellknown-tests.ts
davodesign84: For some reason this fails with
```
ERROR: 4:1  expect  TypeScript@next expected type to be:
  GeoJSONGeometry | null
got:
  GeoJSONPoint | GeoJSONMultiPoint | GeoJSONLineString | GeoJSONMultiLineString | GeoJSONPolygon | GeoJSONMultiPolygon | null
```

davodesign84: Turns out it seems like $ExpectType her only recognizes a single type..

peterblazejewicz: Sorry for the fuss here, I was wrong (looked probably into wrong version). You're correctly on this `GeoJSONGeometry` is OK

peterblazejewicz (commented):
/cc @yairtawil

PR #41427, opened 2020-01-06
@types/through: Add missing queue method to through.ThroughStream

carlegbert:
  (kind of- updating the tsconfig makes the tests catch the issue that I'm fixing)

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://github.com/dominictarr/through/blob/master/index.js#L40>

The `ThroughStream` object uses `queue` as an alias for `push`; both do the same thing (see link to source code above). While the `push` property is inherited from `stream.Transform`, `queue` is not.

Looking at the source code, I also notice that `stream.Transform.push` is a function that takes a `chunk` and an optional `encoding` argument, while `through.ThroughStream.(push | queue)` takes only the chunk. I wonder if the definition for `push` should be overridden to omit that argument?

As per https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#some-packages-have-no-tslintjson-and-some-tsconfigjson-are-missing-noimplicitany-true-noimplicitthis-true-or-strictnullchecks-true, I updated the tsconfig (this update actually makes the test fail to compile without my change).

### DIFF
--- a/types/through/index.d.ts
+++ b/types/through/index.d.ts
@@ -17,7 +17,7 @@ declare function through(write?: (data: any) => void,
 declare namespace through {
     export interface ThroughStream extends stream.Transform {
         autoDestroy: boolean;
-        queue: (chunk: any, encoding?: string | undefined) => any;
+        queue: (chunk: any) => any;
     }
 }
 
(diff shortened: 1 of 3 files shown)


elibarzilay:
Ping @AndrewGaspar ?

PR #41436, opened 2020-01-07
Add types for leaflet.browser.print

ihrigb:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/leaflet.browser.print/index.d.ts
+++ b/types/leaflet.browser.print/index.d.ts
@@ -1,0 +1,95 @@
+// Type definitions for leaflet.browser.print 1.0
+// Project: https://github.com/Igor-Vladyka/leaflet.browser.print
+// Definitions by: Benjamin Ihrig <https://github.com/ihrigb>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.6
+
+import * as leaflet from 'leaflet';
+
+declare module 'leaflet' {
+    namespace control {
+        interface BrowserPrintControlOptions extends ControlOptions {
+            title?: string;
+            documentTitle?: string;
+            printLayer?: L.TileLayer;
+            printModes?: Control.BrowserPrint.Mode[];
+            closePopupsOnPrint?: boolean;
+            contentSelector?: string;
+            pageSelector?: string;
+            manualMode?: boolean;
+            customPrintStyle?: L.PolylineOptions;
+        }
+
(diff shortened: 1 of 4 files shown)


ihrigb:
This already was an open PR (#41209), but I was not able to react, as I was on leave. So, I will reopen this here.

@andrewbranch you commented that I may remove the line `// TypeScript Version: 3.6`. If I do so, `npm run lint leaflet.browser.print` tells me "Compile error in typescript@3.5 but not in typescript@3.6" and that I should fix it with a comment.

types/leaflet.browser.print/index.d.ts
```diff
@@ -0,0 +1,95 @@
...
+            type ModeType = LandscapeModeType | PortraitModeType | AutoModeType | CustomModeType;
+
+            namespace Mode {
+                type Landscape = LandscapeModeType;
+                type Portrait = PortraitModeType;
+                type Auto = AutoModeType;
+                type Custom = CustomModeType;
+            }
+
+            function Mode(mode: ModeType, title?: string, pageSize?: string): any;
```
DanielRosenwasser: Should this return `Mode`?

types/leaflet.browser.print/index.d.ts
```diff
@@ -0,0 +1,95 @@
...
+        }
+    }
+
+    namespace Control {
+        interface BrowserPrint {
+            addTo(map: L.Map): HTMLDivElement;
+        }
+
+        namespace BrowserPrint {
+            class Mode { }
```
DanielRosenwasser: Is this really a `class`? And why is it empty? @sandersn, is there a reason why dtslint isn't getting an error?

If you switch it to an `interface`, then this works in 3.5 and earlier. Earlier versions will freak out because they think that a `class` and a `function` can never merge.

DanielRosenwasser (changes requested):

types/leaflet.browser.print/index.d.ts
ihrigb: Not sure. This was not relevant for my use case, so I left it out. I may improve this on a following PR.

types/leaflet.browser.print/index.d.ts
ihrigb: I believe it is a class in the original code. It is left empty, because I am not working with it in my code, but may fill it once I use it.

ihrigb:
@DanielRosenwasser can you please help reviewing?

PR #41458, opened 2020-01-07
[react-modal] Use correct type for style objects 

FiberJW:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/reactjs/react-modal/blob/master/src/components/ModalPortal.js#L351

The type for React `style` props should be `React.CSSProperties`, but this library is written in JS which can't express that, so the existing typings for this library mimic this behavior. This PR fixes that and simultaneously improves the DX for people using these typings by providing better autocomplete and checks to make writing incorrect style properties harder.

Thanks for taking the time to check this out!

<p align="center">

</p>

### DIFF
--- a/types/react-modal/react-modal-tests.tsx
+++ b/types/react-modal/react-modal-tests.tsx
@@ -1,86 +1,86 @@
-import * as React from 'react';
+import * as React from "react";
 import ReactModal = require('react-modal');
 
 // tests for static method
 // string
-ReactModal.setAppElement('#main');
+ReactModal.setAppElement("#main");
 // HTMLElement
-ReactModal.setAppElement(document.getElementById('#main'));
+ReactModal.setAppElement(document.getElementById("#main"));
 
 class ExampleOfUsingReactModal extends React.Component {
-    contentRef: HTMLDivElement;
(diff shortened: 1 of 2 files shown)


types/react-modal/react-modal-tests.tsx
```diff
@@ -14,7 +14,7 @@ class ExampleOfUsingReactModal extends React.Component {
     const onAfterOpenFn = () => { };
     const onAfterCloseFn = () => { };
     const onRequestCloseFn = (event: React.MouseEvent | React.KeyboardEvent) => { };
-    const customStyle = {
+    const customStyle: { overlay: React.CSSProperties, content: React.CSSProperties } = {
```
peterblazejewicz: Can we change this, for readability:
```ts
const customStyle: ReactModal.Styles = {
```
I doubt I'd ever redeclare styles shape instead of importing declared type

FiberJW: Thanks for catching this! Resolved in latest commit.

peterblazejewicz (approved):
LGTM
@FiberJW thanks!

PR #41459, opened 2020-01-07
[jsonwebtoken] Add Algorithm type

nflaig:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/auth0/node-jsonwebtoken/blob/master/sign.js#L12

### DIFF
--- a/types/jsonwebtoken/index.d.ts
+++ b/types/jsonwebtoken/index.d.ts
@@ -6,7 +6,8 @@
 //                 Veli-Pekka Kestilä <https://github.com/vpk>,
 //                 Daniel Parker <https://github.com/rlgod>,
 //                 Kjell Dießel <https://github.com/kettil>,
-//                 Robert Gajda <https://github.com/RunAge>
+//                 Robert Gajda <https://github.com/RunAge>,
+//                 Nico Flaig <https://github.com/nflaig>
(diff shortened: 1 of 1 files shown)


RunAge (approved):

c100k:
Hello 👋🏽,

Actually this is a "breaking" change. Narrowing the type from `string` to `Algorithm` breaks some code with the following error :

     TypeScript diagnostics (customize using `[jest-config].globals.ts-jest.diagnostics` option):
    main.ts:13:9 - error TS2322: Type 'string' is not assignable to type 'Algorithm'.

    13         algorithm: externalConfig.algorithm,

It happens for example when the value comes from an external config typed with string. Not a big deal, but it is not safe to let dependabot update safely even if it's just a semver patch ;)

nflaig:
@c100k You are right, thanks for noticing. Any suggestion how this could be resolved?

c100k:
Except expliciting it clearly in the changelog with a warning, there was actually not much to do on your side I guess :)

Because since it is a breaking change, the new version should have been `9.0.0` according to semver. But, on the other side, that would make it unclear for anyone since [jsonwebtoken](https://www.npmjs.com/package/jsonwebtoken) is at `8.5.1`. I think we hit a limit of `lib` vs `@types/lib` versioning.

PR #41463, opened 2020-01-08
LokiJS DynamicView.branchResultset: parameter transform is optional

Kaesual:
In DefinitelyTyped, the parameter **transform** is (incorrectly) marked as required. I fixed this.

See also:
https://github.com/techfort/LokiJS/blob/v1.5.3/src/lokijs.js#L3971 (since npmjs.com states the current types are for LokiJS version 1.5.3, although this is correct for the latest version 1.5.8, too)
http://techfort.github.io/LokiJS/DynamicView.html (JSDoc)


TeamworkGuy2 (approved):
LGTM

PR #41480, opened 2020-01-09
[conventional-changelog] Added type definitions

JasonHK:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/conventional-changelog-core/conventional-changelog-core-tests.ts
+++ b/types/conventional-changelog-core/conventional-changelog-core-tests.ts
@@ -1,0 +1,19 @@
+/* tslint:disable:no-mergeable-namespace no-namespace */
+"use strict";
+
+import conventionalChangelogCore from "conventional-changelog-core";
+
+namespace Test {
+    declare const context: conventionalChangelogCore.Context;
+    declare const gitRawCommitsOpts: conventionalChangelogCore.GitRawCommitsOptions;
+    declare const options: conventionalChangelogCore.Options;
+    declare const parserOpts: conventionalChangelogCore.ParserOptions;
+    declare const writerOpts: conventionalChangelogCore.WriterOptions;
+
+    conventionalChangelogCore(); // $ExpectType Readable
+    conventionalChangelogCore(options); // $ExpectType Readable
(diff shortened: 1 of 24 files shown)


types/conventional-changelog-core/conventional-changelog-core-tests.ts
```diff
@@ -0,0 +1,19 @@
+/* tslint:disable:no-mergeable-namespace no-namespace */
+"use strict";
+
+import conventionalChangelogCore from "conventional-changelog-core";
+
+namespace Test {
```
armanio123: Is there a reason for having a namespace here? You should be able to test without it. 

Same question for the other test files.

types/conventional-changelog-core/conventional-changelog-core-tests.ts
```diff
@@ -0,0 +1,19 @@
+/* tslint:disable:no-mergeable-namespace no-namespace */
+"use strict";
```
armanio123: This line is unnecessary. Please remove it.

armanio123 (changes requested):

types/conventional-changelog-core/conventional-changelog-core-tests.ts
JasonHK: I use this approach since I want to isolate the parts of the definition to make the tests easier to identify.

For example in [/types/conventional-recommended-bump/conventional-recommended-bump-tests.ts#L34](https://github.com/JasonHK/DefinitelyTyped/blob/conventional-changelog/types/conventional-recommended-bump/conventional-recommended-bump-tests.ts#L34):

```typescript
namespace Test.Options {
    ...
}
```

The namespace `Test.Options` means I was testing `conventionalRecommendedBump.Options` inside of it.

If this is acceptable should I leave a comment inside the files to prevent confusion by others?

armanio123: It's not unacceptable if it's easier for you to read. It's a little bit confusing as it's not the intended purpose of namespaces. 

I'm ok with it, I'll merge it.

armanio123 (approved):

PR #41484, opened 2020-01-09
[parse] Stricter constructors, and let Cloud.define return Promises

rdhelms:
Please fill in this template.

-----

The stricter constructor types are a result of discussion with @RaschidJFR and @LinusU from https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41051

The Promise return type for `Parse.Cloud.define` is a followup for https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41457 which takes into account the fact that the return types of Cloud functions can be wrapped in a Promise.

### DIFF
--- a/types/parse/index.d.ts
+++ b/types/parse/index.d.ts
@@ -825,7 +825,7 @@ namespace Parse {
     }
     interface UserConstructor extends ObjectStatic {
         new <T extends Attributes>(attributes: T): User<T>;
-        new(): User;
+        new(attributes?: Attributes): User;
 
         allowCustomUserClass(isAllowed: boolean): void;
         become(sessionToken: string, options?: UseMasterKeyOption): Promise<User>;
--- a/types/parse/index.d.ts
+++ b/types/parse/index.d.ts
@@ -461,7 +461,7 @@ namespace Parse {
         unPinAllWithName(name: string, objects: Object[]): Promise<void>;
     }
     interface ObjectConstructor extends ObjectStatic {
-        new <T extends Attributes>(className: string, attributes?: T, options?: any): Object<T>;
+        new <T extends Attributes>(className: string, attributes: T, options?: any): Object<T>;
         new(className?: string, attributes?: Attributes, options?: any): Object;
     }
     const Object: ObjectConstructor;
@@ -825,7 +825,7 @@ namespace Parse {
     }
     interface UserConstructor extends ObjectStatic {
         new <T extends Attributes>(attributes: T): User<T>;
-        new(attributes?: Attributes): User;
+        new(): User;
 
         allowCustomUserClass(isAllowed: boolean): void;
         become(sessionToken: string, options?: UseMasterKeyOption): Promise<User>;
@@ -1084,7 +1084,7 @@ namespace Parse {
             param: { [P in keyof Parameters<T>[0]]: Parameters<T>[0][P] }
         ) => any>(
             name: string,
-            func: (request: FunctionRequest<Parameters<T>[0]>) => ReturnType<T>
+            func: (request: FunctionRequest<Parameters<T>[0]>) => Promise<ReturnType<T>> | ReturnType<T>
         ): void;
         /**
          * Gets data for the current set of cloud jobs.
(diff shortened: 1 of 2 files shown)


rdhelms:
@RaschidJFR @LinusU let me know if you have additional concerns or observations to add on to what we discussed earlier.

Also, I made another attempt at the idea to "make attributes optional in the constructor if all the type parameter's properties are optional"...and I came up with an example that could technically work, but I think the error message is too confusing when the user fails to provide attributes when there ARE required properties. So I didn't include this change, but wanted to provide this boiled-down demo as a reference:
```ts
declare namespace Parse {
    type EmptyObject<T> = {} extends T ? {} : never

    interface ObjectConstructor {
        new <T extends EmptyObject<T>>(params?: T): void
        new <T extends object>(params: T): void
    }
    const Object: ObjectConstructor
}

new Parse.Object<{ optional?: number }>() // Optional case - works fine
new Parse.Object<{ optional?: number }>({ optional: 123 }) // Optional case - works fine

new Parse.Object<{ required: string }>({ required: 'hello' }) // Required case - works fine
new Parse.Object<{ required: string }>() // Required case - errors, as expected...although the error is extremely confusing
// ^Confusing error: Type '{ required: string; }' does not satisfy the constraint 'never'.

```

types/parse/index.d.ts
```diff
@@ -825,7 +825,7 @@ namespace Parse {
     }
     interface UserConstructor extends ObjectStatic {
         new <T extends Attributes>(attributes: T): User<T>;
-        new(attributes?: Attributes): User;
+        new(): User;
```
RaschidJFR: Why not letting create a User with attributes?

RaschidJFR (changes requested):

types/parse/index.d.ts
RaschidJFR: This is the most natural way of using the SDK without using type params, I think that should be up to the user to decide.

rdhelms: So perhaps I'm misinterpreting what you're asking - but the intent is for the overload on line 827 to cover the case where a User is created with initial attributes. In this way `new Parse.User({ attr: 'thing' })` and `new Parse.User<{ attr: string }>({ attr: 'thing' })` are both still allowed. The purpose of line 828 is to prevent `new User<{ attr: string }>()`, since that would create an object with a bad state (similar to the discussion for Parse.Object, TS would think `.get('attr')` was a string when it would actually be undefined).

These are the test cases that summarize those instances:
```ts
function testUser() {
    function testConstructor() {
        // $ExpectType User<Attributes>
        new Parse.User();

        // $ExpectType User<{ example: number; }>
        new Parse.User({ example: 100 });

        // $ExpectError
        new Parse.User<{ example: number }>();

        // $ExpectError
        new Parse.User<{ example: number }>({ example: 'hello' });
    }
}
```

rdhelms: So, after some further thought, I think I realize what you were noticing. The irony here is that all the test cases that we care about work successfully whether `attributes?: Attributes` is a param or not. My instinct was not to include it purely for that reason - that it wasn't necessary. But I did notice that there is a subtle advantage to having it present, which is that it improves the intellisense for IDEs.

In other words, without the param present, someone might type `new Parse.User()` and in that moment, the second overload is being used and so hovering over `User` would not how `attributes` as a possible param. Of course, if they were to go ahead and pass a param anyway, then the first overload would be used, and hovering would correctly show the `attributes` param.

But *with* the param present in the second overload, hovering over `new Parse.User()` would show `attributes` as a possible param, which would be more helpful in guiding the user. And once they pass a parameter, the first overload will be used again as before.

So with that, I think that you're right that there is a subtle advantage there, so I added it back.

RaschidJFR: Hey, actually I was about to comment and tell you that I got your point 😅. I missed the test this time as I was still chewing our previous conversation:
```
// $ExpectType User<{ example: number; }>
new Parse.User({ example: 100 });
```
but it actually works in any case. I'm glad at least a small advantage came out from this :)

RaschidJFR (approved):

RaschidJFR (approved):

PR #41485, opened 2020-01-09
[rdf-js] Add quad generic to DataFactory

thewilkybarkid:
Adds the quad type to `DataFactory` as a generic, rather than having it on its functions. This allows matching a `DataFactory` up with how it's used, such as https://github.com/DefinitelyTyped/DefinitelyTyped/blob/2eccf3b0aa829ea8570b4682aba395a2bf154bc7/types/rdf-dataset-indexed/index.d.ts#L8

I've tried to update its usage as well.


thewilkybarkid:
In draft as I haven't added tests, but will wait on a green light before going through them.

types/rdf-ext/lib/Dataset.d.ts
```diff
@@ -1,13 +1,13 @@
-import { Dataset } from 'rdf-js';
+import { BaseQuad, Dataset } from 'rdf-js';
 import { PropType } from './_PropType';
 import QuadExt = require('./Quad');
 
-interface DatasetExt extends Dataset {
+interface DatasetExt<Q extends BaseQuad = QuadExt> extends Dataset<Q> {
   readonly length: number;
   toJSON(): Array<ReturnType<PropType<QuadExt, 'toJSON'>>>;
```
thewilkybarkid: Because of this line, does this mean it should be `Q extends QuadExt`?

tpluscode: I think it should be a closed type

```
interface DatasetExt extends Dataset<QuadExt> 
```

rdf-ext will always use its `Ext` types AFAICT

thewilkybarkid: Same for https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1163ead296d84e7a3c80d71e7c81ecbd1a130e9a/types/rdf-ext/lib/DataFactory.d.ts#L42-L43 too?

tpluscode: I should think that it does not make sense to make anything in rdf-ext generic. Those implementations as specific specialisations of the base interface

thewilkybarkid:
Holiday rather than abandoned. 😅

But happy to carry on in #41651.

Issue #40050, opened 2019-10-31
META - Updating minimum TypeScript version


RyanCavanaugh:
We're considering moving the minimum-supported TypeScript version on DT from 2.0 to 2.8.

Reasons for doing this:
 * Reduce friction for using popular 2.8 features like conditional types
 * Substantially faster CI time since we'd be testing fewer back versions
 * Reduce typings churn for people on ancient versions of TS since they probably don't like change in the first place

Potential risks and outcomes:
 * A developer on 2.7 would be unable to use a DT package written today through Automatic Type Acquisition, even though it would very likely work anyway (~60% of current packages are 2.0-compatible)
   * TS developers can try it anyway, and are already in this state to some extent due to new packages starting development at 2.8+ minimum version
   * JS developers have no real reason to not upgrade to a newer JS LS

Data:
 * Telemetry indicates 99.9% of VS Code users are 3.0 or newer
 * Telemetry indicates 97% of VS users are on 3.0 or newer
 * 20% of DT packages have already opted in to 2.8 or newer

Feedback?

notlaforge:
When was `unknown` added? Go big or go home! All the `any` in DT is by far the largest source of "typescript shoulda caught it" bugs in these here parts.

SlurpTheo:
> When was `unknown` added?

[3.0](https://github.com/Microsoft/TypeScript/wiki/Roadmap#30-july-2018)

notlaforge:
ty. So, 99.99%!

johnnyreilly:
I agree we should do this. Let's do this!

JoshuaKGoldberg:
Please, could you instead increase it slowly over time? 2.1 one month, 2.2 the next, and so on?

Each version has some small breaking changes. It'll be easier to understand what troubles this brings by going slowly and methodically at first.

Projects in maintenance mode often don't have the bandwidth to overcome those breaking changes, especially if they're core to infrastructure.

SimonSchick:
Yes please!

notlaforge:
@JoshuaKGoldberg projects in maintenance mode are not likely to be upgrading versions of type definitions month after month, yeah?

johnnyreilly:
@JoshuaKGoldberg do remember that upgrading DT does not remove the historic type definitions that have already been published to npm. `@types` historic versions live forever like Mumm-Ra.

You'll be able to depend upon those regardless of where DT goes versioning-wise.

[...]

PR #41491, opened 2020-01-09
Node: update `global` to include `typeof globalThis`

OliverJAsh:
`window` has type `Window & typeof globalThis`, so I think Node's `global` should do the same.

This allows the following which previously would have been a type error:

```ts
declare global {
    var foo: number;
}
{
    foo;
    global.foo; // this used to be a type error
}
```

This means we don't have to define globals twice, so we can remove repeat definitions inside the `Global` interface, such as `process`.

`globalThis` is only available in recent versions of TS, so I'm not sure how we go about making this change. We can't re-declare `global`, so unfortunately using TS versioned types won't help in this case.

What does everyone think?

---


OliverJAsh:
> @microsoft @DefinitelyTyped @jkomyno @a-tarasyuk @alvis @r3nya @btoueg @BrunoScheufler @smac89 @tellnes @Touffy @DeividasBakanas @eyqs @Flarna @Hannes-Magnusson-CK @KSXGitHub @hoo29 @kjin @ajafff @islishude @mwiktorczyk @mohsen1 @n-e @octo-sniffle @galkin @parambirs @eps1lon @SimonSchick @ThomasdenH @WilcoBakker @wwwy3y3 @ZaneHannanAU @samuela @kuehlein @j-oliveras @bhongy @chyzwar @trivikr @nguymin4 @yoursunny @qwelias 

Please could I get some feedback on this proposal?

Flarna:
Ff types versions don't help I see no chance to get this in soon.
It took a long time to get min ts version from 2.0 to 2.8 in this repo therefore I expect it will take up to two or three years till min version is move to >= 3.4.

see https://github.com/DefinitelyTyped/DefinitelyTyped/issues/40050

PR #41496, opened 2020-01-09
react-dom: When `act` is not async, don't return a Promise-like

whatisaphone:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/react/blob/2e4948a34d4504eb852f7473d3c0315037f75d22/packages/react-dom/src/test-utils/ReactTestUtilsAct.js#L212-L216
  - The package spits out a warning at runtime, but this is TypeScript, so we can fix this problem at compile time instead

---

This fixes the return type of `ReactTestUtils.act`. The return value depends on whether its callback is **sync** or **async**. Presently, the return type for **async** is correct, but the return type for **sync** is incorrect—it returns a `Promise`-like, when it shouldn't. That causes downstream lint errors like this one:

```text
  57:5   error  Promises must be handled appropriately          @typescript-eslint/no-floating-promises
```

If you try to fix those errors, you instead get a runtime warning (from the source code linked above):

```text
  console.error ../../node_modules/react-dom/cjs/react-dom-test-utils.development.js:72
    Warning: Do not await the result of calling act(...) with sync logic, it is not a Promise.
```

This PR fixes all that, by making sure the non-Promise return value is not shaped like a Promise.

### DIFF
--- a/types/react-dom/test-utils/index.d.ts
+++ b/types/react-dom/test-utils/index.d.ts
@@ -287,7 +287,7 @@ export function createRenderer(): ShallowRenderer;
  * @see https://reactjs.org/blog/2019/02/06/react-v16.8.0.html#testing-hooks
  */
 // the "void | undefined" is here to forbid any sneaky "Promise" returns.
-export function act(callback: () => void | undefined): DebugPromiseLike;
+export function act(callback: () => void | undefined): void;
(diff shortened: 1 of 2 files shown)


elibarzilay:
Ping @MartynasZilinskas @theruther4d @Jessidhia

PR #41498, opened 2020-01-09
[mustache] More correct definitions and reduce the amount of `any`

phillipj:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [janl/mustache.js/mustache.mjs](https://github.com/janl/mustache.js/blob/master/mustache.mjs)

<hr>

As a maintainer of mustache.js, I noticed the were was quite a lot of `any`s in the current definitions and other tweaks that makes these definitions more correct in general:

- More strict definition of tags, going from an array of string to an array of only two string items
- Make `Mustache.version | .name` fields `readonly`
- Partials from `any` to `Record<string, string>` or a lookup function
- `originalTemplate` from `any` to `string`

### DIFF
--- a/types/mustache/index.d.ts
+++ b/types/mustache/index.d.ts
@@ -1,6 +1,8 @@
-// Type definitions for Mustache 0.8.4
+// Type definitions for Mustache 3.2.1
 // Project: https://github.com/janl/mustache.js
-// Definitions by: Mark Ashley Bell <https://github.com/markashleybell>, Manuel Thalmann <https://github.com/manuth>
+// Definitions by: Mark Ashley Bell <https://github.com/markashleybell>,
+//                 Manuel Thalmann <https://github.com/manuth>,
+//                 Phillip Johnsen <https://github.com/phillipj>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 /**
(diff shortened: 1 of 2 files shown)


phillipj:
Just realised that these typings are more than welcome to live inside the actual mustache.js repo and its npm package.

That would be preferable in my eyes, as it's easier to remember to update the defs when package changes are made.

Might be good reasons to keep them here that I'm not considering tho? 🤷‍♂️

markashleybell (approved):
All looks good to me!

markashleybell:
@phillipj I agree: in my opinion, it would be nicer to have these distributed with the mustache.js package itself and avoid having to install a separate @types package. 

However, according to the [TypeScript publishing docs](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html), having separate typings is the recommended approach for packages not written in TypeScript. I'm not totally sure why—and they don't go into any more detail—but I guess sticking to the recommendations is advisable.

Disclaimer: I'm by no means an expert here!

phillipj:
@markashleybell thanks for having a look and pointers to the TypeScript docs!

Agreed, let's continue for now with the typedefs here. I'll try to ask around to find the reasons *why* that is actually recommended, until then these changes shouldn't be blocked by that.

Thanks a bunch for creating these typedefs in the first place though, much appreciated 💯

armanio123:
There's pros and cons for both options. Ultimately depends on what the library prefers. 

I personally prefer DefinitelyTyped as it allows the community to easily update definitions without the library maintainers to jump in (in case it goes AWOL and decides not to work anymore on the project). Also, have more trained eyes reviewing and submitting changes.

For what it's worth, if the library is not in TypeScript already there's also another alternative. The TS version 3.7 allows to create declarations by using any JSDoc-annotated code. This might help with improving the accuracy concerns: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#--declaration-and---allowjs

phillipj:
@armanio123 thanks for merging and the added input on where to place typedefs.

Didn't know about the possibilities of creating typedefs our of JSDocs, that's worth considering. Also your point of library authors loosing interest in keeping typedefs up-to-date in a JavaScript project is valid. I imagine that happens as time goes by, where it makes sense that the community can contribute typedefs on the side, without nagging on library authors to be involved.

PR #41511, opened 2020-01-10
dropzone: Add options to dropzone instances, add type for options, add type for options.headers

JereKuusela:
Please fill in this template.

1. Headers are given to XMLHttpRequest.setRequestHeader which accepts strings (https://gitlab.com/meno/dropzone/blob/master/src/dropzone.js#L2026)
2. Dropzone.options is a map from HTMLElement id (string) to DropzoneOptions (https://gitlab.com/meno/dropzone/blob/master/src/dropzone.js#L2304)
3. Options exists as an instance property and is the only way to change options of an existing instance (https://gitlab.com/meno/dropzone/blob/master/src/dropzone.js#L959)

### DIFF
--- a/types/dropzone/dropzone-tests.ts
+++ b/types/dropzone/dropzone-tests.ts
@@ -174,6 +174,10 @@ dropzoneWithOptionsVariations = new Dropzone('.test', {
 const dropzone = new Dropzone('.test');
 
 dropzone.options.clickable = true;
+if (!dropzone.options.headers) {
+	dropzone.options.headers = {};
+}
+dropzone.options.headers.test = 'test';
 dropzone.enable();
 dropzone.disable();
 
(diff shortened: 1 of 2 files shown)


PikachuEXE (changes requested):
Headers: https://www.dropzonejs.com/#config-headers
Seems fine

`Dropzone.option`: https://www.dropzonejs.com/#configuration
Seems fine except the value can also be set to `false`

Dropzone instance option: Can't find anything in doc
Since it does not appear in doc please ask with the project maintainer(s) if that's an expected public instance

JereKuusela:
@PikachuEXE Here is an official example: https://gitlab.com/meno/dropzone/-/wikis/set-url-dynamically

PikachuEXE (approved):
That's very hidden doc :P
Looks good then

PR #41526, opened 2020-01-10
[lazy-brush] Add definitions

ksocha:
Definitions for [lazy-brush](https://github.com/dulnan/lazy-brush)

### DIFF
--- a/types/lazy-brush/index.d.ts
+++ b/types/lazy-brush/index.d.ts
@@ -1,0 +1,131 @@
+// Type definitions for lazy-brush 1.0
+// Project: https://github.com/dulnan/lazy-brush#readme
+// Definitions by: Kamil Socha <https://github.com/ksocha>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export interface Coordinates {
+    x: number;
+
+    y: number;
+}
+
+export class Point implements Coordinates {
+    x: number;
+
+    y: number;
+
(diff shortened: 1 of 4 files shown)


peterblazejewicz (commented):
What about public instance properties? I assume author suggests using access methods, but still, in the sample app these props are used directly as being public:
https://github.com/dulnan/lazy-brush-demo/blob/master/src/classes/Scene.js#L250
https://github.com/dulnan/lazy-brush/blob/master/src/LazyBrush.js#L16-L20

ksocha:
@peterblazejewicz I added instance properties to the definitions.

peterblazejewicz:
@ksocha Kamil, public ones, not private (I guess `_` were to be treated as private). Sorry for not being strict on this, waisting your time :(

ksocha:
@peterblazejewicz All of these props all accessible in vanilla JS, so I think all should be typed. After all `_` prefix is just naming convention, not an access modifier.

peterblazejewicz:
I Ithink I have an understanding of the JS semantics, I've just pointed long-established convention being now replaced by new specs (`# is the new _`)
https://rfrn.org/~shu/2018/05/02/the-semantics-of-all-js-class-elements.html#ex9
Anyway, that's ok and can be amended if required, so  LGTM

peterblazejewicz (approved):
LGTM
@ksocha thanks for contribution!

PR #41532, opened 2020-01-11
Harden highland types

SupernaviX:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: http://highlandjs.org/

### DIFF
--- a/types/highland/index.d.ts
+++ b/types/highland/index.d.ts
@@ -1296,23 +1296,41 @@ declare namespace Highland {
 		each(f: (x: R) => void): Pick<Stream<R>, 'done'>;
 
 		/**
-		 * Pipes a Highland Stream to a [Node Writable Stream](http://nodejs.org/api/stream.html#stream_class_stream_writable)
-		 * (Highland Streams are also Node Writable Streams). This will pull all the
-		 * data from the source Highland Stream and write it to the destination,
-		 * automatically managing flow so that the destination is not overwhelmed
-		 * by a fast source.
+		 * Pipes a Highland Stream to a [Node Writable
+		 * Stream](http://nodejs.org/api/stream.html#stream_class_stream_writable).
+		 * This will pull all the data from the source Highland Stream and write it to
+		 * the destination, automatically managing flow so that the destination is not
+		 * overwhelmed by a fast source.
 		 *
-		 * This function returns the destination so you can chain together pipe calls.
+		 * Users may optionally pass an object that may contain any of these fields:
+		 *
+		 * - `end` - Ends the destination when this stream ends. Default: `true`. This
+		 *   option has no effect if the destination is either `process.stdout` or
+		 *   `process.stderr`. Those two streams are never ended.
+		 *
+		 * Like [Readable#pipe](https://nodejs.org/api/stream.html#stream_readable_pipe_destination_options),
+		 * this function will throw errors if there is no `error` handler installed on
+		 * the stream.
+		 *
+		 * This function returns the destination so you can chain together `pipe` calls.
+		 *
+		 * **NOTE**: While Highland streams created via `_()` and [pipeline](#pipeline)
+		 * support being piped to, it is almost never appropriate to `pipe` from a
+		 * Highland stream to another Highland stream. Those two cases are meant for
+		 * use when piping from *Node* streams. You might be tempted to use `pipe` to
+		 * construct reusable transforms. Do not do it. See [through](#through) for a
+		 * better way.
 		 *
 		 * @id pipe
-		 * @section Streams
-		 * @name Stream.pipe(dest)
+		 * @section Consumption
+		 * @name Stream.pipe(dest, options)
 		 * @param {Writable Stream} dest - the destination to write all data to
+		 * @param {Object} options - (optional) pipe options.
 		 * @api public
 		 */
 		pipe<U>(dest: Stream<U>): Stream<U>;
-		pipe<U>(dest: NodeJS.ReadWriteStream): Stream<U>;
-		pipe(dest: NodeJS.WritableStream): void;
+		pipe<U>(dest: NodeJS.ReadWriteStream, options?: { end?: boolean }): Stream<U>;
+		pipe(dest: NodeJS.WritableStream, options?: { end?: boolean }): Stream<void>;
 
 		/**
 		 * Consumes a single item from the Stream. Unlike consume, this function will
--- a/types/highland/index.d.ts
+++ b/types/highland/index.d.ts
@@ -127,20 +127,20 @@ interface HighlandStatic {
 	 * @api public
 	 */
 	<R>(): Highland.Stream<R>;
-	<R>(xs: Highland.Stream<R>[]): Highland.Stream<R>;
-	<R>(xs: R[]): Highland.Stream<R>;
-	<R>(xs: (push: (err: Error | null, x?: R | Highland.Nil) => void, next: () => void) => void): Highland.Stream<R>;
+	<R>(source: Highland.Stream<R>[]): Highland.Stream<R>;
+	<R>(source: R[]): Highland.Stream<R>;
+	<R>(source: (push: (err: Error | null, x?: R | Highland.Nil) => void, next: () => void) => void): Highland.Stream<R>;
 
-	<R>(xs: Highland.Stream<R>): Highland.Stream<R>;
-	<R>(xs: NodeJS.ReadableStream, onFinished?: Highland.OnFinished): Highland.Stream<R>;
-	<R>(eventName: string, eventEmitter: NodeJS.EventEmitter, mappingHint?: Highland.MappingHint): Highland.Stream<R>;
+	<R>(source: Highland.Stream<R>): Highland.Stream<R>;
+	<R>(source: NodeJS.ReadableStream, onFinished?: Highland.OnFinished): Highland.Stream<R>;
+	<R>(source: string, eventEmitter: NodeJS.EventEmitter, mappingHint?: Highland.MappingHint): Highland.Stream<R>;
 
 	// moar (promise for everything?)
-	<R>(xs: PromiseLike<Highland.Stream<R>>): Highland.Stream<R>;
-	<R>(xs: PromiseLike<R>): Highland.Stream<R>;
+	<R>(source: PromiseLike<Highland.Stream<R>>): Highland.Stream<R>;
+	<R>(source: PromiseLike<R>): Highland.Stream<R>;
 
-	<R>(xs: Iterable<R>): Highland.Stream<R>;
-	<R>(xs: Iterator<R>): Highland.Stream<R>;
+	<R>(source: Iterable<R>): Highland.Stream<R>;
+	<R>(source: Iterator<R>): Highland.Stream<R>;
 
 	// - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 	// UTILS
@@ -390,7 +390,7 @@ interface HighlandStatic {
 	 * @param args... - the arguments to apply to the function
 	 * @api public
 	 */
-	partial(f: Function, ...args: any[]): Function;
+	partial(fn: Function, ...args: any[]): Function;
 
 	/**
 	 * The reversed version of compose. Where arguments are in the order of
@@ -431,7 +431,7 @@ interface HighlandStatic {
 	 * _.not(true)   // => false
 	 * _.not(false)  // => true
 	 */
-	not<R>(a: any): boolean;
+	not<R>(x: any): boolean;
 }
 
 declare namespace Highland {
@@ -776,7 +776,7 @@ declare namespace Highland {
 		 * @id intersperse
 		 * @section Transforms
 		 * @name Stream.intersperse(sep)
-		 * @param {R} sep - the value to intersperse between the source elements
+		 * @param {R} separator - the value to intersperse between the source elements
 		 * @api public
 		 */
 		intersperse(separator: R): Stream<R>;
@@ -893,8 +893,7 @@ declare namespace Highland {
 		 * @param {Function} iterator - the function which reduces the values
 		 * @api public
 		 */
-		// TODO: convert this to this.scan(z, f).last()
-		reduce<U>(memo: U, f: (memo: U, x: R) => U): Stream<U>;
+		reduce<U>(memo: U, iterator: (memo: U, x: R) => U): Stream<U>;
 
 		/**
 		 * Same as [reduce](#reduce), but uses the first element as the initial
@@ -906,7 +905,7 @@ declare namespace Highland {
 		 * @param {Function} iterator - the function which reduces the values
 		 * @api public
 		 */
-		reduce1<U>(f: (memo: U, x: R) => U): Stream<U>;
+		reduce1<U>(iterator: (memo: U, x: R) => U): Stream<U>;
 
 		/**
 		 * The inverse of [filter](#filter).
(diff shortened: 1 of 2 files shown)


types/highland/index.d.ts
```diff
@@ -734,6 +768,19 @@ declare namespace Highland {
...
 
+		/**
+		 * Creates a new Stream with the separator interspersed between the elements of the source.
+		 *
+		 * `intersperse` is effectively the inverse of [splitBy](#splitBy).
+		 *
+		 * @id intersperse
+		 * @section Transforms
+		 * @name Stream.intersperse(sep)
+		 * @param {R} sep - the value to intersperse between the source elements
```
NotBobTheBuilder: minor nitpick - could we change sep to separator to match the argument name in typescript?

types/highland/index.d.ts
```diff
@@ -842,7 +906,7 @@ declare namespace Highland {
 		 * @param {Function} iterator - the function which reduces the values
 		 * @api public
 		 */
-		reduce1<U>(memo: U, f: (memo: U, x: R) => U): Stream<U>;
+		reduce1<U>(f: (memo: U, x: R) => U): Stream<U>;
```
NotBobTheBuilder: similar nitpick here, could we change the parameter name to be `iterator` to match the doc string

types/highland/index.d.ts
```diff
@@ -884,7 +948,82 @@ declare namespace Highland {
 		 *
 		 * _([1, 2, 3, 4]).scan1(add) // => 1, 3, 6, 10
 		 */
-		scan1<U>(memo: U, x: (memo: U, x: R) => U): Stream<U>;
+		scan1<U>(x: (memo: U, x: R) => U): Stream<U>;
```
NotBobTheBuilder: likewise with the name iterator

types/highland/index.d.ts
```diff
@@ -132,11 +166,11 @@ interface HighlandStatic {
 	 * @param x - the object to test
 	 * @api public
 	 */
-	isStream(x: any): boolean;
+	isStream(x: any): x is Highland.Stream<any>;
```
NotBobTheBuilder: Nice touch

types/highland/index.d.ts
```diff
@@ -706,7 +740,7 @@ declare namespace Highland {
 		 * firstBlogpost(docs)
 		 * // => {type: 'blogpost', title: 'foo'}
 		 */
-		findWhere(props: Object): Stream<R>;
+		findWhere(props: Partial<R>): Stream<R>;
```
NotBobTheBuilder: Nice touch here too

types/highland/index.d.ts
```diff
@@ -784,6 +831,22 @@ declare namespace Highland {
 		 */
 		map<U>(f: (x: R) => U): Stream<U>;
 
+		pick<Prop extends keyof R>(props: Prop[]): Stream<Pick<R, Prop>>;
```
NotBobTheBuilder: could we add the docs for pick?

NotBobTheBuilder (commented):
Thanks for filing this PR!

It's looking great. I still need to manually test, but left some minor style comments in the meantime. I can approve after I get a chance to test, hopefully tomorrow

types/highland/index.d.ts
```diff
@@ -884,7 +948,82 @@ declare namespace Highland {
...
+		 *
+		 * `splitBy` is effectively the inverse of [intersperse](#intersperse).
+		 *
+		 * @id splitBy
+		 * @section Transforms
+		 * @name Stream.splitBy(sep)
+		 * @param {String | RegExp} sep - the separator to split on
+		 * @api public
+		 */
+		splitBy(this: Stream<string>, sep: string): Stream<string>;
```
NotBobTheBuilder: sep should be `string | RegExp`

types/highland/index.d.ts
```diff
@@ -784,6 +831,22 @@ declare namespace Highland {
...
+		 * only enumerable elements are selected. Both enumerable
+		 * and non-enumerable elements are selected when using ES5.
+		 *
+		 * @id pickBy
+		 * @section Transforms
+		 * @name Stream.pickBy(f)
+		 * @param {Function} f - the predicate function
+		 * @api public
+		 */
+		pickBy(f: (key: string, value: any) => boolean): Stream<Partial<R>>
```
NotBobTheBuilder: Could this be the following?

```
		pickBy<K extends keyof R>(f: (key: K, value: R[K]) => boolean): Stream<Partial<R>>
```

types/highland/index.d.ts
```diff
@@ -734,6 +768,19 @@ declare namespace Highland {
...
+		 *
+		 * `intersperse` is effectively the inverse of [splitBy](#splitBy).
+		 *
+		 * @id intersperse
+		 * @section Transforms
+		 * @name Stream.intersperse(sep)
+		 * @param {R} sep - the value to intersperse between the source elements
+		 * @api public
+		 */
+		intersperse(separator: R): Stream<R>;
```
NotBobTheBuilder: I suppose strictly speaking there is no reason the intersperse element has to be of the same type as R. What would your thoughts be on this?

```
		intersperse<T>(separator: T): Stream<R | T>;
```

With these tests?

```
// $ExpectType Stream<Foo>
fooStream = fooStream.intersperse(foo);

// $ExpectType Stream<Foo|Bar>
fooStream.intersperse(bar)
```

types/highland/index.d.ts
NotBobTheBuilder: Additionally, I think the type signature can be hairy for this -- on the first iteration, memo will be R, but can be anything after that depending on what the iterator returns. What do you think of this?

```
		reduce1<U>(iterator: (a: R | U, b: R) => U): Stream<U>;
```

One would expect that R=U most of the time, but it doesn't have to be. Propose these tests for this:

```
// $ExpectType Stream<Foo>
fooStream = fooStream.reduce1((memo: Foo, x: Foo) => {
  return memo;
});

// $ExpectType Stream<Bar>
barStream = fooStream.reduce1((memo: Foo | Bar, x: Foo) => {
  return bar;
});
```

types/highland/index.d.ts
NotBobTheBuilder: Similar situation to reduce1 -- how about

```
scan1<U>(iterator: (a: R | U, x: R) => U): Stream<U>;
```

And as tests are already missing lets add these:

```

// $ExpectType Stream<Foo>
fooStream = fooStream.scan1((memo, x) => {
  return memo;
});

// $ExpectType Stream<Bar>
barStream = fooStream.scan1((memo, x) => {
  return bar;
});
```

types/highland/index.d.ts
```diff
@@ -884,7 +948,82 @@ declare namespace Highland {
...
+		 * returned.
+		 *
+		 * @id slice
+		 * @section Transforms
+		 * @name Stream.slice(start, end)
+		 * @param {Number} start - integer representing index to start reading from source (inclusive)
+		 * @param {Number} end - integer representing index to stop reading from source (exclusive)
+		 * @api public
+		 */
+		slice(start: number, end: number): Stream<R>;
```
NotBobTheBuilder: Lets add a test

```
// $ExpectType Stream<Foo>
fooStream = fooStream.slice(10, 100)
```

types/highland/index.d.ts
```diff
@@ -884,7 +948,82 @@ declare namespace Highland {
...
+		 *
+		 * @id sort
+		 * @section Transforms
+		 * @name Stream.sort()
+		 * @api public
+		 *
+		 * var sorted = _(['b', 'z', 'g', 'r']).sort().toArray(_.log);
+		 * // => ['b', 'g', 'r', 'z']
+		 */
+		sort(): Stream<R>;
```
NotBobTheBuilder: Lets add a test for this too

```
// $ExpectType Stream<Foo>
fooStream = fooStream.sort()
```

types/highland/index.d.ts
```diff
@@ -884,7 +948,82 @@ declare namespace Highland {
...
+		 * order](https://en.wikipedia.org/wiki/Partially_ordered_set). If it does not,
+		 * the resulting ordering is undefined.
+		 *
+		 * @id sortBy
+		 * @section Transforms
+		 * @name Stream.sortBy(f)
+		 * @param {Function} f - the comparison function
+		 * @api public
+		 */
+		sortBy(f: (a: R, b: R) => number): Stream<R>;
```
NotBobTheBuilder: Another test case:

```
// $ExpectType Stream<Foo>
fooStream = fooStream.sortBy((a: Foo, b: Foo) => 1)
```

types/highland/index.d.ts
```diff
@@ -884,7 +948,82 @@ declare namespace Highland {
...
+
+		/**
+		 * [splitBy](#splitBy) over newlines.
+		 *
+		 * @id split
+		 * @section Transforms
+		 * @name Stream.split()
+		 * @api public
+		 */
+		split(this: Stream<string>): Stream<string>;
```
NotBobTheBuilder: Test cases:

```
// $ExpectError
fooStream.split()

// $ExpectType Stream<string>
_(['']).split()
```

types/highland/index.d.ts
NotBobTheBuilder:
```

// $ExpectError
fooStream.splitBy(",")

// $ExpectType Stream<string>
_(['']).splitBy(',')

// $ExpectType Stream<string>
_(['']).splitBy(/,/)
```

types/highland/index.d.ts
```diff
@@ -966,7 +1105,7 @@ declare namespace Highland {
 		 * @param {Object} props - the properties to match against
 		 * @api public
 		 */
-		where(props: Object): Stream<R>;
+		where(props: Partial<R>): Stream<R>;
```
NotBobTheBuilder: Lets also add

```
bazStream = bazStream.where({'baz': true})
```

types/highland/index.d.ts
```diff
@@ -1064,7 +1203,36 @@ declare namespace Highland {
...
+		 * var txt = _(['foo.txt', 'bar.txt']).flatMap(readFile)
+		 * var md = _(['baz.md']).flatMap(readFile)
+		 * var js = _(['bosh.js']).flatMap(readFile)
+		 *
+		 * _([txt, md, js]).mergeWithLimit(2);
+		 * // => contents of foo.txt, bar.txt, baz.txt and bosh.js in the order
+		 * // they were read, but bosh.js is not read until either foo.txt and bar.txt
+		 * // has completely been read or baz.md has been read
+		 */
+		mergeWithLimit<U>(this: Stream<Stream<U>>, n: number): Stream<U>;
```
NotBobTheBuilder: Lets add these too

```
// $ExpectError
fooStream.merge();

fooStream = fooStreamStream.mergeWithLimit(1);

// $ExpectError
fooStream.mergeWithLimit(1);
```

types/highland/index.d.ts
```diff
@@ -1204,6 +1372,38 @@ declare namespace Highland {
...
+		 *
+		 * *Note:* This transform will be renamed `zipAll` in the next major version
+		 * release.
+		 *
+		 * @id zipAll0
+		 * @section Higher-order Streams
+		 * @name Stream.zipAll0()
+		 * @api public
+		 */
+		zipAll0(this: Stream<Stream<R>>): Stream<R[]>;
```
NotBobTheBuilder: I tried to add a test case for zipAll0 but typescript complained - do you know why that might be? The method definition looks right on first glance to me

```
fooStreamStream.zipAll0()
```

```
The 'this' context of type 'Stream<Stream<Foo>>' is not assignable to method's 'this' of type 'Stream<Stream<Stream<Foo>>>'.
  Types of property 'write' are incompatible.
    Type '(x: Stream<Foo>) => boolean' is not assignable to type '(x: Stream<Stream<Foo>>) => boolean'.
      Types of parameters 'x' and 'x' are incompatible.
        Type 'Stream<Stream<Foo>>' is not assignable to type 'Stream<Foo>'.
          Types of property 'write' are incompatible.
            Type '(x: Stream<Foo>) => boolean' is not assignable to type '(x: Foo) => boolean'.
              Types of parameters 'x' and 'x' are incompatible.
                Type 'Foo' is not assignable to type 'Stream<Foo>'.ts(2684)
```

NotBobTheBuilder (commented):
Took another pass and did some manual reviewing. This is looking great - I suggested some test cases to include, couple of further type narrowing opportunities and may have found an issue in `zipAll0`. 

After these changes, I don't think there is anything further needed before I approve. Sorry to surface so many things but they're hopefully all small enough to be easily addressed!

types/highland/index.d.ts
SupernaviX: That's a good point, I hadn't thought of that use case.

types/highland/index.d.ts
```diff
@@ -1262,23 +1462,41 @@ declare namespace Highland {
...
+		 * @name Stream.pipe(dest, options)
 		 * @param {Writable Stream} dest - the destination to write all data to
+		 * @param {Object} options - (optional) pipe options.
 		 * @api public
 		 */
 		pipe<U>(dest: Stream<U>): Stream<U>;
-		pipe<U>(dest: NodeJS.ReadWriteStream): Stream<U>;
-		pipe(dest: NodeJS.WritableStream): void;
+		pipe<U>(dest: NodeJS.ReadWriteStream, options?: { end?: boolean }): Stream<U>;
+		pipe(dest: NodeJS.WritableStream, options?: { end?: boolean }): Stream<void>;
```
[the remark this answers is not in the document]
SupernaviX: Good call, yes it should be.  I must have gotten this behavior mixed up with `through`.

types/highland/index.d.ts
SupernaviX: Sure, that makes sense. The type of `value` isn't as useful as it could be here, because narrowing the type of `key` doesn't affect the type of `value`, but it beats an `any`.

types/highland/index.d.ts
SupernaviX: That makes sense. I honestly hadn't thought about it for cases besides strings, but that's easy to add and doesn't interfere with normal consumption.

types/highland/index.d.ts
SupernaviX: Oh yeah, that's an interesting mistake on my part. I think that as written, `this` must be both `Stream<R>` (from the class definition) and `Stream<Stream<R>>` (from the method definition). I tried switching it to
```
zipAll0<T>(this: Stream<Stream<T>>): Stream<T[]>;
```
and the test started to pass.

Sorry for submitting broken code; my typescript codebase doesn't use much highland now, so I don't have real code to test most of this against. More reason I should have been adding test cases for all the new methods.

types/highland/index.d.ts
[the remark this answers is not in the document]
SupernaviX: After reading the docs/code, I'm replacing the node stream overloads with
```
		pipe<U extends NodeJS.WritableStream>(dest: U, options?: { end?: boolean }): U
````
If a consumer passes something more specific than a generic WritableStream, this won't lose type information.

SupernaviX:
I've pushed fixes for all of the issues above. Sorry again about pushing that broken method, and I really appreciate you taking the time to write all those test cases.

types/highland/index.d.ts
NotBobTheBuilder: Interesting, I haven't seen that kind of error before, nice work debugging it!

NotBobTheBuilder (approved):
Thanks again for filing this PR! Looks great to me, approving now

PR #41538, opened 2020-01-11
@types/highlight.js - Add lib and language types

### LINKED ISSUE #41351: [@types/highlight.js] Add types for `highlight.js/lib/highlight.js`
As described in https://github.com/highlightjs/highlight.js/issues/2340 I would like to import file `highlight.js/lib/highlight.js` but I it complains that I don't have types for it since types were just defined for the index.js file.

@JoshuaKGoldberg I read the links you sent me, but I am still not understanding how to type a file inside a module. Can you maybe provide a bit more help in how I can accomplish that?

I have esModuleInterop enabled and the types for that file are exactly the same as the ones for the main entrypoint of the module, but I still cannot get it to work.

rtfpessoa:
Fixes https://github.com/DefinitelyTyped/DefinitelyTyped/issues/41351

- [Err] Add or edit tests to reflect the change. (Run with `npm test`.)

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: No changes just missing types
- [N.A.] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [N.A.] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

### DIFF
--- a/types/highlight.js/index.d.ts
+++ b/types/highlight.js/index.d.ts
@@ -158,6 +158,13 @@ declare namespace hljs
 	}
 }
 
+declare module 'highlight.js/lib/highlight.js' {
+	export = hljs;
+}
+
+declare module 'highlight.js/lib/languages/*' {
+	export default function(hljs?: hljs.HLJSStatic): hljs.IModeBase;
+}
 
 export = hljs;
 export as namespace hljs;


rtfpessoa:
@JoshuaKGoldberg can you take a look here? I tested in my code and it worked. If you have suggestions on how to make tests here, I tried with `import * as hljs from 'highlight.js/lib/highlight.js';` but got a module not found error.

rtfpessoa:
Manually tested in https://github.com/rtfpessoa/diff2html/pull/279

rtfpessoa:
@JoshuaKGoldberg if you can check this one would be nice to merge it too.

types/highlight.js/index.d.ts
```diff
@@ -158,6 +158,13 @@ declare namespace hljs
 	}
 }
 
+declare module 'highlight.js/lib/highlight.js' {
```
nikeee: Could you add a test for these module exports?

Something like this would suffice:
https://github.com/rtfpessoa/diff2html/pull/279/files#diff-b343dabfda22e9aaae3240a5add32efb

nikeee (changes requested):

types/highlight.js/index.d.ts
rtfpessoa: @nikeee As I tried to explain in the main text. I cannot get the import of those modules in here to work. It fails saying they were not found. While in diff2html it worked fine.
Any suggestions?

rtfpessoa:
@nikeee any ideas about my comment?

types/highlight.js/index.d.ts
nikeee: When using your new type defintions on a clean project, only the first one wroks:

```
import * as c from "highlight.js";
import * as b from "highlight.js/lib/highlight.js";
import a = require("highlight.js/lib/highlight.js");
```

rtfpessoa: Then, how can I add tests for those new modules?

muuvmuuv:
Would love to see this merged. What is missing?

rtfpessoa:
I could not get any tests for the added types to compile. Since I could not get help this was eventually closed. 
In my case I solved this by putting them in my project and has been working great.

muuvmuuv:
Can also confirm that this is working in my Angular 9 Ivy/Ionic project with the tyings you have added to diff2html.

PR #41552, opened 2020-01-13
[reactstrap] Update Tooltip and Popover to add RefObject as a target

ngbrown:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/reactstrap/reactstrap/blob/0eade39d84e7a3e1cd8f6780b4040a49331568d8/src/__tests__/Uncontrolled.spec.js#L118


kyletsang (changes requested):
Can you add a test for this please?  Thanks.

ngbrown:
I've looked through the `reactstrap` code and found more potential uses of `RefObject`.  I've created tests for everything I found and updated the component prop types.

types/reactstrap/lib/Collapse.d.ts
```diff
@@ -23,7 +23,7 @@ export interface CollapseProps extends React.HTMLProps<HTMLElement> {
 
 export interface UncontrolledCollapseProps extends CollapseProps {
     defaultOpen?: boolean;
-    toggler: string;
+    toggler: string | React.RefObject<HTMLElement>;
```
kyletsang: According the to reactstrap documentation, this should be a string.

`The toggler prop is a string which will run querySelectorAll to find dom elements which will trigger toggle.`

https://reactstrap.github.io/components/collapse/

kyletsang (changes requested):

types/reactstrap/lib/Collapse.d.ts
ngbrown: It takes an id (e.g. `#toggler`), which is exactly the sort of thing you would rather use RefObject for instead. Infact, in the code, the `toggler` value is passed to `findDOMElements()`, which as its [very first lines is](https://github.com/reactstrap/reactstrap/blob/6b46b59a39e665ecd837538059bf43e36a5a841e/src/utils.js#L274):

```js
  if (isReactRefObj(target)) {
    return target.current;
  }
```

But I did notice that the `propTypes` has [it narrowed](https://github.com/reactstrap/reactstrap/blob/6b46b59a39e665ecd837538059bf43e36a5a841e/src/UncontrolledCollapse.js#L10) to `toggler: PropTypes.string.isRequired,`, so your probably right that it should be a string only without checking with the reactstrap project.

I'll remove that change for now.

kyletsang: It does feel like the reactstrap documentation is outdated in this part.  We can address this once we get confirmation from the reactstrap project. Everything looks good otherwise. Thanks!

kyletsang (approved):

PR #41559, opened 2020-01-13
fix(nyaapi): cleanups and fixes

infanf:
remove unnecessary interfaces
add category and filter definitions
fix options definitions for si

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: Categories and Filters from https://nyaa.si and https://nyaa.net (Search form options)

### DIFF
--- a/types/nyaapi/index.d.ts
+++ b/types/nyaapi/index.d.ts
@@ -6,32 +6,46 @@
 export namespace si {
     type UserStatus = number;
 
-    interface Comment {
-        username: string;
-        user_id: number;
-        user_avatar: string;
-        user_status: string;
-        content: string;
-        date: string;
+    enum Filter {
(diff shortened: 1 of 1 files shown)


types/nyaapi/index.d.ts
```diff
@@ -134,7 +175,7 @@ export namespace pantsu {
     interface SearchOptions {
         term?: string;
         n?: number;
-        c?: string[];
+        c?: Category;
```
elibarzilay: @infanf: You've replaced `string` with `Category` in a few places, but here you're replacing `string[]` -- is this intentional?

infanf: @elibarzilay yes it is. The nyaa pantsu REST API docs where the array came from are outdated (or were wrong to begin with)

PR #41356, opened 2020-01-02
[@types/node] Add worker.SHARE_ENV and WorkerOptions.env

dnalborczyk:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [ ] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: 
- https://nodejs.org/dist/latest-v13.x/docs/api/worker_threads.html#worker_threads_worker_share_env
- https://nodejs.org/dist/latest-v13.x/docs/api/worker_threads.html#worker_threads_new_worker_filename_options
- [ ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

PR #41562, opened 2020-01-13
add missing field "env" in WorkerOptions in worker_threads

saartochner-lumigo:
Please fill in this template.

- [X ] Test the change in your own code. (Compile and run.)
- [X ] Add or edit tests to reflect the change. (Run with `npm test`.)
- [X ] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).

- [X ] Provide a URL to documentation or source code which provides context for the suggested changes: 
https://nodejs.org/docs/latest-v12.x/api/worker_threads.html#worker_threads_worker_share_env
https://nodejs.org/docs/latest-v12.x/api/worker_threads.html#worker_threads_new_worker_filename_options

### DIFF
--- a/types/node/v12/worker_threads.d.ts
+++ b/types/node/v12/worker_threads.d.ts
@@ -60,6 +60,7 @@ declare module "worker_threads" {
         stdout?: boolean;
         stderr?: boolean;
         execArgv?: string[];
+        env?: symbol;
     }
 
     class Worker extends EventEmitter {


types/node/v12/worker_threads.d.ts
```diff
@@ -60,6 +60,7 @@ declare module "worker_threads" {
         stdout?: boolean;
         stderr?: boolean;
         execArgv?: string[];
+        env?: symbol;
```
Flarna: I think #41356 has a better solution in this as besides `worker.SHARE_ENV` also an object is possible here.

Flarna (changes requested):

PR #23169, opened 2018-01-24
express-serve-static-core: add support for custom locals

OliverJAsh:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: http://expressjs.com/en/api.html#app.locals and http://expressjs.com/en/api.html#res.locals
- [ ] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

PR #41568, opened 2020-01-13
`history`: use interface for `LocationState` to allow augmentation

### LINKED ISSUE #27012: `history`: no way to customise or override `LocationState`
`LocationState` has type `any`:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/566e0394859fdc1dc893658ccec6b06372d56a91/types/history/index.d.ts#L46

There should be a way of customising/overriding this, so we're not forced to cast/validate an `any` type every time we try to use it.

I am currently working around this by patching the types to remove the `LocationState` type, and adding it back in via module augmentation and namespace merging:

``` ts
declare module 'history' {
  namespace History { type LocationState = CustomLocationState; }
}
```

Hopefully one day TS will allow `any` types to be overriden: https://github.com/Microsoft/TypeScript/issues/4062

---

- [ ] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).
  - Authors: @sergey-buturlakin @ngbrown @rokoroku 

OliverJAsh:
Related:

- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23169
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/27699
- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37718
- https://github.com/microsoft/TypeScript/issues/36178

Fixes https://github.com/DefinitelyTyped/DefinitelyTyped/issues/27012

Currently, `LocationState` defaults to any. A custom type can be provided via generics:

```ts
{
    const anything: any = {};
    const history: History<{ foo: string }> = anything;
    history.location.state; // $ExpectType { foo: string; }
    history.location.state.foo; // $ExpectType string
    history.location.state.bar; // $ExpectError
}
```

However, it is very easy to make the mistake of forgetting to provide the generic:

```ts
{
    const anything: any = {};
    const history: History = anything;
    history.location.state.bar; // $ExpectError, but got none, because `state` is `any`
}
```

Furthermore, this will not help when `history` or `location` are provided to you (meaning their types must be inferred):

```ts
withRouter(({ history, location }) => {
  history.location.state.bar; // $ExpectError, but got none, because `state` is `any`
  location.state.bar; // $ExpectError, but got none, because `state` is `any`
});
```

This PR proposes replacing `type LocationState = any` with `interface LocationState {}`. This means it will be an empty object by default, and custom properties can be added via module augmentation:

```ts
declare module 'history' {
    namespace History {
        interface LocationState {
            foo: string;
        }
    }
}

{
    const anything: any = {};
    const history: History = anything;
    history.location.state; // $ExpectType LocationState
    history.location.state.foo; // $ExpectType string
    history.location.state.bar; // $ExpectError
}
```

With this change, I don't see any reason to keep the `S` location state generic. If you're happy with these changes, I'll go ahead and remove that.

/cc @gwmccull

---

Please fill in this template.

### DIFF
--- a/types/history/history-tests.ts
+++ b/types/history/history-tests.ts
@@ -125,3 +125,23 @@ let input = { value: "" };
     let supportsDOM = ExecutionEnvironment.canUseDOM;
     let isExtraneousPopstateEvent = DOMUtils.isExtraneousPopstateEvent;
 }
+
+//
+// Location state augmentation
+//
+
+declare module 'history' {
+    namespace History {
+        interface LocationState {
+            foo: string;
+        }
+    }
+}
+
+{
+    const anything: any = {};
+    const history: History = anything;
+    history.location.state; // $ExpectType LocationState
+    history.location.state.foo; // $ExpectType string
+    history.location.state.bar; // $ExpectError
+}
--- a/types/history/history-tests.ts
+++ b/types/history/history-tests.ts
@@ -133,7 +133,7 @@ let input = { value: "" };
 declare module 'history' {
     namespace History {
         interface LocationState {
-            foo: string;
+            foo?: string;
         }
     }
 }
--- a/types/history/history-tests.ts
+++ b/types/history/history-tests.ts
@@ -4,6 +4,29 @@ import * as PathUtils from 'history/PathUtils';
 import * as DOMUtils from 'history/DOMUtils';
 import * as ExecutionEnvironment from 'history/ExecutionEnvironment';
 
+//
+// Location state augmentation
+//
+
+declare module 'history' {
+    namespace History {
+        interface LocationState {
+            some?: string;
+            foo?: string;
+            value?: number;
+            the?: string;
+        }
+    }
+}
+
+{
+    const anything: any = {};
+    const history: History = anything;
+    history.location.state; // $ExpectType LocationState | undefined
+    history.location.state!.foo; // $ExpectType string | undefined
+    history.location.state!.bar; // $ExpectError
+}
+
 let input = { value: "" };
 
 {
@@ -43,7 +66,7 @@ let input = { value: "" };
 }
 
 {
-    let history: MemoryHistory<{the: 'state'}> = createMemoryHistory();
+    let history: MemoryHistory = createMemoryHistory();
 
     // Pushing a path string.
     history.push('/the/path');
@@ -65,7 +88,7 @@ let input = { value: "" };
     unblock();
 
     history.entries.forEach(function (location) {
-        let typedLocation: Location<{ the: 'state' }> = location;
+        let typedLocation: Location = location;
     });
 }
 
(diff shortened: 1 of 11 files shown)


types/history/history-tests.ts
```diff
@@ -96,14 +119,14 @@ let input = { value: "" };
 }
 
 {
-    let location1 = LocationUtils.createLocation('path/1', 1);
```
OliverJAsh: Trade off: you can no longer use non-object types

types/react-router/test/hooks.tsx
```diff
@@ -5,13 +5,17 @@ interface Params {
     id: string;
 }
 
-interface LocationState {
-    s: string;
+declare module 'history' {
+    namespace History {
+        interface LocationState {
+            s?: string;
```
OliverJAsh: I had to make these augmentations optional because the augmentation will be used for all tests!

OliverJAsh:
One of the test fails and I'm not sure how to fix it.

types/history/history-tests.ts
gwmccull: does this mean that these types don't match the reality of the library?

gwmccull:
I'm not very familiar with module augmentation. Would that have to be done in a `.d.ts` file or can that be done in a component file (for example, in a React component)?

types/history/history-tests.ts
OliverJAsh: Correct, that's the trade off.

Ideally we would be use the same technique without forcing this to be an object type, but declaration merging doesn't allow that. For example, declaration merging works with interfaces:

```ts
interface Bar {}

interface Foo {
  bar: Bar;
}

interface Bar {
  a: number;
}

declare const foo: Foo;

foo.bar.a; // number
```

… but it will not override a type alias:

```ts
type Bar = unknown;

interface Foo {
  bar: Bar;
}

// Compile error
// Duplicate identifier 'Bar'
type Bar = number;

declare const foo: Foo;

foo.bar; // unknown
```

This does mean it's a breaking change, but it shouldn't be too difficult to workaround this by wrapping primitives in an object, like I did here (`2` -> `{ value: 2 }`).

OliverJAsh:
@gwmccull It can be done in either.

types/history/history-tests.ts
OliverJAsh: https://github.com/microsoft/TypeScript/issues/4062#issuecomment-360528573

gwmccull: well, I'm not the maintainer so I'm not sure if I get a vote around here but I would vote against a change that means the type signatures no longer match with a valid input of the library

if it's critical to change the location state to only accept an object, it seems like that first needs to occur in the `history` library itself

PR #41569, opened 2020-01-13
`express-serve-static-core`: use interface for locals to allow augmentation

OliverJAsh:
Reinstates https://github.com/DefinitelyTyped/DefinitelyTyped/pull/23169

Should we make the same change for params? 🤔

### DIFF
--- a/types/express-serve-static-core/express-serve-static-core-tests.ts
+++ b/types/express-serve-static-core/express-serve-static-core-tests.ts
@@ -1,9 +1,27 @@
 import * as express from 'express-serve-static-core';
 
 const app: express.Application = {} as any;
+
+declare module 'express-serve-static-core' {
+    interface ApplicationLocals {
+        foo: string;
+    }
+    interface ResponseLocals {
+        foo2: string;
+    }
+}
+
+app.locals.foo; // $ExpectType string
+app.locals.bar; // $ExpectError
+
 app.listen(3000);
 app.listen(3000, (err: any) => {
     // no-op error callback
+});
+
+app.use((_req, res) => {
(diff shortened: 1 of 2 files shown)


JoseLion:
It seems to not be that simple. According to the documentation res.locals is:
> An object that contains response local variables scoped to the request, and therefore available only to the view(s) rendered during that request / response cycle (if any). Otherwise, this property is identical to app.locals.

We might have type mismatches at some point 🤔 any thoughts?

https://expressjs.com/en/api.html#res.locals
https://expressjs.com/en/api.html#app.locals

OliverJAsh:
I guess response locals could be generic (like params currently are), and application locations could be augmentable interfaces (as seen in this PR). OTOH, a user might have multiple applications, so then perhaps they ought to be generic as well. It's just a PITA to specify the generic everywhere :-/

JoseLion (changes requested):
@OliverJAsh personally I do like the idea of being able to augment an interface instead a using `any`, locals should at least be of type `{}` and we will be able to have safer types 👍 

I was thinking maybe we could use [conditional types](http://www.typescriptlang.org/docs/handbook/advanced-types.html#conditional-types) on `res.locals` to determine if it's of type `ResponseLocals` or `ApplicationLocals`. However, I'm not quite sure that would be possible at this point, what do you think? Any ideas?

If we cannot use conditional types, I think at least we can add documentation to `app.locals` and `res.locals` properties to explain that `ApplicationLocals`/`ResponseLocals` are empty interfaces that can be augmented (we can add a small example too!). Otherwise, it might be a bit confusing for users to understand what `ApplicationLocals`/`ResponseLocals` types are supposed to be 🙂

Issue #25677, opened 2018-05-10
[meta] Contribution guide should document best practices for breaking changes


thw0rted:
- [X] I have a question that is inappropriate for [StackOverflow](https://stackoverflow.com/).  (Please ask any appropriate questions there).

BLUF: I broke somebody's build by removing a long-deprecated type from a package.  DT should take steps to prevent this sort of thing in the future.

Long version: For reasons too complex for this question, having the deprecated type `NodeBuffer` in `@types/node` caused deprecation warnings to show up in my code.  I looked into it and there hasn't been an actual NodeBuffer class in node since at least since DT started tracking, so I put in a PR (#25500) to remove it.  Because DT doesn't practice semver, this caused at least one user of `@types/node` to break a build where a second- or third-order dependency referenced `NodeBuffer`.

I have looked through the issues here and found a few tangentially related (like #7719) but I don't see any discussion of how to mark changed typings to prevent consumers from breaking on an automated update.  It's my understanding that the current convention has `@types/foo` versioned with a number that matches the library `foo` it describes.  The problem with this is of course that I can release 10 typings for `foo@1.2.3` as I improve the description, adding more specific return types or correcting errors.  Each of these typings changes (for the same version of the target library) could potentially be breaking or non-breaking, but the vocabulary we have today does not allow me to express this.

What I'd like to be able to say is "this is version 4.5.6 of the typings for node 9.X", *and have npm understand* that, per semver, it can update automatically to 4.6.0 of the node 9.X typings, but it should *not* automatically update to typings version 5.0.0 when that is created.  Then, when I want to make a breaking change to the typings -- *not* based on changes to the underlying library, which would necessitate a major version increment in the library itself -- I could increment the typings version and users would know to check for breakage.

Is this possible? Desirable?  I'm certain I'm not the first to have this kind of thought, but if it was discussed here before, I haven't found it.  Regardless of the solution, this kind of problem -- how to handle breaking changes for typings within a major version of the library being described -- should get an entry in the contribution section of the README.

(Probably cc @mhegazy @RyanCavanaugh ?)

RyanCavanaugh:
> Each of these typings changes (for the same version of the target library) could potentially be breaking or non-breaking

[...]

PR #41580, opened 2020-01-14
`history`: switch default `LocationState` generic from `any` to `unknown`

OliverJAsh:
`any` disables type checking, so it should be avoided:

```ts
declare const any: any;
any.foo.bar; // no compile error, but runtime exception!
```

Related:

- https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41568
- https://github.com/microsoft/TypeScript/issues/36178
- https://github.com/DefinitelyTyped/DefinitelyTyped/issues/27012

---

Please fill in this template.

### DIFF
--- a/types/history/index.d.ts
+++ b/types/history/index.d.ts
@@ -49,7 +49,10 @@ export namespace History {
       location: Location<S>,
       action: Action,
     ) => void;
-    export type LocationState = any;
+    // The value type here is a "poor man's `unknown`". When these types support TypeScript
+    // 3.0+, we can replace this with `unknown`.
+    type PoorMansUnknown = {} | null | undefined;
+    export type LocationState = PoorMansUnknown;
     export type Path = string;
     export type Pathname = string;
     export type Search = string;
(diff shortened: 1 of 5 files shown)


types/history/index.d.ts
```diff
@@ -49,7 +49,10 @@ export namespace History {
       location: Location<S>,
       action: Action,
     ) => void;
-    export type LocationState = any;
+    // The value type here is a "poor man's `unknown`". When these types support TypeScript
+    // 3.0+, we can replace this with `unknown`.
+    type PoorMansUnknown = {} | null | undefined;
```
danielnixon: https://github.com/gcanti/unknown-ts/blob/master/src/index.ts#L1 maybe?

danielnixon (commented):
I suspect this change would break a lot of code. Worth reading:
* https://github.com/DefinitelyTyped/DefinitelyTyped/issues/25677#issuecomment-388215959

OliverJAsh:
> I suspect this change would break a lot of code. Worth reading:

Perhaps we could wait until the next major version of `history` comes out, although it would be difficult to co-ordinate the releases so they're both released at exactly the same time.

types/history/index.d.ts
OliverJAsh: This has worked well for me so far, I've also used it in the Node types.

johnnyreilly:
Thanks!

PR #41584, opened 2020-01-14
[braintree] fix subscription requests according to the docs

Qunapi:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://developers.braintreepayments.com/reference/request/subscription/create/node>

### DIFF
--- a/types/braintree/index.d.ts
+++ b/types/braintree/index.d.ts
@@ -958,6 +958,8 @@ declare namespace braintree {
             update?: DiscountUpdateRequest[];
         };
         firstBillingDate?: Date;
+        id?: string;
+        merchantAccountId?: string;
         neverExpires?: boolean;
         numberOfBillingCycles?: number;
         options?: {
@@ -974,8 +976,6 @@ declare namespace braintree {
         trialDuration?: number;
         trialDurationUnit?: string;
         trialPeriod?: boolean;
-        id?: string;
-        merchantAccountId?: string;
     }
 
     export interface SubscriptionHistory {
--- a/types/braintree/index.d.ts
+++ b/types/braintree/index.d.ts
@@ -174,7 +174,7 @@ declare namespace braintree {
 
     interface SubscriptionGateway {
         cancel(subscriptionId: string): Promise<void>;
-        create(request: SubscriptionCreateRequest): Promise<ValidatedResponse<Subscription>>;
+        create(request: SubscriptionRequest): Promise<ValidatedResponse<Subscription>>;
         find(subscriptionId: string): Promise<Subscription>;
         retryCharge(
             subscriptionId: string,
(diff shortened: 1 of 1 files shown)


types/braintree/index.d.ts
```diff
@@ -978,6 +976,16 @@ declare namespace braintree {
         trialPeriod?: boolean;
     }
 
+    export interface SubscriptionUpdateRequest extends SubscriptionRequest {
```
smrubin: From what I'm seeing on the docs, the `id` and `merchantAccountId` as part of the update request object seem to be optional as well. I think we can keep the existing `SubscriptionRequest` interface and make both these fields optional on that interface.

smrubin (changes requested):
@Qunapi Thanks for identifying this inconsistency with the docs. Per other comment, I think we should just mark the two fields as optional as part of the request object for both create and update. As a result, I don't think we need new interfaces for the create and update request objects as they'd be the same.

types/braintree/index.d.ts
Qunapi: But then we will have a slightly incorrect check for both update and create. Wouldn't that be a problem?

smrubin: Not sure I follow, what do you mean by slightly incorrect check? The `id` is required for update but only as the first argument in the `update` method, not as part of the request object itself. The way I'm reading the docs, the request object marks both of the touched fields as optional.

sharno: Actually it seems that `id` is optional in `create` but required in `update`
While `merchantAccountId` is optional in both of them

smrubin: @sharno So, there are two separate `id`s we may be talking about here. There is an `id` passed as the first argument to the `update` method. This is required, and is currently marked as such in the TS definition.

The second `id` is passed as part of the request object, and per the docs says "If this option is provided", leading me to believe it is optional.

Note the differences between:
1) The id argument (https://developers.braintreepayments.com/reference/request/subscription/update/node#arg.id)
2) The id as part of the request object (https://developers.braintreepayments.com/reference/request/subscription/update/node#id)

sharno: @smrubin Oh I see what you are referring to now. Your suggestion makes sense then of just making both fields optional in the current TS definition.

sharno (approved):
Looks good to me

types/braintree/index.d.ts
```diff
@@ -976,6 +974,8 @@ declare namespace braintree {
         trialDuration?: number;
         trialDurationUnit?: string;
         trialPeriod?: boolean;
+        id?: string;
```
smrubin: /nit Can we keep the props in alphabetical order? We've done this throughout and I think it makes lookups easier.

smrubin (approved):
Awesome, thanks @Qunapi

PR #41587, opened 2020-01-14
[parse] Parse.Object.toJSON and some generic static methods

rdhelms:
Please fill in this template.

-----

Up until now, `Parse.Object.toJSON` has been typed as `any`, which ultimately defeats the purpose of having a typed Parse Object at all. This PR types the return of `toJSON` in accordance with the Parse library's [ParseObject.toJSON](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseObject.js#L436) logic. The `ToJSON` type uses another new `Encode` type to map types in accordance with the library's [encode function](https://github.com/parse-community/Parse-SDK-JS/blob/master/src/encode.js#L22). 

In order for the type to reflect the recursive behavior of `encode` itself, this did require bumping the TS version to 3.7 - but we had already been bumped up to 3.5, so this isn't as large of a jump as we made a few weeks ago from 2.x to 3.x.

Additionally, some methods needed to be made generic in order to have better control over their return type.

### DIFF
--- a/types/parse/index.d.ts
+++ b/types/parse/index.d.ts
@@ -460,7 +460,7 @@ namespace Parse {
         pinAll(objects: Object[]): Promise<void>;
         pinAllWithName(name: string, objects: Object[]): Promise<void>;
         registerSubclass<T extends Object>(className: string, clazz: new (options?: any) => T): void;
-        saveAll<T extends Object>(list: T[], options?: Object.SaveAllOptions): Promise<T[]>;
+        saveAll<T extends readonly Object[]>(list: T, options?: Object.SaveAllOptions): Promise<T>;
         unPinAll(objects: Object[]): Promise<void>;
         unPinAllObjects(): Promise<void>;
         unPinAllObjectsWithName(name: string): Promise<void>;
--- a/types/parse/index.d.ts
+++ b/types/parse/index.d.ts
@@ -21,7 +21,7 @@
 //                  Linus Unnebäck <https://github.com/LinusU>
 //                  Patrick O'Sullivan <https://github.com/REPTILEHAUS>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 3.5
+// TypeScript Version: 3.7
 
 /// <reference types="node" />
 /// <reference path="node.d.ts" />
@@ -181,6 +181,12 @@ namespace Parse {
         createdAt: Date;
         objectId: string;
         updatedAt: Date;
+    }
+
+    interface JSONBaseAttributes {
+        createdAt: string;
+        objectId: string;
+        updatedAt: string;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


types/parse/index.d.ts
```diff
@@ -828,15 +855,15 @@ namespace Parse {
...
 
         allowCustomUserClass(isAllowed: boolean): void;
-        become(sessionToken: string, options?: UseMasterKeyOption): Promise<User>;
-        current<T extends Attributes>(): User<T> | undefined;
-        currentAsync(): Promise<User | null>;
-        signUp(username: string, password: string, attrs: any, options?: SignUpOptions): Promise<User>;
-        logIn(username: string, password: string, options?: FullOptions): Promise<User>;
-        logOut(): Promise<User>;
-        requestPasswordReset(email: string, options?: SuccessFailureOptions): Promise<User>;
+        become<T extends User>(sessionToken: string, options?: UseMasterKeyOption): Promise<T>;
```
LinusU: Would you mind explaining why you chose `<T extends User>(...): Promise<T>` over
`<T extends Attributes>(...): Promise<User<T>>`? Not saying that it's wrong, just curious...

rdhelms: Yep, I actually initially typed it as `<T extends Attributes>` but I changed it so that users would be able to pass a type that extended User but was not User itself. In other words, if someone made a custom User class that extended Parse.User, then I wanted them to be able to have these methods return their custom class instead of the parent Parse.User class.

Open to suggestions, though, if you have any thoughts on that.

LinusU: Ah, of course, I think that sounds very good 👍

rdhelms:
I just added a quick addition to this that we noticed today - `Parse.Object.saveAll` previously only returned array types, and so trying to destructure the response only ever gave you a union of the types of the elements of the array being passed...but using a tuple type gives better inferences when destructuring.

One of the test cases I just added demonstrates this:
```ts
// $ExpectType [Object<Attributes>, Object<{ example: string; }>]
await Parse.Object.saveAll<[ typeof objUntyped, typeof objTyped ]>([ objUntyped, objTyped ]);
```

Previously, the `saveAll` type parameter would not have allowed passing a tuple type, and the return would have been `(Object<Attributes> | Object<{ example: string; }>)[]`

types/parse/index.d.ts
RaschidJFR: Hey this works great! Actually this has been always bugging me since forever. Tnx @rdhelms

RaschidJFR:
Hey guys, I'm doubting now that upgrading TS to v7 is a good idea. This is a **breaking change** that got through unnoticed.

For example. **Angular 8** [that do not support ts +v3.5](https://gist.github.com/LayZeeDK/c822cc812f75bb07b7c55d07ba2719b3), and v9 is not yet released. 

I personally work a lot on Angular and sub frameworks (such as Ionic). Shouldn't we try a workaround or maybe revert part of this PR until the environment is more mature?

PR #41594, opened 2020-01-14
feat: typings for rdf-transform-triple-to-quad

tpluscode:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/rdf-js/index.d.ts
+++ b/types/rdf-js/index.d.ts
@@ -9,7 +9,6 @@
 
 import * as stream from "stream";
 import { EventEmitter } from "events";
-import { Readable } from "stream";
 
 /* Data Model Interfaces */
 /* https://rdf.js.org/data-model-spec/ */
@@ -329,7 +328,7 @@ export interface DataFactory {
  * Optional events:
  * * prefix(prefix: string, iri: RDF.NamedNode): This event is emitted every time a prefix is mapped to some IRI.
  */
-export interface Stream<Q extends BaseQuad = Quad> extends Readable {
+export interface Stream<Q extends BaseQuad = Quad> extends EventEmitter {
     /**
      * This method pulls a quad out of the internal buffer and returns it.
      * If there is no quad available, then it will return null.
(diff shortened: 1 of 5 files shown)


types/rdf-js/index.d.ts
```diff
@@ -328,7 +329,7 @@ export interface DataFactory {
  * Optional events:
  * * prefix(prefix: string, iri: RDF.NamedNode): This event is emitted every time a prefix is mapped to some IRI.
  */
-export interface Stream<Q extends BaseQuad = Quad> extends EventEmitter {
+export interface Stream<Q extends BaseQuad = Quad> extends Readable {
```
tpluscode: @rubensworks @LaurensRietveld @bergi

Any reason not to make a stream explicitly a `Readable`? I know it's not like that in the spec itself but the JSDoc clearly says it's a readable stream. Otherwise one need to jump additional hoops to too `.pipe()`

rubensworks: [Readable](https://nodejs.org/api/stream.html#stream_class_stream_readable) has a much stricter contract than EventEmitter. Since the RDFJS spec does not put those requirements on [Stream](http://rdf.js.org/stream-spec/#stream-interface), we should not add those in the typings. People from the pure JS world don't expect such an interface, so we shouldn't have this in TS either.

rubensworks (changes requested):

types/rdf-js/index.d.ts
tpluscode: > People from the pure JS world don't expect such an interface, so we shouldn't have this in TS either.

Really? This from stream spec, right? So how do not expect someone to see a `stream.Readable` as the base of a "readable quad stream"?

But ok, this is not the right place to discuss as we'd need to apply such change in the spec first. Let me revert for now

rubensworks (approved):
I approve the reverted changes to rdf-js.
Since I'm not a user of `rdf-transform-triple-to-quad`, it'd be good if someone else could also have a look at this.

PR #41601, opened 2020-01-15
Add declarations for chrome.enterprise.reportingPrivate

chabster4g:
- [*] Use a meaningful title for the pull request. Include the name of the package modified.
- [*] Test the change in your own code.
- [*] Provide a URL to documentation or source code which provides context for the suggested changes: https://cs.chromium.org/chromium/src/chrome/common/extensions/api/enterprise_reporting_private.idl

### DIFF
--- a/types/chrome-apps/index.d.ts
+++ b/types/chrome-apps/index.d.ts
@@ -2839,6 +2839,64 @@ declare namespace chrome {
          */
         function challengeUserKey(challenge: ArrayBuffer, registerKey: boolean, callback: (response: ArrayBuffer) => void): void;
     }
+    /**
+     * @requires Permissions: 'enterprise.reportingPrivate'
+     * @requires Explicit whitelisting.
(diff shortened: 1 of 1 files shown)


niikoo (changes requested):
Thanks for your contribution. This is an API that requires whitelisting. Please se our comments on the other PR you made.

chabster4g:
> Thanks for your contribution. This is an API that requires whitelisting. Please se our comments on the other PR you made.

Whitelisting has been done initially, along with the implementation. This is a private API explicitly created for our extension. I  tried to declare this namespace separately,  but it made other APIs inaccessible. I did not find any other way to declare those APIs in TS.

pyle:
@chabster4g namespace merging should work: https://www.typescriptlang.org/docs/handbook/declaration-merging.html#merging-namespaces
I have also sent you some internal mail for how this might work for your situation.

PR #41611, opened 2020-01-15
Changed definition of IncomingMessage

arel0ve:
TYPES/CONNECT: Changed definition of IncomingMessage from 'class' to 'interface'.
Fixed bug `Cannot extend an interface 'http.IncomingMessage'. Did you mean 'implements'?` from angular build.

### DIFF
--- a/types/connect/index.d.ts
+++ b/types/connect/index.d.ts
@@ -18,7 +18,7 @@ declare function createServer(): createServer.Server;
 declare namespace createServer {
     export type ServerHandle = HandleFunction | http.Server;
 
-    export class IncomingMessage extends http.IncomingMessage {
+    export interface IncomingMessage extends http.IncomingMessage {
         originalUrl?: http.IncomingMessage["url"];
     }
 


EvanHahn (changes requested):
Requesting changes because I have questions, not because this is necessarily wrong.

What problem were you running into during your Angular build? I'm surprised that `connect` would be involved at all.

And to me, this seems like it's a class (as it's an instance of `http.IncomingMessage` with a minor augmentation), not an interface. What's the rationale for the switch?

arel0ve:
@EvanHahn 
In angular build I get next error:
`ERROR in ../node_modules/@types/connect/index.d.ts:21:42 - error TS2689: Cannot extend an interface 'http.IncomingMessage'. Did you mean 'implements'?`
This error can be because of:
Class can't extend interface it can implement it and `http.IncomingMessage` is interface, not class. I tested, and after this little fix build works fine. (For build I used typescript@3.5.3 and angular/core@8.2.14).

EvanHahn:
It looks like [`http.IncomingMessage` is a class, not an interface](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c47a34ead1637f6f34e7d630dc88ea3f6e5562cb/types/node/http.d.ts#L274). I worry that changing it might cause some unexpected problems. What do you think?

arel0ve:
@EvanHahn 
Thanks, and sorry that I wasted your time. It looks like problem was after Angular Universal have been installed, it used automatically too old version of @types/node (6.0.118). And in this version http.IncomingMessage was an interface.
So, this pull request can be rejected. Sorry again, for wasting your time and many thanks.

Issue #41179, opened 2019-12-20
Add support for TypeScript's Assertion Functions


hjr3:
## 🚀 Feature Proposal

Originally posted at: https://github.com/facebook/jest/issues/9146
 
 Support TypeScript 3.7's new [assertion functions](http://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions).

## Motivation
 
 Jest should support this TypeScript language feature to make authoring tests simpler.
 ## Example
 
 Let's say I have a function under test that returns a nullable value:
 
 ```ts
 export interface Data {
   anotherCoolProp: any
   somethingNeat: any
 }
 
 export function gimmeData(): Data | null {
   // Implementation details
   // ...
 }
 ```
 
 Currently, testing the results requires null checks for every assertion:
 
 ```ts
 test('gimme that data', () => {
   const data = gimmeData()
 
   expect(data).toBeTruthy()
 
   expect(data!.anotherCoolProp).toEqual('coolio')
   expect(data!.somethingNeat).toEqual('neato')
 })
 
 // or:
 test('gimme that data', () => {
   const data = gimmeData()
 
   if (!data) {
     throw new Error('Expected data to be non-null')
   }
 
   expect(data.anotherCoolProp).toEqual('coolio')
   expect(data.somethingNeat).toEqual('neato')
 })
 ```
 
 The first uses the `!` non-null operator, which opts out of type safety and can lead to confusing error reports. The second seems non-idiomatic: why not write `expect(data).toBeTruthy()`?
 
 With TypeScript 3.7 you can now define [Jest's global `expect`](https://jestjs.io/docs/en/expect) to behave this way:
 
 ```ts
 declare function <T>expect(value: T): asserts value is NonNullable<T>
 ```
 
 (Although typing w/ the Jest's chaining will be a bit more involved.)

Lots of people have worked on these types, but I will start with @sandersn to see what he thinks about this.

kevinbarabash:
This is exciting!  Having to add type refinements in addition to test assertions is a pain.

Igmat:
This type definitions should be updated too:

[`@types/assert`](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/assert)

@nicoabie, @LinusU

If you don't have enough time, I'll probably update your definition tomorrow.

LinusU:
Nice, absolutely!

LinusU:
[...]

PR #41616, opened 2020-01-15
[assert] Add support for TypeScript's Assertion Functions

LinusU:
re. #41179

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/node/blob/master/lib/assert.js


LinusU:
Oh boy, quite a few packages depends on this just because some tests uses it... I'm working on it 💪

types/sanctuary/sanctuary-tests.ts
```diff
@@ -1,10 +1,11 @@
...
 const S = create({checkTypes, env});
 
-assert.equal(S.map(S.concat('@'))(['foo', 'bar', 'baz']), ["@foo", "@bar", "@baz"]);
-assert.equal(S.reduce(S.add)(0)([1, 2, 3, 4, 5]), 15);
-assert.equal(S.flip(S.concat)('foo')('bar'), "barfoo");
+// $ExpectType number
+S.reduce(S.add)(0)([1, 2, 3, 4, 5]);
+
+// $ExpectType string
+S.flip(S.concat)('foo')('bar');
```
davidchambers: Did you mean to drop the `S.map` expression?

LinusU: Yeah, it compiled to a really weird expression so it felt very strange to type check it. I'm guessing that it might actually be a bug that it didn't compile to `string[]`, but I didn't have time to look into it.

Since it didn't test anything before (`assert.equal` accepted `any`) I didn't feel like anything was lost by just dropping it...

What do you think?

davidchambers: Dropping it seems reasonable to me. Thanks for the explanation.

LinusU:
That should be all 💪 

ping @RyanCavanaugh, @orta, @sandersn: not really sure how the review process works when I'm touching a ton of packages 😅 any pointers?

beeequeue (approved):
superagent stuff looks good at least  💪

orta:
Impressive work, so most of the diff comes from you removing the assert lib in libraries which included it in their tests. Given that we don't eval the tests, only run the type checker on them - that seems like a very reasonable switch.

The biggest question which remains for me is: is it ok to move assert up to TS 3.7 and above? That's up to you and @nicoabie

LinusU:
> is it ok to move assert up to TS 3.7 and above?

I feel so, let's see what @nicoabie thinks ☺️

nicoabie (approved):

nicoabie:
@orta let's move on!
Thanks @LinusU for the heavy lifting!

LinusU:
Nooooo 😬

LinusU:
@orta it would be great if we could get this merged before any more conflicts arrive ❤️

uniqueiniquity:
@LinusU My mistake - I'll make sure this gets merged as soon as the build break gets fixed.

uniqueiniquity:
Seems like you just need to add "ES2016.Array.Include" to the lib section of oracledb's tsconfig.

LinusU:
Pushed a patch that should fix it 👍 

Not really sure how my change triggered that error from the beginning though, it should have been caught by the previous PR oracledb 🤔

LinusU:
@uniqueiniquity should be good to go now 👏

uniqueiniquity:
Hi @LinusU,
Unfortunately, I gave you some bad guidance regarding this PR.
It's causing problems for downstream packages that our automation failed to predict.
I'm going to have to revert it.

If you would like to resubmit this PR, I would recommend creating a `ts3.7` subfolder for this package and making your change there. That way, consumers with lower version requirements won't need to be modified. Details on how to do that can be found [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions).

I apologize for not catching this sooner!

vegerot:
very happy to see this 🙂

PR #41623, opened 2020-01-15
[parse] Fix typo in Parse.Object.save attrs type

rdhelms:
Please fill in this template.

-----

Fixing a typo:
`[key in K]: T[K];` -> `[key in K]: T[key];`
The `K` type parameter represents the union of all property names while `key` is the particular property name. Each property should be typed according to its own type, not the union of all property types.

The following test case also ensures that this is now correct:
```ts
// objTyped: Parse.Object<{ example: boolean; someString: string }>

// $ExpectError
await objTyped.save({ example: 'hello', someString: true });
```


rdhelms:
Just realized that this same typo is in `set`, so I'm going to fix that here as well

rdhelms:
👍🏻Fixed the same issue for `set` - all good to go now.

LinusU (approved):

PR #41624, opened 2020-01-15
Update react-simple-maps geography type from object to any

kennethbigler:
the object type is too restrictive, forcing people to use any in their code. The geography prop comes through as a very complex object, and using the object type prevents you from using it properly. If we want to allow any geography type (I am not sure if they vary by map, but I assume they would), we should type it as any, not as object.

That being said, I would happily contribute my type definition for geography if it is the same across all maps provided.

A quote straight from your common mistakes document:
> Using the types Function and Object is almost never a good idea. In 99% of cases it's possible to specify a more specific type. Examples are (x: number) => number for functions and { x: number, y: number } for objects. If there is no certainty at all about the type, any is the right choice, not Object. If the only known fact about the type is that it's some object, use the type object, not Object or { [key: string]: any }.

### DIFF
--- a/types/react-simple-maps/index.d.ts
+++ b/types/react-simple-maps/index.d.ts
@@ -84,7 +84,7 @@ export interface ZoomableGroupProps extends React.SVGAttributes<SVGGElement> {
 }
 
 interface GeographiesChildrenArgument {
-    geographies: object[];
+    geographies: any[];
     path: GeoPath;
     projection: GeoProjection;
 }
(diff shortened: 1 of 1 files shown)


thepocp (approved):

Zemnmez:
What about `unknown` instead of `any`?

kennethbigler:
@Zemnmez good idea, I updated the PR, thank you!

kennethbigler:
@Zemnmez changing it back to `any` to keep TypeScript 2 support

Zemnmez:
haha – you could split versions but maybe that's too much! nice PR anyway :)

PR #41627, opened 2020-01-15
Add types for yawn-yaml

JamieMagee:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/yawn-yaml/cjs/index.d.ts
+++ b/types/yawn-yaml/cjs/index.d.ts
@@ -2,6 +2,6 @@ export default class YAWN {
     constructor(content: string);
     json: any;
     yaml: string;
-    getRemark(path: string): any;
+    getRemark(path: string): string;
     setRemark(path: string, remark: string): boolean;
 }
--- a/types/yawn-yaml/cjs/index.d.ts
+++ b/types/yawn-yaml/cjs/index.d.ts
@@ -1,0 +1,7 @@
+export default class YAWN {
+    constructor(content: string);
+    json: any;
+    yaml: string;
+    getRemark(path: string): any;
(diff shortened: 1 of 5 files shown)


types/yawn-yaml/index.d.ts
```diff
@@ -0,0 +1,11 @@
+// Type definitions for yawn-yaml 1.4
+// Project: https://github.com/mohsen1/yawn#readme
+// Definitions by: Jamie Magee <https://github.com/me>
```
elibarzilay: Use your GH username

types/yawn-yaml/cjs/index.d.ts
```diff
@@ -0,0 +1,7 @@
+export default class YAWN {
+    constructor(content: string);
+    json: any;
+    yaml: string;
+    getRemark(path: string): any;
```
elibarzilay: Are you sure that you want to use so many `any`s?  For example, looking at the code, *I think* that `getRemark` returns an `undefined | string`...

elibarzilay (changes requested):

types/yawn-yaml/cjs/index.d.ts
JamieMagee: Good catch! I think it might actually just be `string`, as there's a `return ''` [in `getNodeRemark`](https://github.com/mohsen1/yawn-yaml/blob/master/src/index.js#L429-L455).

As for `json` that type comes directly [from `js-yaml#load`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/js-yaml/index.d.ts#L10)

elibarzilay (approved):

PR #41629, opened 2020-01-15
[swiper] Add missing controller property to options

AlexReff:
The definition is wrong, needs this added.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://swiperjs.com/api/#controller

All components listed on the left are included - except controller. This adds controller.

### DIFF
--- a/types/swiper/index.d.ts
+++ b/types/swiper/index.d.ts
@@ -439,7 +439,7 @@ export interface SwiperOptions {
     wrapperClass?: string;
 
     // Components
-    controller?: ControllerOptions;
+    controller?: ControllerOptions | boolean;
     navigation?: NavigationOptions;
     pagination?: PaginationOptions;
     scrollbar?: ScrollbarOptions;
--- a/types/swiper/index.d.ts
+++ b/types/swiper/index.d.ts
@@ -439,6 +439,7 @@ export interface SwiperOptions {
     wrapperClass?: string;
 
     // Components
+    controller?: ControllerOptions;
     navigation?: NavigationOptions;
     pagination?: PaginationOptions;
     scrollbar?: ScrollbarOptions;


types/swiper/index.d.ts
```diff
@@ -439,6 +439,7 @@ export interface SwiperOptions {
     wrapperClass?: string;
 
     // Components
+    controller?: ControllerOptions;
```
peterblazejewicz: @AlexReff question, the docs linked states this could be also a `boolean`:
https://swiperjs.com/api/#controller
```ts
const swiper = new Swiper(containerSelector, {
        controller: true,
```
so:
```ts
controller?: ControllerOptions | true;
```

AlexReff: Good catch, I didn't even realize that when adding this.

`controller?: ControllerOptions | boolean;`

Would be correct, but I'm not sure how to update my PR.

peterblazejewicz: > Would be correct, but I'm not sure how to update my PR.

Using desktop app:
https://help.github.com/en/desktop/contributing-to-projects
simply: add new changes as a new commit and push changes (all from your branch), your commits will be squashed later into single one if accepted

If you have used the Github online editor:
https://www.youtube.com/watch?v=e41HPOHX9aE

odahcam:
```suggestion
    controller?: ControllerOptions | true;
```

You can update the PR by accepting the suggestion. 😉

AlexReff: Updated my PR

AlexReff: Thanks, updated!

AlexReff:
@peterblazejewicz @odahcam Updated, think it needs another review, not familiar with this process.

peterblazejewicz (approved):
LGTM
@AlexReff, thx for contribution! and no worries, this is technically OK, can be updated later!

odahcam (approved):

peterblazejewicz:
@odahcam  🙇

PR #41631, opened 2020-01-16
Improve type definitions for mdast.

Zemnmez:
- Replaces tests to be based off the examples in the mdast spec

- Adds `null` to all potentially empty fields -- as noted mdast
interchangeably uses `null` and `undefined` for 'nullable' fields.

- Split out `List` into `OrderedList` and `UnorderedList` to better
reflect conditional fields

- Make `Root` a union of `RootT` instances (the old root interface)
to ensure the consistency in children specified in the spec

- Add JSDoc text and examples from the spec.

- Split `Code` into a union of `CodeWithoutLang` and `CodeWithLang` to better
reflect the availability of the `meta` field.

A key takeaway: before the `null` change, this mdast definition was rejecting the examples in the mdast spec.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/syntax-tree/mdast/blob/e0902c8acfecb9680ee9fa066da725c3a1201774/readme.md#table


types/mdast/index.d.ts
```diff
@@ -1,27 +1,103 @@
...
+// 3. In MDAST's definition, List is a type with rather complex field relationships -- some fields are only
+//    potentially present if other fields have some value. To express this better, I, similarly to 'root'
+//    express List in terms of a type union of OrderedList and UnorderedList.
+//
+// 4. It's definitely arguable that these typings should focus more on ease of use, rather than accurately reflecting
+//    the mdast spec documentation. I've picked somewhat of a middleground by making the most minor changes I can
+//    where I think it's useful.
+
+// explicit exports
+export {};
```
Rokt33r: We probably don't need this. Let's finish discussing in https://github.com/unifiedjs/unified-message-control/pull/7#discussion_r368266506 and come back to this pr later.

Rokt33r (changes requested):

PR #41635, opened 2020-01-16
feat(ioredis): make Command constructable

ethanresnick:
Per the JSDoc [here](https://github.com/luin/ioredis/blob/1d06cf4bd968fd9762b87d7cd3d756c396158ce0/lib/command.ts#L54), Command is an exported, consructable class with the options listed in this PR

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/luin/ioredis/blob/1d06cf4bd968fd9762b87d7cd3d756c396158ce0/lib/command.ts#L54


types/ioredis/index.d.ts
```diff
@@ -57,11 +57,33 @@ declare namespace IORedis {
...
+        /**
+         * Set the encoding of the reply, by default buffer will be returned.
+         */
+        replyEncoding?: string | null;
+        errorStack?: string;
+        keyPrefix?: string;
+    }
+
+    interface CommandStatic {
+        new(
```
SimonSchick: At this point making it a class seems more reasonable.

SimonSchick (changes requested):

types/ioredis/index.d.ts
ethanresnick: Done.

ethanresnick:
Any update on this? I believe I've addressed the feedback. Not sure who I'm supposed to ping. @SimonSchick?

SimonSchick (approved):

types/ioredis/index.d.ts
```diff
@@ -50,18 +50,33 @@ declare class Commander {
...
+}
+
 declare namespace IORedis {
     type KeyType = string | Buffer;
 
     type BooleanResponse = 1 | 0;
 
     type ValueType = string | Buffer | number | any[];
 
-    interface Command {
```
weswigham: You should _probably_ add a `type _Command = typeof Command` outside the IORedis namespace, and then a `type Command = _Command` inside the namespace, so as to not break external users who referenced the `Command` interface, no?

ethanresnick: Thanks @weswigham. Good catch, and I've pushed that change. I was thinking about how/whether to avoid a break actually, but couldn't immediately figure out how to get it to work with namespaces.

PR #41646, opened 2020-01-16
Improve types of highland filter and toPromise

SupernaviX:
`toPromise` now returns an instance of the constructor passed to it, instead of just a `PromiseLike<T>`.
`filter` now accepts a type guard, just like the array filter function.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: http://highlandjs.org/

### DIFF
--- a/types/highland/highland-tests.ts
+++ b/types/highland/highland-tests.ts
@@ -66,6 +66,11 @@ interface StrBarArrMap {
   [key: string]: Bar[];
 }
 
+declare class MyPromise<T> implements PromiseLike<T> {
+  constructor(executor: (resolve: (value: T | PromiseLike<T>) => void, reject: (err: any) => void) => void);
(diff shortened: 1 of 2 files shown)


types/highland/highland-tests.ts
```diff
@@ -243,6 +250,9 @@ fooStream = fooStream.filter((x: Foo) => {
   return bool;
 });
 
+// $ExpectType Stream<Baz>
+fooStream.filter(isBaz);
+
```
NotBobTheBuilder: Could we add one more test case for this?

```js
// $ExpectType Stream<number>
numStream = numStream.filter((n: number) => n > 3);
```

(This was mostly for my own satisfaction that we can use `is` without breaking vanilla predicates that aren't type based)

NotBobTheBuilder (commented):
Looks awesome! I just had one suggestion for an extra test case. The `toPromise` method implementation is really cool.

types/highland/highland-tests.ts
SupernaviX: Sure, makes sense

NotBobTheBuilder (approved):
Thanks!

PR #41650, opened 2020-01-16
[@types/chai] Add sub package for chai v4 without Object.should

villasv:
Please fill in this template.

If adding a new definition:


villasv:
This is another approach to what I proposed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/37571

Because removing the declaration merging of Object.should would break people's code.

villasv:
Do you have any suggestions @jedmao?

villasv:
Will convert to an issue that at least can be forever waiting for discussion

PR #41668, opened 2020-01-17
React phone number input

AdrienEtienne:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/react-phone-number-input/index.d.ts
+++ b/types/react-phone-number-input/index.d.ts
@@ -7,15 +7,13 @@
 
 import * as React from "react";
 
-/// <reference types="react">
 export type NumberFormat = "NATIONAL" | "National" | "INTERNATIONAL" | "International";
 
-export function formatPhoneNumber(value?: string): string;
 export function formatPhoneNumber(value: string, format?: NumberFormat): string;
 export function formatPhoneNumberIntl(value?: string): string;
 export function isValidPhoneNumber(value?: string): boolean;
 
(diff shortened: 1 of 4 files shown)


types/react-phone-number-input/index.d.ts
```diff
@@ -0,0 +1,76 @@
...
+    onChange: (value: string) => void;
+    value: string;
+    // Optional props
+    onFocus?: () => void;
+    onBlur?: () => void;
+    country?: string;
+    metadata?: object;
+}
+
+export interface PhoneInputProps {
```
fleck: from the docs "See the list of all available props for <PhoneInput/>. All properties not listed there will be passed through to the phone number <input/> component." 
The props for this should also take any valid `<input/>` props.

```typescript
export interface PhoneInputProps extends Omit<React.InputHTMLAttributes<HTMLInputElement>, "onChange" | "id"> {
```

AdrienEtienne: Thanks.
Have created a PR for that: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41898

JeremyKirkham:
@AdrienEtienne  - also missing `numberInputProps` optional property.

AdrienEtienne:
> @AdrienEtienne - also missing `numberInputProps` optional property.

Thank you for your comment.
See this PR should fix that: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41957

PR #41675, opened 2020-01-17
Revert "`history`: switch default `LocationState` generic from `any` to `unknown`"

### LINKED ISSUE #41674: Property 'from' does not exist on type '{}'
I now get compilation errors in a project where i use react-router-dom. react-router-dom is using createBrowserHistory() and I'm accessing location.state in my code.

Can I prevent the TypeScript compiler from caring about this?

Error introduced by the following change:
#41580

  - Authors: @sergey-buturlakin, @ngbrown, @rokoroku, @danielnixon 

(edit: grammar)

danielnixon:
Reverts DefinitelyTyped/DefinitelyTyped#41580 to fix https://github.com/DefinitelyTyped/DefinitelyTyped/issues/41674

CC @johnnyreilly @OliverJAsh @fr3gu

### DIFF
--- a/types/history/history-tests.ts
+++ b/types/history/history-tests.ts
@@ -9,8 +9,6 @@ let input = { value: "" };
 {
     let history = createBrowserHistory<{ some: 'state' }>();
 
-    history.location.state; // $ExpectType { some: "state"; }
-
     // Listen for changes to the current location. The
     // listener is called once immediately.
     let unlisten = history.listen(function (location) {
@@ -116,8 +114,7 @@ let input = { value: "" };
 }
 
 {
-    const anything: any = {};
-    const eventTarget: EventTarget = anything;
+    let eventTarget: EventTarget;
     DOMUtils.addEventListener(eventTarget, 'onload', function (event) { event.preventDefault(); });
     DOMUtils.removeEventListener(eventTarget, 'onload', function (event) { event.preventDefault(); });
     DOMUtils.getConfirmation('confirm?', (result) => console.log(result));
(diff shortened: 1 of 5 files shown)


johnnyreilly:
It's fine to change behaviour back to a previous version of type definitions. But what I'd like to avoid is changes that go back and forth multiple times (this has happened previously, particularly with React definitions)

@danielnixon could you provide a clear rationale for this please?

@OliverJAsh could you respond and let's see where this goes?

Thanks all.

danielnixon:
> @danielnixon could you provide a clear rationale for this please?

It's a breaking change (one that I'm sympathetic with, don't get me wrong, `any` is terrible). And it's a breaking change that I suspect is going to break a lot of code.

On the topic of breaking type definitions, the TypeScript lead has this to say (https://github.com/DefinitelyTyped/DefinitelyTyped/issues/25677#issuecomment-388215959):

> Our only guidance is "Don't break people if you don't have to" which is generally naturally adhered to.

OliverJAsh:
As I said in the original PR, I'm happy to wait for a new major version
before we ship this change—the PR was merged before we had a chance to
discuss it any further. But I suspect it would be difficult to coordinate
those changes to go out together, as one release, so then I think we may as
well just ship it now.

johnnyreilly:
Thanks for responding both. @oliverjash it's not clear to me what you mean by:

> But I suspect it would be difficult to coordinate those changes to go out together, as one release, so then I think we may as well just ship it now.

Are you in favour of merging this PR? Or are you saying, "since the previous PR has been merged we may as well leave as is"?

OliverJAsh:
Sorry I'm on mobile so have to be brief.

In the long term I'm keen for the type to be corrected (to unknown), as
that provides type safety.

Whether that happens now or later (i.e. when updating the types to support
a new major version of history), I don't mind, but in my experience it's
too difficult to coordinate breaking changes so that they all go out
together. How would we remember to make this change when updating the types
to the next major version of history? I don't think there's any way to
coordinate these things. If we forget to make the change then, we'll have
to wait for another major version to come along, which could take a very
long time.

Because of that, I think it's easier just to make the change now (i.e.
close this revert and keep it as unknown).

uniqueiniquity (changes requested):
Marking as changes requested since it seems like this PR has been requested to be closed.

PR #41678, opened 2020-01-17
fix:(rdf-js) relax input params of dataset and dataset factory

tpluscode:
I found that the dataset with a derived quad (eg. `Dataset<QuadExt>`) prevents it from being used with the base quads (such as coming from generic parsers or the default data factory). For example:

```ts
import factory from 'rdf-data-model'
import Parser from '@rdfjs/parser-n3'
import rdf from 'rdf-ext'

const quad = factory.quad(/* ... */)
const parser = new Parser()

// both invocations below will fail because Quad is not assignable to QuadExt
const dataset = rdf.dataset([ quad ])
    .import(parser.import(string))
```

I think this is incorrect, and empirically verified in the case of `rdf-ext` that any quad produced by other datasets can be successfully added into instances of `DatasetExt`. 
That should also be the requirement to ensure compatibility between spec implementations.

Hence, I propose to relax the input parameters of `Dataset` and `DatasetFactory` methods.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://rdf.js.org/dataset-spec/#dataset-interface>


tpluscode:
Also cc @thewilkybarkid

rubensworks (changes requested):
I understand the problem, but I am not so fond of the solution.
By removing the generics from methods such as Dataset.add, (TS) implementors may run into issues.

I think an alternative solution may be to do something like the following:
```
add<Q extends BaseQuad>(quad: Q): this;
```

This is however exactly the opposite of #41651, which makes me question whether we really want to make the change suggested in #41651.

tpluscode:
Please reflect on my statement from the PR description:

> I think this is incorrect, and empirically verified in the case of rdf-ext that any quad produced by other datasets can be successfully added into instances of DatasetExt.

How should it be? Should `DatasetExt` only safely accept its specialised quads as params?

Again, should this be allowed?

```
import { quad } from 'rdf-data-model' // creates Quad
import rdf from 'rdf-ext'

const dataset: Dataset<QuadExt> = rdf.dataset()

// error?
dataset.add(quad())
```

@rubensworks @rubenverborgh @LaurensRietveld  @bergos

----

If this is not the expected behaviour (ie. handling more basic quads in specialised stores), then I conclude that the `factory` parameter more strictly to enforce the correct type of `Quad` within a project.

tpluscode:
Okay, I think I find my own answer. The code above fails because `Quad` is not assignable to `QuadExt`. Remember though that it is legally (at least as far a language in concerned) to cast a the dataset to its base form. Nothing changes at runtime but suddenly the code becomes "correct":

```diff
import { quad } from 'rdf-data-model' // creates Quad
import rdf from 'rdf-ext'

+// it is possible to downcast a more specialised
-const dataset: Dataset<QuadExt> = rdf.dataset()
+const dataset: Dataset<Quad> = rdf.dataset()

// error disappears?
dataset.add(quad())
```

Hence, I conclude that my changes are not wrong...

rubensworks:
@tpluscode I fully agree with your [https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41678#issuecomment-575687652](comment). It's just the implementation of the solution I have problems with.

See my review above for an alternative solution. Could you share your thoughts on that?

tpluscode:
> It's just the implementation of the solution I have problems with.

I don't understand your reservation TBH.

> Could you share your thoughts on that?

Of course. Adding generic type arguments to individual methods changes nothing.

```ts
add<Q extends BaseQuad>(quad: Q): this;
```

A method defined like this is exactly the same as `add(quad: BaseQuad)`. It will be legal to call it with any quad which inherits `BaseQuad`. So why obscure the interface?

thewilkybarkid:
Only quickly read this, but looks similar to https://github.com/linkeddata/rdflib.js/issues/374#issuecomment-564544961.

types/rdf-js/index.d.ts
```diff
@@ -472,29 +472,29 @@ export interface DatasetCore<Q extends BaseQuad = Quad> {
...
+export interface DatasetCoreFactory<Q extends BaseQuad = Quad, D extends DatasetCore<Q> = DatasetCore<Q>> {
     /**
      * Returns a new dataset and imports all quads, if given.
      */
-    dataset(quads?: Q[]): DatasetCore<Q>;
+    dataset(quads?: BaseQuad[]): D;
 }
 
-export interface Dataset<Q extends BaseQuad = Quad> extends DatasetCore<Q> {
+export interface Dataset<Q extends BaseQuad = Quad> extends DatasetCore<Q>, Iterable<Q> {
```
thewilkybarkid: `Iterable<Q>` would be on `DatasetCore`? (But seems an unrelated change.)

tpluscode:
Okay, let me summarise for clarity:

RDF/JS spec only defines a single type of Quad/Term/etc. A dataset is a collection of standard-compliant quads. Nothing more and nothing less.
Any custom dataset must accept quads from other implementations, as long as they implement the specification.

This is the only way to keep interoperability. `@types/rdf-js` introduces generics to have more accurate **design-time** types but at runtime all quads are equal in terms of the specification.

types/rdf-js/index.d.ts
tpluscode: Well spotted. I meant to revert that. Thanks.

types/rdf-js/index.d.ts
```diff
@@ -615,14 +615,14 @@ export interface Dataset<Q extends BaseQuad = Quad> extends DatasetCore<Q> {
...
+    union(quads: Dataset<BaseQuad>): this;
 }
 
-export interface DatasetFactory<Q extends BaseQuad = Quad> extends DatasetCoreFactory<Q> {
+export interface DatasetFactory<Q extends BaseQuad = Quad, D extends Dataset<Q> = Dataset<Q>> extends DatasetCoreFactory<Q, D> {
     /**
      * Returns a new dataset and imports all quads, if given.
      */
-    dataset(quads?: Dataset<Q>|Q[]): Dataset<Q>;
+    dataset(quads?: DatasetCore<BaseQuad>|BaseQuad[]): D;
```
thewilkybarkid: While I agree that it should be `DatasetCore` rather than `Dataset`, the spec does say `Dataset`. (Though given there's quite a few mistakes in it...)

tpluscode: Fair enough, it does say `Dataset`

types/rdf-dataset-indexed/dataset.d.ts
```diff
@@ -0,0 +1,39 @@
+import * as RDF from 'rdf-js';
+import { Readable } from 'stream';
+
+declare class DatasetIndexed implements RDF.Dataset<RDF.Quad> {
```
thewilkybarkid: Could be `Q extends RDF.Quad = RDF.Quad` and the factory/some of the methods updated?

types/rdf-js/index.d.ts
tpluscode: And I have not change `Dataset#addAll` which has similar signature so best keep consistent.

thewilkybarkid: Ideally this would just be an `Iterable`, but outside the scope here.

tpluscode: Had that thought but that would deviate greatly from the spec

tpluscode:
Any additional thoughts @rubensworks ?

rubensworks:
Ok, I just re-read the whole PR. You can ignore my last suggestion, it indeed doesn't make much sense.
Nevertheless, my issues with the changes in `types/rdf-js/index.d.ts ` remain.

It looks like the problem you are experiencing is related to the type of quads a `Dataset` accepts as *input*.
Right now, `DatasetExt` is defined as `DatasetExt extends Dataset<QuadExt>`. This means that every operation to `DatasetExt` must be done using a `QuadExt`.
Obviously, this causes problems when you for example want to add triples that were created with a different datafactory, which is the issue you refer to in the OP.

To me, it looks like the root cause of this problem is the fact that the current **`Dataset` typings don't make a distinction between *input* quads and *output* quads**.
For instance (if I understand the `DatasetExt` implementation correctly), `DatasetExt` will always produce `QuadExt`'s as *output*. However, it can take any `Quad` as *input*.

This means that the problem may be solvable by **adding two generic types to `Dataset`**.
For instance: `DatasetCore<QIn extends BaseQuad = Quad, QOut extends BaseQuad = Quad>`.
Where `DatasetExt` could then be defined as `DatasetExt extends Dataset<QIn extends BaseQuad = Quad, QIn extends BaseQuad = QuadExt>`.

This should avoid the need to use raw `BaseQuad` types in certain methods, which is a thing I'd like to avoid at all costs, as strict type checking for more specific quad types is lost then.

@tpluscode Does this clarify my problem to you?

tpluscode:
> Right now, DatasetExt is defined as DatasetExt extends Dataset<QuadExt>. This means that every operation to DatasetExt must be done using a QuadExt.

Yes, that is the current limitation, which I think is incorrect.

> To me, it looks like the root cause of this problem is the fact that the current Dataset typings don't make a distinction between input quads and output quads.

I do not think they should. Like I said above, a dataset is a collection of quads. It is nice to have a generic type argument to make types more specific but at its core the `DatasetExt` is an implementation of `DatasetCore<Quad>`. Again, do consider this line:

```ts
const dsExt: DatasetCore = $rdf.dataset();
```

It works currently (with `DatasetExt extends Dataset<Quad>`) and it should keep being valid when `DatasetExt extends Dataset<QuadExt>`.

> `DatasetExt` will always produce `QuadExt`'s as output. However, it can take any `Quad` as input.

Any dataset should take any `(Base)Quad` as input. That is the promise of interoperability between independent RDF/JS Dataset components/implementations.

While the approach with two generic type arguments will work, I think it's unnecessary and allows for noncompliant declaration like `Dataset<QuadExt, QuadExt>`.

> This should avoid the need to use raw BaseQuad types in certain methods, which is a thing I'd like to avoid at all costs, as strict type checking for more specific quad types is lost then.

This might be possible. `BaseQuad` is indeed hardly used anywhere

thewilkybarkid:
> Any dataset should take any `(Base)Quad` as input.

That's true, but the typing will depend on whether the implementation will convert input into the expected output type.  If I have a `Dataset<CustomQuad>` (where `CustomQuad` is the same as `Quad` but does not take variables), and I `add()` a generic `Quad` containing a variable, what happens? Unless the implementation will convert this `Quad` into a `CustomQuad`, the type would now be `Dataset<CustomQuad | Quad>`...

tpluscode:
> That's true, but the typing will depend on whether the implementation will convert input into the expected output type

Nope. The typing is just a reflection of the specification. We add stuff which makes the declarations more convenient but otherwise any dataset must work correctly when treated as `Dataset<Quad>`.

tpluscode:
Otherwise I shall question the entire idea of datasets being generic and reduce them all to `Dataset<Quad>` indeed to make all quads equal

rubensworks:
> Any dataset should take any (Base)Quad as input. That is the promise of interoperability between independent RDF/JS Dataset components/implementations.

Sure, that is the basis. However, the TS definitions allow developers to do *more*. If people want to create their own `BasedQuad` type (for example, only allow NamedNode in S, P, and O), they should be able to restrict structures such as `Dataset` to work specifically for their custom types.

Right now, I'm just talking about the high-level rdf-js typings for `Dataset`. If you want to create more specific typings for `DatasetExt`, go ahead, I just want to keep the common high-level typings as flexible as possible.

> While the approach with two generic type arguments will work, I think it's unnecessary and allows for noncompliant declaration like Dataset<QuadExt, QuadExt>.

This looks compliant to me. If people want to create a `Dataset` that is restricted to `QuadExt` as I/O, and have strong type checks based on that, they should be able to do so.

tpluscode:
Fair enough. Let me do that

tpluscode (commented):
Okay @rubensworks, do have a look now. Based on your suggestions here's what I got, with minor differences:

1. reversed the order of in/out so that new argument(s) come last. his way will not break existing types
2. by default `QIn` is the same as `QOut`. not exactly what I wanted but again, least impactful and I still can get the result I want with rdf-ext

types/rdf-js/index.d.ts
```diff
@@ -469,32 +469,32 @@ export interface DatasetCore<Q extends BaseQuad = Quad> {
      */
     match(subject?: Term | null, predicate?: Term | null, object?: Term | null, graph?: Term | null): this;
 
-    [Symbol.iterator](): Iterator<Q>;
+    [Symbol.iterator](): Iterator<OutQuad>;
 }
 
-export interface DatasetCoreFactory<Q extends BaseQuad = Quad> {
+export interface DatasetCoreFactory<OutQuad extends BaseQuad = Quad, InQuad extends BaseQuad = BaseQuad, D extends DatasetCore<OutQuad, InQuad> = DatasetCore<OutQuad, InQuad>> {
```
thewilkybarkid:
```suggestion
export interface DatasetCoreFactory<OutQuad extends BaseQuad = Quad, InQuad extends BaseQuad = OutQuad, D extends DatasetCore<OutQuad, InQuad> = DatasetCore<OutQuad, InQuad>> {
```

types/rdf-js/index.d.ts
```diff
@@ -469,32 +469,32 @@ export interface DatasetCore<Q extends BaseQuad = Quad> {
      */
     match(subject?: Term | null, predicate?: Term | null, object?: Term | null, graph?: Term | null): this;
```
rubensworks: I suggest to refer to the terms of `InQuad` here as well.

rubensworks (changes requested):
Good point on backwards-compat! I agree with these changes. Just one minor comment.

types/rdf-js/index.d.ts
tpluscode: Like this?

```ts
match(
    subject?: PropType<InQuad, 'subject'> | null,
    predicate?: PropType<InQuad, 'predicate'> | null,
    object?: PropType<InQuad, 'object'> | null,
    graph?: PropType<InQuad, 'graph'> | null): this;
```

thewilkybarkid: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/40821 was rejected (but still makes sense to me).

rubensworks: @tpluscode Something like `InQuad['subject']` is a bit shorter, and consistent with #40821.

@thewilkybarkid Good point, it's the same issue. Let's try to figure this out...
In any case, whatever option we choose, I think it's important to be consistent between `Dataset` and `Store`.

So the question is if incompatible quad subtypes `MyQuad1` and `MyQuad2` are created, should this be allowed?:
```
const source = new MySource<MyQuad1>();
const matchQuad = new SomeQuad<MyQuad2>();
source.match(matchQuad.subject, matchQuad.predicate, matchQuad.object);
```

In hindsight, perhaps it doesn't make much sense to allow this. What do you think @tpluscode @thewilkybarkid?

tpluscode: Yea, I think I stand by your comment from #40821. Not because you might have different kinds of quads (not really sold on that still) but because in code you might have a `Term` and declarations will prevent it from being used to match:

```ts
// any term we got from other lib or whatever
let subject: Term
let dataset: Dataset

// restricting match will cause this to fail
dataset.match(term)
```

thewilkybarkid: Are there any genuine cases for not knowing what type of `Term` you have?

tpluscode: It may just be inaccurate types in other, potential 3rd party code. Limiting the match method would likely force a bunch of `if (node.termType === 'NamedNode')` kind of statement to ensure matching terms. I expect this to quickly go out of hand with little benefit

rubensworks: > It may just be inaccurate types in other, potential 3rd party code.

IMO, we should assume that typings will be accurate and correct.
Otherwise we risk moving more in the direction of weakly-typed/typeless JS code.

> would likely force a bunch of if (node.termType === 'NamedNode')

Except for the inaccurate typings case you mentioned, I don't expect this to happen (unless desired by the developer).

tpluscode: I'm not sure it's a good idea, and also seems unrelated to the current changes. Agree to try in a separate PR?

rubensworks: Sure, a separate PR sounds good.

rubensworks (approved):

types/rdf-dataset-indexed/dataset.d.ts
```diff
@@ -0,0 +1,39 @@
+import * as RDF from 'rdf-js';
+import { Readable } from 'stream';
+
+declare class DatasetIndexed<Q extends RDF.Quad = RDF.Quad> implements RDF.Dataset<Q> {
```
thewilkybarkid: `InQuad` should be `BaseQuad`. But more importantly, it only implements `DatasetCore` doesn't it? There are some differences, hence https://github.com/rdf-ext/rdf-dataset-indexed/pull/3.

Might be best to extract (most of) this change.

tpluscode: > InQuad should be BaseQuad.

👍 

> But more importantly, it only implements DatasetCore doesn't it?

TBH, I'm not sure. The implementation is similar to that of `rdf-ext` and the API surface does look fairly compatible with `Dataset` 🤷‍♂

thewilkybarkid: They're not all implemented (eg `toCanonical`, `union`). I want to say I found some with different signatures too, but was a while ago when I tried to use it.

tpluscode: Do you propose to go back to `DatasetCore` and keep it that way until rdf-ext/rdf-dataset-indexed#3 is merged?

thewilkybarkid: Yeah. Can keep the methods though, as long as they match what it actually implements.

tpluscode: Shoot, you're right. The methods are not exactly the same. Let me revisit that

thewilkybarkid: Better to do these in a separate PR IMO (unrelated changes).

PR #41692, opened 2020-01-17
fix: add PropsWithChildren to ExoticComponent

maddhruv:
Extends `ExoticComponent` (`memo`) with `PropsWithChildren`

Please fill in this template.

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -22,6 +22,7 @@
 //                 Sebastian Silbermann <https://github.com/eps1lon>
 //                 Kyle Scully <https://github.com/zieka>
 //                 Cong Zhang <https://github.com/dancerphil>
+//                 Dhruv Jain <https://github.com/maddhruv>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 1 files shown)


eps1lon:
We actually want to remove implicit children since not every component handles children. Could you clarify what you want to accomplish here by adding a test?

maddhruv:
@eps1lon Just like how `FunctionComponent` accepts props with children, wanted to make sure `ExoticComponent` also has the same nature.
Faced a problem while using `React.memo` and I explicitly didn't declare the `children` prop so had to add it manually.

> We actually want to remove implicit children

If that is the story, we should remove it from the `FunctionComponent` as well

eps1lon:
> If that is the story, we should remove it from the FunctionComponent as well

It's a breaking change and one that impacts a lot of code. We have the loose plan of removing them when react 17 comes out to align breaking changes. Until then we don't want to repeat our mistakes.

maddhruv:
> Until then we don't want to repeat our mistakes.

can we not release this as a patch, was causing me trouble
now I have to explicitly write type for `children`

eps1lon:
That is the point. You handle children explicitly so you should type them explicitly.

maddhruv:
Then what about the `FunctionComponent` can we remove from there as well or will remove when React 17 lands

uniqueiniquity (changes requested):
Marking as changes requested until @eps1lon's concerns are satisfied.

maddhruv:
@eps1lon Can you have a look at it again?

eps1lon (changes requested):
Still rejecting since we don't want to add more implicit children. We will eventually remove them and don't want to add more features that will be eventually removed.

PR #41695, opened 2020-01-17
[parse] Add overload for Parse.Cloud.define

rdhelms:
Please fill in this template.

-----

Previously, `Parse.Cloud.define<() => void>(request => {})` would result in `request.params` being typed as `undefined`. This PR instead causes it to be `{}` in that situation, which is more accurate.

### DIFF
--- a/types/parse/index.d.ts
+++ b/types/parse/index.d.ts
@@ -1107,6 +1107,10 @@ namespace Parse {
         function afterFind(arg1: any, func?: (request: AfterFindRequest) => any): void;
         function beforeLogin(func?: (request: TriggerRequest) => any): void;
         function define(name: string, func: (request: FunctionRequest) => any): void;
+        function define<T extends () => any>(
+            name: string,
+            func: (request: FunctionRequest<{}>) => Promise<ReturnType<T>> | ReturnType<T>
+        ): void;
         function define<T extends (
             param: { [P in keyof Parameters<T>[0]]: Parameters<T>[0][P] }
         ) => any>(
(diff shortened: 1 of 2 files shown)


types/parse/index.d.ts
```diff
@@ -1107,6 +1107,10 @@ namespace Parse {
         function afterFind(arg1: any, func?: (request: AfterFindRequest) => any): void;
         function beforeLogin(func?: (request: TriggerRequest) => any): void;
         function define(name: string, func: (request: FunctionRequest) => any): void;
+        function define<T extends () => any>(
```
RaschidJFR: Hey @rdhelms I just bumped into this when updated a project. Just curious: why `T extends () => any` instead of just `T = any`?. If it's not sensitive I can add another overload in my next push, just to make a bit shorter the syntax.

RaschidJFR: Oh I think the reason is in `parse-tests.ts:578`: 

```
Parse.Cloud.define<(params: { something: string }) => number>('AFunc', (request) => {
```

rdhelms: Yep the goal here was to have `define`'s type parameter be an actual function type (not just `any`) with the same structure as `Parse.Cloud.run`'s type parameter. We wanted to be able to define the type of `request.params` and the expected return type with one type parameter.

Our idea was to be able to create a "contract" of sorts between the client and the server with these function types. We have a npm package where we write our function types, and then we import and pass those types into `Parse.Cloud.run` on the client and into `Parse.Cloud.define` on the server. If the function type changes, it will point us to the necessary changes on both client and server.

PR #41709, opened 2020-01-18
Added "react-native" TVEventHandler class definition

reallynotburner:
React-Native provides a TVEventHandler class that works on Android and iOS.

Please fill in this template.

- [ X ] Use a meaningful title for the pull request. Include the name of the package modified.
- [ X ] Test the change in your own code. (Compile and run.)
- [ X ] Add or edit tests to reflect the change. (Run with `npm test`.)
- [ X ] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [ X ] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [ X ] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

- [ X ] Provide a URL to documentation or source code which provides context for the suggested changes: `https://github.com/facebook/react-native/blob/72890965f34ecf87add998131b2c6a6fe3d54c8b/Libraries/Components/AppleTV/TVEventHandler.js`
- [ - ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ - ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -5035,6 +5035,11 @@ export class TouchableNativeFeedback extends TouchableNativeFeedbackBase {
     static canUseNativeForeground(): boolean;
 }
 
+export class TVEventHandler {
+  enable(context: any, callback: Function): Function;
+  disable: Function;
+}
+
 export interface Route {
     component?: React.ComponentType<any>;
     id?: string;
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -30,6 +30,7 @@
 //                 Valentyn Tolochko <https://github.com/vtolochk>
 //                 Sergey Sychev <https://github.com/SychevSP>
 //                 Kelvin Chu <https://github.com/RageBill>
+//                 Joshua Brown <https://github.com/reallynotburner>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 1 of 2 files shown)


types/react-native/index.d.ts
```diff
@@ -5035,6 +5036,11 @@ export class TouchableNativeFeedback extends TouchableNativeFeedbackBase {
     static canUseNativeForeground(): boolean;
 }
 
+export class TVEventHandler {
+  enable(context: any, callback: Function): undefined;
```
Naturalclar: is there a specific reason to use `undefined` instead of `void`?
```suggestion
  enable(context: any, callback: Function): void;
```

types/react-native/index.d.ts
```diff
@@ -5035,6 +5036,11 @@ export class TouchableNativeFeedback extends TouchableNativeFeedbackBase {
...
+  disable(): undefined;
```
Naturalclar:
```suggestion
  disable(): void;
```

Naturalclar (changes requested):
@reallynotburner the change LGTM :)
made a few comments.
could you resolve the conflict and see my comments?

alloy (changes requested):
Agreed with @Naturalclar’s review :+1:

PR #41710, opened 2020-01-19
[chart.js] Add types of `ChartPoint`, `ChartData.labels`.

KentarouTakeda:
`Moment` can be specified for chartpoint in addition to `number`, `string` and `Date`.
Also specify it in labels.

---

Please fill in this template.

  * https://github.com/chartjs/Chart.js/blob/v2.9.3/test/specs/scale.time.tests.js#L134  
  * https://www.chartjs.org/samples/latest/scales/time/line.html - source
  * https://www.chartjs.org/docs/latest/axes/cartesian/time.html#date-adapters  
    > By default, Chart.js includes an adapter for **Moment.js** .

### DIFF
--- a/types/chart.js/chart.js-tests.ts
+++ b/types/chart.js/chart.js-tests.ts
@@ -1,4 +1,5 @@
 import { BorderWidth, Chart, Point, ChartColor } from 'chart.js';
+import moment = require('moment');
 
 // alternative:
 // import chartjs = require('chart.js');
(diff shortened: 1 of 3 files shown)


lekoaf:
Is it wise to have Moment as a hard dependency when this is in the documentation?
> The time scale requires both a date library and corresponding adapter to be present. By default, Chart.js includes an adapter for Moment.js. **You may wish to exclude moment and choose from other available adapters instead.**

KentarouTakeda:
Thanks for the review.

I once thought so. But Chart.js itself [contains Moment as hard](https://github.com/chartjs/Chart.js/blob/v2.9.3/package.json#L67) . To remove it, [opt out by webpack.config.js](https://www.chartjs.org/docs/latest/getting-started/integration.html#bundlers-webpack-rollup-etc) is required. For this reason, I thought that type definitions should be hard dependencies.

By the way, [chartjs-adapter-moment](https://github.com/chartjs/chartjs-adapter-moment) was recently created. Chart.js seems to be making its reliance on Moment optional. If this adapter is required when using Moment (it is not required now), then it is reasonable to move your dependency on Moment to this adapter.

simonbrunel:
@KentarouTakeda I'm late to react but it looks like a breaking change. moment **is an optional dependency**, even if Chart.js provides a build with moment in it. The problem with this change is that it forces everyone to change the module resolution method in projects that don't need it (e.g. `chartjs-plugin-datalabels`).

PR #41718, opened 2020-01-19
yargs: Add 'sort-commands' option to `parserConfiguration`

cameronhunter:
 
- http://yargs.js.org/docs/#api-parserconfigurationobj
- https://github.com/yargs/yargs/blob/05324b426025f7bc52b0caeb0794dd61abdd6845/docs/api.md#parserconfigurationobj
- https://github.com/yargs/yargs/blob/63b3dd31a455d428902220c1992ae930e18aff5c/lib/usage.js#L209

Although it looks as if this belongs in `yargs-parser`, it's handled directly in `yargs` (see last code link)

### DIFF
--- a/types/yargs/index.d.ts
+++ b/types/yargs/index.d.ts
@@ -427,7 +427,10 @@ declare namespace yargs {
         parsed: DetailedArguments | false;
 
         /** Allows to configure advanced yargs features. */
-        parserConfiguration(configuration: Partial<Configuration>): Argv<T>;
+        parserConfiguration(configuration: Partial<Configuration & {
+            /** Sort commands alphabetically. Default is `false` */
+            'sort-commands': boolean;
+        }>): Argv<T>;
 
         /**
          * Similar to `config()`, indicates that yargs should interpret the object from the specified key in package.json as a configuration object.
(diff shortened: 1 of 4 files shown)


types/yargs/index.d.ts
```diff
@@ -427,7 +427,10 @@ declare namespace yargs {
         parsed: DetailedArguments | false;
 
         /** Allows to configure advanced yargs features. */
-        parserConfiguration(configuration: Partial<Configuration>): Argv<T>;
+        parserConfiguration(configuration: Partial<Configuration & {
+            /** Sort commands alphabetically. Default is `false` */
+            'sort-commands': boolean;
```
pushplay: Intersection type in the generic type param is kind of gross.  Can we pull it out and name it to clean this up?

pushplay (changes requested):
I think the intersection type in the generic type param should be cleaned up.

types/yargs/index.d.ts
cameronhunter: Sure thing. I've pulled it out to a defined type instead of being inline. Thanks for the feedback.

pushplay (approved):
Great.  Thanks for the clean up.

PR #41725, opened 2020-01-20
Add preset to Stats object

nikuda:
It's possible to use preset in the Stats objects itself as base to extend/enable other flags on:

https://webpack.js.org/configuration/stats/#statspreset

Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/webpack/index.d.ts
+++ b/types/webpack/index.d.ts
@@ -1423,6 +1423,8 @@ declare namespace webpack {
             warningsFilter?: string | RegExp | Array<string | RegExp> | ((warning: string) => boolean);
             /** Show performance hint when file size exceeds `performance.maxAssetSize` */
             performance?: boolean;
+            /** Sets the preset for the type of information that gets displayed */
+            preset?: Preset;
             /** Show the exports of the modules */
             providedExports?: boolean;
         }


ofhouse (changes requested):
@nikuda Thanks for your PR!
The change you want to add seems to be a new option in webpack v5 which is **not** available in v4.

See v4 documentation of `Stats` object (no `preset` key is present):
- https://v4.webpack.js.org/configuration/stats/
- https://github.com/webpack/webpack/blob/v4.41.5/declarations/WebpackOptions.d.ts#L1242

vs v5 documentation (where the `preset` key is available):
- https://webpack.js.org/configuration/stats/#statspreset
- https://github.com/webpack/webpack/blob/v5.0.0-beta.12/declarations/WebpackOptions.d.ts#L1607

Since the header of the declaration file [refers to v4](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/f0260dd21f463b22a44d8cf2aa47e3460ac6a9c8/types/webpack/index.d.ts#L1) I think it is not a good idea to add it to the v4 branch.
Since v5 is in beta right now I guess it makes sense to start a v5 version of this declaration file?
 
I'll take a look what it takes to create a v5 declaration (We need to make sure that it don' t get simply overridden, because there are many v4 users out there) version file later today, but if you do not want to wait you can also go ahead and modify it by yourself & update the PR.

ofhouse:
@nikuda I created a PR #41736 which introduces a way to add types for the upcoming v5 release of webpack.
The PR also adds the changes you requested here as an example case so hopefully you are not to worried that I hijacked your PR for this 🤗

ofhouse:
@nikuda This is now fixed in [`@types/webpack@4.41.3`](https://www.npmjs.com/package/@types/webpack):

```ts
import { Configuration } from 'webpack';
import {} from 'webpack/next';

const config: Configuration = {
  stats: {
    preset: 'none',
  },
};
```

nikuda:
Sorry for the late reply @ofhouse, thanks for doing that, looks good to me 👍

PR #41728, opened 2020-01-20
fix(parse/Schema): update options in functions

RaschidJFR:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/parse-community/Parse-SDK-JS/blob/master/src/ParseSchema.js

### DIFF
--- a/types/parse/index.d.ts
+++ b/types/parse/index.d.ts
@@ -114,6 +114,11 @@ namespace Parse {
         error?: Function;
     }
 
+    interface FieldOptions {
+      required?: boolean;
+      defaultValue?: any;
+    }
+
     interface FullOptions {
         success?: Function;
         error?: Function;
@@ -888,18 +893,18 @@ namespace Parse {
 
         /**
          * Static method to get all schemas
-         * @param options Valid options are:
-         * - useMasterKey: In Cloud Code and Node only, causes the Master Key to be used for this request.
-         * - sessionToken: A valid session token, used for making a request on behalf of a specific user.
+         *
+         * @return {Promise} A promise that is resolved with the result when
+         * the query completes.
          */
-        static all(options?: ScopeOptions): Promise<Schema[]>;
+        static all(): Promise<Schema[]>;
 
-        addArray(name: string): this;
-        addBoolean(name: string): this;
-        addDate(name: string): this;
-        addField(name: string, type?: Schema.TYPE): this;
-        addFile(name: string): this;
-        addGeoPoint(name: string): this;
+        addArray(name: string, options?: FieldOptions): this;
+        addBoolean(name: string, options?: FieldOptions): this;
+        addDate(name: string, options?: FieldOptions): this;
+        addField(name: string, type?: Schema.TYPE, options?: FieldOptions): this;
+        addFile(name: string, options?: FieldOptions): this;
+        addGeoPoint(name: string, options?: FieldOptions): this;
 
         /**
          * Adding an Index to Create / Update a Schema
@@ -913,8 +918,8 @@ namespace Parse {
          */
         addIndex(name: string, index: Schema.Index): this;
 
-        addNumber(name: string): this;
-        addObject(name: string): this;
+        addNumber(name: string, options?: FieldOptions): this;
+        addObject(name: string, options?: FieldOptions): this;
 
         /**
          * Adding Pointer Field
@@ -922,9 +927,9 @@ namespace Parse {
          * @param targetClass  Name of the target Pointer Class
          * @return Returns the schema, so you can chain this call.
          */
-        addPointer(name: string, targetClass: string): this;
+        addPointer(name: string, targetClass: string, options?: FieldOptions): this;
 
-        addPolygon(name: string): this;
+        addPolygon(name: string, options?: FieldOptions): this;
 
         /**
          * Adding Relation Field
@@ -932,20 +937,16 @@ namespace Parse {
          * @param targetClass  Name of the target Pointer Class
          * @return Returns the schema, so you can chain this call.
          */
-        addRelation(name: string, targetClass: string): this;
+        addRelation(name: string, targetClass: string, options?: FieldOptions): this;
 
-        addString(name: string): this;
+        addString(name: string, options?: FieldOptions): this;
 
         /**
          * Removing a Schema from Parse Can only be used on Schema without objects
-         * @param options
-         * Valid options are:
-         * - useMasterKey: In Cloud Code and Node only, causes the Master Key to be used for this request.
-         * - sessionToken: A valid session token, used for making a request on behalf of a specific user.
          * @returns A promise that is resolved with the result when the query completes.
          */
         // @TODO Fix Promise<any>
-        delete(options?: ScopeOptions): Promise<any>;
+        delete(): Promise<any>;
 
         /**
          * Deleting a Field to Update on a Schema
@@ -964,7 +965,7 @@ namespace Parse {
         /**
          * Get the Schema from Parse
          */
-        get(options?: ScopeOptions): Promise<Schema>;
+        get(): Promise<Schema>;
 
         /**
          * Removes all objects from a Schema (class) in Parse. EXERCISE CAUTION, running this will delete all objects for this schema and cannot be reversed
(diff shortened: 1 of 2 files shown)


types/parse/index.d.ts
```diff
@@ -114,6 +114,11 @@ namespace Parse {
         error?: Function;
     }
 
+    interface FieldOptions {
```
JeffGuKang: How about move this under `Schema` namespace?

types/parse/index.d.ts
```diff
@@ -114,6 +114,11 @@ namespace Parse {
...
+      required?: boolean;
+      defaultValue?: any;
+    }
+
```
JeffGuKang:
```
interface FieldOptions<T> {
    required?: boolean;
    defaultValue?: T;
}
```

Generic type can be helpful for `defaultValue`.

types/parse/index.d.ts
```diff
@@ -888,18 +893,18 @@ namespace Parse {
...
-        static all(options?: ScopeOptions): Promise<Schema[]>;
+        static all(): Promise<Schema[]>;
 
-        addArray(name: string): this;
-        addBoolean(name: string): this;
-        addDate(name: string): this;
-        addField(name: string, type?: Schema.TYPE): this;
-        addFile(name: string): this;
-        addGeoPoint(name: string): this;
+        addArray(name: string, options?: FieldOptions): this;
```
JeffGuKang: If we use generic type for `defaultValue`, it can be 
`addArray(name: string, options?: Schema.FieldOptions<[]>): this;`
.

JeffGuKang:
I found this PR and close my own #41801 . This seems to be complementary.

types/parse/index.d.ts
RaschidJFR: I think I'm not following you :/
How does a generic type help here?
I'm seeing the test cases in `parse-tests.ts` from [#41801](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41801/files)  but just not getting it

RaschidJFR: Do you have an example of this use case with/without this feature maybe?

JeffGuKang: If we use a generic type for adding `Schema` functions, it will catch an compile error when another type is used. It will prevent run time error.

```js
addBoolean(name: string, options?: Schema.SchemaOptions<boolean>)
```

Catch an error on TS
```js
schema.addBoolean('boolean', {defaultValue: 'test'}) // DefaultValue should be boolean type.
```

RaschidJFR: Oh gotcha! I'll add it

types/parse/index.d.ts
```diff
@@ -888,18 +888,18 @@ namespace Parse {
...
-        addBoolean(name: string): this;
-        addDate(name: string): this;
-        addField(name: string, type?: Schema.TYPE): this;
-        addFile(name: string): this;
-        addGeoPoint(name: string): this;
+        addArray(name: string, options?: Schema.FieldOptions<any[]>): this;
+        addBoolean(name: string, options?: Schema.FieldOptions<boolean>): this;
+        addDate(name: string, options?: Schema.FieldOptions<Date>): this;
+        addField<T extends Schema.TYPE = any>(name: string, type?: T, options?: Schema.FieldOptions): this;
+        addFile(name: string, options?: Schema.FieldOptions<Parse.File>): this;
```
JeffGuKang: Just File is enough. 
`addFile(name: string, options?: Schema.FieldOptions<File>): this;`

types/parse/parse-tests.ts
```diff
@@ -757,51 +757,102 @@ async function test_local_datastore() {
...
+    notString: Exclude<FieldType, string>,
+    notNumber: Exclude<FieldType, number>,
+    notboolean: Exclude<FieldType, boolean>,
+    notDate: Exclude<FieldType, Date>,
+    notFile: Exclude<FieldType, Parse.File>,
+    notGeopoint: Exclude<FieldType, Parse.GeoPoint[]>,
+    notArray: Exclude<FieldType, any[]>,
+    notObject: Exclude<FieldType, object>,
+    notPointer: Exclude<FieldType, Parse.Pointer>,
+    notPolygon: Exclude<FieldType, Parse.Pointer>
```
JeffGuKang: `Parse.Pointer` looks typo.

types/parse/parse-tests.ts
```diff
@@ -757,51 +757,102 @@ async function test_local_datastore() {
...
 
-    schema.get({ sessionToken: '' }).then(results => {});
+    schema.addObject('field');
+    schema.addObject('field', { defaultValue: { }, required: true });
+    schema.addObject('field', { defaultValue: { abc: 'def' } });
+    // $ExpectError
+    schema.addObject('field', { defaultValue: notObject });
 
-    schema.purge().then(results => {});
+    schema.addPointer('field');
```
JeffGuKang: Expect error by arguments

types/parse/index.d.ts
```diff
@@ -975,16 +971,23 @@ namespace Parse {
...
         /**
          * Update a Schema on Parse
          */
-        update(options?: ScopeOptions): Promise<Schema>;
+        update(): Promise<Schema>;
     }
 
     namespace Schema {
-        type TYPE = string | number | boolean | Date | File | GeoPoint | any[] | object | Pointer | Relation;
+        type TYPE = 'String' | 'Number' | 'Boolean' | 'Date' | 'File' | 'GeoPoint' | 'Polygon' | 'Array' | 'Object' | 'Pointer' | 'Relation';
```
JeffGuKang: The official Parse JS doc is wrong. Do not follow it. 🤕 
It may be better to restore. and I create another PR for wrong documentation of Parse JS.
https://github.com/parse-community/Parse-SDK-JS/pull/1071

types/parse/index.d.ts
```diff
@@ -913,18 +913,18 @@ namespace Parse {
          */
         addIndex(name: string, index: Schema.Index): this;
```
JeffGuKang: `Schema.Index` will make type error cause 983 line.

types/parse/index.d.ts
RaschidJFR: I'm not taking it from the documentation, but from the actual implementation in the SDK: https://github.com/parse-community/Parse-SDK-JS/blob/487bbfb158977a9311a947938b54680538259f1d/src/ParseSchema.js#L15

JeffGuKang: You link is right thing. 
And `index.d.ts` has all types already as GeoPoint, File and so on. So you can use directly.

RaschidJFR: I'm not sure what you mean... the second parameter for `addField()` is not a type but a string!
So one must type, for example:

```ts
schema.addField('defaultFieldString', 'Number');
```
and not:
```ts
schema.addField('defaultFieldString', number);
```

types/parse/index.d.ts
RaschidJFR: I didn't get this one... what should I write here?

types/parse/index.d.ts
JeffGuKang: Oh. You are right. It will be better to make another one for FieldType only.

How about this? 

```ts
type Type = string | number | boolean | Date | File | GeoPoint | any[] | object | Pointer | Relation;
type FieldType = 'String' | 'Number' | 'Boolean' | 'Date' | 'File' | 'GeoPoint' | 'Polygon' | 'Array' | 'Object' | 'Pointer' | 'Relation'
```

And it will fix another error as below comment.

types/parse/index.d.ts
JeffGuKang: It will be fixed naturally when you fix the 983 line as `type TYPE = 'String' | 'Number' | 'Boolean' | 'Date' | 'File' | 'GeoPoint' | 'Polygon' | 'Array' | 'Object' | 'Pointer' | 'Relation'`

types/parse/index.d.ts
RaschidJFR: That's already done in line 986. Is that what you mean?

JeffGuKang: > That's already done in line 986. Is that what you mean?
You are already done. I missed it. :)

JeffGuKang:
LGTM

PR #41733, opened 2020-01-20
fix(redux-state-sync): add missing isActionSynced

jaulz:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/AOHUA/redux-state-sync/blob/master/src/syncState.js#L59

### DIFF
--- a/types/redux-state-sync/index.d.ts
+++ b/types/redux-state-sync/index.d.ts
@@ -37,3 +37,4 @@ export function withReduxStateSync(
     appReducer: Reducer
 ): (state: any, action: AnyAction) => Reducer;
 export function initStateWithPrevTab(store: Store): Store;
+export function isActionSynced(action: AnyAction): boolean;


types/redux-state-sync/index.d.ts
```diff
@@ -37,3 +37,4 @@ export function withReduxStateSync(
     appReducer: Reducer
 ): (state: any, action: AnyAction) => Reducer;
 export function initStateWithPrevTab(store: Store): Store;
+export function isActionSynced(action: AnyAction): boolean;
```
weswigham: This is marked as [only exported for test](https://github.com/AOHUA/redux-state-sync/blob/master/src/syncState.js#L58) in the library source; I'm not sure we should be blessing its external usage by including it in the type definitions for the package.

jaulz: Hm, this is actually not supposed to be marked as "for test only" because its purpose is to check if the incoming action is synced or not. The function was introduced via PR https://github.com/AOHUA/redux-state-sync/pull/48 and you can see one specific use case of it here: https://github.com/AOHUA/redux-state-sync/pull/48#issuecomment-541507549

You can also see that `isActionSynced` is only used in source and not test files: https://github.com/AOHUA/redux-state-sync/search?q=isActionSynced&unscoped_q=isActionSynced

PR #41739, opened 2020-01-20
[multer] add `MulterError`, `File.stream`, better documentation

lojgmyrht:
Added the missing MulterError class. Updated documentation. 

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [multer-error.js](https://github.com/expressjs/multer/blob/master/lib/multer-error.js)


types/multer/index.d.ts
```diff
@@ -1,127 +1,305 @@
...
+             * `fields()`, and `any()` middleware.
+             */
+            files: {
+                [fieldname: string]: Multer.File[];
+            } | Multer.File[];
+        }
+    }
+}
+
+declare class Multer {
```
hlian: @	weswigham I think this class has to be `export`ed, or else there is no way to write this code:


```
import multer from 'multer';

class Foo {
  multerFactory: ???;

  constructor() {
    this.multerFactory = multer({...});
  }  
}
```

Prior to this PR, we had `multer.Instance`, but this patch removes that from the exported namespace

weswigham: Re-adding something like `export type Instance = Multer` is something that could be done.

PR #41742, opened 2020-01-20
[@types/object-fit-images] added new type definitions

rann91:
Please fill in this template.

If adding a new definition:
- [ x ] The package does not already provide its own types, or cannot have its `.d.ts` files generated via `--declaration`
- [ x ] If this is for an NPM package, match the name. If not, do not conflict with the name of an NPM package.
- [ x ] Create it with `dts-gen --dt`, not by basing it on an existing project.
- [ x ] Represents shape of module/library [correctly](https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html)
- [ x ] `tslint.json` should be present and it shouldn't have any additional or disabling of rules. Just content as `{ "extends": "dtslint/dt.json" }`. If for reason the some rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]`  and not for whole package so that the need for disabling can be reviewed.
- [ x ] `tsconfig.json` should have `noImplicitAny`, `noImplicitThis`, `strictNullChecks`, and `strictFunctionTypes` set to `true`.

### DIFF
--- a/types/object-fit-images/index.d.ts
+++ b/types/object-fit-images/index.d.ts
@@ -1,0 +1,11 @@
+// Type definitions for object-fit-images 3.2
+// Project: https://github.com/bfred-it/object-fit-images#readme
+// Definitions by: Ranjit Singh <https://github.com/rann91>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare function objectFillImages(
+    images: string | HTMLElement | HTMLElement[] | NodeList | null,
+    options?: { watchMQ: boolean },
+): void;
+
+export default objectFillImages;
--- a/types/object-fit-images/index.d.ts
+++ b/types/object-fit-images/index.d.ts
@@ -3,9 +3,11 @@
 // Definitions by: Ranjit Singh <https://github.com/rann91>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
+export as namespace objectFillImages;
+
+export = objectFillImages;
+
 declare function objectFillImages(
     images: string | HTMLElement | HTMLElement[] | NodeList | null,
-    options?: { watchMQ: boolean },
+    options?: { watchMQ?: boolean; skipTest?: boolean },
 ): void;
-
-export default objectFillImages;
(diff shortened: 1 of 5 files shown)


types/object-fit-images/index.d.ts
```diff
@@ -0,0 +1,11 @@
+// Type definitions for object-fit-images 3.2
+// Project: https://github.com/bfred-it/object-fit-images#readme
+// Definitions by: Ranjit Singh <https://github.com/rann91>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+declare function objectFillImages(
+    images: string | HTMLElement | HTMLElement[] | NodeList | null,
+    options?: { watchMQ: boolean },
```
peterblazejewicz: There seems to be undocumented `skipTest` option (also bool):
https://github.com/fregante/object-fit-images/blob/master/index.js#L182

types/object-fit-images/index.d.ts
```diff
@@ -0,0 +1,11 @@
...
+): void;
+
+export default objectFillImages;
```
peterblazejewicz: You should also think how this will be consumed, is that only expected to be consumed via ES6 module import?
It looks like browser loaded library with option to be consumed via es6 module, so e.g. could work AFAIK:
```ts
export as namespace objectFillImages;

export = objectFillImages;

declare function objectFillImages(
    images: string | HTMLElement | HTMLElement[] | NodeList | null,
    options?: { watchMQ?: boolean, skipTest?: boolean },
): void;
```

types/object-hash/tsconfig.json
```diff
@@ -6,7 +6,7 @@
         ],
         "noImplicitAny": true,
         "noImplicitThis": true,
-        "strictNullChecks": false,
+        "strictNullChecks": true,
```
peterblazejewicz: This seems like unintentional edit, does it?

rann91:
Thanks for you feedback, @peterblazejewicz. I've updated my PR.

peterblazejewicz (approved):
LGTM

@rann91  thx for the contribution!

PR #41744, opened 2020-01-20
[contract-proxy-kit] Add definition

cag:
Please fill in this template.

<s>- [ ] Test the change in your own code. (Compile and run.)</s>

If adding a new definition:

### DIFF
--- a/types/contract-proxy-kit/index.d.ts
+++ b/types/contract-proxy-kit/index.d.ts
@@ -56,5 +56,8 @@ declare class CPK {
 
     getOwnerAccount(): Promise<string>;
     get address(): string;
-    execTransactions(transactions: Transaction[], options?: TransactionOptions): Promise<TransactionResult>;
+    execTransactions(
+        transactions: ReadonlyArray<Transaction>,
+        options?: TransactionOptions,
+    ): Promise<TransactionResult>;
 }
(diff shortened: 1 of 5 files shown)


types/contract-proxy-kit/index.d.ts
```diff
@@ -0,0 +1,63 @@
...
+// Project: https://github.com/gnosis/contract-proxy-kit#readme
+// Definitions by: Alan Lu <https://github.com/cag>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// Minimum TypeScript Version: 3.6
+
+import { ethers } from 'ethers';
+
+export = CPK;
+
+interface NetworkConfigEntry {
```
uniqueiniquity: If you want your users to be able to import these types from the module, you should wrap them in a namespace called `CPK`.

types/contract-proxy-kit/index.d.ts
```diff
@@ -0,0 +1,63 @@
+// Type definitions for contract-proxy-kit 1.0
+// Project: https://github.com/gnosis/contract-proxy-kit#readme
+// Definitions by: Alan Lu <https://github.com/cag>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// Minimum TypeScript Version: 3.6
```
uniqueiniquity: Is this really the minimum version of TypeScript that is compatible with these type definitions?

uniqueiniquity (changes requested):

types/contract-proxy-kit/index.d.ts
cag: Currently, yes, because accessors were introduced in 3.6. Should I instead be making it a readonly property?

uniqueiniquity: No, that makes sense. Thanks!

uniqueiniquity:

uniqueiniquity (approved):

PR #41754, opened 2020-01-21
[ioBroker] make the options parameter in adapter.getObjectList optional

AlCalzone:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/ioBroker/ioBroker.js-controller/commit/1bd8b3896b074f1a29326341696cd499ce68d1de

### DIFF
--- a/types/iobroker/index.d.ts
+++ b/types/iobroker/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for ioBroker 2.0
+// Type definitions for ioBroker 2.2
 // Project: https://github.com/ioBroker/ioBroker, http://iobroker.net
 // Definitions by: AlCalzone <https://github.com/AlCalzone>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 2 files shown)


types/iobroker/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for ioBroker 2.0
+// Type definitions for ioBroker 2.2
```
uniqueiniquity: The major and minor version of these type definitions should match that of the package they reflect, so I recommend leaving this as 2.0 (unless this is a pre-emptive change for a forthcoming release of iobroker v2.2).

uniqueiniquity (changes requested):

types/iobroker/index.d.ts
AlCalzone: @uniqueiniquity This is indeed for the 2.2 release.

By the way, why is your nickname so hard to type? 😅

uniqueiniquity (approved):

types/iobroker/index.d.ts
uniqueiniquity: I thought it would be fun when I came up with it years ago... clearly wasn't thinking ahead enough. 😆 Might be worth a change someday.

uniqueiniquity: I thought it would be fun when I came up with it years ago... clearly wasn't thinking ahead enough. 😆 Might be worth a change someday.

PR #41758, opened 2020-01-21
[react-twitter-embed] Add definitions

zozoens31:
Please fill in this template.

---
This change is [Reviewable](https://reviewable.io/reviews/definitelytyped/definitelytyped/41758)

### DIFF
--- a/types/react-twitter-embed/index.d.ts
+++ b/types/react-twitter-embed/index.d.ts
@@ -1,0 +1,85 @@
+// Type definitions for react-twitter-embed 3.0
+// Project: https://github.com/saurabhnemade/react-twitter-embed#readme
+// Definitions by: Cyrille <https://github.com/zozoens31>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// Minimum TypeScript Version: 3.5
+import * as React from 'react';
+
+export interface BaseProps {
+    onLoad?: (element: HTMLElement) => void;
+    options?: {};
+    placeholder?: string | React.ReactElement;
+}
+
+export namespace timeline {
+    interface TimelineBaseProps extends BaseProps {
+        autoHeight?: boolean;
+        borderColor?: string;
+        id?: string | number;
+        lang?: string;
+        linkColor?: string;
+        noBorders?: boolean;
+        noFooter?: boolean;
+        noHeader?: boolean;
+        noScrollbar?: boolean;
+        ownerScreenName?: string;
+        screenName?: string;
+        slug?: string;
+        theme?: 'dark' | 'light';
+        transparent?: boolean;
+        url?: string;
+        userId?: number;
+        widgetId?: string;
+    }
+
+    interface ScreenNameSourcesProps {
+        sourceType: 'profile' | 'likes';
+        screenName: string;
+    }
+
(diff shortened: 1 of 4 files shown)


types/react-twitter-embed/index.d.ts
```diff
@@ -0,0 +1,85 @@
...
+// Minimum TypeScript Version: 3.5
+import * as React from 'react';
+
+export interface BaseProps {
+    onLoad?: (element: HTMLElement) => void;
+    options?: {};
+    placeholder?: string | React.ReactElement;
+}
+
+export namespace timeline {
```
uniqueiniquity: Is there a particular reason you need to wrap these types in a namespace?
As far as I can tell, it's not really helping you in any way, and you don't seem to be using it in your test file.

uniqueiniquity (changes requested):

zozoens31 (commented):
> *[Reviewable](https://reviewable.io/reviews/definitelytyped/definitelytyped/41758)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @uniqueiniquity)

___
*[types/react-twitter-embed/index.d.ts, line 14 at r1](https://reviewable.io/reviews/definitelytyped/definitelytyped/41758#-LzBhhkICi_e7_-nWXjc-r1-14:-LzBhuhqA3Syv8U4H-68:b-w5hqaz) ([raw file](https://github.com/definitelytyped/definitelytyped/blob/283c90709ccd625e75a57fbd72acbc838496dadd/types/react-twitter-embed/index.d.ts#L14)):*
<details><summary><i>Previously, uniqueiniquity (Ben Lichtman) wrote…</i></summary><blockquote>

Is there a particular reason you need to wrap these types in a namespace?
As far as I can tell, it's not really helping you in any way, and you don't seem to be using it in your test file.
</blockquote></details>

There's no real reason. I was surprised by the fact that all top lever declaration should be exported, and I didn't want to add confusion to users with all those internal interfaces. What would you suggest would be better? Maybe I can move these to another file, like src/components/TwitterTimelineEmbed.ts, to reflect the JS lib.

uniqueiniquity:
> There's no real reason. I was surprised by the fact that all top lever declaration should be exported, and I didn't want to add confusion to users with all those internal interfaces. What would you suggest would be better? Maybe I can move these to another file, like src/components/TwitterTimelineEmbed.ts, to reflect the JS lib.

I would recommend removing the namespace `timeline` and adding `export {};` to the bottom of the file. This will prevent those types from being exported.

types/react-twitter-embed/index.d.ts
```diff
@@ -0,0 +1,85 @@
+// Type definitions for react-twitter-embed 3.0
+// Project: https://github.com/saurabhnemade/react-twitter-embed#readme
+// Definitions by: Cyrille <https://github.com/zozoens31>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// Minimum TypeScript Version: 3.5
```
uniqueiniquity: Is 3.5 really the minimum version of TypeScript required to compile these types? That seems overly restrictive.

uniqueiniquity (changes requested):

PR #41760, opened 2020-01-21
`react`: `Children` fixes

OliverJAsh:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://reactjs.org/docs/react-api.html#reactchildrenmap

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -2788,6 +2788,8 @@ declare namespace React {
     // ----------------------------------------------------------------------
 
     interface ReactChildren {
+        map(children: null, fn: (child: never, index: number) => never): null;
+        map(children: undefined, fn: (child: never, index: number) => never): undefined;
         map<T, C>(children: C | C[], fn: (child: C, index: number) => T): T[];
         forEach<C>(children: C | C[], fn: (child: C, index: number) => void): void;
         count(children: any): number;
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -2790,7 +2790,7 @@ declare namespace React {
     interface ReactChildren {
         map(children: null, fn: (child: never, index: number) => never): null;
         map(children: undefined, fn: (child: never, index: number) => never): undefined;
-        map<T, C>(children: C | C[], fn: (child: C, index: number) => T): T[];
+        map<T, C>(children: C | C[], fn: (child: C, index: number) => T): Array<Exclude<T, boolean | null | undefined>>;
         forEach<C>(children: C | C[], fn: (child: C, index: number) => void): void;
         count(children: any): number;
         only<C>(children: C): C extends any[] ? never : C;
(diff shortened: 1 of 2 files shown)


types/react/test/index.ts
```diff
@@ -519,13 +519,17 @@ declare const mixedChildren: Array<JSX.Element | string>;
 declare const singlePluralChildren: JSX.Element | JSX.Element[];
 declare const renderPropsChildren: () => JSX.Element;
 
+// $ExpectType null
+const mappedChildrenArray0 = React.Children.map(null, num => num);
+// $ExpectType undefined
+const mappedChildrenArray1 = React.Children.map(undefined, num => num);
```
OliverJAsh: > If children is null or undefined, this method will return null or undefined rather than an array.

https://reactjs.org/docs/react-api.html#reactchildrenmap

types/react/test/index.ts
```diff
@@ -519,13 +519,17 @@ declare const mixedChildren: Array<JSX.Element | string>;
...
 // $ExpectType number[]
 const mappedChildrenArray2 = React.Children.map(numberChildren, num => num);
 // $ExpectType Element[]
 const mappedChildrenArray3 = React.Children.map(elementChildren, element => element);
 // $ExpectType (string | Element)[]
 const mappedChildrenArray4 = React.Children.map(mixedChildren, elementOrString => elementOrString);
-// $ExpectType (string | number | null)[]
+// $ExpectType ReactText[]
```
OliverJAsh:
```ts
console.log(React.Children.map([1, 'foo', null], x => x))
// => [1, 'foo']
```

types/react/index.d.ts
```diff
@@ -2788,11 +2788,13 @@ declare namespace React {
     // ----------------------------------------------------------------------
 
     interface ReactChildren {
-        map<T, C>(children: C | C[], fn: (child: C, index: number) => T): T[];
+        map(children: null, fn: (child: never, index: number) => never): null;
+        map(children: undefined, fn: (child: never, index: number) => never): undefined;
+        map<T, C>(children: C | C[], fn: (child: C, index: number) => T): Array<Exclude<T, boolean | null | undefined>>;
```
eps1lon: Never consciously thought about this :+1: I'm actually a bit scared now considering I always thought this works just like Array.prototype.map.

I feel like the documentation should definitely mention that the result is filtered. I would expect the length of the array to stay the same when mapping i.e. `React.Children.count(children) === React.Children.count(React.Children.map(children, c => c))` should always be `true`.

eps1lon (approved):

types/react/index.d.ts
OliverJAsh: > I feel like the documentation should definitely mention that the result is filtered.

I completely agree. The docs don't describe the full situation :-/

DovydasNavickas (approved):

johnnyreilly:
Nice work!

PR #41761, opened 2020-01-21
use union type instead of intersection type

abdulkareemnalband:
item can be one of given type not all at once

### DIFF
--- a/types/office-js/index.d.ts
+++ b/types/office-js/index.d.ts
@@ -12176,7 +12176,7 @@ declare namespace Office {
          * {@link Office.MessageCompose | MessageCompose}, {@link Office.MessageRead | MessageRead},
          * {@link Office.AppointmentCompose | AppointmentCompose}, {@link Office.AppointmentRead | AppointmentRead}
          */
-        item: Item & ItemCompose & ItemRead & MessageRead & MessageCompose & AppointmentRead & AppointmentCompose;
+        item: Item | ItemCompose | ItemRead | MessageRead | MessageCompose | AppointmentRead | AppointmentCompose;
         /**
          * Gets an object that provides methods to manage the categories master list associated with a mailbox.
          *


AlexJerabek:
@ElizabethSamuel-MSFT or @Zlatkovsky, could you please take a look at this Outlook Item change?

ElizabethSamuel-MSFT:
@abdulkareemnalband Thanks for taking time out to work on this! It's a little unusual to do intersection instead of union but it is our intention to merge all the members of the entire Item object hierarchy as the item property type, hence the [intersection notation](https://www.typescriptlang.org/docs/handbook/advanced-types.html#intersection-types).

As such, we won't take this change at this time.

We appreciate your choosing to develop Office Add-ins and if you come across any other discrepancies or want to suggest improvements, do let us know.

Thanks again!

ElizabethSamuel-MSFT (changes requested):
We would like to leave the property type as is with & instead of |. Thanks.

abdulkareemnalband:
@ElizabethSamuel-MSFT AFAIK item can be one of `ItemCompose , ItemRead , MessageRead , MessageCompose , AppointmentRead , AppointmentCompose`, so it will contain properties/methods of only one type, not all at once.

By using union type we can implement [type garuad](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards) by checking value of `item.itemType` for example
```ts
function isAppointmentCompose(item: Item | ItemCompose | ItemRead | MessageRead | MessageCompose | AppointmentRead | AppointmentCompose): item is Office.AppointmentCompose {
    return item.itemType === Office.MailboxEnums.ItemType.Appointment && !!item.saveAsync;
}
```

with intersection type we cannot do that.

ElizabethSamuel-MSFT:
@abdulkareemnalband That's a good point you raise. However, for now, we will leave it as is since it'll give us the correct IntelliSense behavior (i.e., showing all members available to the item property).

We're in the process of re-organizing Item and its subclasses and will keep in mind your point about type guard.

Thanks.

abdulkareemnalband:
@ElizabethSamuel-MSFT so i will close this PR for now
we could revisit this later

PR #41765, opened 2020-01-21
[@types/async] Add promise-returning overloads for each* functions

wcauchois:
For the async methods `each` and `eachLimit`, and their aliases, if a callback is omitted the function returns a Promise. This PR adds those overloads.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://caolan.github.io/async/v3/docs.html#each

### DIFF
--- a/types/async/index.d.ts
+++ b/types/async/index.d.ts
@@ -114,15 +114,19 @@ export interface AsyncCargo {
 }
 
 // Collections
-export function each<T, E = Error>(arr: IterableCollection<T>, iterator: AsyncIterator<T, E>, callback?: ErrorCallback<E>): void;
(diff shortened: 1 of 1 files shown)


nbibler:
These definitions don't quite cover the implementation correctly. They're requiring either:

```typescript
export function each<T, E = Error>(arr: IterableCollection<T>, iterator: AsyncIterator<T, E>, callback: ErrorCallback<E>): void;
export function each<T, E = Error>(arr: IterableCollection<T>, iterator: AsyncIterator<T, E>): Promise<void>;
```

But ignore that implementation allows for `callback` to be a non-function value (like `undefined`). In that case it returns a `Promise`, as well:

https://github.com/caolan/async/blob/667e71760e7f2dd558e868796bd6e6008a3bac16/lib/internal/awaitify.js#L6-L8

This causes the updated types to incorrectly error when `callback` is an optional/undefinable value.

PR #41766, opened 2020-01-21
Fix Jest Matchers type

mattphillips:
This PR addresses the `Matchers<R>` -> `Matchers<R, T>` change in #39243

The above change is not inline with the types of `Jest` [see code here](https://github.com/facebook/jest/blob/master/packages/expect/src/types.ts#L87) and has broken any libraries extending Jest's `expect` (see: https://github.com/jest-community/jest-extended/issues/234).

The change in this PR adds a default to the new `T` type parameter of `void` which means that other libraries that extend `expect` can write their matchers in terms of `Matchers<R>` or `Matchers<R, T>` (should they need the `T`).

cc/ @tonyhallett

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/jest/blob/master/packages/expect/src/types.ts#L87

### DIFF
--- a/types/jest/index.d.ts
+++ b/types/jest/index.d.ts
@@ -727,7 +727,7 @@ declare namespace jest {
         not: T
     };
     // should be R extends void|Promise<void> but getting dtslint error
-    interface Matchers<R, T = void> {
+    interface Matchers<R, T = {}> {
         /**
          * Ensures the last call to a mock function was provided specific args.
          *
--- a/types/jest/index.d.ts
+++ b/types/jest/index.d.ts
@@ -727,7 +727,7 @@ declare namespace jest {
         not: T
     };
     // should be R extends void|Promise<void> but getting dtslint error
-    interface Matchers<R, T> {
+    interface Matchers<R, T = void> {
         /**
          * Ensures the last call to a mock function was provided specific args.
          *


wsmd (approved):
Thanks for this @mattphillips!

types/jest/index.d.ts
```diff
@@ -727,7 +727,7 @@ declare namespace jest {
         not: T
     };
     // should be R extends void|Promise<void> but getting dtslint error
-    interface Matchers<R, T> {
+    interface Matchers<R, T = void> {
```
uniqueiniquity: It seems like `T` isn't used in return type positions, but `void` only really makes sense as the return type of a function.
If your intent is to have a default with no properties, could you use `{}` here instead? Thanks!

uniqueiniquity (changes requested):

types/jest/index.d.ts
wsmd: Good catch @uniqueiniquity! 

Using `void` is likely to cause issues with `toMatchSnapshot`/`toMatchInlineSnapshot` when `T` remains unspecified. Looks like we're expecting an indexed type here:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c640483291a20a1e7d3ebe1f13a444c697d7be4c/types/jest/index.d.ts#L977

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/c640483291a20a1e7d3ebe1f13a444c697d7be4c/types/jest/index.d.ts#L988

wsmd (changes requested):

types/jest/index.d.ts
SimenB: it was gonna be `= {}` but @orta suggested `= void` instead - maybe he has some thoughts he'd like to share?

orta: Hah, yeah, I recommended both - but I figured `void` would be closer to your average expect-y function type. Seems fine to me to use `{}` instead.

wsmd (approved):

uniqueiniquity (approved):

uniqueiniquity:

PR #41771, opened 2020-01-21
`react`: `Children.map`: replace overload with conditional type

OliverJAsh:
Adds to #41760 

---

Please fill in this template.

### DIFF
--- a/types/react/test/index.ts
+++ b/types/react/test/index.ts
@@ -514,6 +514,7 @@ onlyChild = React.Children.only([null, [[["Hallo"], true]], false]); // error
 const childrenToArray: Array<Exclude<React.ReactNode, boolean | null | undefined>> = React.Children.toArray(children);
 
 declare const numberChildren: number[];
+declare const nodeChildren: React.ReactNode;
 declare const elementChildren: JSX.Element[];
 declare const mixedChildren: Array<JSX.Element | string>;
 declare const singlePluralChildren: JSX.Element | JSX.Element[];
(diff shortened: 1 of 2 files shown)


types/react/test/index.ts
```diff
@@ -533,6 +534,8 @@ const mappedChildrenArray4 = React.Children.map(mixedChildren, elementOrString =
 const mappedChildrenArray5 = React.Children.map(singlePluralChildren, element => element.key);
 // $ExpectType string[]
 const mappedChildrenArray6 = React.Children.map(renderPropsChildren, element => element.name);
+// $ExpectError
+const mappedChildrenArray7 = React.Children.map(nodeChildren, node => node).map;
```
OliverJAsh: Previously this would not error.

eps1lon (commented):
Would this work by sticking with two overloads instead using conditionals?

```ts
map<T, C extends null | undefined>(children: C | C[], fn: (child: C, index: number) => T): C;
map<T, C>(children: C | C[], fn: (child: C, index: number) => T): Array<Exclude<T, boolean | null | undefined>>;
```

OliverJAsh:
@eps1lon I just tried it and it fails the new test (the one added in this PR). It seems to be using the second overload.

eps1lon:
> @eps1lon I just tried it and it fails the new test (the one added in this PR). It seems to be using the second overload.

Well then lets stick with your version.

eps1lon (approved):

PR #41773, opened 2020-01-21
Added react-native-easy-toast

cavarzan:
Please fill in this template.

If adding a new definition:


peterblazejewicz:
@cavarzan 
Q: Is that intended to land here as the referenced package already ships it own types specified in the `package.json`?

cavarzan:
The current library are not being updated for a long time. There is a lot
of PRs fixing the typings but none was merged. The library is stable but
seems to be abandoned by the owner.

Deividi Cavarzan

Em ter, 21 de jan de 2020 19:07, Piotr Błażejewicz (Peter Blazejewicz) <
notifications@github.com> escreveu:

peterblazejewicz (approved):
LGTM
see also comment here:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41773#issuecomment-576908208

@cavarzan Thanks for contribution!

uniqueiniquity (changes requested):
@cavarzan I don't think this package will have the effect that you really want.
If a user has `react-native-easy-toast` and `@types/react-native-easy-toast` installed, the types from the original package will be chosen over those from the `@types` package.

I would suggest forking the original project if it's unmaintained and there are improvements that should be made.

cavarzan:
@uniqueiniquity  OK! I didn't knew that the original package definition has a higher priority rather than `@types` packages. I've considered as override definition just like I use using a local `d.ts` files in the working project folder. Unfortunately, I was unable to override as a local type definition, so I create this PR to possibly fix this.
Indeed the fork and republish the library as a secondary option seems to be a better solution, since we didn't have any updates from the current owner and the MIT license allow the code usage (AFAIK).
Thanks for the observation. I won't delete this PR and let you take the action if you agree.

uniqueiniquity:
Makes sense to me. I'll go ahead and close this PR.

PR #41789, opened 2020-01-22
yup: required fields is not nullable

ekfn:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [link to source code](https://github.com/jquense/yup/blob/820f910fabcb5333986dded0b2443a77e7fd43ff/src/mixed.js#L336)

required fields never return null after validation:
```
return value != null;
```

so `null` should be excluded from types

### DIFF
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -93,7 +93,7 @@ export interface MixedSchema<T = any> extends Schema<T> {
     nullable(isNullable?: true): MixedSchema<T | null>;
     nullable(isNullable: false): MixedSchema<Exclude<T, null>>;
     nullable(isNullable?: boolean): MixedSchema<T>;
-    required(message?: TestOptionsMessage): MixedSchema<Exclude<T, undefined>>;
+    required(message?: TestOptionsMessage): MixedSchema<Exclude<T, null | undefined>>;
     notRequired(): MixedSchema<T | undefined>;
     concat(schema: this): this;
     concat<U>(schema: MixedSchema<U>): MixedSchema<T | U>;
@@ -123,7 +123,7 @@ export interface StringSchema<T extends string | null | undefined = string> exte
     nullable(isNullable?: true): StringSchema<T | null>;
     nullable(isNullable: false): StringSchema<Exclude<T, null>>;
     nullable(isNullable?: boolean): StringSchema<T>;
-    required(message?: TestOptionsMessage): StringSchema<Exclude<T, undefined>>;
+    required(message?: TestOptionsMessage): StringSchema<Exclude<T, null | undefined>>;
     notRequired(): StringSchema<T | undefined>;
 }
 
@@ -145,7 +145,7 @@ export interface NumberSchema<T extends number | null | undefined = number> exte
     nullable(isNullable?: true): NumberSchema<T | null>;
     nullable(isNullable: false): NumberSchema<Exclude<T, null>>;
     nullable(isNullable?: boolean): NumberSchema<T>;
-    required(message?: TestOptionsMessage): NumberSchema<Exclude<T, undefined>>;
+    required(message?: TestOptionsMessage): NumberSchema<Exclude<T, null | undefined>>;
     notRequired(): NumberSchema<T | undefined>;
 }
 
@@ -158,7 +158,7 @@ export interface BooleanSchema<T extends boolean | null | undefined = boolean> e
     nullable(isNullable?: true): BooleanSchema<T | null>;
     nullable(isNullable: false): BooleanSchema<Exclude<T, null>>;
     nullable(isNullable?: boolean): BooleanSchema<T>;
-    required(message?: TestOptionsMessage): BooleanSchema<Exclude<T, undefined>>;
+    required(message?: TestOptionsMessage): BooleanSchema<Exclude<T, null | undefined>>;
     notRequired(): BooleanSchema<T | undefined>;
 }
 
@@ -173,7 +173,7 @@ export interface DateSchema<T extends Date | null | undefined = Date> extends Sc
     nullable(isNullable?: true): DateSchema<T | null>;
     nullable(isNullable: false): DateSchema<Exclude<T, null>>;
     nullable(isNullable?: boolean): DateSchema<T>;
-    required(message?: TestOptionsMessage): DateSchema<Exclude<T, undefined>>;
+    required(message?: TestOptionsMessage): DateSchema<Exclude<T, null | undefined>>;
     notRequired(): DateSchema<T | undefined>;
 }
 
@@ -258,7 +258,7 @@ export interface ObjectSchema<T extends object | null | undefined = object> exte
     nullable(isNullable?: true): ObjectSchema<T | null>;
     nullable(isNullable: false): ObjectSchema<Exclude<T, null>>;
     nullable(isNullable?: boolean): ObjectSchema<T>;
-    required(message?: TestOptionsMessage): ObjectSchema<Exclude<T, undefined>>;
+    required(message?: TestOptionsMessage): ObjectSchema<Exclude<T, null | undefined>>;
     notRequired(): ObjectSchema<T | undefined>;
     concat(schema: this): this;
     concat<U extends object>(schema: ObjectSchema<U>): ObjectSchema<T & U>;
@@ -291,7 +291,7 @@ export interface TestContext {
     parent: any;
     schema: Schema<any>;
     resolve: (value: any) => any;
-    createError: (params?: { path?: string; message?: string, params?: object }) => ValidationError;
+    createError: (params?: { path?: string; message?: string; params?: object }) => ValidationError;
 }
 
 export interface ValidateOptions {
--- a/types/yup/yup-tests.ts
+++ b/types/yup/yup-tests.ts
@@ -636,8 +636,7 @@ const definitionBC: yup.ObjectSchemaDefinition<BC> = {
 };
 const combinedSchema = yup.object(definitionAB).shape(definitionBC); // $ExpectType ObjectSchema<Shape<AB, BC>>
 
-// $ExpectError
-yup.object<MyInterface>({
+const schemaWithoutNumberField = {
     stringField: yup.string().required(),
     subFields: yup
         .object({
@@ -645,10 +644,13 @@ yup.object<MyInterface>({
         })
         .required(),
     arrayField: yup.array(yup.string()).required(),
-});
+};
 
 // $ExpectError
-yup.object<MyInterface>({ stringField: yup.number().required(),
+yup.object<MyInterface>(schemaWithoutNumberField);
+
+const schemaWithWrongType = {
+    stringField: yup.number().required(),
     numberField: yup.number().required(),
     subFields: yup
         .object({
@@ -656,17 +658,22 @@ yup.object<MyInterface>({ stringField: yup.number().required(),
         })
         .required(),
     arrayField: yup.array(yup.string()).required(),
-});
+};
 
 // $ExpectError
-yup.object<MyInterface>({
+yup.object<MyInterface>(schemaWithWrongType);
+
+const schemaWithoutSubtype = {
     stringField: yup.string().required(),
     numberField: yup.number().required(),
     arrayField: yup.array(yup.string()).required(),
-});
+};
 
 // $ExpectError
-yup.object<MyInterface>({ subFields: yup
+yup.object<MyInterface>(schemaWithoutSubtype);
+
+const schemaWithWrongSubtype = {
+    subFields: yup
         .object({
             testField: yup.number().required(),
         })
@@ -674,7 +681,10 @@ yup.object<MyInterface>({ subFields: yup
     stringField: yup.string().required(),
     numberField: yup.number().required(),
     arrayField: yup.array(yup.string()).required(),
-});
+};
+
+// $ExpectError
+yup.object<MyInterface>(schemaWithWrongSubtype);
 
 enum Gender {
     Male = 'Male',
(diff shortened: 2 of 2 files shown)


YashdalfTheGray (changes requested):
@ekfn can you add to the tests file too to make sure we're testing this case properly?

ekfn:
Added tests
@YashdalfTheGray

YashdalfTheGray (approved):
This looks good to me. Thanks for contributing this!

uniqueiniquity:

mvestergaard:
I just noticed this change. I believe this is wrong tbh.
This makes it impossible to reflect the initial state of a form without hacks.

For example let's say I have a field for selecting the assignee of the task, the field is defined `yup.number().nullable().required()`. Until the user selects something, the field is invalid. Before the type of this field would be `number | null`, and i could initialize it to null, which is correct. Now the type is `number`... so I guess I could initialize it to 0, but that's not really correct.
And even worse, let's say the type was `yup.object({...}).nullable().required()`, then I would potentially need to initialize that field with some hacky object structure.

This should really be reverted imo.

I'm pinning to the previous version for now. This breaks a ton of code for us.

ekfn:
@mvestergaard it's a type that returns *after* validation

If it was before validation why can't you initialize with `undefined`?

mvestergaard:
@ekfn Just to establish context. I'm talking about usage in conjunction with Formik.
We would typically have a FormikContext which uses the inferred type of the schema as the form values.
Formik expects an `initialValues` given to it which cannot be undefined. Also if you need to initialize the form state with anything other than a completely empty state (for editing) then undefined wouldn't be correct.

Undefined also leads to different problems. Having undefined as the "null" value rather than null, leads to components going in and out of controlled state.

You're saying this only applies to "after validation" but this has changed the behavior of the `InferType` helper.

If this change only affected the outcome of a validation, then I wouldn't have an issue.

mvestergaard:
I just read the issue you filed on yup for this https://github.com/jquense/yup/issues/761 and honestly I believe this is all just based on confusion.

Copied from there:
```ts
const validationSchema = Yup.object().shape({
  title: Yup.string()
    .nullable()
    .required('Title is required')
});

type ValidationSchema = InferType<typeof validationSchema>;
```

If `ValidationSchema` is supposed to reflect the type of the schema, then `string | null` is correct. An object described by the schema, can certainly have a null value. The schema will say that's it is invalid, but it's a perfectly valid value from a typing perspective.

ekfn:
@mvestergaard according to this https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1911a1fbbe30af03f0f38a915c4bf0620d251fc6/types/yup/index.d.ts#L57
it's a type that returns after validation, `value` can be any (which is not nice).

ekfn:
I think your point is valid and it would be nice to get both types, but current types do not reflect the real code, which is why i created this PR.

mvestergaard:
That's fine and all, but your change altered the outcome of `InferType` to something invalid:

```ts
const schemaWithNullable = Yup.object({
  title: Yup.string().nullable().required(),
});

const schemaWithoutNullable = Yup.object({
  title: Yup.string().required(),
});

type ValuesWithNullable = InferType<typeof schemaWithNullable>; // title = string | null
type ValuesWithoutNullable = InferType<typeof schemaWithoutNullable>; // title = string;

const valuesWithNullable: ValuesWithNullable = {
  title: null
}

const valuesWithoutNullable: ValuesWithoutNullable = {
  //  title: null ** type error, cannot be null 
  title: '' // only way to make this not error
}
```

Now imagine the same thing with a schema like this (or even worse in real life scenarios):
```ts
const nestedSchema = Yup.object({ foo: Yup.string().required() });

const schema = Yup.object({
  nested: Yup.object({ 
    nested: nested: Yup.object({ 
      foo: Yup.string().nullable().required(),
    }).nullable().required();
  }).nullable().required();
});

// ways to define a value for this schema

// before
const value: InferType<typeof schema> = { nested: null }

// after
const value: InferType<typeof schema> = {
  nested: {
    nested: {
      foo: ''
    }
  }
}
```

ekfn:
Your point is to use inner type of yup for `initialValues` with Formik, my point is to sync types of `.validate` method with code.

How do you cast data from Formik after validation? It's same problem, just mirrored.

mvestergaard:
I think you're missing the point. Even outside of formik, this is causing problems. Since `InferType` no longer reflects the type of object described by the schema. `InferType` needs to create a type that describes the object in both valid and invalid state.

@YashdalfTheGray @uniqueiniquity Can you guys please weigh in here?

mvestergaard:
@ekfn going back to this:
```ts
const validationSchema = Yup.object().shape({
  title: Yup.string()
    .nullable()
    .required('Title is required')
});

type ValidationSchema = InferType<typeof validationSchema>;
```

Why not just leave out `.nullable()`? Then you would have the type you expected. 
Why do you need `.nullable()` in this case, when as you mentioned `.validate()` accepts `any`.

mauricedb:
@ekfn @mvestergaard 

Sorry for being late to the party but I am the original implementer of `InferType<T>` and have a few thought on this.

TLDR: This PR is wrong and should not have been accepted/merged. My appologies for not stepping in before that happend.

First of all TypeScript type checking and Yup validation re two very different concerns that occur at very different times and with different constraints. Here `InferType<T>` is used to infer a TypeScript type to satisfy type checking. This does not do validation and ca't do so because the actual value is not known at compile time. The Yup schema is used at runtime to validate but isn't type aware. After all at runtime there is no TypeScript and everything is JavaScript.

Consider the following schema:
```js
const validationSchema = Yup.object({
  age: Yup.number()
    .nullable()
    .min(1)
    .max(100)
    .required()
});
```

Using `InferType<T>` we should get a type like:
```js
type ValidationSchema = {
  age: number | null
}
```

Now null is a valid type but not a valid value. Any number is a valid type but by no means a valid value. For example:
```js
console.log("null", validationSchema.isValidSync({age: null}))
console.log("0", validationSchema.isValidSync({age: 0}))
console.log("50", validationSchema.isValidSync({age: 50}))
console.log("250", validationSchema.isValidSync({age: 250}))
```
All these objects are valid acording to the TypeScript type but only `{age: 50}` is valid acording to the validation rules.

But the difference is bigger. Take for example:
```js
console.log("'50'", validationSchema.isValidSync({age: "50"}))
```
The type would be invalid according to TypeScript as `age` is `string` and not a `number`. Yet it is valid according to Yup because it will try, and succeed here, to convert the string to a number. So any string that can be parsed to a number can work. No way you can express that in a TypeScript type.

And when I start including arrays the difference between validation and TypeScript type is even bigger.

Take the following schema and code:
```js
const validationSchema = Yup.array(Yup.number()).required()

// Is NOT valid
console.log("[]", validationSchema.isValidSync([]))
// Is valid
console.log("[1, 2]", validationSchema.isValidSync([1, 2]))
```

With an array using `.required()` means that an empty array is actually invalid. So the first validation will fail. No way you can express that in a TypeScript type.

Back the the original example:
```js
const validationSchema = Yup.object().shape({
  title: Yup.string()
    .nullable()
    .required('Title is required')
});
```

The `.nullable()` has no effect on validation because a null value will fail the required check. If that is what you are interested in just leave the `.nullable()` out. When infering TypeScript types it can make a lot of sense and should be allowed because it is perfectly acceptable to have a type that is valid with a value that doesn't pass the constraints.

mvestergaard:
@mauricedb Glad to hear I haven't gone mad. There's an open PR to revert this https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41890

mauricedb:
@mvestergaard I am going to give the other contributors and @ekfn a chance to respond first but my vote is to approve that PR and revert this change ASAP.

ekfn:
@mauricedb 
> just leave the .nullable() out
I would love to do it right away, but yup throwing own exceptions in this case, and I'm not sure I can overwrite it.

what about:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41789#issuecomment-578841240
I got the impression that this is the type of validated value, not the type of accepted value.

ekfn:
And I'm fine with reverting this PR, I wanted to clarify the work of `InferType` more than just merging changes.

mvestergaard:
@ekfn In your original example, you don't really need or probably want title to be nullable at all.
The value of an empty input isn't null, it's an empty string, but what you're probably seeing is that you haven't created an initial value to be validated, as i alluded to earlier, and when the field isn't changed, the value of title is undefined, not null, and fails validation.
Illustrated here:
```ts
const validationSchema = Yup.object().shape({
  title: Yup.string()
    .nullable()
    .required('Title is required')
});

const values: any = {};
// some sort of code that renders an input and binds the value to values['title'];
// but the input isn't changed

validationSchema.validate(values); // title at this point is any.
```
What you should have done, is remove `.nullable()` from `title` and then initialize values properly:
```ts
const values: Yup.InferType<typeof validationSchema> = { 
// use InferType to help with creating this value correctly, that's what typescript is for
  title: '';
}

// same as above
```

ekfn:
> The value of an empty input isn't null, it's an empty string

In my case (custom input component) empty value is `null`

mvestergaard:
Then in the final value you need to compensate for the value being able to be null, although you know it never will be, this is a limitation of what can be accomplished with typing this sort of thing, as @mauricedb explained earlier.
```ts
const title = values.title ?? ''
```

ekfn:
what about `number` type? I dont think that `n ?? 0` is a good practice.

mvestergaard:
Since `n` can never be null or undefined in the validated value, it doesn't really matter. It's only to get around the limitations of typings.

ekfn:
> Since n can never be null or undefined in the validated value

An empty value is `null` (when the user did not select anything),` 0` is a valid value (when the user has selected something)

mvestergaard:
You specified in your schema that it is required, null and undefined will never pass that validation, so the validated value cannot be null, hence `n ?? 0` will never use the fallback to 0.

mauricedb:
@ekfn 

The case of `.validate()` is a bit weird but that is the way Yup works. It will accept anyting as input and will resolve that value if it is valid or reject it if it isn't. So the value can't be typed as `T` but the result is `Promise<T>`.

Not the only weird bit in the API. For example `.nullable(true/false)` versus `.required()/.notRequired()`. And `.required()` sort of maps to not allowing `undefined` while in Yup it actually has a more complex meaning like an array that must have at least one elmement in it.

It works well enough but the Yup APi was not designed with static typing like TypeScript does in mind.

mauricedb:
Just to clarify how I go about these things.

If I need to know the difference between user entered data and defaults I will make properties nullable. So something like:
```js
  firstName: yup
    .string()
    .required()
    .nullable()
    .default(null),
  age: yup
    .number()
    .nullable()
    .default(null)
});

const newPerson = personSchema.default()
```

Validating `newPerson` fails because both properties default to `null`
```js
// False
console.log(personSchema.isValidSync(newPerson))
```

Trying to print the uppercase of the first name is actually a runtime error because `firstName` is `null`.
```js
// Runtime error
console.log(newPerson.firstName.toUpperCase())
```

When we infer the TypeScript type the last line should give us a compile time error as well.
```
type Person = yup.InferType<typeof personSchema>;

// Same as 
type Person = {
  firstName: string | null;
  age: number | null
}

// Compile error and would fail at runtime
console.log(newPerson.firstName.toUpperCase())

// Compiles and runs fine with null check
console.log(newPerson.firstName?.toUpperCase())
```

So having types that are more premisive than validations is perfectly fine and something I do all the time. They are both valid and perfectly usefull.

HTH

YashdalfTheGray:
Sorry for the oversight on this merge y'all! I'm happy to go along with @mauricedb here and proceed with the revert.

mauricedb:
No problem, I should have taken a look sooner. I just aproved #41890.

mauricedb:
@ekfn Why do you have to cast to `any`? If an object has a `title` that is `null | string` then having a string in there perfectly valid typing wise. With the JSON you are sending to the server there is no type information, for that matter at runtime there is no TypeScript type info at all. So a `validate()` makes sense to prevent sending invalida data but should have no effect type wise.

ekfn:
> Why do you have to cast to any

because I have a complex nested object with dynamic parts

ekfn:
> Why do you have to cast to any?

because I have a complex nested object with dynamic parts

> With the JSON you are sending to the server there is no type information, for that matter at runtime there is no TypeScript type info at all. So a validate() makes sense to prevent sending invalida data but should have no effect type wise.

We have strict contract with server (we use graphql and all requests are typed) and we have linter, so project won't build if you trying to send `string | null` to a `string` parameter.

mvestergaard:
@ekfn The project I'm working on uses graphql too and has the same sort of constraints. It really isn't a problem.
We have just defined some mapping functions that map from the graphql types to the form types, and vice versa.

ekfn:
It works fine on flat data and not so fine on complex data.

What I'm trying to achieve is something like this:
```ts
type InitialValues = {
    requiredValue: string | null;
    notRequiredValue: string | null;
};

type ValidatedValues = {
    requiredValue: string;
};

type ObjectExtract<A, B> = {
    [P in Extract<keyof A, keyof B>]: Extract<A[P], B[P]>;
} & {
    [P in Exclude<keyof A, keyof B>]: A[P]
};

type ResultType = ObjectExtract<InitialValues, ValidatedValues>;
```

I don't even want to validate `notRequiredValue` because it's always valid, so it's kind of extract of initial type and validated type.

AmitNeuhaus:
Hey guys three years after i still cant understand how can I infer-type from schema to use as default values type null and set it required so null will not be valid value but still won't override the null type from the infered typed.

PR #41792, opened 2020-01-22
feat(async): `retry` method definition update. Closes #41767

### LINKED ISSUE #41767: @types/async callback should be optional in retry() interface
Greetings Authors,
The interface for `async.retry()` has the callback marked as a mandatory parameter instead of optional.

It seems the retry interface should be formatted as this:
```
export function retry<T, E = Error>(
    opts: number | {
        times: number,
        interval: number | ((retryCount: number) => number),
        errorFilter?: (error: Error) => boolean
    },
    task: (callback: AsyncResultCallback<T, E>, results: any) => void,
    callback?: AsyncResultCallback<any, E>
    ): void;
```

- [x] I tried using the `@types/async` package and had problems.
- Authors: @notarseniy  @notarseniy @Penryn @fenying @pascalmartin @Dmitri1337 @erossignon @Juliiii

peterblazejewicz:
- support for optional parameters
- support for differnet return depending on invocation params

See #41767 
Thanks!

Please fill in this template.

https://caolan.github.io/async/v3/docs.html#retry

### DIFF
--- a/types/async/index.d.ts
+++ b/types/async/index.d.ts
@@ -222,14 +222,26 @@ export function auto<R extends Dictionary<any>, E = Error>(tasks: AsyncAutoTasks
 export function autoInject<E = Error>(tasks: any, callback?: AsyncResultCallback<any, E>): void;
 
 export function retry<T, E = Error>(
-    opts: number | {
-        times: number,
-        interval: number | ((retryCount: number) => number),
-        errorFilter?: (error: Error) => boolean
-    },
-    task: (callback: AsyncResultCallback<T, E>, results: any) => void,
-    callback: AsyncResultCallback<any, E>
-    ): void;
+    opts?:
+        | number
+        | {
+              times?: number;
+              interval?: number | ((retryCount: number) => number);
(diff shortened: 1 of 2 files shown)


types/async/index.d.ts
```diff
@@ -218,14 +218,25 @@ export function auto<R extends Dictionary<any>, E = Error>(tasks: AsyncAutoTasks
...
-    callback: AsyncResultCallback<any, E>
-    ): void;
+    opts?:
+        | number
+        | {
+              times?: number;
+              interval?: number | ((retryCount: number) => number);
+              errorFilter?: (error: Error) => boolean;
+          },
+): Promise<void>;
```
weswigham: I'm _pretty sure_ this overload needs the `task` parameter, since it's the only "required" parameter according to the docs, no?

weswigham (changes requested):

types/async/index.d.ts
peterblazejewicz: Done, thanks! It was not covered by existing or added tests, never though about this. Amended

PR #41804, opened 2020-01-23
[react-native] update Lists props to match flow Maybe Types.

chubillkelvin:
Fixes props of Lists to match the usage of [Maybe Types](https://flow.org/en/docs/types/maybe/) in flow. Refer to flow typing in [Lists](https://github.com/facebook/react-native/tree/master/Libraries/Lists).

### DIFF
--- a/types/react-native/index.d.ts
+++ b/types/react-native/index.d.ts
@@ -52,7 +52,6 @@
 /// <reference path="Devtools.d.ts" />
 /// <reference path="LaunchScreen.d.ts" />
 
-import * as PropTypes from 'prop-types';
 import * as React from 'react';
 
 type Constructor<T> = new (...args: any[]) => T;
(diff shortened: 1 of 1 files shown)


TheSavior:
Are you sure these props shouldn't also include undefined? We typically include undefined to support

```
function myComponent(props) {
 return <ScrollView {...props} />
}
```

Which at least in flow requires the pros supporting undefined.

RageBill:
> Are you sure these props shouldn't also include undefined? We typically include undefined to support
> 
> ```
> function myComponent(props) {
>  return <ScrollView {...props} />
> }
> ```
> 
> Which at least in flow requires the pros supporting undefined.

`undefined` is already covered in those cases since they are optional props. If you try adding `undefined` to these props, you will see the message
```
Property is optional, so no need to specify `undefined` as a possible value.
```
when you run `npm run lint react-native`.

TheSavior:
Ah, I forgot the situation. How does TypeScript handle this?

```
<MyOtherComponent />

function MyOtherComponent(props) {
  return (
    <MyComponent 
       horizontal={props.horizontal} // This is actually the one that ruins it. If you don't pass in foo, it becomes undefined here. 
    />
  )
}

function MyComponent(props) {
  // props now is {foo: undefined}, so it isn't missing from the object, it's explicitly marked as undefined
 return <ScrollView {...props} />
}
```

RageBill:
For Typescript, when we define a type, let's say `foo?: number`, it is automatically interpreted as `foo?: number | undefined`.

The behaviour for TS should be the same as in flow in your example.

TheSavior:
Thanks for walking me through this! I can't seem to repro the error our team remembers hitting in Flow that caused us to always put the `?` on the right hand side as well. Maybe that behavior was changed in Flow and we don't need it anymore!

Regardless, good to know that TypeScript doesn't need it. :)

RageBill:
You're welcome :)

PR #41814, opened 2020-01-23
[uuid] Avoid depending on "node" typings

### LINKED ISSUE #328: @types/uuid references @types/node
Unfortunately I'm trying to use this library with Typescript, so I need to use `@types/uuid`, but that contains a reference to `Buffer` in `globals.d.ts` which means it pulls in `@types/node`, which breaks my build. I'm writing code for browsers, not node.

There would be a few possible fixes for this, including just removing the reference to `Buffer` and just using `Array` / `ArrayLike` - or creating an interface for `Buffer` instead of pulling in the one from node...

LinusU:
Fixes https://github.com/uuidjs/uuid/issues/328

-----

There is no code in uuid that specifically depends on `Buffer`, and `Buffer` conforms to `ArrayLike<number>` so we don't need to specifically declare it as accepted ☺️ 

-----

### DIFF
--- a/types/uuid/index.d.ts
+++ b/types/uuid/index.d.ts
@@ -4,6 +4,7 @@
 //                 Felipe Ochoa <https://github.com/felipeochoa>
 //                 Chris Barth <https://github.com/cjbarth>
 //                 Rauno Viskus <https://github.com/rauno56>
+//                 Linus Unnebäck <https://github.com/LinusU>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.2
 
--- a/types/uuid/interfaces.d.ts
+++ b/types/uuid/interfaces.d.ts
@@ -1,8 +1,6 @@
-/// <reference types="node" />
-
 // Uses ArrayLike to admit Unit8 and co.
-export type OutputBuffer = ArrayLike<number> | Buffer;
-export type InputBuffer = ArrayLike<number> | Buffer;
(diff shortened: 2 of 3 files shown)


MattyBalaam:
@LinusU This has completely broken all Typescript in our app somehow!

With a bit of fliddling I can see that re-adding ` /// <reference types="node" />` to the beginning on `interfaces.d.ts` makes this work again. Do you have any ideas?

Screenshot 2020-01-31 at 14 48 49

LinusU:
@MattyBalaam interesting 🙈

Is this a project targeting Node.js? In that case you should add `@types/node` directly to your project.

If not, what is your `"lib"` setting in your tsconfig set to?

MattyBalaam:
This is for a client-side web app.

The libs are `["es2017.intl", "dom"]`

The former is used by `react-intl`. One thing possibly non-standard is that we are using these `typeRoots:  ["./node_modules/@types", "./typings"],`

weswigham:
Oh, it's because your lib setting is missing the base lib - it was being added by node, implicitly. Add es2015 to your lib list.

MattyBalaam:
That's interesting. I assumed that the es2017 setting would have included
anything earlier.

weswigham:
`es2017` includes everything - `es2017.intl` is _just_ the `intl` API and nothing else.

sam-s4s:
Thank you for doing this - my problem is solved <3

MattyBalaam:
Yes that makes sense now so think about it. Thanks.

Issue #41612, opened 2020-01-15
@types/color Type parameter 'T' has a circular default


micobarac:
```
ERROR in node_modules/@types/color/index.d.ts:13:40 - error TS2716: Type parameter 'T' has a circular default.

13 interface Color<T extends ColorParam = ColorParam> {
```

peterblazejewicz:
shouldn't it read?
```ts
interface Color<T  = Color<ColorParam>> {
```

dmihal:
Getting this as well

taktakpeops:
Hello, got into the same issue. Opened PR #41815 to fix it

BeeeQueue:
Make sure your TypeScript versions are up to date. Circular dependencies were introduced in 3.7.

You can wait with updating the package until you've updated TS, as the update doesn't add missing types or change anything else, it's just for developer experience.

taktakpeops:
Coming back to what I said in the PR, in this case we are just mirroring the version of the modules and not really sticking to semver for the typings.

Now, my PR is not only for dev experience but mainly for building my projects in my CI.

Switching from 3.0.0 to 3.0.1 shouldn’t imply upgrading side dependencies based on semver since it’s supposedly a patch and not a minor nor major

BeeeQueue:
All types packages are indirectly dependant on TypeScript - I don't know if updating the required TS version is enough for a new major version (I don't believe it should be).

I don't know who to @ about this either, but it'd be nice to know for sure

jmthibault79:
Pinning the `@types/color` version to 3.0.0 resolved this compile error for me on Typescript 3.1.6.  Thanks for filing this bug.

orta:
Hi thread, we're moving DefinitelyTyped to use [GitHub Discussions](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/53377) for conversations the `@types` modules in DefinitelyTyped.

To help with the transition, we're closing all issues which haven't had activity in the last 6 months, which includes this issue. If you think closing this issue is a mistake, please pop into the [TypeScript Community Discord](https://discord.gg/typescript) and mention the issue in the `definitely-typed` channel.

PR #41815, opened 2020-01-23
fix(@types/color): fix circular ref to T

taktakpeops:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/41612

### DIFF
--- a/types/color/index.d.ts
+++ b/types/color/index.d.ts
@@ -8,7 +8,7 @@
 
 import convert = require('color-convert');
 
-type ColorParam = Color | string | ArrayLike<number> | number | { [key: string]: any };
+type ColorParam = string | ArrayLike<number> | number | { [key: string]: any };
 
 interface Color<T extends ColorParam = ColorParam> {
     toString(): string;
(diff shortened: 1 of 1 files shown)


beeequeue (changes requested):
Are you sure you're on the latest version of TypeScript?

It supports circular type defs since 3.7

taktakpeops:
Hi,

It works with 3.7.

However, for 3.6 and lower it doesn’t.

Since 3.0.1 was a minor, I assume it shouldn’t be a breaking change for the version lower than 3.7. 

That’s why the pull-request.

BeeeQueue:
IMO it's better to keep the types package's major and minor versions aligned with the package it belongs to, but if other maintainers agree we should fix it

The update doesn't add any missing types or anything either way, so you can wait with updating the package until you've updated you TS version.

taktakpeops:
Ok, I understand your point of view.

However, can we rely on semver for the typings? 

It seems to me that we mirror the published version of the modules using the typings more than applying the logic behind semver

jameswlane:
I agree with @BeeeQueue that: 

> it's better to keep the types package's major and minor versions aligned with the package it belongs to

BeeeQueue:
See also my comments in #41612

PR #41825, opened 2020-01-23
fix(carbon-components-react): `title` prop of the AccordionItem. Closes #41811

### LINKED ISSUE #41811: [Accordian][AccordianItem] - AccordionItem does not accepts 'title' as node element in props.
AccordionItem only accepts title as string.
does not accepting title as react node elememt.

If You see in below url- 
title accepts string as default, need to add title as -

`export interface AccordionItemProps extends InheritedProps {
    onHeadingClick?(data: HeadingClickData): void,
    open?: boolean,
    renderExpando?: React.ReactNode,
    title?: React.ReactNode,
}`

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/carbon-components-react/lib/components/Accordion/AccordionItem.d.ts

peterblazejewicz:
- redeclare `title` property to supports React.Node

Thanks!

FYI: omit for:
> type 'ReactNode' is not assignable to type 'string'

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/carbon-design-system/carbon-components-react/blob/master/src/components/AccordionItem/AccordionItem.js#L36

### DIFF
--- a/types/carbon-components-react/lib/components/Accordion/AccordionItem.d.ts
+++ b/types/carbon-components-react/lib/components/Accordion/AccordionItem.d.ts
@@ -8,10 +8,12 @@ export interface HeadingClickData {
     isOpen: boolean;
 }
 
-export interface AccordionItemProps extends InheritedProps {
+export interface AccordionItemProps extends Omit<React.HTMLProps<InheritedProps>, "title"> {
     onHeadingClick?(data: HeadingClickData): void,
     open?: boolean,
     renderExpando?: React.ReactNode,
+    /** The accordion title. */
+    title?: React.ReactNode;
 }
 
 declare class AccordionItem extends React.Component<AccordionItemProps> { }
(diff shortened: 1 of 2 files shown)


types/carbon-components-react/lib/components/Accordion/AccordionItem.d.ts
```diff
@@ -8,10 +8,12 @@ export interface HeadingClickData {
     isOpen: boolean;
 }
 
-export interface AccordionItemProps extends InheritedProps {
+export interface AccordionItemProps extends Omit<React.HTMLProps<InheritedProps>, "title"> {
```
kalbert312:
```suggestion
export interface AccordionItemProps extends Omit<InheritedProps, "title"> {
```
Any HTML props and common carbon props are already on the `InheritedProps` interface. There's no constraint on that type parameter of `HTMLProps`  but it is used as the type for the event handlers and ref so we're no longer extending from `InheritedProps`.

kalbert312 (changes requested):

kalbert312 (approved):

types/carbon-components-react/lib/components/Accordion/AccordionItem.d.ts
peterblazejewicz: Done! Appreciated, thanks!

PR #41830, opened 2020-01-24
[@types/plupload] Enable import

s-hakase:
fix `'@types/plupload/index.d.ts is not a module` when import as a module like `import plupload from 'plupload'`

---

Please fill in this template.

### DIFF
--- a/types/plupload/index.d.ts
+++ b/types/plupload/index.d.ts
@@ -770,5 +770,5 @@ declare namespace plupload {
 }
 
 declare module 'plupload' {
-    export default plupload
+    export = plupload;
 }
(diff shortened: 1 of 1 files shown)


types/plupload/index.d.ts
```diff
@@ -768,3 +768,7 @@ declare namespace plupload {
      */
     function addFileFilter(name: string, cb: Function): void;
 }
+
+declare module 'plupload' {
+    export default plupload
```
weswigham:
```suggestion
    export = plupload;
```
See [here](https://github.com/moxiecode/plupload/blob/master/src/plupload.js#L2466), it's cjs entrypoint is still a module assignment. Locally, you should use `esModuleInterop` to still allow importing it via the `default` in a module.

weswigham (changes requested):
Different export assignment, please~

types/plupload/index.d.ts
s-hakase: Thank you so mach for telling me!

weswigham (approved):

PR #41833, opened 2020-01-24
Mongoose: Add database name, host and port to ConnectionBase

teolag:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://mongoosejs.com/docs/api.html

### DIFF
--- a/types/mongoose/index.d.ts
+++ b/types/mongoose/index.d.ts
@@ -311,6 +311,15 @@ declare module "mongoose" {
      * Each state change emits its associated event name.
      */
     readyState: number;
+      
+    /** Connected database name */
+    name: string
+
+    /** Connected host */
+    host: string
+
+    /** Connected port number */
+    port: number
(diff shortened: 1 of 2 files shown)


ToucheSir (changes requested):
Could you add some tests that check for the correct types on each of these fields? There should already be examples for other fields on `Connection`. Thanks!

teolag:
Thanks for the response @ToucheSir 
Is this the tests you was thinking about? I had a hard time find any tests here, just a lot of statements with .toLowerCase() and .toFixed(). In my mind, test has expects or asserts or something =)

ToucheSir (approved):
That works great. The tests here are definitely different from the usual expect/assert workflow because it is the type definition itself that's being tested. Calling `toLowerCase` and `toFixed` would fail to type check on a differently-typed property, so they work well as asserts here :)

PR #41842, opened 2020-01-24
Fix the return type on getFormError Selector

tremarkley:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/redux-form/lib/selectors.d.ts
+++ b/types/redux-form/lib/selectors.d.ts
@@ -1,7 +1,8 @@
 import { FormErrors, GetFormState } from "../index";
 
 export type DataSelector<FormData = {}, State = {}> = (formName: string, getFormState?: GetFormState) => (state: State) => FormData;
-export type ErrorSelector<State = {}> = (formName: string, getFormState?: GetFormState) => (state: State) => string;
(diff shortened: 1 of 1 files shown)


tremarkley:
documentation on the `getFormError` selector can be found here: https://redux-form.com/8.2.2/docs/api/selectors.md/

types/redux-form/lib/selectors.d.ts
```diff
@@ -2,6 +2,7 @@ import { FormErrors, GetFormState } from "../index";
 
 export type DataSelector<FormData = {}, State = {}> = (formName: string, getFormState?: GetFormState) => (state: State) => FormData;
 export type ErrorSelector<FormData = {}, State = {}, ErrorType = string> = (formName: string, getFormState?: GetFormState) => (state: State) => FormErrors<FormData, ErrorType>;
+export type GetFormErrorSelector<State = {}, ErrorType = string> = (formName: string, getFormState?: GetFormState) => (state: State) => ErrorType;
```
weswigham: The one place this is used, the `getFormError` export, passes no type parameters, so the type is always essentially
```ts
export type GetFormErrorSelector = (formName: string, getFormState?: GetFormState) => (state: {}) => string;
```
since the generics seem extraneous, maybe it's best to just write it like that?

weswigham (changes requested):

PR #41843, opened 2020-01-24
feat: add typings for react-messenger-checkbox

remiroyc:
Please fill in this template.

If adding a new definition:


types/react-messenger-checkbox/index.d.ts
```diff
@@ -0,0 +1,36 @@
+// Type definitions for react-messenger-checkbox 0.1.0
```
peterblazejewicz: I think this will have to be a `major.minor`: `0.1`

types/react-messenger-checkbox/tslint.json
```diff
@@ -0,0 +1,80 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
```
uniqueiniquity: Please do not disable any lint rules in a new package.

uniqueiniquity (changes requested):

types/react-messenger-checkbox/index.d.ts
```diff
@@ -0,0 +1,31 @@
...
+    version?: string;
+    language?: string;
+    debug?: boolean;
+    onEvent?: (event: any) => void;
+}
+
+/**
+ * <MessengerCheckbox />
+ */
+export default class MessengerCheckbox extends React.Component<ReactMessengerProps> {}
```
uniqueiniquity: It seems like the [underlying library](https://unpkg.com/react-messenger-checkbox@0.1.0/lib/index.js) is not using `export default`, so this .d.ts file should not either. Instead, you should follow the class template from [the TypeScript handbook](http://www.typescriptlang.org/docs/handbook/declaration-files/templates.html). Note that if you want to use `import name from 'package';` in your consuming code, you can use the `esModuleInterop` compiler setting.

uniqueiniquity (changes requested):

types/react-messenger-checkbox/index.d.ts
remiroyc: I'm so sorry, this is my first typescript contribution and I don't understand well your request.

Here is the official exemple : https://github.com/Yoctol/react-messenger-checkbox/blob/master/example/src/App.js

I guess package use defaut import. So if I understand your message, I should set esModuleInterop option to true in my .tsconfig ?

Could you expand on that a little bit for me, please?

uniqueiniquity: Sure thing - no apology necessary!
As you probably know, when using React, it is common to write imports as:
```
import MessengerCheckbox from 'react-messenger-checkbox';
```
What this means in NodeJS, without the help of any extra tools, is the following:
```
const MessengerCheckbox = require('react-messenger-checkbox').default;
```
Because this is common in the React community, many component packages are written in the following way:
```
module.exports.default = MyReactComponent;
```
However, this particular library isn't doing that, as can be seen in my link above; instead, it's assigning the component directly to `module.exports`, leaving the `default` property undefined.

Tools like Babel are able to hide this detail by including helper functions (for example, `_interopRequireDefault` in that example) that make imports work correctly. However, TypeScript will only include such a helper function if the `esModuleInterop` option is set to true. Because of this, we want to make sure the definitions on DefinitelyTyped match the real implementation of the library as closely as possible. 

In summary, my recommendation is to write this type declaration file as if there is no export default (because, in reality, there isn't), but when you use it, feel free to use the `esModuleInterop` setting to match the example.

Please let me know if I can clarify further!

remiroyc: Thanks for your reply and your explanation. It was very useful to me.
I just push a new commit to update the type definition and delete the default export.  I don't know if I should declare namespace and export ReactMessengerProps interface inside, because the documentation explains : 

```
 *~ Note that if you decide to include this namespace, the module can be
 *~ incorrectly imported as a namespace object, unless
 *~ --esModuleInterop is turned on:
 *~   import * as x from '[~THE MODULE~]'; // WRONG! DO NOT DO THIS!
```

types/react-messenger-checkbox/index.d.ts
```diff
@@ -0,0 +1,33 @@
...
+    debug?: boolean;
+    onEvent?: (event: any) => void;
+}
+
+/**
+ * <MessengerCheckbox />
+ */
+declare class MessengerCheckbox extends React.Component<ReactMessengerProps> {}
+
+export {};
```
uniqueiniquity: You should do `export = MessengerCheckbox` instead.

types/react-messenger-checkbox/index.d.ts
uniqueiniquity: Given the most recent change, does it work now (that is, are you able to import it as you want with esModuleInterop turned on)?

types/react-messenger-checkbox/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+    "compilerOptions": {
+        "esModuleInterop": true,
+        "module": "commonjs",
+        "lib": ["es6"],
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "strictNullChecks": false,
```
uniqueiniquity: This needs to be `true`.

uniqueiniquity (changes requested):

types/react-messenger-checkbox/index.d.ts
remiroyc: Yes, it works now.

uniqueiniquity (approved):

uniqueiniquity:

PR #41846, opened 2020-01-24
Add debug boolean to UseTableOptions

priley86:
See table options: 
https://github.com/tannerlinsley/react-table/blob/master/docs/api/useTable.md#table-options


ggascoigne (commented):
I'm sorry but the docs are out of date, the debug flag was removed from react-table just before rc-2, you can find it mentioned in the change log towards the end of that section.
https://github.com/tannerlinsley/react-table/blob/master/CHANGELOG.md

priley86:
ok , good to know. thanks. closing for now then.

PR #41848, opened 2020-01-24
chrome: Adds `chrome.devtools.panels.themeName`.

dgp1130:
Official API definition: https://developer.chrome.com/extensions/devtools_panels#property-themeName

Apparently Chrome extensions don't get the `prefers-color-scheme` media query and instead need to use this value. It can only be `"dark"` or `"default"` (even when users explicitly set their theme to `"light"`).

N/A, I don't think there is an effective way to test this since the property only exists within a Chrome extension. I don't see other tests covering `chrome.devtools.panels.*`. Let me know if there is a means of testing this.

I did test this in my extension directly and it seems to be working just fine. I also manually verified that even when setting DevTools theme explicitly to "Light theme" it still outputs the value `"default"` 🤷‍♂️.

https://bugs.chromium.org/p/chromium/issues/detail?id=608869
https://developer.chrome.com/extensions/devtools_panels#property-themeName

`themeName` was added in Chrome 59, but later versions already exist. This PR is just correcting a missed property.

### DIFF
--- a/types/chrome/index.d.ts
+++ b/types/chrome/index.d.ts
@@ -2091,6 +2091,12 @@ declare namespace chrome.devtools.panels {
      * function() {...};
      */
     export function openResource(url: string, lineNumber: number, callback: () => void): void;
+
+    /**
+     * @since Chrome 59.
+     * The name of the color theme set in user's DevTools settings.
+     */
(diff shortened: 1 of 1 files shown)


fredrikhr (approved):
LGTM :+1:

PR #33030, opened 2019-02-13
[jest]: add type reference to jest-diff

SimenB:
- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [ ] Test the change in your own code. (Compile and run.)
- [ ] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [ ] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: From #32891
- [ ] Increase the version number in the header if appropriate.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`.

Is this enough to add the dependency?

It need to import `@types/jest-diff`

PR #41852, opened 2020-01-25
fix: remove jest interface types as they can be imported from jest itself

SimenB:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/facebook/jest/issues/7807

Since Jest has been rewritten to TypeScript, all of these interface types can be imported directly from Jest's individual modules. The major of Jest 25 is the perfect time to remove these 🙂


types/jest/index.d.ts
```diff
@@ -29,8 +29,6 @@
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
-/// <reference types="jest-diff" />
```
SimenB: I added this in #33030 due to a bug in the publisher meaning it wasn't added as a dependency. However, it was not removed when `@types/jest-diff` was deprecated in #39589.

r3nya (approved):

PR #41859, opened 2020-01-25
feat(node): v13.7

SimonSchick:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V13.md#13.7.0

### DIFF
--- a/types/node/module.d.ts
+++ b/types/node/module.d.ts
@@ -1,6 +1,43 @@
 declare module "module" {
     import { URL } from "url";
-    namespace Module {}
+    namespace Module {
+        /**
+         * Updates all the live bindings for builtin ES Modules to match the properties of the CommonJS exports.
+         * It does not add or remove exported names from the ES Modules.
+         */
+        function syncBuiltinESMExports(): void;
+
+        /**
+         * @experimental
+         */
+        function findSourceMap(path: string, error?: Error): SourceMap;
+        interface SourceMapPayload {
+            file: string;
+            version: number;
+            sources: string[];
+            sourcesContent: string[];
+            names: string[];
+            mappings: string;
+            sourceRoot: string;
+        }
+
+        interface SourceMapping {
+            generatedLine: number;
+            generatedColumn: number;
+            originalSource: string;
+            originalLine: number;
+            originalColumn: number;
+        }
+
+        /**
+         * @experimental
+         */
+        class SourceMap {
+            readonly payload: SourceMapPayload;
+            constructor(payload: SourceMapPayload);
+            findEntry(line: number, column: number): SourceMapping;
+        }
+    }
     interface Module extends NodeModule {}
     class Module {
         static runMain(): void;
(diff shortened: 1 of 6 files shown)


SimonSchick:
Please note that I failed to update the version to `13.6` in my last PR, hence the jump in this one.

SimonSchick:
Build failure is unrelated to changes.

types/node/module.d.ts
```diff
@@ -1,6 +1,43 @@
...
+            originalColumn: number;
+        }
+
+        /**
+         * @experimental
+         */
+        class SourceMap {
+            readonly payload: SourceMapPayload;
+            constructor(payload: SourceMapPayload);
+            findEntry(line: number, column: number): SourceMapping;
```
Flarna:
```suggestion
            findEntry(line: number, column: number): SourceMapping | {};
```
see https://github.com/nodejs/node/blob/bbab963db80528f3f932269774f1d5d6012fbb68/lib/internal/source_map/source_map.js#L195

types/node/module.d.ts
```diff
@@ -1,6 +1,43 @@
...
+        /**
+         * Updates all the live bindings for builtin ES Modules to match the properties of the CommonJS exports.
+         * It does not add or remove exported names from the ES Modules.
+         */
+        function syncBuiltinESMExports(): void;
+
+        /**
+         * @experimental
+         */
+        function findSourceMap(path: string, error?: Error): SourceMap;
```
ExE-Boss: Shouldn’t&nbsp;these be&nbsp;defined as&nbsp;static&nbsp;members of&nbsp;the&nbsp;`Module`&nbsp;class?

types/node/module.d.ts
```diff
@@ -1,6 +1,43 @@
...
+         * @experimental
+         */
+        class SourceMap {
+            readonly payload: SourceMapPayload;
+            constructor(payload: SourceMapPayload);
+            findEntry(line: number, column: number): SourceMapping;
+        }
+    }
     interface Module extends NodeModule {}
     class Module {
```
ExE-Boss: Can’t&nbsp;this&nbsp;just be&nbsp;<code>class&nbsp;Module implements&nbsp;NodeModule</code>?

SimonSchick: No, we would have to duplicate all properties in that case.
Also out of scope of this PR.

types/node/module.d.ts
SimonSchick: They should be both but similarly to `events` the exports are hard to model without duplicating a bunch of things.
It's also not _really_ intended to be used as a static but rather just slapped on the module for exports.
See https://github.com/nodejs/node/blob/master/lib/module.js#L7
For that reason I have decided to not model it as such.

ExE-Boss:
`@types/assert` has&nbsp;now been&nbsp;fixed in&nbsp;<https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41916>.

ExE-Boss (approved):

galkin (approved):

types/node/module.d.ts
SimonSchick: @bcoe @MylesBorins for reference.

PR #41878, opened 2020-01-27
[umzug] fix: support array of migrations as input

jaulz:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/sequelize/umzug/blob/master/src/index.js#L65
https://github.com/sequelize/umzug/blob/master/test/fixtures/index.js#L61

### DIFF
--- a/types/umzug/index.d.ts
+++ b/types/umzug/index.d.ts
@@ -211,6 +211,8 @@ declare namespace umzug {
     }
 
     interface Migration {
+        file: string;
+        
         migration(): Promise<any>;
         up(): Promise<any>;
         down(): Promise<any>;
--- a/types/umzug/index.d.ts
+++ b/types/umzug/index.d.ts
@@ -174,7 +174,7 @@ declare namespace umzug {
         /**
          * Options for defined migration
          */
-        migrations?: MigrationOptions;
+        migrations?: MigrationOptions | Migration[];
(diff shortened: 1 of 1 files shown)


types/umzug/index.d.ts
```diff
@@ -213,6 +213,11 @@ declare namespace umzug {
     interface Migration {
```
jessetrinity: I'm not familiar with umzug, but do you need [options](https://github.com/sequelize/umzug/blob/4e3169dd9d6051af46987a97836ed47aa013ebb2/src/migration.js#L37) as well?

jessetrinity (changes requested):

types/umzug/index.d.ts
jaulz: Actually not and I also just removed the other two props (`file` and `path`) from the interface. Migration is the interface and https://github.com/sequelize/umzug/blob/master/src/migration.js is just one specific implementation of the same. `file`, `path` and `options` are only used in the internal implementation but if you want to pass a custom migration (via my originally suggested changes) then you just need to implement the functions.

jessetrinity: Cool. Thanks!

jessetrinity (approved):

Issue #40381, opened 2019-11-14
[@types/mongodb] $addToSet not allowing to set value to array that can be undefined


helios1138:
Authors: @CaselIT, @alanmarcell, @bitjson, @dante-101, @mcortesi, @EnricoPicci, @AJCStriker, @julien-c, @daprahamian, @denys-bushulyak, @BastienAr, @sindbach, @geraldinelemeur, @jishi, @various89, @angela-1, @lirbank, @hector7, @floric, @erikc5000, @Manc, @jloveridge, @ranguna, @HosseinAgha, @albertossilva, @Rauno56

I have a DB record interface with a field like
```ts
array?: ObjectID[] | null
```
then I'm trying to call an update operation with
```ts
$addToSet: { array: value },
```
and I get a TS error
```
Types of property '$addToSet' are incompatible.
  Type '{ array: ObjectID; }' is not assignable to type 'SetFields<Card>'.
    Type '{ array: ObjectID; }' is not assignable to type 'NotAcceptedFields<Card, any[]>'.
      Types of property 'array' are incompatible.
        Type 'ObjectID' is not assignable to type 'undefined'.
```
if I define the field simply as
```ts
array: ObjectID[]
```
it works as expected

I believe this is a bug because I want to describe the database record in a safe way, where fields might or might not exist and it is a perfectly valid mongo operation to call `$addToSet` with a non-existing field, which will result in creating the array with the provided value the first time it's called.

Manc:
I just hit the same problem after upgrading my own project from `"@types/mongodb": "^3.1.22"` to `^3.3.13`.

The issue must be somewhere here:

```
$addToSet?: SetFields<TSchema>;
```

And…

```
export type SetFields<TSchema> = ({
    readonly [key in KeysOfAType<TSchema, any[]>]?: Unpacked<TSchema[key]> | AddToSetOperators<TSchema[key]>;
} &
    NotAcceptedFields<TSchema, any[]>) & {
    readonly [key: string]: AddToSetOperators<any> | any;
};
```

I don't fully understand this yet, but we somehow have to not only allow Array types (`any[]`), but also optional arrays.

I also discovered that this "only" causes an error if `strictNullChecks` are turned off. I highly recommend that we turn on `strictNullChecks` in `tsconfig.json` to make it compatible with projects in `strict` mode, which currently also causes a few tests to fail.

[...]

PR #41879, opened 2020-01-27
In SetFields, take into consideration optional array types.

nicu-chiciuc:
Please fill in this template.

This PR aims to fix this open issue #40381 .
The problem can be summarized by the following example:
```
type Something = {
  optionalArray?: string[];
  normalArray: string[];
  notArray: string;
};

const setFields: SetFields<Something> = {
  // Type 'string' is not assignable to type 'undefined'.ts(2322)
  optionalArray: 'test'
};

// Will have type `type KeysOfSomething = "normalArray" | undefined`
type KeysOfSomething = KeysOfAType<Something, any[]>;
```

The `optionalArray` type is not included as an allowed property by `KeysOfAType`.

A simple solution, and the one in this PR is to replace:
```
KeysOfAType<TSchema, any[]>
```
with
```
KeysOfAType<TSchema, any[] | undefined>
``` 
so as to allow optional types.

Although I'm not sure if this is the best solution since this might also fail for other specific cases. Another approach would be to augment `KeysOfType` (and `KeysOfOtherType`) from:
```
type KeysOfAType<TSchema, Type> = {
  [key in keyof TSchema]: TSchema[key] extends Type ? key : never;
}[keyof TSchema];
```
to
```
type KeysOfAType<TSchema, Type> = {
  [key in keyof TSchema]:
    TSchema[key] extends Type ? key :
    Type extends TSchema[key] ? key :
    never;
}[keyof TSchema];
```
Or something similar, which would check if there's any intersection at all between the types.

### DIFF
--- a/types/mongodb/index.d.ts
+++ b/types/mongodb/index.d.ts
@@ -1209,9 +1209,9 @@ export type ArrayOperator<Type> = {
 };
 
 export type SetFields<TSchema> = ({
-    readonly [key in KeysOfAType<TSchema, any[]>]?: Unpacked<TSchema[key]> | AddToSetOperators<TSchema[key]>;
+    readonly [key in KeysOfAType<TSchema, any[] | undefined>]?: Unpacked<TSchema[key]> | AddToSetOperators<TSchema[key]>;
 } &
-    NotAcceptedFields<TSchema, any[]>) & {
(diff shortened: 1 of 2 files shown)


CaselIT (approved):
Looks good to me. Thanks for the effort

ps: github seems not to have pick up that this pr fixes 40381 .

PR #41880, opened 2020-01-27
Add `addEventListener()` typings to `MediaRecorder`

AppLover69:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder>

### DIFF
--- a/types/dom-mediacapture-record/index.d.ts
+++ b/types/dom-mediacapture-record/index.d.ts
@@ -51,10 +51,7 @@ declare class MediaRecorder extends EventTarget {
 
     addEventListener(type: 'dataavailable', listener: (this: MediaRecorder, event: BlobEvent) => void, useCapture?: boolean): void;
     addEventListener(type: 'error', listener: (this: MediaRecorder, event: MediaRecorderErrorEvent) => void, useCapture?: boolean): void;
-    addEventListener(type: 'pause', listener: (this: MediaRecorder, event: Event) => void, useCapture?: boolean): void;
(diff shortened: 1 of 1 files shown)


types/dom-mediacapture-record/index.d.ts
```diff
@@ -48,6 +49,13 @@ declare class MediaRecorder extends EventTarget {
 
     constructor(stream: MediaStream, options?: MediaRecorderOptions);
 
+    addEventListener(type: 'dataavailable', listener: (this: MediaRecorder, event: BlobEvent) => void, useCapture?: boolean): void;
```
eliasmeire: I don't think this is needed. MediaRecorder already extends EventTarget which has the correct type for `addEventListener`.

elsmr (changes requested):

types/dom-mediacapture-record/index.d.ts
AppLover69: Right, but it lacks the typings for `'dataavailable'` and `'error'` events

eliasmeire: Ah I see, good improvement! maybe you could mimic the way TypeScript does it: https://github.com/microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L2381-L2400. They use an EventMap to enforce both type and event.

AppLover69: That's what I did locally in my own projects, I'll gladly do so 😄

elsmr (approved):

PR #41884, opened 2020-01-27
[rdf-dataset-indexed] Fix data factory type

thewilkybarkid:
Bit surprised, but the data factory is actually meant to be a dataset (core) factory too. Given there's a test for its usage, seems like it really is intended.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/rdf-ext/rdf-dataset-indexed/blob/7ea6c12626be211080c25af6c6eb853318474fbc/test/dataset.test.js#L143-L152


thewilkybarkid:
Maybe `rdf-js` should have something like:

```typescript
export type ExtendedDataFactory<Q extends BaseQuad = Quad, D extends DatasetCoreFactory<Q> = DatasetCoreFactory<Q>> = DataFactory<Q> & D;
```

/cc @tpluscode

PR #41890, opened 2020-01-27
yup: Revert "required is not nullable (#41789)"

mvestergaard:
This reverts commit 0a32b87d146dc3efe1ef47e82e0b02bf89248b5a.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes:  https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41789

```ts
const validationSchema = Yup.object().shape({
  title: Yup.string()
    .nullable()
    .required('Title is required')
});

type ValidationSchema = InferType<typeof validationSchema>;
```

If `ValidationSchema` is supposed to reflect the type of the schema, then `string | null` is correct. An object described by the schema, can certainly have a null value. The schema will say that it is invalid, but it's a perfectly valid value from a typing perspective.

If you wanted the type of the field to be `string` then define it as `Yup.string().required()`.

See comments on original MR for additional reasons why this causes issues.

### DIFF
--- a/types/yup/index.d.ts
+++ b/types/yup/index.d.ts
@@ -93,7 +93,7 @@ export interface MixedSchema<T = any> extends Schema<T> {
     nullable(isNullable?: true): MixedSchema<T | null>;
     nullable(isNullable: false): MixedSchema<Exclude<T, null>>;
     nullable(isNullable?: boolean): MixedSchema<T>;
-    required(message?: TestOptionsMessage): MixedSchema<Exclude<T, null | undefined>>;
+    required(message?: TestOptionsMessage): MixedSchema<Exclude<T, undefined>>;
     notRequired(): MixedSchema<T | undefined>;
     concat(schema: this): this;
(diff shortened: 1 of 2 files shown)


ekfn:
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41789#issuecomment-578841240

mauricedb (approved):

weswigham:
@YashdalfTheGray as the approver on the PR this reverts, I'd like you to weigh in.

mvestergaard:
@weswigham I guess it drowned a bit on the original PR, but he already weighed in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41789#issuecomment-579283847

YashdalfTheGray:
I'll weigh in here too. But I'm sorry, it was an oversight on my part. I fully support this revert.

weswigham:
Ah, OK, sorry.

PR #41894, opened 2020-01-27
[@pollyjs] Add missing adapter and persister packages

offirgolan:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/pollyjs__adapter-puppeteer/index.d.ts
+++ b/types/pollyjs__adapter-puppeteer/index.d.ts
@@ -6,4 +6,6 @@
 
 import Adapter from '@pollyjs/adapter';
 
-export default class PuppeteerAdapter extends Adapter {}
+declare class PuppeteerAdapter extends Adapter {}
+
+export = PuppeteerAdapter;
--- a/types/pollyjs__adapter-puppeteer/index.d.ts
+++ b/types/pollyjs__adapter-puppeteer/index.d.ts
@@ -1,0 +1,9 @@
+// Type definitions for @pollyjs/adapter-puppeteer 4.0
+// Project: https://github.com/netflix/pollyjs/tree/master/packages/@pollyjs/adapter-puppeteer
+// Definitions by: offirgolan <https://github.com/offirgolan>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.4
+
+import Adapter from '@pollyjs/adapter';
+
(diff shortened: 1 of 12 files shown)


types/pollyjs__adapter-puppeteer/index.d.ts
```diff
@@ -0,0 +1,9 @@
+// Type definitions for @pollyjs/adapter-puppeteer 4.0
+// Project: https://github.com/netflix/pollyjs/tree/master/packages/@pollyjs/adapter-puppeteer
+// Definitions by: offirgolan <https://github.com/offirgolan>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.4
+
+import Adapter from '@pollyjs/adapter';
+
+export default class PuppeteerAdapter extends Adapter {}
```
weswigham:
```suggestion
class PuppeteerAdapter extends Adapter {}
export = PuppeteerAdapter;
```

If you look at their [runtime behavior](https://runkit.com/embed/597ogce4iypk) they transpile a `default` into a `module.exports` assignment, so the types need to reflect the runtime reality, not the original source.

types/pollyjs__persister-local-storage/index.d.ts
```diff
@@ -0,0 +1,9 @@
+// Type definitions for @pollyjs/persister-local-storage 4.0
+// Project: https://github.com/netflix/pollyjs/tree/master/packages/@pollyjs/persister-local-storage
+// Definitions by: offirgolan <https://github.com/offirgolan>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.4
+
+import Persister from '@pollyjs/persister';
+
+export default class LocalStoragePersister extends Persister {}
```
weswigham: Same here; [at runtime](https://runkit.com/embed/sqgcmm893xy7) the module doesn't use a default, but rather assigns to the module as a whole.

weswigham (changes requested):
Use `export =` in the export assignments and then add `esModuleInterop: true` in the `tsconfig`s for the tests.

types/pollyjs__adapter-puppeteer/index.d.ts
offirgolan: Updated. There are other polly packages that need this change. Should I address that in a follow up PR or include it here?

weswigham (approved):

PR #41899, opened 2020-01-27
fix(karma): add missing `restartOnFileChange` config option

peterblazejewicz:
This fixes issues in Angular CLI default karma configs (when typed)
https://git.io/Jvm92

Thanks!

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: http://karma-runner.github.io/3.0/config/configuration-file.html https://git.io/Jvm92

### DIFF
--- a/types/karma/karma-tests.ts
+++ b/types/karma/karma-tests.ts
@@ -99,7 +99,8 @@ module.exports = (config: karma.Config) => {
       },
     },
 
-    singleRun: true
+    singleRun: true,
+    restartOnFileChange: true
   });
 };
 
@@ -119,5 +120,6 @@ console.log(karma.constants.DEFAULT_HOSTNAME);
 console.log(karma.VERSION);
 
 karma.config.parseConfig('karma.conf.js', {
-    singleRun: true
+    singleRun: true,
+    restartOnFileChange: true
 });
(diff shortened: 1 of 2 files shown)


types/karma/karma-tests.ts
```diff
@@ -121,3 +121,29 @@ console.log(karma.VERSION);
 karma.config.parseConfig('karma.conf.js', {
     singleRun: true
 });
+
+// angular-cli sample config: https://git.io/Jvm9i
+module.exports = (config: karma.Config) => {
```
43081j: can we please narrow this test config down to what we're expecting to test? rather than having a copy from another project.
most things should have been tested in the earlier config, so here you should only need a minimum config object with `restartOnFileChange` set

43081j (changes requested):

types/karma/karma-tests.ts
peterblazejewicz: sure, tbd, thanks!

peterblazejewicz: Done, commit amended. Thanks!

devoto13 (approved):

43081j (approved):

PR #41539, opened 2020-01-11
Improve styled-system type safety by using generics for all properties

META-DREAMER:
This allows for type safety for values in a theme using generics (e.g. you wont be allowed to use breakpoints that arent defined in your theme), as well as being able to  use styled-system in Platforms like React Native where the valid values for CSS properties are different.

The generics have been set to the defaults they were before, so backwards compatibility is preserved, but now you have the power to customize the typings to be specific to the values in your theme.

Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [ ] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: N/A, just an improvement on type safety / extensibility for existing functionality
- [ ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

PR #41903, opened 2020-01-28
Improve styled-system type safety by using generics for all properties

META-DREAMER:
Reopening #41539

This allows for type safety for values in a theme using generics (e.g. you wont be allowed to use breakpoints that arent defined in your theme), as well as being able to  use styled-system in Platforms like React Native where the valid values for CSS properties are different.

The generics have been set to the defaults they were before, so backwards compatibility is preserved, but now you have the power to customize the typings to be specific to the values in your theme.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: N/A, just an improvement on type safety / extensibility for existing functionality

### DIFF
--- a/types/styled-system/index.d.ts
+++ b/types/styled-system/index.d.ts
@@ -20,7 +20,7 @@
 //                 Nicholas Hehr <https://github.com/HipsterBrown>
 //                 Hammad Jutt <https://github.com/hammadj>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.9
+// Minimum TypeScript Version: 3.0
 
 import * as CSS from 'csstype';
 
--- a/types/styled-system/index.d.ts
+++ b/types/styled-system/index.d.ts
@@ -26,7 +26,7 @@ import * as CSS from 'csstype';
 
 export function get(obj: any, ...paths: Array<string | number>): any;
 
-export type ObjectOrArray<T, K extends keyof any = keyof any> = T[] | Record<K, T | Record<K, T> | T[]>;
+export type ObjectOrArray<T, K extends keyof any = keyof any> = T[] | Record<K, T | Record<K, T | Record<K, T> | T[]> | T[]>;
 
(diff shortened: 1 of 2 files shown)


hammadj:
@zephraph It seems like updating the minimum TS version to 3.0 causes the CI to fail because of rebass/grid having a lower required version of TS. What's the best way to address this? Update the version number in the header? Problem is that styled-system is still on 5.1 so would get out of sync. Or should I go into the rebass typings and update that to have 3.0 as minimum?

alloy:
Hmm, I think it should be fine for dependencies to have a lower minimum TS requirement.

@orta Any idea on why this shouldn’t be ok?

orta:
if `rebass__grid`is imported into an old typescript project the build might not compile because this uses features which don't exist in that version of typescript - you'd be breaking their API contract - you'd have to also bump their version in this PR basically

alloy:
Ah yes, of course–I read [the message](https://travis-ci.org/DefinitelyTyped/DefinitelyTyped/builds/642739292#L241) wrong and thought they were dependencies, not dependents 👍

zephraph:
Ha, @orta the owner approved message there looks unintentional. 

@hammadj, hopefully @orta's message above helps... sounds like we might need to do some extra type gymnastics. 

Do you _have_ to update the version to 3.0? I thought I saw 2.9 in the build last time, which (hopefully) is a lower lift?

hammadj:
@zephraph I think if I remove the `unknown` type and use `any` instead, I can get it to work with TS 2.9. Not as typesafe, but still better than it was before this PR. Will give it a shot now.

zephraph:
Definitely looking better!

hammadj:
@zephraph Think I got it passing now

zephraph:
Good job! I’ll go through it a bit more today, but it’s looking good.

zephraph:
Hey @alloy, let's ship it!

🚢

alloy:
Aight! 👍

PR #41906, opened 2020-01-28
[rdf-js] Add generics to BaseQuad and formalise Triple

thewilkybarkid:
In draft while I'm experimenting, but this adds generics to `BaseQuad`, and makes `Triple` a real triple (that is, `graph` _has_ to be `DefaultGraph`).

### DIFF
--- a/types/n3/index.d.ts
+++ b/types/n3/index.d.ts
@@ -100,8 +100,6 @@ export class Quad extends BaseQuad implements RDF.Quad {
     toJSON(): string;
 }
 
-export class Triple extends Quad implements RDF.Triple {}
-
 export namespace DataFactory {
     function namedNode(value: string): NamedNode;
     function blankNode(value?: string): BlankNode;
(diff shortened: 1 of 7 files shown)


types/n3/index.d.ts
```diff
@@ -110,8 +108,8 @@ export namespace DataFactory {
     function defaultGraph(): DefaultGraph;
     function quad(subject: RDF.Quad_Subject, predicate: RDF.Quad_Predicate, object: RDF.Quad_Object, graph?: RDF.Quad_Graph): Quad;
     function quad<Q_In extends RDF.BaseQuad = RDF.Quad, Q_Out extends BaseQuad = Quad>(subject: Q_In['subject'], predicate: Q_In['predicate'], object: Q_In['object'], graph?: Q_In['graph']): Q_Out;
-    function triple(subject: RDF.Quad_Subject, predicate: RDF.Quad_Predicate, object: RDF.Quad_Object): Quad;
-    function triple<Q_In extends RDF.BaseQuad = RDF.Quad, Q_Out extends BaseQuad = Quad>(subject: Q_In['subject'], predicate: Q_In['predicate'], object: Q_In['object']): Q_Out;
+    function triple(subject: RDF.Quad_Subject, predicate: RDF.Quad_Predicate, object: RDF.Quad_Object): RDF.Triple<Quad>;
```
thewilkybarkid: I think the `This is the default value for this type parameter, so it can be omitted` error is wrong here? `Quad` is the local N3 quad, not the `rdf-js` `Quad`?

tpluscode:
What issue does this solve? I find the construct `Triple<Quad>` rather weird...

types/rdf-js/index.d.ts
```diff
@@ -174,27 +174,27 @@ export type Quad_Graph = DefaultGraph | NamedNode | BlankNode | Variable;
 /**
  * An RDF quad, taking any Term in its positions, containing the subject, predicate, object and graph terms.
  */
-export interface BaseQuad {
+export interface BaseQuad<S extends Term = Term, P extends Term = Term, O extends Term = Term, G extends Term = Term> {
```
tpluscode: I don't think you have to genarlise S/P/O to get the "real triple". 

How about something like this instead:

```ts
export interface BaseQuad<TGraph extends Term = Term> {
  // ...

  graph: TGraph;
}

export interface Quad extends BaseQuad<Quad_Graph> {
  // ...
}

export interface Triple extends Quad<DefaultGraph> {
}
```

tpluscode (changes requested):

rubensworks:
FYI, the typings here are a bit outdated.
`Triple` has been removed from the spec: https://github.com/rdfjs/data-model-spec/pull/153

So they should (eventually) be removed from these typings here as well.
(For now, I suggest to mark them as deprecated to avoid breaking existing usages)

thewilkybarkid:
Wasn't aware of the removal of `triple`, will close then!

PR #41909, opened 2020-01-28
[@types/wnumb] Update types to wNumb version 1.2.0

eikooc:
This change corrects the types and exports them.

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/leongersen/wnumb/releases

UPDATE:
Can you squash the commits to only leave the first commit and it's corresponding commit description?

### DIFF
--- a/types/wnumb/index.d.ts
+++ b/types/wnumb/index.d.ts
@@ -4,7 +4,7 @@
 //                 Jamie Neubert Pedersen <https://github.com/eikooc>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export = wNumb;
+export default wNumb;
 export as namespace wNumb;
 
 /**
--- a/types/wnumb/index.d.ts
+++ b/types/wnumb/index.d.ts
@@ -1,7 +1,10 @@
-// Type definitions for wnumb 1.0
+// Type definitions for wnumb 1.2.0
 // Project: https://github.com/leongersen/wnumb
-// Definitions by: Corey Jepperson <https://github.com/acoreyj>
+// Definitions by: Corey Jepperson <https://github.com/acoreyj> and Jamie Neubert Pedersen <https://github.com/eikooc>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export = wNumb;
+export as namespace wNumb;
 
 /**
  * Create a wNumb
@@ -28,7 +31,7 @@ declare namespace wNumb {
         /**
          * A number to append to a number. For example: ',-'.
          */
-        postfix?: string;
+        suffix?: string;
         /**
          * The prefix for negative values. Defaults to '-' if negativeBefore isn't set.
          */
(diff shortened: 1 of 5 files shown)


types/wnumb/index.d.ts
```diff
@@ -1,8 +1,11 @@
-// Type definitions for wnumb 1.0
+// Type definitions for wnumb 1.2.0
```
peterblazejewicz: this should read `1.2` I believe:

types/wnumb/index.d.ts
```diff
@@ -1,8 +1,11 @@
...
 // Project: https://github.com/leongersen/wnumb
-// Definitions by: Corey Jepperson <https://github.com/acoreyj>
+// Definitions by: Corey Jepperson <https://github.com/acoreyj> and Jamie Neubert Pedersen <https://github.com/eikooc>
```
peterblazejewicz: Can you follow the readme guide? (append via comma separated entry or wrap into new line):
https://github.com/DefinitelyTyped/DefinitelyTyped#edit-an-existing-package

eikooc: Alright :)

eikooc:
Can you squash the commits to only leave the first commit and it's corresponding commit description?

types/wnumb/index.d.ts
eikooc: @peterblazejewicz everything should be in order now

eikooc:
@peterblazejewicz is it okay now?

types/wnumb/index.d.ts
```diff
@@ -1,8 +1,12 @@
-// Type definitions for wnumb 1.0
+// Type definitions for wnumb 1.2
 // Project: https://github.com/leongersen/wnumb
 // Definitions by: Corey Jepperson <https://github.com/acoreyj>
+//                 Jamie Neubert Pedersen <https://github.com/eikooc>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
+export default wNumb;
+export as namespace wNumb;
```
jessetrinity: I think you want to remove the `declare namespace` below so you have

`export as namespace wNumb`
`export interface Options`
`export interface Instance`

jessetrinity (changes requested):

types/wnumb/index.d.ts
eikooc: Alright. Latest commit should do that. Thanks!

jessetrinity: You do need to export those interface if you want that type info for intelliSense, etc:

```ts
export interface Options {
    decimals?: number;
    ...
}

export interface Instance {
    ...
}
```

You should modify the test to use these types to make sure it actually works:

```ts
const options: wNumb.Options = {decimals: 1}
const moneyFormat: wNumb.Instance = wNumb(options);
```

Otherwise your types will just show up as `any` which is probably not what you want.

jessetrinity (changes requested):

types/wnumb/index.d.ts
eikooc: Okay, I didn't catch that. Tests and types should be updated now :)

jessetrinity:
Thanks!

PR #41910, opened 2020-01-28
Add browser-or-node types

jussikinnula:

### DIFF
--- a/types/browser-or-node/index.d.ts
+++ b/types/browser-or-node/index.d.ts
@@ -1,0 +1,9 @@
+// Type definitions for browser-or-node 1.2
+// Project: https://github.com/flexdinesh/browser-or-node
+// Definitions by: Jussi Kinnula <https://github.com/jussikinnula>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+export const isBrowser: boolean;
+export const isWebWorker: boolean;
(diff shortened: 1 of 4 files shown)


jussikinnula:
Note! There's already a PR (https://github.com/flexdinesh/browser-or-node/pull/6) for adding types into the package itself, but it's been open for a while. Let's add this and when the package gets types, then remove these.

types/browser-or-node/index.d.ts
```diff
@@ -0,0 +1,9 @@
+// Type definitions for browser-or-node 1.2
+// Project: https://github.com/flexdinesh/browser-or-node
+// Definitions by: Jussi Kinnula <https://github.com/jussikinnula>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 3.7
```
weswigham: You can omit this line from the header - these type definitions are super simple and should work all the way back into typescript's past.

weswigham (changes requested):
The TS version header seems a bit gratuitous for these types

types/browser-or-node/index.d.ts
jussikinnula: I changed the TS version to 2.8, as the tests were passing with that. You're right that it should work even with the oldest TypeScript versions.

weswigham (approved):

PR #41914, opened 2020-01-28
Add types for serviceworker-webpack-plugin

remcohaszing:
These type definitions are for the injected code, not for the plugin itself.

If adding a new definition:

### DIFF
--- a/types/serviceworker-webpack-plugin/index.d.ts
+++ b/types/serviceworker-webpack-plugin/index.d.ts
@@ -11,7 +11,11 @@ export interface ServiceWorkerOption {
     jsonStats?: Stats.ToJsonOutput;
 }
 
-export interface ServiceWorkerWebpackPluginOptions<T = Pick<ServiceWorkerOption, 'assets'>> {
+export interface ServiceWorkerDefaultOption {
+    assets: string[];
+}
+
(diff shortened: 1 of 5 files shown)


types/serviceworker-webpack-plugin/index.d.ts
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Whether to minimize output.
+     *
+     * @default process.env.NODE_ENV === 'production'
+     */
+    minimize?: boolean;
+}
+
+export default class ServiceWorkerWebpackPlugin<T = ServiceWorkerDefaultOption> extends Plugin {
```
weswigham:
```suggestion
export = ServiceWorkerWebpackPlugin;
class ServiceWorkerWebpackPlugin<T = ServiceWorkerDefaultOption> extends Plugin {
```
While the original source does use a default export, their build process makes it so that [at runtime the default becomes the commonjs module as a whole.](https://runkit.com/embed/aqnvd2ci25qv)

You'll need to use an export assignment here and the `esModuleInterop` flag in projects using it as a default to reflect this.

weswigham (changes requested):

weswigham (approved):

Issue #40897, opened 2019-12-07
[@types/theme-ui] Error on spreading styled system interpolation objects


hasparus:
- [X] I tried using the `@types/xxxx` package and had problems.
- [X] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript
- [x] [Mention](https://github.com/blog/821-mention-somebody-they-re-notified) the authors (see `Definitions by:` in `index.d.ts`) so they can respond.

@erikdstock @ifiokjr @sbardian @pomber @wKovacs64 @prateekkathal

---

Hi! I've encountered a type error while trying to compose reusable style objects. 

**TLDR**: Type 'SystemCssProperties' is not assignable to type 'UseThemeFunction'.

### Reproduction

[CodeSandbox](https://codesandbox.io/s/theme-ui-sx-type-error-repro-3uvyb)

```tsx
import { SystemCssProperties } from "@styled-system/css";

const buttonStyles: SystemCssProperties = {
  font: "inherit",
  background: "papayawhip",
  border: "none"
};

const styles = { buttonStyles };

const _ = (
  <button
    type="button"
    sx={styles.buttonStyles} // 🔥
  >
    click me
  </button>
);
```

Typing `buttonStyles` as `SxStyleProp` works, but then you can't spread it onto other objects like

```tsx
  sx={{ ...styles.buttonReset, background: "blue" }}
```

### Errors

For the first snippet:

```
error TS2322: Type 'SystemCssProperties' is not assignable to type '(SystemCssProperties & Record<string, string | number | SystemCssProperties | CSSPseudoSelectorProps | ... 4 
more ... | Record<...>>) | ... 4 more ... | undefined'.
  Type 'SystemCssProperties' is not assignable to type 'UseThemeFunction & Record<string, string | number | 
SystemCssProperties | CSSPseudoSelectorProps | CSSSelectorObject | VariantProperty | UseThemeFunction | (string | ... 1 more ... | null)[] | Record<...>>'.       
    Type 'SystemCssProperties' is not assignable to type 'UseThemeFunction'.
      Type 'SystemCssProperties' provides no match for the signature '(theme: any): SystemStyleObject'.     

165   <button type="button" sx={styles.buttonStyles}> 
                            ~~

  node_modules/@types/theme-ui/index.d.ts:157:5       
    157     sx?: SxStyleProp;
            ~~
[...]

Issue #7719, opened 2016-01-21
Definitely Not Versioned


snoopcheri:
What's the recommended way to find a TypeScript definition file for a specific version of a library and a specific TypeScript version? (for example a definition file for lodash v2.4.2, which compiles with TypeScript 1.5 and above)

What I know is the following:
#### Library Version

Some TypeScript definition files contain a comment at the beginning of the file, which mentions the version of the library this definition file refers to (e.g. AngularJS which specifies version 1.4+). Other TypeScript definition files don't mention anything in this respect - not even in the commit history.

Is there no reliable way to know what library version is covered by a TypeScript definition file?
#### TypeScript version

The DefinitelyTyped repository has branches starting with 0.8 up to 1.4.1. I'm assuming that this represents the TypeScript version itself. Unfortunately it ends with 1.4.1. Therefore I assumed that TypeScript definition files on the master branch are valid for TypeScript 1.4.1 and above. Unfortunately this is not always the case - there are TypeScript definition file which are only valid for TypeScript 1.6 and above.

Is there no reliable way to know what TypeScript version is required for a TypeScript definition file?

What we're doing at the moment is storing commit hashes of TypeScript definition files that compile with the TypeScript version we're using currently (v1.5.3). While this is a pain, it kinda works. But since we usually don't know, which library version the definition file is for, we're just hoping for the best. This doesn't sound like a promising and reliable process though.

snoopcheri:
Anyone?

An answer along the lines "there is no recommended way, apart from being up-to-date with TypeScript and libraries is the best you can do" or "it depends on the library" are perfectly acceptable answers. :)

I'm mainly asking because I want to be sure that I'm not missing something very obvious here...

alainsahli:
+1

sir-marc:
No one?

simonalbrecht:
+1 👍

mhegazy:
@types packages have TS version tags on them.  if you want a specific version of the compiler then fix your dependency on `@ts2.0` for instance.

simonalbrecht:
@mhegazy Yes that is clear. 

The issue here is that there is no clear way to see for which version of a library the typings are for. The version numbers of the `@types` package only very rarely match the version of the dependency you're pulling the typings in. For example, `jquery@3.2.1` (which is the latest version at this time) is paired with `@types/jquery@2.0.42`.

[...]

PR #41915, opened 2020-01-28
Reexport @theme-ui/components from theme-ui

hasparus:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://theme-ui.com/migrating/, https://github.com/system-ui/theme-ui/blob/master/packages/theme-ui/src/index.js

**Linked to** https://github.com/system-ui/theme-ui/issues/121.

### DIFF
--- a/types/theme-ui/index.d.ts
+++ b/types/theme-ui/index.d.ts
@@ -8,7 +8,7 @@
 //                 Prateek Kathal <https://github.com/prateekkathal>
 //                 Piotr Monwid-Olechnowicz <https://github.com/hasparus>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 3.1
+// TypeScript Version: 3.5
 
 import { SystemStyleObject } from '@styled-system/css';
 import * as CSS from 'csstype';
--- a/types/theme-ui/index.d.ts
+++ b/types/theme-ui/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for theme-ui 0.2
+// Type definitions for theme-ui 0.3
 // Project: https://github.com/system-ui/theme-ui#readme
 // Definitions by: Erik Stockmeier <https://github.com/erikdstock>
 //                 Ifiok Jr. <https://github.com/ifiokjr>
@@ -6,23 +6,20 @@
 //                 Rodrigo Pombo <https://github.com/pomber>
 //                 Justin Hall <https://github.com/wKovacs64>
 //                 Prateek Kathal <https://github.com/prateekkathal>
+//                 Piotr Monwid-Olechnowicz <https://github.com/hasparus>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.1
 
-import { ResponsiveStyleValue, SystemStyleObject } from '@styled-system/css';
+import { SystemStyleObject } from '@styled-system/css';
 import * as CSS from 'csstype';
 import * as React from 'react';
-import { lineHeight, Theme as StyledSystemTheme } from 'styled-system';
+import { Theme as StyledSystemTheme } from 'styled-system';
 import { Interpolation, SerializedStyles } from '@emotion/serialize';
+export * from '@theme-ui/components';
 
 export {};
 
-type Omit<T, K> = Pick<T, Exclude<keyof T, K>>;
 type ObjectOrArray<T> = T[] | { [K: string]: T | ObjectOrArray<T> };
-
-interface Object<T> {
-    [k: string]: T | Object<T>;
-}
 
 export interface ThemeProviderProps<Theme> {
     theme: Partial<Theme> | ((outerTheme: Theme) => Theme);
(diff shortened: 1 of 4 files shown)


prateekkathal:
@hasparus I don't think it is a good idea to introduce breaking changes. Maybe we should keep the old keys/properties to support older versions. Please add the below as a comment.

```js
DEPRACATED: Not supported after theme-ui:^0.3.*
```

Also, I'd suggest adding tests for the newer versions separately rather than changing the old ones.

hasparus:
> Maybe we should keep the old keys/properties to support older versions. Please add the below as a comment.

I get the point. It leads to another problem though.
Comments appear in tooltips and that's just it. No errors. So we can update theme-ui to 0.3 without seeing the changelog. Maybe dependabot or greenkeeper made the PR.
We merge to PR, continous deployment runs and our default color mode changed, because **`initialColorMode` is not handled by theme-ui anymore**. This is a bug at runtime.

`@types` packages are versioned. I believe we should match breaking changes of theme-ui (best effort) to help users avoid bugs in runtime. This PR won't break old versions of @types/theme-ui.

I'd find deprecation comment fitting, if theme-ui added a deprecation warning at runtime. There is no warning, thus, this is a breaking change, and we can either help the user deal with it by removing `initialColorMode` or don't.

Connected issues: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/7719 https://github.com/DefinitelyTyped/DefinitelyTyped/issues/25677

hasparus:
We needed to upgrade required version to 3.5 because @types/theme-ui__components explodes on TS3.4 (due to "as" prop).

types/theme-ui/index.d.ts
```diff
@@ -1,29 +1,26 @@
...
 // TypeScript Version: 3.1
 
-import { ResponsiveStyleValue, SystemStyleObject } from '@styled-system/css';
+import { SystemStyleObject } from '@styled-system/css';
 import * as CSS from 'csstype';
 import * as React from 'react';
-import { lineHeight, Theme as StyledSystemTheme } from 'styled-system';
+import { Theme as StyledSystemTheme } from 'styled-system';
 import { Interpolation, SerializedStyles } from '@emotion/serialize';
+export * from '@theme-ui/components';
```
prateekkathal: I am not sure why this belongs here. The goal is to only make types exportable, not components by itself.

types/theme-ui/index.d.ts
```diff
@@ -157,14 +154,6 @@ export interface SxProps {
 
 type SxComponent<T extends SxProps = IntrinsicSxElements['div']> = React.ComponentClass<T & { as?: React.ElementType }>;
 
-export const Box: SxComponent;
```
prateekkathal: Again, the goal is to define component props, not make them exportable from the main package itself.

prateekkathal (changes requested):

prateekkathal:
@hasparus It seems like there is a PR sitting for `styled-system` #41903 & issue still unresolved for `theme-ui` #40897 

I'd defer merging this PR as the proposed changes might fix older versions of theme-ui types. If a breaking change must be introduced, it should be made ensuring previous builds don't have errors.

types/theme-ui/index.d.ts
hasparus: Take a look at https://github.com/system-ui/theme-ui/blob/b0257f0aa3dda6e65482552780fda715783205c3/packages/theme-ui/src/index.js#L18

This is the **title of my pull request**.
ThemeUI 0.3 reexports @theme-ui/components. We want to provide typings for these reexported components, so we can have typechecking and autoimports while using theme-ui.

prateekkathal: @hasparus Just to be clear, re-exporting can also be understood in terms of making improvements. Also, theme-ui exporting it's own components package directly is completely different than a types package exporting a theme-ui package.

Again, like I said... this package is only meant to define & export types, not the whole library or components. If this is the case, then this PR doesn't make sense at all. The package is meant to export only types not the whole components itself. 

Also, just to be clear, theme-ui doesn't support types locally, so there is no way type checking can happen. We have to manually define them. Removing them altogether defeats the purpose of this package itself.

prateekkathal: @hasparus It seems I completely overlooked the fact that a types/theme-ui components was released by you. 

Hmm... I think in that case, my comments are irrelevant! 

Thanks for making the PR 👍

prateekkathal:
@hasparus I am gonna wait on the updated version of Styled System types to be merged and will then approve this one. 

Thanks again for the PR! 👍

types/theme-ui/index.d.ts
hasparus: Sorry I didn’t express myself clearly. I assumed "types" is implicit in the context of DefinitelyTyped. A bit lazy of me.

prateekkathal (approved):

prateekkathal:
@hasparus The commit from #41903 requires `Omit`. Maybe we should import it from the components package.

PR #41916, opened 2020-01-28
Revert "[assert] Add support for TypeScript's Assertion Functions"

uniqueiniquity:
Reverts DefinitelyTyped/DefinitelyTyped#41616 due to issues caused in consuming packages (e.g. Node).

The core change should be submitted using [types versions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-1.html#version-selection-with-typesversions) to avoid breaking consumers.


LinusU:
@uniqueiniquity hmm, was there a reason to also revert all changes to the packages that before (and now again) depends unnecessary on `assert`? Wouldn't it be good to have them not depend on assert anyway? 🤔 

Also, sorry for not seeing your response in that PR! Just saw it now, must have missed it in my notifications, would have reacted sooner if I had...

PR #41919, opened 2020-01-28
Fix node's assert module resolution in tests

sandersn:
Previously it incorrectly resolved to @types/assert, which is *almost* the same, but a port that necessarily lags a bit, and is compatible with the browser.

### DIFF
--- a/types/node/tsconfig.json
+++ b/types/node/tsconfig.json
@@ -50,6 +50,9 @@
         "typeRoots": [
             "../"
         ],
+        "paths": {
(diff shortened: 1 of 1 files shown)


sandersn:
types-publisher has a check that throws unless path mappings are of the form `"packageName/v3". But it already has a list of the built-in node modules; it might be worth making an exclusion in the code to allow this.

SimonSchick:
Isn't the same needed for the ts3.5 folder?

PR #41920, opened 2020-01-28
[relay-runtime] update store `lookup` arg

hellatan:
allow `selector` arg to be same type returned from `getSelector`

Type returned from here:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/relay-runtime/lib/store/RelayModernSelector.d.ts#L49

### DIFF
--- a/types/relay-runtime/lib/store/RelayStoreTypes.d.ts
+++ b/types/relay-runtime/lib/store/RelayStoreTypes.d.ts
@@ -482,7 +482,7 @@ export interface Environment {
      * Optionally takes an owner, corresponding to the operation that
      * owns this selector (fragment).
      */
-    lookup(selector: SingularReaderSelector): Snapshot;
+    lookup(selector: SingularReaderSelector | ReaderSelector): Snapshot;
 
     /**
      * Send a query to the server with Observer semantics: one or more


alloy (changes requested):
The current state of the types matches upstream. That is, [`getSelector` returns a `ReaderSelector`](https://github.com/facebook/relay/blob/597d2a17aa29d401830407b6814a5f8d148f632d/packages/relay-runtime/store/RelayModernSelector.js#L153), which is [a union of `SingularReaderSelector` and `PluralReaderSelector`](https://github.com/facebook/relay/blob/aa60b530d378910d88ba1a41499e6817ee0b370d/packages/relay-runtime/store/RelayStoreTypes.js#L74). If you think `lookup` should accept either, then please first PR a Flow change upstream so we can stay in sync.

types/relay-runtime/lib/store/RelayStoreTypes.d.ts
```diff
@@ -482,7 +482,7 @@ export interface Environment {
      * Optionally takes an owner, corresponding to the operation that
      * owns this selector (fragment).
      */
-    lookup(selector: SingularReaderSelector): Snapshot;
+    lookup(selector: SingularReaderSelector | ReaderSelector): Snapshot;
```
hellatan: @alloy should this really be `lookup(selector: ReaderSelector)` then if that `ReaderSelector` is a union of both `SingularReaderSelector` and `PluralReaderSelector`?

hellatan:
thanks for the guidance @alloy. Walking through the relay source code (instead of just looking at type def files), gives it a lot more clarity as to what's actually going on under the hood. I'm still a little confused as to how these types work within the relay package itself, but from what I was experiencing, I think I can solve this on my app level instead of needing to update these types defs.

alloy:
Sounds good–happy to be of assistance 👍

PR #41931, opened 2020-01-29
@types/node: http2: Add protocol option to http2.connect

murgatroid99:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: the option is read at https://github.com/nodejs/node/blob/70c32a6d190e2b5d7b9ff9d5b6a459d14e8b7d59/lib/internal/http2/core.js#L3010 and checked at https://github.com/nodejs/node/blob/70c32a6d190e2b5d7b9ff9d5b6a459d14e8b7d59/lib/internal/http2/core.js#L3028-L3036

I am unsure about restricting the option so strongly. On one hand, in the code path that checks the value, any value other than those two results in an error being thrown. On the other hand, there is a code path that does not check the value at all, so technically in that case any value is allowed.

### DIFF
--- a/types/node/http2.d.ts
+++ b/types/node/http2.d.ts
@@ -440,6 +440,7 @@ declare module "http2" {
     export interface ClientSessionOptions extends SessionOptions {
         maxReservedRemoteStreams?: number;
         createConnection?: (authority: url.URL, option: SessionOptions) => stream.Duplex;
+        protocol?: 'http:' | 'https:';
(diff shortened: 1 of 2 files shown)


Flarna (approved):
You could consider to a additionally a PR for node docs.

murgatroid99:
The corresponding Node docs PR is at nodejs/node##31560.

PR #41932, opened 2020-01-29
[@testing-library/dom] v6.11 - add new history popstate event

bochap:
Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/testing-library/dom-testing-library/pull/433>>

### DIFF
--- a/types/testing-library__dom/events.d.ts
+++ b/types/testing-library__dom/events.d.ts
@@ -80,7 +80,8 @@ export type EventType =
     | 'pointerOut'
     | 'pointerLeave'
     | 'gotPointerCapture'
-    | 'lostPointerCapture';
+    | 'lostPointerCapture'
+    | 'popstate';
 
 export type FireFunction = (element: Document | Element | Window, event: Event) => boolean;
 export type FireObject = {


types/testing-library__dom/events.d.ts
```diff
@@ -80,7 +80,8 @@ export type EventType =
     | 'pointerOut'
     | 'pointerLeave'
     | 'gotPointerCapture'
-    | 'lostPointerCapture';
+    | 'lostPointerCapture'
+    | 'popstate';
```
kentcdodds: Hmmm... Shouldn't this be `popState`?

weyert: I had a look earlier but https://developer.mozilla.org/en-US/docs/Web/API/WindowEventHandlers/onpopstate seems to suggest it is `popstate`?

weyert: Also they seem to  use `popstate` in JSDOM [1] so I think it's correct?

[1] https://github.com/jsdom/jsdom/blob/d6f8a97b5fb7709d6ad0215c1ae95fd4cab58489/lib/jsdom/living/nodes/WindowEventHandlers-impl.js#L18

weyert (approved):

types/testing-library__dom/events.d.ts
weyert: Wait, there is also this PR for `dom-testing-library` which uses `popState`: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41951/files 

I am confused now

eps1lon: [It is `popState`](https://github.com/testing-library/dom-testing-library/pull/433/files#diff-4607e46b15da8ae44c4b47730c6d5963R342)

eps1lon (changes requested):
Seems like we were working on the same things. I very much appreciate the work but this was already implemented in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41951

types/testing-library__dom/events.d.ts
seetdev: Thanks I will cancel this PR since there is another one. And I got it wrong it should be popState

PR #41935, opened 2020-01-29
fix(d3-axis): Allow use of scales with toString property

robertmoura:
---
Changing an existing definition
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/d3/d3-axis#axisTop
---
I've changed the axis domain to allow any scale that has a `toString` domain to create an axis. This helps with previous typings for scales like the [ordinal scale](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/d3-scale/index.d.ts#L1451). For example this was an error: 
```typescript
const bandScale = d3.scaleBand<{ toString(): string }>();

d3.axisLeft(bandScale); // Error here
```

### DIFF
--- a/types/d3-axis/d3-axis-tests.ts
+++ b/types/d3-axis/d3-axis-tests.ts
@@ -35,6 +35,7 @@ let num: number;
 let axisScaleNumber: d3Axis.AxisScale<number>;
 let axisScaleDate: d3Axis.AxisScale<Date>;
 let axisScaleString: d3Axis.AxisScale<string>;
+let axisScaleToString: d3Axis.AxisScale<{ toString(): string }>;
 
 // --------------------------------------------------------------------------
 // Test AxisScale Helper Interface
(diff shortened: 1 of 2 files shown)


denisname:
Will review it soon...

denisname:
The problem with the type `{toString (): string}` is that almost everything is of that type:

```typescript
let x: {toString(): string};

x = {};
x = 2;
x = false;
x = () => null;
x = new Object();
```

So, by expanding the type of `AxisDomain`, you have made it possible to write something like:

```typescript
declare let axis: d3.AxisScale<Function>;
d3.axisLeft(axis); // Should be an error
```

The simplest would be to do:

```typescript
const bandScale = d3.scaleBand()
    .domain(objs.map(_ => _.toString()))
    .range([0, width]);

d3.axisLeft(bandScale);
```

I am not convinced that adding
```typescript
export function scaleBand<Domain extends { toString(): string }>(): ScaleBand<Domain>;
```
in the type definition was a god idea in the first place.

sandersn (changes requested):
requesting changes so the bot understands the status of the PR

robertmoura:
Yeah, I see what you mean. It does limit what you can do with the axis but I don't think it's worth the risk of allowing runtime errors.

Thanks @denisname ! I'll close this pull request.

PR #41957, opened 2020-01-29
 [react-phone-number-input] - Update phone-input props

jlismore:
Replace oudated props with updated forms
Add jsdoc help to most functions
Update util function types

- [ ] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/catamphetamine/react-phone-number-input>> <<https://catamphetamine.github.io/react-phone-number-input/docs/styleguide/index.html#phoneinputwithcountry>>

### DIFF
--- a/types/react-phone-number-input/index.d.ts
+++ b/types/react-phone-number-input/index.d.ts
@@ -1,77 +1,185 @@
 // Type definitions for react-phone-number-input 3.0
 // Project: https://github.com/AdrienEtienne/react-phone-number-input
 // Definitions by: Peter Yang <https://github.com/PeterYangIO>
-//                Adrien Etienne <https://github.com/AdrienEtienne>
-//                Jonathan Fleckenstein <https://github.com/fleck>
+//                 Adrien Etienne <https://github.com/AdrienEtienne>
+//                 Jonathan Fleckenstein <https://github.com/fleck>
+//                 James Lismore <https://github.com/jlismore>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // Minimum TypeScript Version: 3.5
 
-import * as React from "react";
+import * as React from 'react';
 
-export type NumberFormat = "NATIONAL" | "National" | "INTERNATIONAL" | "International";
+/**
+ * Formats value as a "local" phone number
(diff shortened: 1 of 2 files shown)


AdrienEtienne:
Thank you @jlismore looks good to me.
Just one comment, should we move the old code into a v1 folder?
And keep your code as a v2?

I know its a quite new package but still there is breaking changes

AdrienEtienne (approved):
Thank you @jlismore looks good to me.

lucksp:
can the `international` prop be added as per `react-phone-number-input` 
> international: boolean? — If country is specified and international property is true then the phone number can only be input in "international" format for that country, but without "country calling code" part. For example, if country is "US" and international property is not passed then the phone number can only be input in the "national" format for US ((213) 373-4253). But if country is "US" and international property is true then the phone number will be input in the "international" format for US (213 373 4253) without "country calling code" part (+1). This could be used for implementing phone number input components that show "country calling code" part before the input field and then the user can fill in the rest of their phone number in the input field.

types/react-phone-number-input/index.d.ts
```diff
@@ -1,77 +1,185 @@
...
     error?: string;
-    ext?: React.ReactElement;
-    flagComponent?: React.Component<{ country: string, flagsPath: string, flags: FlagsMap, }, object>;
+    flagComponent?: React.ComponentType<{ country: string; flagsPath: string; flags: FlagsMap }>;
     flags?: FlagsMap;
-    flagsPath?: string;
-    getInputClassName?: (params: { disable?: boolean, invalid?: boolean }) => string;
-    id?: string | number;
-    inputComponent?: React.Component<InputComponentProps, object>;
-    international?: boolean;
```
lucksp: why was this item removed?

lucksp: added back in as part of #42996

gcoombe:
re: https://github.com/DefinitelyTyped/DefinitelyTyped/pull/41957/files#diff-7c250e244beb8218eae22310cb1711b50f59f3c393a925b94be69b307150ab3dR63 I think this should be `InputHTMLAttributes<HTMLInputElement>` not `InputHTMLAttributes<string>`. Otherwise it messes the types up for other event handlers like onBlur etc.

PR #41986, opened 2020-01-30
Fix value `Cookies` used in place of a type in @types/koa

spudly:
Fixes error TS2749: 'Cookies' refers to a value, but is being used as a type here.

Please fill in this template.

- [n/a] Add or edit tests to reflect the change. (Run with `npm test`.)

- [n/a] Provide a URL to documentation or source code which provides context for the suggested changes: <<url here>>
- [n/a] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

### DIFF
--- a/types/koa/index.d.ts
+++ b/types/koa/index.d.ts
@@ -716,7 +716,7 @@ declare namespace Application {
         req: IncomingMessage;
         res: ServerResponse;
         originalUrl: string;
-        cookies: Cookies;
+        cookies: typeof Cookies;
         accept: accepts.Accepts;
         /**
          * To bypass Koa's built-in response handling, you may explicitly set `ctx.respond = false;`


types/koa/index.d.ts
```diff
@@ -716,7 +716,7 @@ declare namespace Application {
         req: IncomingMessage;
         res: ServerResponse;
         originalUrl: string;
-        cookies: Cookies;
+        cookies: typeof Cookies;
```
jessetrinity: Is this a common pattern? I'm not sure what you're importing here. Drop a link?

spudly: Line 23 has this:

```ts
import * as Cookies from "cookies";
```

Cookies is a value, not a type (hence the error I'm trying to fix). It is an object representing the exports of the "cookies" module, which has this type:

```ts
{
    default: CookiesStatic;
    get(key: string): string;
    set(key: string, value: any, options?: CookieOptions | undefined): CookiesStatic;
    expire(key: string, options?: CookieOptions | undefined): CookiesStatic;
    defaults: CookieOptions;
    enabled: boolean;
}
```

You can verify that this is correct by reading the [context.cookies section of the koa docs](https://koajs.com/#ctx-cookies-get-name-options-) where it documents `context.cookies.get` and `context.cookies.set`

So, to fix the error that indicated I had was using a value (`Cookies`) where a type was expected, I grabbed the type from the value (using the typescript `typeof` operator) and used that instead.

sandersn (changes requested):
can you add a test that shows what this change fixes?

Edit: Once there are tests I think it will be easier for reviewers to decide if the change makes sense.

spudly:
@typescript-bot Please don't close this. It's still an issue; I'm just very busy.

@sandersn, is there a guide somewhere for adding tests to this repo? If so, I couldn't find it. Any tips you can provide would be much appreciated.

spudly:
@typescript-bot, seriously? Ughhh

sandersn:
@spudly It's only given one line in the readme: https://github.com/DefinitelyTyped/DefinitelyTyped#create-a-new-package. Basically, you put code in koa-tests.ts that will be typechecked with TS 2.8-3.8, but not actually run.

PR #41996, opened 2020-01-31
Adds DragSelect types

IceCreamYou:
This PR adds types for https://github.com/ThibaultJanBeyer/DragSelect ([dragselect](https://www.npmjs.com/package/dragselect) on npm). This is not my first time adding types to DefinitelyTyped, but it has been awhile, and I'm happy to fix things if I made mistakes.

Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/dragselect/index.d.ts
+++ b/types/dragselect/index.d.ts
@@ -3,7 +3,7 @@
 // Definitions by: IceCreamYou <https://github.com/me>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-export class DragSelect {
+declare class DragSelect {
     // These properties store the options passed to the constructor, or the
     // defaults if the relevant options aren't passed. See the docs below.
     selectedClass: string;
@@ -298,3 +298,5 @@ export class DragSelect {
         height: number;
     };
 }
+
+export = DragSelect;
--- a/types/dragselect/index.d.ts
+++ b/types/dragselect/index.d.ts
@@ -1,0 +1,307 @@
+// Type definitions for DragSelect v1.13.1
+// Project: https://ThibaultJanBeyer.com/
+// Definitions by: IceCreamYou <https://github.com/me>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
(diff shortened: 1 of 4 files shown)


types/dragselect/index.d.ts
```diff
@@ -0,0 +1,302 @@
+// Type definitions for DragSelect 1.13
+// Project: https://ThibaultJanBeyer.com/
+// Definitions by: IceCreamYou <https://github.com/me>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+export as namespace DragSelect;
```
weswigham: So this means that in a global context, you'd have to write `new DragSelect.DragSelect`, which seems wrong. You might want to say `export = DragSelect`, and remove the `export` modifier from the class.

weswigham (changes requested):
I'm pretty sure the exports as written here don't match the intended usage just yet.

types/dragselect/index.d.ts
IceCreamYou: I removed the `export as namespace` because you're clearly right about that - `new DragSelect.DragSelect` is not what I meant to express. However, I'm unclear why you're suggesting `export =` could be a solution. In a file with imports/exports, I don't think that would make a difference other than changing the import syntax from `import { DragSelect } from 'dragselect'` to `import DragSelect = require('dragselect')`. In a file with no imports/exports, I don't think `export =` would make the class available as a global, would it? Is there a better way to export a (constructable) class as "available to import and also maybe global"?

weswigham: `export as namespace DragSelect` makes the whole module available in a script context as the global `DragSelect`. If the whole module _is_ the class declaration (doable with an `export =`), then `new DragSelect` in a global context will work. Also, [looking at it,](https://runkit.com/embed/f35hmlvj2oxl) it seems like `import DragSelect = require("dragselect")` is the _correct_ way to import it.

IceCreamYou: Thanks for responding. I updated the PR as you suggested.

PR #42008, opened 2020-01-31
Add render method to Koa Context

ghost:
koa-ejs adds a render method to the ParamterizedContext. This update adds that definition.


jessetrinity:
@christyatsegment filling out the PR template (especially with links to give context for the change) will help reviewers.

PR #42017, opened 2020-01-31
Ember Data: add re-exports for new packages

chriskrycho:
Adds types for [Ember RFC #0395 Ember Data Packages](https://emberjs.github.io/rfcs/0395-ember-data-packages.html).

If adding a new definition:
- [ ] ~~`tslint.json` should be present and it shouldn't have any additional or disabling of rules. Just content as `{ "extends": "dtslint/dt.json" }`. If for reason the some rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]`  and not for whole package so that the need for disabling can be reviewed.~~ **Note:** because these packages are just re-exporting existing package definitions, they *must* use the same `tslint.json` rules as those packages do.


types/ember-data__adapter/error.d.ts
```diff
@@ -0,0 +1,31 @@
+import DS from 'ember-data';
+
+export const AdapterError: typeof DS.AdapterError;
```
weswigham: proposal:
```suggestion
export import AdapterError = DS.AdapterError;
```

reexports all meanings of the thing, which is your intent here, right?

chriskrycho: Huh. I had no idea that was a thing; yes indeed that captures our intent. I'm happy to update all of these accordingly!

chriskrycho:
This is a *really* weird failure; it's entirely unre

PR #42022, opened 2020-01-31
[@testing-library/jest-dom] Add new type for toHaveStyle matcher

lourenci:
Update `toHaveStyle` matcher type according to https://github.com/testing-library/jest-dom/pull/196.

### DIFF
--- a/types/testing-library__jest-dom/index.d.ts
+++ b/types/testing-library__jest-dom/index.d.ts
@@ -28,7 +28,7 @@ declare namespace jest {
         toHaveClass(...classNames: string[]): R;
         toHaveFocus(): R;
         toHaveFormValues(expectedValues: Record<string, unknown>): R;
-        toHaveStyle(css: string): R;
+        toHaveStyle(css: string | object): R;
         toHaveTextContent(text: string | RegExp, options?: { normalizeWhitespace: boolean }): R;
         toHaveValue(value?: string | string[] | number): R;
         toBeChecked(): R;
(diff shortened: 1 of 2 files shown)


types/testing-library__jest-dom/index.d.ts
```diff
@@ -28,7 +28,7 @@ declare namespace jest {
         toHaveClass(...classNames: string[]): R;
         toHaveFocus(): R;
         toHaveFormValues(expectedValues: Record<string, unknown>): R;
-        toHaveStyle(css: string): R;
+        toHaveStyle(css: string | object): R;
```
jgoz: I think this might be better expressed as `string | Record<string, unknown>` similar to the one above.

types/testing-library__jest-dom/test/testing-library__jest-dom-global-tests.ts
```diff
@@ -22,6 +22,7 @@ expect(element).toHaveClass('cls1', 'cls2', 'cls3', 'cls4');
 expect(element).toHaveFocus();
 expect(element).toHaveFormValues({ foo: 'bar', baz: 1 });
 expect(element).toHaveStyle('display: block');
+expect(element).toHaveStyle({ display: 'block' });
```
jgoz: Minor thing, could you add another property that uses a different type, like `width: 100` (just to verify that non-string values are also accepted)?

jgoz (changes requested):
Thanks for updating this @lourenci, just a couple minor suggestions from me

jgoz (approved):
Looks great, thanks for doing this!

PR #42028, opened 2020-02-01
[xml-flow] Add xml-flow@1.0.4 definition 

kodexxx:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/xml-flow/tsconfig.json
+++ b/types/xml-flow/tsconfig.json
@@ -1,0 +1,23 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6"
+        ],
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "strictNullChecks": true,
+        "strictFunctionTypes": false,
+        "baseUrl": "../",
+        "typeRoots": [
+            "../"
+        ],
+        "types": [],
+        "noEmit": true,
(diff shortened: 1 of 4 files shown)


types/xml-flow/tsconfig.json
```diff
@@ -0,0 +1,23 @@
+{
+    "compilerOptions": {
+        "module": "commonjs",
+        "lib": [
+            "es6"
+        ],
+        "noImplicitAny": true,
+        "noImplicitThis": true,
+        "strictNullChecks": true,
+        "strictFunctionTypes": false,
```
jessetrinity: Can you [set to true](https://github.com/definitelytyped/definitelytyped#some-packages-have-no-tslintjson-and-some-tsconfigjson-are-missing-noimplicitany-true-noimplicitthis-true-or-strictnullchecks-true)

types/xml-flow/tslint.json
```diff
@@ -0,0 +1,7 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "strict-export-declare-modifiers": false,
```
jessetrinity: For documentation purposes and to help future contributors, it is preferable to [disable rules with comments](https://github.com/definitelytyped/definitelytyped#linter), and only when necessary.

jessetrinity (changes requested):

PR #42030, opened 2020-02-01
[braintree] Use discriminated union for WebhookNotifications

sgarner:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions
- [ ] ~If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.~

### Summary of Change

This changes the `WebhookNotification` types to know their own values for the `kind` property, so that the compiler can automatically infer types from it as a [discriminated union](https://www.typescriptlang.org/docs/handbook/advanced-types.html#discriminated-unions).

This also requires changing the notification types to be interfaces instead of classes, and changing `WebhookNotification` to be a union type instead of an abstract class.

Result: A guard on the returned `notification.kind` will now have the effect of narrowing the type of `notification` itself, so that explicit casts aren’t always necessary. See the test for a demonstration.

### DIFF
--- a/types/braintree/braintree-tests.ts
+++ b/types/braintree/braintree-tests.ts
@@ -105,9 +105,11 @@ const gateway: BraintreeGateway = new braintree.BraintreeGateway({
     const sampleResponse = await gateway.webhookTesting.sampleNotification(kind, subscriptionId).catch(console.error);
     if (!sampleResponse) return;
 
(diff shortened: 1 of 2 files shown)


smrubin (approved):
Looks good. Thanks for helping to improve the definition @sgarner

PR #42042, opened 2020-02-02
sinon: allow specifying stub type parameters

### LINKED ISSUE #36436: Sinon types are broken with overloads
This was mentioned a little in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/31606 but there isn't an open issue describing it. That PR broke stubbing overloaded methods:

```ts
class Bar {
	public foo(a: string): void;
	public foo(a: number): void;
	public foo(a: string | number): void {};
}

const bar = new Bar();
const barStub = sinon.stub(bar, "foo");

// This line gives an error:
// Argument of type '"test"' is not assignable to parameter of type 'number | SinonMatcher'.
barStub.withArgs("test"); 
```

I don't know if it's possible to fix this without just changing back to `any`, but it does seem like this should work.

43081j:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: N/A

---

Fixes #36436 

Its a bit of a pain having to specify `TObj` and `TKey` too but they are dependencies of the other two type params so we don't have much of a choice. Its better than now: not working at all.

So edge cases where the method can't be inferred can at least specify the type manually.

### DIFF
--- a/types/sinon/ts3.1/index.d.ts
+++ b/types/sinon/ts3.1/index.d.ts
@@ -336,6 +336,8 @@ declare namespace Sinon {
     }
 
     interface SinonSpyStatic {
+        /* tslint:disable:no-unnecessary-generics */
+
         /**
          * Creates an anonymous function that records arguments, this value, exceptions and return values for all calls.
          */
@@ -351,16 +353,23 @@ declare namespace Sinon {
          * The original method can be restored by calling object.method.restore().
          * The returned spy is the function object which replaced the original method. spy === object.method.
          */
-        <T, K extends keyof T>(obj: T, method: K): T[K] extends (
-            ...args: infer TArgs
-        ) => infer TReturnValue
-            ? SinonSpy<TArgs, TReturnValue>
-            : SinonSpy;
+        <
+            TObj,
+            TKey extends keyof TObj,
+            TArgs extends any[] = TObj[TKey] extends (
+                ...args: infer TArgsValue
+            ) => any ? TArgsValue : any[],
+            TReturn = TObj[TKey] extends (
+                ...args: any[]
+            ) => infer TReturnValue ? TReturnValue : any
+        >(obj: TObj, method: TKey): SinonSpy<TArgs, TReturn>;
 
         <T, K extends keyof T>(obj: T, method: K, types: Array<('get'|'set')>): PropertyDescriptor & {
             get: SinonSpy<[], T[K]>;
             set: SinonSpy<[T[K]], void>;
         };
+
+        /* tslint:enable:no-unnecessary-generics */
     }
 
     interface SinonStub<TArgs extends any[] = any[], TReturnValue = any>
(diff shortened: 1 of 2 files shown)


43081j:
@DanTup @murkey @arkhamRejek would you mind letting me know your thoughts on this?

its not as ideal as it somehow magically understanding overloads but it at least means in the edge cases affected, you can specify the types to override the inference. 

```ts
// force the stub to have args [number]
sinon.stub<typeof myObject, 'myMethod', [number]>(myObject, 'myMethod');
```

jessetrinity:
You're listed as owner but I will give others more time to review before looking at this. Thanks!

sandersn:
@43081j Looks like other knowledgeable people haven't reviewed this. Want me to merge it anyway?

My opinion is that this is a dangerous change for a widely-used utility package. 

1. Adding a lot of conditional types means that performance may slow down unpredictably when used with large union types, like those from React.
2. Since the new usage is for explicit type arguments, that means type parameter inference has already failed. In that case we recommend casts anyway, since explicit type parameters are just a prettier hacky cast.

43081j:
@sandersn i wouldn't merge it yet.

you have the same concerns i did. its kinda nice that it works and you can override the inferred types but its still something im not entirely sure about.

it didnt used to infer the types, so by introducing the type inference we broke a lot of builds (quite a while back now). the problem is when someone has overloads, the compiler can't possibly infer the type correctly so it'll pick one of many and often picks the one you didn't want.

for the overload situation, casting doesn't quite work because you can't make a stub in the first place. any idea what we can suggest/do for that?

sandersn:
The explicit `any` argument forces the compiler to use the first overload, or maybe the second? I think that's a decent solution if it doesn't happen very often. If a failure of inference is common, I'd embrace it and try to make to the explicit cast as simple as possible: probably by removing most of the overloads and type parameters and returning a kind of useless type that is easy to cast. Probably `unknown` or `never` are good candidates.

43081j:
overloads on the thing you're stubbing, i meant. here's an example:

```ts
class Foo {
  bar(val: string): void;
  bar(val: number): void;
  bar(val: string|number): void {};
}

const obj = new Foo();
```

If we now do `sinon.stub(obj, 'bar')`, it'll try infer the args and return value. but because there are 3 possibilities, you can't infer the right one 2/3 of the times.

```ts
const stub = sinon.stub(obj, 'bar');
stub.calledWith(5); // it infers that the args to the method were [string] rather than [number]
```

So if we tell people to pass `any` (`sinon.stub<any>(obj, 'bar')`), it just means it will won't be able to infer so the args will also be `any[]`. They can then cast to the right type of stub.

Either works for me, guess its a question of slowness/complexity vs. usability. im not sure myself

sandersn:
Oh, that makes sense. It doesn't change my advice though. I'd go with instructions to add the cast. Basically, inference from overloads is not very good in the compiler, and the overall user experience will be better if users manually work around it instead of the package having complex code to work around it.

sandersn (changes requested):
requesting changes so the bot understands the PR state

PR #42043, opened 2020-02-02
feat(body-parser): bump to 1.19

peterblazejewicz:
- version update
- docs amended
- tests updated
- default DT configuration

Thanks!

- [x] `tslint.json` should be present and it shouldn't have any additional or disabling of rules. Just content as `{ "extends": "dtslint/dt.json" }`.
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/expressjs/body-parser/tree/1.19.0#body-parser

### DIFF
--- a/types/body-parser/body-parser-tests.ts
+++ b/types/body-parser/body-parser-tests.ts
@@ -1,11 +1,6 @@
 import * as http from 'http';
 import express = require('express');
-import {
-    json,
-    raw,
-    text,
-    urlencoded,
-} from 'body-parser';
+import { json, raw, text, urlencoded } from 'body-parser';
 
 const app = express();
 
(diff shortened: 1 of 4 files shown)


types/body-parser/body-parser-tests.ts
```diff
@@ -1,35 +1,37 @@
 import * as http from 'http';
 import express = require('express');
-import {
```
jwalton: Why remove these import tests?

peterblazejewicz: They're valid. I've removed them to test and follow examples from project:
https://github.com/expressjs/body-parser#examples
(as users usually start that way).
I can amend this, to use exported names, let me know.

jwalton: It would be nice if we put the exported names back - it would be nice if they didn't break (although I can't imagine how they would.  :P)

peterblazejewicz: sure, np., done. Changes are squashed, just white-space related changes via Prettier.
Thanks!

PR #42044, opened 2020-02-02
Re-introduce prematurely deprecated rdflib typings

Vinnl:
This reverts commit 974e3815b6d2e6f71adb932c2048dbae82dfe824, introduced in #39502. That deprecated the type definitions, because a prerelease of rdflib (1.0.7-4) had inadvertently been tagged as `latest`. That's now been reverted, so the current latest version (again 1.0.6) no longer includes type definitions - this puts the DT ones back. Once 1.0.7 gets released, they can be removed again.

Please fill in this template.

If adding a new definition:

### DIFF
--- a/notNeededPackages.json
+++ b/notNeededPackages.json
@@ -3781,12 +3781,6 @@
             "asOfVersion": "2.4.0"
         },
         {
-            "libraryName": "rdflib",
-            "typingsPackageName": "rdflib",
-            "sourceRepoURL": "http://github.com/linkeddata/rdflib.js",
-            "asOfVersion": "1.0.5"
-        },
-        {
             "libraryName": "re2",
             "typingsPackageName": "re2",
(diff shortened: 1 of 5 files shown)


peterblazejewicz (approved):
LGTM!
context: linkeddata/rdflib.js#369

jessetrinity (changes requested):
remove from not needed types? 
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/2eebcb4507471f4ce0f42dee6dff303603f6d8ec/notNeededPackages.json#L3802

Vinnl:
@jessetrinity Huh, good catch. Apparently it was moved around in [a later commit](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/35a3ce65890eea789464a9710229bc05b9efcda9), which is why the revert did not catch it.

I manually removed it now.

PR #42062, opened 2020-02-03
SuperAgent: add toJSON methods

morigs:
Those declarations can be useful for `toJSON` method overriding. For that reason, we should return `object`.

https://github.com/visionmedia/superagent/pull/466
https://github.com/visionmedia/superagent/blob/a913704346d9477001820a063b3546f4de9e37be/lib/node/index.js#L972
https://github.com/visionmedia/superagent/blob/a913704346d9477001820a063b3546f4de9e37be/lib/node/response.js#L213

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/visionmedia/superagent/pull/466

### DIFF
--- a/types/superagent/index.d.ts
+++ b/types/superagent/index.d.ts
@@ -117,6 +117,7 @@ declare namespace request {
         status: number;
         statusType: number;
         text: string;
+        toJSON(): object;
         type: string;
         unauthorized: boolean;
         xhr: XMLHttpRequest;
@@ -157,6 +158,7 @@ declare namespace request {
         set(field: string, val: string): this;
         set(field: 'Cookie', val: string[]): this;
         timeout(ms: number | { deadline?: number, response?: number }): this;
+        toJSON(): object;
         type(val: string): this;
         unset(field: string): this;
         use(fn: Plugin): this;


beeequeue (changes requested):
The type should definitely be more specific than `object`. It's very clear what the functions will return in the files you linked.

I don't really get what you mean with functions overriding.

morigs:
> The type should definitely be more specific than `object`. It's very clear what the functions will return in the files you linked.
> 
> I don't really get what you mean with functions overriding.

@BeeeQueue 

By default, these implementations of `toJSON` methods are very minimalistic. So some people may want to override them to get a more verbose JSON representation of request and response. If we'll use the more specific types, for users it'll be more difficult to override these methods to something else

BeeeQueue:
I feel like the proper way to handle that would be to have a `formatResponse(res)` function rather than overriding the function. Most users shouldn't "suffer" incomplete typings because of a niche use-case.

If you really want to override the function you can make the override's return type compatible with the normal types, or override the type with a `declare module "superagent"`.

morigs:
@BeeeQueue I agree with you. But in our case, we have automated tracing which serializes response to JSON. So we cannot see response/request details in tracing. So basically it's better for users to know that there is a custom `toJSON` method

BeeeQueue:
I still don't see the problem. Since this is an issue for your specific implementation only you should have to override the type for your project to your unorthodox solution.

PR #42066, opened 2020-02-03
Add object type to the MJML package

eiskalteschatten:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://mjml.io/documentation/#using-mjml-in-json (the function also takes a JSON object; not just a string)

### DIFF
--- a/types/mjml/index.d.ts
+++ b/types/mjml/index.d.ts
@@ -2,6 +2,7 @@
 // Project: https://github.com/mjmlio/mjml, https://mjml.io
 // Definitions by: aahoughton <https://github.com/aahoughton>
 //                 marpstar   <https://github.com/marpstar>
+//                 eiskalteschatten   <https://github.com/eiskalteschatten>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 interface MJMLParsingOpts {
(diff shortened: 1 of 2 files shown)


marpstar (changes requested):
@eiskalteschatten thanks for this PR. 

this looks pretty good. Ideally, we'd use a more specific type that just `object` here, to provide better support for autocomplete / type-checking. 

After checking the docs, it looks like the schema would be something like the following:

```ts
interface MJMLJsonObject {
    tagName: string;
    attributes: object;
    children: MJMLJsonObject[];
}
```

and then the function signature would be changed to

```ts
declare function mjml2html(inp: string | MJMLJsonObject, opts?: MJMLParsingOpts): MJMLParseResults;
```

Would you be willing to make this small change?

eiskalteschatten:
@marpstar No problem. You're right, that does make more sense. I'll make the change now.

eiskalteschatten:
Just pushed it.

marpstar (changes requested):
Perfect.  

I didn't think about this until after I submitted my last review, but would you mind adding a usage example to the test file?  Apologies for not realizing it earlier. 

Once that's done, I'll happily to approve.

eiskalteschatten:
No problem. I also just noticed we forgot the "content" property. I'll add that as well and make both it and children optional since neither of those are required.

marpstar (approved):
Looks great. Thank you!

eiskalteschatten:
You're welcome!

PR #42068, opened 2020-02-03
Update various socketcluster types to newest versions

DanielRose:

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/SocketCluster/socketcluster-client https://github.com/SocketCluster/sc-broker https://github.com/SocketCluster/scc-broker-client https://github.com/SocketCluster/socketcluster-server https://github.com/SocketCluster/sc-channel

Lots of updates related to the new version of socketcluster.

Note that the socketcluster repository itself changed from a "master" coordinating package to a cli-tool (which needs no more type definitions). However, there are a bunch of packages which refer to the old version and never were updated (since they were obsoleted).


aruniverse:
Shouldnt you have made like a types/socketcluster-client/v13 directory and move all the existing/previous types in there and then update the files in types/socketcluster-client/ with the v15 types?

DanielRose:
@aruniverse Keeping an old version of a type definition around is only needed if either:
* the old type definition is going to be updated later
* another current type definition is using the old version

Neither is the case here. Since types/socketcluster-client was published in v13 you can always get those type definitions from npm.

The reason for socketcluster-server/v14, for example, is because socketcluster changed to a cli-tool in its new version (v15). As a result, it needs no more current type definitions. Instead of deleting its folder, the type definitions for the older version (v14) remain (when it was a "master" controller package). However, those type definitions refer to (and the actual package works with) socketcluster-server in v14. Thus, those older type definitions need to remain, otherwise the repository wouldn't compile anymore.

PR #42070, opened 2020-02-03
aws-lambda: Make Handle.callback optional

dobrynin:
The comment itself already says it's optional. Just updating type to reflect that.

### DIFF
--- a/types/aws-lambda/index.d.ts
+++ b/types/aws-lambda/index.d.ts
@@ -1115,7 +1115,7 @@ export interface LexResult {
 export type Handler<TEvent = any, TResult = any> = (
     event: TEvent,
     context: Context,
-    callback: Callback<TResult>,
+    callback?: Callback<TResult>,
 ) => void | Promise<TResult>;
 
 /**


simonbuchan (changes requested):
Nope, this is a long-standing misunderstanding, see the issue #38342 and previous PRs #30820, #37649, #38828.

To summarize: you are misunderstanding how typescript works here: `callback` being optional means you can instead return a promise when implementing a handler, not that the caller (the AWS Lambda runtime, here) might not provide it, which is what this change is making. If you don't want to use `callback`, you don't need to add it to your signature: `export const fooHandler: FooHandler = async event => ...` works fine. This change will break everyone who is currently using `callback`, as they will now need to check if it's undefined, when it never will be in the real environment.

Please don't argue "but I'm writing tests" or "you could add this as a new `AsyncHandler<E, R>`, etc. here: those discussions have been had already in the above links. Mostly they end up making things worse for both existing and new users, sometimes due to typescript bugs. This typing is accurate: you will receive three arguments and you can either return a promise or nothing. We can't enforce using `callback` if you don't return a promise, and there's no reason you can't use both styles in the same handler.

I'll probably update the comment to try to avoid PRs like this in the future.

dobrynin:
@simonbuchan thanks for the detailed explanation and sorry for not finding those past issues myself. Your explanations in the other posts are clear, but I agree that a comment in the code itself would be useful. Closing.

PR #42075, opened 2020-02-03
[redux-form] Add additional params to ValidateCallback type

jkillian:
Adds a couple missing fields to the `redux-form` `ValidateCallback` type. This change is based off of the [flow types in the codebase](https://github.com/redux-form/redux-form/blob/ca8184418fde8f45833c5d386d97ca29715737d2/src/defaultShouldError.js#L5-L13).

---

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/redux-form/redux-form/blob/ca8184418fde8f45833c5d386d97ca29715737d2/src/defaultShouldError.js#L5-L13

### DIFF
--- a/types/redux-form/lib/reduxForm.d.ts
+++ b/types/redux-form/lib/reduxForm.d.ts
@@ -37,6 +37,8 @@ export interface ValidateCallback<FormData, P, ErrorType> {
     nextProps: P & InjectedFormProps<FormData, P, ErrorType>;
     props: P & InjectedFormProps<FormData, P, ErrorType>;
     initialRender: boolean;
+    lastFieldValidatorKeys: string[];
+    fieldValidatorKeys: string[];
     structure: any;
 }
 
--- a/types/redux-form/redux-form-tests.tsx
+++ b/types/redux-form/redux-form-tests.tsx
@@ -242,7 +242,16 @@ const testFormWithChangeFunctionDecorator = reduxForm<TestFormData, TestFormComp
 
 type TestProps = {} & InjectedFormProps<TestFormData>;
 const Test = reduxForm<TestFormData>({
-    form : "test"
+    form : "test",
+    shouldError: ({
+        values,
+        nextProps,
+        props,
+        initialRender,
+        lastFieldValidatorKeys,
+        fieldValidatorKeys,
+        structure
+    }) => true,
 })(
     class Test extends React.Component<TestProps> {
         handleSubmitForm = (values: Partial<TestFormData>, dispatch: Dispatch<any>, props: {}) => {};


m-b-davis:
Hi @JKillian thanks for the PR!

I wonder if these properties are meant to be exposed...there's no mention in the docs that I can see. I'm not an active redux-form user anymore...could you provide some more context on why we should add this and what it's usage is ?

JKillian:
Good questions @m-b-davis!

These fields are actually used in the default implementations of [`shouldError`](https://github.com/redux-form/redux-form/blob/ca8184418fde8f45833c5d386d97ca29715737d2/src/defaultShouldError.js#L29) and [`shouldWarn`](https://github.com/redux-form/redux-form/blob/ca8184418fde8f45833c5d386d97ca29715737d2/src/defaultShouldWarn.js#L29).

So, if you want to implement a custom but slightly different version of `shouldWarn` or `shouldError`, you need to be able to access these values.

For example, our codebase currently has a snippet like this, where we want `shouldWarn` to keep its original default behavior with some additional logic:

```ts
// code mainly copied from the default redux-form implementation:
// https://github.com/redux-form/redux-form/blob/ca8184418fde8f45833c5d386d97ca29715737d2/src/defaultShouldWarn.js
const shouldWarn = useCallback<NonNullable<Props['shouldWarn']>>(
  ({
    values,
    nextProps,
    props,
    initialRender,
    // @ts-ignore (until redux-form typings are updated)
    lastFieldValidatorKeys,
    // @ts-ignore
    fieldValidatorKeys,
    structure,
  }) => {
    if (initialRender) {
      return true;
    }

    // By default, redux-form doesn't rerun warnings when its `warn` prop changes.
    // We want that to happen though, so we add this block:
    if (props.warn !== nextProps.warn) {
      return true;
    }

    return (
      !structure.deepEqual(values, (nextProps as any).values) ||
      !structure.deepEqual(lastFieldValidatorKeys, fieldValidatorKeys)
    );
  },
  [],
);
```

I suppose I could add typings for `defaultShouldWarn` also, and then could provide some `shouldWarn` implementation which uses the default implementation under-the-hood, but there'd still be the same issue where the default implementation expects to get `lastFieldValidatorKeys` and `fieldValidatorKeys` provided to it.

m-b-davis:
@JKillian thanks for the info. From digging around the source I can't see that this would create any problems, so long as we're sure lastFieldValidatorKeys and fieldValidatorKeys are always passed into shouldError and shouldWarn.

I guess my only worry is for this to be removed from the implementation - after users start relying on it. At that point I supposed it can be removed with a breaking change, and it would be replaced by equivalent properties.

I'll approve for now - any other reviewers want to weigh in?

m-b-davis (approved):

PR #42081, opened 2020-02-04
Add @types for ntlm-client

Borrajo:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/ntlm-client/index.d.ts
+++ b/types/ntlm-client/index.d.ts
@@ -23,10 +23,7 @@ interface NtlmType2 {
     };
 }
 
-declare namespace ntlm {
-    function createType1Message(workstation: string, domain: string): string;
-    function decodeType2Message(type1Message: string): NtlmType2;
-    function createType3Message(type2Message: NtlmType2, username: string, password: string, workstation: string, domain: string): string;
-}
-
-export { ntlm };
+declare function createType1Message(workstation: string, domain: string): string;
+declare function decodeType2Message(type1Message: string): NtlmType2;
+declare function createType3Message(type2Message: NtlmType2, username: string, password: string, workstation: string, domain: string): string;
+export { createType1Message, decodeType2Message, createType3Message };
--- a/types/ntlm-client/index.d.ts
+++ b/types/ntlm-client/index.d.ts
@@ -3,7 +3,7 @@
 // Definitions by: Manuel Borrajo <https://github.com/borrajo>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-import "node";
+/// <reference types="node"/>
 
 interface NtlmType2 {
     flags: number;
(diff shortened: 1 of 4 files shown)


types/ntlm-client/index.d.ts
```diff
@@ -0,0 +1,32 @@
...
+            SERVER: string;
+            DNS: string;
+            FQDN: string;
+            PARENT_DNS: string;
+        },
+        buffer: Buffer;
+    };
+}
+
+declare namespace ntlm {
```
jessetrinity: Does this need to be namespaced?

jessetrinity (changes requested):

types/ntlm-client/index.d.ts
jessetrinity: if you use 

```ts
import ntlm = require('ntlm-client');
```

in the test, you will see that you have to type
```ts
ntlm.ntlm.createType3Message
```
with the extra namespace layer.

Borrajo:
Removed inneeded namespace.

types/ntlm-client/index.d.ts
jessetrinity: Do you need the ntlm types as well? You can export those as interfaces:
https://www.typescriptlang.org/docs/handbook/declaration-files/templates/module-d-ts.html

types/ntlm-client/index.d.ts
```diff
@@ -0,0 +1,29 @@
+// Type definitions for ntlm-client 0.1
+// Project: https://github.com/clncln1/node-ntlm-client
+// Definitions by: Manuel Borrajo <https://github.com/borrajo>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import "node";
```
sheetalkamat: This import is not needed.

sheetalkamat (changes requested):

types/ntlm-client/index.d.ts
Borrajo: if i remove the import, running the lint throws me this error:

`ERROR: 17:17  expect  TypeScript@next compile error: 
Cannot find name 'Buffer'. Do you need to install type definitions for node? Try npm i @types/node and then add node to the types field in your tsconfig.`

Then add 'node' in types in tsconfig 
`"types": [ "node" ]`
And when I run the lint again, error:
`Error: Expected compilerOptions["types"] === []`

So the only way I found to solve it was by `import "node"`. 
Can you think of any other way to solve this?

types/ntlm-client/index.d.ts
Borrajo: That interface does not need to be exported, it is for internal use. The user has nothing to modify.

types/ntlm-client/index.d.ts
sheetalkamat: If thats the case you need to change this to `/// <reference types="node"/>`

PR #42085, opened 2020-02-04
[@babel/core] Add missing merge options

antonku:
This PR adds the following merging options:

- [exclude](https://babeljs.io/docs/en/options#exclude)
- [include](https://babeljs.io/docs/en/options#include)
- [test](https://babeljs.io/docs/en/options#test)
- [overrides](https://babeljs.io/docs/en/options#overrides)

`MatchPattern` function declaration in this PR is based on the [source code](https://github.com/babel/babel/blob/779f00b8e9ce568ece4859501dd79682798c7fbf/packages/babel-core/src/config/config-chain.js#L698-L700) and a bit different from the one in the [docs](https://babeljs.io/docs/en/options#matchpattern). There is an [opened PR](https://github.com/babel/website/pull/2168) for the documentation update.

### DIFF
--- a/types/babel__core/babel__core-tests.ts
+++ b/types/babel__core/babel__core-tests.ts
@@ -58,6 +58,30 @@ checkOptions({ caller: { name: '', tomato: true } });
 checkOptions({ rootMode: 'upward-optional' });
 // $ExpectError
 checkOptions({ rootMode: 'potato' });
+checkOptions({ exclude: '../node_modules' });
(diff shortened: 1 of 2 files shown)


sandersn:
This change looks fine to me, but I don't like merging changes to a widely used package without author signoff.

@yortus @marvinhagemeister @mgroenhoff @Jessidhia can you take a look?

PR #42089, opened 2020-02-04
fix(carbon-components-react) Dropdown itemToElement prop should be ReactNode

sgregoire:
Allowed `Dropdown` to transform items into any non-nullable ReactNode.

Please fill in this template.

- [X] Run `npm run lint carbon-components-react` (or `tsc` if no `tslint.json` is present).
- [X] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://react.carbondesignsystem.com/?path=/story/dropdown--default>>


kalbert312 (changes requested):
I don't think this will work. https://github.com/carbon-design-system/carbon/blob/62024063d8c298e17eea54e77ec593ee01a2a63f/packages/react/src/components/Dropdown/Dropdown.js#L305-L309 it is used as a JSX element tag name. For a React node to work, I think carbon's code would need to be
```javascript
{itemToElement ? itemToElement : itemToString(item)}
```
Because it's a JSX tag, that value of that prop will be passed as the first parameter to `React.createElement`, which does not accept strings except HTML element tags (`div`, `span`, etc).

Seems like `itemToString` would need to be used instead.

sgregoire:
> I don't think this will work. https://github.com/carbon-design-system/carbon/blob/62024063d8c298e17eea54e77ec593ee01a2a63f/packages/react/src/components/Dropdown/Dropdown.js#L305-L309 it is used as a JSX element tag name. For a React node to work, I think carbon's code would need to be
> 
> ```js
> {itemToElement ? itemToElement : itemToString(item)}
> ```
> 
> Because it's a JSX tag, that value of that prop will be passed as the first parameter to `React.createElement`, which does not accept strings except HTML element tags (`div`, `span`, etc).
> 
> Seems like `itemToString` would need to be used instead.

Hello, I changed the code with `itemToElement` as a `NonNullable<React.ReactElement>`.
You can find [here](https://codesandbox.io/s/agitated-moore-z5wje?fontsize=14&hidenavigation=1&theme=dark) an example of expected behaviour with JS, and [here](https://codesandbox.io/s/billowing-voice-hhpwm?fontsize=14&hidenavigation=1&theme=dark) the issue with the current typings.

kalbert312 (approved):

PR #42097, opened 2020-02-04
fix(carbon-components-react) TooltipDefinition missed some props

sgregoire:
Added the missing props `align` and `triggerClassName` from `TooltipDefinition`.

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/carbon-design-system/carbon/blob/master/packages/react/src/components/TooltipDefinition/TooltipDefinition.js>>


types/carbon-components-react/lib/components/TooltipDefinition/TooltipDefinition.d.ts
```diff
@@ -1,10 +1,12 @@
 import * as React from "react";
-import { Direction, ReactDivAttr } from "../../../typings/shared";
+import { Direction, ReactDivAttr, TooltipAlignment } from "../../../typings/shared";
 
 interface InheritedProps extends ReactDivAttr { }
 
 export interface TooltipDefinitionProps extends InheritedProps {
     direction?: Extract<Direction, "bottom" | "top">, // required but has default value
+    align?: TooltipAlignment,
+    triggerClassName?: string
```
kalbert312: It looks like this is not in `carbon-components-react` v.7.6.x so this line will need to be updated: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d56b018b0d6fcbedf0053400c80def5efddd53ab/types/carbon-components-react/index.d.ts#L1 . If it was introduced in v7.7.x, it should be `7.7`, `v8.0.x` would be `8.0` etc.

kalbert312 (changes requested):

types/carbon-components-react/lib/components/TooltipDefinition/TooltipDefinition.d.ts
kalbert312: That line affects the version of the types package published to the registry.

sgregoire: The `align` prop is there in branch `v10.6.4`, see [TooltipDefinition#78](https://github.com/carbon-design-system/carbon/blob/v10.6.4/packages/react/src/components/TooltipDefinition/TooltipDefinition.js#L78).

You can also see in the [package.json#4](https://github.com/carbon-design-system/carbon/blob/v10.6.4/packages/react/package.json) from the same branch that the version `7.6.0` is still there.

kalbert312: Looks like I mixed myself up.

kalbert312 (approved):

PR #42108, opened 2020-02-04
feat(mkdirp): upgrade mkdirp types to 1.0

### LINKED ISSUE #42122: mkdirp typings are for 0.x.0 versions, but 1.0 has breaking changes
The `mkdirp` package's recent 1.0 release completely changed the method signatures, making the `@types/mkdirp` typings for it incorrect. Node terminates (with no error) when we follow the typings, oddly.

Please update the typings for the new API.

JustinBeckwith:
Fixes #42122


JustinBeckwith:
Greetings folks!  Not sure what the right thing to do here is :) The `@types/mkdirp-promise` module I think was relying on types in `@types/mkdirp`, but I don't see how it chooses which version to use.  I am going to try adding a `package.json` that pins to the right semver major version.

peterblazejewicz:
@JustinBeckwith 
Nice, we could join efforts. Can you take from this PR the v0.5 creation to your  PR?
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42136
(I would close mine). Your one seems better aligned with the changes.
Also can you amend the commit/pr title, so the #42122 could be closed automatically?
Let me know,
Thanks!

JustinBeckwith:
@Bartvds greetings!  Would love to get this patched up, if you have a minute for a look :)

types/mkdirp-promise/index.d.ts
```diff
@@ -1,13 +0,0 @@
-// Type definitions for mkdirp-promise 5.0
```
mrmlnc: I think that deleting typings for this package is unnecessary, because they can continue to be used (600k downloads per week).

types/mkdirp/index.d.ts
```diff
@@ -35,6 +53,46 @@ declare namespace mkdirp {
         fs?: FsImplementationSync;
     }
 
-    function sync(dir: string, opts?: Mode | OptionsSync): Made;
+    /**
+     * Synchronously create a new directory and any necessary subdirectories at
```
mrmlnc: We can use `code block` to highlight the code as it is done in the documentation. This is useful when reading documentation in the editor.

Here and below.

types/mkdirp/index.d.ts
```diff
@@ -35,6 +53,46 @@ declare namespace mkdirp {
...
+     * implementation is overridden.
+     */
+    function manual(dir: string, opts?: Mode | Options): Promise<string|undefined>;
+
+    /**
+     * Use the manual implementation (not the native one). This is the default
+     * when the native implementation is not available or the stat/mkdir
+     * implementation is overridden.
+     */
+    function manualSync(dir: string, opts?: Mode | Options): string|null;
```
mrmlnc: Shouldn't there be a `undefined` (instead of `null`) in this place?

Code:

* https://github.com/isaacs/node-mkdirp/blob/7cd4e9eb9a4376b551414a154b26697cf21d9b93/lib/mkdirp-manual.js#L36
* https://github.com/isaacs/node-mkdirp/blob/7cd4e9eb9a4376b551414a154b26697cf21d9b93/lib/mkdirp-manual.js#L43

`fs.mkdirSync`:

* https://nodejs.org/dist/latest-v13.x/docs/api/fs.html#fs_fs_mkdirsync_path_options

types/mkdirp/index.d.ts
```diff
@@ -35,6 +53,46 @@ declare namespace mkdirp {
...
+     * overridden.
+     */
+    function native(dir: string, opts?: Mode | Options): Promise<string|undefined>;
+
+    /**
+     * Use the native implementation (not the manual one). This is the default
+     * when the native implementation is available and stat/mkdir are not
+     * overridden.
+     */
+    function nativeSync(dir: string, opts?: Mode | Options): string|null;
```
mrmlnc: https://github.com/isaacs/node-mkdirp/blob/7cd4e9eb9a4376b551414a154b26697cf21d9b93/lib/mkdirp-native.js#L33

The `mkdirpManualSync` method returns `string` or `undefined`.

The `findMade` too:

* https://github.com/isaacs/node-mkdirp/blob/7cd4e9eb9a4376b551414a154b26697cf21d9b93/lib/mkdirp-native.js#L27
* https://github.com/isaacs/node-mkdirp/blob/7cd4e9eb9a4376b551414a154b26697cf21d9b93/lib/find-made.js#L16

types/mkdirp/index.d.ts
```diff
@@ -35,6 +53,46 @@ declare namespace mkdirp {
...
+     * it will be treated as the opts.mode. If opts.mode isn't specified, it
+     * defaults to 0o777 & (~process.umask()).
+     * You can optionally pass in an alternate fs implementation by passing in
+     * opts.fs. Your implementation should have opts.fs.mkdirSync(path, mode)
+     * and opts.fs.statSync(path).  You can also override just one or the other
+     * of mkdirSync and statSync by passing in opts.statSync or opts.mkdirSync,
+     * or providing an fs option that only overrides one of these.
+     * @returns Returns the first directory that had to be created, or undefined if everything already exists.
+     */
+    function sync(dir: string, opts?: Mode | Options): string|null;
```
mrmlnc: The `sync` method calls `mkdirpNativeSync` or `mkdirpManualSync` that returns `string` or `undefined`:

https://github.com/isaacs/node-mkdirp/blob/7cd4e9eb9a4376b551414a154b26697cf21d9b93/index.js#L25

mrmlnc (changes requested):
Thanks for the contribution!

Looks like we have some confusion about the return types. Please check the returned types.

types/mkdirp-promise/index.d.ts
JustinBeckwith: This is a great topic for conversation :) I removed it based on the conversation here:
https://github.com/microsoft/types-publisher/pull/736

If I don't remove these types, we would need to make changes to them to make this change work.  `@types/mkdirp-promise` assumes that it can use `@types/mkdirp` without pinning to any particular semver range.  When making this change and keeping it around, it caused build failures. 

The alternative to this approach (I think..) is to introduce a package.json to `@types/mkdirp-promise`, and pin that to the `~0.5.0` range.  I started down that path, which led to the whole whitelist thing above and well...  yeah, I followed the advice in that thread :)  The part that swayed me was knowing that a.) updates to `@types/mkdirp-promise` are unlikely at this stage, and b.) the package will still exist - it won't get taken down or anything.

All that having been said - I'm super flexible here!

types/mkdirp/index.d.ts
JustinBeckwith: Hah, good catch!  I was following the information specifically called out in the readme:
https://github.com/isaacs/node-mkdirp#mkdirpsyncdir-opts---stringnull

But the code don't lie

mrmlnc (approved):

types/mkdirp-promise/index.d.ts
sandersn: Also, we can restore the mkdir-promise source to the DT repo at any time in case we need to publish a new version &mdash; although the PR that did that would still need to introduce a package.json etc etc.

PR #42109, opened 2020-02-04
Add graphite module declarations

ctumolosus:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/graphite/graphite-tests.ts
+++ b/types/graphite/graphite-tests.ts
@@ -1,0 +1,75 @@
+import graphite = require('graphite');
+
+interface Metrics {
+    available: number;
+    free: number;
+    total: number;
+}
+
+interface OptionalMetrics {
+    disk?: Metrics;
+    memory?: Metrics;
+}
+
+const memory: Metrics = {
+    available: 30000000,
+    free: 25000000,
+    total: 40000000,
+};
+
+const os: OptionalMetrics = {
+    memory,
+};
+
+const client = graphite.createClient('plaintext://graphite.example.org:2003/');
+
+// Check metric values
+client.write({ foo: 23 });
+client.write({ foo: { bar: { qux: 23 } } });
+// Check interfaces as metric values
+client.write(memory);
+// Check interfaces with optional fields as metric values
+client.write(os);
+// Check nested interfaces as metric values
+client.write({ os: { memory } });
+
+// The following metric values should yield compilation errors, but these are
+// ignored because of the use of the any keyword as type annotation.
+client.write(1);
+client.write('foo');
+client.write(false);
+client.write([]);
+client.write({ foo: 'foo' });
+client.write({ foo: { baz: 'baz' } });
+client.write({ foo: new Date() });
+client.write({ foo: [] });
+
+// Check overloaded arguments
+client.write({ foo: 23 }, (err: any) => {});
+client.write({ foo: 23 }, Date.now(), (err: any) => {});
+
+// Check metric and tag values
+client.writeTagged({ foo: 23 }, { foo: 'baz' });
+client.writeTagged({ foo: { bar: 23 } }, { foo: { bar: 'qux' } });
+
+// Check overloaded arguments
+client.writeTagged({ foo: 23 }, { foo: 'baz' }, (err) => {});
+client.writeTagged({ foo: 23 }, { foo: 'baz' }, Date.now(), (err) => {});
+
+// Check utilities
+
+// Graphite metric values should be numeric, but outside the graphite context
+// flatten can be used to flatten nested object containing any value.
+graphite.flatten({ foo: 23 });
+graphite.flatten({ foo: { bar: 1, baz: '2' }, qux: true });
+graphite.flatten(memory);
+
+// Check other parameters
+graphite.flatten({ foo: 23 }, { baz: 'bar' });
+graphite.flatten({ foo: 23 }, { baz: 'bar' }, 'prefix');
+
+graphite.appendTags({ foo: 23 }, { foo: 'foo' });
+
+// Should yield compilation error when a nested object is provided, but won't
+// happen when using any as the type annotation.
+graphite.appendTags({ foo: { bar: 23 } }, { foo: 'foo' });
--- a/types/graphite/index.d.ts
+++ b/types/graphite/index.d.ts
@@ -1,0 +1,24 @@
+// Type definitions for graphite 0.1
+// Project: https://github.com/felixge/node-graphite
+// Definitions by: Christian Muñoz <https://github.com/ctumolosus>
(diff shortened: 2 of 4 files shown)


ctumolosus:
Initial commit added intentionally to show direction I tried to take prior to switching to using `any` as the type annotation for all parameters and return types. Technically, the [`graphite`](https://www.npmjs.com/package/graphite) module accepts anything but won't report invalid values. For example:

```js
import graphite from 'graphite';
const client = graphite.createClient('...');
client.write({ foo: new Date() }); // reports {}
client.write({ foo: 1 }); // reports { foo: 1 }
client.write({ foo: { bar: { qux: 3 } } }); // reports { 'foo.bar.qux': 3 }
client.write('foo'); // reports { 0: 'f', 1: 'o', 2: 'o' }
// Etc.
```

ctumolosus:
This pull requests introduces type definitions for the [`graphite`](https://www.npmjs.com/package/graphite) package.

P.S.

I intended to use the following types for the `metrics` and `tags` parameters, but apparently I cannot use an index signature to constraint the type for interface properties. So, I decided to use `any` as the type annotation instead.

```ts
type MetricValue = number | MetricObject;
type MetricObject = { [key: string]: MetricValue };

type TagValue = string | TagObject;
type TagObject = { [key: string]: TagValue };

declare class GraphiteClient {
  write(metrics: MetricObject): void;
  writeTagged(metrics: MetricObject, tags: TagObject): void;
}
```

jessetrinity:
>I intended to use the following types for the metrics and tags parameters, but apparently I cannot use an index signature to constraint the type for interface properties. So, I decided to use any as the type annotation instead.

Do you have an example of the usage that wasn't working?

ctumolosus:
You can preview the example below in the [TypeScript Playground](http://www.typescriptlang.org/play/?ssl=1&ssc=1&pln=58&pc=40#code/C4TwDgpgBAshwCcCWBjAagQwDYFdoF4oA7HAWwCMIEoAfWeZFAeXICsIVgBuAWAChQkeolQt2nKIQDeUANoBrCCABcUAM4iiAcwC6quCPTY8UAL68+-AGY4inJAHsiUAO7JgEABSkGqNft9mNg5gAEpVADcHJAATKClTfn4kIg8EKwwUaDhSBwQQA0Z4-igoYAdgbFUSCioLUpw1CBjqskoEeqgrBAgIVtqO-kTLPhS0jKyoJjVC1GK+Up9c-IB+AOWCwK5SoaS+FCcNKCW8lXoN2ZRJedLu3tUARgAGABoSqEbmx9f38sqsVQAJh+fHMewORCODn8UxmgWuCQs-AA9MioABhDBEIgVKBgBAOCKxaBYqBjKgTElqKAYY7wsAYBAYHxpZR7VFQACCCC0ZAgqSgDisZXA0AA5DlTpcxWTqTjgDS1GokFoiBhyFhoOU8YzmfAqILhYJxZcxCExQA6FFo0oASSIMQgAA91Cq1cAcD1ZcckEqUloyc5jVAJRALoFLZ5AQBmAAsAFZQvw3EgPN4w6dQki+BzOa48vJGQ5bC09sH0Q5SAzgEgNRAADwAFReUAAqgA+BHvWQAaUDUEUICFUEbehHvZ0UGdHgd1NbUBW457k+qEAidV2IxsdhrTlc7ggaEBTfb6cMMIrVYwNbrTZbl0wuAg7fCUCisXiw2TB6P6Y2WagDkiFQCAAEJ2TRAAJBwXDXKgWxcPJgAAC2ICp-TKZDr0w318wQQsCRLKAYgcCA5VxTIsjABUHGoxw1SwPECUgBAazIltrVcZDUFQ3DKIgaj1U1fsUNwlAMCaKBKHEz4oHErBNWoUhGgVCBU2Qg18UJYkaS6LBsLyKBOKIMiPDiBxggkYcanaNQrRGXMeT5AVh2DMVpmlb15UVZVVSErUHB1JkWQNVzRRDS9q1rTV6xkE5VnWKV4ToEsICsFJmm2Ux70CR88HbS1OJHUVqWHLSWNAEN4pAGVGWgFIDivG9NXsjlSmKoRQ3DQxaA+B00oymIZVw7yJN8tU6zKQK3NS9KTKG1qbXajqTQzfJPJGii-T8ybtRm-q5uaSMYwTJM+BTDxf2hLMIK5LA1AcFsULI6BEKIMUFUQ-D7M4xtwq6pLDGG8iFTGt1-KmkVOpM9cEGO6NAUBM6LsPQFPDitaQDMG6HLRP7OsldaIy8rbxohvb-phqh4cR5GfzRmRoVUDGNjMbHs1zBVROpMBogFW1iI4Yk4m1WSsRAC0oCAA):


```ts
type MetricValue = number | MetricObject;
type MetricObject = { [key: string]: MetricValue };

function write(metrics: MetricObject): void {}

interface MemoryMetric {
  total: number;
  used: number;
  free: number;
}

interface OsMetric {
  memory?: MemoryMetric;  
}

const memory: MemoryMetric = {
  free: 10,
  used: 10,
  total: 20,
};

const os: OsMetric = {};

// Cannot provide an interface as a metric parameter:

// Argument of type 'MemoryMetric' is not assignable to parameter of type 'MetricObject'.
//   Index signature is missing in type 'MemoryMetric'.(2345)
write(memory);

// A workaround:

type Compatible<T, U> = {
  [K in keyof T]: T[K] extends U ? T[K] : never;
}

function writeV2<T>(metrics: Compatible<T, MetricValue>): void {}

writeV2(memory); // nice!

// However, worth noting that this workaround does not accept optional properties, 
// which is acceptable in this case because caller must either provide a flat or 
// nested object of numbers.

// Argument of type 'OsMetric' is not assignable to parameter of type 'Compatible<{ memory?: MemoryMetric | undefined; }, MetricValue>'.
//  Types of property 'memory' are incompatible.
//    Type 'MemoryMetric | undefined' is not assignable to type 'undefined'.
//      Type 'MemoryMetric' is not assignable to type 'undefined'.(2345)
writeV2(os);

// Also, these won't work.

// Type 'MemoryMetric' is not assignable to type 'never'.(2322)
writeV2({ memory });

// Type 'MemoryMetric' is not assignable to type 'never'.(2322)
writeV2({ os: { memory } });

// At this point I decided to use any. 
```

ctumolosus:
I believe it's related to this issue: 
https://github.com/microsoft/TypeScript/issues/15300

jessetrinity:
Nice find. Does it makes sense to change those interfaces to [types](https://github.com/microsoft/TypeScript/issues/15300#issuecomment-332366024)?

ctumolosus:
I rather not enforce a preference between interfaces and type aliases. 

If acceptable by the community, I would be happy to introduce the workaround with the caveat that nesting interfaces in object literals will yield compilation errors.

PR #42111, opened 2020-02-05
feat(node): Add Buffer.from(object) override

wemeetagain:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_object_offsetorencoding_length

### DIFF
--- a/types/node/globals.d.ts
+++ b/types/node/globals.d.ts
@@ -244,7 +244,7 @@ declare class Buffer extends Uint8Array {
      * A `TypeError` will be thrown if {obj} has not mentioned methods or is not of other type appropriate for `Buffer.from()` variants.
      * @param obj An object supporting `Symbol.toPrimitive` or `valueOf()`.
      */
-    static from(obj: object, byteOffset?: number, length?: number): Buffer;
+    static from(obj: { valueOf(): string | object } | { [Symbol.toPrimitive](hint: 'string'): string }, byteOffset?: number, length?: number): Buffer;
     /**
      * Creates a new Buffer containing the given JavaScript string {str}.
      * If provided, the {encoding} parameter identifies the character encoding.
--- a/types/node/globals.d.ts
+++ b/types/node/globals.d.ts
@@ -240,6 +240,12 @@ declare class Buffer extends Uint8Array {
     static from(data: number[]): Buffer;
     static from(data: Uint8Array): Buffer;
     /**
+     * Creates a new buffer containing the coerced value of an object
+     * A `TypeError` will be thrown if {obj} has not mentioned methods or is not of other type appropriate for `Buffer.from()` variants.
+     * @param obj An object supporting `Symbol.toPrimitive` or `valueOf()`.
+     */
+    static from(obj: object, byteOffset?: number, length?: number): Buffer;
+    /**
      * Creates a new Buffer containing the given JavaScript string {str}.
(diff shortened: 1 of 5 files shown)


types/node/globals.d.ts
```diff
@@ -239,6 +239,12 @@ declare class Buffer extends Uint8Array {
      */
     static from(data: number[]): Buffer;
     static from(data: Uint8Array): Buffer;
+    /**
+     * Creates a new buffer containing the coerced value of an object
+     * A `TypeError` will be thrown if {obj} has not mentioned methods or is not of other type appropriate for `Buffer.from()` variants.
+     * @param obj An object supporting `Symbol.toPrimitive` or `valueOf()`.
+     */
+    static from(obj: object, byteOffset?: number, length?: number): Buffer;
```
SimonSchick: Can we narrow this to something more specific?
As describe by the `@param` directive we could create an interface that supports either toPrimitive or valueOf?

wemeetagain: :+1:
Are you thinking something like: `{ valueOf(): any } | { [Symbol.toPrimitive](hint: string): any }` ?

SimonSchick: See https://github.com/nodejs/node/blob/master/lib/buffer.js#L311

I find the broadness of this function somewhat concerning, lol.

wemeetagain: I see,
`{ valueOf(): string | object } | { [Symbol.toPrimitive](hint: 'string'): string }`

lmao, yeah, its surprisingly vague

SimonSchick (approved):
Going to approve this since it does match the API.
On a personal note: I wish this didn't exist, these implicit conversions are very bad for type safety and we will probably see a bug report due to this in the future 🤷‍♂

jessetrinity:
Roses are red
Violets are blue
BenchmarkPR failed so I've run it anew

sandersn (changes requested):
I don't think we should merge something in node unless it meets a real need. Obscure parts of the API may not actually be used.

@wemeetagain Can you give examples of types that would use this overload and would need a cast otherwise? I'd especially like to see usage in a real code base.

wemeetagain:
This is meeting a real need. I am using this part of the API.

I have a library that wraps an opaque datastructure (in our case, a merkle tree) with a `Proxy` that exposes an API that resembles `Uint8Array`. These `Proxy` objects are used interchangably with real `Uint8Array`s as long as the limited API is used. Included in this is an implementation of `valueOf` that presents a real Uint8Array when necessary.

something like:
```typescript
Uint8ArrayLike {
  [index: number]: number;
  length: number;
  // In order for `Uint8Array` to fulfil `Uint8ArrayLike`,
  // valueOf return type must be `object`
  // because Uint8Array share the `valueOf` typing inherited from `Object`.
  // otherwise this would be:
  // valueOf(): Uint8Array;
  valueOf(): object; 
}

const a: Uint8ArrayLike = new Uint8Array();
const b: Uint8Arraylike = opaquething();

// I would like to do:

Buffer.from(a);
Buffer.from(b);

// I must currently do:

Buffer.from(a as Uint8Array);
Buffer.from(b as Uint8Array);
```

sandersn (approved):

sandersn:
Sounds reasonable, although I still have reservations.

PR #42120, opened 2020-02-05
Fix onRangeChange types from CalendarProps

catruzz:
The types declaration of onRangeChange function seem to be wrong: typescript returns error if I try to manipulate the returning range array like the following:
```
<BigCalendar
  onRangeChange={
    range => {
      range.map((v, k) => { ... }) // typescript error here
    }
  }
/>
```

Typescript error:
```
Property 'map' does not exist on type 'Date[] | { start: stringOrDate; end: stringOrDate; }'.
Property 'map' does not exist on type '{ start: stringOrDate; end: stringOrDate; }'.
```

Furthermore, if you look at the variables returned by the function with a spread operator, you'll get an array of 2 elements:
[0] => range: Date[]
[1] => view: View

### DIFF
--- a/types/react-big-calendar/index.d.ts
+++ b/types/react-big-calendar/index.d.ts
@@ -268,7 +268,7 @@ export interface CalendarProps<TEvent extends object = Event, TResource extends 
     onDoubleClickEvent?: (event: TEvent, e: React.SyntheticEvent<HTMLElement>) => void;
     onSelectEvent?: (event: TEvent, e: React.SyntheticEvent<HTMLElement>) => void;
     onSelecting?: (range: { start: stringOrDate; end: stringOrDate }) => boolean | undefined | null;
-    onRangeChange?: (range: Date[] | { start: stringOrDate; end: stringOrDate }) => void;
+    onRangeChange?: (range: Date[], view: View) => void;
     selected?: any;
     views?: ViewsProps;
     drilldownView?: View | null;
--- a/types/react-big-calendar/index.d.ts
+++ b/types/react-big-calendar/index.d.ts
@@ -268,7 +268,7 @@ export interface CalendarProps<TEvent extends object = Event, TResource extends 
     onDoubleClickEvent?: (event: TEvent, e: React.SyntheticEvent<HTMLElement>) => void;
     onSelectEvent?: (event: TEvent, e: React.SyntheticEvent<HTMLElement>) => void;
     onSelecting?: (range: { start: stringOrDate; end: stringOrDate }) => boolean | undefined | null;
-    onRangeChange?: (range: Date[] | { start: stringOrDate; end: stringOrDate }, view: View) => void;
+    onRangeChange?: (range: Date[] | { start: Date; end: Date }, view?: View) => void;
     selected?: any;
     views?: ViewsProps;
     drilldownView?: View | null;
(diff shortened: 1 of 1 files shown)


types/react-big-calendar/index.d.ts
```diff
@@ -268,7 +268,7 @@ export interface CalendarProps<TEvent extends object = Event, TResource extends
     onDoubleClickEvent?: (event: TEvent, e: React.SyntheticEvent<HTMLElement>) => void;
     onSelectEvent?: (event: TEvent, e: React.SyntheticEvent<HTMLElement>) => void;
     onSelecting?: (range: { start: stringOrDate; end: stringOrDate }) => boolean | undefined | null;
-    onRangeChange?: (range: Date[] | { start: stringOrDate; end: stringOrDate }) => void;
+    onRangeChange?: (range: Date[], view: View) => void;
```
eps1lon: This doesn't seem to be what the docs state:

> Returns an Array of dates or an object with start and end dates for BUILTIN views.

-- http://intljusticemission.github.io/react-big-calendar/examples/index.html#prop-onRangeChange

eps1lon (changes requested):
Thanks for opening this issue.

Could you add a codesandbox that illustrates this type behavior? It seems to not match the documentation. If the documentation is outdated, please open an issue in the actual repository as well.

types/react-big-calendar/index.d.ts
PaitoAnderson: https://github.com/intljusticemission/react-big-calendar/blob/master/src/Calendar.js#L253 ◀️ Code comment seems to agree with @catruzz, it says the type is Date only though :/

PaitoAnderson: Clicking next/prev repros this on the console: https://codesandbox.io/s/weathered-dawn-4rijx

PaitoAnderson (changes requested):
Could you make `stringOrDate` just `Date` or show an example of string?

types/react-big-calendar/index.d.ts
catruzz: Thanks to the code provided by @PaitoAnderson I found out that `onRangeChange` actually returns an array of Dates or an object having `start` and `end` keys (as stated in the docs) depending on the current view (that is also returned) and Typescript returns errors accordingly.

1. Day: returns array with one element
2. Week: returns array of 7 elements
3. Month: returns object `{start: ..., end: ...}`
4. Agenda: returns object `{start: ..., end: ...}`

Here you can see the working example with the `view` parameter evidence (Typescript is included but doesn't fire any error here though, as it does in my VSCode)
https://codesandbox.io/s/nifty-lake-t99ky

types/react-big-calendar/index.d.ts
```diff
@@ -268,7 +268,7 @@ export interface CalendarProps<TEvent extends object = Event, TResource extends
     onDoubleClickEvent?: (event: TEvent, e: React.SyntheticEvent<HTMLElement>) => void;
     onSelectEvent?: (event: TEvent, e: React.SyntheticEvent<HTMLElement>) => void;
     onSelecting?: (range: { start: stringOrDate; end: stringOrDate }) => boolean | undefined | null;
-    onRangeChange?: (range: Date[] | { start: stringOrDate; end: stringOrDate }) => void;
+    onRangeChange?: (range: Date[] | { start: stringOrDate; end: stringOrDate }, view: View) => void;
```
eps1lon: I can follow the implementation but this is also missing from the docs, no? Would be really nice if you could contribute this to the docs as well.

eps1lon (approved):

types/react-big-calendar/index.d.ts
catruzz: Sorry but I can't find a way to edit the docs. It seems that the page http://intljusticemission.github.io/react-big-calendar/examples/index.html#prop-onRangeChange isn't available on react-big-calendar repo

eps1lon (changes requested):
Please add links to documentation and a test.

eps1lon:
@RyanCavanaugh The bot is misbehaving.

PR #42123, opened 2020-02-05
fix(pino): Add missing params in PrettyOptions

Pegase745:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/pinojs/pino-pretty#options

### DIFF
--- a/types/pino/index.d.ts
+++ b/types/pino/index.d.ts
@@ -465,7 +465,7 @@ declare namespace P {
         /**
          * Format output of message, e.g. {level} - {pid} will output message: INFO - 1123 Default: `false`.
          */
-        messageFormat?: boolean | string;
+        messageFormat?: false | string;
         /**
          * If set to true, will add color information to the formatted output message. Default: `false`.
          */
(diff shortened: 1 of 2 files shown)


types/pino/index.d.ts
```diff
@@ -461,6 +462,10 @@ declare namespace P {
          * The key in the JSON object to use for timestamp display. Default: "time".
          */
         timestampKey?: string;
+        /**
+         * Format output of message, e.g. {level} - {pid} will output message: INFO - 1123 Default: `false`.
+         */
+        messageFormat?: boolean | string;
```
AdamVig: Based on [the code](https://github.com/pinojs/pino-pretty/blob/c0c6b52ecd58a3844d276ad1a5d8caa263ce0081/lib/utils.js#L200), it looks like `true` would never be a valid value for this property, so we can narrow the type accordingly:
```suggestion
        messageFormat?: false | string;
```

AdamVig (approved):

PR #42125, opened 2020-02-05
fix(carbon-components-react) TileGroup's onChange signature: all arguments are mandatory

sgregoire:
The `onChange` prop is called in [TileGroup.js#110](https://github.com/carbon-design-system/carbon/blob/v10.6.4/packages/react/src/components/TileGroup/TileGroup.js#L110)
The new selection, the name, and the event are sent.

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://github.com/carbon-design-system/carbon/blob/v10.6.4/packages/react/src/components/TileGroup/TileGroup.js#L110>>


kalbert312 (approved):
Looks good. I guess there's no current possibility for that callback to be called with `null/undefined` based on the usages.

PR #42128, opened 2020-02-05
fix(carbon-components-react) Updated DataTableCustomSelectionProps accordingly to the Element's props

sgregoire:
The issue can be reproduced [here](https://codesandbox.io/s/sleepy-silence-8wk9d?fontsize=14&hidenavigation=1&theme=dark) at line 52, once the `radio` will be fixed, `checked` will crash.

The PR fixes two issues on the prop generation for DataTableRow.

* The prop `checked`: R["isSelected"] can lead to `undefined`.
* The prop `radio`: The type was misleading.
  * Since the prop is not mandatory ([see here](https://github.com/carbon-design-system/carbon/blob/v10.6.4/packages/react/src/components/DataTable/TableSelectRow.js#L78)), I changed it to optional.
  * The method `Extract` expects the second argument property to be a type and not a value, it led to some compilation issues.

Please fill in this template.

I could not provide an URL for this.

### DIFF
--- a/types/carbon-components-react/carbon-components-react-tests.tsx
+++ b/types/carbon-components-react/carbon-components-react-tests.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import {
     AccordionItem,
-    DataTable, DataTableCustomRenderProps,
+    DataTable,
+    DataTableCustomRenderProps,
     DataTableHeader,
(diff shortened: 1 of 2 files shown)


kalbert312 (approved):
Cool test.

PR #42142, opened 2020-02-06
[angular-aria] Add types for AngularJS $aria service

chivesrs:
Please fill in this template.

If adding a new definition:

The `tslint.json` is explicitly modified to disable interface naming rules. The reason for this is that AngularJS typings follow the pattern listed [here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/angular/README.md#services-and-other-injectables).

Regarding documentation on the actual types, see:
https://docs.angularjs.org/api/ngAria
https://docs.angularjs.org/guide/accessibility
https://github.com/angular/angular.js/blob/master/src/ngAria/aria.js#L97
https://docs.angularjs.org/api/ngAria/provider/$ariaProvider#config
https://github.com/angular/angular.js/blob/master/src/ngAria/aria.js#L205

Also try some sample code:
```
s = document.createElement('script'); s.src = "https://ajax.googleapis.com/ajax/libs/angularjs/1.7.9/angular-aria.js"; document.getElementsByTagName('head')[0].appendChild(s);

i = angular.injector(); i.loadNewModules(['ngAria']); aria = i.get('$aria'); console.log(aria); console.log(aria.config('tabindex')); console.log(aria.config('ariaInvalid'));
```

Paste into https://angularjs.org in the JS console. First line loads `ngAria` script (it's not built into main AngularJS, but it is a core library), second line instantiates the service and uses it.

/cc @peterblazejewicz

### DIFF
--- a/types/angular-aria/angular-aria-tests.ts
+++ b/types/angular-aria/angular-aria-tests.ts
@@ -1,5 +1,5 @@
 function testConfig($aria: angular.aria.IAriaService): void {
-    // $ExpectType void
+    // $ExpectType boolean
     $aria.config('tabindex');
 
     // $ExpectError
--- a/types/angular-aria/angular-aria-tests.ts
+++ b/types/angular-aria/angular-aria-tests.ts
@@ -1,0 +1,7 @@
+function testConfig($aria: angular.aria.IAriaService): void {
+    // $ExpectType void
+    $aria.config('tabindex');
+
+    // $ExpectError
+    $aria.config('unknown-string');
+}
--- a/types/angular-aria/index.d.ts
+++ b/types/angular-aria/index.d.ts
@@ -2,6 +2,7 @@
 // Project: http://angularjs.org, https://docs.angularjs.org/api/ngAria
 // Definitions by: Chives <https://github.com/chivesrs>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.3
 
 import * as angular from 'angular';
 
--- a/types/angular-aria/index.d.ts
+++ b/types/angular-aria/index.d.ts
@@ -1,0 +1,19 @@
+// Type definitions for angular-aria 1.7
+// Project: http://angularjs.org, https://docs.angularjs.org/api/ngAria
+// Definitions by: Chives <https://github.com/chivesrs>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import * as angular from 'angular';
+
+declare module 'angular' {
+    namespace aria {
+        type IAriaAttribute = 'ariaHidden'|'ariaChecked'|'ariaReadonly'|'ariaDisabled'|'ariaRequired'|'ariaInvalid'|'ariaValue'|'tabindex'|'bindKeydown'|'bindRoleForClick';
+
+        /**
+         * $aria service.
+         */
+        interface IAriaService {
+            config(attribute: IAriaAttribute): boolean;
+        }
+    }
+}
--- a/types/angular-aria/tslint.json
+++ b/types/angular-aria/tslint.json
@@ -1,0 +1,1 @@
+{ "extends": "dtslint/dt.json" }
--- a/types/angular-aria/tslint.json
+++ b/types/angular-aria/tslint.json
@@ -1,1 +1,6 @@
-{ "extends": "dtslint/dt.json" }
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "interface-name": false
+    }
+}
--- a/types/angular-aria/tsconfig.json
+++ b/types/angular-aria/tsconfig.json
@@ -2,6 +2,7 @@
     "compilerOptions": {
         "module": "commonjs",
         "lib": [
+            "dom",
             "es6"
         ],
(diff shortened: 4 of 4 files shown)


types/angular-aria/index.d.ts
```diff
@@ -0,0 +1,20 @@
...
+
+declare module 'angular' {
+    namespace aria {
+        type IAriaAttribute = 'ariaHidden'|'ariaChecked'|'ariaReadonly'|'ariaDisabled'|'ariaRequired'|'ariaInvalid'|'ariaValue'|'tabindex'|'bindKeydown'|'bindRoleForClick';
+
+        /**
+         * $aria service.
+         */
+        interface IAriaService {
+            config(attribute: IAriaAttribute): boolean;
```
peterblazejewicz: So I think this is wrong, the ngAria, as per author intent, provides a configuration object (provider) to configure supported attributes:
https://gist.github.com/marcysutton/8e4f6e51d82cb36710c4#example
This provider comes with `config` method that accepts object:
https://github.com/angular/angular.js/blob/master/src/ngAria/aria.js#L97
the `$ariaService` is for internal purpose only and not exposed as public api I think:
https://docs.angularjs.org/api/ngAria/service/$aria

types/angular-aria/tslint.json
```diff
@@ -0,0 +1,6 @@
+{
+    "extends": "dtslint/dt.json",
+    "rules": {
+        "interface-name": false
```
peterblazejewicz: So the current DT style guide suggests to not put the exclude rule here in the config file, but explicitely use this (if possible) in the source code, so for example:
```ts
/**
* see {@link https://docs.angularjs.org/api/ngAria/provider/$ariaProvider}
*/
// tslint:disable-next-line interface-name
interface IAriaOptions {
```

types/angular-aria/angular-aria-tests.ts
```diff
@@ -0,0 +1,7 @@
+function testConfig($aria: angular.aria.IAriaService): void {
+    // $ExpectType boolean
+    $aria.config('tabindex');
```
peterblazejewicz: IMO this is not how that api is expected to be used:
https://docs.angularjs.org/api/ngAria/provider/$ariaProvider#overview

types/angular-aria/tslint.json
chivesrs: Yeah but all AngularJS typings break this rule. Which is why I would rather disable it package wide.

types/angular-aria/index.d.ts
chivesrs: $aria is a public API, the doc you see is broken.
See https://github.com/angular/angular.js/blob/master/src/ngAria/aria.js#L160 for docs.

types/angular-aria/angular-aria-tests.ts
chivesrs: This is not the provider.

chivesrs: Let me find you some working code on our application that uses $aria typed as any.

```
function clickDirectiveLinkFunction($aria: any): angular.IDirectiveLinkFn {
  return ($scope, $element, $attrs) => {
    const runCallback = ($event: JQueryEventObject, $isMiddleClick = false) => {
      // Evaluates "my-click" expression with $event and $isMiddleClick as
      // locals.
      $scope.$eval($attrs['myClick'], {$event, $isMiddleClick});
    };

    $element.on('click', runCallback);
    $element.on('mouseup', ($event) => {
      if ($event.which === 2) {
        // A `which` value of 2 indicates a middle click.
        // See https://api.jquery.com/event.which/ for more details.
        runCallback($event, true);
      }
    });

    if ($aria.config('bindKeydown')) {
      $element.on('keydown', ($event) => {
        if ($event.key === 'Enter') {
          runCallback($event);
        }
      });
    }

    if ($aria.config('tabindex') && $element.attr('tabindex') === undefined) {
      $element.attr('tabindex', '0');
    }
  };
}
```

I'm fairly certain that it's a public service, otherwise I'm not sure why there would be an `@ngDoc` annotation.

types/angular-aria/index.d.ts
chivesrs: I'll see if I can get the docs functional, but you can see on the main ngAria page that $aria is meant to be a public service.

Yes, this PR doesn't add typings for $ariaProvider at all. That's because my code doesn't use it, so I didn't bother adding them in. I can add them in as a second PR, just so this one as the initial creation of the module doesn't change too much from PR creation time to submission time.

chivesrs: I got the docs working. Creating PR for it.

chivesrs: See https://github.com/angular/angular.js/pull/16945. You can find the working docs for $aria in the PR as a screenshot.

sandersn:
Thanks for the review, @peterblazejewicz. I'm going to merge so the package is available, but @chivesrs you might consider adding a test case based on @peterblazejewicz 's last comment.

chivesrs:
@sandersn that's the `$ariaProvider` which I'll create a second PR shortly for (it's a different thing entirely even though the names are similar). Expect one soon tm.

PR #42143, opened 2020-02-06
[Properties-reader] Include missing methods

oborrego:
Please fill in this template.

### DIFF
--- a/types/properties-reader/index.d.ts
+++ b/types/properties-reader/index.d.ts
@@ -18,6 +18,9 @@ declare namespace PropertiesReader {
         each<T>(iterator: (this: T, key: string, value: Value) => void, scope: T): Reader;
         getAllProperties(): { [key: string]: Value; };
         clone(): Reader;
+        save(destFile: string): Promise<any>;
+        getByRoot(root: any): {};
+        bindToExpress(app: object, basePath?: string, makePaths?: boolean): Reader;
     }
 }
 
--- a/types/properties-reader/index.d.ts
+++ b/types/properties-reader/index.d.ts
@@ -18,7 +18,7 @@ declare namespace PropertiesReader {
         each<T>(iterator: (this: T, key: string, value: Value) => void, scope: T): Reader;
         getAllProperties(): { [key: string]: Value; };
         clone(): Reader;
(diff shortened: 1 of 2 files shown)


types/properties-reader/index.d.ts
```diff
@@ -18,6 +18,9 @@ declare namespace PropertiesReader {
         each<T>(iterator: (this: T, key: string, value: Value) => void, scope: T): Reader;
         getAllProperties(): { [key: string]: Value; };
         clone(): Reader;
+        save(destFile: string): Promise<any>;
```
Goldsmith42: According to the documentation and source, there's also a callback that can be provided as a second parameter.

types/properties-reader/index.d.ts
```diff
@@ -18,6 +18,9 @@ declare namespace PropertiesReader {
...
+        getByRoot(root: any): {};
```
Goldsmith42: As I understand the source, this should return `{ [key: string]: value }`.

Goldsmith42 (changes requested):

types/properties-reader/index.d.ts
oborrego: Yes, you are right. Fixed

types/properties-reader/index.d.ts
oborrego: Also fixed

types/properties-reader/index.d.ts
```diff
@@ -18,8 +18,8 @@ declare namespace PropertiesReader {
         each<T>(iterator: (this: T, key: string, value: Value) => void, scope: T): Reader;
         getAllProperties(): { [key: string]: Value; };
         clone(): Reader;
-        save(destFile: string): Promise<any>;
-        getByRoot(root: any): {};
+        save(destFile: string, onComplete?: Function ): Promise<any>;
```
Goldsmith42: Since the documentation specifies it, we can use a more specific callback type here, something like `(err: any, data: string) => void` instead of `Function`.

I can see now in the source that the returned result is actually the contents of the saved file, so we can also specify the return type as `Promise<string>`.

Goldsmith42 (approved):
Remove the whitespace after the function parameter type so that the CI build passes and you should be good to go.

PR #42158, opened 2020-02-06
feat(react-native-appsflyer): complete definition

fabianlee1211:
Please fill in this template.

If adding a new definition:

-----
Hello there,

This is a package definition for React Native AppsFlyer.

Package: https://www.npmjs.com/package/react-native-appsflyer
Source: https://github.com/AppsFlyerSDK/react-native-appsflyer

Cheers!

### DIFF
--- a/types/react-native-appsflyer/index.d.ts
+++ b/types/react-native-appsflyer/index.d.ts
@@ -1,0 +1,86 @@
+// Type definitions for react-native-appsflyer 5.1
+// Project: https://github.com/AppsFlyerSDK/react-native-appsflyer#readme
+// Definitions by: Fabian Lee <https://github.com/fabianlee1211>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.8
+
+export interface InitSdkOptions {
+    devKey: string;
+    isDebug: boolean;
+    appId: string;
+}
+
+export enum EmailCryptType {
+    EmailCryptTypeNone = 0,
+    EmailCryptTypeSHA1 = 1,
+    EmailCryptTypeMD5 = 2,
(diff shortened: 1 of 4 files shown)


types/react-native-appsflyer/index.d.ts
```diff
@@ -0,0 +1,96 @@
...
+    function setAdditionalData(additionalData: AdditionalData, success?: SuccessCallback): void;
+    function setCustomerUserId(userId: string, callback?: SuccessCallback): void;
+    function setCollectIMEI(isCollect: boolean, callback?: SuccessCallback): void;
+    function setCollectAndroidID(isCollect: boolean, callback?: SuccessCallback): void;
+    function setAppInviteOneLinkID(oneLinkId: string, callback?: SuccessCallback): void;
+    function setCurrencyCode(currencyCode: string, callback?: SuccessCallback): void;
+    function setDeviceTrackingDisabled(isDeviceTrackingDisabled: boolean, success?: SuccessCallback): void;
+}
+
+export default appsFlyer;
```
jessetrinity: I'm not familiar at all with what exporting a default namespace is supposed to do (I didn't think it was even supported). Can you help me out? Thanks!

fabianlee1211: I'll be honest that I'm not an expert in TS, I referenced other libraries here and came up with writing it this way.
The way I think of it was to follow how the library is implemented, which is just a bunch of methods attached to a JS object. 
So I was trying to do the same with namespace so that it follows how the library APIs are used i.e `appsFlyer.<method>`.
About supporting `export default namespace`, I suppose it is related to this PR.
https://github.com/microsoft/TypeScript/pull/18628

Thanks for reviewing though! I'm learning along the way.

PR #42161, opened 2020-02-06
[pouchdb-find] Add missing field to PouchDB.Find.FindResponse 

codingismy11to7:
> You may also want to pay attention to the "warning" value included in your results set, indicating that there was no index that matched the given query.

https://pouchdb.com/guides/mango-queries.html


jakubnavratil (approved):
Thanks for commit

PR #42179, opened 2020-02-07
[jwplayer] Add missing method signatures for JWPlayer.off()

soutar:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://developer.jwplayer.com/jwplayer/docs/jw8-reference#section-event-listening-with-the-jw-player-api

<hr />

This PR adds extra call signatures for the `off()` method which can be used in two ways:
1. to remove all event listeners for a given event 
2. to remove a specific event listener for a given event

The signature for removing _all_ listeners was already present, but the signatures for removing a specific listener were missing.


djcsdy (approved):
LGTM

PR #42185, opened 2020-02-07
Fix return types for filter funcs without callback

bertyhell:
according to the docs, you can omit the callback. In this case the function will return a promise containing the filtered array.

http://caolan.github.io/async/v3/docs.html#filter

Please fill in this template.


sandersn:
This change looks fine to me, but I don't like merging changes to such a widely used package without some authors signing off.

@Kern0 @Penryn @fenying @pascalmartin @Dmitri1337 @erossignon @Juliiii Can you take a look?

fenying (approved):
I think this should be ok.

PR #42229, opened 2020-02-09
update i2c-bus typings

kou029w:
  - documentation: https://github.com/fivdi/i2c-bus/blob/e4418626f76f1369cb425a5a2eefaf49954f3691/README.md#api
  - source code: https://github.com/fivdi/i2c-bus/blob/e4418626f76f1369cb425a5a2eefaf49954f3691/i2c-bus.js

### DIFF
--- a/types/i2c-bus/index.d.ts
+++ b/types/i2c-bus/index.d.ts
@@ -1,11 +1,17 @@
-// Type definitions for i2c-bus v1.0.0
+// Type definitions for i2c-bus v5.1.0
 // Project: https://github.com/fivdi/i2c-bus
 // Definitions by: Jason Heard <https://github.com/101100>
+//                 Kohei Watanabe <https://github.com/kou029w>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+// TypeScript Version: 3.7
 
 /// <reference types="node" />
 
+type I2CDeviceId = { manufacturer: number; product: number; name: string };
+
+type BytesWritten = { bytesWritten: number; buffer: Buffer };
+
+type BytesRead = { bytesRead: number; buffer: Buffer };
 
 type CompletionCallback = (error: any) => any;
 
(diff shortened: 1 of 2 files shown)


types/i2c-bus/index.d.ts
```diff
@@ -33,8 +44,7 @@ export interface I2cBusFuncs {
     smbusWriteI2cBlock: boolean;
 }
 
-export interface I2cBus {
-
+export interface Bus {
```
101100: I think it would be nice to keep the old name for those updating the typings to the latest version.  Also, if people are using other hardware, there might be a collision of type names if it is left so general.

101100 (commented):
Thanks for updating these with all of the changes!  I have one recommendation/discussion point.

types/i2c-bus/index.d.ts
kou029w: Thanks for reviewing this PR. Fixed in c0e7903.

101100 (approved):
Thanks for the changes.  Looks good.  :cookie:

types/i2c-bus/index.d.ts
```diff
@@ -1,19 +1,30 @@
-// Type definitions for i2c-bus v1.0.0
+// Type definitions for i2c-bus v5.1.0
 // Project: https://github.com/fivdi/i2c-bus
 // Definitions by: Jason Heard <https://github.com/101100>
+//                 Kohei Watanabe <https://github.com/kou029w>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
-// TypeScript Version: 2.3
+// TypeScript Version: 3.7
```
sandersn: This line specifies typescript 3.7 as the *minimum* version. I think these types will actually work with 2.8+, which is what DT supports right now. In a follow-up PR, you might try removing this line entirely and seeing whether tests still pass.

sandersn (commented):
I'll publish with the current changes, but consider a follow-up PR.

PR #42230, opened 2020-02-09
fix(intercom-client): fix find methods

sspatari:
Change return type find method that returns promise and find method with callback from User to UserLis


johnnyreilly (approved):

ErikMartensson:
@sspatari In what scenario do you get a list back instead of just one user? When I use this function I only get one user back, not a list.

I use it this way:
```javascript
client.users.find({ user_id: 'my-id' });
```

PR #42242, opened 2020-02-10
feat(markdown-magic): type definitions for version 1.0

peterblazejewicz:
- types definition
- tests

Thanks!

https://github.com/DavidWells/markdown-magic#api

- [x ] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] `tslint.json` should be present and it shouldn't have any additional or disabling of rules. Just content as `{ "extends": "dtslint/dt.json" }`.

### DIFF
--- a/types/markdown-magic/index.d.ts
+++ b/types/markdown-magic/index.d.ts
@@ -1,0 +1,59 @@
+// Type definitions for markdown-magic 1.0
+// Project: https://github.com/DavidWells/markdown-magic#readme
+// Definitions by: Piotr Błażejewicz (Peter Blazejewicz) <https://github.com/peterblazejewicz>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// Minimum TypeScript Version: 3.4
+
+/**
+ * @param path Path or glob pattern
+ * @param config See configuration options below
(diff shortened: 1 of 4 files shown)


types/markdown-magic/index.d.ts
```diff
@@ -0,0 +1,59 @@
+// Type definitions for markdown-magic 1.0
+// Project: https://github.com/DavidWells/markdown-magic#readme
+// Definitions by: Piotr Błażejewicz (Peter Blazejewicz) <https://github.com/peterblazejewicz>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// Minimum TypeScript Version: 3.4
```
sandersn: This code doesn't look like it uses complicated types, so it probably doesn't require 3.4.

sandersn (approved):
I'm going to merge this so it's available right away, but if you can switch `readonly string[]` to `ReadonlyArray<string>` in a followup PR, it will be usable by a few more people who are stuck on old versions of TS.

types/markdown-magic/index.d.ts
peterblazejewicz: Yup, read. Will do this asap, thanks!

PR #42249, opened 2020-02-10
twig: Added overload to Template.render when allow_async=false

dmelcer9:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/twigjs/twig.js/blob/51544ce66e355cf7b142d7f58d52f71088548bad/ASYNC.md#switching-render-mode

### DIFF
--- a/types/twig/index.d.ts
+++ b/types/twig/index.d.ts
@@ -3,6 +3,7 @@
 // Definitions by: Carlos Ballesteros Velasco <https://github.com/soywiz>
 //                 Tim Schumacher <https://github.com/enko>
 //                 Maik Tizziani <https://github.com/mtizziani>
+//                 Daniel Melcer <https://github.com/dmelcer9>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
 // Imported from: https://github.com/soywiz/typescript-node-definitions/twig.d.ts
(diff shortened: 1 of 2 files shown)


types/twig/index.d.ts
```diff
@@ -24,6 +25,7 @@ export interface Parameters {
 
 export interface Template {
     reset(blocks: any): void;
+    render(context?: any, params?: any, allow_async?: false): string;
     render(context?: any, params?: any, allow_async?: boolean): string | Promise<string>;
```
enko: @dmelcer9 wouldn't it be better to explicit define `false` and `true`? `false` returns `string` and `true` returns `Promise<string>` and `undefined` whatever the standard is.

enko (changes requested):
Please have a look at my comment.

types/twig/index.d.ts
dmelcer9: Does true always return a promise? The name of the allow_async parameter is somewhat ambiguous about this. In any case, we still need the general Boolean overload for when the parameter is unknown at compile time.

dmelcer9: Bumping so @typescript-bot doesn't automatically close.

enko: @dmelcer9 Sorry did have time to look into this and @typescript-bot seems to be very strict. Maybe we merge as it is. I looked at the code and it wasn't very clear to me, how the default is computed.

enko:
@typescript-bot lgtm

dmelcer9:
I created https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42593 since there doesn't seem to be a way to get typescript-bot to open this PR.

PR #42254, opened 2020-02-10
Leaflet: add _getIconUrl to Default class in Icon namespace

geopic:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/Leaflet/Leaflet/issues/6991 (Leaflet issue providing context)

### DIFF
--- a/types/leaflet/index.d.ts
+++ b/types/leaflet/index.d.ts
@@ -16,7 +16,7 @@ export as namespace L;
 import * as geojson from 'geojson';
 
 export class Class {
-    static extend(props: any): { new(...args: any[]): any } & typeof Class;
+    static extend(props: any): {new(...args: any[]): any} & typeof Class;
     static include(props: any): any & typeof Class;
     static mergeOptions(props: any): any & typeof Class;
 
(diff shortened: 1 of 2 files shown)


henrythasler (changes requested):
I suggest to 

- add a link to the leaflet source code where this property is defined
- add a test that uses this property

geopic:
Sorted, @henrythasler.

> * add a link to the leaflet source code where this property is defined

https://github.com/Leaflet/Leaflet/blob/master/src/layer/marker/Icon.Default.js#L33-L43

henrythasler:
I'm not totally comfortable with this change. Can anyone else state their opinion please?

types/leaflet/index.d.ts
```diff
@@ -1840,6 +1841,10 @@ export namespace Icon {
 
     class Default extends Icon<DefaultIconOptions> {
         static imagePath?: string;
+
+        // Intended to be a private property but is required to be exposed for this bugfix by vue2-leaflet to work: https://vue2-leaflet.netlify.com/quickstart/#marker-icons-are-missing
+        _getIconUrl: (name: string) => string;
```
ronikar: It should be local because it's relevant only for `Vue Leaflet`

ronikar (changes requested):

atd-schubert:
I would not add a not public part of the API to the type definition.

types/leaflet/index.d.ts
geopic: I'll raise this with vue2-leaflet.

geopic:
@atd-schubert In retrospect you're right, I'll close the pull request and will get this resolved with the `vue2-leaflet` library authors instead.

types/leaflet/index.d.ts
arjun-mavonic: It's not just relevant to just Vue Leaflet. issue is with @types/leaflet. Basically the types definition is missing this type `_getIconUrl`.
 
This was a valid change and should have merged instead of closing.

Issue #12395, opened 2016-11-01
How to create a type definition for a CommonJS module that exports a function?


mohsen1:
I'm trying to contribute a type definition for a module that exports like this:

```js
  module.exports = function (store) {
    return store;
  };
```

But using `export =` or `export default` gave me all sort of errors. 

What is the best of exporting for modules like this today?

vvakame:
```
function foo(store:any): void;
export = foo;
```
try this form.

mohsen1:
@vvakame without declaring any module? 

let's say that file that is exporting function is at `node_modules/foo/index.js`

and I'm putting my `.d.ts` file in `node_modules/@types/foo/index.d.ts`

If content of `index.d.ts` is the following, it won't work:

```ts
dclare module 'foo' {
function foo(store:any): void;
export = foo;
}
```

When I import `"foo"`, I get following error: 

```ts
import fn from 'foo';
```
```
 error TS2497: Module ''foo'' resolves to a non-module entity and cannot be imported using this construct.
```

vvakame:
```
function foo(store:any): void;
declare namespace foo {}
export = foo;
```

```
import * as fn from 'foo';
```

http://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html

mohsen1:
That's not working either. It's the same problem as https://github.com/DefinitelyTyped/DefinitelyTyped/issues/11720

if a npm module is exporting a function, there is that problem...

mohsen1:
@mhegazy what you recommend?

andy-ms:
The error message is correct: `"foo"` is a non-module entity -- a function. It should be imported as `import foo = require("foo")`. And @vvakame is correct that you shouldn't need the `declare module "foo"` wrapper, since the location of the types (`node_modules/@types/foo/index.d.ts`) already determines the module being declared.

orta:
Hi thread, we're moving DefinitelyTyped to use [GitHub Discussions](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/53377) for conversations the `@types` modules in DefinitelyTyped.

To help with the transition, we're closing all issues which haven't had activity in the last 6 months, which includes this issue. If you think closing this issue is a mistake, please pop into the [TypeScript Community Discord](https://discord.gg/typescript) and mention the issue in the `definitely-typed` channel.

PR #42261, opened 2020-02-11
fix(defaults): Limit function definition to module scope

carmanchris31:
Please fill in this template.

---

This fixes an issue where the `defaults` function is declared globally even though it's only available when imported as a module.

Before:
Types and no errors when not imported
defaults-scope-before

After:
Error when module not imported (as expected)
defaults-scope-after-1

Types available when module is imported (as expected)
defaults-scope-after-2

Note: If this is a common issue, perhaps we could consider adding a note about it to the "Common mistakes" documentation.

### DIFF
--- a/types/defaults/index.d.ts
+++ b/types/defaults/index.d.ts
@@ -3,8 +3,8 @@
 // Definitions by: Ibtihel CHNAB <https://github.com/IbtihelCHNAB>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
-declare function defaults(options: any, defaultOptions: any): any;
-
 declare module "defaults" {
+  function defaults(options: any, defaultOptions: any): any;
+
   export = defaults;
 }


peterblazejewicz (commented):
@carmanchris31 
I believe it should read:
```ts
// Type definitions for defaults 1.0.3
// Project: https://github.com/tmpvar/defaults/
// Definitions by: Ibtihel CHNAB <https://github.com/IbtihelCHNAB>

declare function defaults(options: any, defaultOptions: any): any;

export = defaults;
```
That would correct the CommonJS module imports, remove not needed module definition (file itself is an module), see similar:
https://github.com/DefinitelyTyped/DefinitelyTyped/issues/12395#issuecomment-257572414
https://github.com/DefinitelyTyped/DefinitelyTyped#should-i-add-an-empty-namespace-to-a-package-that-doesnt-export-a-module-to-use-es6-style-imports
https://github.com/DefinitelyTyped/DefinitelyTyped#a-package-uses-export--but-i-prefer-to-use-default-imports-can-i-change-export--to-export-default

carmanchris31:
Hey @peterblazejewicz that looks promising. I’ll give it a look and update.

What are your thoughts on having an extra test for the scope? Is there any precedent for this on other packages?

peterblazejewicz:
@carmanchris31 
This one is a CommonJS module with single export:
https://github.com/tmpvar/defaults/blob/master/index.js#L3
This extra tests IMO can be usefull when the module under test supports multiple module loaders, like UMD, e.g. see example in DT project here:
https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/ejs/test
So that one should just use CommonJS TS syntax I think, thanks!

PranavSenthilnathan:

PR #42264, opened 2020-02-11
chrome: goForward and goBack methods & properties

ozomer:
Api is documented here:

* https://developer.chrome.com/extensions/tabs#method-goForward
* https://developer.chrome.com/extensions/tabs#method-goBack

### DIFF
--- a/types/chrome/index.d.ts
+++ b/types/chrome/index.d.ts
@@ -7672,6 +7672,32 @@ declare namespace chrome.tabs {
      * @param callback Called after the operation is completed.
      */
     export function discard(tabId?: number, callback?: (tab: Tab) => void): void;
+    /**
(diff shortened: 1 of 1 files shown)


fredrikhr (changes requested):
Don't the optional `tabId` parameters mean that there are overloads to both functions that do not take the `tabId` parameter? (Even though Google doesn't document these overloads, that's how they work, right?).

i.e.
``` ts
export function goForward(callback?: () => void): void;
export function goBack(callback?: () => void): void;
```
(with their repective comments and maybe a `@see` link.

ozomer:
@couven92 You are correct. Fixed the commit.

fredrikhr (approved):
:+1:

PR #42271, opened 2020-02-11
feat: types for @rdfjs/to-ntriples

tpluscode:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/rdfjs__to-ntriples/OTHER_FILES.txt
+++ b/types/rdfjs__to-ntriples/OTHER_FILES.txt
@@ -1,0 +1,5 @@
+lib/blankNode.d.ts
+lib/defaultGraph.d.ts
+lib/literal.d.ts
+lib/namedNode.d.ts
+lib/variable.d.ts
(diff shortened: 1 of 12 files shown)


types/rdfjs__to-ntriples/OTHER_FILES.txt
```diff
@@ -0,0 +1,5 @@
+lib/blankNode.d.ts
+lib/defaultGraph.d.ts
+lib/literal.d.ts
+lib/namedNode.d.ts
+lib/variable.d.ts
```
tpluscode: Those files are imported in the tests. I think it should be unnecessary to list them here but the build failed otherwise...

sandersn: weird! I think there might be a bug in the dependency-following code. If I had to guess, it's because the code doesn't handle deep references into scoped packages correctly. I opened https://github.com/microsoft/types-publisher/issues/737 to track it.

sandersn (approved):
Looks good. I'll merge this so you can use it right away, despite the workaround where you had to put a bunch of source files in OTHER_FILES.txt.

PR #42282, opened 2020-02-11
feat(react): exclude number from keyof JSX.IntrinsicElements

lttb:
It's required to exclude `number` type from `keyof JSX.IntrinsicElements` because of the case when `JSX.IntrinsicElements` has some additional declarations with indexer type, for example

```
declare namespace JSX {
    interface IntrinsicElements extends JSX.IntrinsicElements {
        [name: string]: SomeJSXElementType;
    }
}
```

Thank you!

Please fill in this template.

- [X] Provide a URL to documentation or source code which provides context for the suggested changes: the context is in the Pull Request description

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -115,7 +115,7 @@ declare namespace React {
     interface ReactComponentElement<
         T extends keyof JSX.IntrinsicElements | JSXElementConstructor<any>,
         P = Pick<ComponentProps<T>, Exclude<keyof ComponentProps<T>, 'key' | 'ref'>>
-    > extends ReactElement<P, T> { }
+    > extends ReactElement<P, Exclude<T, number>> { }
 
     /**
      * @deprecated Please use `FunctionComponentElement`


ferdaber:
React definitions is already very heavily used across the board, and any changes (especially when it comes to adding additional layers of a conditional type) can heavily impact performance. I would err on the side of fixing whatever is merging that extra declaration, rather than try to protect against it from the React definitions side.

eps1lon:
> I would err on the side of fixing whatever is merging that extra declaration, rather than try to protect against it from the React definitions side.

My point as well. Without knowing who does this and why we have quite a slippery slope here: Why shouldn't we exclude other types as well?

It would help a lot if you could show a concrete example (preferably using http://www.typescriptlang.org/play/) that illustrates the issue.

lttb:
Thank you for the response!

Please forgive me in advance for so long read, but I think this issue requires a bit more context 🙂 

## intro

I'm working on the cssinjs library with static styles extraction called reshadow (https://github.com/lttb/reshadow), and we're working around some semantic level over JSX and styles.

Some abstract basic example:

```js
import styled from 'reshadow'

const Button = ({children, color, disabled, icon}) => styled`
    button {
        color: ${color};
    }

    button[disabled] {
        opacity: .5;
    }

    content {
        font-size: 14px;
    }
`(
    <button disabled={disabled}>
        <content>{children}</content>
        <icon as="span">{icon}</icon>
    </button>
)
```

As you might notice, there are `content` and `icon` elements, that have markup semantics on the context of the component, but there are no such elements in the DOM. By default, such elements fallback to the `div`, but with `as` attribute it is possible to set the DOM presentation (i.e. `as="span"` goes to `span` in render)

BTW, there is a bit more information and examples about the concept in the docs: https://reshadow.dev/concepts#elements

In fact, it is an abstraction over `className` in an attempt to achieve more simple and declarative markup, especially with complex components with different styling states.
For example, the same code with `css-modules`:

```js
import cn from 'classnames'
import styles from './styles.module.css'

// ...
(
    <button disabled={disabled} className={cn(styles.button, disabled && styles.disabled)}>
        <div className={styles.content}>{children}</div>
        <span className={styles.icon}>{icon}</span>
    </button>
)
```

Of course, it is possible to write something, like

```js
const Content = 'div'
const Icon = 'span'

const Button = ({children, color, disabled, icon}) => styled`
    button {
        color: ${color};
    }

    button[disabled] {
        opacity: .5;
    }

    Content {
        font-size: 14px;
    }
`(
    <button disabled={disabled}>
        <Content>{children}</Content>
        <Icon>{icon}</Icon>
    </button>
)
```

and in some cases, it would be really handy, but in a common way, in my experience, it was a bit worse Developer Experience, especially for newcomers

## typescript

So, to provide a `typescript` support for this library,  it's required to have an ability to use typed custom semantic elements in the JSX.

In the case of `reshadow` there is no way to know which custom element names will be used and to support them, we've added such declaration:

*reshadow/elements.d.ts*
```js
type CustomElement<T extends keyof JSX.IntrinsicElements> = T extends any
    ? ({as: T} & JSX.IntrinsicElements[T]) | JSX.IntrinsicElements['div']
    : never;

declare namespace JSX {
    interface IntrinsicElements extends JSX.IntrinsicElements {
        [name: string]: CustomElement<keyof JSX.IntrinsicElements>;
    }
}
```

Unfortunately, I can't add an example by http://www.typescriptlang.org/play/, because it looks like it doesn't merge namespace declaration, but on the codesandbox it works as expected (autocomplete and so on works as well):
- [codesandbox example](https://codesandbox.io/s/reshadow-next-typescript-se741?fontsize=14&hidenavigation=1&theme=dark), `typescript 3.7.5`, `@types/react 16.9.19` 
 - in `next-env.d.ts` there is a reference to `node_modules/reshadow/elements.d.ts`, which has an extension above for the `JSX`
- the usage example in `pages/index.tsx`

image

And everything should be okay while an option `"skipLibCheck": true` used.

However, there is a problem with `"skipLibCheck": false` (which is a bit rare, but valid), because of the way how `keyof SomeType` works (https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types-and-index-signatures).

If there is an `indexer` property in the type, like `{[key: string]: any}`, `keyof` will return `string | number`.

Thus, `keyof JSX.IntrinsicElements` with declaration above also returns `string | number`, which causes a conflict with `T` used in the `ReactElement` type:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/28ff2e3a70ee7172902b2efb0f0c527e8ada0430/types/react/index.d.ts#L109-L118

## summary

I've been trying hack around to solve that issue on the library definitions side, and of course, the change in React definitions is the last thing  😄but it looks like there is no way to exclude the `number` type from `keyof` of the type if the indexer property was used.

And, definitely, this issue is kind of very specific use case.
Customizations, and even more so indexer property in `JSX.IntrinsicElements` isn't that common of course.

But, at the same time, it would be super great to consider such fix, if there are no any tradeoffs like performance, maintainability and so on

For example, it seems that something like this of should not affect the performance there because of its static evaluation?

```js
// exclude the `number` type from keys in case of indexer property
type IntrinsicElementsKeys = Exclude<keyof JSX.IntrinsicElements, number>

interface ReactComponentElement<
    T extends IntrinsicElementsKeys | JSXElementConstructor<any>,
    P = Pick<ComponentProps<T>, Exclude<keyof ComponentProps<T>, 'key' | 'ref'>>
> extends ReactElement<P, T> { }
```

And maybe change `keyof JSX.IntrinsicElements` to `IntrinsicElementsKeys` in other cases too.

Anyway, thank you very much for your attention

eps1lon:
> keyof T will be string | number (and not just string, since in JavaScript you can access an object property either by using strings (object['42']) or numbers (object[42]))

-- https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types-and-index-signatures

Oh no why do they still allow weak typing in that spot :laughing: 

Would a union of keys work e.g. `'content' | 'icon'`? 

Either way we can't accept this because we don't know the performance implications. You would need to push empty commits until the benchmark task is able to be complete. I don't think the current failure is deterministic.

Personally I wouldn't use this library in the original form. I find `const Content = 'div';` way more readable (compiler agrees with me here) i.e. who defines that one becomes a `div` the other a `span`?

lttb:
> Oh no why do they still allow weak typing in that spot 😆

Feeling the same 😄 

> Would a union of keys work e.g. 'content' | 'icon'?

Yeah, not just `[key: 'content' | 'icon']: Type`, but with mapped object type ([playground](https://www.typescriptlang.org/play/?jsx=2&ssl=1&ssc=1&pln=24&pc=1#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wChSYBPMJOAYQFcBnGXAUQBskQkA7GADwAVOEgAeMPgBMmcANZJKETHABSAZQAaAOgCS-KMF5NgaTtz4wmAPjgBeOCPGTeMuCl6VScH3AD8cAAUAN4oTABcjgC+cABkalp6BkYmZlw8-EwA2kIAugCUcAA+CTr6MIbGpuYZVln4UsAAbvi53r6RvEhNSFAA3ORUNHA1lrIOwe0+WQqUcEYEaBD8lvjFBKbLayX4S-woRr2tkYws7OmWArPKpUkVKdUXmdakUeRGklCY6LTllamjTKiCTSWQaMrJKppCyZAA0IyeVjgwTgb1IAHp0XAAO7QOSybFIDgcd4rL4-RDEGCAwQABXs7k88KcINcshYlQA5gyOUZObYUVMfEMkJEhH04EK4GAcGAInBaQNUZKfN4Pr1vmhaMh0DA6LhIF1+DSBCrfI5gS43NcVH8HtDakx4VL6Q4PHMhbZnKDKbqTbTmQLlaQgA)):

```js

type CustomElement<T extends keyof JSX.IntrinsicElements> = T extends any
    ? ({as: T} & JSX.IntrinsicElements[T]) | JSX.IntrinsicElements['div']
    : never;

type CustomElements = {
    [key in 'content' | 'icon' | 'container']: CustomElement<keyof JSX.IntrinsicElements>
}

interface IntrinsicElements extends JSX.IntrinsicElements, CustomElements { }
``` 

But it requires developers to predefine the elements, which is also might be useful, but increases the barrier to entry

> i.e. who defines that one becomes a div the other a span

That's an option for the babel plugin, so by default all custom elements (that not defined in DOM) fallback to the `div`, unless the `as` attribute provided

developer might choose another element for the fallback, we've just found `div` the most common tag for such cases

ferdaber:
I see, in that case I would actually modify the change to be:
```ts
    interface ReactComponentElement<
        T extends keyof JSX.IntrinsicElements | JSXElementConstructor<any>,
        P = Pick<ComponentProps<T>, Exclude<keyof ComponentProps<T>, 'key' | 'ref'>>
    > extends ReactElement<P, Exclude<T, number>> { }
```

sheetalkamat (changes requested):
Please look into feedback

lttb:
@ferdaber thank you very much for the feedback, and sorry please for the late response.

I've checked the suggested change, and everything seems fine. Hope benchmark task will be completed too

types/react/index.d.ts
```diff
@@ -115,7 +115,7 @@ declare namespace React {
     interface ReactComponentElement<
         T extends keyof JSX.IntrinsicElements | JSXElementConstructor<any>,
         P = Pick<ComponentProps<T>, Exclude<keyof ComponentProps<T>, 'key' | 'ref'>>
-    > extends ReactElement<P, T> { }
+    > extends ReactElement<P, Exclude<T, number>> { }
```
lttb: should I add any comments about the indexer type here, or the commit description is enough?

PR #42291, opened 2020-02-12
Update for evernote v2.0.5

fcastilloec:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://dev.evernote.com/doc/reference/ and https://github.com/evernote/evernote-sdk-js

### DIFF
--- a/types/evernote/index.d.ts
+++ b/types/evernote/index.d.ts
@@ -1,6319 +1,1855 @@
-// Type definitions for evernote v 1.25.8
+// Type definitions for evernote v2.0.5
 // Project: https://www.npmjs.com/package/evernote
 // Definitions by: Zachary Collins <https://github.com/corps>
+//                 Felipe Castillo <https://github.com/fcastilloec>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+//              https://dev.evernote.com/doc/reference/
 // TypeScript Version: 2.3
 
-import { Thrift } from 'thrift';
-
-declare namespace Evernote {
-    interface Callback<T> {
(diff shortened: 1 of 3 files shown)


PranavSenthilnathan:
@corps Could you take a look since it looks like almost the whole definition is rewritten.

types/evernote/index.d.ts
```diff
@@ -1,6319 +1,1855 @@
-// Type definitions for evernote v 1.25.8
+// Type definitions for evernote v2.0.5
```
PranavSenthilnathan: Could you enable the `dt-header` in the `tslint.json` file? The version should only be major and minor.

fcastilloec: Sure, I'll enable it. I just didn't even check `tslint.json`. I just notice than a lot of rules are turned off.
When running `dts-gen`, it creates a `tslint.json` with only:

```json
{ "extends": "dtslint/dt.json" }
```
Should I just leave `tslint.json` as above (deleting all those disabled rules)? And, of course, I'll fix any that can be fixed.

types/evernote/index.d.ts
```diff
@@ -1,6319 +1,1855 @@
...
-         *     <li> "NoteAttributes.applicationData.key" - note not found, by key</li>
-         * </ul>
-         */
-        getNoteApplicationDataEntry(guid: string, key: string, cb: Callback<string>): void;
-
-        /**
-         * Update, or create, an entry in the applicationData map for
-         * the note identified by guid.
-         */
-        setNoteApplicationDataEntry(guid: string, key: string, value: string, c
```
sheetalkamat: You dont need module definition here

sheetalkamat (changes requested):

PR #42303, opened 2020-02-12
[node-notifier]: add `timeout: false` option

SimenB:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/mikaelbr/node-notifier/pull/271

### DIFF
--- a/types/node-notifier/index.d.ts
+++ b/types/node-notifier/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for node-notifier 5.4.0
+// Type definitions for node-notifier 6.0
 // Project: https://github.com/mikaelbr/node-notifier
 // Definitions by: Qubo <https://github.com/tkQubo>
 //                 Lorenzo Rapetti <https://github.com/loryman>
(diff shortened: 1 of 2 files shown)


types/node-notifier/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for node-notifier 5.4.0
+// Type definitions for node-notifier 5.4.1
```
PranavSenthilnathan: This should be only major and minor version (one of the lint rules catches this but currently they're all disabled in the tslint.json file in this folder). So you can either leave it as it is or change it to 5.4 or 6.0 (the latest package version).

PranavSenthilnathan (changes requested):

types/node-notifier/index.d.ts
SimenB: Could you fix it? I'm away from the computer for a few days

SimenB:
updated 👍

PR #42316, opened 2020-02-12
Fix new errors from TS@next (3.9)

sandersn:
1. Bogus trailing } and > are now detected in JSX.
2. Properties that are overwritten in JSX and object spreads are now detected.

For react-motion, note that the fix correctly started to use a previously unused attribute `style`, which exposes a type error in the tests:

```
config.style: number | OpaqueConfig
```

but the test assigns it to

```
div.style: string | number
```

I squelched the error with a cast to `number`, but probably either the
test or the types are actually wrong.


johnnyreilly (commented):
I approve this change 🤗

PR #42318, opened 2020-02-12
mixpanel-browser: Add transport support

cedrics:
Adds request options to `mixpanel-browser`'s types. See mixpanel's [documentation](https://developer.mixpanel.com/docs/javascript-full-api-reference#section-mixpanel-track) for details.
Even though this is was released in a minor version, this is actually a breaking change


peterblazejewicz (approved):
Introduced here (2.33):
https://github.com/mixpanel/mixpanel-js/commit/6537ae19712c9077c671e04239a42be3a2179c77
LGTM!
@cedrics thanks!

PR #42331, opened 2020-02-13
fix(sharedb): correct sharedb.DB.getOps so it has nullable arguments

pxpeterxu:

  - https://github.com/share/sharedb#getops
  - Usage in tests: https://github.com/share/sharedb/blob/8de47b6dc2f4ff2c2c2cbaac08665593cebaa386/test/db.js#L76

**Details**

sharedb.DB.getOps was typed as taking `from` and `to` as always being `numbers`, when in fact, either of them could be `null`:

- If `from` is null, it means we should get ops from the beginning of time
- If `to` is null, it means we should get ops up to the latest

This is documented in [docs](https://github.com/share/sharedb#getops) and in [tests](https://github.com/share/sharedb/blob/8de47b6dc2f4ff2c2c2cbaac08665593cebaa386/test/db.js#L76)

I've edited `getOps` and `getOpsToSnapshot`, and added a test for `getOps`

### DIFF
--- a/types/sharedb/index.d.ts
+++ b/types/sharedb/index.d.ts
@@ -2,6 +2,7 @@
 // Project: https://github.com/share/sharedb
 // Definitions by: Steve Oney <https://github.com/soney>
 //                 Eric Hwang <https://github.com/ericyhwang>
+//                 Peter Xu <https://github.com/pxpeterxu>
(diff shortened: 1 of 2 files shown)


PranavSenthilnathan:

PR #42333, opened 2020-02-13
Add typings for `json-ptr@1.2`

treybrisbane:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/json-ptr/json-ptr-tests.ts
+++ b/types/json-ptr/json-ptr-tests.ts
@@ -1,0 +1,52 @@
+import JsonPointer = require('json-ptr');
+
+// JsonPointer static
+const JsonPointer1: typeof JsonPointer = JsonPointer.JsonPointer;
+const JsonReference1: typeof JsonPointer.JsonReference = JsonPointer.JsonReference;
+const create1: JsonPointer = JsonPointer.create('');
+const staticHas1: boolean = JsonPointer.has({}, '');
+const staticGet1: unknown = JsonPointer.get({}, '');
+const staticSet1: unknown = JsonPointer.set({}, '', undefined);
+const staticSet2: unknown = JsonPointer.set({}, '', undefined, false);
+const staticSet3: unknown = JsonPointer.set({}, '', undefined, true);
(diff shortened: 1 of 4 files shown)


treybrisbane:
I'm assuming I don't need to wrap these in a `declare module 'json-ptr' { ... }`, right? First time submitting. 🙂

types/json-ptr/tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
+    "compilerOptions": {
+        "baseUrl": "../",
+        "esModuleInterop": true,
```
sandersn: this needs to be false to test backward compatibility

types/json-ptr/json-ptr-tests.ts
```diff
@@ -0,0 +1,52 @@
+import JsonPointer from 'json-ptr';
```
sandersn: Use the commonjs syntax `import JsonPointer = require('json-ptr')`.

Consumers of the package who want to use `import JsonProinter from 'json-ptr'` will need to add `"esModuleInterop": true`, but Definitely Typed tests need to be backward compatible to show that the types will work with old code bases.

sandersn (changes requested):

types/json-ptr/tsconfig.json
treybrisbane: Done. 🙂

types/json-ptr/json-ptr-tests.ts
treybrisbane: Done. 🙂

treybrisbane:
I've just been informed that the author of `json-ptr` is himself working on a TypeScript reimplementation, which would render these typings redundant.

I'm currently seeking clarification on the ETA of that work. Depending on when it's likely to land, it may be worth merging this anyway during the interim, but it may also be pointless. Can we delay merging this temporarily until I know more? 🙂

sandersn:
Sure, just ping me when you want me to sign off so that the bot will know it's ready to go.

treybrisbane:
Closing this due to the reasons mentioned above. :)

PR #42340, opened 2020-02-13
[angular-mocks] Fix error when using typescript 3.7.5

hwildwood:
Importing these types into a typescript 3.7.5 project resulted in an exception from the typescript compiler.

```
.../node_modules/typescript/lib/tsc.js:78602
                throw e;
                ^

TypeError: Cannot read property 'get' of undefined
```

My hunch is that it was due to a clash between the `mock` namespace and the `mock` property on the IAngularStatic interface.  The IAngularStatic interface is exported from the angular types module as angular so you end up with a clash between `angular.mock` referring to the namespace in the angular-mocks index.d.ts and also referring to the IMockStatic interface. Splitting the angular-mocks `mock` namespace to `http` and `browser` for their respective interfaces fixes the problem for us.

Please fill in this template.

### DIFF
--- a/types/angular-mocks/angular-mocks-tests.ts
+++ b/types/angular-mocks/angular-mocks-tests.ts
@@ -134,7 +134,7 @@ $componentController<{}, { test: { name: string } }>('Test controller', {
 // IHttpBackendService
 ///////////////////////////////////////
 let httpBackendService: ng.IHttpBackendService;
-let requestHandler: ng.mock.IRequestHandler;
+let requestHandler: ng.http.IRequestHandler;
 
 httpBackendService.flush();
 httpBackendService.flush(1234);
--- a/types/angular-mocks/index.d.ts
+++ b/types/angular-mocks/index.d.ts
@@ -206,9 +206,9 @@ declare module 'angular' {
       method: string,
       url: string | RegExp | ((url: string) => boolean),
       data?: string | RegExp | object | ((data: string) => boolean),
-      headers?: mock.IHttpHeaders | ((headers: mock.IHttpHeaders) => boolean),
+      headers?: http.IHttpHeaders | ((headers: http.IHttpHeaders) => boolean),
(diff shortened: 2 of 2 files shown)


thorn0:
If this is a bug in TS, which it apparently is, it should be reported. Did you try to isolate the issue to create a minimal reproduce case?

HerbiePorter:
That's a good point, thanks @thorn0.  I'm relatively new to typescript, I had assumed that the angular-mocks type module was doing something wrong (clashing a namespace with a static property) but the error is being thrown from the typescript compiler.  I'll try and isolate it this weekend in a test case, raise a bug with typescript and link it to this PR so we can track them both.

thorn0:
> was doing something wrong (clashing a namespace with a static property

Nothing wrong with this. TS supports [declaration merging](https://www.typescriptlang.org/docs/handbook/declaration-merging.html).

thorn0 (changes requested):
https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42340#issuecomment-585983834

HerbiePorter:
@thorn0 I've tried to reproduce the exception in a standalone project but I couldn't get it to work.  I'm happy to close this PR as it sounds like it might be something related to our project setup rather than the angular-mocks types or typescript compiler.

thorn0:
Most probably it's something in your project setup that revealed a bug in the compiler.

When trying to reproduce, did you start from scratch or were you removing / simplifying parts of your project to understand at what point this error stops happening?

HerbiePorter:
Our project is relatively large and complex (due to a migration from angular to react) so it would be difficult to work backwards from there.  Instead, I tried to start from scratch with a simple hello-world angular webpack project with typescript 3.7.5.  With that simple setup, the compiler didn't throw any warnings.

thorn0:
Yeah, I tried that too.

PR #42342, opened 2020-02-13
Change koa-ejs ctx.render 'properties' values type to 'any' from 'string'

scoobster17:
Change koa-ejs ctx.render properties type to 'any' to allow for nested objects and other types than 'string'.

v4.2.1 introduced the ctx.render types, but incorrectly sets the second argument to type { [key: string]: string }. Whilst the keys should be strings, the values can be booleans, numbers, nested objects and more.
____________________________________________________

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/koajs/ejs/blob/master/example/app.js (you can see on L36-38 users property is an array of objects).

### DIFF
--- a/types/koa-ejs/index.d.ts
+++ b/types/koa-ejs/index.d.ts
@@ -8,7 +8,8 @@ import * as Koa from "koa";
 
 declare module "Koa" {
   interface ExtendableContext {
-    render: (template: string, properties?: {[name: string]: string}) => Promise<string>;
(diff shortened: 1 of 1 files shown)


scoobster17:
Problem introduced in v.4.2.1 published as a part of this pull request https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42008

PranavSenthilnathan:

PR #42354, opened 2020-02-13
[react-table] v7: fix getRowId type signature

ggascoigne:
The signature for `getRowId` was incorrect, this fixes it. See https://github.com/tannerlinsley/react-table/blob/master/docs/api/useTable.md for details.

### DIFF
--- a/types/react-table/index.d.ts
+++ b/types/react-table/index.d.ts
@@ -112,7 +112,7 @@ export type UseTableOptions<D extends object> = {
     defaultColumn: Partial<Column<D>>;
     initialRowStateKey: IdType<D>;
     getSubRows: (originalRow: D, relativeIndex: number) => D[];
-    getRowId: (originalRow: D, relativeIndex: number) => IdType<D>;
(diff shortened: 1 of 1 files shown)


types/react-table/index.d.ts
```diff
@@ -112,7 +112,7 @@ export type UseTableOptions<D extends object> = {
     defaultColumn: Partial<Column<D>>;
     initialRowStateKey: IdType<D>;
     getSubRows: (originalRow: D, relativeIndex: number) => D[];
-    getRowId: (originalRow: D, relativeIndex: number) => IdType<D>;
+    getRowId: (originalRow: D, relativeIndex: number, parent?: Row<D>) => string;
```
stramel: I believe this should still return and `IdType<D>`?

ggascoigne: No, it returns an arbitrary identifier. The default implementation returns the row number as a string, or a concatenation of row and parent row for grouped records.

ggascoigne: No, it returns an arbitrary identifier. The default implementation returns the row number as a string, or a concatenation of row and parent row for grouped records.

PR #42369, opened 2020-02-14
fix FilterQuery when use RootQuerySelector

kabytaa:
The following code is supposed to compile.
```javascript
const queries:Array<FilterQuery<ReturnType>> = [];
const condition :RootQuerySelector<ReturnType> = { $or: queries};
const result = await collection.find<ReturnType>(condition).toArray();
```
But getting the following error:
```
  Overload 1 of 2, '(query?: FilterQuery<ReturnType>): Cursor<ReturnType>', gave the following error.
    Argument of type 'RootQuerySelector<ReturnType>' is not assignable to parameter of type 'FilterQuery<ReturnType>'.
      Type 'RootQuerySelector<ReturnType>' is not assignable to type '{ [P in keyof ReturnType]?: Condition<ReturnType[P]>; }'.
  Overload 2 of 2, '(query: FilterQuery<ReturnType>, options?: FindOneOptions): Cursor<ReturnType>', gave the following error.
    Argument of type 'RootQuerySelector<ReturnType>' is not assignable to parameter of type 'FilterQuery<ReturnType>'.
      Type 'RootQuerySelector<ReturnType>' is not assignable to type '{ [P in keyof ReturnType]?: Condition<ReturnType[P]>; }'.ts(2769)
```

### DIFF
--- a/types/mongodb/index.d.ts
+++ b/types/mongodb/index.d.ts
@@ -1386,7 +1386,7 @@ export type Condition<T> = MongoAltQuery<T> | QuerySelector<MongoAltQuery<T>>;
 
 export type FilterQuery<T> = {
     [P in keyof T]?: Condition<T[P]>;
-} &
+} |
     RootQuerySelector<T>;
 
 /** http://docs.mongodb.org/manual/reference/command/collStats/ */


HosseinAgha (changes requested):
1. I don't think it is a good idea to use internal lib types inside your codebase.
2. You are changing a very sensitive piece of code and your change is breaking a lot of test cases. What if user had a key from the both Schema and root query in their query? (and many many other edge cases)
3. The TypeScript is actually right here. type of `condition` (RootQuerySelector) is a super (smaller) type of FilterQuery. In OOP language every Dog (FilterQuery) is an Animal (RootQuerySelector) but every Animal is not a Dog.

PR #42374, opened 2020-02-14
[jasmine] Add typings for resolveTo() and rejectWith() introduced in jasmine 3.5

coyoteecd:
See:
- https://jasmine.github.io/api/3.5/SpyStrategy.html
- https://github.com/jasmine/jasmine/pull/1688
- https://github.com/jasmine/jasmine/issues/1715

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://jasmine.github.io/api/3.5/SpyStrategy.html

### DIFF
--- a/types/jasmine/jasmine-tests.ts
+++ b/types/jasmine/jasmine-tests.ts
@@ -500,6 +500,53 @@ describe("A spy, when configured to fake a series of return values", () => {
     });
 });
 
+describe("A spy, when configured to fake a promised return value", () => {
+    const bar = 10;
+    const foo = {
+        getAsyncBar: () => {
+            return Promise.resolve(bar);
+        }
+    };
+
+    it("verifies return value type", () => {
+        spyOn(foo, "getAsyncBar").and.resolveTo(745);
+        spyOn(foo, "getAsyncBar").and.resolveTo("42"); // Is an error with TS 3.1+ typings.
+    });
+
+    it("tracks that the spy was called", async () => {
+        await foo.getAsyncBar();
+        expect(foo.getAsyncBar).toHaveBeenCalled();
+    });
+
+    it("when called returns the requested value", async () => {
+        spyOn(foo, "getAsyncBar").and.resolveTo(745);
+        await expectAsync(foo.getAsyncBar()).toBeResolvedTo(745);
+    });
+});
+
+describe("A spy, when configured to fake a promised rejection", () => {
+    const bar = 10;
+    const foo = {
+        getAsyncBar: () => {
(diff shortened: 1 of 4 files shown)


types/jasmine/jasmine-tests.ts
```diff
@@ -500,6 +500,53 @@ describe("A spy, when configured to fake a series of return values", () => {
...
+    const bar = 10;
+    const foo = {
+        getAsyncBar: () => {
+            return Promise.resolve(bar);
+        }
+    };
+
+    it("verifies return value type", () => {
+        spyOn(foo, "getAsyncBar").and.resolveTo(745);
+        spyOn(foo, "getAsyncBar").and.resolveTo("42"); // Is an error with TS 3.1+ typings.
```
djungowski: Not sure what the meaning of this test case is? It only creates spies but does nothing else

types/jasmine/jasmine-tests.ts
```diff
@@ -500,6 +500,53 @@ describe("A spy, when configured to fake a series of return values", () => {
     });
 });
 
+describe("A spy, when configured to fake a promised return value", () => {
+    const bar = 10;
+    const foo = {
```
djungowski: Please use
```
const foo = jasmine.createSpyObj(['getAsyncBar']);
foo.getAsyncBar.and.resolveTo(bar);
```

Plus: I would prefer more useful names than foo and bar ;-)

djungowski (changes requested):
Please rework the tests

types/jasmine/jasmine-tests.ts
coyoteecd: Meaning of the test case is to give a compile error if resolveTo() is removed from typings. Remember the tests are for the typings, not for the jasmine library (at least this is my informed guess by looking at existing tests). The exact same set of tests are there for and.returnValue().

Goal here was to be roughly consistent with the tests for and.returnValue, so unless you want to turn this into "refactor tests", I'd leave it as-is.

types/jasmine/jasmine-tests.ts
coyoteecd: I might prefer useful names too, but we got a bunch of foo bar everywhere and I'd stick to the existing conventions

coyoteecd:
@djungowski I explained in comments why I prefer to keep the tests as they are, if you need further changes let me know.

devoto13 (approved):

djungowski (approved):

PranavSenthilnathan:

PR #42378, opened 2020-02-14
[chrome] Match chrome.tabCapture.CaptureOptions with IDL definition

EchoAbstract:
In both the documentation and the IDL for tabCapture chrome uses
the confusing name `MediaStreamConstraint` which is not similar to
the actual `MediaStreamConstraints` browser type.  In the IDL the
documentation claims (incorrectly) that it's MediaTrackConstraints,
but there are additional undocumented supported fields so the type
ends up as `Object`:

See:

https://chromium.googlesource.com/chromium/src/+/b79156a434f215444989be10c007a1a992280108/chrome/common/extensions/api/tab_capture.idl#27

NOTE: This should be a non-breaking change as the `object` type is less strict and the previous definition would not work with current chrome.

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://chromium.googlesource.com/chromium/src/+/b79156a434f215444989be10c007a1a992280108/chrome/common/extensions/api/tab_capture.idl#27

### DIFF
--- a/types/chrome/index.d.ts
+++ b/types/chrome/index.d.ts
@@ -8,6 +8,7 @@
 //                 MatCarlson <https://github.com/MatCarlson>
 //                 ekinsol <https://github.com/ekinsol>
 //                 Thierry Régagnon <https://github.com/tregagnon>
+//                 Brian Wilson <https://github.com/echoabstract>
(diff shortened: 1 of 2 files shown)


fredrikhr (approved):
LGTM :+1:

PranavSenthilnathan:

PR #42392, opened 2020-02-15
feat(@blazor/javascript-interop): type definition for Blazor interop

peterblazejewicz:
This is the outcome of the discussion from the AspNetCore project:
dotnet/aspnetcore#18902
The definition covers JavaScript part of the .Net Core Blazor integration with the JavaScript:
https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interop?view=aspnetcore-3.1

- type definition based on @microsoft/dotnet-js-interop
- tests

The version number (major.minor) is taken off the .Net Core version (3.1)
The definition has exemption for `no-unnecessary-generics` TSLint rule, to stay synced strictly with the definition it was based on (`@microsoft/dotnet-js-interop` which has no public content available).

/cc @ryanelian @mkArtakMSFT

Thanks!

Please wait with merging until we get sign-off from MS folks

- [x] If not, do not conflict with the name of an NPM package. (the official package is @microsoft/dotnet-js-interop and has no public api to consume)

### DIFF
--- a/types/blazor__javascript-interop/blazor__javascript-interop-tests.ts
+++ b/types/blazor__javascript-interop/blazor__javascript-interop-tests.ts
@@ -1,0 +1,47 @@
+/// <reference lib="DOM" />
+//
+// see {@link https://docs.microsoft.com/en-us/aspnet/core/blazor/javascript-interop?view=aspnetcore-3.1}
+//
+const exampleJsFunctions = {
+    showPrompt(text: string | undefined) {
+        return prompt(text, 'Type your name here');
+    },
+    displayWelcome(welcomeMessage: string) {
+        document.getElementById('welcome')!.innerText = welcomeMessage;
+    },
+    returnArrayAsyncJs() {
+        DotNet.invokeMethodAsync<number[]>('BlazorSample', 'ReturnArrayAsync').then(data => {
+            data.push(4);
+            console.log(data);
+        });
+    },
+    sayHello(dotnetHelper: DotNet.DotNetObject) {
+        return dotnetHelper.invokeMethodAsync<string>('SayHello').then(r => console.log(r));
+    },
+};
+
+interface ColorFlags {
+    red: boolean;
+    green: boolean;
+    blue: boolean;
+}
+
+const testInteropApi = async (dotNetRef: DotNet.DotNetObject) => {
+    // validate api patterns
+    const tokens = [1, 2, 3];
+    DotNet.invokeMethod<string>('MyCoolApp.Core', 'Foo', 'First', 'Second'); // $ExpectType string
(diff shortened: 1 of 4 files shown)


ryanelian:
Very interesting, I didn't know a package [`@microsoft/dotnet-js-interop`](https://github.com/dotnet/extensions/blob/master/src/JSInterop/Microsoft.JSInterop.JS/src/package.json) exists in npm.

Why not use `@types/@microsoft/dotnet-js-interop` instead of `@types/@blazor/javascript-interop` to match the name?

peterblazejewicz:
> Why not use `@types/@microsoft/dotnet-js-interop` instead of `@types/@blazor/javascript-interop` to match the name?

Thanks! I know, I know. It looks like (at least for now) internal, non-public and advertised for public consumption package used by MS to build the Blazor related components (e.g. in the Web.JS project and related), Hence this `@blazor` derived namespace (because this seems to be a way people name this feature: https://www.google.com/search?q=blazor+javascript+interop). We could use `@microsoft/dotnet-js-interop`, but we need OK for this. It would work though, as no one really installs `@microsoft/dotnet-js-interop` directly I believe atm, one would just install `npm i -D @types/@microsoft/dotnet-js-interop`). Let's wait for feedback

msschl:
This definition would be awesome. Thank you @peterblazejewicz

PranavSenthilnathan:
I'm fine with using the `@microsoft` scope for this package. If the package is published later with its own types then the publisher can just remove the types from this repo. The [source](https://github.com/dotnet/extensions/blob/master/src/JSInterop/Microsoft.JSInterop.JS/src/package.json#L2) for the definitions here does indicate this package name as well.

mkArtakMSFT:
Thanks @peterblazejewicz.
Can you please include @SteveSandersonMS as a reviewer? He's from the Blazor team.

peterblazejewicz:
> Can you please include @SteveSandersonMS as a reviewer? He's from the Blazor team.

Yes, ofc. Can do this only via cc. though:
/cc @SteveSandersonMS

SteveSandersonMS:
This looks good to me! I'm happy with the `.d.ts` contents.

I don't know about the other files as they seem specific to DefinitelyTyped so someone else should review those.

ryanelian:
How to use the `DotNetObject` interface?

image

It is not available nor can be `import`ed...

peterblazejewicz:
@ryanelian 
This is based on global namespace, so:
```ts
class MyClass {
    myFunction(dotNetRef: DotNet.DotNetObject) {
        const msg = dotNetRef.invokeMethod<string>("SayHello", "World");
        dotNetRef.invokeMethodAsync<string>("SayHello", "World").then(msg => {
            console.log(msg);
        });
    }
}
```
hth

ryanelian:
Very strange indeed. I tried reinstalling the package and it still hasn't showed up:

image

PR #42398, opened 2020-02-16
fix(react): forwardRef accepts a special rendering function

### LINKED ISSUE #37660: [@types/react] Incorrect type definitions for React.forwardRef
- [x] I tried using the `@types/react@^16.7` package and had problems.
- [x] I tried using the latest stable version of tsc. `3.5.3`
  - Authors: @johnnyreilly @bbenezech @pzavolinsky @digiguru @ericanderson @DovydasNavickas @theruther4d @guilhermehubner @ferdaber @jrakotoharisoa @pascaloliv @Hotell @franklixuefei @Jessidhia @saranshkataria @lukyth @eps1lon

---

When using `React.forwardRef` I would like to add some `propTypes` and `defaultProps` to the returned component. The definition for the function is:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1709112cb66b86035ea8af59448424a5dc772010/types/react/index.d.ts#L730

Which returns a `ForwardRefExoticComponent`

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1709112cb66b86035ea8af59448424a5dc772010/types/react/index.d.ts#L724-L728

If you follow the type extensions you will see it does not define `propTypes` as an allowed attribute.

Thus you cannot define `propTypes`, only `defaultProps`.

However looking at the argument accepted, `RefForwardingComponent`:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/1709112cb66b86035ea8af59448424a5dc772010/types/react/index.d.ts#L503-L509

Seems to imply that you can have `propTypes` on the function passed, for example:

```tsx
const ExternalLink: React.RefForwardingComponent<
  HTMLAnchorElement,
  React.HTMLProps<HTMLAnchorElement>
> = ({href, ...otherProps}, ref) => <a ref={ref} {...otherProps} href={href} />;

ExternalLink.propTypes = {
  href: PropTypes.string.isRequired,
  target: PropTypes.string,
  rel: PropTypes.string,
};

// Should we allow these to be overridden?
ExternalLink.defaultProps = {
  target: '_blank',
  rel: 'noreferrer noopener',
};

export default React.forwardRef(ExternalLink);
```

However, doing this will produce the runtime error:

```
Warning: forwardRef render functions do not support propTypes or defaultProps. Did you accidentally pass a React component?
```

Seems like the types here are wrong?

eps1lon:
Closes #37660 

The returned component already allows `propTypes` (was probably fixed without referencing the issue). However, `forwardRef` still allowed a proper component (which causes a warning when using `defaultProps` and `propTypes`). This PR fixes this by narrowing the type to a special "render function".

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://reactjs.org/docs/react-api.html#reactforwardref
- ~[ ]~ If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- ~[ ]~ If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

### DIFF
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -746,6 +746,11 @@ declare namespace React {
         (props: PropsWithChildren<P>, ref: Ref<T>): ReactElement | null;
         displayName?: string;
     }
+    /**
+     * @deprecated Use ForwardRefRenderingFunction. forwardRef doesn't accept a
+     *             "real" component.
+     */
+    interface RefForwardingComponent <T, P = {}> extends ForwardRefRenderingFunction<T, P> {}
 
     function forwardRef<T, P = {}>(render: ForwardRefRenderingFunction<T, P>): ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>>;
 
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -522,7 +522,7 @@ declare namespace React {
         displayName?: string;
     }
 
-    interface ForwardRefRenderingFunction<T, P = {}> {
+    interface ForwardRefRenderFunction<T, P = {}> {
         (props: PropsWithChildren<P>, ref: Ref<T>): ReactElement | null;
         displayName?: string;
     }
@@ -531,7 +531,7 @@ declare namespace React {
      * @deprecated Use ForwardRefRenderingFunction. forwardRef doesn't accept a
      *             "real" component.
      */
-    interface RefForwardingComponent <T, P = {}> extends ForwardRefRenderingFunction<T, P> {}
+    interface RefForwardingComponent <T, P = {}> extends ForwardRefRenderFunction<T, P> {}
 
     interface ComponentClass<P = {}, S = ComponentState> extends StaticLifecycle<P, S> {
         new (props: P, context?: any): Component<P, S>;
@@ -753,7 +753,7 @@ declare namespace React {
         propTypes?: WeakValidationMap<P>;
     }
 
-    function forwardRef<T, P = {}>(render: ForwardRefRenderingFunction<T, P>): ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>>;
+    function forwardRef<T, P = {}>(render: ForwardRefRenderFunction<T, P>): ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>>;
 
     /** Ensures that the props do not include ref at all */
     type PropsWithoutRef<P> =
--- a/types/react/index.d.ts
+++ b/types/react/index.d.ts
@@ -522,14 +522,6 @@ declare namespace React {
         displayName?: string;
     }
 
-    interface RefForwardingComponent<T, P = {}> {
-        (props: PropsWithChildren<P>, ref: Ref<T>): ReactElement | null;
-        propTypes?: WeakValidationMap<P>;
-        contextTypes?: ValidationMap<any>;
-        defaultProps?: Partial<P>;
-        displayName?: string;
-    }
-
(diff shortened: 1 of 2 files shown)


types/react/index.d.ts
```diff
@@ -750,7 +763,7 @@ declare namespace React {
         propTypes?: WeakValidationMap<P>;
     }
 
-    function forwardRef<T, P = {}>(Component: RefForwardingComponent<T, P>): ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>>;
+    function forwardRef<T, P = {}>(render: ForwardRefRenderFunction<T, P>): ForwardRefExoticComponent<PropsWithoutRef<P> & RefAttributes<T>>;
```
eps1lon: Follows naming of https://github.com/facebook/react/blob/0cf22a56a18790ef34c71bef14f64695c0498619/packages/react/src/forwardRef.js#L11

ferdaber (commented):
Should we use `undefined` instead of `never`? Also can you test what this does to `LibraryManagedAttributes`?

eps1lon:
> Should we use undefined instead of never? 

I prefer `never` here because it's closer to our intention. The error message isn't helpful ("{} is not assignable to undefined") but that wouldn't change by using `undefined`, no? Or what are you trying to achieve by using `undefined`.

> Also can you test what this does to LibraryManagedAttributes

I'm not following how this relates. I didn't change the return type of `forwardRef`. Only what it accepts.

ferdaber:
`never` tends to mess up conditional types for me (any naked type parameter that evaluates to `never` in a conditional type will short-circuit to the _else_ branch), so I just usually avoid it unless absolutely necessary. And yeah I misread the PR and thought it changed the exotic component type, but this looks good.

ferdaber (approved):

eps1lon:
The types-publisher is causing some issues with versioned types. Do you know what's up @sandersn?

sandersn:
Yep, it's a known issue which is now fixed. I restarted CI and now that Travis has passed, typescript-bot should put this in the "To merge" pile.

types/react/index.d.ts
blasmedina: help, this change I generate an error in project...

wulucxy:
@eps1lon

hi，after I update to this version，ts compile warning me this：

```
Argument of type 'FunctionComponent<InfoSelectProps>' is not assignable to parameter of type 'ForwardRefRenderFunction<unknown, InfoSelectProps>'.
  Types of property 'defaultProps' are incompatible.
    Type 'Partial<InfoSelectProps> | undefined' is not assignable to type 'undefined'.
      Type 'Partial<InfoSelectProps>' is not assignable to type 'undefined'.
```

this is my code:

```
import { SelectProps } from 'antd/es/select'

export interface InfoSelectProps extends Omit<SelectProps, 'onChange'> {
  axiosInstance?: (params: any) => Promise<any>
  url?: string
  className?: string
  options?: Record<string, any>
  labelKey?: string
  valueKey?: string
  format?: (option: OptionType, item?: OptionType) => OptionType
  type?: 'flat' | 'object'
  dataSource?: any[]
  dynamicItems?: OptionType[]
  transformDynamicItems?: (list: OptionType[], dynamicItems: OptionType[]) => void
  transform?: (list: OptionType[], prevList?: OptionType[]) => OptionType[]
  onChange?: (value: SelectValue, row?: OptionType | OptionType[], list?: OptionType[]) => void
  onFinish?: (data: DataType) => void
}

export default React.forwardRef<unknown, InfoSelectProps>(InfoSelect)
```

so，how can I fix this?

PR #42405, opened 2020-02-16
[@hapi/catbox-memcached] add types for @hapi/catbox-memcached

btmorex:
Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/hapi__catbox-memcached/index.d.ts
+++ b/types/hapi__catbox-memcached/index.d.ts
@@ -1,0 +1,25 @@
+// Type definitions for @hapi/catbox-memcached 3.0
+// Project: https://github.com/hapijs/catbox-memcached#readme
+// Definitions by: Avery Fay <https://github.com/btmorex>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import { CacheKey, ClientApi, ClientOptions } from '@hapi/catbox';
+
+declare namespace CatboxMemcached {
+    interface Options extends ClientOptions {
+        host?: string;
+        port?: number;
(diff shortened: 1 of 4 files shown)


types/hapi__catbox-memcached/index.d.ts
```diff
@@ -0,0 +1,25 @@
...
+    interface Options extends ClientOptions {
+        host?: string;
+        port?: number;
+        location?: string;
+        timeout?: number;
+        idle?: number;
+    }
+}
+
+interface CatboxMemcached<T> extends ClientApi<T> {}
```
PranavSenthilnathan: Is this needed? I think the class declaration below is sufficient.

PranavSenthilnathan (changes requested):

types/hapi__catbox-memcached/index.d.ts
btmorex: Doing that or listing out all the members is required. Without it, you get:
```
Class 'CatboxMemcached<T>' incorrectly implements interface 'ClientApi<T>'.
  Type 'CatboxMemcached<T>' is missing the following properties from type 'ClientApi<T>': start, stop, get, set, and 3 more.ts(2420)
```
Originally, I had them listed out, but I took a peek at the types for @hapi/catbox-memory and @hapi/catbox-redis and they're both using this trick, so I followed suit.

btmorex: I'm going to hit resolve because I think that's the way to push this forward, but to be clear, I'm happy to change it if I'm not doing it in the right way.

Issue #26532, opened 2018-06-14
whether @types/gsap needs update


Renhj90520:
Hi, I tried using the `@types/gsap` package and had problems.
gsap suggests that import CSSPlugin as following 
import { CSSPlugin } from "gsap/all";  or import CSSPlugin from 'gsap/CSSPlugin'
I do this in my angular 4 project, but it seems not working, and gsap is already at version 2.0.1, is this one needed update?
@codeBelt  @ProbablePrime

ProbablePrime:
@Renhj90520 Yes, it would need an update to support this. 

Unfortunately the way things are configured right now each plugin will need types added to: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/gsap/Plugins.d.ts

GSAP has a ton of plugins and when i did the last update I only touched ones I was used or that were already present.

Gabb1995:
Do Types help with tree shaking when using the build optimizer? Because I'm trying to use GSAP in my Angular 7 project and with build optimizer enabled I keep getting that TimelineLite is not a constructor.

ZachSaucier:
GSAP 3 now has its own official Typescript definitions included in the package. Please post issues on [the official GSAP repo](https://github.com/greensock/GSAP).

PR #42410, opened 2020-02-16
Add type for TweenLite.version

fuleinist:
Add type for TweenLite.version


ProbablePrime:
As per: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/26532

GSAP has its own types. https://github.com/greensock/GSAP/tree/master/types

We should most likely remove the DT types.

PR #42417, opened 2020-02-17
feat(node): Add support for TypeScript 3.7 assertion functions

### LINKED ISSUE #40415: @types/node Use Typescript 3.7 asserts syntax for node's assert function
About the `asserts` feature:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions
https://github.com/microsoft/TypeScript/pull/32695

Basically I would like to change:
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/9f57df89f30bacee1d46dfbf40a08132b4b059ef/types/node/assert.d.ts#L2

into this:
```ts
function internal(value: any, message?: string | Error): asserts value;
```

Since this TypeScript feature has been released only recently, how should this best be added to the types without breaking compilation on older TypeScript versions?

cc @Flarna @rbuckton @jstewmon (This is obviously not the complete list of authors because I didn't want to ping too many people. While browsing the source these names came up. Please ping anyone who you think should be involved in the discussion)

G-Rath:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [assertion functions](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions)

I've made a start on this, but it was getting a bit tangly so wanted to push it up for the diff & for some guidance on how to handle a few things :)

Closes #40415


G-Rath:
@ExE-Boss would be cool to have your eyes on this, as you did the `assert` package types.

I'm having issues having the type versions actually take effect, which I think relates to `export =`. I can't actually find an example of a `@types` library with a similar shape as this, so I'm wondering if it's even possible...?

ExE-Boss:
No, it’s because the `ts3.7` need `index.d.ts` similar to what's in `ts3.5`, but extending from `ts3.5`:

<dl>
<dt><code>/types/node/ts3.7/index.d.ts</code>:</dt>
<dd><p></p>

```ts
// NOTE: These definitions support NodeJS and TypeScript 3.5.

// Base definitions for all NodeJS modules that are not specific to any version of TypeScript:
// tslint:disable-next-line: no-bad-reference
/// <reference path="../ts3.5/index.d.ts" />

// TypeScript 3.7-specific augmentations:
/// <reference path="assert.d.ts" />
```
</dd></dt></dl>

Also, you don’t have to do this for **Node < v13**, as the `assert` module is as old as **Node** itself.

G-Rath:
@ExE-Boss ah that makes sense - I wondered if I needed to do that; however that doesn't seem to have resolved the issue.

> Also, you don’t have to do this for Node < v13, as the assert module is as old as Node itself.

I'm not sure what you mean by this - while assert has been around since the start, it's had properties deprecated and api changes.

G-Rath:
@typescript-bot I'm waiting on input from people to help resolve the build issues.

types/node/v11/package.json
```diff
@@ -6,6 +6,11 @@
             "*": [
                 "ts3.2/*"
             ]
+        },
```
SimonSchick: Should probably omit v11, it's EOL.

types/node/ts3.7/assert.d.ts
```diff
@@ -0,0 +1,53 @@
...
+
+            constructor(options?: {
+                message?: string; actual?: any; expected?: any;
+                operator?: string; stackStartFn?: Function
+            });
+        }
+
+        type AssertPredicate = RegExp | (new() => object) | ((thrown: any) => boolean) | object | Error;
+
+        function fail(message?: string | Error): never;
```
SimonSchick: Can we omit all the functions that do not use the `assert` functionality, TS should fall back iirc.

SimonSchick (approved):
Little bit worried with fragmentation here, we have 2.8, (3.2), 3.5 and now 3.7, getting complicated enough that we should probably add some docs (readme.md).

types/node/ts3.7/assert.d.ts
G-Rath: It'd be great if it does, but sadly I can't get them working at all right now :(

Judging by the types in `ts3.5`, I'd say TS does fallback.

SimonSchick: That's odd, partial augmentation usually works, see `fs.d.ts` etc. in the ts3.5 folder 👀

G-Rath: It's not the partial - it's just the whole thing entire.

It's actually weird, as TS seems to completely ignore `assert.d.ts` unless I move it into `ts3.5`, at which point it starts complaining about `export = internal` being a duplicate.

It's like it's actually completely ignoring the `ts3.7` in favor of `ts3.5` - I wonder if it's FIFO?

types/node/ts3.7/assert.d.ts
```diff
@@ -0,0 +1,53 @@
+declare module "assert" {
```
ExE-Boss:
```suggestion
// tslint:disable-next-line:no-bad-reference
/// <reference path="../assert.d.ts" />

declare module "assert" {
```

types/node/ts3.7/index.d.ts
```diff
@@ -0,0 +1,8 @@
+// NOTE: These definitions support NodeJS and TypeScript 3.7.
+
+// Base definitions for all NodeJS modules that are not specific to any version of TypeScript:
+// tslint:disable-next-line:no-bad-reference
+/// <reference path="../base.d.ts" />
```
ExE-Boss: This needs to extend from the TS3.5 types.

```suggestion
/// <reference path="../ts3.5/index.d.ts" />
```

types/node/ts3.7/assert.d.ts
ExE-Boss: You need to put `ts3.7` before `ts3.5` in the `typesVersions` field.

ExE-Boss (changes requested):

G-Rath:
@ExE-Boss cheers for confirming that - that change results in the expected behaviour, but unfortunately TypeScript still seems to complain about the identifier being duplicated:

```
node_modules/@types/node/assert.d.ts:52:14 - error TS2300: Duplicate identifier 'internal'.

52     export = internal;
                ~~~~~~~~

  node_modules/@types/node/ts3.7/assert.d.ts:55:14
    55     export = internal;
                    ~~~~~~~~
    'internal' was also declared here.

node_modules/@types/node/ts3.7/assert.d.ts:55:14 - error TS2300: Duplicate identifier 'internal'.

55     export = internal;
                ~~~~~~~~

  node_modules/@types/node/assert.d.ts:52:14
    52     export = internal;
                    ~~~~~~~~
    'internal' was also declared here.
```

This happens regardless of if `<reference path="../assert.d.ts" />`  is in `assert.d.ts`, which is why I've not yet committed that change.

Removing the `export = internal` results in:

```
node_modules/@types/node/assert.d.ts:52:5 - error TS2309: An export assignment cannot be used in a module with other exported elements.

52     export = internal;
       ~~~~~~~~~~~~~~~~~~
```

G-Rath:
@SimonSchick if you've got any ideas on how to fix the errors, I'm happy to take another crack at this, but for now I'm out of ideas :/

SimonSchick:
I tried to play around with your PR but overriding/augmenting `export = ...` modules (what are those even called?) just doesn't work.

If we ditched the requirement for declaration merging and removed the top level `function internal` we could probably make it work but we would be breaking anyone that used:

```ts
import * as assert from 'assert';

assert(something) // would error
```

G-Rath:
Yeah, I suspect it just isn't possible, which means that either TS needs to change or support for <3.7 needs dropping.

Do you think it's worth pinging the ts team? - given that the `assert` module was their perfect showcasing of this feature, and that I can't find any reference of this exact code layout anywhere (as you say: "what are those even called?" - the `assert` library works b/c it's not wrapped in `module` - so the issue seems to actually be `export = ` + `module` = impossible to override module) maybe they didn't consider this code structure :/

SimonSchick:
@G-Rath this is what I am talking about https://github.com/DefinitelyTyped/DefinitelyTyped/commit/3e79429a7b45d750cd07157813f36f4b14acd91a (credits for tests etc. go to you obv.)

G-Rath:
Yeah, but as you said that prevents you from doing `assert(...)`.

However, I just noticed #41919 - is there anyway we could leverage that?

One possible somewhat ideal solution would be if we could tell TS to replace rather than merge - we'd be required to include the entire assert file, but it would do the job 🤔

G-Rath:
@SimonSchick I think I know the issue & how to fix it, but need to catch a train (I have amazing timing) - I'll push something up for you to see soon.

PR #42419, opened 2020-02-17
[aws-lambda] Add protocol to APIGatewayRequestContext.

### LINKED ISSUE #42049: [@types/aws-lambda] Missing key in Request context.

  - Authors: @darbio, @skarum, @StefH, @tobyhede, @buggy, @y13i

=====

`APIGatewayEventRequestContext` is missing a `protocol` key of type `string`. Mine shows this inside the `requestContext` object:
```json
{
  "protocol": "HTTP/1.1",
}
```

simonbuchan:
Fixes #42049.

No decent docs from AWS for the actual proxy integration, on the
Lambda side or the API Gateway side, but there's an approximation on
the API Gateway docs for the VTL $context variable:

https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html#context-variable-reference

From experimentation, not all of the variables referenced above exist,
at least most of the time, but everything except the `protocol` that
was present in what I received already existed. If that makes any sense.

### DIFF
--- a/types/aws-lambda/common/api-gateway.d.ts
+++ b/types/aws-lambda/common/api-gateway.d.ts
@@ -9,6 +9,7 @@ export interface APIGatewayEventRequestContext {
     domainPrefix?: string;
     eventType?: string;
     extendedRequestId?: string;
+    protocol: string;
     httpMethod: string;
     identity: {
         accessKey: string | null;
--- a/types/aws-lambda/test/api-gateway-tests.ts
+++ b/types/aws-lambda/test/api-gateway-tests.ts
@@ -28,6 +28,7 @@ let proxyHandler: APIGatewayProxyHandler = async (event, context, callback) => {
     requestContext = event.requestContext;
     str = event.resource;
 
+    str = requestContext.protocol;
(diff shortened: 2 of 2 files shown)


joakimsandstromfinnair:
I think https://github.com/DefinitelyTyped/DefinitelyTyped/blob/b9b6b0e9bd5de6ed18be0a3e07816f60fe6f2cfa/types/aws-lambda/trigger/api-gateway-proxy.d.ts#L17 should also have the protocol added. Since now we get:

```
[tsl] ERROR in /Users/xx/yy/src/docs-get.ts(39,11)
      TS2345: Argument of type 'import("/Users/xx/node_modules/@types/aws-lambda/index").APIGatewayProxyEvent' is not assignable to parameter of type 'AWSLambda.APIGatewayProxyEvent'.
  Types of property 'requestContext' are incompatible.
    Property 'protocol' is missing in type 'import("/Users/xx/node_modules/@types/aws-lambda/index").APIGatewayEventRequestContext' but required in type 'AWSLambda.APIGatewayEventRequestContext'.
```

PR #42420, opened 2020-02-17
[aws-lambda] New API Gateway Authorizer types, deprecating old ones.

### LINKED ISSUE #24178: Split up CustomAuthorizerEvent into separate interfaces for REQUEST and TOKEN

- [X] I tried using the `@types/aws-lambda` package and had problems.
- [X] I tried using the latest stable version of tsc. https://www.npmjs.com/package/typescript (Using typescript 2.6.1, however this is unrelated to tsc versions)
  - Not sure who to tag since there are currently 22 contributors listed so I picked 5 of the most recent from the index.d.ts history please let me know if somebody should be tagged but isn't, Authors: @Haydabase @simonbuchan @andy-ms @coderbyheart @sheetalkamat 

I propose that the CustomAuthorizerEvent should be split up into two different types representing the REQUEST type custom authorizer and the TOKEN type custom authorizer. We just experienced a case where our custom authorizer Lambda was configured to be of type TOKEN but we were attempting to use the REQUEST attributes. Having this more explicit in the types would probably have made us realize this mistake earlier.

simonbuchan:
Noticed this testing #42419. Fixes #24178.

When integrating a custom authorizer, you actually have two options,
creating a token or a request authorizer, which changes what payload
you will get. You nearly certainly know which you will be called with!

Just deprecating the old version as it's kinda broken in a way thats
hard to fix without breaking someone, but we want to guide devs to the
new version.

It is possible to fix the existing type by adding a bunch of
`foo?: never` fields to each alternative so existing accesses don't,
error but this makes things more complex, and confusing for the common
case.

Other ideas welcome!


simonbuchan:
To the maintainer: probably give this one a day or so in case someone wants to complain. Also, is there any etiquette on using `/** @deprecated */`? I've used it here as the existing type is worse experience, but the justification for the error isn't particularly clear.

furudean (approved):

simonbuchan:
To the maintainer: Please don't merge for now, found an issue when [testing merging the change from #42421](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42421#issuecomment-592585406)

simonbuchan:
OK, I think that should be good, @c-bandy - does this change work for you? With more testing I found the integration response properties except for null are all converted to string, which is an interesting quirk to test.

furudean (approved):
Verified that the changes to `CustomAuthorizerResult` and `CustomAuthorizerCallback` now restrict the `context`s properties as expected. Nice work!

JasonTruter:
methodArn isn't added to the APIGatewayRequestAuthorizerEvent.

simonbuchan:
@JasonTruter: just double checked, you're right, it is sent on both. Should be a very easy fix. If you open this as an issue and @ me on it, I'll be reminded and you'll get notified when the fix is merged, but it's not required.

Issue #34059, opened 2019-03-21
[aws-lambda] Claims property is missing for Authorizer


bykka:
According to the documentation "https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-mapping-template-reference.html" AuthResponseContext class can also contain property "claims".

So, an idea is to introduce a new interface Claims and to add it to the possible values of AuthResponseContext

`
export interface Claims {
[name: string]: boolean | number | string;
}
`
`
export interface AuthResponseContext {
    [name: string]: boolean | number | string | Claims;
}
`

Authors: @darbio, @buggy, @MichaelMarner

simonbuchan:
Going back through `aws-lambda` issues, since they seem to be being neglected.

Those docs are more detective work than authoritative on what can be in the context, since they are for the mapping templates, not the proxy integration, but that's fine if its confirmed by experimentation. Unfortunately I'm pretty suspicious of tightening the context, at least without end developer request, given #33437 got reverted, and AWS are happy to let their own authenticator break the rules on lambda authenticaros.

I'm instead thinking the change should be making the context a type parameter, so you can reliably expect things from a custom lambda authenticator or declare you are using a cognito authenticator, e.g. something like:

```ts
// handlers.ts
import { APIGatewayProxyHandler, CustomAuthorizerHandler, CognitoAuthContext } from "aws-lambda";

export const cognitoUsername: APIGatewayProxyHandler<CognitoAuthContext> = async (event) => {
  return event.authorizer.claims.username;
};

interface MyAuthContext {
  token: string;
}

export const customAuthorizer: CustomAuthorizer<MyAuthContext> = async (event) => {
  return { ..., context: { token: "123" } };
};

export const getCustomToken: APIGatewayProxyHandler<MyAuthContext> = async (event) => {
  return event.authorizer.token;
};
```

With appropriate defaults for compatibility. I'm a bit concerned that `APIGatewayProxyHandler<T>` implies that `T` is the *body* type (which is actually, of course, stringified and thus untypable), so maybe this is better:

```ts
export interface APIGatewayProxyHandlerWithAuthorizer<TAuthContext> {
  ... as before
  authorizer: TAuthContext;
}
export type APIGatewayProxyHandler = APIGatewayProxyHandlerWithAuthorizer<any>;
[...]

PR #42421, opened 2020-02-17
[aws-lambda] Narrow AuthResponseContext type

### LINKED ISSUE #42418: AuthResponseContext in aws-lambda should not accept props of type any
The context properties of an AuthResponseContext (types/aws-lambda/common/api-gateway.d.ts) can be of types string, bools and numbers hence **any** is misleading.

See: https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html

The interface AuthResponseContext should not accept properties of type **any**, instead it should accept properties of type **string | boolean | number**.

furudean:
Fixes #42418

> You can access the stringKey, numberKey, or booleanKey value (for example, "value", "1", or "true") of the context map in a mapping template by calling $context.authorizer.stringKey, $context.authorizer.numberKey, or $context.authorizer.booleanKey, respectively. The returned values are all stringified. Notice that you cannot set a JSON object or array as a valid value of any key in the context map.

From https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html

### DIFF
--- a/types/aws-lambda/common/api-gateway.d.ts
+++ b/types/aws-lambda/common/api-gateway.d.ts
@@ -42,5 +42,5 @@ export interface APIGatewayEventRequestContext {
  * http://docs.aws.amazon.com/apigateway/latest/developerguide/use-custom-authorizer.html#api-gateway-custom-authorizer-output
  */
 export interface AuthResponseContext {
-    [name: string]: any;
+    [name: string]: string | number | boolean;
 }
--- a/types/aws-lambda/common/api-gateway.d.ts
+++ b/types/aws-lambda/common/api-gateway.d.ts
@@ -42,5 +42,5 @@ export interface APIGatewayEventRequestContext {
  * http://docs.aws.amazon.com/apigateway/latest/developerguide/use-custom-authorizer.html#api-gateway-custom-authorizer-output
  */
 export interface AuthResponseContext {
-    [name: string]: string | number | boolean;
+    [name: string]: string | number | boolean | undefined;
 }
--- a/types/aws-lambda/index.d.ts
+++ b/types/aws-lambda/index.d.ts
@@ -33,6 +33,7 @@
 //                 Alex Bolenok <https://github.com/alex-bolenok-centralreach>
 //                 Marian Zange <https://github.com/marianzange>
 //                 Alexander Pepper <https://github.com/apepper>
+//                 Cassidy Bandy <https://github.com/c-bandy>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 2.8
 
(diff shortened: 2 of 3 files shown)


simonbuchan:
Docs are a bit dumb here, there's been an attempt to narrow before is #33437 that was reverted since the cognito authorizer will give you an object in `event.authorizer.claims`. This can be fixed by splitting the authorizer (output) and proxy handler (input) types, but that's a bigger breaking change.

I've put some thoughts about where I think this should go in the related issue #34059, any feedback?

c-bandy:
Ah! Shame it isn't as straight-forward as documented.

What do you think about using a union type to represent a generic `context`? This could cover the custom authorizer context and the cognito context property.

c-bandy:
On #34059, I basically agree. I think your last solution is as elegant as it gets. 

If not for that, at the very least we should differentiate the `APIGatewayProxyEvent.requestContext`, `CustomAuthorizerResult.context` and `CustomAuthorizerEvent.requestContext` types.

simonbuchan:
Yeah, I'm happy narrowing the authorizer result context, it should only catch actual errors (unless someone's playing tricks with `toString()` / `toJSON()`? They knew what they were getting into.)

I also have #42420 open now which will conflict with this if it goes in. If I'm bringing in a new type anyway I can make these changes at the same time?

c-bandy:
I'll wait for your PR to be merged until I make any changes here

c-bandy:
@simonbuchan sorry, looks like i completely parsed your question wrong. feel free to do so!

StefH (approved):

darbio:
Isn't this going to be addressed by @simonbuchan in #42420? If so this PR should **not** be merged.

simonbuchan (changes requested):
More relevantly, a very similar PR #33437 has previously had to have been rolled back, so whether this gets added to my PR or not (I was just commenting that it would conflict) this shouldn't be merged.
I'm going to mark this request changes so it doesn't slip through.

simonbuchan:
... @c-bandy you thumbed that up so fast it was there before the page loaded after my post 😲

c-bandy:
@simonbuchan I was already here writing essentially the same message you were, but you were faster than me :P

simonbuchan:
Since it's been open for a week and at least nobody seems to be complaining about it, I'll throw the changes in.

simonbuchan:
Let me know if the latest changes in #42420 look like they fix this for you.

c-bandy:
I havent had time to test your changes but the source code looks good to me

c-bandy:
Hey @simonbuchan, I tried testing this in one of our repos, it looks like the issue raised here is not solved by your work so far, I can still put `any` inside the context for a custom authorizer. Am I expected to use a different type here?

image

simonbuchan:
Ah, yes, by default my PR doesn't narrow the response context type, it probably should. You can instead declare a type that describes the context exactly, then use that with `APIGatewayAuthorizerWithContextResult<MyContext>`, and matching `APIGatewayProxyWithContextEvent<MyContext>`.

I can quickly add narrowing the response type in my PR on my Monday, which should be before a maintainer merges it, since I'm UTC+12, but I might forget. Won't be a huge problem to PR just that though.

PR #42422, opened 2020-02-17
fix(ejs): add missing options property

peterblazejewicz:
- `outputFunctionName` configuration option
- test updated

Thanks!

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://ejs.co/#docs

### DIFF
--- a/types/ejs/index.d.ts
+++ b/types/ejs/index.d.ts
@@ -414,7 +414,10 @@ export interface Options {
 	 *
 	 * @default ejs.localsName
 	 */
-	localsName?: string;
+    localsName?: string;
+
+    /** Set to a string (e.g., 'echo' or 'print') for a function to print output inside scriptlet tags. */
+    outputFunctionName ?: string;
 }
 
 export interface Cache {
--- a/types/ejs/index.d.ts
+++ b/types/ejs/index.d.ts
@@ -416,7 +416,7 @@ export interface Options {
 	 */
 	localsName?: string;
 
-    /** Set to a string (e.g., 'echo' or 'print') for a function to print output inside scriptlet tags. */
+	/** Set to a string (e.g., 'echo' or 'print') for a function to print output inside scriptlet tags. */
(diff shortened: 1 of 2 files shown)


types/ejs/index.d.ts
```diff
@@ -414,7 +414,10 @@ export interface Options {
 	 *
 	 * @default ejs.localsName
 	 */
-	localsName?: string;
+    localsName?: string;
```
ExE-Boss:
```suggestion
	localsName?: string;
```

types/ejs/index.d.ts
```diff
@@ -414,7 +414,10 @@ export interface Options {
...
+
+    /** Set to a string (e.g., 'echo' or 'print') for a function to print output inside scriptlet tags. */
```
ExE-Boss:
```suggestion
	/** Set to a string (e.g., 'echo' or 'print') for a function to print output inside scriptlet tags. */
```

types/ejs/index.d.ts
```diff
@@ -414,7 +414,10 @@ export interface Options {
...
+    outputFunctionName ?: string;
```
ExE-Boss:
```suggestion
	outputFunctionName?: string;
```

ExE-Boss (changes requested):
This package uses tabs, because they’re better for developer accessibility (<https://github.com/prettier/prettier/issues/7475>, [Nobody talks about the real reason to use Tabs over Spaces : javascript](https://www.reddit.com/r/javascript/comments/c8drjo/nobody_talks_about_the_real_reason_to_use_tabs/))

ExE-Boss (approved):

ExE-Boss:
You might also want to add yourself to the “Definitions by:” field:

https://github.com/DefinitelyTyped/DefinitelyTyped/blob/308f467a1788ba1a70bc6115b949ace208bd7979/types/ejs/index.d.ts#L3-L4

PR #42427, opened 2020-02-17
feat(ejs): package level custom .editorconfig

peterblazejewicz:
The DT project's Prettier uses global .editorconfig, this change allows to
use package-level editor config settings, here TABS (vs. SPACES):
https://git.io/Jv8Ar

/cc @ExE-Boss

Thanks!

### DIFF
--- a/types/ejs/.editorconfig
+++ b/types/ejs/.editorconfig
@@ -1,0 +1,3 @@
+[*]
+indent_style = tab
+indent_size = 4
--- a/types/ejs/.editorconfig
+++ b/types/ejs/.editorconfig
@@ -1,3 +1,2 @@
 [*]
 indent_style = tab
-indent_size = 4


types/ejs/.editorconfig
```diff
@@ -0,0 +1,3 @@
+[*]
+indent_style = tab
+indent_size = 4
```
ExE-Boss: This line is unnecessary with tabs:
```suggestion
```

ExE-Boss (approved):
While I do approve of this, the @DefinitelyTyped team has rejected my PRs for containing this file in the past.

PranavSenthilnathan:
Let's just have one prettier config for the whole repo. We prefer having a single consistent style across Definitely Typed as much as possible.

peterblazejewicz:
> Let's just have one prettier config for the whole repo. We prefer having a single consistent style across Definitely Typed as much as possible.

The point of EditorConfig is to allow composition (one root, other are composed). Here we have just tried to avoid mistake I've done updating the types/ejs (converted tabs to spaces). This local editor config effects only the package (types/ejs) format.
Just discussing this, no problem with revisiting that in the future.
Thanks!

PR #42429, opened 2020-02-17
[@types/lingui__core]: Add number and date into i18n type

igorDolzh:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: [definition in the source code](https://github.com/lingui/js-lingui/blob/master/packages/core/src/i18n.js#L68)

### DIFF
--- a/types/lingui__core/i18n.d.ts
+++ b/types/lingui__core/i18n.d.ts
@@ -55,6 +55,10 @@ export class I18n {
 
     selectOrdinal(id: string, config: PluralProps): string;
 
+    number(value: number, format?: Intl.NumberFormatOptions): string;
+
+    date(value: string | Date, format?: Intl.DateTimeFormatOptions): string;
+
     constructor();
 
     availableLanguages: string[];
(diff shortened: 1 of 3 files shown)


huan086 (commented):
These methods do not exists on the i18n object at runtime. You should be importing from lingui/macro

huan086:
You are doing it wrong. Follow the tutorial https://lingui.js.org/tutorials/react.html. Import from lingui macro e.g. `import { Trans } from '@lingui/macro'`

igorDolzh:
Thanks for your guidance, but read this.
I tried this method first https://lingui.js.org/ref/macro.html#number for numbers
```
i18n._(t`There were ${number(10000)} people.`)
``` 
returns `There were [object Object] people. `

If I try without `t`
```
i18._(number(amount))
```
returns `amount,number,number0`

Also, I found this issue opened, people facing the same problem with `date` and `number` https://github.com/lingui/js-lingui/issues/560

Then, I was searching for other options to use "number", and `i18n.number` works as expected. It returns a string with the value.
And as you could see here https://github.com/lingui/js-lingui/blob/master/packages/core/src/i18n.js#L68, `number` and `date` are inside I18n. That's why I think it should be included in types as well.

Of course, `number` and `date` from `macro` should be fixed as well, but not here :)

If you know what I'm doing wrong, and how I could fix it, you will be a savior!)

sheetalkamat (changes requested):
please handle feedback already given

igorDolzh:
It is sad, that no discussion happened here. But apparently, my solution is controversial. I close the PR. I will try to fix `lingui/macro` instead.

PR #42438, opened 2020-02-17
Update to uuid@v7.x

### LINKED ISSUE #328: @types/uuid references @types/node

### LINKED ISSUE #368: Publish TypeScript types with package
Congratulations on releasing the big ES modules rewrite! I gave it a try in a simple toy app, and it works wonderfully. It was great to see tree-shaking working correctly!

The only feedback I have is that it'd be really nice to have the TypeScript definitions published as part of the package instead of a separate `@types` package. This would ensure that all consumers of v7 and up are able to use the package with TypeScript without installing another package that is potentially out of date. As an example, I'm unable to test the new v7 beta in my TypeScript projects since the published types are only for v3. In order to use the beta, I'd need to write my own declaration file, and if I'm doing that I figured I might as well offer to do it officially.

If you're interested, I'm willing to write new type declarations and send a PR adding them.

ctavan:
Fixes uuidjs/uuid#368, fixes uuidjs/uuid#328

We are about to do a major version bump in the `uuid` library and the recommended way of importing the library has changed, see https://github.com/uuidjs/uuid#upgrading-from-v3x-of-this-module

So far we only have beta versions of the new major version released which is why I open this as a draft PR so that I can already collect some feedback.

@LinusU since you recently fixed the types for `uuid` (thank you for that!!) I'd be really curious about your feedback on this PR.

This PR also relates to uuidjs/uuid#328

----

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/uuidjs/uuid/blob/master/README.md


LinusU:
The changes looks good to me 👍 

Not sure what the policy on removing old version are, so I cannot comment on the removal of v2.x, but if 7.x is the newest version it seems a bit excessive to still have the v2 typings ☺️ 

Potentially the old typings should be named 6.x instead of 3.x? 🤔

ctavan:
@LinusU thank you for taking a look!

> Not sure what the policy on removing old version are, so I cannot comment on the removal of v2.x, but if 7.x is the newest version it seems a bit excessive to still have the v2 typings ☺️

Well, the jump from 3.x to 7.x looks larger than it is: In fact the current stable version is 3.x and we decided to call the next major release 7.x for one specific reason: The UUID spec also uses the term "Version" to specify 4 different algorithms that are all supported by the `uuid` JS library, namely Version 1,3,4,5. To avoid any confusion of having a `uuid@v4` release which might be mistakenly related to the UUID Version 4 algorithm we decided to simply jump all version numbers that might cause confusion (there's a UUID Version 6 proposal out in the wild) and went for v7.x.

I can understand that this may look unexpected to people who are upgrading from 3.x and I think I will add a short section to the readme.

> Potentially the old typings should be named 6.x instead of 3.x? 🤔

Well, as it stands the current typings are valid for the `v3.x` series of the `uuid` module, so I think `v3` is an appropriate naming.

LinusU:
> Well, the jump from 3.x to 7.x looks larger than it is [...]

Ahh, hehe, that makes sense 👍 

I just assumed that the changes between 3 -> 4 -> 5 -> 6 didn't affect the typings and that no-one had bothered updating the version.

Everything looks good then 👏

ctavan:
Thanks @LinusU!

I think I'll have to let this one sit until the final `v7.0.0` release of the `uuid` module is done in order for the CI to be able to pass. Will kindly ask you for a formal review then.

ctavan:
@LinusU `v7.0.0` of `uuid` has been released so it's time to review and merge this one 😄

ctavan:
The CI complains with

```
Error: Could not find version [object Object]
```

Does anybody have a clue what's wrong @iamolivinius @felipeochoa @cjbarth @Rauno56 @LinusU?

LinusU (approved):

PR #42447, opened 2020-02-18
proxy-from-env: Add missing URL type for input parameter

flopes89:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/Rob--W/proxy-from-env/blob/master/index.js#L24

### DIFF
--- a/types/proxy-from-env/index.d.ts
+++ b/types/proxy-from-env/index.d.ts
@@ -1,7 +1,10 @@
 // Type definitions for proxy-from-env 1.0
 // Project: https://github.com/Rob--W/proxy-from-env#readme
 // Definitions by: JasonHK <https://github.com/JasonHK>
+//                 FloPes <https://github.com/flopes89>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+import { Url } from "url";
 
 /**
  * Takes an input URL and returns the desired proxy URL. If no proxy is set, an
@@ -10,4 +13,4 @@
  * @returns The URL of the proxy that should handle the request to the given
  *          URL.
  */
-export function getProxyForUrl(url: string|URL): string;
+export function getProxyForUrl(url: string|Url): string;
--- a/types/proxy-from-env/index.d.ts
+++ b/types/proxy-from-env/index.d.ts
@@ -4,6 +4,8 @@
 //                 FloPes <https://github.com/flopes89>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
+/// <reference types="node" />
+
 import { Url } from "url";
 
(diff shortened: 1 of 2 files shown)


JasonHK (approved):

JasonHK (changes requested):
Your build was failed, please check again

types/proxy-from-env/index.d.ts
```diff
@@ -10,4 +10,4 @@
  * @returns The URL of the proxy that should handle the request to the given
  *          URL.
  */
-export function getProxyForUrl(url: string): string;
+export function getProxyForUrl(url: string|URL): string;
```
JasonHK: The URL object `getProxyForUrl` accept is [this](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/d129b9b777901659b1559f9355f9548496e72a2b/types/node/url.d.ts#L20), not the standard `URL`.

JasonHK (changes requested):

types/proxy-from-env/index.d.ts
flopes89: I updated the types to use the `url` type, but the tests fail telling me the `"url"` module cannot be found. How can I fix this so the import works as expected? I checked other type definitions and they do this exactly the same as I do? (e.g. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/parseurl/index.d.ts#L9 )

flopes89:
I updated the types to use the url type, but the tests fail telling me the "url" module cannot be found. How can I fix this so the import works as expected? I checked other type definitions and they do this exactly the same as I do? (e.g. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/parseurl/index.d.ts#L9 )

types/proxy-from-env/index.d.ts
```diff
@@ -1,13 +1,16 @@
 // Type definitions for proxy-from-env 1.0
 // Project: https://github.com/Rob--W/proxy-from-env#readme
 // Definitions by: JasonHK <https://github.com/JasonHK>
+//                 FloPes <https://github.com/flopes89>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
+import { Url } from "url";
```
JasonHK: @flopes89 You're missing the triple-slash directive. Please put the code below before line 7.

```typescript
```

JasonHK (changes requested):

flopes89:
Thanks for the hint! I updated the reference type. Sorry for the mistake.

JasonHK (approved):
Thank you!

PranavSenthilnathan:

PR #38113, opened 2019-09-04
mongoose - Add AsyncIterator to DocumentQuery

muteshen:
Please fill in this template.

- [x] Use a meaningful title for the pull request. Include the name of the package modified.
- [x] Test the change in your own code. (Compile and run.)
- [x] Add or edit tests to reflect the change. (Run with `npm test`.)
- [x] Follow the advice from the [readme](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#make-a-pull-request).
- [x] Avoid [common mistakes](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/README.md#common-mistakes).
- [x] Run `npm run lint package-name` (or `tsc` if no `tslint.json` is present).

Select one of these and delete the others:

If changing an existing definition:
- [x] Provide a URL to documentation or source code which provides context for the suggested changes: <<https://mongoosejs.com/docs/api.html#query_Query-Symbol.asyncIterator>>
- [ ] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header.
- [ ] If you are making substantial changes, consider adding a `tslint.json` containing `{ "extends": "dtslint/dt.json" }`. If for reason the any rule need to be disabled, disable it for that line using `// tslint:disable-next-line [ruleName]` and not for whole package so that the need for disabling can be reviewed.

PR #42452, opened 2020-02-18
[mongoose] Add AsyncIterator on Query and Aggregate

Unibozu:
This PR adds support for AsyncIterator on mongoose `DocumentQuery` and `Aggregate`.

I had to add `ESNext.AsyncIterator` to `lib` which means upstream projects will need to add this lib themselves to update. There was no mention of how to deal with that in the Readme so let me know the best way. A few alternative can be to bring the types for `AsyncIterator` in the file or type the iterator with an `AsyncIterator`-like typing...

Tested on a fairly big API project using `for await`

---

.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes:  https://mongoosejs.com/docs/api/query.html#query_Query-Symbol.asyncIterator and https://mongoosejs.com/docs/api/aggregate.html#aggregate_Aggregate-Symbol.asyncIterator

### DIFF
--- a/types/mongoose/index.d.ts
+++ b/types/mongoose/index.d.ts
@@ -35,6 +35,7 @@
 //                 Thomas Pischulski <https://github.com/nephix>
 //                 Sam Kim <https://github.com/rlaace423>
 //                 Dongjun Lee <https://github.com/ChazEpps>
+//                 Clement Debiaune <https://github.com/Unibozu>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 // TypeScript Version: 3.0
 
@@ -2296,8 +2297,11 @@ declare module "mongoose" {
 
     /** Flag to opt out of using $geoWithin. */
     static use$geoWithin: boolean;
-  }
 
+    /** AsyncIterator - https://mongoosejs.com/docs/api/query.html#query_Query-Symbol.asyncIterator */
+    [Symbol.asyncIterator](): AsyncIterator<DocType>;
+  }
+  
   // https://github.com/aheckmann/mquery
   // mquery currently does not have a type definition please
   //   replace it if one is ever created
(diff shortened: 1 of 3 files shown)


Unibozu:
I may also add that another possible BC change is that `Aggregate<T>` now requires the `T` generic to be a non-array `DocType`. Typings were appropriately amended but there is a small risk that some codebases rely on this previous assumption

ToucheSir (commented):
Correct me if I'm wrong, but doesn't this have the same compat issues as #38113? Also not sure how this is usually handled, so any insight from @minestarks or any of the other maintainers would be much appreciated!

Unibozu:
@ToucheSir It's 100% the same PR.

I am not sure what are the rules regarding compat issues with typings especially if they pass the `lint` tests (works from TS 2.8). The main issue is the requirement to add `esnext.AsyncIterator` to actually be able to use these typings now.

A workaround is to bring the declarations in `lib.esnext.asynciterable.ts` to mongoose typings and return an `AsyncIterator`-like object, which will work equally well and allow us to only keep `es6` in libs. This seems like a reasonnable choice as mongoose only adds these iterators if `Symbol.AsyncIterator` are available.

Waiting on hearing from someone else

sandersn:
You could also create a new (non-DT) package named mongoose-modern-types that uses the module augmentation syntax and interface-class merging to add *just* `[Symbol.asyncIterator]()`. Something like:

```ts
interface DocumentQuery<T, DocType extends Document, QueryHelpers = {}> {
  [Symbol.asyncIterator](): AsyncIterator<DocType>;
}
```

Unibozu:
@sandersn Yeah, it's an interesting way to do it.

Can we add it to this repo as a sub-folder? I've seen it as a recommended way to deal with newer TS features (using `typesVersions`), so perhaps we can add it here as well and let people include the modern types by requiring `@types/mongoose/esnext`?

It's a shame TS doesn't have a feature similar to `typesVersions` to target `libs` instead.

sandersn:
@Unibozu I think that would work, although it would have to be a triple-slash reference since a real `import` statement will end up in the emitted js:

```ts
/// <reference types="mongoose/esnext" />
```

One reason we haven't done `libs` yet is that we already support two dimensions: multiple versions of TS and multiple versions of the source package. A third dimension would be even more confusing.

sheetalkamat (changes requested):
Needs changes per discussion here

Unibozu:
I will send an updated version as soon as I find time to do it! thanks all for your comments

PR #42454, opened 2020-02-18
Add definitions for apple-signin-api

voidpumpkin:
Adding definitions for [apple-signin-api](https://developer.apple.com/documentation/signinwithapplejs)

Apple documentation is actually not completely in sync with all the data we have, so here are some pictures objects in use:
image
image
image

Please fill in this template.

If adding a new definition:

### DIFF
--- a/types/apple-signin-api/apple-signin-api-tests.ts
+++ b/types/apple-signin-api/apple-signin-api-tests.ts
@@ -6,7 +6,7 @@ const ClientConfig: AppleSignInAPI.ClientConfigI = {
     usePopup: false,
 };
 
-const signinResponse: AppleSignInAPI.signInResponse = {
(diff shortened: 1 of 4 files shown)


PranavSenthilnathan:

koenpunt:
Thanks for this, but how am I supposed to use this?

koenpunt:
I've now added a custom declaration, but I think it's supposed to work straight away?

```ts
declare var AppleID: AppleSignInAPI.AppleID;
```

voidpumpkin:
@koenpunt It seems you are using it perfectly with `AppleSignInAPI.AppleID` since `AppleSignInAPI` namespace is just available globally.

Unless im being a noob and i was supposed to do this in some other way.

In my use case we just add to the Window interface `AppleID: AppleSignInAPI.AppleID;`.

PR #42457, opened 2020-02-18
[codemirror] Add addon display/autorefresh

mtgto:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://codemirror.net/doc/manual.html#addon_autorefresh

### DIFF
--- a/types/codemirror/addon/display/autorefresh.d.ts
+++ b/types/codemirror/addon/display/autorefresh.d.ts
@@ -1,0 +1,16 @@
+// Type definitions for CodeMirror
+// Project: https://github.com/marijnh/CodeMirror
+// Definitions by: mtgto <https://github.com/mtgto>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+
+// See docs https://codemirror.net/doc/manual.html#addon_autorefresh
+
+import * as CodeMirror from "codemirror";
+
+declare module "codemirror" {
+    interface EditorConfiguration {
+        // if true, it will be refreshed the first time the editor becomes visible.
+        // you can pass delay (msec) time as polling duration
+        autoRefresh?: boolean | { delay: number };
+    }
+}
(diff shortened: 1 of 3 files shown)


types/codemirror/addon/display/autorefresh.d.ts
```diff
@@ -0,0 +1,16 @@
...
+
+// See docs https://codemirror.net/doc/manual.html#addon_autorefresh
+
+import * as CodeMirror from "codemirror";
+
+declare module "codemirror" {
+    interface EditorConfiguration {
+        // if true, it will be refreshed the first time the editor becomes visible.
+        // you can pass delay (msec) time as polling duration
+        autoRefresh?: boolean | { delay: number };
```
azoson: Can we make this property nullable? 🤔 
[This source](https://codemirror.net/addon/display/autorefresh.js) assigns `null` to the `cm.state.autoRefresh` like following: 

```javascript
    if (cm.state.autoRefresh) {
      stopListening(cm, cm.state.autoRefresh)
      cm.state.autoRefresh = null
    } 
```

Do you know whether is this assignment allowed to addon users?

azoson (changes requested):

types/codemirror/addon/display/autorefresh.d.ts
mtgto: [Document](https://codemirror.net/doc/manual.html#addon_autorefresh) says autoRefresh is optional param either `true` or `{ delay: xxx (msec) }` object.
Implementation of autorefresh addons, you notice, is nullable.
IMO, We should not be nullable because doc does not mention it can be null.

mtgto: @azoson My opinion is non-nullable.
But I don't have confident.
Can you choose nullable or non-nullable?

mtgto:
@azoson My opinion is non-nullable.
But I don't have confident.
Can you choose nullable or non-nullable?

types/codemirror/addon/display/autorefresh.d.ts
mtgto: hmm... @typescript-bot requests to me "new comment or commit", but nothing changes after I commented.

PR #42468, opened 2020-02-19
add(mustache): better mustache.parse return type

sbdchd:
Currently mustache.parse has `any` as its return type which is more
permissive than necessary.

I've ran though the mustache docs to create some test cases for the return
types:

https://mustache.github.io/mustache.5.html

and the test cases:

https://runkit.com/sbdchd/5e4bf8f70a83de0014e082ed

I also did a partial conversion of mustache.mjs to typescript to figure
out the return types:

https://gist.github.com/sbdchd/449c409648fd4c5120386bbc3fc683f2#file-mustache-ts-L113-L277

So I think this return types should be accurate.

Note that since this is a recursive type we need to set the minimum
version in the index.d.ts header otherwise the linter complains.

---

Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/janl/mustache.js/blob/aca97b82c80e8fd1d36162e05e4b289380965d96/mustache.mjs#L83-L109

### DIFF
--- a/types/mustache/index.d.ts
+++ b/types/mustache/index.d.ts
@@ -3,7 +3,9 @@
 // Definitions by: Mark Ashley Bell <https://github.com/markashleybell>,
 //                 Manuel Thalmann <https://github.com/manuth>,
 //                 Phillip Johnsen <https://github.com/phillipj>
+//                 Steve Dignam <https://github.com/sbdchd>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// Minimum TypeScript Version: 3.7
 
 /**
  * Provides the functionality to render templates with `{{mustaches}}`.
@@ -81,7 +83,7 @@ interface MustacheStatic {
      * @param tags
      * The tags to use.
      */
-    parse(template: string, tags?: OpeningAndClosingTags): any;
+    parse(template: string, tags?: OpeningAndClosingTags): TemplateSpans;
 
     /**
      * Renders the `template` with the given `view` and `partials` using the default writer.
@@ -337,6 +339,14 @@ declare class MustacheWriter {
      */
     rawValue(token: string[]): string;
 }
+
+type TemplateSpanType = 'text' | 'name' | '#' | '&' | '^' | '!' | '>' | '=';
+
+type TemplateSpans = Array<
+    | [TemplateSpanType, string, number, number]
+    | [TemplateSpanType, string, number, number, TemplateSpans, number]
+    | [TemplateSpanType, string, number, number, string, number, boolean]
+>;
 
 /**
  * An array of two strings, representing the opening and closing tags respectively, to be used in the templates being rendered.
--- a/types/mustache/index.d.ts
+++ b/types/mustache/index.d.ts
@@ -340,7 +340,24 @@ declare class MustacheWriter {
     rawValue(token: string[]): string;
 }
 
-type TemplateSpanType = 'text' | 'name' | '#' | '&' | '^' | '!' | '>' | '=';
+type RAW_VALUE = "text";
+type ESCAPED_VALUE = "name";
+type UNESCAPED_VALUE = "&";
+type SECTION = "#";
+type INVERTED = "^";
+type COMMENT = "!";
+type PARTIAL = ">";
+type EQUAL = "=";
+
+type TemplateSpanType =
+    | RAW_VALUE
+    | ESCAPED_VALUE
+    | SECTION
+    | UNESCAPED_VALUE
+    | INVERTED
+    | COMMENT
+    | PARTIAL
+    | EQUAL;
 
 type TemplateSpans = Array<
     | [TemplateSpanType, string, number, number]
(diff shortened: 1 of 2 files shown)


sbdchd:
Hmm, `node_vault` has a lower required typescript version and depends on mustache

https://dev.azure.com/definitelytyped/DefinitelyTyped/_build/results?buildId=35410&view=logs&j=8635df7d-ecde-52a7-c7a8-84b998e35690&t=c264d665-0b79-59b7-7e04-9965cbe74860

sbdchd:
Also I think changing the return type from `any` to a more concrete type is a breaking change so I'm wondering if the version should be bumped or kept the same as the source code

types/mustache/index.d.ts
```diff
@@ -338,6 +340,14 @@ declare class MustacheWriter {
     rawValue(token: string[]): string;
 }
 
+type TemplateSpanType = 'text' | 'name' | '#' | '&' | '^' | '!' | '>' | '=';
```
phillipj: Would it make sense to turn these string literals into type aliases or an enum that uses the "domain language" of mustache (`section | inverted | partial` etc..)? That would make sense when seeing the below in the source code at least.

```js
if (symbol === '#') value = this.renderSection(token, context, partials, originalTemplate);
else if (symbol === '^') value = this.renderInverted(token, context, partials, originalTemplate);
else if (symbol === '>') value = this.renderPartial(token, context, partials, tags);
else if (symbol === '&') value = this.unescapedValue(token, context);
else if (symbol === 'name') value = this.escapedValue(token, context);
else if (symbol === 'text') value = this.rawValue(token);
```
[janl/mustache.js mustache.mjs#L560-565](https://github.com/janl/mustache.js/blob/aca97b82c80e8fd1d36162e05e4b289380965d96/mustache.mjs#L560-L565)

phillipj:
Impressive work you've done to ensure all edge cases are covered 👍

I'm not too familiar with the versioning of typedefs to be honest. So far I've had the understanding that the `@types/xyz` package should be versioned as close the the `xyz` package itself. This have made sense to me as a `@types/` user myself lately.

It would also surprise me if there are a great number of projects using mustache.js, that use `.parse()` at all. Throughout the years it has been a hidden trick to pre-parse the templates, ensuring the subsequent calls to `.render()` would perform as best as possible. But still those use cases, doesn't actually care about the output of `.parse()` 🤷‍♂

types/mustache/index.d.ts
sbdchd: `enum` didn't work because the function returns string literals instead of enum values, but we can add aliases for the variants of the type.

```typescript
type RAW_VALUE = "text"
type ESCAPED_VALUE = "name"
type UNESCAPED_VALUE = "&"
type SECTION = "#"
type INVERTED = "^"
type COMMENT = "!"
type PARTIAL = ">"
type EQUAL = "="

type TemplateSpanType =
  | RAW_VALUE
  | ESCAPED_VALUE
  | SECTION
  | UNESCAPED_VALUE
  | INVERTED
  | COMMENT
  | PARTIAL
  | EQUAL

type TemplateSpans = Array<
  | [TemplateSpanType, string, number, number]
  | [TemplateSpanType, string, number, number, TemplateSpans, number]
  | [TemplateSpanType, string, number, number, string, number, boolean]
>
```

phillipj: Nice 👍

phillipj (approved):

Tr1gs:
When I transpile (typescript: 3.8.3 @types/mustache: 4.0.1) with this change I get a circular reference error:

```
node_modules/@types/mustache/index.d.ts:362:6 - error TS2456: Type alias 'TemplateSpans' circularly references itself.

362 type TemplateSpans = Array<
```
which is at

```
type TemplateSpans = Array<
    | [TemplateSpanType, string, number, number]
    | [TemplateSpanType, string, number, number, TemplateSpans, number]
    | [TemplateSpanType, string, number, number, string, number, boolean]
>;
```

sbdchd:
@Tr1gs working for me on typescript playground with 3.8.3

Are you sure you're using the typescript >=3.7?

```typescript
type TemplateSpans = Array<
    | [string, string, number, number]
    | [string, string, number, number, TemplateSpans, number]
    | [string, string, number, number, string, number, boolean]
>;
```

[Playground Link](https://www.typescriptlang.org/play/index.html#code/C4TwDgpgBAKhC2YA2BDYEDKYUDsDOUAvFAIIBOZKIAPALABQUTUAPlANp7BkCWOA5gBooXXgOE4ArvABGEMhOlyyAXQbNWHUXyEjuOxbPmHlwuIlTosuPCflrGzNp33i9Y3VKMKoX0+4NfJWMoGQB7MKQIXAcAPgBuIA)

Tr1gs:
Yeah, I'm definitely using typescript 3.8.3, before that I was using 3.7.5 with @types/mustache 4.0.0

Tr1gs:
Apologies, I found the issue.  I was transpiling with the global typescript which was pre 3.7 when I ran `./node_modules/typescript/bin/tsc --build tsconfig.prod.json` (3.8.3) it worked fine.  Sorry for the confusion

PR #42476, opened 2020-02-19
Added method definition for new method in SDK

kilinkis:
The method can be found in the official documentation under "Removing a Discount": https://shopify.github.io/js-buy-sdk/

[ ] The package does not already provide its own types, or cannot have its .d.ts files generated via --declaration


straiforos (changes requested):
Looks good thank you for adding the functionality and updating the version. If you would not mind adding yourself to the authors list that would be greatly appreciated.

kilinkis:
hi and thanks @straiforos
I'm already added in the "Definitions by: ", is that what you mean? If not, how do I add myself as an author? 🤔Also, the PR is tagged as "author is owner".

straiforos:
Hi @kilinkis sorry about that. Yes that is what I meant. Github condensed the definitions by so I didn't see it. I'll approve

straiforos (approved):

types/shopify-buy/index.d.ts
```diff
@@ -1,4 +1,4 @@
-// Type definitions for shopify-buy 1.4.0
+// Type definitions for shopify-buy 1.11.0 (WIP)
```
PranavSenthilnathan: Could you turn on the `dt-header` lint rule in the `tslint.json` file? This version should just be major and minor (this is automatically extracted when publishing to npm so having the correct format will prevent any issues).

PranavSenthilnathan (changes requested):

types/shopify-buy/index.d.ts
straiforos: @PranavSenthilnathan great catch! Sorry I didn't see this. I should have been more thorough.

kilinkis: Just updated that version reference here :) good catch @PranavSenthilnathan https://github.com/DefinitelyTyped/DefinitelyTyped/pull/42476/files#diff-0cdb71e33d6ce9abc56e43ec7419ae4a

sheetalkamat (changes requested):
Please enable "dt-header": true in tslint.json of this package since you are changing the version

kilinkis:
done @sheetalkamat :)

kilinkis:
error says:

```Error: /home/travis/build/DefinitelyTyped/DefinitelyTyped/types/shopify-buy/index.d.ts:1:25
ERROR: 1:25  dt-header               
Error parsing header. Expected: foo MAJOR.MINOR (patch version not allowed). 
See: https://github.com/Microsoft/dtslint/blob/master/docs/dt-header.md
ERROR: 92:1  no-trailing-whitespace  trailing whitespace```

In the  changes I did, there is no trailing space and the version changed from `1.4.0` to `1.11.0`. Not sure what the problem is here. Is it the double middle digits? 🤔

**Comment by sheetalkamat** [comment]
You need it to be `1.11` instead

**Comment by kilinkis** [comment]
@sheetalkamat sadly, that didn't work neither :(

**Comment by sheetalkamat** [comment]
It isn't complaining about the version any more.. 
```

Error in shopify-buy
301
Error: /home/travis/build/DefinitelyTyped/DefinitelyTyped/types/shopify-buy/index.d.ts:92:1
302
ERROR: 92:1  no-trailing-whitespace  trailing whitespace
```

kilinkis:
all done @sheetalkamat, sorry I didn't see that error before. Thanks for your help!

PR #42483, opened 2020-02-19
Fix js-cookie types

track0x1:
Calls to `set()` return a string or undefined, not void.

  - https://github.com/js-cookie/js-cookie/blob/v2.2.1/src/js.cookie.js#L48
  - https://github.com/js-cookie/js-cookie/blob/v2.2.1/src/js.cookie.js#L98


peterblazejewicz (approved):
Nicely spotted, the docs are missing this detail 👍 
@track0x1 thanks!
LGTM!

PR #42485, opened 2020-02-19
Improve type definitions for package shuffle-array

jwulf0:

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/pazguille/shuffle-array/blob/master/index.js you can see that the `pick` function returns a single element or an array
- [x] If this PR brings the type definitions up to date with a new version of the JS library, update the version number in the header. Set to 1.0.1; did not have any version number previously.

=====

It's 3 changes, the most importent is the return type of the `pick`-method (which is wrong in the previous version; might be due to a change in `shuffle-array`, idk). Also, I updated the parameter type to `ReadonlyArray<T>` as is encouraged in the readme. Last, I simply changed names of option-interfaces from having the suffix `Option` to `Options` as that seems to be a better fit intuitively; but I can revert that if such changes are discouraged.

This is my first contribution to `DefinetelyTyped`, please let me know if I missed something in the contribution guidelines.

### DIFF
--- a/types/shuffle-array/index.d.ts
+++ b/types/shuffle-array/index.d.ts
@@ -36,7 +36,7 @@ interface ShuffleArray {
      * arr - The given array.
      * options - Optional configuration options.
      */
-    pick<T>(arr: ReadonlyArray<T>, options?: PickOptions): T | T[];
+    pick<T>(arr: ReadonlyArray<T>, options?: PickOptions): T | T[];
 }
 declare var shuffle: ShuffleArray;
 export = shuffle;
--- a/types/shuffle-array/index.d.ts
+++ b/types/shuffle-array/index.d.ts
@@ -1,4 +1,4 @@
-// Type definitions for shuffle-array
+// Type definitions for shuffle-array 1.0.1
 // Project: https://github.com/pazguille/shuffle-array
 // Definitions by: rhysd <https://rhysd.github.io>
 //                 jwulf0 <https://github.com/jwulf0>
--- a/types/shuffle-array/index.d.ts
+++ b/types/shuffle-array/index.d.ts
@@ -1,6 +1,6 @@
-// Type definitions for shuffle-array 1.0.1
+// Type definitions for shuffle-array 1.0
 // Project: https://github.com/pazguille/shuffle-array
-// Definitions by: rhysd <https://rhysd.github.io>
+// Definitions by: rhysd <https://github.com/rhysd>
 //                 jwulf0 <https://github.com/jwulf0>
 // Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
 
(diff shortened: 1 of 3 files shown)


types/shuffle-array/index.d.ts
```diff
@@ -1,22 +1,23 @@
-// Type definitions for shuffle-array
+// Type definitions for shuffle-array 1.0.1
```
PranavSenthilnathan: This should only be major and minor. If you enable the lint rule in dtlint.json it should catch this (`dt-header` IIRC).

PranavSenthilnathan (changes requested):

types/shuffle-array/index.d.ts
jwulf0: Thank you for your feedback!

I updated the version info to `1.0´.

About the `dt-header` rule, I saw that it's disabled in shuffle-array's tslint.json. I enabled it and had to change the first contributor's URL (the dt-header rule seems to allow only github URLs in the header). I would say that's fine since it is evidently the correct github profile and a link to their personal website can be found there.

jwulf0:
Fixed the merge conflict (sorry for the multiple commits, was confused there for a second) and the whitespace error.

jwulf0:
@PranavSenthilnathan sorry to bother but did you get a chance to check the changes?

types/shuffle-array/index.d.ts
```diff
@@ -27,14 +28,15 @@ interface ShuffleArray {
...
     /**
-     * Pick one or more random elements from the given array.
+     * Pick one or more random elements from the given array. If options.picks is
+     * omitted or === 1, a single element will be returned; otherwise an array.
      *
      * arr - The given array.
      * options - Optional configuration options.
      */
-    pick<T>(arr: T[], options?: Object): T[];
+    pick<T>(arr: ReadonlyArray<T>, options?: PickOptions): T | T[];
```
andrewbranch, line 39: This could probably be written as a series of overloads since it’s often statically analyzable what `option.picks` is:

```suggestion
    pick<T>(arr: readonly T[], options: PickOptions & { picks: 1 }): T;
    pick<T>(arr: readonly T[], options: PickOptions): T | T[];
    pick<T>(arr: readonly T[]): T;
```

It would need some tests. [Example](https://www.typescriptlang.org/play/#code/JYOwLgpgTgZghgYwgAgArAQawPIAczAD2IAzsgN4BQyNyuGmJA-AFzIgCuAtgEbQDclAL6VKAEwgIANnCgoYHEAgLE6DADwAVAHwAKWVDZy4Y4lICeyTQG0AugBpkhfEVJt0WPCtLIAZBTUsEjYARmQhAEo2TUEJaVl5RWVXQMwtPQMjCBMzSxsHJxdiYLQGL1cSKKtkAB8rO1jJGTlkBSVvVPT9KENkY1MQC3rbKpjxJoTkADdZZEz2bj4oBtEEYrAnEAgQtk5eaGQAXlTuqEdyVJKwyME10g3iCAAmXcWD4-osU4jb9eRFTAgQgAdxAr32UFqCwhdiOJwM50ubAAsnAwAALAB0UDgIFMXF0EXCP0oQA)

andrewbranch (approved):

orta (approved):

PR #42487, opened 2020-02-19
update customKeys and extraKeys attribute for showHintOptions

BlueSpider929:
@rohit-gohri , this [commit](https://github.com/DefinitelyTyped/DefinitelyTyped/commit/d66b9a5834cef7f5d2b9e98208d03bf71fb0fc1e) is really exciting and very helpful to use codemirror `hint` addon within typescript projects.
However, I get into some trouble using `customKeys` and `extraKeys` attributes because we can't pass any arguments for the handler function as it's declared as `() => void`.
Please refer to `customKeys` and `extraKeys` implementation on codemirror [here](https://github.com/codemirror/CodeMirror/blob/5f722737883bbeadb2a629857d3961490e4134ca/addon/hint/show-hint.js#L183).
Actually, when we are defining `customKeys` and `extraKeys`, its purpose is to write custom handlers for specific keys, so the arguments `editor: Editor` and `handler: Handler` is mandatory.
It'd be helpful if we update these two attributes `customKeys` and `extraKeys` with
```
customKeys?: {[x: string]: (editor: cm.Editor, handle: cm.Handle) => void} | null;
extraKeys?: {[x: string]: (editor: cm.Editor, handle: cm.Handle) => void} | null;
```
instead of 
```
customKeys?: {[x: string]: () => void} | null;
extraKeys?: {[x: string]: () => void} | null;
```
Thank you for your attention.

### DIFF
--- a/types/codemirror/addon/hint/show-hint.d.ts
+++ b/types/codemirror/addon/hint/show-hint.d.ts
@@ -61,8 +61,8 @@ declare module "codemirror" {
         closeOnUnfocus?: boolean;
         completeOnSingleClick?: boolean;
         container?: HTMLElement | null;
-        customKeys?: {[x: string]: () => void} | null;
-        extraKeys?: {[x: string]: () => void} | null;
+        customKeys?: {[key: string]: (editor: Editor, handle: Handle) => void} | null;
+        extraKeys?: {[key: string]: (editor: Editor, handle: Handle) => void} | null;
     }
 
     /** The Handle used to interact with the autocomplete dialog box.*/
--- a/types/codemirror/addon/hint/show-hint.d.ts
+++ b/types/codemirror/addon/hint/show-hint.d.ts
@@ -61,8 +61,8 @@ declare module "codemirror" {
         closeOnUnfocus?: boolean;
         completeOnSingleClick?: boolean;
         container?: HTMLElement | null;
-        customKeys?: {[key: string]: (editor: Editor, handle: Handle) => void} | null;
-        extraKeys?: {[key: string]: (editor: Editor, handle: Handle) => void} | null;
+        customKeys?: {[key: string]: (editor: Editor, handle: Handle) => void | string} | null;
+        extraKeys?: {[key: string]: (editor: Editor, handle: Handle) => void | string} | null;
     }
 
     /** The Handle used to interact with the autocomplete dialog box.*/
--- a/types/codemirror/addon/hint/show-hint.d.ts
+++ b/types/codemirror/addon/hint/show-hint.d.ts
@@ -62,7 +62,7 @@ declare module "codemirror" {
         completeOnSingleClick?: boolean;
         container?: HTMLElement | null;
         customKeys?: {[key: string]: (editor: Editor, handle: Handle) => void | string} | null;
-        extraKeys?: {[key: string]: (editor: Editor, handle: Handle) => void | string} | null;
+        extraKeys?: {[key: string]: ((editor: Editor, handle: Handle) => void) | string} | null;
     }
 
     /** The Handle used to interact with the autocomplete dialog box.*/
--- a/types/codemirror/addon/hint/show-hint.d.ts
+++ b/types/codemirror/addon/hint/show-hint.d.ts
@@ -61,7 +61,7 @@ declare module "codemirror" {
         closeOnUnfocus?: boolean;
         completeOnSingleClick?: boolean;
         container?: HTMLElement | null;
-        customKeys?: {[key: string]: (editor: Editor, handle: Handle) => void | string} | null;
+        customKeys?: {[key: string]: ((editor: Editor, handle: Handle) => void) | string} | null;
         extraKeys?: {[key: string]: ((editor: Editor, handle: Handle) => void) | string} | null;
     }
 


types/codemirror/addon/hint/show-hint.d.ts
```diff
@@ -61,8 +61,8 @@ declare module "codemirror" {
         closeOnUnfocus?: boolean;
         completeOnSingleClick?: boolean;
         container?: HTMLElement | null;
-        customKeys?: {[x: string]: () => void} | null;
-        extraKeys?: {[x: string]: () => void} | null;
+        customKeys?: {[key: string]: (editor: Editor, handle: Handle) => void} | null;
+        extraKeys?: {[key: string]: (editor: Editor, handle: Handle) => void} | null;
```
azoson: Thanks for refinement the type 🙏 
I wonder whether the `extraKeys` properties can have `string` or `false` values.
And I wonder whether the functions can have the second argument.
Do you have any thoughts about these?

I found [the document](https://codemirror.net/doc/manual.html#option_extraKeys) says that `extraKeys` should be `a valid key map value`.
And this document says `The values of properties in key maps can be either functions of a single argument (the CodeMirror instance), strings, or false`.
Thus, I think we shold define `extraKeys` types as following:
```typescript
extraKeys?: { [key: string]: (editor: Editor) => void | string | false | null }
```
Is it correct? 🤔

azoson (changes requested):

types/codemirror/addon/hint/show-hint.d.ts
BlueSpider929: @azoson , Thanks for reviewing my PR.
I think [the document](https://codemirror.net/doc/manual.html#option_extraKeys) is describing the `extraKeys` for `codemirror` instance itself. The `extraKeys` we are refining is for `hint` addon of `codemirror`. Please refer to [this document](https://codemirror.net/doc/manual.html#addon_show-hint). It has its own `extraKeys` option.

> customKeys: keymap
> Allows you to provide a custom key map of keys to be active when the pop-up is active. The handlers will be called with an extra argument, a handle to the completion menu, which has moveFocus(n), setFocus(n), pick(), and close() methods (see the source for details), that can be used to change the focused element, pick the current element or close the menu. Additionally menuSize() can give you access to the size of the current dropdown menu, length give you the number of available completions, and data give you full access to the completion returned by the hinting function.
> extraKeys: keymap
> Like customKeys above, but the bindings will be added to the set of default bindings, instead of replacing them.

And its implementation is found [here](https://github.com/codemirror/CodeMirror/blob/5f722737883bbeadb2a629857d3961490e4134ca/addon/hint/show-hint.js#L183). 
I hope this helps you to understand

BlueSpider929:
@azoson , Thanks for reviewing my PR.
I think [the document](https://codemirror.net/doc/manual.html#option_extraKeys) is describing the `extraKeys` for `codemirror` instance itself. The `extraKeys` we are refining is for `hint` addon of `codemirror`. Please refer to [this document](https://codemirror.net/doc/manual.html#addon_show-hint). It has its own `extraKeys` option.

> customKeys: keymap
> Allows you to provide a custom key map of keys to be active when the pop-up is active. The handlers will be called with an extra argument, a handle to the completion menu, which has moveFocus(n), setFocus(n), pick(), and close() methods (see the source for details), that can be used to change the focused element, pick the current element or close the menu. Additionally menuSize() can give you access to the size of the current dropdown menu, length give you the number of available completions, and data give you full access to the completion returned by the hinting function.
> extraKeys: keymap
> Like customKeys above, but the bindings will be added to the set of default bindings, instead of replacing them.

And its implementation is found [here](https://github.com/codemirror/CodeMirror/blob/5f722737883bbeadb2a629857d3961490e4134ca/addon/hint/show-hint.js#L183). 
I hope this helps you to understand

types/codemirror/addon/hint/show-hint.d.ts
BlueSpider929: @azoson , Thanks for reviewing my PR.
I think [the document](https://codemirror.net/doc/manual.html#option_extraKeys) is describing the `extraKeys` for `codemirror` instance itself. The `extraKeys` we are refining is for `hint` addon of `codemirror`. Please refer to [this document](https://codemirror.net/doc/manual.html#addon_show-hint). It has its own `extraKeys` option.

> customKeys: keymap
> Allows you to provide a custom key map of keys to be active when the pop-up is active. The handlers will be called with an extra argument, a handle to the completion menu, which has moveFocus(n), setFocus(n), pick(), and close() methods (see the source for details), that can be used to change the focused element, pick the current element or close the menu. Additionally menuSize() can give you access to the size of the current dropdown menu, length give you the number of available completions, and data give you full access to the completion returned by the hinting function.
> extraKeys: keymap
> Like customKeys above, but the bindings will be added to the set of default bindings, instead of replacing them.

And its implementation is found [here](https://github.com/codemirror/CodeMirror/blob/5f722737883bbeadb2a629857d3961490e4134ca/addon/hint/show-hint.js#L183). 
I hope this helps you to understand

BlueSpider929 (commented):
@azoson , Thanks for reviewing my PR.
I think [the document](https://codemirror.net/doc/manual.html#option_extraKeys) is describing the `extraKeys` for `codemirror` instance itself. The `extraKeys` we are refining is for `hint` addon of `codemirror`. Please refer to [this document](https://codemirror.net/doc/manual.html#addon_show-hint). It has its own `extraKeys` option.

> customKeys: keymap
> Allows you to provide a custom key map of keys to be active when the pop-up is active. The handlers will be called with an extra argument, a handle to the completion menu, which has moveFocus(n), setFocus(n), pick(), and close() methods (see the source for details), that can be used to change the focused element, pick the current element or close the menu. Additionally menuSize() can give you access to the size of the current dropdown menu, length give you the number of available completions, and data give you full access to the completion returned by the hinting function.
> extraKeys: keymap
> Like customKeys above, but the bindings will be added to the set of default bindings, instead of replacing them.

And its implementation is found [here](https://github.com/codemirror/CodeMirror/blob/5f722737883bbeadb2a629857d3961490e4134ca/addon/hint/show-hint.js#L183). 
I hope this helps you to understand

azoson (approved):
Thanks!
I understand codemirror's `extraKeys` is not the addon's `extraKeys`.
Now I understand that the `(editor: Editor, handle: Handle) => void}` is valid, so approve!
Thanks for your great contribution and PR conversation 🙏

types/codemirror/addon/hint/show-hint.d.ts
BlueSpider929: @azoson , Thanks for reviewing my PR.
I think [the document](https://codemirror.net/doc/manual.html#option_extraKeys) is describing the `extraKeys` for `codemirror` instance itself. The `extraKeys` we are refining is for `hint` addon of `codemirror`. Please refer to [this document](https://codemirror.net/doc/manual.html#addon_show-hint). It has its own `extraKeys` option.

> customKeys: keymap
> Allows you to provide a custom key map of keys to be active when the pop-up is active. The handlers will be called with an extra argument, a handle to the completion menu, which has moveFocus(n), setFocus(n), pick(), and close() methods (see the source for details), that can be used to change the focused element, pick the current element or close the menu. Additionally menuSize() can give you access to the size of the current dropdown menu, length give you the number of available completions, and data give you full access to the completion returned by the hinting function.
> extraKeys: keymap
> Like customKeys above, but the bindings will be added to the set of default bindings, instead of replacing them.

And its implementation is found [here](https://github.com/codemirror/CodeMirror/blob/5f722737883bbeadb2a629857d3961490e4134ca/addon/hint/show-hint.js#L183). 
I hope this helps you to understand

BlueSpider929:
@azoson, very good point.
I think I missed that.
While `this mechanism is deprecated`, `codemirror hint` addon did not cross out that code snippet, so I think we need to support that.
Let me update the interface and will push again.
Thanks.

types/codemirror/addon/hint/show-hint.d.ts
```diff
@@ -61,8 +61,8 @@ declare module "codemirror" {
         closeOnUnfocus?: boolean;
         completeOnSingleClick?: boolean;
         container?: HTMLElement | null;
-        customKeys?: {[key: string]: (editor: Editor, handle: Handle) => void} | null;
-        extraKeys?: {[key: string]: (editor: Editor, handle: Handle) => void} | null;
+        customKeys?: {[key: string]: (editor: Editor, handle: Handle) => void | string} | null;
+        extraKeys?: {[key: string]: (editor: Editor, handle: Handle) => void | string} | null;
```
azoson:
```suggestion
        extraKeys?: {[key: string]: ((editor: Editor, handle: Handle) => void) | string} | null;
```

I find that this type is inferred as a function returning 'string' or 'void'.
`((editor: Editor, handle: Handle) => void) | string` is suitable, I think. 🙇 

https://www.typescriptlang.org/play/#code/C4TwDgpgBAKlC8UDeAoK6oG0DWEQC4oBnYAJwEsA7AcwF1CAKASgQD4oA3Ae3IBMoAPsTJVqKAL6ColAK4AbOQG4UAYy6USUAGZcuhOIlQYoAIwCGpQgHISFGlYA0ExUA

types/codemirror/addon/hint/show-hint.d.ts
```diff
@@ -61,8 +61,8 @@ declare module "codemirror" {
         closeOnUnfocus?: boolean;
         completeOnSingleClick?: boolean;
         container?: HTMLElement | null;
-        customKeys?: {[key: string]: (editor: Editor, handle: Handle) => void} | null;
-        extraKeys?: {[key: string]: (editor: Editor, handle: Handle) => void} | null;
+        customKeys?: {[key: string]: (editor: Editor, handle: Handle) => void | string} | null;
```
azoson:
```suggestion
        customKeys?: {[key: string]: ((editor: Editor, handle: Handle) => void) | string} | null;
```

azoson (changes requested):

BlueSpider929:
@azoson , I approved your suggestion regarding types of the `extraKeys` and `customKeys` and committed the changes.

azoson (approved):
sorry for late.
and thanks for the update 🙏 
LGTM! 👍

PR #42489, opened 2020-02-19
[reactstrap]: Export Util.setGlobalCssModule

kyletsang:
Please fill in this template.

- [x] Provide a URL to documentation or source code which provides context for the suggested changes: https://github.com/reactstrap/reactstrap/issues/1789

As per discussion in https://github.com/reactstrap/reactstrap/issues/1789, `setGlobalCssModule` should be exposed.  As for the others, I believe it should be added on a case by case basis?


types/reactstrap/index.d.ts
```diff
@@ -193,3 +193,5 @@ export {
   UncontrolledCarouselProps,
   UncontrolledPopoverProps
 } from './lib/Uncontrolled';
+import * as Util from './lib/Util';
```
PranavSenthilnathan: `export { Util } from './lib/Util':` to match the style in the file?

kyletsang: reactstrap exports the functions inside Util with `export * as Util from './utils'` but it seems this is not available until Typescript 3.8. 

Would it be better to put the functions in Util under a namespace and export that?

Util.d.ts
```
declare namespace Util {
    export function setGlobalCssModule(cssModule: CSSModule): void;
}

export default Util;
```

index.d.ts
```
export { default as Util } from './lib/Util';
```

sheetalkamat: This looks ok

types/reactstrap/lib/Util.d.ts
```diff
@@ -0,0 +1,3 @@
+import { CSSModule } from '../index';
```
sheetalkamat: Can you please name this file utils.d.ts instead to match the lib structure in the package.

sheetalkamat (changes requested):

kyletsang:
I've made the file name change. Thanks

PR #42493, opened 2020-02-19
Use generic to allow use of a class as view model. Knockout

ReyPena:
Please fill in this template.

If adding a new definition:

If removing a declaration:

### DIFF
--- a/types/knockout/index.d.ts
+++ b/types/knockout/index.d.ts
@@ -947,7 +947,7 @@ interface KnockoutComputedContext {
 // and used Union Types to simplify overloads (requires TypeScript 1.4)
 //
 declare namespace KnockoutComponentTypes {
-    type viewModel = ViewModelFunction | ViewModelSharedInstance | ViewModelFactoryFunction | AMDModule;
+    type ViewModel = ViewModelFunction | ViewModelSharedInstance | ViewModelFactoryFunction | AMDModule;
     
     interface Config<T> {
         viewModel?: T;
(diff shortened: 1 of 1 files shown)


ffMathy:
I would PascalCase ViewModel.

types/knockout/index.d.ts
```diff
@@ -946,15 +947,16 @@ interface KnockoutComputedContext {
 // and used Union Types to simplify overloads (requires TypeScript 1.4)
 //
 declare namespace KnockoutComponentTypes {
-
-    interface Config {
-        viewModel?: ViewModelFunction | ViewModelSharedInstance | ViewModelFactoryFunction | AMDModule;
+    type ViewModel = ViewModelFunction | ViewModelSharedInstance | ViewModelFactoryFunction | AMDModule;
+    
+    interface Config<T> {
```
Retsam: Should there be some restriction on `T` here?  It seems weird that you could have something like `viewModel: string`, but maybe that is reasonable?  (I haven't actually used KO's component APIs, myself)

types/knockout/index.d.ts
```diff
@@ -946,15 +947,16 @@ interface KnockoutComputedContext {
...
+        viewModel?: T;
         template: string | Node[] | DocumentFragment | TemplateElement | AMDModule;
         synchronous?: boolean;
     }
 
-    interface ComponentConfig {
-        viewModel?: ViewModelFunction | ViewModelSharedInstance | ViewModelFactoryFunction | AMDModule;
+    interface ComponentConfig<T = ViewModel> {
+        viewModel?: ViewModel;
```
Retsam: Should be T, right?

ReyPena: Thank you! good catch

Retsam (approved):

types/knockout/index.d.ts
ReyPena: I couldn't find a better way to do this, and this will allow specifying the class allowing to be more strict.
this is the is idea whit this change:
`
components.register<ComponentClass>('component-name', {
    viewModel: ComponentClass,
    template: require('./containers/units-set-list/units-set-list-component.html')
});
`

